PR #39, opened 2018-01-30
dotnet core upgrade

CloneDeath:
I've upgraded the two projects to dotnet core (using netstandard2.0), and updated the csproj files to the new format. I moved the demo project into its own directory, simply because of the way the new csproj file format works... otherwise I would have had to make the project exclude everything and only bring in demo.cs, making it a lot more complex.

In order to get around the mono dependency, I've added a nuget reference to the beta netstandard posix nuget package, which looks to be working fine.

I've added dlerror in order to help diagnose issues.

Also, when lazy loading ncurses on my platform, I always got intptr.zero. But, when I changed it to load now (1 instead of 0 for the mode), it ran perfectly for me.

Lastly, i changed the dllimport bindings from libc to libdl. Again, I had issues binding to libc, and I'm not entirely sure why. (libdl just wraps libc for the dl calls, btw). If you have a problem with this (it doesn't run), then I can go ahead and create two bindings, one for libc, and one for libdl, and try one before falling back to the other. I don't know if this is necessary, so I held back on doing this to keep my changes minimal for this pull request.


tom-corwin:
Have you thought about adding `<EnableDefaultCompileItems>false</EnableDefaultCompileItems>` to the demo project, and add the line `<Compile Exclude="demo.cs" />` to the main project? This would avoid reorganizing the whole project, and still have the same result. Just a thought, since I as looking at this project. :P

CloneDeath:
Yes, I considered that. And ultimately, the change I made only involved moving two files: demo.cs and Terminal.csproj

Most projects usually are separated with one project per folder, and not mixed with the main solution... I'm not sure why this was an exception.

I'm not against doing exactly as you said though, if for no other reason than to keep this pull request focused solely on the dotnet core upgrade.

tom-corwin:
I wasn't against what you did, as I prefer a single project per folder, I was just mentioning it. :)

migueldeicaza:
I applied the changes to bindings.cs to the upstream project (mono-curses), which is regularly synced here.

The only downside of going to .NET Standard is that it results in a very large executable if I want to keep using Mono and it is slower to startup than what I am using now.   Perhaps we should build the .NET standard library separately for now.

migueldeicaza:
I had another problem: the `poll()` call does not seem to be returning the data about activity on a file descriptor, so the application goes into a tight loop.   Need to debug this.

CloneDeath:
Sorry, I was sick all last week.

I think this stems from the fact that I used the beta release of Mono.Posix. I think the time to revisit this is when that package leaves beta.

PR #54, opened 2018-02-17
Unify projects in SDK-style and PackageReference

kzu:
Currently there's a bit of a mix between `packages.config` and `PackageReference`, in that despite projects having the latter, they still contain `Reference` items with `HintPath` pointing to @migueldeicaza's machine ;). 

This PR cleans that, switches the two remaining projects to SDK-style too, and bumps Mono so that it all builds nicely in travis.


kzu:
This makes working with the projects much better @migueldeicaza, while you're around ;)

PR #69, opened 2018-04-16
Renamed Demo projects and added NetCore to sln

seasoned-code:
Added netstandard1.6 to Terminal.Gui.csproj, while maintaining net461 and netcoreapp2.0.
Changed Terminal.Demo.NetCore reference Terminal.Gui to project instead of nuget.
Changed mainloop.cs to use a Task instead of Thread to support .Net Standard.
Added an OsVersion helper for NetStandard and added preprocessor to Core.cs for UseSystemConsole.


migueldeicaza:
The "Standalone" example is intended not to be compiled against the current source code, but against a NuGet package.   The idea being to show to users how they would use this in practice.

migueldeicaza:
On what scenario would I want this to be a .NET standard project?

It already supports the two environments that support the console.

seasoned-code:
I did not realize that was the intention of the Standalone (I think some documentation may help with that). The demo project is not in msbuild15 format, so I assumed it was not meant to be demoed against .Net Core (since the Standalone is).

My argument for supporting .Net Standard is for consumption. Speaking for myself, naturally, I am working on a framework project that I just want to target one .Net implementation, not a multitude because I do not always have the luxuries of having multiple environments to compile against (and I think that would be the assumption for others as well). Without .Net Standard, I will be forced to add multiple frameworks for my library and that will end with more maintenance.

Maybe I am thinking a bit one dimensional here, but is that not the intention of .Net Standard, to support multiple frameworks?

Before posting this PR, I looked at the other PRs and it appears your intention is to the maintain current functionality, this is why I was not recommending the removal of the other explicit environments and to add .Net Standard unhindering the existing archs.

migueldeicaza:
Got it - I think it might make sense to surface a .NET standard API, but I do not necessarily think the implementation should be limited to .NET standard 1.5.   The Nugetizer should support this scenario by computing the API intersection, just need to figure out how to do that.

I agree on docs and probably restructuring the directory to better show what is going on.

seasoned-code:
We can target a number of Standard frameworks, if that is what you would prefer.

Issue #67, opened 2018-04-10
Windows: mouse support, resizing support


migueldeicaza:
Need to add mouse support on Windows, which seems to require this:

https://docs.microsoft.com/en-us/windows/console/reading-input-buffer-events

https://docs.microsoft.com/en-us/windows/console/console-input-buffer

migueldeicaza:
Added some of the support for input buffers now.

migueldeicaza:
Fixed by @nickvdyck pull request.

PR #82, opened 2018-05-10
Windows: Implementing mouse events and writing to console buffers

nickvdyck:
This is a first initial implementation of making mouse events work on windows. Related to #67 although this does not solve the resizing issues.

It still pretty rough at the moment, just wondering what you think. It is built upon the initial work that was already in place, just had to wire some bits up.

Currently tested on .net core 2.1.0.rc1 and full framework 471.
Also verified that I didn't break any unix stuff. Ubuntu 16.04 dotnet core 2.1.0.rc1 and mono 5.10.
Don't have a mac, so haven't tested that one. But given that the changes or isolated to the windowsdriver this should be fairly safe. 😁 (famous last words)

### DIFF
--- a/Terminal.Gui/Core.cs
+++ b/Terminal.Gui/Core.cs
@@ -1590,14 +1590,12 @@ namespace Terminal.Gui {
 			if (Top != null)
 				return;
 
-			if (!UseSystemConsole) {
-				var p = Environment.OSVersion.Platform;
-				if (p == PlatformID.Win32NT || p == PlatformID.Win32S || p == PlatformID.Win32Windows)
-					UseSystemConsole = true;
-			}
-			//UseSystemConsole = true;
+			var p = Environment.OSVersion.Platform;
+
 			if (UseSystemConsole)
 				Driver = new NetDriver ();
+			else if (p == PlatformID.Win32NT || p == PlatformID.Win32S || p == PlatformID.Win32Windows)
+				Driver = new WindowsDriver();
 			else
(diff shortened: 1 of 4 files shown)


nickvdyck:
Thanks,

I like what you did with reshuffling those drivers around. I rebased this branch on top of master.

migueldeicaza:
Wow - this is amazing!   This is exactly what I wanted to do!

I came down to my office tonight to start reading documentation on the screen buffer and decided to check on where this pull request was.

The issue with the cursor should be fixable by implementing the UpdateCursor method by calling [SetConsoleCursorPosition](https://docs.microsoft.com/en-us/windows/console/setconsolecursorposition).

Can you tell me more about the mouse events not working?

Finally, before I merge, just as a sanity check, could you state on this commit that the code is MIT licensed?   Then I can go ahead and merge.

nickvdyck:
I fixed the cursor positioning by implementing the `UpdateScreen` method just like you said. Didn't realize it was that easy 🤦‍♂️, thanks for pointing me the right direction.

The basic implementation for mouse events is working. But I did notice that when clicking on the checkbox it checks it but than also immediately unchecks the control. I haven't really looked in to this so am not really sure what is happening here.

I'll see if I can find some time later today to do some debugging.

I also added an MIT license to the top of the file. I'm not really familiar with licensing so I'm not sure this is what you meant.

migueldeicaza:
Fabulous news!   Going to merge this now.

Thanks!

PR #87, opened 2018-05-15
Fix bug in ListView that caused it to render spaces off the window

pmsanford:
This was causing the ListView to render outside its frame and off the side of the terminal starting at the end of the last item in the ListView


migueldeicaza:
Thanks for your contribution!

Issue #81, opened 2018-05-06
Enter special non-standard characters in TextField


YauhenPyl:
Is it possible somehow processing non-standard characters in TextField like '=', '\' or numbers? Thanks.

luetm:
Just ran into this as well. Is there only support for text without numbers?

migueldeicaza:
Team, I do not understand this, what platform is this, and what is the problem?

Perhaps you are using the Keypad, instead of the main keyboard?

YauhenPyl:
It is about non-standard characters, e.g. how to put the next string to TextField: `Holly-Molly_23`.

nickvdyck:
This only happens on windows. For windows it uses a different key map function eg `MapKey` https://github.com/migueldeicaza/gui.cs/blob/master/Terminal.Gui/Driver.cs#L1056 . This does some filtering to only include characters between A-z. Because the mapping into a ConsoleKey only allows you to map to standard keys. The unix code path uses a different map function eg `MapCursesKey` https://github.com/migueldeicaza/gui.cs/blob/master/Terminal.Gui/Driver.cs#L545. 


Currently tested on Microsoft Windows [Version 10.0.17134.1] with dotnet core 2.1.0.rc1

I can try to make a put together a patch if given some pointers. 😁

luetm:
@migueldeicaza Thank you! It now works for the en-US keyboard. With the de-CH (Swiss German) keyboard layout we also have keys like AltGr, which we use for thinkgs like @,# or \ . These don't work yet. Is this easy to implement or do you have to create keymaps for all keyboard layouts?

luetm:
@YauhenPyl  For me it already works with 0.8.0

migueldeicaza:
@luetm: could you print out the results of using Console.ReadKey for any characters that are failing for you?

migueldeicaza:
Excellent - will close

luetm:
Sorry for the late reply.

I still cannot use the @ in 0.8.0. I think this might be a windows bug in relation to the de-CH layout. To make an @ with this layout I need to use this combo: https://i.imgur.com/k5WAs2y.png

Pressing this combination gives me the '2' character in a field. Any steps I can do to help debugging?

migueldeicaza:
Yes, can you use Visual Studio and debug what happens in read key event processing?

luetm:
[...]

PR #90, opened 2018-05-17
#81 fixed AltGr not being recognized.

luetm:
Found another issue testing this, but it fixes the AltGr part. Some special characters (triggered with AltGr or not doesn't matter, e.g. the § character, which doesn't need any modifiers at all, trigger a Â to be added to the console, in addition to the character. So pressin the key § yields Â§. I suspect something with unicode, but I don't have the overview (yet).

### DIFF
--- a/Terminal.Gui/Drivers/WindowsDriver.cs
+++ b/Terminal.Gui/Drivers/WindowsDriver.cs
@@ -38,7 +38,7 @@ namespace Terminal.Gui {
 		public const int STD_OUTPUT_HANDLE = -11;
 		public const int STD_INPUT_HANDLE = -10;
 		public const int STD_ERROR_HANDLE = -12;
-
+		
 		internal IntPtr InputHandle, OutputHandle;
 		IntPtr ScreenBuffer;
 		uint originalConsoleMode;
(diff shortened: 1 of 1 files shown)


nickvdyck:
I'm not super familiar with NStack (but am familiar to go programming) so am not sure if I'm right. But the textfield class is directly manipulating the `ustring` via its index. And I don't think this takes into account that a Unicode character can be made up out of surrogate pairs. It is important to make sure astral symbols are treated as one.

So if you would change the following lines (https://github.com/migueldeicaza/gui.cs/blob/master/Terminal.Gui/Views/TextField.cs#L124)  to something like this:

```csharp
if (p < text.RuneCount) {
        Driver.AddRune (Secret ? (Rune)'*' : (Rune)text.ToRunes () [p]);
}
```

And update the following (https://github.com/migueldeicaza/gui.cs/blob/master/Terminal.Gui/Views/TextField.cs#L266) to something like this:
```csharp
runes = text.ToRunes().ToList();
runes.InsertRange(point, kbstr.ToRunes());
SetText (ustring.Join("", runes.Select(r => ustring.Make((Rune)r)).ToArray()));
```

Than adding an astral character should work. Although there are few more places that need updating. But again I'm not really sure that this is the right solution.

migueldeicaza:
Good bug report, and good catch.

The whole reason for using `ustring` and NStack was to get this right, and I got it wrong :-) - I was not doing the proper Utf8 string processing.   I have now a patch, there are really two indexes to be used in the tex

PR #97, opened 2018-05-21
Handle case where there's no previous focus

pmsanford:
Ran into this with an app with no views other than the menu.


csnewman:
I've experienced this issue as well. I was going to submit a pr, but you've beaten me to it.

pmsanford:
Woops, didn't mean to add that last commit to this same PR, but it's here. I'll start branching :)

migueldeicaza:
Thanks for your contribution!

PR #100, opened 2018-06-07
Quick one: Ensure SelectionChanged is called on mouse click too

pmsanford:


migueldeicaza:
Thanks for your contribution!

PR #103, opened 2018-06-09
Add support FullColor (24bit) for Windows

prifio:
I add api for support 24bit color and 8bit color(256 color). Change struct Attribute to class Attribute and add class FullAttibute : Attribute. Add FullColor class (just r-g-b) and static method to convert 24bit color &larr;&rarr; 8bit color (Linux-like) &larr;&rarr; 4bit color (Old-windows-like). 
Change WindowsDriver implementation for support 24bit color.
But i can't implement checking windows version (old verson don't support 24bit color). I tried to use registry, but .Net Core don't support it. (see SetSupportFullColor() in WindowsConsole)
Change background of Box10x in example.
image

### DIFF
--- a/Terminal.Gui/Drivers/ConsoleDriver.cs
+++ b/Terminal.Gui/Drivers/ConsoleDriver.cs
@@ -82,7 +82,82 @@ namespace Terminal.Gui {
 		/// </summary>
 		White
 	}
+	
+	/// <summary>
+	/// 24bit color. Support translate it to 4bit(Windows) and 8bit(Linux) color
+	/// </summary>
+	public class FullColor {
+	    public int R { get; private set; }
+	    public int G { get; private set; }
+	    public int B { get; private set; }
+	    public FullColor(int r, int g, int b) {
+		R = r;
+		G = g;
+		B = b;
+	    }
 
+	    public static FullColor LinuxColor(int code) {
+		if (code == 7)
+		    code = 8;
+		else if (code == 8)
+		    code = 7;
+		if (code == 8)
+		    return new FullColor(192, 192, 192);
+		if (code <= 15) {
+		    int k = 128;
+		    if (code >= 9)
+			k = 255;
+		    return new FullColor(code % 2 * k, code / 2 % 2 * k, code / 4 % 2 * k);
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Thanks for the contribution!   I am currently on vacation, with limited time to review, but I think this looks good.

I will take some code, but I suspect that this might force me to finally revisit the color allocation code, and come up with something better.

prifio:
I rewrote some part of code. I removed FullAttribute and back to struct Attibute and expanded Attribute (it's strange, but I more time didn't see easy way). Refactored code, restored cursor position after ConsoleWrite. And I tried implement TrueColor support detection. `System.Runtime.InteropServices.RuntimeInformation.OSDescription` Work fine (unlike `System.Environment.OSVersion`), but CI server can't build bkz he can't find this property.

Terminal.Gui/Drivers/ConsoleDriver.cs
```diff
@@ -10,6 +10,7 @@
 using Mono.Terminal;
 using NStack;
 using Unix.Terminal;
+using System.Drawing;
```
migueldeicaza: System.Drawing will be problematic, I would rather avoid bringing another assembly dependency.

migueldeicaza:
This looks great.   I will check after this 0.16 bug fix releease

Terminal.Gui/Drivers/ConsoleDriver.cs
prifio: Sorry. I forgot remove it. Now, System.Drawing not used

tig:
Is this a PR that we should take? Is this still valid?

tig:
@migueldeicaza - please comment on this PR. Should we pursue it or abandon?

AndreR:
Pretty sad, that this was abandoned!

tig:
I agree. It would be awesome if someone took it on. The tracking issue is here: https://github.com/migueldeicaza/gui.cs/issues/48

AndreR:
@tig What do you think? How complicated would it be to re-implement this pull request into the current state of development?

PR #104, opened 2018-06-11
Fixes cases where subviews can be null

MihaMarkic:
subviews can be null in GetEnumerator() and also on Remove(view).


migueldeicaza:
Thanks for your contribution!

PR #109, opened 2018-06-28
Using newer version of ncurse

Urgau:
Add support for ncurse 6


migueldeicaza:
I do not think how this could work, since there is a large amount of code that needs to be updated to the new name.   See the file binding.cs for details.

PR #110, opened 2018-07-05
Handle shift-modifier if tab is pressed on System.Console and Windows drivers

mholo65:
- This fixes backtab not working on Windows and System.Console.


migueldeicaza:
Beautiful!   Thank you!

PR #113, opened 2018-07-08
Ensure SelectedChanged is called on mouse click.

mholo65:


migueldeicaza:
Thanks for the patch!

PR #114, opened 2018-07-09
Fix for setting and removing focus when using mouse.

mholo65:
- Ensure focus is always sent up the chain of subviews.
- Ensure that removed focus is sent down the chain of subviews.


migueldeicaza:
Thanks for the patch!

Issue #83, opened 2018-05-11
HexViewer pending issues


migueldeicaza:
Last line of the buffer is not displayed properly.

migueldeicaza:
Closing fixed bug.

PR #115, opened 2018-07-10
Fixed HexView Display Offset Bug

Daniel-McCarthy:
This pull request is in response to the open issue #83 "HexViewer pending issues".

The HexView would previously display lines that don't contain any data with a longer offset than lines that displayed hexadecimal data. This was due to filling in three spaces per byte instead of two. After correcting this the HexView has been displaying lines evenly in my tests.

I have included a comparison displaying the original issue and the results of the change.

hexviewfix


migueldeicaza:
Thank you for the patch Daniel!

PR #117, opened 2018-07-26
Error on RemoveAll

shargon:
already is removed on `Remove(View view)`


migueldeicaza:
Thanks for the bug fix!

PR #129, opened 2018-09-18
fix(repainting subviews): When `LayoutSubviews` needs to be called on…

GaikwadPratik:
… subview, the modified subviews are not painted as the `layoutNeeded` is set to `false` on parent view which prevent redraw of the parent as well child view. Hence we need to explicitly call `SetNeedsDisplay` which will recalculate UI layout based on modifications. But we still need to set `layoutNeeded` to `true` so that parent view is actually repainted.

Fixes: repainting issue related to https://github.com/migueldeicaza/gui.cs/issues/127


Terminal.Gui/Core.cs
```diff
@@ -423,7 +423,8 @@ internal void SetNeedsLayout ()
 		/// </summary>
 		/// <param name="region">The region that must be flagged for repaint.</param>
 		public void SetNeedsDisplay (Rect region)
-		{
+		{			
+			SetNeedsLayout();
 			if (NeedDisplay.IsEmpty)
```
migueldeicaza: This is not a good idea, as this will cause a very expensive computation on every request to redisplay a region.   The culprit for this problem is likely elsewhere.

SetNeedsLayout is intended to reduce the amount of work done, this does the opposite.

Conceptually this is not right either.

Terminal.Gui/Views/TextView.cs
```diff
@@ -392,14 +392,34 @@ void ColorSelection ()
 				Driver.SetAttribute (ColorScheme.Normal);
 		}
 
-		bool isReadOnly;
+		bool isReadOnly = false;
```
migueldeicaza: False is already the default.

Terminal.Gui/Views/TextView.cs
```diff
@@ -392,14 +392,34 @@ void ColorSelection ()
...
+		/// <value>Number of lines(Default 5).</value>
+		public int PageShift 
+		{
+			get => nPageShift;
+			set 
+			{
+				nPageShift = value;
+			}
+		}
+
```
migueldeicaza: We should likely change this to be computed and follow the same rules that Emacs uses based on the height of the view.

Terminal.Gui/Views/TextView.cs
```diff
@@ -670,6 +690,8 @@ public override bool ProcessKey (KeyEvent kb)
 			case Key.CursorDown:
 			case Key.ControlP:
 			case Key.CursorUp:
+			case Key.PageUp:
+			case Key.PageDown:
 				lastWasKill = false;
```
migueldeicaza: Other aliases include Alt-v for for page up, and control-v for page down

Terminal.Gui/Core.cs
GaikwadPratik: This is the only solution I could find. Else will need to trigger `TerminalResized` as suggested by @IKoshelev. But I don't like that either as it is not the correct event. Can you please help me here? I am kind of stuck because of this. TIA.

Terminal.Gui/Views/TextView.cs
GaikwadPratik: I know. :) Just wanted to be explicit.

Terminal.Gui/Views/TextView.cs
GaikwadPratik: I will add those too.

Terminal.Gui/Views/TextView.cs
GaikwadPratik: Can you point me to an example in the code?

Terminal.Gui/Core.cs
GaikwadPratik: I added that because, as mentioned in the commit message, in `LayoutSubViews` function call `if layoutNeeded` is false which we set at the end of the same function call. For redrawing/repainting we need it to be true.

GaikwadPratik:
@migueldeicaza Can you please review this PR and let me know if something needs to be changed?

Terminal.Gui/Views/TextView.cs
```diff
@@ -78,6 +84,27 @@ public static List<List<Rune>> StringToRunes (ustring content)
 			return lines;
 		}
 
+		// Splits a string into a List that contains a List<Rune> for each line
+		public static List<List<Rune>> StringToRunes (ustring content, int nMaxRuneCount)
```
migueldeicaza: Ok, I see what is going on here.   The parameter is not really the maxRuneCount, but the desired column. 

This should be updated accordingly.

migueldeicaza:
Ok, I see what is going on in this patch now.

I have to think a bit about how I want to implement wrapped lines.   I am not convinced that this is what I want to maintain in its current form.

If you can extract the page down/page up features into a separate PR, I can merge that while I think about the word wrap code.

Terminal.Gui/Views/TextView.cs
```diff
@@ -78,6 +84,27 @@ public static List<List<Rune>> StringToRunes (ustring content)
...
+		{
+			var lines = new List<List<Rune>> ();
+			int start = 0, i = 1;
+			for (; i < content.Length; i++)
+			{
+				if ((i % nMaxRuneCount).Equals(0) || content [i].Equals(10))
```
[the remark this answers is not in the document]
GaikwadPratik: Isn't `.Equals` better than `==`?

GaikwadPratik:
@migueldeicaza I have created https://github.com/migueldeicaza/gui.cs/pull/138. For Page up/down. How would you write logic for Text Wrapping? I kept both datastructure so that we can support both formats. However I am inclined towards making wrapped text default.

migueldeicaza:
Merged the other bit.

ikkeflikkeri:
Is there a fix for this drawing issue? The view is not updating if I add another one to it. I'm trying to create a wizard application. Same issue GaikwadPratik was talking about up there.

tig:
I am going to close this PR. I believe it is too stale to be valid at this point in time. Please feel free to submit another PR on this topic.

PR #135, opened 2018-09-25
Fixing bug in TextView, clicking line below text end causes IndexOutOfRangeException

IKoshelev:
To reproduce bug in original, take following application and click any line below text end (for example, line 8).

 ```
                    Application.Init();
		    var top = Application.Top;

		    top.Add(new TextView() {
			Text = @"11111
	22222222222
	3333333333
	4444444444444
	555555555555"
		    });

		    Application.Run();
```


migueldeicaza:
Thank you for your patch!

PR #137, opened 2018-09-30
Use an upper character for MenuItem's hotkey

sangkilc:
When a MenuItem contains a lowercase hotkey (e.g., `MenuItem("C_ut", "", null)`), the hotkey `u` does not work because `Mneu::ProcessKey()` function always compares it with an uppercase key.


migueldeicaza:
Thanks for your contribution!

PR #138, opened 2018-10-03
Pageupdown

GaikwadPratik:
PR for adding page up and down in text view


GaikwadPratik:
Fixes https://github.com/migueldeicaza/gui.cs/issues/139

PR #140, opened 2018-10-10
Clean up directory structure

McNeight:
* Place all source projects under src\ except StandaloneExample


migueldeicaza:
I wont merge this request, too much to review, if I decide to make this, I would likely do this by hand myself.

Generally, it is best to bring this topic up before submitting a patch:

McNeight:
It was my original intent to just move the source into a src\ directory (first commit), and I neglected to move off to a new branch before the second commit. I'll try to back the second commit out of the structure branch, if that would work for you?

PR #141, opened 2018-10-16
Fixed a bug with multiline labels not calculating their bounds correctly

HLFrye:
I noticed multiline labels didn't seem to be calculating their bounding rectangle correctly.  I've updated Label.GetRect to return a Rect using the mw variable instead of cols, which will cause the width to be calculated based on the longest line in the label, instead of the last one.


HLFrye:
Realized a second fix was necessary as well - I updated Label.Recalc so that the last line of a multiline string gets added to the lineResult List.

PR #144, opened 2018-10-30
Delete key on keyboard did not delete a char from TextView 

gamidon:
Added Key.DeleteChar to Key.ControlD switch condition. 

Verified it performs as I would expect.  Hope you do to.


migueldeicaza:
Thanks for your contribution!

PR #159, opened 2018-12-14
Fixed subview layout

distantcam:
When adding or removing child views the parent view was not recalculating it's size, or relaying the items. This fixes that problem.


jamie-davis:
Can I second this, as I was about to submit a PR with a similar fix and it would be very nice to get this into the released library.

migueldeicaza:

PR #160, opened 2018-12-14
Fixed multiline dialog layout

distantcam:
Multiline dialogs were still using the full width of the text line, not counting newlines. This lead to extremely wide dialogs with all the text left justified.

This fixes the issue by correctly calculating the width of a multiline label and using that as the dialog width.


migueldeicaza:
Woohoo!   Thanks!

PR #161, opened 2018-12-14
Made IListDataSource render usable without external fields set

distantcam:
Moves the `container` and `driver` fields into the `Render` method on the interface so the interface is all that's needed to render, and so the `ListView` doesn't need to set fields to work with an implementation of `IListDataSource`.


migueldeicaza:
Thanks for your contribution!

PR #168, opened 2019-01-31
Make cursor settable

sangkilc:
Hi,

Currently we cannot set the cursor position of a TextField, but it can be inconvenient sometimes. For example, one may want to insert a string in the middle of a Text in a TextField. This patch creates a setter for `CursorPosition`.

Thanks,
Sang Kil


migueldeicaza:

sangkilc:
Just realized that I haven't added boundary checks. For example, when I set CursorPosition to be a negative integer, it should make it as `0` instead. Do you want me to create another PR for this?

PR #186, opened 2019-04-29
Correct the ScrollView behaviour for vertical scrolling

DieselMeister:
Corrected the ScrollView behavior for vertical scrolling. In case of a vertical scrollbar the position of the "arrow" can not be determinated by the Bound.Width.


migueldeicaza:
Beautiful!   Thank you!

PR #187, opened 2019-05-01
Change Cursor Position before "Changed Event" is triggert. (Issue #2 …

DieselMeister:
…on Termina.Gui.Elmish)

In the elmish wrapper for the Termial.Gui I need the "right" cursor position when the "Changed" event is triggered. In the current Implementation the cursor position is set after the event is triggered, so either I have to wait some ms asynchronously before I can process the event or I change the order inside the ProcessKey Method by using a helper variable "oldCurPos" and set the cursor position before the SetText method is called.

Related to: https://github.com/DieselMeister/Terminal.Gui.Elmish/issues/2

### DIFF
--- a/Terminal.Gui/Views/TextField.cs
+++ b/Terminal.Gui/Views/TextField.cs
@@ -215,6 +215,11 @@ namespace Terminal.Gui {
 
 		public override bool ProcessKey (KeyEvent kb)
 		{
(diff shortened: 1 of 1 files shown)


DieselMeister:
Wow, That was fast! 🥇 👍

Issue #192, opened 2019-05-04
Menuitem separator


Slesa:
It would be nice to separate some menu items, e.g.
```
File
  Open
  Load 
  Save
  -------
  Quit
```
The line could be drawn if the title of the item is empty.

tig:
This was fixed ages ago. This issue should be closed.

PR #193, opened 2019-05-04
Fixing menu item separator (#192)

Slesa:
Fixing menu item separator (#192)
- do not crash when item is null
- when using cursor, do not activate line if item is empty
- use the right color scheme


migueldeicaza:
This looks fine, but it made a lot of white space changes.

Slesa:
Sorry for that. I **really** habe trouble with your decision to use Tabs. VStudio always reformats a few lines and I tried my best to repair it... 
I think the spaces were right again in the follow-up with the disabled items?

PR #199, opened 2019-05-05
Added first try of a statusbar

Slesa:
Added new Control StatusBar (#9) in the same style as the MenuBar.


Slesa:
Where exactly comes this build error from? I did not change the menu.cs here?

PR #200, opened 2019-05-06
Fix build. Fix spaces in Menu.

Slesa:
Master build is broken, at least on my machine. Changed my settings to insert TABs, too.


migueldeicaza:
I think I applied some of these fixes independently, is that why some code is commented out here?

Slesa:
Yes, the code is part of a fix. But it must be inside a new function and be called by the ctor. The way it was implemented did not work - at least on mu machine.

migueldeicaza:
This patch should no longer be needed in master

PR #201, opened 2019-05-06
Feature/status bar

Slesa:
First try for a StatusBar (#9). It uses the same layout as the MenuBar. 
Includes the fix of the build (#200) to be able to compile.

### DIFF
--- a/StandaloneExample/Program.cs
+++ b/StandaloneExample/Program.cs
@@ -165,6 +165,14 @@ class Demo {
 		MessageBox.ErrorQuery (50, 5, "Error", "There is nothing to close", "Ok");
 	}
 
+ 	static void Help() {
+		MessageBox.Query(50, 7, "Help", "This is a small help\nBe kind.", "Ok");
+ 	}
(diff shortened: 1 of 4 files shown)


migueldeicaza:
This looks good - would you be so kind in adding inline API docs, so folks can figure out how to use this?

Slesa:
I was not sure if you would like it. 
Next plan would be to show the help texts of the menu in such a statusbar, but draw the hotkey of a menu item after its text. As there is no event system, there may could be a static member StatusBar.Current to draw a message after the items. 

The problem is - I'd need my other commits of the menu system ;) I have too many PRs here.

migueldeicaza:
@BDisp  Do we still need this patch, now that we merged the other branch?

BDisp:
I think it's not needed now. But @Slesa had the merits of this.

Slesa:
Well, if it is not needed anymore...

migueldeicaza:
Thank you folks!

Issue #154, opened 2018-11-20
Menuitem - Disable


SemperFu:
It would be nice if the menu items had an enable/disable option. If I don't have a file open, I'd like the save/save as/close menus to be grayed out.

Slesa:
I have a PR for this feature, but it depends on another PR for the menu separators

tig:
This can be accomplished by setting the `CanExecute` property to a method that returns false (or true to enable).

```c#
item.CanExecute = delegate () { return false; };
```

closing.

PR #203, opened 2019-05-06
Feature/DisableMenuItem

Slesa:
Allow menu items to be disabled (#154)
They cannot be selected when using arrow keys.

### DIFF
--- a/StandaloneExample/StandaloneExample.csproj
+++ b/StandaloneExample/StandaloneExample.csproj
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Terminal.Gui" Version="0.17.0" />
+    <ProjectReference Include="..\Terminal.Gui\Terminal.Gui.csproj" />
+    <!-- PackageReference Include="Terminal.Gui" Version="0.17.0" / -->
   </ItemGroup>
 </Project>
--- a/Terminal.Gui/Drivers/CursesDriver.cs
+++ b/Terminal.Gui/Drivers/CursesDriver.cs
@@ -266,6 +266,7 @@ namespace Terminal.Gui {
 				Colors.Menu.Focus = Curses.A_BOLD | MakeColor (Curses.COLOR_WHITE, Curses.COLOR_BLACK);
 				Colors.Menu.HotNormal = Curses.A_BOLD | MakeColor (Curses.COLOR_YELLOW, Curses.COLOR_CYAN);
 				Colors.Menu.Normal = Curses.A_BOLD | MakeColor (Curses.COLOR_WHITE, Curses.COLOR_CYAN);
+				Colors.Menu.Disabled = MakeColor(Curses.COLOR_WHITE, Curses.COLOR_CYAN);
 
 				Colors.Dialog.Normal = MakeColor (Curses.COLOR_BLACK, Curses.COLOR_WHITE);
 				Colors.Dialog.Focus = MakeColor (Curses.COLOR_BLACK, Curses.COLOR_CYAN);
--- a/Terminal.Gui/Views/Menu.cs
+++ b/Terminal.Gui/Views/Menu.cs
@@ -131,13 +131,13 @@ namespace Terminal.Gui {
 				maxW = Math.Max (l, maxW);
 			}
 
-			current = -1;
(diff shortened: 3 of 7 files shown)


StandaloneExample/StandaloneExample.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Terminal.Gui" Version="0.17.0" />
+    <ProjectReference Include="..\Terminal.Gui\Terminal.Gui.csproj" />
+    <!-- PackageReference Include="Terminal.Gui" Version="0.17.0" / -->
```
migueldeicaza: This is intended to be a standalone example, one that does not require the current project.   If you want to just try things out with the released project, you can edit "Example" in the default solution.

Terminal.Gui/Drivers/CursesDriver.cs
```diff
@@ -266,6 +266,7 @@ public override void Init (Action terminalResized)
 				Colors.Menu.Focus = Curses.A_BOLD | MakeColor (Curses.COLOR_WHITE, Curses.COLOR_BLACK);
 				Colors.Menu.HotNormal = Curses.A_BOLD | MakeColor (Curses.COLOR_YELLOW, Curses.COLOR_CYAN);
 				Colors.Menu.Normal = Curses.A_BOLD | MakeColor (Curses.COLOR_WHITE, Curses.COLOR_CYAN);
+				Colors.Menu.Disabled = MakeColor(Curses.COLOR_WHITE, Curses.COLOR_CYAN);
```
migueldeicaza: While this is fine, it does not handle the black and white setting for Colors.Menu.Disabled, so you need to figure out what color you would like to use for that and set it there.

Terminal.Gui/Views/Menu.cs
```diff
@@ -76,6 +78,20 @@ public MenuItem (ustring title, string help, Action action)
...
+		/// <summary>
+		/// Gets or sets the action to be invoked if the menu can be triggered
+		/// </summary>
+		/// <value>Function to determine if action is ready to be executed.</value>
+		public Func<bool> CanExecute { get; set; }
+		
+		/// <summary>
+		/// Shortcut to check if the menu item is enabled
+		/// </summary>
+		public bool IsEnabled() {
```
migueldeicaza: Since we need some changes in this patch, let me suggest that you get the indentation to follow the existing code here, the brace should go in a new line.

Terminal.Gui/Views/Menu.cs
```diff
@@ -76,6 +78,20 @@ public MenuItem (ustring title, string help, Action action)
...
+			return CanExecute==null ? true : CanExecute();
+		}
```
migueldeicaza:
```suggestion
			return CanExecute == null ? true : CanExecute ();
		}
```

Terminal.Gui/Views/Menu.cs
```diff
@@ -131,13 +147,13 @@ static Rect MakeFrame (int x, int y, MenuItem [] items)
...
-			current = -1;
+			/* current = -1;
 			for (int i = 0; i < items.Length; i++) {
 				if (items [i] != null) {
 					current = i;
 					break;
 				}
-			}
+			} */
 			return new Rect (x, y, maxW + 2, items.Length + 2);
```
migueldeicaza: The above should likely be removed

Terminal.Gui/Views/Menu.cs
```diff
@@ -149,15 +165,25 @@ public Menu (MenuBar host, int x, int y, MenuBarItem barItems) : base (MakeFrame
 			CanFocus = true;
 		}
 
+		internal Attribute DetermineColorSchemeFor(MenuItem item, int index) 
+		{
+			if (item!=null) 
+			{
+				if (index==current) return ColorScheme.Focus;
+				if (!item.IsEnabled()) return ColorScheme.Disabled;
+			}
```
migueldeicaza: Let us follow the style, the return should go in a newline.

tig:
While I love the idea of this feature, this PR is very old. I'm going to close it. @Slesa, if you would be so kind, please resubmit this as a new PR. Thanks.

Issue #177, opened 2019-04-06
Labels don't display Unicode characters correctly on Linux & .NET Core


david-tamar:
The following label:
```
window.Add(new Label(0, 0, "我是一个中国人的一句。"));
```
Displays on console via gui.cs like that:
```
~H~Q ~X   ~@       ~[     ~Z~D  ~@ ~O  ~@~B
```

I'm using **.NET Core 2.2** from Microsoft's official repositories, **gui.cs 0.20.0** from NuGet (appears as “Terminal.Gui 0.20.0” on my Rider IDE NuGet package manager), fully-unicode-aware Rxvt terminal emulator on Linux Debian Buster. I've also tried to use Rider's built-in terminal emulator, it's the same results.

Although the default encoding of .NET Core on Linux systems is already UTF8, I tried to convert strings and IO all over to UTF8 just to make sure it's not something to do with the way .NET Core runs on my Linux Debian setup.

What I've tried so far:
1. Convert strings to UTF8 using `Encoding`.
```
public static string ToUTF8(this string text)
{
	var bytes = Encoding.UTF8.GetBytes(text);
	return Encoding.UTF8.GetString(bytes);
}
```
```
window.Add(new Label(0, 3, "©".ToUTF8()));
```
2. Set console input and output encoding to UTF8.
```
Console.InputEncoding = Encoding.UTF8;
Console.OutputEncoding = Encoding.UTF8;
```
3. Tested with UTF16, UTF32, UTF7 encoding as well, non deliver better results.

I can display Unicode characters without any problem on my console when not using gui.cs.

migueldeicaza:
Can you show me the contents of your environment, in particular I believe LANG and TERM

david-tamar:
This is the output from the `env` command on Xfce4-terminal:
```
TERM=xterm-256color
LANG=C.UTF-8
```
**Or more broadly:**
```
SHELL=/bin/bash
WINDOWID=67110285
QT_ACCESSIBILITY=1
COLORTERM=truecolor
GDK_DPI_SCALE=1.74
SSH_AUTH_SOCK=/tmp/ssh-MjliQuzLWN6X/agent.650
SSH_AGENT_PID=685
GTK_MODULES=appmenu-gtk-module
XDG_SEAT=seat0
PWD=/home/david
LOGNAME=david
QT_QPA_PLATFORMTHEME=gtk2
XDG_SESSION_TYPE=tty
[...]

PR #204, opened 2019-05-06
Feature/Unicode

Slesa:
A new demo under new folder 'examples' to test usage of Unicode characters. Part of the main menu is in Russian. Problems as found so far:
- Menu is not drawn correctly with ncurses (#177) but looks fine in windows
- As MenuBar items are not rune-ified, hotkey cannot be found
- Not sure if the keyboard value of Unicode input is calculated correctly
- NStack still lacks Unicode support

### DIFF
--- a/Terminal.Gui/Views/Menu.cs
+++ b/Terminal.Gui/Views/Menu.cs
@@ -131,13 +131,13 @@ namespace Terminal.Gui {
 				maxW = Math.Max (l, maxW);
 			}
 
-			current = -1;
+			/* current = -1;
 			for (int i = 0; i < items.Length; i++) {
 				if (items [i] != null) {
 					current = i;
 					break;
 				}
-			}
+			} */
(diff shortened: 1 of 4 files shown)


migueldeicaza:
Thanks for doing this work, it is fabulous to have a test case.

While this code started with a focus on doing Unicode, I did not always test, and you can see that there are some faults here and there.   

There are two classes of problem today, one is I have discovered that the API I am using for rendering runes is quite limited, so on Unicode situations, I should use add_wch instead of addch.   The former taking full cchar_t structures.   The second issue are places where I am not taking into account the size of a cell.   This is usually a minor problem, except for the text input capabilities, so I need to redo the work using some sort of (Rune, Byte) pairs to track both the rune *and* the number of columsn that the Rune will use when displaying.

This sample will be very useful.

tig:
I implemented this test case in the "Unicode in Menu" Scenario in `UI Catalog`. Closing this PR.

image

PR #205, opened 2019-05-07
Feature/TimeInput

Slesa:
new control: Time Input (#14) with support for long and short format, the latter w/o seconds.


migueldeicaza:
There is some commented out code, what is the purpose of that?

Slesa:
This code is relevant and part of a new function, but I cannot include it due to dependencies between my changes. 
But having it included, it did not compile anymore. I've no clue how this could have worked on the ci. Has it compiled there?

migueldeicaza:
Thank you!   I applied your patch manually due to:

- The additional changes that conflict with master
- Whitespace changes
- Reformatted to fit the current style

Issue #15, opened 2018-01-11
New Feature: Add Popup menu API


migueldeicaza:

tig:
Fixed by @BDisp with #1615

PR #206, opened 2019-05-07
New control: Popup menu

Slesa:
New Control: Popup menu (#15)
Problem is the hiding of the menu after ESC - it takes a second to redraw the screen. It also has probably too much source code in common with menu.cs.

Open questions:
- Should the menu be closed after clicking on the rest of the screen?
- Should the menu be closed after the action of an item is executed? 
  Here, the popup could also be a main menu where you call different submodules - as the example shows:
   Call the setup part, return to menu
   Call the maintenance part, return to menu
   All done, quit the menu and exit the application

### DIFF
--- a/StandaloneExample/Program.cs
+++ b/StandaloneExample/Program.cs
@@ -64,6 +64,18 @@ class Demo {
 			new Label (new Rect (0, 12, 40, 3), "4-Hello world, how are you doing today") { TextAlignment = TextAlignment.Justified });
 	}
 
+	static void ShowPopupMenu() {
+		var popup = new PopupMenu(22, 4, "Popup", new MenuItem [] {
+				new MenuItem ("_Configuration", "Show configuration", NewFile),
+				new MenuItem ("_Setup", "Change settings", () => {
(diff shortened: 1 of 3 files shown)


migueldeicaza:
Answers: 

* I think that we should close the popup menu if the user clicks outside of the menu
* I think that after the action is triggered, it should hide the menu

Slesa:
I don't think it should close automatically after the action is triggered. If it does not, you can build a small tool which (e.g. a Starter tool for other programs) which simply draws an initial popup menu in the center of the screen, which has submenus also drawn as popup menus. Which means: if you open a submenu and close the main menu after triggering, the submenu has no root window.

migueldeicaza:
I think that nested menus should be first class citizens, so activating a submenu does not close the menu, but beyond that, the action would make sense.

tig:
This is an old/stale PR. I'm closing it due to lack of activity.

@Slesa, if you would be so kind, please submit another PR with this change as having Popup menus (per #206) would be awesome!

PR #208, opened 2019-05-08
Implement WakeUp in WindowsDriver

SzymonPobiega:
My (probably naive) attempt to deal with https://github.com/migueldeicaza/gui.cs/issues/207 by implementing the `WakeUp` method in the `WindowsDriver`.


migueldeicaza:
Oh I like this!

migueldeicaza:
This looks good!

PR #243, opened 2019-08-07
Fixes changing focus using backtab

mholo65:
- Focus is now moved correctly when using tab and backtab

See 
backtab_focus

ping @migueldeicaza


migueldeicaza:

PR #250, opened 2019-08-30
Added DateField view

BarryNolte:
Created DateField control as per issue #13


migueldeicaza:
Thanks for the contribution!    How does this cope with formats in different locales?   For example, outside of the US the dates are usually DD/MM/YYYY

BarryNolte:
Aaarrggg... I should have caught that.  I'll fix it.

migueldeicaza:
@BDisp Do we still need this patch with your mega-patch merged? 

Can I clsoe this issue?

BDisp:
Yes you can close.

PR #253, opened 2019-09-02
bugfix in Button.ProcessKey and expose a new function for all drivers

giladlevi:


Terminal.Gui/Drivers/CursesDriver.cs
```diff
@@ -159,7 +157,7 @@ void ProcessInput (Action<KeyEvent> keyHandler, Action<MouseEvent> mouseHandler)
 			if (code == Curses.KEY_CODE_YES) {
 				if (wch == Curses.KeyResize) {
 					if (Curses.CheckWinChange ()) {
-						terminalResized ();
+						TerminalResized ();
```
biohazard999: TerminalResized?.Invoke()

biohazard999: Could be null or?

Therzok:
```suggestion
						TerminalResized?.Invoke ();
```

PR #259, opened 2019-09-07
Make API for UI construction more fluent

kzu:
The main API for constructing UIs is `Add(view)`. By making the
API return the view (typed to the received view type), we allow
further operations on the view (i.e. save it to a variable, set
additional properties, etc.).

This makes the calling patterns more concise, so instead of:

```
var label = new Label("...");
Add(label);
```

You can just do:

```
var label = Add(new Label("..."));
```

for example. Or:

```
Add(new TextField("")
{
  Y = Pos.Right(Add(new Label("Password: ")))
});
```

This API change is backwards compatible and none of the existing
demo code had to change.

### DIFF
--- a/Terminal.Gui/Core.cs
+++ b/Terminal.Gui/Core.cs
@@ -479,10 +479,10 @@ namespace Terminal.Gui {
 		/// </summary>
 		/// <remarks>
 		/// </remarks>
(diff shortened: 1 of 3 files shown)


kzu:
This is causing a stackoverflow on some code paths, not worth it :)

PR #262, opened 2019-09-12
Allow key events to traverse the TopLevel hierarchy

kzu:
It's quite common for key events to flow up the containment
hierarchy in UI frameworks. This could simplify management
of global hotkeys that can be placed in a top-level "App"
and have all views shown subsequently just "inherit" that
behavior as long as they let the key event to flow.

Make sure we make a copy of the items since a key handler
could modify the top-level collection (i.e. by closing a
dialog in between this processing).

### DIFF
--- a/Terminal.Gui/Core.cs
+++ b/Terminal.Gui/Core.cs
@@ -1315,6 +1315,13 @@ namespace Terminal.Gui {
 			get => true;
 		}
 
+		/// <summary>
+		/// Determines whether the <see cref="Toplevel"/> is modal or not. 
+		/// Causes <see cref="ProcessKey(KeyEvent)"/> to propagate keys upwards 
+		/// by default unless set to <see langword="true"/>.
+		/// </summary>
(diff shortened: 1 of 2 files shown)


migueldeicaza:
Can you explain the use case for this?

I suspect that this is to implement some sort of fallback - the problem that I have with this is that the Toplevel is supposed to consume all events, and not leak those that are not handled.

kzu:
Scenario is a subview handling some command, and letting others bubble up to the parent views. In our case (guit) we have local commands and global ones. Say a subview handles Paste differently than a parent view (think some editor view vs a shell handling pasting of filenames)

migueldeicaza:
We discussed in person the scenarios.

What we should get in place is a "bool Modal = true" flag on the Toplevel.   And those that do not want the modal behavior could set this to "False".

In this patch scenario, the foreach essentially has to stop at the first toplevel that has Modal = true.

kzu:
Addressed the feedback, ready for merging :)

Issue #151, opened 2018-11-18
FileDialog.FilePath


SemperFu:
FileDialog.FilePaths is an array of full paths. I have multiple files off so I should just be able to use FilePath. However this is only the FileName. 

This should be changed to the Full FilePath and there should include a FileName as well.

BDisp:
But joining the `FilePath `with each `FilePaths `you get the full path.

SemperFu:
Closing as this was already merged. Thank you

PR #263, opened 2019-09-14
Fix: Changed FileDialog.FilePath to full path

wangyw15:
Fix: Changed FileDialog.FilePath to full path
Related to #151


migueldeicaza:

PR #264, opened 2019-09-18
Added TextView.TextChanged event

kzu:
This is a generally useful event to expose in a textview :)

### DIFF
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // TextView.cs: multi-line text editing
 //
 // Authors:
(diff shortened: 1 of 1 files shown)


Terminal.Gui/Views/TextView.cs
```diff
@@ -312,6 +314,7 @@ void ResetPosition ()
 			set {
 				ResetPosition ();
 				model.LoadString (value);
+				TextChanged?.Invoke(this, new EventArgs());
```
Therzok: Maybe pass the old string in the event?

Therzok: Otherwise, if that's not needed/desired, please use `EventArgs.Empty`

PR #270, opened 2019-10-02
Upgrade from PackageLicenseUrl to PackageLicenseExpression

kzu:
The existing property is now considered legacy and results in a build warning: 

```
##[warning]C:\Program Files\dotnet\sdk\3.0.100\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(198,5): Warning NU5125: The 'licenseUrl' element will be deprecated. Consider using the 'license' element instead.
```


BDisp:
I don't having that warning...

PR #273, opened 2019-10-23
bugfix in WindowsDriver: removed the TerminalResized member that hide…

giladlevi:
…s the inherited one from ConsoleDriver


migueldeicaza:
Thank you for this bug fix!

PR #274, opened 2019-10-23
wasted allocations in WindowsDriver.EventPending

giladlevi:
[Re]move unwanted allocation (new CancellationTokenSource()) in WindowsDriver.EventsPending to only when needed.


migueldeicaza:

PR #275, opened 2019-10-23
[Core-Timeouts&Idle mechanism] wasted allocations

giladlevi:
Timeouts: move Timeout to a struct and avoid from allocations in mainloop.RunTimers.
Idles: move idleHandlers from type Func<bool> to Action and avoid from allocations in mainloop.RunIdle.

### DIFF
--- a/Terminal.Gui/Core.cs
+++ b/Terminal.Gui/Core.cs
@@ -1691,7 +1691,6 @@ namespace Terminal.Gui {
 			{
 				mainLoop.AddIdle (() => { 
 					d (state);
-					return false;
 				});
 			}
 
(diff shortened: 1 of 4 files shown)


BDisp:
@giladlevi this pull request crash with the actions methods of the menu. In the Example project open File->Text Editor Demo and then try File->Close. You can't exit from there. I already had committed that and only after a few days I saw that issue. The timeouts even and odd are causing this problem.

tig:
@BDisp can you please comment on the validity of this PR? You've been in the related code a bunch and I believe you are the expert on it. If this PR is something valid, please ether re-open it or submit a new one.

BDisp:
At the time I tested it and it didn't work and I didn't see any sense of having even and odd timeouts.

PR #277, opened 2019-10-23
[Core - Subviews]: added InternalSubview property without allocations

giladlevi:
Added InternalSubview property without allocations (the calling to subviews.AsReadOnly() actually allocates and Copy the List) for internal usages and keeps the public Subviews property for api compatibility.


migueldeicaza:
Thanks for your contribution!

PR #287, opened 2019-11-02
[guit] Add NETStandard 2.0 support

kzu:
In order to allow libraries (i.e. control libraries) to target
NS2 and work simultaneously on .NET and .NETCore versions of
guit, we add it as a target framework too, which automatically
causes the package to include that when referenced from a NS2
package.

The upgrade to net472 makes gui.cs fully compatible with
netstandard2.0 without the need for any facade assemblies.
This simplifies and streamlines package installation and
reduces dependencies on other packages.

### DIFF
--- a/Terminal.Gui/Terminal.Gui.csproj
+++ b/Terminal.Gui/Terminal.Gui.csproj
@@ -1,6 +1,6 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


Therzok:
Made a PR for this too: https://github.com/migueldeicaza/gui.cs/pull/284

I would still keep net461 as a target. Net472 does not create any additional output assemblies when targeting netstandard2.0, but net461 does.

I wonder why the netcoreapp target is needed. I think net461/netstandard2.0 are the cleaner targeting solution.

migueldeicaza:
Kzu, there is a different patch - can you tell me if the one I merged from @Therzok  works for you?

kzu:
Yep, that one works too :)

PR #291, opened 2019-11-13
Fixed and Enabled Library reinitialization

KDERazorback:
_These changes allows the library to be shutdown, returning to the initial Console configuration, and then reinitialized again at any time. Changes are cross-platform and tested on Windows and Linux/OSX._

Changes:
- Replaced static driver initialization with property getter for reference passing in Core.cs::View class, this allows the library to be reinitialized at any time.
- Made the Shutdown method on Core.cs::Application class public, since there is no reason to keep it private. Applications can shutdown the library and revert the console to the initial stage by calling it on demand.
- Fixed a memory-leak on Drivers/WindowsDriver class by destroying the generated screen buffers at driver shutdown by calling CloseHandle.
- Minor change to Core.cs::Application.Init(Func<Toplevel>) for better initialization-status tracking, via backend property instead of relying on the Top field of the class.

### DIFF
--- a/Terminal.Gui/Core.cs
+++ b/Terminal.Gui/Core.cs
@@ -243,7 +243,7 @@ namespace Terminal.Gui {
 		/// Points to the current driver in use by the view, it is a convenience property
 		/// for simplifying the development of new views.
 		/// </summary>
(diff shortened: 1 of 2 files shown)


migueldeicaza:

PR #296, opened 2019-11-16
Fixes #282 "Repaint Issue"

### LINKED ISSUE #282: Repaint Issue
Hello,

I've found an issue where the RadioGroup-Control does not repaint correctly. Here is the code to reproduce it:

```csharp
Application.Init();

Window window = new Window("Repaint Issue") {X = 0, Y = 0, Width = Dim.Fill(), Height = Dim.Fill()};
RadioGroup radioGroup = new RadioGroup(1, 1, new[] { "Short", "Longer Text  --> Will not be repainted <--", "Short" });

Button replaceButton = new Button(1, 10, "Replace Texts above")  {
    Clicked = () => { radioGroup.RadioLabels = new string[] { "Longer than before", "Shorter Text", "Longer than before" }; }
};

window.Add(radioGroup, replaceButton);
Application.Top.Add(window);
Application.Run();
```

The "--> Will not be repainted <--" will not be repainted with "blanks".
I think there are more issues with repainting especially when replacing SubViews and so on.

Regards,
Jens

BDisp:


kbilsted:
Why the use of #if ?

BDisp:
Because I use various samples differents situations and is simple to separate them. Isn't necessary for the PR but copy and paste causes this situations :)

kbilsted:
I think the PR should be 100% ready when you submit and not require the maintainer to do stuff "at the edges "

BDisp:
@kbilsted sorry but I disagree with you at this point because the demo only serves to demonstrate that a core feature is functional. Therefore, the demo is only intended to demonstrate how a particular functionality of a library is implemented and can be at the discretion of each person. With #if, #elif and #endif you can have many samples testing you want, only needing to set false what you don't want execute and set to true what you want execute. Sometimes there are mistakes that escape to us. In that case, @migueldeicaza will only be merged if all the code is correct.

PR #302, opened 2019-11-17
Allowing list items selection

BDisp:
Try List Demos->Select Items from menu and selection work well now


migueldeicaza:

PR #303, opened 2019-11-17
Timefield format with bounds values

BDisp:
Uses 0-24H format and implemented lower and upper bounds to TimeField

### DIFF
--- a/Terminal.Gui/Views/TextField.cs
+++ b/Terminal.Gui/Views/TextField.cs
@@ -35,7 +35,7 @@ namespace Terminal.Gui {
 		///   Client code can hook up to this event, it is
 		///   raised when the text in the entry changes.
 		/// </remarks>
-		public event EventHandler Changed;
+		public event EventHandler<ustring> Changed;
 
 		/// <summary>
 		///    Public constructor that creates a text field, with layout controlled with X, Y, Width and Height.
@@ -98,7 +98,10 @@ namespace Terminal.Gui {
 			}
 
 			set {
+				ustring oldText = ustring.Make(text);
 				text = TextModel.ToRunes (value);
+				Changed?.Invoke(this, oldText);
+
 				if (point > text.Count)
 					point = Math.Max (text.Count-1, 0);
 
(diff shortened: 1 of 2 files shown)


kbilsted:
Not sure what problem you are trying to solve ?

BDisp:
@kbilsted you are right, in which case we should also use sepChar in the format. Thanks.

kbilsted:
`sepChar` is a strange name for a `string` variable.  But more imporantly Not sure what problem you are trying to solve ? Why not trust the standard parser ?

BDisp:
@kbilsted sepChar can be interpreted as a separator character and not as a Char type. I think it is safer to use the culture info derived separator than to use only the standard parser and also when using boundary checking. Anyway, coding has many different ways to achieve the same goal, depending on one's perspective :)

BDisp:
@kbilsted with this commit you will understand why the sepChar variable was declared as char :)

kbilsted:
Not sure what problem you are trying to solve ? Why not trust the standard parser ?

I Think it ia better to use the full string not just the first char. Some locales may define sevetal chars as separator

BDisp:
@kbilsted in my commit I commented that you would realize why it was declared as char. Is that this change forced me to change the checking of the statements like if (Text [CursorPosition] == ':')
to if (Text [CursorPosition] == sepChar.ToCharArray () [0]). Perhaps this is why it was originally in the repository declared as char to make it easier to verify. However, there is nothing wrong with your observation.

kbilsted:
Your can compare using `string.StartsWith()` and provide an index

BDisp:
@kbilsted does not work because Text is a ustring that contains the ascii character code and not a string. To compare we have to convert to string in case of ustring or convert to char in case of string, which is what I did. As there has to always be this conversion the original author decided to declare the separator as char and I think it should continue like this, so I do not take the PR not to be accepted. But, however, there is an important comparison because I am comparing with the separator variable instead of the literal character ':'

migueldeicaza:
Fantastic patch - thank you for your contribution!

PR #310, opened 2019-11-23
extract methods on ListView to make it controlable from other controls

kbilsted:


migueldeicaza:

Issue #306, opened 2019-11-19
When awaiting a method it is awaited until the cursor moves


timothyparez:
I am probably just doing it the wrong way but the idea is 
that I have one (or more) buttons which when pressed/clicked
load some data and display that data in a ListBox.  
  
When I click the button the first time I can see the first `WriteLine`
and then the `LoadItemsAsync` method is awaited.  
  
If I don't do anything it will just await forever.   
When I click the button again or move focus to the next control  
the awaited method completes and the next `WriteLine` is executed.  

Update: *If I move the mouse the awaited method also completes*
  
I have tried without `MainLoop.Invoke` as well.  
I'm guessing I am just handling this the wrong way. 

Any suggestions?

```
button.Clicked += async () =>
{
    Application.MainLoop.Invoke (async () => {
            
        //When the button is 'clicked' the following is executed
        Debug.WriteLine($"Clicked the button");
        var items = await LoadItemsAsync();

        //However the following line is not executed
        //until the button is clicked again or
        //until the cursor is moved to the next view/control 
        Debug.WriteLine($"Got {items.Count} items)");                                                     
        itemsList.SetSource(items);

        //Without calling this the UI is not updated
        this.LayoutSubviews();                    
    });
};
```

System: **Ubuntu 19.10**

BDisp:
Can you provide the code of what you want to do? I needed your code to look for more if I could make it available, of course. But in principle, the problem is that code is not executed because it is awaitable and will only execute after the execution of the asynchronous method. When the asynchronous method is fully executed it is in a different thread than the one in which it was called. Try the following code:

button.Clicked += async () => { 
				//When the button is 'clicked' the following is executed
				Debug.WriteLine ($"Clicked the button");
				using (CancellationTokenSource cancellationTokenSource = new CancellationTokenSource ()) {
					syncContextTaskScheduler = TaskScheduler.FromCurrentSynchronizationContext ();
					await Task<List<string>>.Run (() => LoadItemsAsync (cancellationTokenSource.Token).Result).ContinueWith (task => {
						//However the following line is not executed
						//until the button is clicked again or
						//until the cursor is moved to the next view/control
						var items = task.Result;
						Debug.WriteLine ($"Got {items.Count} items)");
						itemsList.SetSource (items);

						//Without calling this the UI is not updated
						this.LayoutSubviews ();
					}, CancellationToken.None,
						TaskContinuationOptions.OnlyOnRanToCompletion,
[...]

PR #312, opened 2019-11-25
Fixes issue #306 async/await hang

BDisp:


BDisp:
> 
> 
> [a683345](https://github.com/migueldeicaza/gui.cs/commit/a68334551098b3d97ffffd41771aee8c2f002375)
> @timothyparez is the last one. I think it must always be the last. Someone said that changes submitted to an already submitted PR can always be done on the same PR. However, I had a doubt that I would be grateful if someone with true knowledge of the subject please clarify me. Must the changes be submitted based on the last submission, as it really was, or will I have to make the changes again from a new upstream/master checkout? Thanks in advance.

I put a question on https://github.com/migueldeicaza/gui.cs/issues/306#issuecomment-558539249 regarding this PR and although I am waiting for a reply by someone well understood on this subject, I reflected and I am forward with this my following opinion. I really think we should only merge the upstream because there may have been new commited merged from uptream after the last PR and then resubmitting the new changes. By the way, if there have been changes in the upstream and submit without merging first does git warn you before or not?

timothyparez:
@BDisp I have no idea. I mostly use Gitlab (hosted and on-premise) because Github is blocked every other day here.

BDisp:
The question is for anyone who might be able to answer me correctly. Anyway thanks

timothyparez:
@BDisp so I checked out your latest. 
The weird behavior persists.

BDisp:
@migueldeicaza can you help me with this CI error, please. In my travis all went well with no errors was you can see below:
travis

In your travis I got errors with nuget unsigned packages as you can see below:
yourtravis

 After merged with your master I also started receiving the same errors as yours. Looks like you might have to update some nuget packages for CI to stop giving errors.

migueldeicaza:

Issue #313, opened 2019-11-25
Make `ListWrapper` public so it can be used by consumers


timothyparez:
Currently if you want to assign an `IList` to `Source` have to use the `SetSource` method  
because the `SourceProperty` is actually of type `IListDataSource`.

But when implementing DataBinding that means you need to convert `IList` to `IListDataSource`  
and you do this by wrapping the `IList` into an `IListDataSource` implementation.

For example, here I define 2 bindings, one for a `Label` the other for a `ListView`

    var ustringConverter = new UStringValueConverter();
    var listWrapperConverter = new ListWrapperConverter();

    var labelBinding = new Binding(this, "Text", label2, "Text", ustringConverter);
    var listBinding = new Binding(this, "Heroes", listView, "Source", listWrapperConverter);

Looking at the code for the converter:

    public class ListWrapperConverter : IValueConverter
    {
        public object Convert(object value, object parameter = null)
        { 
            var wrapper = new CustomListWrapper((IList)value);
            return wrapper;            
        }
    }

All it does it wrap the `IList` into a `CustomListWrapper` instance.  
But this `CustomListWrapper` class is just a 1:1 copy of the internal `ListWrapper` class.
Id would be nice to just write `var wrapper = new CustomListWrapper((IList)value);`

kbilsted:
I think if you override `ToString()` on your list element types you can do without the ListWrapper class

timothyparez:
@kbilsted not sure we are talking about the same thing.
I may not have explained it properly.

Imagine we have a `List<Person>` called `people` and a `ListView` called `theList`.
   
     //This way, the render process will call ToString() on every instance of Person
     theList.SetSource(people); 

So to customize what gets rendered, you can just override `ToString()` that is true.

I`m creating a `Binding` class which takes 
 - a `Source` instance of type `object`
 - a `SourceProperty` property of type `string`
 - a `Target` instance of type object
 - a `TargetProperty` property of type `string`
 - an optional instance of `IValueConverter` 

When the `Source` value is changed the `Target` is updated by setting the `TargetProperty`
on the `Target`:
    
    var targetProperty = Target.GetType().GetProperty(TargetPropertyName);                 
    targetProperty.SetValue(Target, finalValue);   
  
Since the `Binding` does not know the difference between a `Label` and a `ListView` it does not  
know about `ListView.SetSource(IList source)` so it tries to `ListView.Source = value;`  

[...]

PR #314, opened 2019-11-26
Making ListView.ListWrapper public #313

timothyparez:


kbilsted:
I would prefer moving it out so it is no longer an inner class :-)

timothyparez:
> 
> 
> I would prefer moving it out so it is no longer an inner class :-)

That would be even better yes, I do not like nested classes.    
Not sure what the original reasoning behind it was.   
I'll take a look in the morning if it would break anything.

kbilsted:
It won't

BDisp:
@timothyparez, I was already working on this change :).
As @kbilsted stated it is better to move outside the ListView class to avoit the "Do not nest ListWrapper type. Alternatively, change its accessibility so that it is not externally visible."
Also I had prepare a litle change to ListView to allow setting SetSource async. If you want to add to include your PR, better.

```
		/// <summary>
		/// Sets the source to an IList value asynchronously, if you want to set a full IListDataSource, use the Source property.
		/// </summary>
		/// <value>An item implementing the IList interface.</value>
		public Task SetSourceAsync (IList source)
		{
			return Task.Factory.StartNew (() => {
				if (source == null)
					Source = null;
				else
					Source = MakeWrapper (source);
				return source;
			}, CancellationToken.None, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
		}

```

timothyparez:
@BDisp It should be ok now.
Currently I just moved it to ListWrapper.cs but since it is not a `View`,  
should we move it to a separate folder and what should we name it?

timothyparez:
Azure pipelines failed but I can't see why.

BDisp:
You could leave it in the same file it was before because it belong to ListView. See in the travis about the error

timothyparez:
I always stick to the one class per file rule.

BDisp:
> 
> 
> I always stick to the one class per file rule.

Me too. But in others' houses we have to respect the rules I think. If you look at the Menu.cs file you will find that there are four classes and only two are derived from View, because the author thought that as it was referring to the same subject decided to keep everything in the same file. In these cases it even looks good and is very common. If I were to decide, I would keep it in the same file but outside the ListView class. But it's up to you, of course. But the biggest problem is that you are submitting your PR through your master. I've done it before :)

timothyparez:
Let me follow @BDisp's advice and I will create a new PR later.

Issue #317, opened 2019-12-08
Log output in textView 


oSumAtrIX:
I have to refresh the application everytime I change text of a textview which is a bad solution. Is there anything to refresh the text of the textview without refreshing the application?

migueldeicaza:
Fixed by the above

PR #320, opened 2019-12-10
Textview reviewed

### LINKED ISSUE #318: System.ArgumentOutOfRangeException on click on empty TextView
When the TextView is empty, clicking on it causes an out of range exception

BDisp:
Fixes #318 out of range exception and #317 text redraw when navigate backward


migueldeicaza:

PR #333, opened 2020-03-01
Added DeepestView property in MouseEvent

cekkr:


migueldeicaza:
Why would this be used, as this should always be sent to the deepest view?

BDisp:
@migueldeicaza his idea have some logic that I used in the menu and textfield to notify which view the mouse is hover. And besides the view property I added two more, OfX and OfY to the MouseEvent struct.

migueldeicaza:
Got it - is this now a repeat?   There are two conflicting files

BDisp:
@migueldeicaza, however the author deleted the branch and I was unable to change it. Can I do anything?

tig:
Can we close this?

BDisp:
Yes, I already implement it.

PR #348, opened 2020-03-12
Switch to stable version of .NET framework reference assemblies

kzu:


migueldeicaza:
It looks like this broke the build?

kzu:
Might be worth bumping mono too, the CI is using a pretty old version too AFAICS. Maybe switch to latest table 6.8.0?

migueldeicaza:
How do I change the version?   Do I need to pick a different VM?

And would love to move to something else, what do I need to do?

kzu:
Here's a guide to creating a public project in Azure DevOps: https://docs.microsoft.com/en-us/azure/devops/organizations/public/create-public-project?view=azure-devops

I'm bumping the mono version now too (it's just that line in the ci definition)

kzu:
Ok, Travis is green now. I don't know what's failing on the devops since it's in a private project so I have no access. I don't know how Mono is provisioned on those images or how it's updated. TBH, it would be simpler to just switch to a windows-2019 image...

migueldeicaza:
Could I make the build logs public?

migueldeicaza:
Here is the failure:

https://gist.github.com/migueldeicaza/f91dfad46b90b6818503c430285fe98f

PR #361, opened 2020-04-13
What it does is that if it doesn't allow multi-selection and you already have a different item selected, select the new one and deselect the previous one.

BDisp:


migueldeicaza:
Thank you!

PR #367, opened 2020-04-14
Fixes #355 stack overflow with Pos based on the size of windows at startup. Added a OnResized action to set the Pos after the terminal are resized.

### LINKED ISSUE #355: Pos.Left, Pos.Bottom, etc... stack overflow if passed view has not resized/shown
To repro, modify `demo.cs` so the end of the file is as shown:

```csharp
		win.Add (drag, dragText);

		var bottom = new Label ("This should go on the bottom!") {
			X = Pos.Left(win),
			Y = Pos.Bottom (win)
		};
		win.Add (bottom);

		top.Add (win);
		//top.Add (menu);
		top.Add (menu, statusBar, ml);
		Application.Run ();
	}
}
```

In other words, it's currently (seemingly) impossible to position items relatively using `Pos.Left`, `Pos.Right`, `Pos.Top`, or `Pos.Bottom` before the app runs. 

I'm assuming this is all because the 'View` that is being passed (`win` above) does not yet have a size or location because resizing hasn't happened.

A possible fix would be to implement an `OnResize` event on `View` that would let these things be set.

BDisp:


migueldeicaza:
Thank you for tracking this down and the patch!

PR #375, opened 2020-04-15
Fixed a moving window issue. Added OnLoad Action because there are settings that need to be accessed only once.

BDisp:


BDisp:
@tig Can you please tell me which of the two situations you intended?

`a) "This should go on the bottom of the same top-level!"`
`b) "This should go on the bottom of another top-level!"`

If it is a) you can attach the label to the window, so there is no problem.

If it is b) you cannot attach the label to the window because the label is outside the window's limits, so the window cannot be responsible for repainting it, so the label must be assigned to another top-level.

tig:
> @tig Can you please tell me which of the two situations you intended?
> 
> `a) "This should go on the bottom of the same top-level!"`
> `b) "This should go on the bottom of another top-level!"`
> 
> If it is a) you can attach the label to the window, so there is no problem.
> 
> If it is b) you cannot attach the label to the window because the label is outside the window's limits, so the window cannot be responsible for repainting it, so the label must be assigned to another top-level.

@migueldeicaza  actually did that, but I've spent a bunch of time on `StatusBar` and the whole 'position at bottom' problem, so I now have clarity. This PR is NOT ready:

https://github.com/migueldeicaza/gui.cs/pull/372

The summary is: Windows Terminal & ConEmu support vertical resize. cmd.exe does not. I doubt curses or Net do either. But for those that do we should provide two options:

1) The behavior currently in demo.c where "This should go on the bottom" is tied to `win`.
2) The behavior currently in StatusBar.c where it tracks to `Application.Driver.Rows -1`.

In my next `StatusBar` PR, I introduce a `Parent` property. If `Parent` is set (e.g. `Parent = win`) we get behavior #1 (last line of parent view).

If `Parent` is null, then the `StatusBar` sticks to bottom of console window. Not sure why anyone would really want this, but it mimics current behavior.

Make sense?

BDisp:
I think it's a good idea. I will try to dedicate myself a little to handling the mouse in linux because it behaves competently differently from Windows. It is impossible to drag on windows, as well as clicking on the menu. It really fails.

PR #407, opened 2020-04-27
Fsharp example updated with the Demo version.

BDisp:


BDisp:
I need to figure out why the scrollview 2 with the type of Box10x does not redraw.

PR #409, opened 2020-04-28
Faster handling of the mouse in ScrollView with the addition of an uninterrupted click on the mouse features. Changed the Wakeup method to reset the events to update the screen.

BDisp:


migueldeicaza:
Beautiful!

tig:
I think this broke this: https://github.com/migueldeicaza/gui.cs/pull/208.

Or, I'm doing something wrong. The progress bars do not update unless I move the mouse around:

```csharp
		private void Pulse ()
		{
			if (_activityProgressBar.Fraction + 0.01F >= 1) {
				_activityProgressBar.Fraction = 0F;
			} else {
				_activityProgressBar.Fraction += 0.01F;
			}
			_pulseProgressBar.Pulse ();
			
		}

		private void Start ()
		{
			if (_timer != null) {
				_timer.Dispose ();
				_timer = null;
			}

			_activityProgressBar.Fraction = 0F;
			_pulseProgressBar.Fraction = 0F;

			_timer = new Timer ((o) => Application.MainLoop.Invoke (() => Pulse ()), null, 0, 250);
		}

		private void Stop ()
		{
			if (_timer != null) {
				_timer.Dispose ();
				_timer = null;
			}

			_activityProgressBar.Fraction = 1F;
			_pulseProgressBar.Fraction = 1F;
		}
```

BDisp:
I'll check that but is strange because I already fixed before.

tig:
FWIW, I don't get automatic updates when I run under Linux either. Am I doing something wrong in my code above?

BDisp:
To work in WSL I think the fixes are in PR #435

tig:
This is still broken. The `Progress` example in #501 demonstrates this for me using BOTH `WindowsDriver.cs` and `CursesDriver.cs`:

BDisp:
With this sample it will work
https://github.com/migueldeicaza/gui.cs/pull/501#issuecomment-632962091

PR #410, opened 2020-05-02
Improved a better clipped screen. Fixes some bugs with ScrollView, Menu. Added some virtual methods.

BDisp:


BDisp:
TextField now has a historical record of all entries, allowing undo and redo. It also allows you to select text with the keyboard and not just the mouse. The terminal is no longer restricted by the OS's proprietary keys, so you can now use Ctrl + C, Ctrl + V and Ctrl + P directly.

Terminal.Gui/Core.cs
```diff
@@ -1471,7 +1563,7 @@ public override bool ProcessKey (KeyEvent keyEvent)
 				return true;
 			case Key.CursorLeft:
 			case Key.CursorUp:
-			case Key.BackTab:
+			case Key.BackTab | Key.ShiftMask:
```
migueldeicaza, line 1566: I suspect that this is a problem on Windows - on Unix, this should be returning Key.BackTab, so I suspect the driver on Windows should clear "ShiftMask" in this case?

migueldeicaza: start review

Terminal.Gui/Core.cs
```diff
@@ -1779,7 +1881,7 @@ public override bool MouseEvent (MouseEvent mouseEvent)
 
 			int nx, ny;
 			if ((mouseEvent.Flags == (MouseFlags.Button1Pressed | MouseFlags.ReportMousePosition) ||
-				mouseEvent.Flags == MouseFlags.Button4Pressed)) {
+				mouseEvent.Flags == MouseFlags.Button3Pressed)) {
 				if (dragPosition.HasValue) {
```
migueldeicaza, line 1885: I do not understand this change.

Terminal.Gui/Core.cs
```diff
@@ -2218,6 +2351,8 @@ public static void Shutdown ()
 
 		static void Redraw (View view)
 		{
+			Application.CurrentView = view;
```
migueldeicaza, line 2354: Should we not restore this value after Redraw?

But generally, this looks like some global state that is prime to be trampled over, I would rather not have it.

What is the rationale for having this?

Terminal.Gui/Drivers/ConsoleDriver.cs
```diff
@@ -443,6 +443,10 @@ public abstract class ConsoleDriver {
 		/// </summary>
 		public abstract int Rows { get; }
 		/// <summary>
+		/// If true, it allows to write at the beginning of the next line if it's exceed the column limit.
+		/// </summary>
+		public abstract bool AllowNewLine { get; set; }
+		/// <summary>
```
migueldeicaza, line 449: I do not think that we should have this, if we did, it should be called "Auto-wrap", but this really belongs as a higher-level concept.

Instead, if we are out of boundary, this should cut the text.

If a view has a desire to wrap around, it should do this on its own, this should not be magic in the driver, otherwise, well intended code with bugs will have additional bugs to work around.

Terminal.Gui/Core.cs
BDisp: For the rightmost mouse button in Windows Driver is defined the Button 4 but in the Unix is defined the Button 3. So all the code in the drag where if the user click with the right mouse button in Windows the drag work, but not in Unix that uses the Button 3. It's not working in the Unix yet because limitations of bitwise flags combinations.

Terminal.Gui/Core.cs
BDisp: I will.

Terminal.Gui/Core.cs
BDisp: That static variable it's very important for to know what view is being drawn when a Rune is added to the screen. In case it exceeds the screen limits it's not drawn unless the AllowNewLine is setted. I can change it to Auto-Wrap wich I thing it's looks a beter name.

Terminal.Gui/Drivers/ConsoleDriver.cs
BDisp: That it's what is it is actually doing. It's only let drawn if it isn't not out of boundary or only let drawn if the AllowNewLine is defined.

Terminal.Gui/Core.cs
BDisp: > Should we not restore this value after Redraw?

Application.CurrentView is restored at Redraw method of the Application class, View class, Toplevel class and in the Window class. Thus is restored in all most top-levels.

BDisp:
> 
> 
> Clipping should not implement wrap-around. Clipping should instead enforce that if the current drawing position is outside the clipping view, that the operation does not draw anything on the screen.
> 
> So we should drop all the code that contains support for wrapping around and instead do clipping.

BDisp:
> Clipping should not implement wrap-around. Clipping should instead enforce that if the current drawing position is outside the clipping view, that the operation does not draw anything on the screen.
> 
> So we should drop all the code that contains support for wrapping around and instead do clipping.

So in that case you want me to eliminate the AllowWrap property and force the clipping instead of the wrapping around? If that is the case I also fully agree. I just did it because I thought there was some need to be allowed if someone wanted it.

Terminal.Gui/Core.cs
BDisp: Done

Terminal.Gui/Core.cs
BDisp: Done

Terminal.Gui/Drivers/ConsoleDriver.cs
BDisp: Done

BDisp: > If a view has a desire to wrap around, it should do this on its own, this should not be magic in the driver, otherwise, well intended code with bugs will have additional bugs to work around.

All views can write directly to the console driver, which would require having to redraw all views to obtain the same result. So managing this directly in the driver avoids all that work that could even be useless, overwhelming and possibly not have the desired result.

Designer/Program.cs
```diff
@@ -15,19 +15,45 @@ static void Close ()
 			MessageBox.ErrorQuery (50, 7, "Error", "There is nothing to close", "Ok");
 		}
 
+		static void Copy ()
+		{
+			TextField textField = menu.LastFocused as TextField;
+			if (textField != null && textField.SelectedLength != 0) {
+				textField.Copy ();
+			}
```
migueldeicaza, line 23: I think this is fine for now, but generally, I think that we should have a global "Copy" operation that copies whatever the currently focused view wants to provide.

So we might need to introduce a "IInputView" interface or something like that, following the spirit of UIKit.

Terminal.Gui/Core.cs
```diff
@@ -1445,8 +1535,9 @@ public override bool ProcessKey (KeyEvent keyEvent)
 				return true;
 
 			switch (keyEvent.Key) {
-			case Key.ControlC:
-				// TODO: stop current execution of this container
+			case Key.ControlQ:
```
migueldeicaza, line 1538: Oh, I forgot about this.

This is what we had been discussing with @tig on a different thread.   I forgot I had added this as a convenience.  Guess it was a bad call.

Terminal.Gui/Drivers/CursesDriver.cs
```diff
@@ -62,8 +65,49 @@ public override void AddRune (Rune rune)
...
-			foreach (var rune in str)
+			foreach (var rune in str) {
+				if (IsOutBounds (rune))
+					return;
 				AddRune (rune);
+			}
+		}
+
+		bool IsOutBounds (Rune rune)
+		{
```
migueldeicaza, line 76: Thank you for this heroic effort, but I suspect this is the wrong place to put this, the result is too complicated.

There are two ways of properly solving this, one is to track the "last known position" in View (the Move method already calls Driver.Move) so we would keep "current-col" and "current-row", and then introduce a `View.AddRune` method there, which only needs to check in the View whether it is out of bounds or not, and update its own counter there.

And then we move the code from using `Driver.AddRune` to `view.AddRune`.

We should then add a big fat [Obsolete] to `Driver.AddRune` to discourage its use (and the same goes for `AddStr`).

An alternative design to not keep track of a per-view col/row position, is to have AddRune in the view take a column/row, and advance that on each view that does drawing manually.  But I suspect this is too cumbersome.

A third design option is to do what UI toolkits do and is to create a "drawing-context" upon entering draw, so draw code would be like this (more or less, this is from the top off my head):

```
void Draw (Rect rect)
{
    var context = GetDrawingContext ();
    context.AddRune ("h");
    context.AddString ("ello");
}

And the x/y coordinates would be on the context.    Personally, I think that keeping this on the View is an acceptable tradeoff.

Designer/Program.cs
BDisp: I agree. I had already thought about it after having implemented it in TextField. Work for later.

Terminal.Gui/Core.cs
BDisp: I needed the generic Ctrl + C to copy and Ctrl + X to cut so the best alternative I saw was to change to Ctrl + Q which I think is a good alternative.

Terminal.Gui/Drivers/CursesDriver.cs
BDisp: I think you can put that in the View so that it can control that flow. These are ideas that come with time :)

PR #413, opened 2020-05-08
Fix topological sort in view class

En3Tho:
In View class there is a method TopologicalSort which actually worked not quite the way it intended to work. Instead of getting all nodes with no incoming edges it got all nodes with ones.

Simple case to repro: n1->n2->n3->n4.
Expected nodes with no incoming edges: n1
Actual: n2,n3,n4

Another case:
n1->n2, n1->n3, n1->n4
Expected: n1,
Actual: same n2,n3,n4

Because of this at start elements were at wrong positions because pos calculation order was wrong and it needed either resizing to call layout update or somehow did it on it's own in like 15 seconds (i guess it's about internal timers or smth)

### DIFF
--- a/Terminal.Gui/Core.cs
+++ b/Terminal.Gui/Core.cs
@@ -1248,12 +1248,12 @@ namespace Terminal.Gui {
 		}
 
 		// https://en.wikipedia.org/wiki/Topological_sorting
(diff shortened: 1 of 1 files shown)


migueldeicaza:
Thank you!

PR #414, opened 2020-05-09
Feature/ComboBox

fergusonr:
Simple text field with auto-complete.   List Demos -> Search Single Item

https://drive.google.com/open?id=1uxh5a6y8kqIv6ZMdy6-k_CJQWv1bhz_y


BDisp:
Good job @fergusonr. My advice is to change this code:

```
this.Enter += (object sender, EventArgs e) => {
	this.SetFocus (search);
	search.CursorPosition = search.Text.Length;
};
```

to this one:

```
public override bool OnEnter ()
{
	if (!search.HasFocus)
		this.SetFocus (search);
	search.CursorPosition = search.Text.Length;

	return true;
}
```

and let the Enter event to the end user. That was the reason why I changed that.

fergusonr:
Thanks @BDisp change checked-in

fergusonr:
Hi @BDisp  (I think) your last check-in has changed some behaviour of TextField.Text The change event:-

    void Search_Changed (object sender, ustring text)

the text argument is the *previous* text, not the current. Which is not a big deal as I can use the actual TextField instance value.

However if I programatically set the Text value, it only takes effect on the second assignment

   search.Text = "missed"; // does not set
   search.Text = "found"; // this one works

BDisp:
@fergusonr you have both values of the `TextField.Text` in the event. The new one through ((TextField)sender).Text and the old one through the `e`. So I think you must deal with this two.

BDisp:
@fergusonr that is breaking in the linux. Since this project is multi-platform you must check that.

BDisp:
@fergusonr it's still breaking in the linux. See the DirListView implementeation on the File->Open. You are dealing folders with a Windows's way only that not is compatible with Unix. Use WSL and install dotnet, mono and vsdbg on it to run your project on linux.

BDisp:
Well done. It's working in both now. I haven't macOS but as it is based Unix OS, it must working there too.

fergusonr:
@BDisp these should be my last check-ins. Tidy up behaviour when hosted in a Window

https://drive.google.com/open?id=1nS6GY8qSW9Ka4TcmtMvQ46whdTOnRB-X

BDisp:
That's great. Your control is near to be a combo that you can add properties to a read only feature too. How about that? :-)

fergusonr:
> That's great. Your control is near to be a combo that you can add properties to a read only feature too. How about that? :-)

Your right

BDisp:
@fergusonr don't forget to rebase your branch to reflect the changes in the master for the files you are modifying.

tig:
@fergusonr Can you please implement a Scenario on `UI Catalog` that exercises this control as part of this PR?

See instructions here: https://github.com/migueldeicaza/gui.cs/tree/master/UICatalog

fergusonr:
@tig that's done. My scenario appears at the end for some reason I cannot immediately figure out

tig:
This is very cool.

I think that change to `Label` was done by @BDisp and I just merged it as part of that PR.

image

I will open an issue to track it as I've seen it as well (see BUGBUG here: ).

If you start typing in the edit and then pick an item, I would think the list would update to reflect that the edit now has that item in it full.

image

E.g. in the above the list should JUST have `debian_version` in it.

Also, when open it's not clear that it's an edit and a list. I wonder if making the edit one or two char narrower would help?

image

Or... use a different color for the list? Or both:

image

fergusonr:
Thanks @tig agree the list should be differentiated from search. I do not change the list on selection because that's the behaviour of the Windows ComboBox. ie: it disappears on selection (unless I'm misunderstand you)

All said, I do not have a strong opining either way, so if this gets merged to master, its open to public PR's

tig:
What about my suggestion for making the list narrower than the edit?

BDisp:
A suggestion may even be of variable size within a minimum and maximum limit depending on the size of the item.

tig:
Also, please add 

```csharp
///<inheritdoc cref="name"/>
```

comments to all of your `public overrides`. This will ensure we don't have any warnings on build. 

e.g.

```csharp
///<inheritdoc cref="OnEnter"/>
public override bool OnEnter ()
		{
			if (!search.HasFocus)
				this.SetFocus (search);

			search.CursorPosition = search.Text.Length;

			return true;
		}
```

fergusonr:
@tig ok, thats done.

BDisp:
@fergusonr this is getting fantastic. I'm already visualizing the ComboBox with a 'V' on the right and it is possible to open and close the search list :-)

tig:
I verified this is AWESOME by fetching the PR locally.

Great freaking job!

BDisp:
Really he is doing a fabulous job. I'm really amazed.

PR #420, opened 2020-05-14
Corrects the position of the StatusBar so that it is always at the bottom of the screen. Works perfectly with curses.

BDisp:
The top-level must have the responsiblility for positioning the status bar in its proper place. The problem that happened before, mainly with curses, is that Redraw was not executed because it was outside the top-level frame. I also changed the MenuBar to possibly position a scrollable top-level.


BDisp:
I think that this PR fixes the issue stated in #442 and #437.

tig:
Great work. Makes me jealous on Windows ;-)

PR #422, opened 2020-05-15
Fixes the return value of the position cursor in the TextField.

BDisp:
I detected the issue when I double clicked on the right side of the TextField outside the text.


BDisp:
@migueldeicaza should I change the shortcuts for TextField's Undo and Redo to Ctrl+U and Ctrl+R, respectivally, due the fact Ctrl+Z to beeing used to suspend unix terminal?
			
```
case Key.ControlZ:
	Driver.Suspend ();
	return true;
```

migueldeicaza:
Makes sense - I was not aware that C-z had been bound at some point to undo.

BDisp:
Done. Thanks.

tig:
Not a fan of this, guys. 

Maybe it's because the last time I was a Unix guy was about 1989. 

Both Mac and Windows use Ctrl-Z for undo/Ctrl-Y redo.

How frequently do people need to "suspend" console GUI apps? I argue never. 

Please don't 'undo' (pun-intended) 30 years of GUI standardization by making undo be Ctrl-U. 

The same thing applies to Ctrl-c/x/v.

BDisp:
@tig that why I waited until @migueldeicaza pronounce about this. But as you had read he prefer that and I have to do what he said.

tig:
> @tig that why I waited until @migueldeicaza pronounce about this. But as you had read he prefer that and I have to do what he said.

No, he said "Makes sense - I was not aware that C-z had been bound at some point to undo."

I did not read this as him saying he preferred breaking how both Mac & Windows work in favor of supporting a concept (suspending a terminal) that a) only applies to Linux, and b) doesn't really make sense for a "console GUI" app (it's for pausing long listings scrolling by, IIRC).

BDisp:
@tig I thought he was answering my question in the affirmative.

Issue #429, opened 2020-05-16
ListView needs events


tig:
When using a `ListView` I need to know 

1. When the user changes which item is selected
2. When the user checks (or unchecks) an item (presses space)
3. When the user presses enter when on an item
4. When the user clicks on an item with the mouse
5. When the user double-clicks on an item with the mouse

I can't do any of this easily right now.

tig:
Fixed. Well done @BDisp !

PR #440, opened 2020-05-17
Added a OpenSelectedItem event to the ListView #429

BDisp:


tig:
I've tested this with `UI Catalog` and it works wonderfully. Really nice work! Thanks.

BDisp:
Usage example:

```
private static void ListView_OpenSelectedItem (object sender, EventArgs e)
{
	var item = ((ListView)sender).SelectedItem;
	var itemName = ((ListView)sender).Source.ToList ();
	MessageBox.Query (50, 7, "Selected Item is: ", $"{item}:{itemName[item]}", "Ok");
}

```

BDisp:
In agreement and subscribing to what @tig mentioned in #447, I present the respective changes to the pending ListView PR. Now, by invoking any of these events, the desired values ​​in the '`e`' argument can be obtained directly.

```
private static void ListView_SelectedChanged (object sender, ListViewItemEventArgs e)
{
	MessageBox.Query (50, 7, "Selected Item is: ", $"{e.Item}:{e.Value}", "Ok");
}

private static void ListView_OpenSelectedItem (object sender, ListViewItemEventArgs e)
{
	MessageBox.Query (50, 7, "Selected Item is: ", $"{e.Item}:{e.Value}", "Ok");
}

```

migueldeicaza:
I love the change - a conflict came up after some recent merges, can we fix this, and then merge?

BDisp:
Ok after all the merges are done I'll fixes them.

Issue #246, opened 2019-08-17
TimeField is not fully implemented


fstiffo:
I'm using the NuGet package for version 0.24 in an F# (core 4.7) project , but TimeField view is no present. Any idea?

kbilsted:
I have the same problem. So I cannot run the example from https://github.com/migueldeicaza/gui.cs/blob/master/Example/demo.cs 

The code is in the repo (https://github.com/migueldeicaza/gui.cs/blob/master/Terminal.Gui/Views/TimeField.cs) but inspecting the nuget it does not contain the type.

tig:
Renamed Issue to reflect state of control.

See `UI Catalog` scenario to illustrate. See PR #475

tig:
I think we're good here. Reopen if you don't think so.

PR #490, opened 2020-05-22
Date/TimeField - Added features stated in #475 and  #246

BDisp:
It's depend on the PR #488


tig:
Looks like you need to merge this with another of your PRs as it's failing:

 Views/TimeField.cs(98,14): error CS0103: The name 'ReadOnly' does not exist in the current context [/home/vsts/work/1/s/Terminal.Gui/Terminal.Gui.csproj]

BDisp:
I already merged it.

tig:
BDDisp, I don't think this is ready. 

- Layout still doesn't work (Calculated or Absolute).
- The TimeDate Scenario causes crashes easy.

I'll push an update to UI Catalog with a repro asap.

I think we should agree that any new functionality checked in to this project must have a `Scenario` in `UI Catalog` that sufficiently exercises it.

BDisp:
I'll see it. Remember that you can't force the width because they are fixed to be stay centered in the text field.

BDisp:
I just verified that you already implemented it in the same way as my PR. Do you want it to close?

tig:
> I just verified that you already implemented it in the same way as my PR. Do you want it to close?

Not sure what you mean. Can you elaborate?

PR #499, opened 2020-05-22
Using the Mainloop timer.

BDisp:


BDisp:
Is not only return true, but a boolean. So if is return true the timer continues and stop otherwise. #483

tig:
I don't think you and I are on the same page here. You've expanded the use of `MainLoop.AddTimer` which was not my intent.

In the `Progress` sample, my intent is to demonstrate multi-threading with `Terminal.Gui`. I is NOT to use (or demonstrate) `MainLoop.AddTimer`.

This line of code:

```
timeoutToken = Application.MainLoop.AddTimeout (TimeSpan.FromMilliseconds (10), Pulse);
```

Is a HACK intended to work around the fact that `Mainloop` processing does not continue when `Invoke` is called. I should have not included it in the code I checked in. 

I do not want to use MainLoop timer in this sample. I want to have the thread created by `System.Timer` to cause my code to run and update the UI every n milliseconds.

BDisp:
Was only to demonstrate that it work

PR #512, opened 2020-05-23
Fixes some DateField and TimeField bugs.

BDisp:


BDisp:
@tig If you merge this PR you will see that the problem of the TimeAndDate.cs scenario will solved. The problem is that it is using absolute values ​​and computed at the same time. As the PR already has new constructors to receive only the value of the date and time is resolved.

BDisp:
It seems that DateTime does not handle the various date formats well when switching from long format to short format so I will have to create my own patch for that purpose. If, however, you are able to solve it in another way, please tell me.

tig:
> It seems that DateTime does not handle the various date formats well when switching from long format to short format so I will have to create my own patch for that purpose. If, however, you are able to solve it in another way, please tell me.

Building a control that handles all date/time formats is a Hard problem. I did it in native Windows back in the '90s and it was a BITCH.

BDisp:
I'll try to tame it :-)

BDisp:
Problem solved. You can merge now if you want. PR #524

BDisp:
> 
> 
> It seems that DateTime does not handle the various date formats well when switching from long format to short format so I will have to create my own patch for that purpose. If, however, you are able to solve it in another way, please tell me.

Solved with PR #524

PR #514, opened 2020-05-23
Fixes #409 - Invoke does not cause Wakeup #501.

BDisp:
If there are no mainloop timers or other SO events it stay suspended forever. Now it responds to any event.

### DIFF
--- a/Terminal.Gui/Drivers/WindowsDriver.cs
+++ b/Terminal.Gui/Drivers/WindowsDriver.cs
@@ -522,9 +522,9 @@ namespace Terminal.Gui {
 		void WindowsInputHandler ()
 		{
 			while (true) {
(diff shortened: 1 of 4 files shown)


tig:
I know I've been guilty of this too, but we really should make sure PRs include only the code relevant to the PR. For myself, I commit to being more careful.

tig:
I confirmed this PR Closes #501. Nice job!!!

BDisp:
> 
> 
> I know I've been guilty of this too, but we really should make sure PRs include only the code relevant to the PR. For myself, I commit to being more careful.

I totally agree, but as they were just small things that could go on afterwards, I did it this way. As I also know that grammar is very important, I detected and changed it. But I promise to be more concise in separating the themes.

PR #525, opened 2020-05-23
Edge case handling and Quality-of-life improvements on Core, ListView and MenuBar

KDERazorback:
*This commit is intended to fix an elusive crash that may occur on the library when working with multiple independent TopLevel views; Adds an IsCurrentTop convenience property to the base TopLevel and improves both ListView and MenuBar views to avoid raising obscure exceptions on some edge cases.*

**Changes:**

+ Added `IsCurrentTop` read-only property to `Gui.Toplevel` class to allow a more convenient checking if a view's instance is currently on top (*active and displayed*).
- Fixed an elusive crash that may occur in the `Application.RunLoop` method due to a null'ed `Toplevel.NeedDisplay` property. This issue appears to be caused by a race condition that may occur when switching Views (TopLevel) too fast. Extended to all other `NeedDisplay` checks too.
- `ListView` control now displays empty rows for Null items in its Items collection, instead of crashing with a `NullReferenceException`.
- Improved `MenuBarItem` constructor behaviour by performing an additional sanity check on the `MenuItem[]` children parameter to ensure its not null. If so, raise an `ArgumentNullException`. Using an empty array of type `MenuItem` for this parameter still displays an empty menu as intended.

### DIFF
--- a/Terminal.Gui/Core.cs
+++ b/Terminal.Gui/Core.cs
@@ -351,6 +351,15 @@ namespace Terminal.Gui {
 		public ustring Id { get; set; } = "";
 
 		/// <summary>
+		/// Returns a value indicating if this View is currently on Top (Active)
+		/// </summary>
+		public bool IsCurrentTop {
+			get {
+				return Application.Current == this;
+			}
+		}
+
+		/// <summary>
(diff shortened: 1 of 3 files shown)


tig:
Looks great. Thanks.

PR #534, opened 2020-05-24
I just raised NetDriver up. I don't know if it's worth it.

BDisp:


tig:
Much better! The `System Console` scenario still crashes on exit though.

BDisp:
To be honest I just tested it on the Example project :-)

tig:
Haha. I do the same thing, where I only run `UI Catalog` ;-)

PR #543, opened 2020-05-25
Fixes #540

BDisp:


tig:
Dude, you are amazing. Well done!

I've further updated `Progress` such you can change the tick-count of either demo at runtime.

It's a pretty neat demonstration of how threading is more deterministic. Change the tick time of both to something small (like 100ms) and you can see how `AddTimer` starts to fall behind. 

Here I use 100ms and then do a bunch of stuff like holding down TAB and resizing the console. After a bit you can see that `AddTimer` gets behind. I don't think this is a bug BTW, this is just the nature of preemptive vs. non-preemptive multi-tasking.

BDisp:
I toggled the `idleHandlers `before `waitTimeout`. Toggle it again to compare.

```
if (mainLoop.idleHandlers.Count > 0 || CkeckTimeout (wait, ref waitTimeout)) {
	return true;
}

```

BDisp:
I think `System.Timer` is more inaccurate.

Issue #536, opened 2020-05-24
Unix "ReportMousePosition" not working in UI Catalog


BDisp:
But works on the Example project.

BDisp:
The cause is the` Curses.endwin ()` call which forces the restoration of linux to default and does not allow any further calls to be made to any `Curses `functions. It is not easy to solve, because as there are no more `Toplevels `you can not just call this function if it really is to close the program completely. In `windows `this is no problem.

Issue #518, opened 2020-05-23
SystemConsole (aka NetDriver) appears to have broke


tig:
I figured I'd write a Scenario to illustrate `Application.UseSystemConsole = true;`

I immediately discovered I had to change how `Scenario.Init, .Setup, and .Run` worked. Which I did.

Then I couldn't figure out why I was getting weird behavior and crashes.

So I modified `demo.c` to `Application.UseSystemConsole = true;` and ran it.

Crashed on startup.

I don't have time to look at this now, but from what I can see nobody's been testing `SystemConsole` for a while and it's quite broken...

BDisp:
SystemConsole does not support the gui. It is as if you are writing characters directly on the console. So this just uses the OS's default console.

tig:
Oh. I thought it worked for gui, but just didn't support mouse.

BDisp:
I tested it once and nothing works, at least I got this impression that made me think right away, why use this option if it doesn't work lol

tig:
To be clear, `UseSystemConsole` means the `NetConsole` will be used. Just so we're talking about the same thing:

```csharp
		static void Init (Func<Toplevel> topLevelFactory)
		{
			if (_initialized) return;

			var p = Environment.OSVersion.Platform;
			Mono.Terminal.IMainLoopDriver mainLoopDriver;

			if (UseSystemConsole) {
				mainLoopDriver = new Mono.Terminal.NetMainLoop ();
				Driver = new NetDriver ();
			} else if (p == PlatformID.Win32NT || p == PlatformID.Win32S || p == PlatformID.Win32Windows) {
				var windowsDriver = new WindowsDriver ();
				mainLoopDriver = windowsDriver;
				Driver = windowsDriver;
			} else {
				mainLoopDriver = new Mono.Terminal.UnixMainLoop ();
				Driver = new CursesDriver ();
			}
			Driver.Init (TerminalResized);
			MainLoop = new Mono.Terminal.MainLoop (mainLoopDriver);
			SynchronizationContext.SetSynchronizationContext (new MainLoopSyncContext (MainLoop));
			Top = topLevelFactory ();
			Current = Top;
			CurrentView = Top;
			_initialized = true;
		}
```

Here are related Issues:

- #370
- #48 
- #84 
- #418 

I'll stop now and get back to something important.

BDisp:
[...]

Issue #551, opened 2020-05-27
UI-Catalog window is not updating after size restored


BDisp:
To repo: With a normal window, click on any scene and maximize the window. Ctrl + Q to quit and the main window remains at its initial size when it should be maximized as well.

tig:
Interesting...

BDisp:
I'm almost to send a PR to solves all this. #536, #515, #518

tig:
I think this is related to the way `WindowsDriver` attaches/detaches to the console. 

Right now it never explicitly attaches or detaches, but just uses the implied attach by calling `GetStdHandle`. I think the fix may be to explicitly call `AllocConsole` in the `WindowsConsole` constructor and `FreeConsole` in `Cleanup`.

BDisp:
But that I think does not solve the problem in Unix and NetDriver because they have other methods of freeing up their resources. Anyway, I will present and we will exchange ideas.

tig:
Oh, so this repros on Unix as  well? Gotcha!

BDisp:
As a matter of fact this is not specific because in Unix refresh always updates the screen size, but the mouse and color problem does.

PR #553, opened 2020-05-27
Fixes #515, #518 and #536.

### LINKED ISSUE #515: Strange behavior with the color of the scenario changing from time to time in linux
I am observing a strange behavior with the color of the scenario changing from time to time in linux, verified here.
color-behaviour
color-behaviour2

BDisp:
The Driver and MainLoop should only be started and closed once.


BDisp:
And also #551

tig:
This is really good stuff.

PR #556, opened 2020-05-27
Fixes #437. Must explicit that is the current Top.

### LINKED ISSUE #437: A secondary Toplevel with a Menubar and no other child window behaves badly
See `UI Catalog`'s `TopLevelNoWindowBug` for repro. When run, the menu appears, but draws incorrectly and some keystrokes don't work.

```csharp
using Terminal.Gui;

namespace UICatalog {
	[ScenarioMetadata (Name: "TopLevelNoWindowBug", Description: "Illustrates that not having a Window causes Application.Run to wedge. #")]
	[ScenarioCategory ("Bug Repro")]

	class TopLevelNoWindowBug : Scenario {

		public override void Run ()
		{
			var ntop = new Toplevel (new Rect (0, 0, Application.Driver.Cols, Application.Driver.Rows));

				new MenuBarItem ("_Файл", new MenuItem [] {
					new MenuItem ("_Создать", "Creates new file", null),
					new MenuItem ("_Открыть", "", null),
					new MenuItem ("Со_хранить", "", null),
					new MenuItem ("_Выход", "", () => ntop.Running = false )
				}),
				new MenuBarItem ("_Edit", new MenuItem [] {
					new MenuItem ("_Copy", "", null),
					new MenuItem ("C_ut", "", null),
					new MenuItem ("_Paste", "", null)
				})
			});
			ntop.Add (menu);

			// BUGBUG: This being commmented out causes Application.Run to wedge.
			//var win = new Window ($"Scenario: {GetName ()}") {
			//	X = 0,
			//	Y = 1,
			//	Width = Dim.Fill (),
			//	Height = Dim.Fill ()
			//};
			//ntop.Add (win);

			Application.Run (ntop);
		}
	}
}
```

Note, because this code is running in `UICatalog` it is possible this won't reproduce if there is not ANOTHER `Toplevel` in the mix.

BDisp:


UICatalog/Scenarios/TopLevelNoWindowBug.cs
```diff
@@ -8,31 +8,31 @@ class TopLevelNoWindowBug : Scenario {
...
-			var ntop = new Toplevel (new Rect (0, 0, Application.Driver.Cols, Application.Driver.Rows));
+			Top = new Toplevel (new Rect (0, 0, Application.Driver.Cols, Application.Driver.Rows));
 
 			var menu = new MenuBar (new MenuBarItem [] {
 				new MenuBarItem ("_Файл", new MenuItem [] {
 					new MenuItem ("_Создать", "Creates new file", null),
 					new MenuItem ("_Открыть", "", null),
 					new MenuItem ("Со_хранить", "", null),
-					new MenuItem ("_Выход", "", () => ntop.Running = false )
+					new MenuItem ("_Выход", "", () => Top.Running = false )
```
tig, line 18: I do not like the use of `Top.Running = false`. When I first wrote this sample, I didn't fully appreciate what it did. In all of our examples we should use `Top.RequestStop` which does the same thing. I would like to remove `Top.Running` as a public API at some point (having public APIs be based on setting a property is just bad design).

BDisp: I agree.

PR #563, opened 2020-05-28
Fixes #562 background drawing issue.

### LINKED ISSUE #562: Something broken the background drawing.
It is also advisable to always test with a modal dialog box to check the background design of the previous visible toplevel.
background-paint

BDisp:


tig:
Please resubmit. Note `core.cs` no longer exists and you'll have to merge these changes into the right place.

PR #584, opened 2020-05-31
Added class KeyModifiers which improvements better control over the combinations keys.

### LINKED ISSUE #531: HexView: Captures TAB- doesn't let you tab forward to next view
Oddly enough BACK-TAB *does* tab back out.

But when the focus is in the text editing portion TAB just takes focus back to the hex editing portion.

If focus is on the text editing portion and the user presses Shift-TAB the focus does go to the previous `View`.

BDisp:
Fixes #531

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -379,12 +379,39 @@ namespace Terminal.Gui {
 			};
 		}
 
+		KeyModifiers keyModifiers;
+
+		KeyModifiers MapKeyModifiers (Key key)
+		{
+			if (keyModifiers == null)
+				keyModifiers = new KeyModifiers ();
+
+			if (!keyModifiers.Shift && key.HasFlag (Key.ShiftMask))
+				keyModifiers.Shift = true;
+			if (!keyModifiers.Alt && key.HasFlag (Key.AltMask))
+				keyModifiers.Alt = true;
+			if (!keyModifiers.Ctrl && key.HasFlag (Key.CtrlMask))
+				keyModifiers.Ctrl = true;
+			//if (!keyModifiers.Capslock)
+			//	keyModifiers.Capslock = true;
+			//if (!keyModifiers.Numlock)
+			//	keyModifiers.Numlock = true;
+			//if (!keyModifiers.Scrolllock)
+			//	keyModifiers.Scrolllock = true;
+
+			return keyModifiers;
+		}
+
 		void ProcessInput (Action<KeyEvent> keyHandler, Action<KeyEvent> keyDownHandler, Action<KeyEvent> keyUpHandler, Action<MouseEvent> mouseHandler)
 		{
 			int wch;
 			var code = Curses.get_wch (out wch);
 			if (code == Curses.ERR)
 				return;
+
+			keyModifiers = new KeyModifiers ();
+			Key k;
+
 			if (code == Curses.KEY_CODE_YES) {
 				if (wch == Curses.KeyResize) {
 					if (Curses.CheckWinChange ()) {
@@ -398,8 +425,8 @@ namespace Terminal.Gui {
 					mouseHandler (ToDriverMouse (ev));
 					return;
 				}
-				keyHandler (new KeyEvent (MapCursesKey (wch)));
-				keyUpHandler (new KeyEvent (MapCursesKey (wch)));
+				keyHandler (new KeyEvent (MapCursesKey (wch), keyModifiers));
+				keyUpHandler (new KeyEvent (MapCursesKey (wch), keyModifiers));
 				return;
 			}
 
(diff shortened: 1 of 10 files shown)


Terminal.Gui/Views/HexView.cs
```diff
@@ -312,9 +312,13 @@ public override bool ProcessKey (KeyEvent keyEvent)
 				MoveUp (bytesPerLine);
 				break;
 			case Key.Tab:
-				leftSide = !leftSide;
-				RedisplayLine (position);
-				firstNibble = true;
+				if (keyEvent.IsCtrl) {
```
tig, line 315: If you included this change to `HexView.cs` in this PR because you thought it fixed the "can't tab out of hex view" bug (#531), I don't think it does (I just pulled this PR down and tried it and the `Unicode` scenario still shows #531 reproduces.

Terminal.Gui/Core/Event.cs
```diff
@@ -8,6 +8,36 @@
 
 namespace Terminal.Gui {
 
+	/// <summary>
+	/// Manage the keys modifiers within a key event.
```
tig, line 12: Should be

/// Identifies the state of the "shift"-keys within a <see cref="KeyEvent"> event.

Terminal.Gui/Core/Event.cs
```diff
@@ -302,6 +332,11 @@ public class KeyEvent {
 		/// </summary>
 		public Key Key;
 
+		/// <summary>
+		/// Check if the Shift key was pressed or not.
```
tig, line 336: I suggest:

    /// Indicates the "shift" state of the various shift-keys (e.g. Shift, Alt, Ctrl, NumLock, ...).

Terminal.Gui/Core/Event.cs
```diff
@@ -313,57 +348,83 @@ public class KeyEvent {
...
 
 		/// <summary>
 		/// Gets a value indicating whether the Alt key was pressed (real or synthesized)
 		/// </summary>
 		/// <value><c>true</c> if is alternate; otherwise, <c>false</c>.</value>
-		public bool IsAlt => (Key & Key.AltMask) != 0;
+		public bool IsAlt => KeyModifiers.Alt;
 
 		/// <summary>
 		/// Determines whether the value is a control key (and NOT just the ctrl key)
```
tig, line 360: I'm not sure this comment is valid anymore.

Terminal.Gui/Core/Event.cs
```diff
@@ -313,57 +348,83 @@ public class KeyEvent {
...
+			if (KeyModifiers.Numlock) {
+				msg += "Numlock-";
+			}
+			if (KeyModifiers.Scrolllock) {
+				msg += "Scrolllock-";
 			}
+
+			msg += $"{(((uint)this.KeyValue & (uint)Key.CharMask) > 27 ? $"{(char)this.KeyValue}" : $"{key}")}";
+
 			return msg;
```
tig, line 428: I love what you've done here!

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -379,12 +379,39 @@ MouseEvent ProcessButtonState (Curses.MouseEvent cev, MouseFlags mf)
...
+			if (keyModifiers == null)
+				keyModifiers = new KeyModifiers ();
+
+			if (!keyModifiers.Shift && key.HasFlag (Key.ShiftMask))
+				keyModifiers.Shift = true;
+			if (!keyModifiers.Alt && key.HasFlag (Key.AltMask))
+				keyModifiers.Alt = true;
+			if (!keyModifiers.Ctrl && key.HasFlag (Key.CtrlMask))
+				keyModifiers.Ctrl = true;
+			//if (!keyModifiers.Capslock)
```
tig, line 395: Please remove old code if it is no longer valid. You either have confidence in your PR or you don't ;-)

BDisp: I was going to remove it but then I noticed that TextField needs it, for now. But in WindowsDriver this method is not really necessary, but I have to change the TextField to use the new features. I can do it if I want to but it can also happen that other Views use it but then it is a matter of changing it.

Terminal.Gui/Core/Event.cs
BDisp: The information is already available in both. Which do you prefer?

Terminal.Gui/Views/HexView.cs
BDisp: With ctrl + tab moves to the next View. Now in the scenario where only HexView is located, the cursor goes to a place that I no longer had the head to investigate.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
BDisp: I think it is better to keep TextField like this for now because CursesDriver does not detect the Ctrl and Shift keys directly and therefore I am not able to map properly and will break functionality. On NetDriver I haven't seriously addressed the issue yet but I will ...

Terminal.Gui/Views/HexView.cs
BDisp: @tig I think we should not break the functionality of the Tab key that `HexView `uses internally which is to move between hexadecimal code and text. As in a text editor, you press the `Tab `key to insert it into the text and to move to another `View `you will have to press `Ctrl + Tab` or `Shift + Tab` to move to the next or previous `View`, respectively.

Terminal.Gui/Core/Event.cs
BDisp: @tig I already know what to do here. I will keep both comparisons for compatibility and completely remove the modifier keys in the `Key `property of the `KeyEvent `class on `WindowsDriver`.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
BDisp: Ahh this is on `CursesDriver`. I just kept it to be the same as `WindowsDriver `but `CursesDriver `doesn't support the other features and that's why I commented. I'll delete it then.

Terminal.Gui/Views/HexView.cs
tig: On Windows, at least, if an Edit control is in a dialog, along with other controls, `Tab` tabs to next control, and `Shift-Tab` tabs to previous. You have to use `Ctrl-Tab` to enter a tab. If the Edit control is the only control, `Tab` adds a tab to the text and `Sift-Tab` does nothing. 

I think we should follow the same behavior for all views that accept input in Terminal.Gui.

BDisp: I think it is debatable because if I am in a text editor I would like to use all the available keys to work on that text and not move to other controls, but anyway ...
Of course, whoever created this control also thought of the same.

tig: > I think it is debatable because if I am in a text editor I would like to use all the available keys to work on that text and not move to other controls, but anyway ...
> Of course, whoever created this control also thought of the same.

I authored a contributors guide that has some tenets for making decisions like this. It's *ALL* debatable, but instead of debating specific examples like this in isolation, I'd prefer we debate the tenets first (using examples like this to pressure test the tenets). https://github.com/migueldeicaza/gui.cs/blob/master/CONTRIBUTING.md

Let me know what you think. Please use a new Issue if you have comments on the CONTRIBUTING.md.

PR #593, opened 2020-06-01
Changed Tab key to Enter key as per the tenets. Closes #589

### LINKED ISSUE #589: [design] Text editors should handle the Tab key and provide Ctrl+Tab to move to the next View.
If I am typing in a text editor I would like to use all the available keys that can help me in its editing and not be worried if when I look at the screen I am already typing in another `View `after all. In `Windows`, the `TextBox `with multiple lines has this behavior, that is, I have to type `Ctrl + Tab` to move to the next control. `Gui.cs` already provides `Shift + Tab` to move to the previous control. Another example is the `Tab `control which when pressed opens the next `Tab`. Now a mandatory rule is `Views `that are justified in having to really need the `Tab `key to have to code `Ctrl + Tab` to allow moving to the next control. Or even better is that we are not too worried about forgetting it is in TopLevel's ProcessKey that we consider by default the two alternatives to move to the next `View `that would be `Tab `and `Ctrl + Tab`. That said, I think that a `View `cannot take hold of both of these alternatives or there can never be an alternative to moving to the next control. Excuse the long will but I would like to put my point of view.

BDisp:


migueldeicaza:
Link to these tenets?   I missed that.

That said, this seems fine.

migueldeicaza:
Feel free to merge.

tig:
> Link to these tenets? I missed that.
> 
> That said, this seems fine.

https://github.com/migueldeicaza/gui.cs/blob/master/CONTRIBUTING.md

PR #595, opened 2020-06-02
Added mouse tasks to the idle handlers. Fixed keyModifiers.

BDisp:


migueldeicaza:
I do not understand the purpose of this change, can you explain?

BDisp:
The Post method is never called to pass the thread to the UI if the thread are created with the Task.Run. In the Threading.cs the Post get called because the async calls are all inside methods, lambdas, actions, that there are previous added to the idle handlers. It's working in the past because the timers was running due the progress bar causing a returning to the main thread. This shouldn't be needed but it was the only way I found to figure it out.

BDisp:
@migueldeicaza  see this comments to understand better what is going on.

https://github.com/migueldeicaza/gui.cs/pull/586#issuecomment-636515735

BDisp:
@tig with this PR you don't need to include anymore the `Application.MainLoop.AddTimeout` in the `Scrolling.cs` to get the continuous pressed button to working.

Issue #436, opened 2020-05-17
If StatusBar handles <key>, <key> is eaten for all Toplevels except Application.Top


tig:
Easily reproduced in the main `UI Catalog` window where opening the `File` menu, selecting `Quit` and hitting `ENTER` starts the selected scenario vs. exiting. 

This is true of any key that `StatusBar` eats: E.g. if there's a `StatusItem` for `Key.Left` left arrow on Menu won't work. 

I'm using `ENTER` this way, because `ListView` currently doesn't support `ENTER` to select an item, nor does it provide a "ItemDoubleClicked" event (#429).

BDisp:
`ENTER` key is working on `Menu`.

BDisp:
I'm going to create an `OpenSelectedItems` event for `ListView` right now. Invoked by the `ENTER` key and `DoubleClicked`.

tig:
> `ENTER` key is working on `Menu`.

Not for me. It causes the selected Scenario to run instead of exiting. Have you tried reproducing on Windows? I can repro on Linux too.

(Note I just remembered OnKeyDown only works on Windows and thus `UI Catalog` scenarios currently can't exit on Linux...)

BDisp:
Yes I did. It's working on both. I'll checkout your PR and see if it's happen too.

BDisp:
> 
> 
> I'm going to create an `OpenSelectedItems` event for `ListView` right now. Invoked by the `ENTER` key and `DoubleClicked`.

Done! #440

tig:
Pretty sure the underlying bug here is the same as what is causing #437.

tig:
@BDisp this is fixed, right?

PR #596, opened 2020-06-02
Improvements the mouse and keys of the MenuBar. 

BDisp:
The hot keys were been called to the wrong Toplevel even it isn't the current. Mouse changes was to get a better response. And there was fixed some minor bugs due the various previous interventions.

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
@@ -695,8 +695,9 @@ namespace Terminal.Gui {
 						keyUpHandler (new KeyEvent (map, keyModifiers));
 					}
 				}
-				if (!inputEvent.KeyEvent.bKeyDown)
+				if (!inputEvent.KeyEvent.bKeyDown) {
 					keyModifiers = null;
+				}
 				break;
 
 			case WindowsConsole.EventType.Mouse:
@@ -879,7 +880,7 @@ namespace Terminal.Gui {
 			};
 		}
 
-		private async Task ProcessButtonDoubleClickedAsync ()
+		async Task ProcessButtonDoubleClickedAsync ()
 		{
 			await Task.Delay (200);
 			IsButtonDoubleClicked = false;
@@ -896,11 +897,11 @@ namespace Terminal.Gui {
 				};
 
 				var view = Application.wantContinuousButtonPressedView;
-				if (view == null)
+				if (view == null) {
 					break;
+				}
 				if (IsButtonPressed && (mouseFlag & MouseFlags.ReportMousePosition) == 0) {
 					mouseHandler (me);
-					//mainLoop.Driver.Wakeup ();
 				}
 			}
 		}
--- a/Terminal.Gui/Core/Application.cs
+++ b/Terminal.Gui/Core/Application.cs
@@ -351,8 +351,9 @@ namespace Terminal.Gui {
 					OfY = me.Y - newxy.Y,
 					View = view
 				};
-				if (OutsideFrame (new Point (nme.X, nme.Y), mouseGrabView.Frame))
+				if (OutsideFrame (new Point (nme.X, nme.Y), mouseGrabView.Frame)) {
 					lastMouseOwnerView?.OnMouseLeave (me);
+				}
 				if (mouseGrabView != null) {
 					mouseGrabView.OnMouseEvent (nme);
 					return;
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -989,7 +989,7 @@ namespace Terminal.Gui {
 			if (subviews == null || subviews.Count == 0)
 				return false;
 			foreach (var view in subviews)
-				if (view.ProcessHotKey (keyEvent))
+				if (view.SuperView.IsCurrentTop && view.ProcessHotKey (keyEvent))
 					return true;
 			return false;
 		}
@@ -1004,7 +1004,7 @@ namespace Terminal.Gui {
 			if (subviews == null || subviews.Count == 0)
 				return false;
 			foreach (var view in subviews)
-				if (view.ProcessColdKey (keyEvent))
+				if (view.SuperView.IsCurrentTop && view.ProcessColdKey (keyEvent))
 					return true;
 			return false;
 		}
@@ -1024,7 +1024,7 @@ namespace Terminal.Gui {
 			if (subviews == null || subviews.Count == 0)
 				return false;
 			foreach (var view in subviews)
-				if (view.OnKeyDown (keyEvent))
+				if (view.SuperView.IsCurrentTop && view.OnKeyDown (keyEvent))
 					return true;
 
 			return false;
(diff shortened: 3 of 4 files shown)


BDisp:
I think this https://github.com/migueldeicaza/gui.cs/pull/596/commits/ce30a837642cc72aef6433e9322a6c4970c13028 fixes #436.

migueldeicaza:
We probably need a better description of what this change does on the description, it is hard to judge the changes without it.

BDisp:
Without merge with https://github.com/migueldeicaza/gui.cs/pull/596/commits/4a1f5a9309cf10d45d13c3b7d87ba7ceb8303a6a `Alt `key doesn't work in the `Menu`.

tig:
@BDisp I'm very confused about what this PR does and doesn't do. 

It makes a lot of changes that are hard to follow in the diff. Can you try to explain the problem and the fix more clearly?

BDisp:
> 
> 
> @BDisp I'm very confused about what this PR does and doesn't do.
> 
> It makes a lot of changes that are hard to follow in the diff. Can you try to explain the problem and the fix more clearly?

`OnKeyUp `and `OnKeyDown `in reality I reset them for what remains the same because my intention was to activate the `Menu `right after pressing and not when releasing the key but this is not effective because it activates right when switching between windows and leaves the `Menu `activated no need. At the mouse level, I activated both `ButtonPressed `and `ButtonDoubleClicked `to open and close the `Menu `more quickly. Now it works alternately, that is, if the `Menu `is closed both events will open and close it, otherwise. There is another commit that also solves a problem in which if you open the "`Text Editor Demo`" in the `Example `project, pressing` Ctrl + Q `would trigger the `Application.Top` Quit method that does not belong to the `Toplevel `current.

BDisp:
@tig here is a video explaining what I was meaning about the Quit that belongs to the `Application.Top` is running in the `Text Editor Demo`. If I choose "Ok" to quit after I close the editor the application exits.

toplevel-keys

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -892,6 +879,32 @@ MouseEvent ToDriverMouse (WindowsConsole.MouseEventRecord mouseEvent)
 			};
 		}
 
+		private async Task ProcessButtonDoubleClickedAsync ()
+		{
+			await Task.Delay (200);
+			IsButtonDoubleClicked = false;
+		}
+
+		async Task ProcessContinuousButtonPressedAsync (WindowsConsole.MouseEventRecord mouseEvent, MouseFlags mouseFlag)
```
tig, line 888: I like this pattern. It makes a lot of sense to me, now that I've studied it.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -892,6 +879,32 @@ MouseEvent ToDriverMouse (WindowsConsole.MouseEventRecord mouseEvent)
 			};
 		}
 
+		private async Task ProcessButtonDoubleClickedAsync ()
```
tig, line 882: Coding style says "don't use `private`" as it's the default and not needed.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -695,7 +695,8 @@ void IMainLoopDriver.MainIteration ()
 						keyUpHandler (new KeyEvent (map, keyModifiers));
 					}
 				}
-				keyModifiers = null;
+				if (!inputEvent.KeyEvent.bKeyDown)
+					keyModifiers = null;
```
tig, line 699: Please wrap all if's with `{` and `}`.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -892,6 +879,32 @@ MouseEvent ToDriverMouse (WindowsConsole.MouseEventRecord mouseEvent)
...
+			while (IsButtonPressed) {
+				await Task.Delay (200);
+				var me = new MouseEvent () {
+					X = mouseEvent.MousePosition.X,
+					Y = mouseEvent.MousePosition.Y,
+					Flags = mouseFlag
+				};
+
+				var view = Application.wantContinuousButtonPressedView;
+				if (view == null)
```
tig, line 899: Wrap `break;` in `{}`

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -892,6 +879,32 @@ MouseEvent ToDriverMouse (WindowsConsole.MouseEventRecord mouseEvent)
...
+					break;
+				if (IsButtonPressed && (mouseFlag & MouseFlags.ReportMousePosition) == 0) {
+					mouseHandler (me);
+					//mainLoop.Driver.Wakeup ();
```
tig, line 903: Can this comment be deleted?

Terminal.Gui/Core/Application.cs
```diff
@@ -352,7 +352,7 @@ static void ProcessMouseEvent (MouseEvent me)
 					View = view
 				};
 				if (OutsideFrame (new Point (nme.X, nme.Y), mouseGrabView.Frame))
-					lastMouseOwnerView.OnMouseLeave (me);
+					lastMouseOwnerView?.OnMouseLeave (me);
```
tig, line 355: Use `{}`.

Terminal.Gui/Core/View.cs
```diff
@@ -989,7 +989,7 @@ public override bool ProcessHotKey (KeyEvent keyEvent)
 			if (subviews == null || subviews.Count == 0)
 				return false;
 			foreach (var view in subviews)
-				if (view.ProcessHotKey (keyEvent))
+				if (view.SuperView.IsCurrentTop && view.ProcessHotKey (keyEvent))
```
tig, line 992: This comment is broader than this PR, but anytime I see three functions that have EXACTLY the same logic in them, with a fair amount of logic, I wonder if they could be refactored to share implementation to reduce code and risk of one of them having a bug.

ARE all 3 the same???

Terminal.Gui/Views/Menu.cs
```diff
@@ -815,6 +816,7 @@ internal void CloseMenu (bool reopen = false, bool isSubMenu = false)
 					IsMenuOpen = false;
 					PositionCursor ();
 				}
+				IsMenuOpen = false;
```
tig, line 819: Both branches of the `if` above set `IsMenuOpen = false`. Why is this needed?

Terminal.Gui/Core/View.cs
BDisp: These processes have different behaviours but with the same logic. They were already there and I don't wanna break anything. Later we can see if it will be possible to congregate all of them.

Terminal.Gui/Views/Menu.cs
BDisp: You're right. My bad :-)

tig:
You are awesome. Thank you!

PR #605, opened 2020-06-03
Fixes #604 Toplevel repaint.

### LINKED ISSUE #604: Toplevel repaint issue.
There are some cases where we only need to clear a certain region of the screen. We don't need clear all screen if it only a minor part needed to repaint. So I think that the `Redraw `methods had a region as parameter because it could be any frame dimensions we want to pass to it. So in the image you see that the `TextField `at the bottom and the `StatusBar `disapear and there are not redrawed again while dragging. Without these features it's not worth it developing dragging at all. I think that my original suggestion to resolve this is better.

background

BDisp:


BDisp:
Even if we have to repaint the entire window to repaint a small fraction?

tig:
Would you please add this comment in `Window.MouseEvent` so future contributors understand what is going on here:

```csharp
// Redraw the entire app window using just our Frame. Since we are 
// Application.Top, and our Frame always == our Bounds (Location is always (0,0))
// our Frame is actually view-relative (which is what Redraw takes).
Application.Top.Redraw (Frame);
```

tig:
Note, that this PR does not fix situations where dragging Windows doesn't repaint when the `Toplevel` that is underneath is NOT `Application.Top`.

`Windows & FrameViews` illustrates this. 



But this PR fixes the specific issue called out, so we should merge it and open another for this other issue.

Terminal.Gui/Core/Window.cs
```diff
@@ -201,7 +201,7 @@ public override bool MouseEvent (MouseEvent mouseEvent)
 				if (dragPosition.HasValue) {
 					if (SuperView == null) {
 						Application.Top.SetNeedsDisplay (Frame);
-						Application.Top.Redraw (Bounds);
+						Application.Top.Redraw (Frame);
```
tig, line 204: Please add this comment

```
// Redraw the entire app window using just our Frame. Since we are 
// Application.Top, and our Frame always == our Bounds (Location is always (0,0))
// our Frame is actually view-relative (which is what Redraw takes).
Application.Top.Redraw (Frame);
```

BDisp: Done!

tig:
Nice work. Sorry about pushing back so hard; I hope you understand why.

BDisp:
> 
> 
> Nice work. Sorry about pushing back so hard; I hope you understand why.

Never mind. We'll see that after. Thanks.

PR #607, opened 2020-06-03
ComboBox. Support parameterless constructor. Update scenario demo

fergusonr:


tig:
Nice work.

See: https://github.com/migueldeicaza/gui.cs/issues/447

I was just reviewing all event code throughout the project and noted this:

- `public event EventHandler<ustring> Changed;`
   - The implementation has some issues: 
             1. The name is "Changed" but the event actually represents "the selection has been confirmed". Rename it to `SelectionChanged`?
             2. It's sending the current text; best practice is for an event to provide previous value. I'm not sure which is more appropriate for this API, but if you KNOW users of the event are going to want the 'new' text, then keep it as is. If you are not sure, then my suggestion is to defined a `SelectionChangedEventArgs` class and include both the old and new value in it.
             2. Calls to `Invoke` are spread around. Ideally there would be `public virtual void OnSelectionChanged()` method that calls `SelectionChanged?.Invoke`.

PR #609, opened 2020-06-04
Fixes #503 Tab stop working.

### LINKED ISSUE #503: UI Catalog: Something in Computed Layout scenario is causing TAB to stop working
1. Run UI Catalog
2. Start `Computed Layout`
3. Focused control will be "[At(1,1)]". 
4. Press TAB. Cursor moves into Text Alignment window.
5. Press TAB.  Nothing happens.

TAB now no longer works. 

6. Click on another button (e.g. [Center]). 
7. Press TAB.
8. TAB works, until it gets back into Text Alignment. Then breaks again.

BDisp:


Terminal.Gui/Core/Toplevel.cs
```diff
@@ -161,16 +165,47 @@ public override bool ProcessKey (KeyEvent keyEvent)
...
 
 			case Key.ControlL:
 				Application.Refresh ();
 				return true;
+
 			}
 			return false;
 		}
 
+		IEnumerable<View> GetSuperViewSubviews (bool isForward)
```
tig, line 181: Is this the best name for this method? If I understand what you are doing here (and I'm not sure I do), you are reversing the order of the subviews. Why not call it `GetSubviewsReverseOrder`?

Have you considered how we may want views to be able to set an explicit tab order in the future? I'm not sure it's needed, but Windows has always had a tab order that was distinct from creation order (`[Control.TabIndex](https://docs.microsoft.com/en-us/dotnet/framework/winforms/controls/how-to-set-the-tab-order-on-windows-forms)`). 

Regardless the idea of using a HashSet to efficiently get the list in reverse order is brilliant. Well done.

BDisp: It really didn't sound right to me, but due to the lack of imagination on my part it went like this. I switched to `GetToplevelSubviews `because the method just reverses the order of the subviews depending on whether isForward is true or not, so the name you suggested would not be entirely correct.
`return isForward ? views : views.Reverse ();`

`TabIndex`'s suggestion is really great and we could think about it but for now this PR solves the issue submitted by you.

Thanks for the compliment, but it was nothing too much because `View `implements this in the `TopologicalSort `method.

Terminal.Gui/Core/Toplevel.cs
```diff
@@ -171,7 +177,35 @@ public override bool ProcessKey (KeyEvent keyEvent)
 			return false;
 		}
 
-		///<inheritdoc/>
+		IEnumerable<View> GetToplevelSubviews (bool isForward)
```
tig, line 180: This crashes if the Toplevel is a dialog and the user presses Tab.

BDisp: Oh sorry. I already fixed.

BDisp:
Fixed here https://github.com/migueldeicaza/gui.cs/pull/615

PR #625, opened 2020-06-05
Add .NET Framework ref assemblies to enable linux build

### LINKED ISSUE #421: How to build on Linux?
I thought I knew how to build dotnet stuff on Linux, but obviously not.

I've got Ubuntu 18.04 running in WSL.

I've got dotnet installed. 

When I do:

```
dotnet restore
dotnet build
``` 

I get:
image

I didn't think you could actually install .NET 4.7.2 on linux. If you can, I don't know how. Hence me asking here.

AArnott:
Fixes #421

### DIFF
--- a/Terminal.Gui/Terminal.Gui.csproj
+++ b/Terminal.Gui/Terminal.Gui.csproj
@@ -169,6 +169,7 @@
     <DebugType></DebugType>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="true" />
     <PackageReference Include="NStack.Core" Version="0.14.0" />
   </ItemGroup>
   <ItemGroup>


tig:
Withi this, can we remove `net472` from: `<TargetFrameworks>net472;netstandard2.0</TargetFrameworks>` now?

I'm eager to do that to enable better CI/CD builds.

AArnott:
@tig: this change enables us to _keep_ `net472`. If we were to remove that, we wouldn't need my PR anyway.
.NET Framework is still very much a popular target framework. IMO most projects should continue to support and test that.
How would removing `net472` enable better CI/CD builds? I suspect we can find a way to get everything we want without dropping net472 support.

migueldeicaza:
Will give it a shot!   Hope it does not break the Mac build with the .NET framework profile :-)

BDisp:
@AArnott I think you are referring that if net472 is not configured it will not work in mono. Example is currently targeting netcore3.1 and does not run in mono, as it gives an error of "File does not contain a valid CIL image." I do not use it practically but for those who use it regularly it is a break. @migueldeicaza if that is the case it will break your profile :-)

AArnott:
@BDisp I don't understand what you're getting at. I'm not sure what you mean by "if net472 is not configured". 
As I understand it, mono doesn't have any problem building or running net472 as it is. But `dotnet build` fails to build net472 targeted projects except on Windows, unless the content of this PR is present. Since I prefer to use `dotnet build` over anything else, this PR allows dotnet commands to "just work".

I don't know anything specifically about how netcoreapp3.1 compiled assemblies work or fail on mono. But I don't expect my change would impact that in any way.

BDisp:
@AArnott Sorry I messed up on another subject.

PR #627, opened 2020-06-05
Added TextAlignment to the Button. Update FSarpExample.

BDisp:


FSharpExample/Program.fs
```diff
@@ -16,40 +16,40 @@ type Demo() = class end
     let mutable menuKeysStyle = Unchecked.defaultof<CheckBox>
     let mutable menuAutoMouseNav = Unchecked.defaultof<CheckBox>
 
-    type Box10x() =
-        inherit View()
-        member val w = 40 with get, set
-        member val h = 50 with get, set
+    type Box10x(x : int, y : int) =
```
tig, line 19: Please try to not combine PRs. This PR should be separate from the Button one.

Terminal.Gui/Views/Button.cs
```diff
@@ -176,24 +186,84 @@ internal void Update ()
...
 		///<inheritdoc/>
 		public override void Redraw (Rect bounds)
 		{
 			Driver.SetAttribute (HasFocus ? ColorScheme.Focus : ColorScheme.Normal);
 			Move (0, 0);
-			Driver.AddStr (shown_text);
 
-			if (hot_pos != -1) {
-				Move (hot_pos, 0);
+			var caption = shown_text;
```
tig, line 197: The text alignment code is complex and we should not duplicate it. In addition, by duplicating this code we hinder out ability to eventually implement line wrap in a unified way, which is a feature that's been requested.

One option would be to add a `Text` property to `View` and move all text alignment down there. Remove all special `Text` handling from all subsclasses where possible. 

Another option would be to implement an interface `IViewHasText` and provide a default implementation.

BDisp: Well, this is an ambitious and great idea. But it will also be necessary to deal with hotkeys at the `View `level (I think you are referring to the `View `class itself, right?) I will need more time to implement it. However, I thought it best to accept this for now and as soon as I have something ready I will find a way to implement it in all the controls that need it.

PR #628, opened 2020-06-05
Added DateTimeEventArgs and improved data and time validation.

BDisp:


grahamehorner:
I would recommend changing the Time backing field from a DateTime to TimeSpan

BDisp:
> 
> 
> I would recommend changing the Time backing field from a DateTime to TimeSpan

@grahamehorner It has already been done.. Thank you for your recommendation.

Terminal.Gui/Views/DateField.cs
```diff
@@ -374,25 +373,25 @@ public override bool MouseEvent (MouseEvent ev)
...
-			DateChanged?.Invoke (this, args);
+			DateChanged?.Invoke (args);
 		}
 	}
 
 	/// <summary>
 	/// Handled the <see cref="EventArgs"/> for <see cref="DateField"/> or <see cref="TimeField"/> events.
 	/// </summary>
-	public class DateTimeEventArgs : EventArgs {
+	public class DateTimeEventArgs<T> : EventArgs {
```
tig, line 385: Can you explain why you used `DateTimeEventArgs<T>` here instead of `DateTimeEventArgs<TimeSpan>` here?

Terminal.Gui/Views/TimeField.cs
```diff
@@ -96,7 +95,7 @@ void TimeField_Changed (object sender, ustring e)
 		/// </summary>
 		/// <remarks>
 		/// </remarks>
-		public DateTime Time {
+		public TimeSpan Time {
```
tig, line 98: What's the logic for using `TimeSpan1` vs. `DateTime` here?

Terminal.Gui/Views/DateField.cs
BDisp: Because this `DateTimeEventArgs `is being used simultaneously by `DateField `and `TimeField `and are of different `types`.

Terminal.Gui/Views/TimeField.cs
BDisp: Because as suggested by @grahamehorner if control is only over `Time`, it doesn't make sense to have the `Date`.

BDisp:
> 
> 
> I don't understand the move from `DateTime` to `TimeSpan`.

@tig I don't think it makes much sense for a `Time `control to present the date as it was initially happening because both `DateField `and `TimeField `are both `DateTime `types. Also as `TimeSpan `it is useful in cases where `TimeField `is used as a timer, timeouts, elapsed time, etc

PR #632, opened 2020-06-06
Changed from Terminal.Gui all EventHandler to Action.

BDisp:


Terminal.Gui/Core/View.cs
```diff
@@ -124,27 +124,27 @@ internal enum Direction {
...
 		/// Event fired when the view looses focus.
 		/// </summary>
-		public event EventHandler<FocusEventArgs> Leave;
+		public Action<FocusEventArgs> Leave;
 
 		/// <summary>
 		/// Event fired when the view receives the mouse event for the first time.
 		/// </summary>
-		public event EventHandler<MouseEventEventArgs> MouseEnter;
+		public Action<MouseEventEventArgs> MouseEnter;
```
tig, line 137: Is it just me, or does `MouseEventEventArgs` seem silly? Should we just make these `MouseEventArgs`?

tig: BTW, I know I'm the one who did this ;-)

Terminal.Gui/Core/View.cs
```diff
@@ -853,7 +853,7 @@ public override bool OnEnter ()
 		public override bool OnLeave ()
 		{
 			FocusEventArgs args = new FocusEventArgs ();
-			Leave?.Invoke (this, args);
+			Leave?.Invoke (args);
```
tig, line 856: Just noticed.... why is this logic in `View` and not in the base (`Responder`)? 

Is there an intentional reason for this? It seems more logical to me that since `Responder` defines `OnLeave` (etc...) it should be the one invoking the event. Am I missing something?

Terminal.Gui/Core/View.cs
BDisp: `MouseEventArgs `really looks much better :-)

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -22,7 +22,7 @@ public class ComboBox : View {
 		///   Client code can hook up to this event, it is
 		///   raised when the selection has been confirmed.
 		/// </remarks>
-		public event EventHandler<ustring> Changed;
+		public Action<ustring> Changed;
```
tig, line 25: Let's use this opportunity to name things consistently. In my opinion any event describing when a selected has changed, where the thing changed has an index, the event name should be `SelectedItemChanged`. 

- `ListView`
- `ComboBox`
- `RadioGroup`

Should all be consistent.

Terminal.Gui/Views/ListView.cs
```diff
@@ -307,12 +307,12 @@ public override void Redraw (Rect bounds)
 		/// <summary>
 		/// This event is raised when the selected item in the <see cref="ListView"/> has changed.
 		/// </summary>
-		public event EventHandler<ListViewItemEventArgs> SelectedChanged;
+		public Action<ListViewItemEventArgs> SelectedChanged;
```
tig, line 310: `SelectedItemChanged`

Terminal.Gui/Core/View.cs
BDisp: In `Responder `there is no virtual method that invokes an `Action`.

Terminal.Gui/Views/TextField.cs
```diff
@@ -41,7 +41,7 @@ public class TextField : View {
 		/// <remarks>
 		///   The passed <see cref="EventArgs"/> is a <see cref="ustring"/> containing the old value. 
 		/// </remarks>
-		public event EventHandler<ustring> Changed;
+		public Action<ustring> Changed;
```
tig, line 44: `TextChanged` (to be consistent with `TextView`)?

Terminal.Gui/Views/ComboBox.cs
BDisp: But wouldn't that break with code already made by users?

Terminal.Gui/Views/ListView.cs
BDisp: But wouldn't that break with code already made by users?

Terminal.Gui/Views/TextField.cs
BDisp: But wouldn't that break with code already made by users?

tig: This already does, right?

BDisp:
> 
> 
> Man, this is really good work. We are going to all appreciate the consistency in the future!
> 
> Given we are inflicting breaking changes here, I think we should use this as an opportunity to improve consistency on how events are named across Views. I've noted a few. Do you think you can update this PR so we don't have to change user code more than once?

In that case, that's fine. I will make the changes suggested by you immediately.

Terminal.Gui/Views/TextField.cs
BDisp: No.

BDisp: But if it is to break it will have to be now because we should not be constantly changing the names of public variables constantly. It is necessary to have the prior approval of @migueldeicaza so we don't move back and forth.

Terminal.Gui/Core/View.cs
BDisp: In the `Responder `have this:

```
		public virtual bool OnLeave ()
		{
			return false;
		}

```

BDisp:
@tig would like you to confirm whether or not I could make the changes you requested, please.

PR #658, opened 2020-06-09
ScrollView enhancements.

BDisp:


BDisp:
Ok @tig. In the `UICatalog `project maybe you don't see any differences but in the `Example `project on the vertical bar the `TopTee `not come first but the `Diamond `instead. In the horizontal bar the content is not totally moved to the left. I think it more abrang to more cases, because depending of the size of the contents t

tig:
Didn't I already accept this?

PR #679, opened 2020-06-12
Removing symbols from MessageBox.

BDisp:


tig:
This is in #669.

BDisp:
Sorry, @tig, I didn't notice that.

tig:
You were on the right path, but here's how I think about it:

- Code that line wraps, e.g. in Edit, Label, etc... should do the right thing w.r.t. \n and \r\n and never end up calling Driver.AddRune with those chars in normal processing. 
- Anytime such a char DOES get sent down the line to Driver.AddRune, #669 should take care of replacing them with glyphs (which will still just be squares on many systems without good fonts). 

So while a function like you've introduced might be nice to have it shouldn't be necessary.

PR #680, opened 2020-06-12
Intended to improve a more faster mouse.

BDisp:
- As you can see in the `Mouse `scenario, if we click very fast we will get more double clicked and triple clicked events, so I added `ButtonTripleClicked `to `MenuBar`.
- I disabled the option that prevented the `ButtonClicked `event if the current position of `X `and `Y `were different from the previous one and thus allow that event to occur.


tig:
Can you please elaborate? "more faster mouse" could mean any one of a dozen things. It's important that PRs and Issues are as descriptive as possible not just for us, now, but for future contributors and maintainers. Thanks.

BDisp:
> 
> 
> Can you please elaborate? "more faster mouse" could mean any one of a dozen things. It's important that PRs and Issues are as descriptive as possible not just for us, now, but for future contributors and maintainers. Thanks.

I add a description. Thanks.

PR #681, opened 2020-06-13
Faster ScrollView. Try it with CharacterMap.

BDisp:


tig:
Did you change something that would cause the `end` key to stop working in scrollview?

BDisp:
> 
> 
> Did you change something that would cause the `end` key to stop working in scrollview?

No. I think it not working before too.

PR #684, opened 2020-06-13
Fixes ScrollDown that was not working.

BDisp:


tig:
Not clear what this actually fixes. Can you change the title of this Issue to actually say what it fixes?

BDisp:
You asking here https://github.com/migueldeicaza/gui.cs/pull/681#issuecomment-643625748 if I change something that would cause the end key to stop working in scrollview? I only put the link to that.

BDisp:
Fixes #689

PR #697, opened 2020-06-16
Closes #683 Resizing issue and Unix change colors.

### LINKED ISSUE #683: UICatalog does not render well when it is in full screen.
full-screen1
full-screen2
full-screen3

BDisp:
- This had already been fixed in #551 but came back.
- It also corrects the color change in Unix.
- The driver can only be closed when closing the application. In Unix this caused the OS to reset the color to its default.


migueldeicaza:

PR #704, opened 2020-06-18
Fixes #703 Label layout issue.

### LINKED ISSUE #703: Layout issue in the Label.
In my opinion I think that the `LayoutSubviews ()` method should not be virtual because it is essential for the redisposition of all views and any overriden may forget to call the base. But for now I just put the Label to call the base method. What is your opinion?

layout-issue

BDisp:


tig:
suggest you pull down my #702 first. You'll see this is already fixed because Label no longer has any code in it. I think.

BDisp:
I will. Thanks.

BDisp:
Closing this because @tig  suggestion is better. Thanks.

BDisp:
> 
> 
> suggest you pull down my #702 first. You'll see this is already fixed because Label no longer has any code in it. I think.

Confirmed. Thank you.

PR #710, opened 2020-06-20
ComboBox multiple fixes

### LINKED ISSUE #709: ComboBox issues
Discovered during extended testing:-

* Dim.Fill() incorrectly calculated

Both ListView and ComboBox have ```Height = Dim.Fill(2)```

Dim Fill()

* List height not resizing

Resizing window, list does not fill Height.

resize

* SetSource() not immediately updating list. 

If ```comboBox.SetSource (items);``` is called a second time after initial construction, the list does not instantly change

* Double click not selecting item

Double clicking the list should be the same as hitting the enter key

fergusonr:
FIxes for #709 

* Dim.Fill() incorrectly calculated
* List height not resizing
* SetSource() not immediately updating list. 
* Double click not selecting item

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1325,8 +1325,10 @@ namespace Terminal.Gui {
 					_x = x.Anchor (hostFrame.Width);
 				if (width == null)
 					w = hostFrame.Width;
+				else if (width is Dim.DimFactor && !((Dim.DimFactor)width).IsFromRemaining ())
+					w = width.Anchor (hostFrame.Width);
 				else
-					w = width.Anchor (hostFrame.Width - _x);
+					w = Math.Max (width.Anchor (hostFrame.Width - _x), 0);
 			}
 
 			if (y is Pos.PosCenter) {
@@ -1342,8 +1344,10 @@ namespace Terminal.Gui {
 					_y = y.Anchor (hostFrame.Height);
 				if (height == null)
 					h = hostFrame.Height;
+				else if (height is Dim.DimFactor && !((Dim.DimFactor)height).IsFromRemaining ())
+					h = height.Anchor (hostFrame.Height);
 				else
-					h = height.Anchor (hostFrame.Height - _y);
+					h = Math.Max (height.Anchor (hostFrame.Height - _y), 0);
 			}
 			Frame = new Rect (_x, _y, w, h);
 		}
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -125,6 +125,16 @@ namespace Terminal.Gui {
 		TextFormatter viewText;
 
 		/// <summary>
+		/// Event fired when a subview is being added to this view.
+		/// </summary>
+		public Action<View> Adding;
+
+		/// <summary>
+		/// Event fired when a subview is being removed from this view.
+		/// </summary>
+		public Action<View> Removing;
+
+		/// <summary>
 		/// Event fired when the view gets focus.
 		/// </summary>
 		public Action<FocusEventArgs> Enter;
@@ -552,6 +562,7 @@ namespace Terminal.Gui {
 				subviews = new List<View> ();
 			subviews.Add (view);
 			view.container = this;
+			OnAdding (view);
 			if (view.CanFocus)
 				CanFocus = true;
 			SetNeedsLayout ();
@@ -596,6 +607,7 @@ namespace Terminal.Gui {
 			if (view == null || subviews == null)
 				return;
 
+			OnRemoving (view);
 			SetNeedsLayout ();
 			SetNeedsDisplay ();
 			var touched = view.Frame;
@@ -931,6 +943,24 @@ namespace Terminal.Gui {
 			/// Its important to set this value to true specially when updating any View's layout from inside the subscriber method.
 			/// </summary>
 			public bool Handled { get; set; }
+		}
+
+		/// <summary>
+		/// Method invoked  when a subview is being added to this view.
+		/// </summary>
+		/// <param name="view">The subview being added.</param>
+		public virtual void OnAdding (View view)
+		{
+			view.Adding?.Invoke (this);
+		}
+
+		/// <summary>
+		/// Method invoked when a subview is being removed from this view.
+		/// </summary>
+		/// <param name="view">The subview being removed.</param>
+		public virtual void OnRemoving (View view)
+		{
+			view.Removing?.Invoke (this);
 		}
 
 		/// <inheritdoc/>
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -127,7 +127,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Event fired when a subview is being added to this view.
 		/// </summary>
-		public Action<View> Adding;
+		public Action<View> Added;
 
 		/// <summary>
 		/// Event fired when a subview is being removed from this view.
@@ -562,7 +562,7 @@ namespace Terminal.Gui {
 				subviews = new List<View> ();
 			subviews.Add (view);
 			view.container = this;
-			OnAdding (view);
+			OnAdded (view);
 			if (view.CanFocus)
 				CanFocus = true;
 			SetNeedsLayout ();
@@ -949,9 +949,9 @@ namespace Terminal.Gui {
 		/// Method invoked  when a subview is being added to this view.
 		/// </summary>
 		/// <param name="view">The subview being added.</param>
-		public virtual void OnAdding (View view)
+		public virtual void OnAdded (View view)
 		{
-			view.Adding?.Invoke (this);
+			view.Added?.Invoke (this);
 		}
 
 		/// <summary>
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -132,7 +132,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Event fired when a subview is being removed from this view.
 		/// </summary>
-		public Action<View> Removing;
+		public Action<View> Removed;
 
 		/// <summary>
 		/// Event fired when the view gets focus.
@@ -607,12 +607,12 @@ namespace Terminal.Gui {
 			if (view == null || subviews == null)
 				return;
 
-			OnRemoving (view);
 			SetNeedsLayout ();
 			SetNeedsDisplay ();
 			var touched = view.Frame;
 			subviews.Remove (view);
 			view.container = null;
+			OnRemoved (view);
 
 			if (subviews.Count < 1)
 				this.CanFocus = false;
@@ -958,9 +958,9 @@ namespace Terminal.Gui {
 		/// Method invoked when a subview is being removed from this view.
 		/// </summary>
 		/// <param name="view">The subview being removed.</param>
-		public virtual void OnRemoving (View view)
+		public virtual void OnRemoved (View view)
 		{
-			view.Removing?.Invoke (this);
+			view.Removed?.Invoke (this);
 		}
 
 		/// <inheritdoc/>
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // Authors:
 //   Miguel de Icaza (miguel@gnome.org)
 //
@@ -13,32 +13,11 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using NStack;
 
 namespace Terminal.Gui {
-	/// <summary>
-	/// Text alignment enumeration, controls how text is displayed.
-	/// </summary>
-	public enum TextAlignment {
-		/// <summary>
-		/// Aligns the text to the left of the frame.
-		/// </summary>
-		Left,
-		/// <summary>
-		/// Aligns the text to the right side of the frame.
-		/// </summary>
-		Right,
-		/// <summary>
-		/// Centers the text in the frame.
-		/// </summary>
-		Centered,
-		/// <summary>
-		/// Shows the text as justified text in the frame.
-		/// </summary>
-		Justified
-	}
-
 	/// <summary>
 	/// Determines the LayoutStyle for a view, if Absolute, during LayoutSubviews, the
 	/// value from the Frame will be used, if the value is Computed, then the Frame
@@ -131,7 +110,8 @@ namespace Terminal.Gui {
 	///    frames for the vies that use <see cref="LayoutStyle.Computed"/>.
 	/// </para>
 	/// </remarks>
-	public class View : Responder, IEnumerable {
+	public partial class View : Responder, IEnumerable {
+
 		internal enum Direction {
 			Forward,
 			Backward
@@ -141,6 +121,8 @@ namespace Terminal.Gui {
 		View container = null;
 		View focused = null;
 		Direction focusDirection;
+
+		TextFormatter viewText;
 
 		/// <summary>
 		/// Event fired when the view gets focus.
@@ -166,6 +148,16 @@ namespace Terminal.Gui {
 		/// Event fired when a mouse event is generated.
 		/// </summary>
 		public Action<MouseEventArgs> MouseClick;
+
+		/// <summary>
+		/// Gets or sets the HotKey defined for this view. A user pressing HotKey on the keyboard while this view has focus will cause the Clicked event to fire.
+		/// </summary>
+		public Key HotKey { get => viewText.HotKey; set => viewText.HotKey = value; }
+
+		/// <summary>
+		/// Gets or sets the specifier character for the hotkey (e.g. '_'). Set to '\xffff' to disable hotkey support for this View instance. The default is '\xffff'. 
+		/// </summary>
+		public Rune HotKeySpecifier { get => viewText.HotKeySpecifier; set => viewText.HotKeySpecifier = value; }
 
 		internal Direction FocusDirection {
 			get => SuperView?.FocusDirection ?? focusDirection;
@@ -389,29 +381,100 @@ namespace Terminal.Gui {
 		/// </remarks>
 		public View (Rect frame)
 		{
+			viewText = new TextFormatter ();
+			this.Text = ustring.Empty;
+
 			this.Frame = frame;
-			CanFocus = false;
 			LayoutStyle = LayoutStyle.Absolute;
 		}
 
 		/// <summary>
-		/// Initializes a new instance of <see cref="LayoutStyle.Computed"/> <see cref="View"/> class.
+		///   Initializes a new instance of <see cref="View"/> using <see cref="LayoutStyle.Computed"/> layout.
 		/// </summary>
 		/// <remarks>
+		/// <para>
 		///   Use <see cref="X"/>, <see cref="Y"/>, <see cref="Width"/>, and <see cref="Height"/> properties to dynamically control the size and location of the view.
-		/// </remarks>
-		/// <remarks>
+		///   The <see cref="Label"/> will be created using <see cref="LayoutStyle.Computed"/>
+		///   coordinates. The initial size (<see cref="View.Frame"/> will be 
+		///   adjusted to fit the contents of <see cref="Text"/>, including newlines ('\n') for multiple lines. 
+		/// </para>
+		/// <para>
+		///   If <c>Height</c> is greater than one, word wrapping is provided.
+		/// </para>
+		/// <para>
 		///   This constructor intitalize a View with a <see cref="LayoutStyle"/> of <see cref="LayoutStyle.Computed"/>. 
 		///   Use <see cref="X"/>, <see cref="Y"/>, <see cref="Width"/>, and <see cref="Height"/> properties to dynamically control the size and location of the view.
+		/// </para>
 		/// </remarks>
-		public View ()
+		public View () : this (text: string.Empty) { }
+
+
+		/// <summary>
+		///   Initializes a new instance of <see cref="View"/> using <see cref="LayoutStyle.Absolute"/> layout.
+		/// </summary>
+		/// <remarks>
+		/// <para>
+		///   The <see cref="View"/> will be created at the given
+		///   coordinates with the given string. The size (<see cref="View.Frame"/> will be 
+		///   adjusted to fit the contents of <see cref="Text"/>, including newlines ('\n') for multiple lines. 
+		/// </para>
+		/// <para>
+		///   No line wrapping is provided.
+		/// </para>
+		/// </remarks>
+		/// <param name="x">column to locate the Label.</param>
+		/// <param name="y">row to locate the Label.</param>
+		/// <param name="text">text to initialize the <see cref="Text"/> property with.</param>
+		public View (int x, int y, ustring text) : this (TextFormatter.CalcRect (x, y, text), text) { }
+
+		/// <summary>
+		///   Initializes a new instance of <see cref="View"/> using <see cref="LayoutStyle.Absolute"/> layout.
+		/// </summary>
+		/// <remarks>
+		/// <para>
+		///   The <see cref="View"/> will be created at the given
+		///   coordinates with the given string. The initial size (<see cref="View.Frame"/> will be 
+		///   adjusted to fit the contents of <see cref="Text"/>, including newlines ('\n') for multiple lines. 
+		/// </para>
+		/// <para>
+		///   If <c>rect.Height</c> is greater than one, word wrapping is provided.
+		/// </para>
+		/// </remarks>
+		/// <param name="rect">Location.</param>
+		/// <param name="text">text to initialize the <see cref="Text"/> property with.</param>
+		public View (Rect rect, ustring text) : this (rect)
 		{
+			viewText = new TextFormatter ();
+			this.Text = text;
+		}
+
+		/// <summary>
+		///   Initializes a new instance of <see cref="View"/> using <see cref="LayoutStyle.Computed"/> layout.
+		/// </summary>
+		/// <remarks>
+		/// <para>
+		///   The <see cref="View"/> will be created using <see cref="LayoutStyle.Computed"/>
+		///   coordinates with the given string. The initial size (<see cref="View.Frame"/> will be 
+		///   adjusted to fit the contents of <see cref="Text"/>, including newlines ('\n') for multiple lines. 
+		/// </para>
+		/// <para>
+		///   If <c>Height</c> is greater than one, word wrapping is provided.
+		/// </para>
+		/// </remarks>
+		/// <param name="text">text to initialize the <see cref="Text"/> property with.</param>
+		public View (ustring text) : base ()
+		{
+			viewText = new TextFormatter ();
+			this.Text = text;
+
 			CanFocus = false;
 			LayoutStyle = LayoutStyle.Computed;
+			// BUGBUG: CalcRect doesn't account for line wrapping
+			var r = TextFormatter.CalcRect (0, 0, text);
 			x = Pos.At (0);
 			y = Pos.At (0);
-			Height = 0;
-			Width = 0;
+			Width = r.Width;
+			Height = r.Height;
 		}
 
 		/// <summary>
@@ -432,6 +495,7 @@ namespace Terminal.Gui {
 			if (SuperView == null)
 				return;
 			SuperView.SetNeedsLayout ();
+			viewText.SetNeedsFormat ();
 		}
 
 		/// <summary>
@@ -763,21 +827,24 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Utility function to draw strings that contain a hotkey.
 		/// </summary>
-		/// <param name="text">String to display, the underscoore before a letter flags the next letter as the hotkey.</param>
+		/// <param name="text">String to display, the hotkey specifier before a letter flags the next letter as the hotkey.</param>
 		/// <param name="hotColor">Hot color.</param>
 		/// <param name="normalColor">Normal color.</param>
 		/// <remarks>
-		/// The hotkey is any character following an underscore ('_') character.</remarks>
+		/// <para>The hotkey is any character following the hotkey specifier, which is the underscore ('_') character by default.</para>
+		/// <para>The hotkey specifier can be changed via <see cref="HotKeySpecifier"/></para>
+		/// </remarks>
 		public void DrawHotString (ustring text, Attribute hotColor, Attribute normalColor)
 		{
-			Driver.SetAttribute (normalColor);
+			var hotkeySpec = HotKeySpecifier == (Rune)0xffff ? (Rune)'_' : HotKeySpecifier;
+			Application.Driver.SetAttribute (normalColor);
 			foreach (var rune in text) {
-				if (rune == '_') {
-					Driver.SetAttribute (hotColor);
+				if (rune == hotkeySpec) {
+					Application.Driver.SetAttribute (hotColor);
 					continue;
 				}
-				Driver.AddRune (rune);
-				Driver.SetAttribute (normalColor);
+				Application.Driver.AddRune (rune);
+				Application.Driver.SetAttribute (normalColor);
 			}
 		}
 
@@ -819,8 +886,13 @@ namespace Terminal.Gui {
 		{
 			if (focused != null)
 				focused.PositionCursor ();
-			else
-				Move (frame.X, frame.Y);
+			else {
+				if (CanFocus && HasFocus) {
+					Move (viewText.HotKeyPos == -1 ? 1 : viewText.HotKeyPos, 0);
+				} else {
+					Move (frame.X, frame.Y);
+				}
+			}
 		}
 
 		/// <inheritdoc/>
@@ -971,6 +1043,15 @@ namespace Terminal.Gui {
 		{
 			var clipRect = new Rect (Point.Empty, frame.Size);
 
+			Driver.SetAttribute (HasFocus ? ColorScheme.Focus : ColorScheme.Normal);
+
+			if (!ustring.IsNullOrEmpty (Text)) {
+				Clear ();
+				// Draw any Text
+				viewText?.SetNeedsFormat ();
+				viewText?.Draw (ViewToScreen (Bounds), HasFocus ? ColorScheme.Focus : ColorScheme.Normal, HasFocus ? ColorScheme.HotFocus : ColorScheme.HotNormal);
+			}
+
 			// Invoke DrawContentEvent
 			OnDrawContent (bounds);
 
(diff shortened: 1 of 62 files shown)


BDisp:
@fergusonr, provide a functional constructor for the `AllViewsTester.cs`. Now there is an error on the `searchset `variable is null in this method.
```

		private int CalculatetHeight ()
		{
			var h = (Height is Dim.DimAbsolute) ? height : Bounds.Height;
			return Math.Min (Math.Max (0, h - 1), searchset.Count);
		}

```
In the `Demo.cs` the `ListView `is not closing on exit.

fergusonr:
Thanks. Fixed AllViewsTester.cs. Investigating Demo.cs issue....

BDisp:
Thanks. Is possible you have the default constructor to accept a ustring and using that to have a search and source filled with that information making visible? It's the way that `AllViewsTester.cs` work. See the `CheckBox `example.

BDisp:
Thanks. It's possible to have that text also into the source?

fergusonr:
The text (what your searching for) is independent to the source (your searching in)

I think this, added to AllViewTesters.cs  CreateClass() would be better for the IUCatalog, but it does not seem to initialize a ListView or ComboBox. Even though I can breakpoint in SetSource() and see its being set. 

```csharp
if(view.GetType().GetMethod("SetSource") != null) {
	IList<string> list = new List<string>() { "Test Text1", "Test Text2", "Test Text3" };
	view?.GetType ().GetMethod ("SetSource")?.Invoke (view, new object [] { list });
}
```

BDisp:
I think you are wright. It seems it's not initializing in the `ListView `which must be fixed first. I also think that adding a `Text `property in the `ListView `with the default string selected would do (may be) the trick.

fergusonr:
I can now get ListView and ComboBox scenario lists to initialize with this code change in addition to the one mentioned above in AllViewTesters.cs CreateClass

Need @tig to comment on the validity of making this change

```csharp
	// Instantiate view
	var view = (View)Activator.CreateInstance (type);

	view.Width = Dim.Fill (); // <-- Added
	view.Height = Dim.Fill (); // <-- Added
```

InitializeLists

BDisp:
I will submit a PR to the `ListView `related and then with that please try if it could have a default search with the default source based on that search.

fergusonr:
Not sure why you need the constructor your asking for? It has no use outside of testing. The same can be achieved with this

```csharp
var combo = new ComboBox ("Test Text")
combo.SetSource(new List<string>() { "Test Text" });
```

BDisp:
I understand your point of view but when @tig insisted on default builders I didn't immediately understand his intention. Check my PR #712 and you will see that even having only one default driver with a simple text it fills the ListView with the generic application `AllViewTesters.cs`. For now it only works with the `Text `and `Title `properties but @tig has already mentioned that it intends to implement other features.

tig:
> Need @tig to comment on the validity of making this change
> 
> ```cs
> 	// Instantiate view
> 	var view = (View)Activator.CreateInstance (type);
> 
> 	view.Width = Dim.Fill (); // <-- Added
> 	view.Height = Dim.Fill (); // <-- Added

I'd rather not add code like this to AllViewsTester, because one of the values of this Scenario is to illustrate the default behavior of views (to expose bugs).

tig:
IMO, given the code I just pushed in #713, when ComboBox is selected, it should show "Test Text" in the `TextField` and `List Item #1` (2 & 3) in the List. It does not do that now.

BDisp:
Of course it is just a record but it follows the same principle as the others. Now reading a `Source `property as @tig intends to implement and not a method as you suggested (although it was also a good idea, don't misinterpret) is more appropriate for the purpose of this class that intends to test its properties.

listview-text

BDisp:
I am checking in the image I sent that the ListView does not keep the focus on the current selected item. @tig could fix this, please :-)

fergusonr:
> IMO, given the code I just pushed in #713, when ComboBox is selected, it should show "Test Text" in the `TextField` and `List Item #1` (2 & 3) in the List. It does not do that now.


Thanks @tig Your not seeing the list for three reasons:-

* 1 There's a fix in this PR required
* 2 The "Test Text" does not match anything in the source list items

Change the list items to this:-
```csharp
var source = new ListWrapper (new List<ustring> () { 
  ustring.Make("Test Text #1"), ustring.Make("Test Text #2"), ustring.Make("Test Text #3") });
```

* 3 The scenario environment is not fully emulating the real environment. Application.Loaded is not getting fired. Need this at the end of ```AllViewsTester.CreateClass()```

```csharp
View CreateClass (Type type) {
    ....
    ....
    ....
    Application.Loaded?.Invoke (new Application.ResizedEventArgs () { 
        Rows = Application.Driver.Rows, Cols = Application.Driver.Cols });

    return view;
}
```

UICatalog ComboBox

There is an issue in that the list does not appear until you give / remove focus on the Combo. Not clear where the problem lies at the moment.

tig:
Application loaded is not the right event. Use LayoutComplete.

-cek
Sent from mobile device. Expect brevity, typos, and dangerous driving.

BDisp:
@fergusonr can you add a down arrow that will open or close the list items of the `ComboBox`, please? In that case both the `TextField `and the `ListView `will be with the same width, that is `Width - 1`. Something like this.
```
		{
			base.Redraw (bounds);
			Move (Bounds.Right - 1, 0);
			Driver.AddRune (Driver.DownArrow);
		}

		public override bool MouseEvent (MouseEvent me)
		{
			if (me.X == Bounds.Right - 1 && me.Y == Bounds.Top && me.Flags == MouseFlags.Button1Clicked) {
				// Process the drop-down.
				return true;
			}
			return base.MouseEvent (me);
		}

```

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -108,18 +118,20 @@ private void Initialize()
...
 			LayoutComplete += (LayoutEventArgs a) => {
 
 				search.Width = Bounds.Width;
 				listview.Width = autoHide ? Bounds.Width - 1 : Bounds.Width;
+				listview.Height = CalculatetHeight ();
 			};
 
 			listview.SelectedItemChanged += (ListViewItemEventArgs e) => {
 
 				if(searchset.Count > 0)
```
tig, line 133: Will you please use brackets (`{` and `}`) around all nested statements?

BDisp:
> 
> 
> I am checking in the image I sent that the ListView does not keep the focus on the current selected item. @tig could fix this, please :-)

I already fixed this in the PR #719.

fergusonr:
> This is great progress. Thanks a ton for all your hard work on this.
> 
> See my notes on using brackets for nested statements.
> 
> Also, ApplicationLoaded is not the right event. Use LayoutComplete.

Thanks @tig The gui.cs API provides Application.Loaded and I'm using it. I need this for one time post constructor initialzation and LayoutComplete to handle the resize. Equivalent in WinForms:-

|gui.cs | WinForms | WinForms Description
 :---    |  :--- | :---
 Loaded | Load| Occurs before a form is displayed for the first time
 LayoutComplete | ReSize | Occurs when the control is resized
 
 If Application.Loaded is depricated then maybe we should do this:-
```csharp 
[Obsolete ("Use LayoutComplete")]
public static Action<ResizedEventArgs> Loaded;
```

```csharp 
#pragma warning disable CS0618 // Type or member is obsolete
Loaded?.Invoke (new ResizedEventArgs () { Rows = Driver.Rows, Cols = Driver.Cols });
#pragma warning restore CS0618 // Type or member is obsolete
```

However, I do think there is a valid reason for providing this function

fergusonr:
> @fergusonr can you add a down arrow that will open or close the list items of the `ComboBox`, please? In that case both the `TextField `and the `ListView `will be with the same width, that is `Width - 1`. Something like this.
> 
> ```
> 		public override void Redraw (Rect bounds)
> 		{
> 			base.Redraw (bounds);
> 			Move (Bounds.Right - 1, 0);
> 			Driver.AddRune (Driver.DownArrow);
> 		}
> 
> 		public override bool MouseEvent (MouseEvent me)
> 		{
> 			if (me.X == Bounds.Right - 1 && me.Y == Bounds.Top && me.Flags == MouseFlags.Button1Clicked) {
> 				// Process the drop-down.
> 				return true;
> 			}
> 			return base.MouseEvent (me);
> 		}
> ```

Make a start. Enabled with  #if COMBO_FEATURE


> Thanks. It's possible to have that text also into the source? May be this? Not tested.
> 
> ```
> 
> 		public ComboBox () : this (string.Empty) { }
> 
> 		/// <summary>
> 		/// Public constructor
> 		/// </summary>
> 		/// <param name="text"></param>
> 		public ComboBox (ustring text) : base ()
> 		{
> 			var source = new List<string> () { text.ToString () };
> 			search = new TextField (text.ToString ());
> 			listview = new ListView (source) { LayoutStyle = LayoutStyle.Computed, CanFocus = true };
> 
> 			Initialize ();
> 			Text = text;
> 			SetSource (source);
> 		}
> ```

Made a start, not quite ready yet. Enabled with  #if COMBO_FEATURE

Terminal.Gui/Views/ComboBox.cs
fergusonr: That's done on my latest check-in

tig:
> Thanks @tig The gui.cs API provides Application.Loaded and I'm using it. I need this for one time post constructor initialzation and LayoutComplete to handle the resize. 

What happens if your control is added to a form dynamically after the application is loaded? 

You'll note your handler never gets called in the `AllViewsTester`.

BDisp:
I submit a PR #721 that will ensure that the `Loaded `event are invoke to all subscribers views, but it will not be invoked in the `AllViewsTester`.

fergusonr:
> > Thanks @tig The gui.cs API provides Application.Loaded and I'm using it. I need this for one time post constructor initialzation and LayoutComplete to handle the resize.
> 
> What happens if your control is added to a form dynamically after the application is loaded?
> 
> You'll note your handler never gets called in the `AllViewsTester`.

Understood. Let me have a re-think. I'll take @BDisp's #721 and have a play around

fergusonr:
> > > Thanks @tig The gui.cs API provides Application.Loaded and I'm using it. I need this for one time post constructor initialzation and LayoutComplete to handle the resize.
> > 
> > 
> > What happens if your control is added to a form dynamically after the application is loaded?
> > You'll note your handler never gets called in the `AllViewsTester`.
> 
> Understood. Let me have a re-think. I'll take @BDisp's #721 and have a play around

One point worth nothing, WinForms has this:-
```csharp
// Summary:
// Called after the control has been added to another container.
protected virtual void InitLayout();
```

tig:
> One point worth nothing, WinForms has this:-
> 
> ```cs
> // Summary:
> // Called after the control has been added to another container.
> protected virtual void InitLayout();
> ```

It would be trivial to add such an event to `View` that was called in `Add` (current code):

```csharp
public virtual void Add (View view)
{
	if (view == null)
		return;
	if (subviews == null)
		subviews = new List<View> ();
	subviews.Add (view);
	view.container = this;
	if (view.CanFocus)
		CanFocus = true;
	SetNeedsLayout ();
	SetNeedsDisplay ();
}
```

But I do not understand the use-case. If you do not think you can make ComboBox work well without this (using `LayoutStarted` or `LayoutComplete`), please explain why. We can then design and add it.

BDisp:
@fergusonr see PR #724 will be enough?

fergusonr:
> > One point worth nothing, WinForms has this:-
> > ```cs
> > // Summary:
> > // Called after the control has been added to another container.
> > protected virtual void InitLayout();
> > ```
> 
> It would be trivial to add such an event to `View` that was called in `Add` (current code):
> 
> ```cs
> public virtual void Add (View view)
> {
> 	if (view == null)
> 		return;
> 	if (subviews == null)
> 		subviews = new List<View> ();
> 	subviews.Add (view);
> 	view.container = this;
> 	if (view.CanFocus)
> 		CanFocus = true;
> 	SetNeedsLayout ();
> 	SetNeedsDisplay ();
> }
> ```
> 
> But I do not understand the use-case. If you do not think you can make ComboBox work well without this (using `LayoutStarted` or `LayoutComplete`), please explain why. We can then design and add it.

Its just helpful to have clean separation between one-time initialization after the control has been laid out, and the resize event that gets called multiple times. The code below only needs to be called once and it cannot go in the constructor because we don't know who the parent is yet.

```csharp
Application.Loaded += (Application.ResizedEventArgs a) => {
	// Determine if this view is hosted inside a dialog
	for (View view = this.SuperView; view != null; view = view.SuperView) {
		if (view is Dialog) {
			autoHide = false;
			break;
		}
	}
```
I did try going down this route:-

```csharp
	LayoutComplete += (LayoutEventArgs a) => {

		if (!initDone) {
			// one-time initialzation
			initDone = true;
		} else {
			// handle resize
		}
	};
```
...and ran into difficulty. This would certainly work if I persevered. Just ran out of time...

fergusonr:
> @fergusonr see PR #724 will be enough?

Thanks @BDisp This does not work. Nothing has been laid out yet. The Dim/Pos, SuperView are not initialized at this point. Probably needs to be done in Application.Begin(). Same place  LayoutComplete and Loaded are called but it does not look trivial.

```csharp
	toplevel.LayoutSubviews (); // this calls LayoutComplete()
	Loaded?.Invoke (new ResizedEventArgs () { Rows = Driver.Rows, Cols = Driver.Cols });
```

fergusonr:
> > @fergusonr see PR #724 will be enough?
> 
> Thanks @BDisp This does not work. Nothing has been laid out yet. The Dim/Pos, SuperView are not initialized at this point. Probably needs to be done in Application.Begin(). Same place LayoutComplete and Loaded are called but it does not look trivial.
> 
> ```cs
> 	toplevel.LayoutSubviews (); // this calls LayoutComplete()
> 	Loaded?.Invoke (new ResizedEventArgs () { Rows = Driver.Rows, Cols = Driver.Cols });
> ```

Actually, this won't work because UICatalog has already called Application.Run() which calls Begin()

BDisp:
@fergusonr I think you really don't need this. As long as the `ListView `is `Layout.Computed` it automatically adjusts to the `TextField`. `Autohide `as false in `Dialog `is unnecessary. If you really want a `ComboBox `that exists as such, you have to work hard and interact with the down arrow. This is what defines autohide. Why the `ListView `must always be visible in the `Dialog`, there is no logic. You originally intended to have an autocomplete control and now you have a `ComboBox `with autocomplete functionality. When typing something in the `TextField`, the `ListView `is visible showing the items that match the search. If the `TextField `is empty and the arrow is clicked, it opens the `ListView `with all the items. I think there is a lack of an option for the `ListView `to continue to display all items and just keep selected the items that correspond to what is typed in the `TextField`. It is challenging but it would be interesting :-)

BDisp:
You can also create a DropDownHeight property.

tig:
@fergusonr & @BDisp: Please work together to incorporate PR #723 into this PR as you refactor to utilize those events. As you do this, please:

1) Ensure the API docs are in well written, clear, English. Not a criticism of @BDisp but English is not his primary language and he needs help there.

2) Ensure there are unit tests for the new events.

3) See if you can find one other existing client code (in one of the UI Scenarios) that should use the Adding/Removing events and update them to show (it's always better to have more than one client).

If you cannot find a CLEAR and obvious use case for "Removing" I suggest you DO NOT introduce it. I have some intuitive concerns about Removing that give me pause (related to my PR on IDisposable support) and we should never add an API unless there is an actual, real, clear use-case for it. We can always add it later.

BDisp:
If an application is closing there are none Removing event firing because the Garbage Collector manages the memory. Now if dynamically remove a view the event will be fired.

tig:
> If an application is closing there are none Removing event firing because the Garbage Collector manages the memory. Now if dynamically remove a view the event will be fired.

This is the whole purpose of `IDisposible` - the garbage collector is not deterministic. Look at my PR for IDisposible. You'll see that it calls `Dispose` for any `View` that has been added with `View:Add`.

BDisp:
> This is the whole purpose of `IDisposible` - the garbage collector is not deterministic. Look at my PR for IDisposible. You'll see that it calls `Dispose` for any `View` that has been added with `View:Add`.

If we really want to cancel an application from closing then the `IDisposable `it's important to be use to avoid that, but if a user click in the X (Close) window would the `IDisposable `run? If it don't then I think it's not needed.

tig:
> > This is the whole purpose of `IDisposible` - the garbage collector is not deterministic. Look at my PR for IDisposible. You'll see that it calls `Dispose` for any `View` that has been added with `View:Add`.
> 
> If we really want to cancel an application from closing then the `IDisposable `it's important to be use to avoid that, but if a user click in the X (Close) window would the `IDisposable `run? If it don't then I think it's not needed.

I'm not sure my IDisposible PR is correct, which is why it's still sitting there waiting for review (from you and others). But I do know that proper design of this should be that `Dispose` ALWAYS gets called even if the user completely exits the process.

BDisp:
> I'm not sure my IDisposible PR is correct, which is why it's still sitting there waiting for review (from you and others). But I do know that proper design of this should be that `Dispose` ALWAYS gets called even if the user completely exits the process.

I already commented about not updating correctly the text changes on Label.

tig:
@fergusonr: I had to tweak this code in Search_Changed to deal with the fact you are not using `ustring` consistently. My fix made its so my test worked, but I"m not 100% sure it is actually correct. Please review how this code is supposed to work and ensure this is right and part of this PR.

``cs
private void Search_Changed (ustring text)
{
	if (source == null) // Object initialization
		return;

	if (string.IsNullOrEmpty (search.Text.ToString ()))
		ResetSearchSet ();
	else
		searchset = source.ToList().Cast<ustring>().Where (x => x.StartsWith (search.Text)).ToList();

	listview.SetSource (searchset);
	listview.Height = CalculatetHeight ();

	listview.Redraw (new Rect (0, 0, width, height)); // for any view behind this
	this.SuperView?.BringSubviewToFront (this);
}
```

**Comment by tig** [comment]
@fergusonr, sorry to keep piling on, but my IDisposed PR has exposed more issues in combobox. Please fix in this PR.

- You were assuming the items in the `ListView` were `strings`. I changed this code to `ustring` thinking it was right, but it didn't fix the problem. The problem is you can't just assume the ListView holds strings. See this:

```csharp
/// <summary>
/// ListView <see cref="View"/> renders a scrollable list of data where each item can be activated to perform an action.
/// </summary>
/// <remarks>
/// <para>
///   The <see cref="ListView"/> displays lists of data and allows the user to scroll through the data.
///   Items in the can be activated firing an event (with the ENTER key or a mouse double-click). 
///   If the <see cref="AllowsMarking"/> property is true, elements of the list can be marked by the user.
/// </para>
/// <para>
///   By default <see cref="ListView"/> uses <see cref="object.ToString"/> to render the items of any
///   <see cref="IList"/> object (e.g. arrays, <see cref="List{T}"/>,
///   and other collections). Alternatively, an object that implements the <see cref="IListDataSource"/>
///   interface can be provided giving full control of what is rendered.
/// </para>
/// <para>
///   <see cref="ListView"/> can display any object that implements the <see cref="IList"/> interface.
///   <see cref="string"/> values are converted into <see cref="ustring"/> values before rendering, and other values are
///   converted into <see cref="string"/> by calling <see cref="object.ToString"/> and then converting to <see cref="ustring"/> .
/// </para>
/// <para>
///   To change the contents of the ListView, set the <see cref="Source"/> property (when 
///   providing custom rendering via <see cref="IListDataSource"/>) or call <see cref="SetSource"/>
///   an <see cref="IList"/> is being used.
/// </para>
/// <para>
///   When <see cref="AllowsMarking"/> is set to true the rendering will prefix the rendered items with
///   [x] or [ ] and bind the SPACE key to toggle the selection. To implement a different
///   marking style set <see cref="AllowsMarking"/> to false and implement custom rendering.
/// </para>
/// </remarks>
public class ListView : View {
```

Thus anywhere in your code where you are referencing `searchset` or `source.ToList()` you need to actually check the type of the list item like `ListView` does and do the right thing:

```csharp
/// <summary>
/// Renders a <see cref="ListView"/> item to the appropriate type.
/// </summary>
/// <param name="container">The ListView.</param>
/// <param name="driver">The driver used by the caller.</param>
/// <param name="marked">Informs if it's marked or not.</param>
/// <param name="item">The item.</param>
/// <param name="col">The col where to move.</param>
/// <param name="line">The line where to move.</param>
/// <param name="width">The item width.</param>
public void Render (ListView container, ConsoleDriver driver, bool marked, int item, int col, int line, int width)
{
	container.Move (col, line);
	var t = src [item];
	if (t == null) {
		RenderUstr (driver, ustring.Make(""), col, line, width);
	} else {
		if (t is ustring) {
			RenderUstr (driver, (ustring)t, col, line, width);
		} else if (t is string) {
			RenderUstr (driver, (string)t, col, line, width);
		} else
			RenderUstr (driver, t.ToString (), col, line, width);
	}
}
```

I personally think this design in `ListView` is a hack, but it is what it is, and we have to work with it.
```

**Comment by fergusonr** [comment]
> > @fergusonr see PR #724 will be enough?
> 
> Thanks @BDisp This does not work. Nothing has been laid out yet. The Dim/Pos, SuperView are not initialized at this point. Probably needs to be done in Application.Begin(). Same place LayoutComplete and Loaded are called but it does not look trivial.
> 
> ```cs
> 	toplevel.LayoutSubviews (); // this calls LayoutComplete()
> 	Loaded?.Invoke (new ResizedEventArgs () { Rows = Driver.Rows, Cols = Driver.Cols });
> ```

Hey, with this View.cs change, the SuperView is correctly set and ComboBox can detect its in a dialog
```csharp
public virtual void OnAdding (View view)
{
	view.Adding?.Invoke (view);
}

**Comment by BDisp** [comment]
> Hey, with this View.cs change, the SuperView is correctly set and ComboBox can detect its in a dialog
> 
> ```cs
> public virtual void OnAdding (View view)
> {
> 	view.Adding?.Invoke (view);
> }
> ```

@fergusonr I had already tried it that way and noticed that the parameter contains the information from the `SuperView `but does not yet contain the final values of the frame. But I think it will be the best option because the event will only occur once for the ComboBox instead of each SubView. You will have to avoid that the `Width `and the `Height `cannot have negative values or you will have an `Exception `of error. Nothing that `Math.Max () `can't solve. So I think it will be better to stay as below so as not to have to repeat the same object in the parameter.

```
public virtual void OnAdding (View view)
{
        view.Adding? .Invoke (this);
}

public virtual void OnRemoving (View view)
{
        view.Removing? .Invoke (this);
}
```

**Comment by fergusonr** [comment]
Please note: This branch now contains PR #724.  "Views now are notified when they are added or removing"  as my changes are built on top of it.

**Comment by BDisp** [comment]
@fergusonr I strongly advise not to use the `Loaded `event on a control. In a toplevel still accepted. Otherwise, the event will be tied to the toplevel even after it has been removed, as shown in the attached image. I recommend changing the `LayoutComplete `event to the following:

```
			LayoutComplete += (LayoutEventArgs a) => {
				if (!autoHide && search.Frame.Width != Bounds.Width ||
					autoHide && search.Frame.Width != Bounds.Width - 1) {
					search.Width = listview.Width = autoHide ? Bounds.Width - 1 : Bounds.Width;
					listview.Height = CalculatetHeight ();
					search.SetRelativeLayout (Bounds);
					listview.SetRelativeLayout (Bounds);
				}
			};

```

loaded-issue

**Comment by BDisp** [comment]
See here https://github.com/BDisp/gui.cs/tree/combobox-multiple-fixes

**Comment by fergusonr** [comment]
> See here https://github.com/BDisp/gui.cs/tree/combobox-multiple-fixes

Perfect, Pulled/pushed. Its now working in UICatalog thanks!

File: Terminal.Gui/Views/ComboBox.cs [hunk 2]
```diff
@@ -33,7 +28,11 @@ public class ComboBox : View {
 			get => source;
 			set {
 				source = value;
-				SetNeedsDisplay ();
+
+				if(isAdded) {
```
tig, line 32: Can you explain why `isAdded` is needed? I don't understand it. There's probably a good reason, but it's not clear to me.

Note you never set it to false.

If it's really needed, you can use `if (SuperView != null && SuperView.SubViews.IndexOf(this) != -1)` to accomplish the same thing; this will also protect against the case where the view has been added twice.

Terminal.Gui/Core/View.cs
```diff
@@ -596,6 +607,7 @@ public virtual void Remove (View view)
 			if (view == null || subviews == null)
 				return;
 
+			OnRemoving (view);
```
tig, line 610: You don't use `Removing`. I don't think we should add APIs that there is no clear use case for. We can always add them later once we have a clear use case, but if we get it wrong we've just created bad legacy. 

Side-note: The concern I raised in another comment on `isAdding` illustrates why `Removing` may be more complicated than first imagined.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -131,63 +153,51 @@ private void Initialize()
...
-					this.Height = h + 1; // adjust view to account for search box
+				if (autoHide) {
+					listview.ColorScheme = Colors.Menu;
 				} else {
-					if (height == 0)
-						height = GetDimAsInt (Height, vertical: true);
-
-					listview.Height = CalculatetHeight ();
-					this.Height = height + 1; // adjust view to account for search box
+					search.ColorScheme = Colors.Menu;
```
tig, line 164: I don't think this is critical for this PR, but I believe the way you've implemented color support means callers can't change what colors are used by setting `ColorScheme`. Something for you to consider.

Terminal.Gui/Core/View.cs
```diff
@@ -124,6 +124,16 @@ internal enum Direction {
 
 		TextFormatter viewText;
 
+		/// <summary>
+		/// Event fired when a subview is being added to this view.
+		/// </summary>
+		public Action<View> Adding;
```
tig, line 130: per the other PR on this topic:

>On names: The actual implementation is synchronous and the call to the virtual method (or the event) happens as the view is being added. In the PR, the call to the virtual function happens AFTER the call to subviews.Add. Thus the proper tense is "past tense". Thus the name of the virtual function (and event if it's required) should be past tense: "Added".

>Since this refers to another method, where "View" is not included in the name, we should not include "View" in the name here either. Thus the virtual function, being consistent with the rest of the project, should be "OnAdded".

BDisp: @tig be consistent in what you say please. This change @fergusonr merged from my PR #723 and you criticized me for using `Added `and `OnAdded `and are now saying the opposite. That way you discourage anyone who has the slightest desire to make improvements to this project.

tig: You are right: I changed my mind between the original code review and subsequent discussions with Miguel that made see I was wrong before. 

In my opinion, adding new APIs should be "highly considered". We should all be open to having strong opinions and then debating things. We should be willing to be wrong. And we should be willing to change our minds if further debate and thought shows we were wrong.

If we were just working on an "App" I'd feel less strongly about getting issues like this right. But we are creating an API and thus the bar is higher. I view it as one of the things that makes working on this project so much fun.

BDisp: > I view it as one of the things that makes working on this project so much fun.

@tig but commenting on a PR that just merged a PR of mine is not that fun. At least you could admit you were wrong in my PR instead of commenting here.

Terminal.Gui/Views/ComboBox.cs
fergusonr: Thanks @tig Its for the case where the source is set **before** the control is added. There is no need to force it to refresh as this will be triggered by  the call to Add()

```csharp
comboBox = new ListView();
comboBox.SetSource(list);
mainWindow.Add(comboBox);
```

Terminal.Gui/Views/ComboBox.cs
fergusonr: Understood. Will revisit this at some point.

Terminal.Gui/Core/View.cs
tig: > > I view it as one of the things that makes working on this project so much fun.
> 
> @tig but commenting on a PR that just merged a PR of mine is not that fun. At least you could admit you were wrong in my PR instead of commenting here.

I was hoping you and @fergusonr were working together on this PR. I had asked that you do so. 

I made the same points on the other threads that I made here about naming and not adding an API for the removed case.

BDisp:
The `TextField `cursor must not be over the down arrow. @fergusonr can you test the changes here (https://github.com/BDisp/gui.cs/tree/combobox-multiple-fixes) please? Thank you.

BDisp:
@fergusonr there is more fixes in my branch. Thanks.

Terminal.Gui/Views/ComboBox.cs
fergusonr: Implemented

Terminal.Gui/Core/View.cs
fergusonr: I'm slightly leaning towards keeping Removing() although I cannot think of a single use case. WinForms has this:-


```csharp
// Occurs when a new control is added to the System.Windows.Forms.Control.ControlCollection.
public event ControlEventHandler ControlAdded;

// Occurs when a control is removed from the System.Windows.Forms.Control.ControlCollection.
public event ControlEventHandler ControlRemoved;
```

Terminal.Gui/Views/ComboBox.cs
fergusonr: Implemented

BDisp:
@fergusonr I inserted a `ScrollView `in the `ComboBox`. I hope you like it.

tig:
Can one of you articulate a plan for finalizing this PR and ensuring all of the feedback has been incorporated? It's not clear to me how this will all come together. 

Also, I'm not sure if this is addressed here yet or not, but it should be:

In `Unicode.cs` I had to comment out Combobox because of a crash due to Unicode handling:
image

BDisp:
You can test my branch here (https://github.com/BDisp/gui.cs/tree/combobox-multiple-fixes) but @fergusonr will make your final decision as the author of the control. I just insisted that it is a `ComboBox `with `AutoComplete `features when typing but it will also have to work like a real `ComboBox`. But @fergusonr has every right to give his opinion and I am willing to collaborate with him to achieve common goals. `Unicode `already works because @fergusonr presented resolution in this PR.

fergusonr:
Yes, the Unicode issue was fixed a good while back in this PR. 

unicode
```csharp
var comboBox = new ComboBox ("git.cs") {
	X = 20,
	Y = Pos.Y (label),
	Width = Dim.Percent (50),
	Height = Dim.Percent(20)
};
```
I believe this PR is ready to go. I've hand picked two of @BDisp changes. Any wider changes can be done in a new PR post v0.90 / v1.00. AutoHide for example should be made a public property so it can be overwritten

Thankyou both for all your help and patience

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -258,115 +275,136 @@ public override bool ProcessKey(KeyEvent e)
...
-		/// <returns></returns>
-		private int CalculatetHeight ()
+		/// 
+		/// Consider making public
+		private void ShowList ()
 		{
-			return Math.Min (height, searchset.Count);
+			listview.SetSource (searchset);
+			listview.Height = CalculatetHeight ();
+			listview.Redraw (new Rect (0, 0, Bounds.Width, Bounds.Height)); // Ensure list shrinks in Dialog
```
tig, line 384: Generally, I view it is a danger sign when code calls `Redraw` directly vs. `SetNeedsDisplay`. Are you sure you need to synchronously force a paint here? I don't fully understand the way this control works, so it may be needed. Thanks.

Terminal.Gui/Core/View.cs
tig: Please remove `Removing` from this PR. I have serious concerns about how it will interact with `IDisposable` and we don't have a clear use case where it's needed. I wish I could explain those concerns in detail, but I'm not confident I understand well enough yet. But I've learned to trust my instincts on things like this. There's no danger in removing the API for now. But there is danger in putting it in and having to change it later. Thanks.

Also, I am arguing for renaming `Adding` to `Added`. I made a comment that argues my point I think pretty clearly. I'll add to that argument that by using `Adding` (current tense`) there's an implication that the operation can be cancelled. However, if we name it `Added` (past tense) it is very clear that the deal is done and there's no going back. 

If you disagree, please explain. If you don't disagree, please rename to `Added`.

Terminal.Gui/Views/ComboBox.cs
BDisp: He need it in the `Dialog`, but he could use the `Clear `method instead.

BDisp:
I think there is no harm in maintaining `OnRemoving`. If it is to be removed then `OnAdded `must also be removed.

tig:
> I think there is no harm in maintaining `OnRemoving`. If it is to be removed then `OnAdded `must also be removed.

I think I've made my argument pretty clear. Beyond my concerns that the semantics around `Removing` will cause conflict with `IDisposable`, in framework/API design a key tenet to be followed is "Do not expose APIs that have no use cases." It just creates complexity and adds more code. History has shown that un-needed APIs just create bad legacy that has to be maintained in the future. We should strive to REMOVE code from the framework, not just add more and more.

If you feel strongly that `Removing` should be part of the API I would appreciate it if you explain why.

BDisp:
> If you feel strongly that `Removing` should be part of the API I would appreciate it if you explain why.

@tig if you prove it will create conflicts with the `IDisposable `implementation, I wouldn't think twice about supporting you. In this specific case of `Removing `I will tell you use cases in which they are very beneficial in implementing them. For example, `ScrollView `has two properties, namely `ShowVerticalScrollIndicator `and `ShowHorizontalScrollIndicator`, where they dynamically add or remove views from his control, which are the cases of the `vertical `and `horizontal ``ScrollBarView`. Now imagine that you create a control that includes the `ScrollView `as a subview and you want, for example, to resize any subview or other control as `AutoHideScrollBars `adds or removes the `ScrollBarViews`. If you had an event that notifies the subscribers of this occurrence, it becomes easier to react according to the needs.

BDisp:
@fergusonr can you take a look for this small adjustments I have made to improve the `ComboBox`, please? Is more to hide the list if the `ComboBox `loses focus. Thanks.

https://github.com/BDisp/gui.cs/tree/pr%23710-combobox-multiple-fixes

Terminal.Gui/Views/ComboBox.cs
fergusonr: I changed it to Clear() and it achieves the same effect. Thanks @BDisp 

*pushed*

BDisp:
@fergusonr, please merge my PR #723 again because I changed the `Removing `event to `Removed `to take action when the view no longer exists in the collection.

After selecting an item, the ComboBox closes but no longer opens either by clicking on the down arrow or with the cursor down.

BDisp:
When `autohide `is `true `the `ComboBox `should close the list when it loses focus.

BDisp:
With the mouse, the `ComboBox `should only get the focus with a click and not just passing the mouse over it.

PR #712, opened 2020-06-20
Fixes #711 and I think perhaps won't break @tig's PR #702, otherwise I'll close this.

### LINKED ISSUE #711: ListView don't have a default source to use with the AllViewsTester.cs.
Adding a `Text `property may be resolve it.

BDisp:


BDisp:
What is promised is owed. I said I would close it if it wasn't necessary. I understand @tig with Source this no longer makes sense.

PR #716, opened 2020-06-21
Fix InputRecord marshalling when built with .NET Native toolchain

InvictusMB:
This is a subset of #707 that fixes the input issue with UWP apps (#700). Properly detecting UWP environment and switching the bindings to pass Windows Store checks seems to be a more sophisticated thing and better be tracked separately.


BDisp:
This is fantastic. @InvictusMB can you find a way to be detected when the window size is being reduced vertically, please?

PR #722, opened 2020-06-23
Fixes #721 The loaded event must be called for all views that subscribe to it.

### LINKED ISSUE #721: Application.Loaded is useless if it is to be called by more than one object.
The problem with `Application.Loaded` is that it belongs to the `Application `class only. It must belong to the `View `class and call all views of subscribers who implement it. The `Application `class should only invoke the `Loaded `event.

BDisp:


Terminal.Gui/Core/View.cs
```diff
@@ -142,6 +142,12 @@ internal enum Direction {
 		View focused = null;
 		Direction focusDirection;
 
+		/// <summary>
+		/// This event is fired once when the application is first loaded. The dimensions of the
+		/// terminal are provided.
+		/// </summary>
+		public Action<Application.ResizedEventArgs> Loaded;
```
tig, line 149: For simple apps this concept is OK.

However, for anything dynamic, having resized be tied to `Application` loading makes no sense and makes this API coutner-productive. As I noted in another thread, this will never get called if the View is created AFTER the app has been running. This is one of the reasons I wrote `AllViewsTester`, to illustrate bugs in Views that depended on the idea they were all created before `Application.Run` was called.

I do not think we should extend `View` to have a `Loaded` event.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -122,7 +122,7 @@ private void Initialize()
 					SetValue ((string)searchset [listview.SelectedItem]);
 			};
 
-			Application.Loaded += (Application.ResizedEventArgs a) => {
+			Loaded += (Application.ResizedEventArgs a) => {
```
tig, line 125: This will not get called in the 'AllViewsTest` scenario. We should not encourage people to do this.

Terminal.Gui/Core/View.cs
BDisp: You are right @tig. I anticipated and did not reflect on it well. I will present another PR that will only invoke an event once when the view is added to the toplevel. I will close this PR next.

Terminal.Gui/Views/ComboBox.cs
BDisp: I know he will not be called, I had said this before in another comment (https://github.com/migueldeicaza/gui.cs/pull/710#issuecomment-648150447). I will present another alternative that will call a view as soon as it is added to the toplevel.

PR #724, opened 2020-06-23
Fixes #723 Views now are notified when they are added or removing.

### LINKED ISSUE #723: The views aren't notified when they are added or removing.
If a view have several views the events will get called for each one and in the event parameter will have the view being added or being removing. If there is no need to execute more than once it only need to add a flag to the event that was already executed.

BDisp:
If a view have several views the events will get called for each one and in the event parameter will have the view being added or being removing. If there is no need to execute more than once it only need to add a flag to the event that was already executed.

### DIFF
--- a/Terminal.Gui/Core/Responder.cs
+++ b/Terminal.Gui/Core/Responder.cs
@@ -165,18 +165,6 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
-		/// Method invoked when a view is added.
-		/// </summary>
-		/// <param name="view">The view added.</param>
-		public virtual void OnAddedView (View view) { }
-
-		/// <summary>
-		/// Method invoked when a view being removing.
-		/// </summary>
-		/// <param name="view">The view being removing.</param>
-		public virtual void OnRemovingView (View view) { }
-
-		/// <summary>
 		/// Method invoked when a view gets focus.
 		/// </summary>
 		/// <returns><c>true</c>, if the event was handled, <c>false</c> otherwise.</returns>
--- a/Terminal.Gui/Core/Responder.cs
+++ b/Terminal.Gui/Core/Responder.cs
@@ -165,6 +165,18 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
+		/// Method invoked when a view is added.
+		/// </summary>
+		/// <param name="view">The view added.</param>
+		public virtual void OnAddedView (View view) { }
+
+		/// <summary>
+		/// Method invoked when a view being removing.
+		/// </summary>
+		/// <param name="view">The view being removing.</param>
+		public virtual void OnRemovingView (View view) { }
+
+		/// <summary>
 		/// Method invoked when a view gets focus.
 		/// </summary>
 		/// <returns><c>true</c>, if the event was handled, <c>false</c> otherwise.</returns>
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -125,14 +125,14 @@ namespace Terminal.Gui {
 		TextFormatter viewText;
 
 		/// <summary>
-		/// Event fired when the view is added.
+		/// Event fired when a subview is being added to this view.
 		/// </summary>
-		public Action<View> AddedView;
+		public Action<View> Adding;
 
 		/// <summary>
-		/// Event fired when the view is being removing.
+		/// Event fired when a subview is being removed from this view.
 		/// </summary>
-		public Action<View> RemovingView;
+		public Action<View> Removing;
 
 		/// <summary>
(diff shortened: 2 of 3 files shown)


migueldeicaza:
Can you share what is the rationale for this change?   

Generally, I would like to keep with the original UIKit idioms if possible, and if so, I would add a 'ViewWillBeAdded" virtual method, and a "ViewWillBeRemoved".   I see that this is being done via events, and I am wondering what is the scenario that would require this.

BDisp:
It is to notify the view that already added to the SuperView. About the names I prefer let @tig comment.

tig:
> Can you share what is the rationale for this change?
> 
> Generally, I would like to keep with the original UIKit idioms if possible, and if so, I would add a 'ViewWillBeAdded" virtual method, and a "ViewWillBeRemoved". I see that this is being done via events, and I am wondering what is the scenario that would require this.

Virtual methods are useful for derived classes (like `ComboBox`). Events are useful for clients. 

In this case, you are correct, the use-case is for a derived class, and thus virtual method should be sufficient. 

If we also apply the tenet "Only Add APIs where there is a real, clear, use-case" then we should ONLY add the virtual method for when a view is added in this PR, because that is all @fergusonr requires to make `ComboBox` work.

tig:
On names: The actual implementation is synchronous and the call to the virtual method (or the event) happens as the view is being added. In the PR, the call to the virtual function happens AFTER the call to `subviews.Add`. Thus the proper tense is "past tense". Thus the name of the virtual function (and event if it's required) should be past tense: "Added". 

Since this refers to another method, where "View" is not included in the name, we should not include "View" in the name here either. Thus the virtual function, being consistent with the rest of the project, should be "OnAdded".

tig:
Really great work!

tig:
If the `Dispose` method on a view is called, would subviews expect to have `Removed` be called?

In my `IDisposable` PR, `View` has this `Dispose` method:

```csharp
/// <inheritdoc/>
protected override void Dispose (bool disposing)
{
	foreach (var subview in InternalSubviews) {
		subview.Dispose ();
	}
	base.Dispose (disposing);
}
```

I'm trying to merge this PR into that one, and I've found the place where my instincts were screaming `Removed` is a bad idea.

As coded above, subviews will never had `Removed` get called when their parent gets cleaned up, because `Remove` is never explicitly called. However they ARE definitely removed (implicitly). 

I think there are four ways to fix this (maybe more):

1. Remove `Removed` from the API until someone really needs it and can say what they want to happen in the case of `Dispose`.
2. Change the code above to pass through `InternalSubviews` calling `Remove (subview)`; this has to be done with a copy of the `List` because you can't `Remove` an item in a `List` while enumerating it. This adds complexity.
3. Change the definition of `Remove` such that it destroys (calls `Dispose`) on the subview. This gets ugly because its a) a breaking change, and b) means clients have to be aware of this.
4. Don't fix it - Clients that depend on the `Removed` event will not be certain the event will be fired.

Please guide me. Thanks.

BDisp:
Ok @tig I will study this. I really already have ideas about using the `Removed `event for some things and one of the reasons that led me to change from `Removing `to `Removed `was to think that the subscriber who wants to use it if called after the view has been removed would give a null argument exception but that is not what happened, because what is being removed is the `List<View>` item and not the view itself. Only the view.container is set to null. So, if there is an error in `IDispose `it has nothing to do with this event but because of the iteration with the `List<View>` which then detects that the list has been changed. In this case, use `for `loop.

BDisp:
@tig you can do more or less this:
```cs
		protected override void Dispose (bool disposing)
		{
			List<View> vToDispose = new List<View> ();
			foreach (var subview in InternalSubviews) {
				subview.Dispose ();
				vToDispose.Add (subview);
			}
			foreach (var v in vToDispose) {
				Remove (v);
			}
			base.Dispose (disposing);
		}
```
Here is the unit test:

```cs
		[Fact]
		public void Added_Removed_IDispose ()
		{
			var v = new View (new Rect (0, 0, 10, 24));
			var t = new View ();

			v.Added += (View e) => {
				Assert.True (v.SuperView == e);
			};

			v.Removed += (View e) => {
				Assert.True (v.SuperView == null);
			};

			t.Add (v);
			Assert.True (t.Subviews.Count == 1);

			t.Dispose ();
			Assert.True (t.Subviews.Count == 0);
		}

```

BDisp:
If decreasing, it would be like this:

````cs
		protected override void Dispose (bool disposing)
		{
			IList<View> vToDispose = new List<View> ();
			foreach (var subview in InternalSubviews.Reverse ()) {
				subview.Dispose ();
				vToDispose.Add (subview);
			}
			foreach (var v in vToDispose.Reverse ()) {
				Remove (v);
			}
			base.Dispose (disposing);
		}

````

PR #735, opened 2020-07-02
ListView - handle null values

daltskin:
Came across this when passing in null collection.

### DIFF
--- a/Terminal.Gui/Views/ListView.cs
+++ b/Terminal.Gui/Views/ListView.cs
@@ -561,11 +561,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Gets the number of items in the <see cref="IList"/>.
 		/// </summary>
-		public int Count {
-			get {
-				return src?.Count != null ? src.Count : 0;
-			}
-		}
+		public int Count => src != null ? src.Count : 0;
 
 		void RenderUstr (ConsoleDriver driver, ustring ustr, int col, int line, int width)
 		{
--- a/Terminal.Gui/Views/ListView.cs
+++ b/Terminal.Gui/Views/ListView.cs
@@ -561,7 +561,11 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Gets the number of items in the <see cref="IList"/>.
 		/// </summary>
-		public int Count => Convert.ToInt32 (src?.Count);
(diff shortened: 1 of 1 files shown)


Terminal.Gui/Views/ListView.cs
```diff
@@ -551,15 +551,17 @@ public class ListWrapper : IListDataSource {
...
+				marks = new BitArray (count);
+				this.src = source;
+			}
 		}
 
 		/// <summary>
 		/// Gets the number of items in the <see cref="IList"/>.
 		/// </summary>
-		public int Count => src.Count;
+		public int Count => Convert.ToInt32 (src?.Count);
```
tig, line 564: Why is the Convert.ToInt32 needed? Thanks.

daltskin: This was to handle src?.Count being null and therefore return 0

tig: Neat trick. The likelyhood is pretty low, but on a 64 bit system, this will be a bug with an insanely large `Count`, right?

daltskin: It's needed because the local IList src could be null.  Without this check it will crap out trying to read the Count property if a null collection was passed in the ctor.

tig: I understand the need for check in the case `src` is `null`. But I'm worried your shortcut of using `Convert` introduces a int64 to int32 dataloss bug.

tig: On 64 bit systems `.Count` is a 64 bit integer.

daltskin: Understand your now :)

Have just made a tweak to keep to the int type.  

You could even do it like this, but IMHO it's not nice to read..

```csharp
public int Count => src?.Count != null ? src.Count : 0;

Comment (tig) [reply Terminal.Gui/Views/ListView.cs]: Uh, don't you mean:

```
public int Count => src != null ? src.Count : 0;
```

daltskin: both actually work - but that's more concise.

Is the backing field int count actually needed?  Couldn't this be replaced by using the property?

PR #738, opened 2020-07-02
Fixes #737.Added an auto hide scroll bars to the ScrollView.

### LINKED ISSUE #737: ScrollView has no Auto Hide Scroll Bars.
- The vertical scroll bar won't show up if it's a subview unless there is some movement.

- `OnDrawContent `can raise an exception if the width or the height are negative.

BDisp:


BDisp:
Fixes #737

PR #741, opened 2020-07-02
Fixes #740. Multi thread toplevels.

### LINKED ISSUE #740: Multithreading not allowed with more than one Toplevel.
- If there are timers or threads running on background on another `TopLevel `than the current but visible, they aren't updated on the screen.

- The `Editor `method in the `Demo.cs` is creating another `Application.Top` unnecessarily wasting resources.

multi-thread-toplevels

BDisp:
Fixes #740

### DIFF
--- a/Example/demo.cs
+++ b/Example/demo.cs
@@ -237,12 +237,13 @@ static class Demo {
 	//
 	static void Editor ()
 	{
-		var tframe = Application.Top.Frame;
-		Application.Top.RemoveAll ();
+		Application.Init ();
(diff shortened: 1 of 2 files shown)


tig:
It would be really great if you built a set of unit tests for this. 

Nice work!

BDisp:
> 
> 
> It would be really great if you built a set of unit tests for this.
> 
> Nice work!

I submitted a little test for this Multi thread toplevels fix at PR #849.

PR #747, opened 2020-07-05
Reset the ListView when the source is defined.

BDisp:


Terminal.Gui/Core/Responder.cs
```diff
@@ -28,7 +28,7 @@ public class Responder {
 		/// Gets or sets a value indicating whether this <see cref="Responder"/> has focus.
 		/// </summary>
 		/// <value><c>true</c> if has focus; otherwise, <c>false</c>.</value>
-		public virtual bool HasFocus { get; internal set; }
+		public virtual bool HasFocus { get; }
```
tig, line 31: Is this PR dependent on the PR that adds the changes for `OnEnter/OnLeave`? If it is not, you should not include those changes in PRs that have not been accepted yet. 

The general idea is that PRs should stand on their own if possible. I can't accept this PR, which seems simple and independent, because it depends on another PR which I have questions/concerns about.

BDisp: Yes you are right. It will be better to wait until you review the other PR and then I merge with this one. Thanks.

BDisp: I really needed to use a feature from the other PR so I really had to merge it. Excuse. I think it's ok now. Thanks.

PR #749, opened 2020-07-07
Setting CanFocus also sets TabIndex and TabStop accordingly.

BDisp:


tig:
Thank you. And Thank you for the Unit Tests!!

PR #751, opened 2020-07-07
Some fixes and improvements to the ComboBox.

BDisp:


fergusonr:
Thanks @BDisp My interpretation of your comment "SelectedItem is required to appear selected..." is that for the below it should return SelectedItem = 791 = d2dl.dll (in the case of my C: drive) but it returns 0

Its so subtle I do not think its a required fix for v0.90 or v1.00

selectedItem

My thought are this:-
```csharp
private void CalculateSelectedItem ()
{
	var found = GetSelectedItemFromSource (search.Text);

	if (found == -1 && searchset.Count > 0) { // No match so return default. Index of first in list
		SelectedItem = GetSelectedItemFromSource (searchset [0].ToString ());
	} else {
		SelectedItem = found;
	}
}
```

BDisp:
I only worried about showing an existing item selected in the list and not when we type part and get out of control and get back in. You can enter the previous valid value when leaving the control. But there was also a problem in which if there was an item previously selected when entering the control, it would not open the list either by clicking on the down arrow or with the cursor, so I had to solve it because it was not functional, sorry. If you think you can improve even more than it already is, go ahead before the new version is released. The `CalculateSelectedItem `method seems to be fine.

BDisp:
`SelectedItem = GetSelectedItemFromSource (searchset [0].ToString ());`

@fergusonr only do this only if `SelectedItem == -1` otherwise you have to assign the `SelectedItem `to the `ListView `so that it doesn't have to select the first item in the list when there was already an item selected previously.

PR #752, opened 2020-07-07
Fixes #386. Window can now be dragged inside another Window.

### LINKED ISSUE #386: Sometimes controls don't activate on mouse clicks
This is really tricky to reproduce and I have not yet found a reliable way of reproducing it with the debugger.

1) Run `demo.c` in cmd.exe or ConEmu on Windows (any terminal that supports mouse).
2) Click on any control (e.g. a top-level `MenuBarItem` or one of the radio buttons.
3) Click on another control.
4) Then another
..
5) Eventually you'll notice one of your clicks does nothing; the target control does not get focus as expected. Clicking again will work.
6) You'll think you're crazy and you just either didn't hit the mouse button hard enough or missed the hit-target. But you didn't. 
7) You'll try again and find you can't precisely reproduce it the same way.

P.S. I thought I had introduced this bug in my upcoming `MenuBar` PR and was actually quite happy to find it also exists in the current `master` and that it exists across all controls.

BDisp:


tig:
Does this fix the mouse click issue? If it does not, then change the title to not reference #386. Thanks.

BDisp:
> Does this fix the mouse click issu

It depends. As I am not having these problems with the mouse I cannot comment. If he persists, tell me to change the title, please. Thanks.

PR #762, opened 2020-07-08
Fixes #760. Never call Curses.endwin if you don't want to exit the application, that is, never close the Driver.

### LINKED ISSUE #760: Colors in Linux changing between windows.
This is happening again and for sure it is the application reinitializing the driver. With `WindowsDriver`, this behavior does not happen because the console settings are only assumed when the application is finished. On `Unix / Linux`, the driver termination runs` Curses.endwin` which immediately restores the console's defaults and the application cannot maintain its own settings. I've had to solve it at least twice, as I recall. Does it have to do with `IDispose`?

unix-close-driver

BDisp:


BDisp:
I already tried but I didn't find any solution to solve this except to avoid closing the driver. Perhaps a flag that notifies the `Driver `that is going to reset and in `Curses `avoid calling `endwin`.

tig:
Ok. I will look at it. If I can't find a fix, i'll accept this PR. For now I'll close it.

PR #764, opened 2020-07-08
ListView support both LayoutStyle to calculate the height.

### LINKED ISSUE #772: ListView does not have the correct height when set the SelectedItem.

BDisp:
Instead of the `top `being set when the `SelectedItem `property is changed it is calculated before being drawn.

Fixes #772

### DIFF
--- a/Terminal.Gui/Views/ListView.cs
+++ b/Terminal.Gui/Views/ListView.cs
@@ -199,6 +199,8 @@ namespace Terminal.Gui {
 			}
 		}
 
+		bool topPending;
+
 		/// <summary>
 		/// Gets or sets the index of the currently selected item.
 		/// </summary>
@@ -214,8 +216,8 @@ namespace Terminal.Gui {
 				OnSelectedChanged ();
 				if (selected < top)
 					top = selected;
-				else if (selected >= top + (LayoutStyle == LayoutStyle.Absolute ? Frame.Height : Height.Anchor (0)))
-					top = selected;
+				else
+					topPending = true;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


BDisp:
When the `SelectedItem `is assigned he will check if he has to move the items by comparing the height. The problem is that if the `LayoutStyle `is `Computed`, the ListView does not have this information and in this case, only when it is going to draw the control does it already have that information. So I created the `topPendent `flag to force it to calculate the width.

tig:
Do you know when ListView started moving the selected item to the top when SelectedItem changed?

I'm not a fan of this behavior and am not sure if it's related to this PR or not. 

`SelectedItem.set` should only change the scroll position of the list box to display the selected item if it was not previously visible. Anything otherwise is disruptive to the UI. See my comment on the other PR for an example.

BDisp:
> Do you know when ListView started moving the selected item to the top when SelectedItem changed?

Do not know. I've always seen him like this but I really shouldn't. Thanks for the clarification.

> Anything otherwise is disruptive to the UI. See my comment on the other PR for an example.

Where?

BDisp:
Of course, when the `ListView `already contains a `SelectedItem `whose index is higher than the height, the selected item must appear at the top. Note that this only happens when you assign a value to `SelectedItem.set` or by scrolling with `PageUp`, `PageDown`, `Home `and `End`.

tig:
> > Do you know when ListView started moving the selected item to the top when SelectedItem changed?
> 
> Do not know. I've always seen him like this but I really shouldn't. Thanks for the clarification.
> 
> > Anything otherwise is disruptive to the UI. See my comment on the other PR for an example.
> 
> Where? I think is this [#765 (comment)](https://github.com/migueldeicaza/gui.cs/pull/765#issuecomment-655652695)

Correct.

Do you know how to fix that issue? Can it be fixed as part of this PR?

BDisp:
I'll merge PR #765 in this one and you'll see.

BDisp:
Done!

BDisp:
I sent the separate PRs for you not to complain. I should have sent it along ;-)

BDisp:
> `SelectedItem.set` should only change the scroll position of the list box to display the selected item if it was not previously visible.

I think this is already fixed.

PR #765, opened 2020-07-08
UICatalog preserves the SelectedItem of both ListViews.

BDisp:


tig:
Nice fix! Thanks.

tig:
This changes the behavior in a way that I find annoying:

When returning from a Scenario, the Scenarios list is scrolled such that the previously selected item is at the top.

BDisp:
Because the PR #764 is missing.

PR #767, opened 2020-07-08
Fixes #766. Added a Load event to the View.

### LINKED ISSUE #766: A Load event for the View is missing.
To allows configurations and assignments to be performed only once before the view is drawn for the first time.

BDisp:


Terminal.Gui/Core/View.cs
```diff
@@ -1208,6 +1208,10 @@ public virtual void Redraw (Rect bounds)
 								view.LayoutSubviews ();
 							Application.CurrentView = view;
 
+							if (!view.loaded) {
```
tig, line 1211: This seems like a very weird place to do this. Shouldn't the call to `OnLoaded` come from `Application`? E.g. within `RunLoop`?

BDisp:
Negative @tig because I want to capture it when it is still running in the `public static RunState Begin (Toplevel toplevel) `method at:

https://github.com/migueldeicaza/gui.cs/blob/4cd307cc6c3f2e6971b9625940373c15decdba7b/Terminal.Gui/Core/Application.cs#L471

I also want to ensure that the event takes place well before the first `Redraw `but after `LayoutSubviews`

https://github.com/migueldeicaza/gui.cs/blob/4cd307cc6c3f2e6971b9625940373c15decdba7b/Terminal.Gui/Core/View.cs#L1207

  and before `vew.Redraw`

https://github.com/migueldeicaza/gui.cs/blob/4cd307cc6c3f2e6971b9625940373c15decdba7b/Terminal.Gui/Core/View.cs#L1213

UnitTests/ViewTests.cs
```diff
@@ -543,5 +543,101 @@ public void CanFocus_Set_Changes_TabIndex_And_TabStop ()
 			Assert.Equal (2, v3.TabIndex);
 			Assert.True (v3.TabStop);
 		}
+
+		[Fact]
+		public void Load_And_Initialize_Event_Comparing_With_Added ()
```
tig, line 548: This test does not test what happens if a `View` is created AFTER `Run` has been called. Right now, `Initialize` will never fire for such cases, which I think is a problem.

BDisp: I don't think so because the `Initialized `event is only useful at the toplevel load and not if you add views dynamically after the `Run`. In this case you already have all the valid configuration off all the current views and you can deal well with them.

BDisp: Remember that the `Initialized `and `Load `events will always run for each toplevel that will be opening.

BDisp: Pease see this:

https://github.com/migueldeicaza/gui.cs/issues/766#issuecomment-656252989

UnitTests/ViewTests.cs
```diff
@@ -543,5 +543,173 @@ public void CanFocus_Set_Changes_TabIndex_And_TabStop ()
...
+				Application.Refresh ();
+			};
+
+			r.Add (v1, v2);
+			t.Add (r);
+
+			Application.Iteration = () => {
+				var sv1 = new View ("sv1") { Width = Dim.Fill (), Height = Dim.Fill () };
+
+				sv1.Added += (e) => {
```
tig, line 683: I may be missing something, but I don't see where you test that `sv1`'s `Initialize` method gets fired.

BDisp: @tig since you do not want the `Initialized `event in the `View `it is no longer possible to trigger for each view. Also triggering the Initialized in `Toplevel `for each view will cause the same event to be fired several times unless `Toplevel `triggers it for each view but then the event must belong to the `View `class. If you want `Toplevel `to trigger the `Initialized `event for each view, I have to choose a method within the view where I can invoke it. Please give suggestions.

tig: > @tig since you do not want the `Initialized `event in the `View `it is no longer possible to trigger for each view. Also triggering the Initialized in `Toplevel `for each view will cause the same event to be fired several times unless `Toplevel `triggers it for each view but then the event must belong to the `View `class. If you want `Toplevel `to trigger the `Initialized `event for each view, I have to choose a method within the view where I can invoke it. Please give suggestions.

I previously suggested we invoke `Initialized` from `RunLoop`. I even provided a code sample that illustrated how it would work. I still do not understand why you rejected that idea.

BDisp: And about the Load event? Is to remove?

BDisp:
https://github.com/migueldeicaza/gui.cs/issues/766#issuecomment-655735329

If you refer to this it is very confusing to me. I don't understand how you can trigger the `Initialized `event inside `RunLoop `for each view by moving to `Responder`. I think I'll give up on the idea anyway.

UnitTests/ViewTests.cs
BDisp: https://github.com/migueldeicaza/gui.cs/issues/766#issuecomment-655735329

If you refer to this it is very confusing to me. I don't understand how you can trigger the `Initialized `event inside `RunLoop `for each view by moving to `Responder`. I think I'll give up on the idea anyway.

BDisp:
I added support for `ISupportInitializeNotification `interface that already exists in the "`public static RunState Begin (Toplevel toplevel)`" method.

PR #774, opened 2020-07-11
Fixes #773. Added a default drop down height to the ComboBox.

### LINKED ISSUE #773: ComboBox not working properly in the Unicode scenario.

BDisp:


fergusonr:
@BDisp 'Height' sets the drop down height. If this is really necessary, maybe the propery name could be clearer?

BDisp:
The intention is to set the ComboBox height. You can merge this and made the changes you want.

BDisp:
> 
> 
> @BDisp 'Height' sets the drop down height. If this is really necessary, maybe the propery name could be clearer?

@fergusonr  In the `Unicode `scenario the `ComboBox `were not shown because there wasn't defined his height. If it is missing I prepared it to assign a default height. The name was what came in mind at the time.

fergusonr:
> > @BDisp 'Height' sets the drop down height. If this is really necessary, maybe the propery name could be clearer?
> 
> @fergusonr In the `Unicode `scenario the `ComboBox `were not shown because there wasn't defined his height. If it is missing I prepared it to assign a default height. The name was what came in mind at the time.

The same is true for ListVIew. If height is not defined, you get no list. I would have just fixed the Unicode scenario by adding height to the ComboBox

BDisp:
> The same is true for ListVIew. If height is not defined, you get no list. I would have just fixed the Unicode scenario by adding height to the ComboBox

@fergusonr I think it is not as imperative to define a default height in `ListView `as it is in `ComboBox`. I know it was enough to define the height of the `ComboBox `in `Unicode ` scenario to be able to display its content, but I think @tig excluded `Height `on purpose in the construction of the `Unicode `scenario to be able to demonstrate not only the presentation of the unicode characters but also the default presentation of the various controls. But if you think the ListView should have a default height, just show that desire.

tig:
> > The same is true for ListVIew. If height is not defined, you get no list. I would have just fixed the Unicode scenario by adding height to the ComboBox
> 
> @fergusonr I think it is not as imperative to define a default height in `ListView `as it is in `ComboBox`. I know it was enough to define the height of the `ComboBox `in `Unicode ` scenario to be able to display its content, but I think @tig excluded `Height `on purpose in the construction of the `Unicode `scenario to be able to demonstrate not only the presentation of the unicode characters but also the default presentation of the various controls. But if you think the ListView should have a default height, just show that desire.

I believe `View.Height` should specify the space the `View` will take up relative to other `Views` in the same `Superview`. In Windows there are two forms of combo boxes: 

1. Has a height of 1, and the drop down covers other controls when it opens; it has a higher z-order than other controls.
2. Has a height of 1 + the size of the drop down, but the drop down is always visible (and has a z-order equal to the edit box.

I believe the current `Combobox` code only supports #2. Right?

If so, I'll assert `Combobox.Height` should equal 1 + the height of the `ListView` and the default value should be 1. In `All Views Tester` it will show only the edit box and `Unicode` should be changed to show it with `1 + some reasonable value` for height.

I have been thinking about how to enhance `All Views` tester to be smart about each View, without forcing each View to put put into production code, code to support it. I was thinking of creating an idiom where the unit tests for each View would include metadata specifying a good 'default test case' and `All Views Tester` would use reflection on the UNIT TESTs to determine the defaults (including fake data). 

Thoughts?

BDisp:
I was more inclined towards 1) but if you prefer 2) I agree with @tig about the `ComboBox `height having a default value of 2. So I'll update this PR by removing the `DropDownHeight `property and initializing with 2 if it's equal to zero. Do you  both agree?

fergusonr:
> I was more inclined towards 1) but if you prefer 2) I agree with @tig about the `ComboBox `height having a default value of 2. So I'll update this PR by removing the `DropDownHeight `property and initializing with 2 if it's less than 2. Do you both agree?

Agree, thanks

PR #777, opened 2020-07-12
Fixes #775. Menu now is working properly on another toplevel and supports unicode.

### LINKED ISSUE #775: Menu auto-navigation not working on another top.
This behavior did not happen previously. I still don't quite understand why this is happening now.

menu-on-another-top

BDisp:
menu-on-another-top-unicode

### DIFF
--- a/Terminal.Gui/Core/PosDim.cs
+++ b/Terminal.Gui/Core/PosDim.cs
@@ -165,7 +165,7 @@ namespace Terminal.Gui {
 			return pCenter;
 		}
 
-		class PosAbsolute : Pos {
+		internal class PosAbsolute : Pos {
 			int n;
 			public PosAbsolute (int n) { this.n = n; }
 
--- a/Terminal.Gui/Core/Toplevel.cs
+++ b/Terminal.Gui/Core/Toplevel.cs
@@ -276,7 +276,7 @@ namespace Terminal.Gui {
 				s = ((Toplevel)SuperView).StatusBar != null;
 			}
 			if (SuperView == null || SuperView is Toplevel) {
-				l = s ? Frame.Height - 1 : Frame.Height;
+				l = s ? Driver.Rows - 1 : Driver.Rows;
 			} else {
 				l = s ? SuperView.Frame.Height - 1 : SuperView.Frame.Height;
 			}
--- a/Terminal.Gui/Core/Toplevel.cs
+++ b/Terminal.Gui/Core/Toplevel.cs
@@ -286,6 +286,7 @@ namespace Terminal.Gui {
 
 		internal void PositionToplevels ()
 		{
+			PositionToplevel (this);
 			foreach (var top in Subviews) {
 				if (top is Toplevel) {
 					PositionToplevel ((Toplevel)top);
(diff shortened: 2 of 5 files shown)


BDisp:
`StatusBar` also work on another toplevel with automatically reposition and with unicode.

statusbar-on-another-top-unicode

Terminal.Gui/Core/PosDim.cs
```diff
@@ -165,7 +165,7 @@ public static Pos Center ()
 			return pCenter;
 		}
 
-		class PosAbsolute : Pos {
+		internal class PosAbsolute : Pos {
```
tig, line 168: Changing `PosAbsolute` to be a 'psuedo-public' by exosing it as `internal` like this gives me angst. 

It is a slippery slope towards taking @migueldeicaza's brilliant (but confusing) design of 'PosDim` to the next level of complexity (and confusion). Right now, the public API for 'PosDim' is very constrained and none of the internals are used anywhere outside of `View` (except the unit tests). By doing this, I think you start to break this.

I have not had the time to fully study the issue represented by this bug/PR, so I can't offer any guidance on if there's a different way to solve this problem without exposing what is really an internal implementation detail (`PosAbsolute`) to other parts of the framework. 

If you say "nope, this is the only way to fix it" then I'll approve it and we can move on. But I'd like to ask you to consider what I'm asking. Thanks!

Terminal.Gui/Core/Toplevel.cs
```diff
@@ -296,17 +297,21 @@ internal void PositionToplevels ()
...
 		{
 			EnsureVisibleBounds (top, top.Frame.X, top.Frame.Y, out int nx, out int ny);
-			if ((nx != top.Frame.X || ny != top.Frame.Y) && top.LayoutStyle != LayoutStyle.Computed) {
-				top.X = nx;
-				top.Y = ny;
+			if ((nx != top.Frame.X || ny != top.Frame.Y) && top.LayoutStyle == LayoutStyle.Computed) {
+				if (top.X is Pos.PosAbsolute && top.Bounds.X != nx) {
+					top.X = nx;
+				}
+				if (top.Y is Pos.PosAbsolute && top.Bounds.Y != ny) {
```
tig, line 304: This is a good example of what I mean above:

Before long someone's going to see this and say "Hey, if I can ask `Y` if it is absolute, why can't I also ask it if it is `PosFactor`? Then they'll want both to be `public`. Etc...

I have a feeling (but I could be wrong) @migueldeicaza  was very intentional in making these classes `private`. 

The good news on this example is at least `Toplevel` is a very "internal" class very tied to `View` (nobody can build their own version of `Toplevel` that replaces it).

Terminal.Gui/Core/PosDim.cs
BDisp: @tig I really am considering what you are asking for but I really need it as I also needed the `internal class DimAbsolute` which is already `internal`. If not, there is no way I know how to differentiate the type. Thanks.

tig: Ok, let's do it. 

But would you do me the favor of creating an Issue to track whether these should be `public` or not? That will force me to continue to think about it, and the implications.

Terminal.Gui/Core/Toplevel.cs
BDisp: @tig I did better, these classes should continue to be `private `rather than `internal`. So I created a PR#778 to remove them.

PR #779, opened 2020-07-12
Fixes #778. Removed internal from PosAbsolute and DimAbsolute.

### LINKED ISSUE #778: PosAbsolute and DimAbsolute should be private.
As @tig rightly stated here https://github.com/migueldeicaza/gui.cs/pull/777#discussion_r453341301, these classes should continue to be private rather than internal.

BDisp:


fergusonr:
> _No description provided._

Curious, whats wrong with 
```csharp
Height is Dim.DimAbsolute
``` 
Why change it to 
```csharp
Height.GetType ().Name == "DimAbsolute"
```
This is dangerous hard-coding

If you really must compare to a string, use  
```csharp
nameof(Dim.DimAbsolute)
```

BDisp:
> 
> 
> > _No description provided._
> 
> @BDisp curious, whats wrong with
> 
> ```cs
> Height is Dim.DimAbsolute
> ```
> 
> Why change it to
> 
> ```cs
> Height.GetType ().Name == "DimAbsolute"
> ```
> 
> This is dangerous hard-coding
> 
> If you really must compare to a string, use
> 
> ```cs
> nameof(Dim.DimAbsolute)
> ```

@fergusonr @tig prefers not to have the private `PosAbsolute `and `DimAbsolute `classes as internal and rightly so. So, if it is just to check what kind of class is derived, I do not necessarily need it to be internal so that it is accessible and in this case it is enough to compare with the name. The `DimAbsolute `class became internal because you needed it for the `ComboBox`, but there are other ways to do it just like I did.

fergusonr:
Thanks understood. I just see this as we, developers of Terminal.Gui, creating restrictions for ourselves and only ourselves. This has been explained to be before, I'm not going to argue the point further!

BDisp:
I know what you mean and I understand you :-)

tig:
Sorry guys, but I'm not understanding what's going on here. 

In my mind, if someone REALLY needs to know the type of `Height` we should expose a REAL API for it. 

Using `GetType` and comparing the string is hacking around the fact there's no API. Because it expose the internal details of `Pos` (by exposing that there's class named `DimAbsolute` it's actually no different, in principle, than making the API public).

@BDisp in my comments about `PosDim` I tried to be careful not to say "we should not expose these internals publicly". Instead I said "I have concerns about doing so (and I think Miguel might too)".  If we have a valid use-case for getting the type of a `PosDim` then we should expose an real API that supports that. 

I am still not convinced either of the use cases requires knowing what type of `Pos` it is. But if YOU guys are convinced then lets define an API that does this in an elegant (not hacky) way.

BDisp:
@tig in yours `PosTests `and `DimTests `you have to do the equality through strings:

`
Assert.Equal ($"Pos.Combine(Pos.Combine(Pos.View(side={side}, target=View()({{X={testRect.X},Y={testRect.Y},Width={testRect.Width},Height={testRect.Height}}}))+Pos.Absolute(0)){(testInt < 0 ? '-' : '+')}Pos.Absolute({testInt}))", pos.ToString ());
`

`
Assert.Equal ("Dim.Absolute(0)", dim.ToString ());
`

What is the difference here?

migueldeicaza:
Hello - I replied on the issue tracking the details, but I think that an interface that probes with `is` an `internal` class is fine and that we do not need this change - the current code is fine.

PR #780, opened 2020-07-12
Fixes #760. Avoiding creating unix driver again after call endwin.

### LINKED ISSUE #760: Colors in Linux changing between windows.

BDisp:


tig:
Ok. I still want to take a look at fixing the Unix driver correctly, but this gets the job done for now! Thanks!

PR #784, opened 2020-07-13
Fixes #783. Menu is now activated by Alt key if the focused view doesn't need it.

### LINKED ISSUE #783: Alt key doesn't activate the Menu anymore.
This happens because only the focused view is called. But if the focused view doesn't handle it, the containers may need to handle them. So, to handle the `Menu `the toplevel must take care on it.

BDisp:


tig:
Great job.

PR #795, opened 2020-07-15
Fixes #226. Added a Initialize method to constructors.

### LINKED ISSUE #226: If ListView is created using the default constructor it cannot be focussed
If you create a `ListView` using the default constructor and set the source later:

```
var listView = new ListView();
listView.SetSource(new string[] { "batman", "robin", "alfred" });
```

then you cannot cycle through the items in the `ListView` even though they are visible.

If you create the `ListView` and set the source like this:

```
var listView = new ListView(new string[] {});
listView.SetSource(new string[] { "batman", "robin", "alfred" });
```

it works just fine.

BDisp:


tig:
Woot!

PR #797, opened 2020-07-15
Fixes #776. TextField and TextView now respect the CanFocus property of his container.

### LINKED ISSUE #776: Tabbing into FrameView containing TextView ignores CanFocus
When using a label within a FrameView you can set the CanFocus property of the label and the FrameView behaves as expected.

This works:

```
            Label appNameView = new Label() { X = 0, Y = 0, Height = Dim.Fill(), Width = Dim.Fill(), CanFocus = false, Text = "some text" };
            _appTitlePane = new FrameView("I shouldn't get focus")
            {
                X = 25,
                Y = 1, // for menu
                Width = Dim.Fill(),
                Height = 9,
                CanFocus = false,
            };
            _appTitlePane.Add(appNameView);
```

However, if you use a TextView within a FrameView and set the CanFocus property - the value is ignored.

```
            TextView appNameView = new TextView() { X = 0, Y = 0, Height = Dim.Fill(), Width = Dim.Fill(), CanFocus = false, ReadOnly = true, Text = "some text" };
            //appNameView.Text = GetAppTitle(true);

            _appTitlePane = new FrameView("I shouldn't get focus")
            {
                X = 25,
                Y = 1, // for menu
                Width = Dim.Fill(),
                Height = 9,
                CanFocus = false,
            };
            _appTitlePane.Add(appNameView);
```

Looking at the TextView the CanFocus property ignores whatever value is set (not sure why?):

https://github.com/migueldeicaza/gui.cs/blob/6bba29813f8c2a8e1403dd5a8119e51887b7c9ad/Terminal.Gui/Views/TextField.cs#L270

This is related to this: https://github.com/migueldeicaza/gui.cs/issues/763 as I'm now using a Label instead of a TextView to honour the leading whitespace.

BDisp:


tig:
@bdisp is on fire! Great stuff.

This reminds me that we need an issue tracking some Focus unit tests. Such a potential for regressions!

Issue #798, opened 2020-07-15
Unit Tests: Add a suite to cover Focus scenarios


tig:
Lots of little things can break in how Focus works. We don't really have any unit tests for this now. It would be a good investment for someone to engineer some. 

Doing so will require some design/engineering of the right test scaffolding to simulate user input beyond what we have now.

tig:
I'm declaring this as fixed by #809.

PR #799, opened 2020-07-15
Just a little Unit Test for Focus #798

BDisp:
- If a container CanFocus property is false it's needed to set all the subviews to false too, it's not magic.
- For the above behavior to work it's needed to change almost all the code that uses SetFocus method to check the CanFocus property before use it.

### DIFF
--- a/UnitTests/ViewTests.cs
+++ b/UnitTests/ViewTests.cs
@@ -718,5 +718,34 @@ namespace Terminal.Gui {
 			Assert.False (v1.CanFocus);
 			Assert.False (v2.CanFocus);
 		}
+
+		[Fact]
+		public void CanFocus_Faced_With_Container ()
+		{
+			var t = new Toplevel ();
+			var w = new Window ();
+			var f = new FrameView ();
+			var v = new View () { CanFocus = true };
+			f.Add (v);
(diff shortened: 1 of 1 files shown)


tig:
> * If a container CanFocus property is false it's needed to set all the subviews to false too, it's not magic.
> * For the above behavior to work it's needed to change almost all the code that uses SetFocus method to check the CanFocus property before use it.

@Bdisp, my apologies, but I'm having a tough time parsing what you've written here. I see nothing wrong with this PR, so I want to just accept it, but your second bullet is making me pause because it seems to imply it won't work. So I'm confused.

BDisp:
> @BDisp, my apologies, but I'm having a tough time parsing what you've written here. I see nothing wrong with this PR, so I want to just accept it, but your second bullet is making me pause because it seems to imply it won't work. So I'm confused.

I understand you :-) What I did meaning with that was that there are much code with the `SetFocus `method to focuse a view when it get focused by mouse that will really going to setting focus to that view even if their `CanFocus `property is setting as false. Maybe it's a good ideia to to a revision for that situations or by on the override mouse events not being calling views that have `CanFocus `setting to false.

PR #801, opened 2020-07-16
Fixes #787. Console default size is always restored on Driver Init.

### LINKED ISSUE #787: UI Catalog - resize scrollbar issue
Title bar not visible after window resize.

- Select Dialog screen
- Double click title bar to go full window
- Double click title bar to minimise window
- Select CTRL Q to go back

Note the vertical scrollbar has not gone to the top.  I've noticed this on ListViews where the current selected item is preserved, causing the scrolling to make that selected item at the top of the screen

Resize scroll issue

BDisp:


tig:
Brilliant.

PR #802, opened 2020-07-16
Fixes #800. Menu IsTopLevel executes on button clicked and setting Run close driver to false because of Unix.

### LINKED ISSUE #800: UICatalog selected screen - menu bar File - Quit causes app to hang (linux only)
Scenario screens that contain a menu bar with the option File - Quit (eg. Editor), cause the application to hang on return to the main menu screen.  This happens on linux only (testing on WSL ubuntu)

- Select Editor screen
- Select File - Quit (from Menu bar)
- You are returned to the main menu - note that the cursor continues to blink - but the app is now unresponsive to any input

Note: Selecting the key combo CTRL-Q (ie. not using the menu bar) from the Editor screen behaves as expected.  So this is related to the menubar only.

BDisp:


tig:
LGTM!

PR #804, opened 2020-07-17
Fixes #803. Init and Shutdown only run once.

### LINKED ISSUE #803: Portable applications must not call initscr() twice
See here https://pubs.opengroup.org/onlinepubs/007908799/xcurses/initscr.html.

I don't know how to do miracles!

BDisp:


tig:
This is great. I'm totally fine with this.

PR #805, opened 2020-07-17
Fixes #755. Click respect the highlight Menu.

### LINKED ISSUE #755: Menu: Mouse clicks outside of text, but within highlight do not register
1. Click on the space just before the "F" in File on the menu like the one one below ; or on the space after the "e".

Expected:

Menu activates.

Actual:

Nothing happens.

Why? Because when the menu IS activated those spaces are painted.

BDisp:


tig:
Thank you!

PR #809, opened 2020-07-18
Fixes #808. Added an automated CanFocus.

### LINKED ISSUE #808: CanFocus can be a little more automatic.
I said here https://github.com/migueldeicaza/gui.cs/pull/799#issue-449767054 that "`If a container CanFocus property is false it's needed to set all the subviews to false too, it's not magic.`" but of course that can be possible. It was a little lazy :-)

BDisp:


BDisp:
Added more unit tests and maybe can Closes #798.

PR #810, opened 2020-07-19
Fixes #731. Pos and Dim validation.

### LINKED ISSUE #731: Frame vs X, Y, Width, Height - Unclear behavior
`LayoutStyle.Absolute` means that `X`, `Y`, `Width`, and `Height` should not/cannot be used. 

If `LayoutStyle.Computed` is set, what happens when setting `Frame` is unclear but it seems it should not be allowed.

We have a lot of code that violates this and this is a source of several bugs.

I think we should either:

a) Throw exceptions if these properties are set when in the wrong mode.

OR

b) Change the `LayoutStyle` to the appropriate value if one of these is set when it shouldn't be, forcing the view into the right mode. 

Currently (but I'd like to debate it to be sure) I like a) best.

An example of where this is clearly messy, and confusing, see this code from `Window` that handles dragging Window with mouse:

```csharp
dragPosition = new Point (nx, ny);
Frame = new Rect (nx, ny, Frame.Width, Frame.Height);
X = nx;
Y = ny;
//Demo.ml2.Text = $"{dx},{dy}";
```

I don't think Views that have `LayoutStyle.Computed` should support being moved because doing so invalidates the `Dim` and `Pos` settings a caller sets. If a `View` has `Width = Dim.Fill()`, for example, 

This code in `Button` is similarly supsect:

```csharp
		internal void Update ()
		{
			if (IsDefault)
				base.Text = ustring.Make (_leftBracket) + ustring.Make (_leftDefault) + " " + text + " " + ustring.Make (_rightDefault) + ustring.Make (_rightBracket);
			else
				base.Text = ustring.Make (_leftBracket) + " " + text + " " + ustring.Make (_rightBracket);

			int w = base.Text.RuneCount - (base.Text.Contains (HotKeySpecifier) ? 1 : 0);
			Width = w;
			Height = 1;
			Frame = new Rect (Frame.Location, new Size (w, 1));
			SetNeedsDisplay ();
		}
```

What thoughts do others have on this?

BDisp:


migueldeicaza:
I would love to have the rationale on the issue, I can only guess what the underlying issue was.

BDisp:
That prevent changing on `Pos `and `Dim `values that aren't null, nor `Pos.Absolute` and nor `Dim.Absolute` giving an exception when they are changed to these values.

tig:
@BDisp do you still want to push this PR. It's been so long that I don't remember the details and why I was reluctant to accept it. 

If you still think we should accept it please let me know and fix the merge issues.

BDisp:
> 
> 
> @BDisp do you still want to push this PR. It's been so long that I don't remember the details and why I was reluctant to accept it.
> 
> If you still think we should accept it please let me know and fix the merge issues.

@tig I think it's useful. I'll fix the merge issues.

Issue #816, opened 2020-07-21
ComboBox render bug at dialog window


lzylinl:
If we add combobox to dialog window, combobox rendered incorrectly.
For example change a bit end of "Dialogs : Scenario"

```csharp
var btnText = new [] { "Zero", "One", "Two", "Three", "Four", "Five", "Six", "Seven", "Eight", "Nine" };
var showDialogButton = new Button ("Show Dialog") {
	X = Pos.Center(),
	Y = Pos.Bottom (frame) + 2,
	IsDefault = true,
	Clicked = () => {
		try {
			int width = int.Parse (widthEdit.Text.ToString ());
			int height = int.Parse (heightEdit.Text.ToString ());
			int numButtons = int.Parse (numButtonsEdit.Text.ToString ());

			var buttons = new List<Button> ();
			var clicked = -1;
			for (int i = 0; i < numButtons; i++) {
				var buttonId = i;
				var button = new Button (btnText [buttonId % 10], is_default: buttonId == 0) {
					Clicked = () => {
						clicked = buttonId;
						Application.RequestStop ();
					},
				};
				buttons.Add(button);
			}

			// This tests dynamically adding buttons; ensuring the dialog resizes if needed and 
			// the buttons are laid out correctly
			var dialog = new Dialog (titleEdit.Text, width, height, buttons.ToArray ());
			var add = new Button ("Add a button") {
				X = Pos.Center (),
				Y = Pos.Center (),
				Clicked = () => {
					var buttonId = buttons.Count;
					var button = new Button (btnText [buttonId % 10], is_default: buttonId == 0) {
						Clicked = () => {
							clicked = buttonId;
							Application.RequestStop ();
						},
					};
					buttons.Add (button);
					dialog.AddButton (button);
				},
			};
			dialog.Add (add);


			////////// BUG

			var testCombo = new ComboBox () {
				ColorScheme = Colors.Dialog,
				X = Pos.Center (),
				Y = Pos.Center () + 3,
				Width = 20,
				Height = 4
			};
			testCombo.SetSource (new List<string> () { "Test 1", "Test 2", "Test 3", "Test 4", });
			dialog.Add (testCombo);

			////////// BUG


			Application.Run (dialog);
			buttonPressedLabel.Text = $"{clicked}";

		} catch (FormatException) {
			buttonPressedLabel.Text = "Invalid Options";
		}
	},
};
Win.Add (showDialogButton);

////////// ALL OK

[...]

PR #819, opened 2020-07-22
ComboBox render fix at dialog after close listview.

lzylinl:
Seems this right way

### DIFF
--- a/Terminal.Gui/Views/ComboBox.cs
+++ b/Terminal.Gui/Views/ComboBox.cs
@@ -499,6 +499,7 @@ namespace Terminal.Gui {
 		private void HideList ()
 		{
 			Reset (SelectedItem > -1);
+			listview.Clear ();
 			listview.TabStop = false;
 		}
 


tig:
@BDisp - do agree this is a good fix?

BDisp:
I'll see and then I'll give feedback

BDisp:
It seems that this does not influence anything but as @fergusonr  submitted a PR earlier, maybe it would be better for him to comment, please.

fergusonr:
This PR has no effect. @lzylinl is there a specific issue this fixes?

lzylinl:
Look at this issue https://github.com/migueldeicaza/gui.cs/issues/816
Use same code for reproduce.
If we open and close combobox (not select anything) we can see this
Wew72V3
Area from listbox does not cleared after combo closed

BDisp:
Ok. @tig I confirm that this PR fixes the issue #816.

fergusonr:
Thankyou both @lzylinl  @BDisp

tig:
nicely done folks.

PR #823, opened 2020-07-24
Fixes #822. Moves the cursor position to the first rune if no hotkey is specified.

### LINKED ISSUE #822: Button text without hotkey should move the cursor position to the first rune.

BDisp:


tig:
Excellent.

PR #824, opened 2020-07-24
Fixes #750. Ensure the correct focus order after call SetFocus.

### LINKED ISSUE #750: SetFocus() change focus order by Tab
Hello,

After applying SetFocus() to TextView the way of switching by Tab elements is changing. For example: 

```csharp
﻿using Terminal.Gui;

namespace UICatalog {
	[ScenarioMetadata (Name: "FocusOrder", Description: "Demonstrate change of focus order after SetFocus()")]
	[ScenarioCategory ("Bug Repro")]
	class FocusOrder : Scenario {
	    class TextFrame : FrameView {
		    private TextView textField;
		    public TextFrame (string title) : base (title) {
			    textField = new TextView () {
				    X = 0,
				    Y = 0,
				    Width = Dim.Fill () - 15,
				    Height = Dim.Fill(),
				    ColorScheme = Colors.Dialog,
				    Text = "0"
			    };
			    var button1 = new Button ("Button 1…") {
				    X = Pos.Right(textField)+1,
				    Y = Pos.Top(textField),
				    Clicked = () => { Button1Handler();}
			    };
			    var button2 = new Button ("Button 2…") {
				    X = Pos.X(button1),
				    Y = Pos.Y(button1)+2,
				    Clicked = () => { Button2Handler();}
			    };
			    Add(textField, button1, button2);
		    }
		    public bool showMessage() {
			    var d =MessageBox.ErrorQuery("Error", "Error", "Ok");
			    this.SetFocus(textField);
			    return true;
		    }
		    private void Button1Handler() {
		    }
		    private void Button2Handler() {
			    showMessage();
		    }
	    }
		public override void Setup () {
		    var tf = new TextFrame ("Frame") {
			    X = 0,
			    Y = 0,
			    Width = Dim.Fill(),
			    Height = Dim.Fill() - 3
		    };
		    var button3 = new Button("Button 3") {
			    X = Pos.Percent(50) - 6,
			    Y = Pos.Bottom(tf) + 1,
		    };
		    Win.Add(tf, button3);
		}
	}
}
```

When focus changed by "Button 2" to TextView pressing Tab switch focus to "Button 3". Estimated order is "Button 1"

BDisp:
- For this to work it's needed to call `SetFocus` with one of the two methods:
`view.SuperView.SetFocus (view)`
or
`view.SetFocus ()`
This is a new method more simple and avoid focus order issues.

- Call like `SuperView.SetFocus (view)` does not ensures that the entire parent hierarchy can really get focus and thus not updating the focus order.

### DIFF
--- a/Example/demo.cs
+++ b/Example/demo.cs
@@ -253,12 +253,7 @@ static class Demo {
 		});
 		ntop.Add (menu);
 
-		string fname = null;
-		foreach (var s in new [] { "/etc/passwd", "c:\\windows\\win.ini" })
-			if (System.IO.File.Exists (s)) {
-				fname = s;
-				break;
-			}
+		string fname = GetFileName ();
 
 		var win = new Window (fname ?? "Untitled") {
 			X = 0,
@@ -275,6 +270,18 @@ static class Demo {
 		win.Add (text);
 
 		Application.Run (ntop);
+	}
+
+	private static string GetFileName ()
+	{
+		string fname = null;
+		foreach (var s in new [] { "/etc/passwd", "c:\\windows\\win.ini" })
+			if (System.IO.File.Exists (s)) {
+				fname = s;
+				break;
+			}
+
+		return fname;
 	}
 
 	static bool Quit ()
@@ -312,7 +319,8 @@ static class Demo {
 		});
 		ntop.Add (menu);
 
-		var win = new Window ("/etc/passwd") {
+		string fname = GetFileName ();
+		var win = new Window (fname) {
 			X = 0,
 			Y = 1,
 			Width = Dim.Fill (),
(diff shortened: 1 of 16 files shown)


BDisp:
In my opinion I think the original method shouldn't be public but only the new method. The problem is the backward compatibility.

tig:
> In my opinion I think the original method shouldn't be public but only the new method. The problem is the backward compatibility.

We have not released 0.90 yet, so breaking backwards compat is an ok thing to do...as long is it isn't super-deep. In fact, if we really think something should be changed we should change it. I don't have a strong opinion on this particular issue myself (I have not dug in deep on how focus works). Your call.

BDisp:
Great I'll do the change so.

tig:
Feels so much more correct. Well done.

tig:
I think you need to rebase though...

BDisp:
> 
> 
> I think you need to rebase though...

Done. Thanks @tig.

BDisp:
@tig I made some changes in the `demo.cs`. Sorry about that.

PR #837, opened 2020-07-31
Fixes #836. Allows a Label display the text initially empty.

### LINKED ISSUE #836: Changes to the Label won't display if x, y and empty text are initially defined.
For example:

`var label = new Label (5, 7, "");`

Later changes on the Label never will be shown. First the location will be empty by the `TextFormatter `if the text is empty. Only the size should be empty and not the location.

BDisp:


tig:
Nice. Sorry I've been absent lately. Life, you know...

PR #839, opened 2020-08-01
Fixes #838. Added a Visible property to the View.

### LINKED ISSUE #838: "Hiding" an element, but keeping it's position for when it's visible again?
Sorry if this seems like a poor question, I've not really worked in a text mode framework. How would one "hide" an element on the screen that I want to make visible later? I have a procedure that uses 3 lines of text to describe the current operations. At various points I would like to make the item 'invisible' but without setting the text blank for the label that it is associated with. Is there a concept for this or am I forced to clear the label to make it not visible? I did not see anything in the documentation for something like this.

BDisp:


tig:
Can you describe the use case where this is useful? I think the code is fine, but I'm curious what drove the need for this since there's no test case included.

BDisp:
> 
> 
> Can you describe the use case where this is useful? I think the code is fine, but I'm curious what drove the need for this since there's no test case included.

@tig that is very simple. You only need to set Visible property to false on any view and all his subviews aren't drawn and set to true to get his drawn again.

tig:
I get that, but I'm asking for a description of a real-life use case where an app would do this.

BDisp:
Oh sorry. In windows is used often. For example, if you have a window with two frame and some conditional query that only shows some data related only for one frame you can hide the other frame from showing. Sometimes is needed to hide instead of disabling. It's only an improvment feature that some users like to use.

PR #843, opened 2020-08-01
Fixes #842. A Toplevel must be created by Application.Init.

### LINKED ISSUE #842: UI-TopLevelNoWindowBug only Application.Init can create Toplevel
In this case, create an implicit toplevel does´t make it act as a really `Application.Top`. So `UICatalog `already has an `Application.Top` that is named of `Top `and that this we should used to recreate a new toplevel. So that is not a bug but the way how `Gui `work.

BDisp:


BDisp:
Uncomment `//Top = new Toplevel (new Rect (0, 0, Application.Driver.Cols, Application.Driver.Rows));` and on the menu click "_Выход" (Quit) and drag the dialog around the window and you'll see that the window will not be redrawning became black that is the color of toplevel.

toplevel-bug

PR #845, opened 2020-08-02
Fixes #844. Now selects the correct Directory.

### LINKED ISSUE #844: OpenDialog directory chooser UI is unintuitive and returns wrong or otherwise illogical paths in regards to what is displayed in the UI
I've been working on some stuff in my application to handle a user picking a location for a folder and I've had some significant challenges myself with the way the OpenDialog folder chooser works. It is not really obvious what will be returned. Here's some examples.

2020-08-01_21h41_12

**What I expect:** C:\Users\Mgamerz\Documents\MEM_Packages
**What is returned:**  C:\Users\Mgamerz \Documents
**Why is this an issue?**
The 'Open' text field, and the 'Open' button, have two different values. When I click Open, which I logically could assume is the folder being selected in the 'Open' text field (plus the directory above it, which is the currently displayed directory), is not what is returned. It returns the top directory instead. This is confusing as the wording makes it look like this is what you are selecting. If this is not the case, I'm not sure why you would ever need to have 'Open' since I can see what is being selected already (I guess unless you have a multi-select).

2020-08-01_21h41_17

**What I expect:** C:\Users\Mgamerz\Documents\MEM_Packages
**What is returned:**  C:\Users\Mgamerz 
**Why is this an issue?**
The directory selected is _two_ whole directories away. Even if I was choosing '..' (why would anyone choose .. as a directory and not just choose the parent itself?), it is clearly not what I was trying to select. This is just plain wrong. 

2020-08-01_21h41_21

**What I expect:** C:\Users\Mgamerz\Documents\MEM_Packages\ME1
**What is returned:**  C:\Users\Mgamerz\Documents\MEM_Packages
**Why is this an issue?**
The dialog looks like I'm choosing the ME1 folder for opening, but it's just opening the current path instead. While ultimately this is the folder I wanted, the way to select this is nothing like I've seen on other platforms such as Windows, Mac, and GUI based Linux.

BDisp:
- I recommend to navigate through keyboard or mouse for the File text because there are no control about the input. Only the directory is controlled to allow changing the drive.

- I changed the label name from Open to File(s) to avoid confusing about directiories and files, because it's only files can be used here and can be multi-selection.

- To select only a directory there must needed to enter to it too.

### DIFF
--- a/Example/demo.cs
+++ b/Example/demo.cs
@@ -305,7 +305,7 @@ static class Demo {
 		Application.Run (d);
 
 		if (!d.Canceled)
-			MessageBox.Query (50, 7, "Selected File", string.Join (", ", d.FilePaths), "Ok");
+			MessageBox.Query (50, 7, "Selected File", d.FilePaths.Count > 0 ? string.Join (", ", d.FilePaths) : d.FilePath, "Ok");
 	}
 
 	public static void ShowHex (Toplevel top)
--- a/Terminal.Gui/Windows/FileDialog.cs
+++ b/Terminal.Gui/Windows/FileDialog.cs
@@ -347,13 +347,15 @@ namespace Terminal.Gui {
 			}
 		}
 
-		internal bool ExecuteSelection ()
+		internal bool ExecuteSelection (bool isPrompt = false)
 		{
 			var isDir = infos [selected].Item2;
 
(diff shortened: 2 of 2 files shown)


tig:
LGTM.

PR #846, opened 2020-08-03
Fixes #644. Added a UICatalog Scenario for a dynamic menu bar.

### LINKED ISSUE #644: MenuBar: Add a rich UI Catalog example showing dynamic menus
We have plenty of examples of `MenuBar` in both the main `UI Catalog` app and `demo.cs`.

However, we are missing an example that illustrates a dynamic menu where:

- MenuBarItems and MenuItems are added and removed dynamically at runtime.
- MenuItem Titles changes dynamically

I'm confident this will expose bugs and design issues (e.g. similar to #643 ).

BDisp:
I removed the sub-menu constructor because the children already are a sub-menu and I replaced by the Parent that wasn't exists and it is indeed needed. I made another adjustments for compatibility to the new changes. I hope you like it.

### DIFF
--- a/Example/demo.cs
+++ b/Example/demo.cs
@@ -586,7 +586,7 @@ static class Demo {
 			new MenuItemDetails ("F_ind", "", null),
 			new MenuItemDetails ("_Replace", "", null),
 			new MenuItemDetails ("_Item1", "", null),
-			new MenuItemDetails ("_Not From Sub Menu", "", null)
+			new MenuItemDetails ("_Also From Sub Menu", "", null)
 		};
 
 		menuItems [0].Action = () => ShowMenuItem (menuItems [0]);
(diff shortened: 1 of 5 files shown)


BDisp:
I added the `StatusBar `fix here because there was already changes in the `Toplevel.cs`. Thanks.

BDisp:
> 
> 
> I added the `StatusBar `fix here because there was already changes in the `Toplevel.cs`. Thanks.

I forgot to mention that this issue only happen on the `Linux`, because `WindowsDriver `don't allow resizing on shrinking.

BDisp:
It only breaking on the `MenuBar `constructor if there are many submenus. See the `demo.cs` that I had to change some of them. But since you said we can make some changes before the launch of the new `GUI `version, I have super confidence that it will work better than before.

PR #848, opened 2020-08-03
Fixes #847. Since the timeout issue are fixed there is no need to use tokenSource.Cancel but just reset the eventReady.

### LINKED ISSUE #847: OperationCanceledException is frustrating to deal with during development
I've been working with this library for about a week and a half now and the 'System.OperationCanceledException' that is thrown on pretty much every interaction (it's caught, but that doesn't stop the side effects) is very frustrating to deal with. I've turned this exception off entirely in my debugger (otherwise it will endlessly pause the app) but it also clogs up the debug output.

image

Is there a way to mitigate this?

BDisp:


Mgamerz:
In things that do a lot UI updates (such as updating a progressbar objects fraction) this GREATLY improves app speed. I was doing a LAN file copy and went from like 80Mbps to 580Mbps simply because the exception output was no longer stalling the application in debug mode when I updated the progress bar progress.

PR #856, opened 2020-08-08
Fixes #844. The SelectionChanged action it's now working correctly.

### LINKED ISSUE #844: OpenDialog directory chooser UI is unintuitive and returns wrong or otherwise illogical paths in regards to what is displayed in the UI

BDisp:


Mgamerz:
Did a checkout of this PR and things seem to behave like expected now. In folder chooser mode, picking .. still picks the same viewed directory. If nothing is selected and you press open, it also chooses the current directory. If you select a folder and hit open, that folder is returned. So now I can select current directory or subdirectory in a UI consistent manner 👍

BDisp:
> In folder chooser mode, picking .. still picks the same viewed directory when returned on `FilePath`

`FilePath `always return the current path plus the selected items. To check what it's returned from `OpenDialog `always use the `FilePaths `for both, folder(s) and file(s). If nothing is selected `FilePath `return the current folder and `FilePaths `return a empty list.
FYI since I implemented the wheel mouse feature into `GUI `it's possible to select many itens at once by clicking the first item and using the Shift key with the mouse select the last item and the items status will be toggled, thus marrking or unmarking dependent the previous item status.

PR #883, opened 2020-08-26
Fix  #882

### LINKED ISSUE #882: NullReferenceException during keyboard navigation
Disclaimer: I'm new to this project, so maybe I did something stupid. That said, my code is an example from the doc...

I'm on the top of master (96208a1c490401b503ee62dde837cbf164180def) and I get a `NullReferenceException at Terminal.Gui.Toplevel.GetDeepestFocusedSubview(View view) in C:\git\gui.cs\Terminal.Gui\Core\Toplevel.cs:line 218` when pressing the `tab` or an arrow key, because the `Focused` property is null.

Here is my application code:

```csharp
Application.Init ();

		new MenuItem ("_Quit", "", () => {
			Application.RequestStop ();
		})
	}),
});

var win = new Window ("Hello") {
	X = 0,
	Y = 1,
	Width = Dim.Fill (),
	Height = Dim.Fill () - 1
};

// Add both menu and win in a single call
Application.Top.Add (menu, win);

Application.Run ();
```

The crash occurs when I click the "File" menu twice (so that it's focused but not opened), and the press tab or the right arrow key.

---

If the fix is just to add a null check, I might be able to open a PR. If it's more deeply rooted, I'm all ears for solutions!

mlaily:


tig (approved):
LGTM. Sorry about the delay.

mlaily:
No problem, thanks :)

Issue #886, opened 2020-09-02
TextView / ScrollToEnd


blakepell:
Build: 0.89.4

New to using this project so there could be a better way to do what I'm intending, I was unable to figure it out without using some of the internals.  Basically, I'm using a TextView to display the contents of a running log.  As new text is added from an `async Task` I want the text view to scroll to the end.

Flow would be:

1. Scroll to the last line minus the frame height so that you see the last line _but also_ the last screen of information (scrolling to just the last line with `ScrollTo` shows only the last line in my testing).
2. Position the cursor on the last line.

I had attempted to add this in my code, wasn't sure how to get the current number of lines from the `TextView` so I was counting line breaks which I'd prefer to avoid since it already exists.  In the `TextView` control the `model.Count` has that value and it additionally has the current frame height which I couldn't figure out how to get at from my own code.  I created this function I put in `TextView`, I'm curious if it would be useful in some form (or if there's a way to accomplish this in my code without adding a function to TextView).

```
/// <summary>
/// Will scroll the <see cref="TextView"/> to the last line and position the cursor there.
/// </summary>
public void ScrollToEnd ()
{
	// Subtract off the frame's height from the current number of lines so that
	// when we scroll to the bottom we show the last screen worth of lines and not
	// just the last line.
	topRow = Math.Max (1, model.Count - Frame.Height);
	SetNeedsDisplay ();

	// Set the cursor to the last row and then position it.
	currentRow = (model.Count > 0) ? currentRow = model.Count - 1 : 0;
	PositionCursor ();
}
```

I tested this in cmd and Powershell on Windows and via Putty on Ubuntu 18.04.

biohazard999:
Want to do the Same thing!

BDisp:
There is a bug in `TextView `at line 1007. It must be `currentRow = model.Count - 1;`
So you can replace this:

```cs
			case Key.CtrlMask | Key.End:
				currentRow = model.Count - 1;
				TrackColumn ();
				PositionCursor ();
				break;

			case Key.CtrlMask | Key.Home:
				currentRow = 0;
				TrackColumn ();
				PositionCursor ();
				break;

```
to this:

```cs
			case Key.CtrlMask | Key.End:
				ScrollToEnd ();
				break;

			case Key.CtrlMask | Key.Home:
				ScrollToHome ();
				break;

public void ScrollToEnd ()
{
	currentRow = model.Count - 1;
	TrackColumn ();
	PositionCursor ();
}

[...]

PR #888, opened 2020-09-03
TextView MoveHome and MoveEnd methods.  #886

blakepell:
These are the public `MoveHome` and `MoveEnd` methods to reposition cursor and scroll the `TextView` discussed in #886.


tig (approved):
Nice. Thanks. and sorry for the delay.

PR #891, opened 2020-09-06
ListView.cs fixes

Mgamerz:
When calling MoveUp and MoveDown, the 'selected' variable is not checked. If I am using a backing data source that can change, this selected value +/- 1 may still not lie in the valid range of values. This PR checks and changes the following:

1) If moving up, and the CURRENT selected value is >= backing data length, the last item is selected
2) If moving down, and the CURRENT selected value is >= backing data length, the last item is selected
3) If the items list has no items, the methods return false as there is nothing to move to. I'm not sure if we should set lastSelectedItem to -1 in this instance.

Sorry if formatting is weird. I wrote this in visual studio code on linux and haven't figured out where the formatter is...

### DIFF
--- a/Terminal.Gui/Views/ListView.cs
+++ b/Terminal.Gui/Views/ListView.cs
@@ -1,4 +1,4 @@
-﻿//
+//
 // ListView.cs: ListView control
 //
 // Authors:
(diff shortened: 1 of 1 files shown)


tig (changes requested):
Please fix the formatting. Thanks.

Mgamerz:
Fix the formatting.

Mgamerz:
Added a commit to fix pressing enter on an empty list throwing an exception.

BDisp:
@Mgamerz I think you have to fix the formatting again :-)

tig:
Closing as I've replaced it with https://github.com/migueldeicaza/gui.cs/pull/906

PR #900, opened 2020-09-18
Add custom resolver for libcoreclr.so

### LINKED ISSUE #889: Doesn't work with .NET 5 single file apps
Follow instructions at https://github.com/benaadams/gui-linked to publish as a single file on Linux.

libcoreclr.so is an internal implementation detail of the .NET runtime. It shouldn't be depended on by 3rd party code.

```
Curses failed to initialize, the exception is: System.DllNotFoundException: Unable to load shared library 'libcoreclr.so' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: liblibcoreclr.so: cannot open shared object file: No such file or directory
   at Mono.Terminal.Internal.UnmanagedLibrary.CoreCLR.dlopen(String filename, Int32 flags)
   at Mono.Terminal.Internal.UnmanagedLibrary.PlatformSpecificLoadLibrary(String libraryPath)
   at Mono.Terminal.Internal.UnmanagedLibrary..ctor(String[] libraryPathAlternatives, Boolean isFullPath)
   at Unix.Terminal.Curses.LoadMethods()
   at Unix.Terminal.Curses.FindNCurses()
   at Unix.Terminal.Curses.initscr()
   at Terminal.Gui.CursesDriver.Init(Action terminalResized)
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Unix.Terminal.Curses.raw()
   at Terminal.Gui.CursesDriver.Init(Action terminalResized)
   at Terminal.Gui.Application.Init(Func`1 topLevelFactory)
   at Terminal.Gui.Application.Init()
   at Demo.Main()
Aborted
```

am11:
This patch is required for "true" SingleFile form factor, introduced in .NET 5.0.
I tested @benadams' repro (by using replacing PackageReference with ProjectReference and commenting out [this line](https://github.com/benaadams/gui-linked/blob/902f49a90742c9fe0d7d3b2f83d5d7885c360b19/Program.cs#L65)) and confirmed this special-casing for .NET5.0 (using SDK's implicitly defined compilation constant `NET5_0`)  fixes the problem.

Fixes #889.

cc @MichalStrehovsky, @benadams

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
@@ -170,7 +170,7 @@ namespace Unix.Terminal {
 			if (ptr == IntPtr.Zero) {
 				throw new MissingMethodException (string.Format ("The native method \"{0}\" does not exist", methodName));
 			}
-            		return Marshal.GetDelegateForFunctionPointer<T>(ptr);  // non-generic version is obsolete
+			return Marshal.GetDelegateForFunctionPointer<T>(ptr);  // non-generic version is obsolete
 		}
 
 		/// <summary>
@@ -258,6 +258,17 @@ namespace Unix.Terminal {
 		/// </summary>
 		static class CoreCLR
 		{
+#if NET5_0
+			// Custom resolver to support true single-file apps
+			// (those which run directly from bundle; in-memory).
+			//     -1 on Unix means self-referencing binary (libcoreclr.so)
+			//     0 means fallback to CoreCLR's internal resolution
+			// Note: meaning of -1 stay the same even for non-single-file form factors.
+			static CoreCLR() =>  NativeLibrary.SetDllImportResolver(typeof(CoreCLR).Assembly,
+				(string libraryName, Assembly assembly, DllImportSearchPath? searchPath) =>
+					libraryName == "libcoreclr.so" ? (IntPtr)(-1) : IntPtr.Zero);
+#endif
+
 			[DllImport ("libcoreclr.so")]
 			internal static extern IntPtr dlopen (string filename, int flags);
 
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -30,6 +30,8 @@ using System.Linq;
 using System.Text;
 using NStack;
 
+using Rune = System.Rune;
+
 namespace Terminal.Gui {
 	class TextModel {
 		List<List<Rune>> lines = new List<List<Rune>> ();
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -33,8 +33,17 @@ jobs:
 
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
-    - name: Autobuild
-      uses: github/codeql-action/autobuild@v1
+    # - name: Autobuild
+    #  uses: github/codeql-action/autobuild@v1
+
+    - name: Setup .NET Core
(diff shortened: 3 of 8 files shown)


Terminal.Gui/Views/TextView.cs
```diff
@@ -30,6 +30,8 @@
 using System.Text;
 using NStack;
 
+using Rune = System.Rune;
```
am11, line 33: This to fix a compiler error; disambiguate .NET5's System.Text.Rune with System.Rune coming from https://github.com/migueldeicaza/NStack/blob/447f61d/NStack/unicode/Rune.cs#L4.

tig (approved):
This all looks great. I don't have time to test and I want to do that before I merge. But I will asap.

Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
```diff
@@ -258,6 +258,18 @@ static class Mono
 		/// </summary>
 		static class CoreCLR
 		{
+#if NET5_0
+			static CoreCLR() =>  NativeLibrary.SetDllImportResolver(typeof(CoreCLR).Assembly,
+				(string libraryName, Assembly assembly, DllImportSearchPath? searchPath) => {
+					if (libraryName == "libcoreclr.so") {
+						var module = typeof(CoreCLR).Module;
+						return Marshal.GetHINSTANCE(module);
```
MichalStrehovsky, line 266: `Marshal.GetHINSTANCE` is just an obscure way to say `(IntPtr)(-1)` outside Windows. Would it be better to just use -1 to make this more explicit?

https://github.com/dotnet/runtime/blob/e7204f5d6fcaca5e097ec854b3be6055229fc442/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.NoCom.cs#L103

MichalStrehovsky: (We are going to add an analyzer to mark single-file unfriendly APIs and GetHINSTANCE is one of such APIs because on Windows, the behavior of the API is going to change once you turn on single-file. On Linux it won't change anything.)

https://github.com/dotnet/runtime/issues/38405

am11: Neat! simplified the resolver. :) I did not realize that GetHINSTANCE simply returns -1 for Unix from GetINSTANCE. I adapted the algorithm in response to Mono's `__Internal` magic string request for self-referencing binary presented here: https://github.com/dotnet/runtime/issues/7267#issuecomment-613516993 (which is a cross-platform solution and here we are only concerned about Linux).

Explicit support of `__Internal` or equivalent would be a nice addition in 6.0 and it can potentially simplify the code in this file. However, as long as we need 5.0 support, we would need the custom resolver here.

tig:
Is .NET 5 at a point where we would be comfortable releasing a nuget package built with it?

am11:
The .NET 5.0 general availability is rumored for Nov. 2020. Some NuGet packages are using dotnet 5.0 RC versions for their preview / beta releases. If there is a stable version planned for `gui.cs` before Nov, we can postpone merging this PR (otherwise a preview version would be welcome meanwhile to obtain feedback from masses). 🙂

BDisp:
> 
> 
> Is .NET 5 at a point where we would be comfortable releasing a nuget package built with it?

Still in `Release Candidate`. I only like to deal with them when they are published with the last `Visual Studio` version.

tig:
My recommendation is that we merge this change now for 0.90 and plan on launching 1.0 after .NET 5 launches in Nov.

Any objections?

BDisp (approved):

am11:
Sounds good, will likely give us more bake time to iron out possible bugs in 5.0 bits. 😎👍

PR #902, opened 2020-09-21
Fixes #254. Processes shortcut menu by combining the keys Ctrl+Alt+Alfa.

### LINKED ISSUE #254: Menu.ShortCut is not hooked up to anything
Menu.ShortCut has no effect.  Looking at the code, it's never used. It can be set, but is never compared against anything so it can fire the MenuItem.Action.

BDisp:
- I decided use the `Ctrl+Alt+Alfa` because some views already subscribe the `Ctrl+Alfa` and thus avoiding conflict.
- Only `Ctrl+Alt+M` in Linux always return `ControlJ` and I don't know why.
- If a editor view has the focus the `Ctrl+Alt+E` will write `€` instead instead run the `ShortCut` menu.

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -434,6 +434,10 @@ namespace Terminal.Gui {
 		public Pos X {
 			get => x;
 			set {
+				if (!ValidatePosDim (x, value)) {
+					throw new ArgumentException ();
+				}
+
 				x = value;
 				SetNeedsLayout ();
 				SetNeedsDisplay (frame);
@@ -450,6 +454,10 @@ namespace Terminal.Gui {
 		public Pos Y {
 			get => y;
 			set {
+				if (!ValidatePosDim (y, value)) {
+					throw new ArgumentException ();
+				}
+
 				y = value;
 				SetNeedsLayout ();
 				SetNeedsDisplay (frame);
@@ -468,6 +476,10 @@ namespace Terminal.Gui {
 		public Dim Width {
 			get => width;
 			set {
+				if (!ValidatePosDim (width, value)) {
+					throw new ArgumentException ();
+				}
+
 				width = value;
 				SetNeedsLayout ();
 				SetNeedsDisplay (frame);
@@ -482,10 +494,27 @@ namespace Terminal.Gui {
 		public Dim Height {
 			get => height;
 			set {
+				if (!ValidatePosDim (height, value)) {
+					throw new ArgumentException ();
+				}
+
 				height = value;
 				SetNeedsLayout ();
 				SetNeedsDisplay (frame);
 			}
+		}
+
+		bool ValidatePosDim (object oldvalue, object newValue)
+		{
+			if (!IsInitialized || layoutStyle == LayoutStyle.Absolute || oldvalue == null || oldvalue.GetType () == newValue.GetType () || this is Toplevel) {
+				return true;
+			}
+			if (layoutStyle == LayoutStyle.Computed) {
+				if (oldvalue.GetType () != newValue.GetType () && !(newValue is Pos.PosAbsolute || newValue is Dim.DimAbsolute)) {
+					return true;
+				}
+			}
+			return false;
 		}
 
 		/// <summary>
--- a/Terminal.Gui/Views/StatusBar.cs
+++ b/Terminal.Gui/Views/StatusBar.cs
@@ -88,11 +88,16 @@ namespace Terminal.Gui {
 			CanFocus = false;
 			ColorScheme = Colors.Menu;
 			X = 0;
-			Y = SuperView != null ? SuperView.Frame.Height - 1 : Pos.AnchorEnd (1);
 			Width = Dim.Fill ();
 			Height = 1;
 
+			Initialized += StatusBar_Initialized;
 			Application.Resized += Application_Resized ();
+		}
+
+		private void StatusBar_Initialized (object sender, EventArgs e)
+		{
+			Y = SuperView.Frame.Height - 1;
 		}
 
 		private Action<Application.ResizedEventArgs> Application_Resized ()
--- a/UnitTests/DimTests.cs
+++ b/UnitTests/DimTests.cs
@@ -233,6 +233,91 @@ namespace Terminal.Gui {
 			Assert.Throws<ArgumentException> (() => dim = Dim.Percent (1000001));
 		}
 
+		[Fact]
+		public void Dim_Validation_Throws_If_NewValue_Is_DimAbsolute_And_OldValue_Is_Another_Type ()
+		{
+			Application.Init (new FakeDriver (), new NetMainLoop (() => FakeConsole.ReadKey (true)));
+
+			var t = Application.Top;
+
+			var w = new Window ("w") {
+				Width = Dim.Fill (0),
+				Height = Dim.Sized (10)
+			};
+			var v = new View ("v") {
+				Width = Dim.Width (w) - 2,
+				Height = Dim.Percent (10)
+			};
+
+			w.Add (v);
+			t.Add (w);
+
+			t.Ready += () => {
+				Assert.Equal (2, w.Width = 2);
+				Assert.Equal (2, w.Height = 2);
+				Assert.Throws<ArgumentException> (() => v.Width = 2);
+				Assert.Throws<ArgumentException> (() => v.Height = 2);
+			};
+
+			Application.Iteration += () => Application.RequestStop ();
+
+			Application.Run ();
+			Application.Shutdown ();
+		}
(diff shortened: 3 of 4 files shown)


tig:
I may be missing something here. 

Wouldn't it be simpler to just to delete the `Menu.Shortcut` field and just have users depend on the fact that any menu item with text preceeded by an underscore (`_`) will get that character as an `Alt-` shortcut? 

I didn't see #254 asking for `menu.Shortcut` to be fixed, specifically.

tig:
I'm not sure why we have to restrict this to ctrl-alt-alpha keystrokes. Why not just allow the user to specify whatever they want. If it conflicts with something else, then too bad. 

I say this because, for example, I couldn't use "F1" for `Help` or "Ctrl-V' for `Edit.Paste` with this PR. 

Right?

tig:
Also, we should provide good samples for this. Right now the main `UICatalog` doesn't try to use this API but should.

For example, it uses `StatusBar` to define `Ctrl-Q` as quit *and* `File._Quit`. 

I intentionally made `_About` be a top-level menu so it could be a test case for single-item top-levels, so I'm leery of suggesting this, but here goes:

Update UI Catalog to have a `_Help` menu. This would include:

     _gui.cs API Overview  -F1
     _gui.cs README
     _About

`F1` or `gui.cs API Overview ` would take the user to https://migueldeicaza.github.io/gui.cs/articles/overview.html

`gui.cs README` would take the user to https://github.com/migueldeicaza/gui.cs

etc...

Also, I don't think we have any code that SHOWS the shortcut in the menu, do we?

BDisp:
> 
> 
> I'm not sure why we have to restrict this to ctrl-alt-alpha keystrokes. Why not just allow the user to specify whatever they want. If it conflicts with something else, then too bad.
> 
> I say this because, for example, I couldn't use "F1" for `Help` or "Ctrl-V' for `Edit.Paste` with this PR.
> 
> Right?

I'm still working to improve this. As you can see in the branch name of PR this a lazy workaround. Of course it's should be better to the user choose what keys they want but for that it's necessary to write in the menu the shortcut keys and the help should be showed throught a popmenu. If a `TextField `has the focus and some `Ctrl+Alfa` keys are subcribed them, the same menu shorcuts won't be triggered and that why I'm using the `Ctrl+Alt+Alfa.`

BDisp:
> Also, I don't think we have any code that SHOWS the shortcut in the menu, do we?

No we don't have. It'll be a challenge but it not to soon.

tig:
> > I'm not sure why we have to restrict this to ctrl-alt-alpha keystrokes. Why not just allow the user to specify whatever they want. If it conflicts with something else, then too bad.
> > I say this because, for example, I couldn't use "F1" for `Help` or "Ctrl-V' for `Edit.Paste` with this PR.
> > Right?
> 
> I'm still working to improve this. As you can see in the branch name of PR this a lazy workaround. Of course it's should be better to the user choose what keys they want but for that it's necessary to write in the menu the shortcut keys and the help should be showed throught a popmenu. If a `TextField `has the focus and some `Ctrl+Alfa` keys are subcribed them, the same menu shortcuts won't be triggered and that why I'm using the `Ctrl+Alt+Alfa.`

I think this is over protecting the programmer. If a programmer uses a Shortcut key in a menu that overrides what should work in an Edit that is their mistake. We shouldn't exclude valid keys like "F1" or "F5" or even "Ctrl-V" just because someone might do something dumb.

BDisp:
> I think this is over protecting the programmer. If a programmer uses a Shortcut key in a menu that overrides what should work in an Edit that is their mistake. We shouldn't exclude valid keys like "F1" or "F5" or even "Ctrl-V" just because someone might do something dumb.

For "Ctrl-V" is not difficult, just to get rid the AltMask check, but for "F1" or "F5" the menu must be refactored to allow that. Maybe it could be doing later as soon it's possible.

BDisp:
In my opinion, I would change the keys from `ControlA to Control Z` into `A to Z`, because it's still necessary to compare to the `Key.CtrlMask` if it just has the ctrl modifier, having a duplicate assigment.
Instead of:
`Key.CtrlMask | Key.ControlA`

I think it would more gracefull like:
`Key.CtrlMask | Key.A`

@migueldeicaza and @tig , if you prefer this change let me know as this would have a huge compatibility impact.

BDisp:
@tig for test the `F1` and `F2` keys in the `WSL 2` you must install in `Windows` a software like [VcXsrv X Server](https://sourceforge.net/projects/vcxsrv/).

tig:
I know you are still trying to rebase this. Perhaps it would be easier to just do a new PR with your Menu specific changes. It's clear this PR has gotten unwieldy with a mix of changes (e.g. all the `event` stuff?

BDisp:
> 
> 
> I know you are still trying to rebase this. Perhaps it would be easier to just do a new PR with your Menu specific changes. It's clear this PR has gotten unwieldy with a mix of changes (e.g. all the `event` stuff?

But without them the menu changes won't work properly. I really having rebase issues and if can't solve it I'll submit a new PR. Thanks.

tig (changes requested):
I need you to back this change out. We don't want to bump to 0.90 yet.

Directory.Build.props
```diff
@@ -1,9 +1,9 @@
 <Project>
 
   <PropertyGroup>
-    <Version>0.89.4</Version>
-    <AssemblyVersion>0.89.4</AssemblyVersion>
-    <FileVersion>0.89.4</FileVersion>    
+    <Version>0.90.0</Version>
```
tig, line 4: I need you to back this change out. We don't want to bump to 0.90 yet.

tig (changes requested):

Directory.Build.props
BDisp: No problem. Do you'll submit a PR to this change and I'll rebase or do you want I made the change?

tig: Just modify this PR to use 0.89.4.

BDisp: I'm finishing some more fixes and improvements and I'll committing when everything is ready. Thanks.

tig (commented):
This appears to be based on a ton of work!

There is a lot in here and I can't tell if it all relates to the menu short cut issue. Does it?

How does all this work? Why does it need to be so complex?

I am worried about accepting such a big, complex PR, just as we're trying to close down the 0.90 release.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -419,8 +419,22 @@ void ProcessInput (Action<KeyEvent> keyHandler, Action<KeyEvent> keyDownHandler,
 					mouseHandler (ToDriverMouse (ev));
 					return;
 				}
-				keyHandler (new KeyEvent (MapCursesKey (wch), keyModifiers));
-				keyUpHandler (new KeyEvent (MapCursesKey (wch), keyModifiers));
+				k = MapCursesKey (wch);
```
tig, line 422: What does all this do? It seems very fragile, just using hard coded numbers.

BDisp: @tig most of the code was changed because the shortcut keys compatibility as much keystrokes has not working. That code is for the functions keys combinations (F1-F12). In the middle I needed to fixes some bugs to test all the necessary code.

BDisp:
There was some problems in Linux because wasn't configure to work with NET5.0. I think the PR is acceptable although it need some more work. In my opinion we should include this to the 0.90 version. Otherwise it no worth lost many time with this.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
BDisp: > 
> 
> What does all this do? It seems very fragile, just using hard coded numbers.

As you know the wch variable is always mapped and converted to `Key `enum before they are passed to a event handler. I only assigned it before if no modifiers keys are necessary to shifted.

tig:
@BDisp I know you've put a lot of work into this PR. I really appreciate that.

Given how close we are to shipping 1.0 and the complexity of this PR, as well as how it is not a core scenario, I think we should wait until after 1.0 to include it.

I'm just not comfortable with all the changes, all the way down to the drivers, at this point. I need to carefully review it all, and there is just too much new code, with a lot of complexity given the time I have right now. I'd much rather spend my limited time focusing on the critical bugs that prevent us from shipping 1.0.

I suggest that instead of this PR, we address the but reported user about `Shortcut` by simply removing the `Shortcut` member from `Menu` for now. I've done this with #909, which is a 1 line change.

Issue #634, opened 2020-06-06
Should the Action be flagged with the event word?


BDisp:
According to this useful comment from @migueldeicaza (https://github.com/migueldeicaza/gui.cs/pull/621#issuecomment-639556768), `Action `doesn't really need to be flagged as an `event`. The only difference is that it appears with a different icon in the dropdown window according to the images. I personally would prefer to have the spark icon, so I ask you to give your opinion too. If you prefer the spark icon, I reset the last push made. Thank you.

![field](https://user-images.githubusercontent.com/13117724/83953326-9124d800-a837-11ea-86e2-b638f17396ba.png)
![spark](https://user-images.githubusercontent.com/13117724/83953327-91bd6e80-a837-11ea-87c8-f3e4fab0fa56.png)

BDisp:
I will close this issue because regardless of the spark icon being more appealing, the event flag prevents an `Action `from being declared inside the construction of a class, such as this:

```
			_xRadioGroup = new RadioGroup (radioItems) {
				X = 0,
				Y = Pos.Bottom (label),
				SelectionChanged = (selected) => DimPosChanged (_curView),
			};

```

tig:
I wish I understood why VS does that. Annoying.

BDisp:
I had already come across that sometimes I was unable to declare events in the construction of the class and after all it was because of this :-)

BDisp:
It only allows for fields.

PR #903, opened 2020-09-22
[Breaking] Use C# Events Instead of Just Actions

worldbeater:
**What kind of change does this PR introduce?**

This PR introduces a massive breaking change, it converts every `public Action<TEventArgs> EventName;` into a C# event `public event Action<TEventArgs> EventName;` Following the [discussion](https://twitter.com/worldbeaterdev/status/1308485191942451206) with @migueldeicaza on Twitter, there are a couple of reasons for doing this.

The first reason is, that an IDE like Microsoft Visual Studio or JetBrains Rider **highlights** `Action`s declared with the `event` keyword with ⚡, making the events easier to discover among the suggested methods, fields, and properties. All events in Xamarin Forms, WPF, Windows Forms, and [Avalonia](https://github.com/AvaloniaUI/Avalonia/blob/75c3707d5d9ba56a356ac3924426dcf478549052/src/Avalonia.Input/IInputElement.cs#L16) are declared with the `event` keyword, and this makes the process of interacting with them a tiny bit easier. Having events highlighted with ⚡ in the IDE is cool stuff.

ezgif-7-7ac02673d0dd

Another reason is, that tools like [Pharmacist](https://github.com/reactiveui/pharmacist) that do code generation are processing **only** those `Action`s that are defined as `event`s. [Pharmacist](https://github.com/reactiveui/pharmacist) is a tool that turns all events in a given assembly into `IObservable` wrappers, and this improves the dev experience for those developers, who decide to use `Terminal.Gui` with [reactive extensions](https://github.com/dotnet/reactive), and with the [ReactiveUI](https://github.com/reactiveui/reactiveui) MVVM framework. In short, Pharmacist allows us to write the following code:

```cs
// Reactive way of processing streams of events.
var item = new CheckBox();
var subscription =
    item.Events()
        .Toggled
        .Where(toggled => toggled)
        .Subscribe(_ => Console.WriteLine("Toggled!"))
        .DisposeWith(cleanup);

// Unsubscribe from the notifications.
subscription.Dispose();
```

Instead of the following code:

```cs
// Traidional way of implementing the pattern shown above.
var item = new CheckBox();

void OnToggled(bool toggled) {
    if (toggled) {
        Console.WriteLine("Toggled!");
    }
}

item.Toggled += OnToggled;

// Unsubscribe from the notifications.
item.Toggled -= OnToggled;
```

It may seem like a matter of taste, but reactive extensions are [really powerful](https://youtu.be/DYEbUF4xs1Q), and there are a lot of tasty extension methods. There is also a [demo view](https://github.com/worldbeater/ReactiveMvvm/blob/main/src/ReactiveMvvm.Terminal/Views/FeedbackView.cs) that shows how to combine MVVM, ReactiveUI, and `Terminal.Gui`. 

**What is the current behavior?**

Currently, most events are defined as follows:

```cs
public Action<SelectedItemChangedArgs> SelectedItemChanged;
```

**What is the new behavior?**

Now, most events are defined as follows:

```cs
public event Action<SelectedItemChangedArgs> SelectedItemChanged;
```

**What might this PR break?**

If folks are using the following syntax:

```cs
scrollView.DrawContent = args => {
    // Do stuff
};
```

Then they'll definitely break. The solution would be to change the code block above to:

```cs
scrollView.DrawContent += args => {
    // Do stuff
};
```

Thanks in advance for reviewing this.

### DIFF
--- a/Example/demo.cs
+++ b/Example/demo.cs
@@ -141,7 +141,7 @@ static class Demo {
 #else
 		var filler = new Filler (new Rect (0, 0, 40, 40));
 		scrollView.Add (filler);
-		scrollView.DrawContent = (r) => {
+		scrollView.DrawContent += (r) => {
 			scrollView.ContentSize = filler.GetContentSize ();
 		};
 #endif
--- a/Example/demo.cs
+++ b/Example/demo.cs
@@ -225,10 +225,11 @@ static class Demo {
 
 	static void NewFile ()
 	{
-		var d = new Dialog (
-			"New File", 50, 20,
-			new Button ("Ok", is_default: true) { Clicked = () => { Application.RequestStop (); } },
-			new Button ("Cancel") { Clicked = () => { Application.RequestStop (); } });
+		var ok = new Button ("Ok", is_default: true);
+		ok.Clicked += () => { Application.RequestStop (); };
+		var cancel = new Button ("Cancel");
+		cancel.Clicked += () => { Application.RequestStop (); };
+		var d = new Dialog ("New File", 50, 20, ok, cancel);
 		ml2 = new Label (1, 1, "Mouse Debug Line");
 		d.Add (ml2);
 		Application.Run (d);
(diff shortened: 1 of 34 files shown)


BDisp:
I've tried do that before but the event keyword prevents from the event being using inside a new class declaration as stated here #634.

tig:
How bad would it be, really, if programmers weren't able to wire up event handlers in object initializers?

My inclination is having `event` and work everywhere else, as eloquently argued by @worldbeater above trumps it not working in object initializers.

BDisp:
I also prefer the `event `flag for `Action `too and I approve this PR.

BDisp (approved):

worldbeater:
Thanks for reviewing this @tig @BDisp! I guess we could declare the actions as optional parameters e.g. `(Action<SelectedItemChangedArgs> selectedItemChanged = null, ...)` and inject them via a constructor. This way it should be even easier for folks to migrate their applications, as they'd need to change their initialization logic only a tiny bit. However, not sure if we really need this. Semantically it may be not as obvious that we need to unsubscribe the delegate passed via the constructor.

tig:
I'm merging this PR as-is with the expectation that we will find issues with it. I think that's ok as long as everyone commits to testing all of their scenarios thoroughly.

PR #920, opened 2020-09-29
Fixes #919 Initialized and CanFocus.

### LINKED ISSUE #919: A SubView must not change the SuperView's CanFocus to true if it is false.
- Of course, when SubViews are added to a SuperView, the CanFocus can be changed to true if the SubView is true. After starting the application if a SuperView has many SubViews and we intend to prevent the focus on all of them, the more practical and faster way is to prevent the focus on SuperView than to individually change all SubViews. If the CanFocus from SuperView is false and the developer inadvertently changing CanFocus to true from a SubView will also change CanFocus to true from SuperView and all SubViews, achieving unexpected behavior. Therefore, only SuperView should change its own state of CanFocus after its initialization.

- For this to be processed along these lines, I noticed that the IsInitialized property should only be set after its complete initialization, that is, in the EndInit method after the initialization of all its SubViews.

BDisp:


tig (commented):
looks good to me!

BDisp:
> 
> 
> looks good to me!

I forgot to remove the new functionality that I am still working on.

tig:
please rebase

PR #930, opened 2020-09-30
Fixes #929 Sometimes ListView throw a exception on the SelectedItem.

### LINKED ISSUE #929: Sometimes ListView throw a exception on the SelectedItem on assignment.
- It's due a bug in `if (selected < 0 || selected >= source.Count)`.
- Other issue is the `OnSelectedChanged `event does not fired when the source becomes null and the selected becomes 0 and we need that event on the `MoveDown `method.
- The `OnEnter `method sometimes does not propagate to the next views who need him. Changing the assigment from `return false` to `return base.OnEnter (view)` made the diference.

BDisp:


tig (commented):
Nice catch.

PR #937, opened 2020-09-30
Add Reactive Extensions & ReactiveUI Sample App

worldbeater:
This PR adds a new project that shows how to use [reactive extensions](https://github.com/dotnet/reactive) and [ReactiveUI](http://reactiveui.net/) with `gui.cs`. After merging https://github.com/migueldeicaza/gui.cs/pull/903, we can now use the full power of [Pharmacist](https://github.com/reactiveui/pharmacist) and generate `IObservable<TEventArgs>` wrappers for every event in the `Terminal.Gui` assembly. This allows us to easily implement the MVVM (model view view-model) pattern and data bindings between views and view models. The new demo app looks like this:


ReactiveExample/ReactiveExample.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageReference Include="Pharmacist.MsBuild" Version="1.8.1" PrivateAssets="all" />
+        <PackageReference Include="Pharmacist.Common" Version="1.8.1" />
+        <PackageReference Include="Terminal.Gui" Version="0.90.3" />
```
tig, line 9: Since this will be part of the Terminal.Gui project, shouldn't this ref the csproj file?

tig (commented):
I'm curious: Is there a unit test that would ensure we never break the ability for Reactive Extensions to work?

If so, would you please consider adding one to UnitTests?

tig (changes requested):
Please add a note about this to the README.md and to the relnotes in the Terminal.Gui.csproj file.

ReactiveExample/ReactiveExample.csproj
BDisp: That way it stay limited do to the 0.90.3 version. It's bad. Good catch.

worldbeater: The issue here is, that [Pharmacist.MSBuild](https://github.com/reactiveui/pharmacist#msbuild) can generate `.Events()` only for `.dll` assemblies and NuGet packages. So if we include `Terminal.Gui` via a `ProjectReference`, the tool won't work. Currently, Pharmacist is being rewritten [to use the new source generators API](https://github.com/reactivemarbles/ObservableEventsSourceGenerator), so hopefully, a new version will be out soon, that won't impose such limitations because the new source generators can work with any source file. We could declare the version attribute as `Version="*"` but this way it has a chance to break the CI build after a new NuGet package version is out, so decided to pin the version temporarily.

Another way of resolving this is to prepare a separate NuGet package e.g. `Terminal.Gui.Events` that will contain the generated observable wrappers (we are [using this approach](https://github.com/AvaloniaUI/Avalonia/blob/master/nukebuild/Build.cs#L188) in Avalonia 0.10). But after the new source generators-based Pharmacist is out, that package will become no longer required, so not sure if publishing a new NuGet makes sense now...

worldbeater:
Stumbled upon an issue related to reactive scheduling, will need some time to implement a custom scheduler.

worldbeater:
Resolved the issues with scheduling by implementing a `TerminalScheduler` that can be used as the main thread scheduler in reactive `Terminal.Gui` apps, also added some getting started documentation. Regarding that unit tests — that reactive stuff won't break until we are marking our `Action`s with the `event` keyword. Assuming [Pharmacist](https://github.com/reactiveui/pharmacist) doesn't currently support `<ProjectReference />` tags, we could wait until [the new source generators-based Pharmacist version](https://github.com/reactivemarbles/ObservableEventsSourceGenerator) is out, and add the tests then. Thoughts?

ReactiveExample/ReactiveExample.csproj
tig: Got it. 

How about this: We can now easily generate pre-release nuget packages for Terminal.Gui by simply updating the version info with a `-pre` indicator via `Terminal.Gui/Directory.build.props`. Could you build a pre-build tool (using `TT`?) to suck the `<version>xxx</version>` string from `Terminal.Gui/Directory.Build.props`?

Oh, that won't work... because the nuget package isn't published and available for several minutes after a push to master.

Hmmm.... 

I guess for now we'll just have to live with the risk that your sample will pull and older nuget package. Coordination will be needed. But it's not a big deal. 

For now, do update to use the `1.0.0-pre.1` nuget package.

tig:
> that reactive stuff won't break until we are marking our `Action`s with the `event` keyword. 

This is in now. I merged your PR on this.

tig:
> we could wait until [the new source generators-based Pharmacist version](https://github.com/reactivemarbles/ObservableEventsSourceGenerator) is out, and add the tests then. Thoughts?

I'm ok with that but am curious what the tests would test for.

ReactiveExample/TerminalScheduler.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Reactive.Concurrency;
+using System.Reactive.Disposables;
+using Terminal.Gui;
+
+namespace ReactiveExample {
+	public class TerminalScheduler : LocalScheduler {
```
tig, line 7: I think you are suggesting this could be built-in to `Terminal.Gui`? How would existing code (e.g. the Progress Bar scenario) be updated to use this?

worldbeater: If we add this scheduler to `Terminal.Gui`, the change will introduce a dependency on `System.Reactive`, and this may result in folks complaining about extra `.dll` files they don't use etc. Probably worth creating a separate package named `Terminal.Gui.ReactiveUI` for this

ReactiveExample/ReactiveExample.csproj
worldbeater: On it. Will update the version soon.

ReactiveExample/TerminalScheduler.cs
tig: That makes tons of sense. Looking forward to that PR.

ReactiveExample/LoginViewModel.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Reactive.Linq;
+using System.Runtime.Serialization;
+using System.Threading.Tasks;
+using NStack;
+using ReactiveUI;
+using ReactiveUI.Fody.Helpers;
+
+namespace ReactiveExample {
+	//
+	// This view model can be easily shared across different UI frameworks.
```
tig, line 12: This gives me warm feels.

Not for this PR, but a pointer to a sample or real-world app that does this would be awesome (I mean, *I* really would love to have one because I'm currently building a XAML app that I'd love to port to Terminal.Gui).

tig (approved):

BDisp (approved):

ReactiveExample/LoginViewModel.cs
worldbeater: We have a few samples listed on this page https://www.reactiveui.net/docs/resources/samples Also, here is a tiny and cross-platform ReactiveUI sample app with a lot of frontends, including Terminal.Gui as well https://github.com/worldbeater/reactivemvvm

tig:
@worldbeater it would be awesome if you tweeted about this!

Issue #944, opened 2020-10-02
StatusBar - SHIFT-CTRL-A shortcut not possible


tig:
`new StatusItem(Key.ShiftMask | Key.CharMask | (Key)65, "~SHIFT-^A~ Select None", () => ...`

This does not work. 

Neither does `Key.ShiftMask | Key.ControlA` which is what I tried first,

Am I missing something or is this just broke?

BDisp:
These combinations never worked all together before. I almost have all that ready but it will get some days to finish.

tig:
> These combinations never worked all together before. I almost have all that ready but it will get some days to finish.

Sweet. Here's the use case: https://github.com/PowerShell/GraphicalTools/issues/120

BDisp:
>On Windows/Linux ^A would be CTRL-A. On Mac it would be Command-A

Well the solution I'll going to submit will return a string like Ctrl+Alt+1 or whatever is choosed but if is the case of print the Ctrl with another name the developer will be deal with that.

migueldeicaza:
While it might be possible to support this on Windows, it will not work on Unix, as there is no way of encoding "control" for uppercase/lowercase, there are only 26 or so control values available, and the kernel does not distinguish those, nor do terminal emulators.

tig:
Won't fix.

Issue #928, opened 2020-09-30
Where the ShortCut should be reside?


BDisp:
I will talk about three places where it will could belong according to my order of preference:

1. On the `TextFormatter `and in the `View`.
2. On the `View `only.
3. On the `Menu `(This going limit the use case of shortcuts than can be used on multiple views)

Please give me suggestions.

BDisp:
On the `TextFormatter `there are no sense, sorry.

tig:
Related: https://github.com/migueldeicaza/gui.cs/issues/944

Whatever you decide, we need to make sure this works. I'm not sure I have a strong opinion yet.

BDisp:
I can guarantee to you it will work better than before, although in unix some combinations won't work at all, mainly Ctrl+Shift and some functions keystroke.

tig:
> I can guarantee to you it will work better than before, although in unix some combinations won't work at all, mainly Ctrl+Shift and some functions keystroke.

Can you explain the technical reason Ctrl+Shift won't work via Curses? I'm curious to know.

BDisp:
> Can you explain the technical reason Ctrl+Shift won't work via Curses? I'm curious to know.

Because the value is always the same no matter the shift key is pressed or not.

tig:
> > Can you explain the technical reason Ctrl+Shift won't work via Curses? I'm curious to know.
> 
> Because the value is always the same no matter the shift key is pressed or not.

That's not a very technical description. What value? 

I can't tell if this is a limitation of Curses or an underlying Linux/mac issue. For example, is it possible that building a Linux/Mac `ConsoleDriver` with some other layer OTHER than Curses, could address this? If so, what are our options. 

Note, if you don't know that's ok. But we need to find someone who does because much like mouse not working in Windows Terminal, Terminal.Gui looks pretty silly if fundamental things like chorded-key-presses don't work on Linux/Mac.

BDisp:
There may be new versions of ncurses that can better handle keyboard and mouse processing. Anyway, I can already process many key combinations in Linux. What is really bothering me is the fact that the mouse movement is no longer being detected in linux. That is why I suspect that there may be new features of libncurses that need to be explored but I confess that I am not a specialist in linux or mac. I can promise to try to search for solutions and find patience for it.

BDisp:
> That's not a very technical description. What value?

[...]

PR #946, opened 2020-10-03
Fixes #881, #928. Allowing more key combinations.

### LINKED ISSUE #881: Modifier key are not fully supported with CursesDriver
It seems modifier key are not fully supported with CursesDriver.

Some examples: 
- `CTRL+1` not supported.
- `CTRL+SHIFT+S` not supported.
- `SHIFT+F5`. not supported.
- `CTRL+S` supported (with `ControlS` key).

However, most of these are supported with the WindowsDriver, except `CTRL+1`.

BDisp:
I decided submit as is for testing.

To personalize a delimiter shortcut separator the `MenuBar.ShortCutDelimiter` must be set before creating the menu.


tig (commented):
Great work!

I don't see any tests in here that prove #944 is fixed.

BDisp:
As I said before the shift+ctrl only will work on WindowsDriver and in this case it is needed to do a check in the SO to use another combination with Unix.

tig:
> As I said before the shift+ctrl only will work on WindowsDriver and in this case it is needed to do a check in the SO to use another combination with Unix.

Then you should change the title to not say it fixes #944.

BDisp:
> Then you should change the title to not say it fixes #944.

@tig you are right, sorry. Done.

BDisp:
To personalize a delimiter shortcut separator the `MenuBar.ShortCutDelimiter` must be set before creating the menu.

tig:
I am not sure what you want to do with this PR. It appears to still be a work in progress.

- It needs to be rebased
- It seems to include unrelated fixes that would ideally be in separate PRs (but it's hard for me to tell because of rebase issues).
- You have not addressed my comments/concerns.

BDisp:
> 
> 
> I am not sure what you want to do with this PR. It appears to still be a work in progress.
> 
Even it's in progress I think better updating it for someone who has interest to testing and detecting failures.


>     * It needs to be rebased
> 
This is what I did now.


>     * It seems to include unrelated fixes that would ideally be in separate PRs (but it's hard for me to tell because of rebase issues).
> 
These are changes that was needed to adapt for this PR go up. Without this PR these changes couldn't be submitted.


>     * You have not addressed my comments/concerns.
I think I already answered. If something escaped me, remind me again, please. Thanks.

.gitignore
```diff
@@ -11,4 +11,8 @@ packages
 docfx/api
 
 #git merge files
-*.orig
\ No newline at end of file
+*.orig
```
tig, line 14: Why is this change in this PR? This is already in master. You need to rebase.

README.md
```diff
@@ -3,21 +3,20 @@
 [![Version](https://img.shields.io/nuget/v/Terminal.Gui.svg)](https://www.nuget.org/packages/Terminal.Gui)
```
tig, line 3: Why is this change in this PR? This is already in master. You need to rebase.

ReactiveExample/FodyWeavers.xml
```diff
@@ -0,0 +1,3 @@
+﻿<Weavers xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="FodyWeavers.xsd">
```
tig, line 1: This was already accepted in e5b0795a.

ReactiveExample/FodyWeavers.xsd
```diff
@@ -0,0 +1,26 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
tig, line 1: Same.

tig:
You did not address any of my review comments on using integer literals. I do not think we should accept code that uses so many literals. 

I see changes in here that are unrelated to either bug, e.g. all of the Reactive sample stuff, the changes to README.md, .gitignore, etc..  Those were all part of another PR. I've commented some more on a few of them.

BDisp:
Do you want I create constant variables to accommodate the integer literals? About Reactive I'm not familiar with it yet.
I deleted my other PR to remake it again. So this one PR is not linked to the original anymore. If you think that this PR has no value to the GUI repo just say it and I will not spend no more time on it.

tig (commented):
> Do you want I create constant variables to accommodate the integer literals? 

I think that would be better than just including all those un-decipherable literals. It is just good coding practice. Do you not agree?

> About Reactive I'm not familiar with it yet.

I was commenting on Reactive because this PR includes files from Reactive sample. I believe this is simply because you did not rebase/merge correctly.

> I deleted my other PR to remake it again. So this one PR is not linked to the original anymore. If you think that this PR has no value to the GUI repo just say it and I will not spend no more time on it.

I think what you are doing here is REALLY important and is good work. I just think it can be further improved. I am not intending to be overly critical, but I do feel strongly that the way it is currently coded is too complex for others to be able to read and understand. This is partially because of the use of all the int literals but also due to the nested if statements. 

When I encounter a problem where I've had to create a lot of complex if/then/elses I try to step back from the problem and try to find a more elegant solution. There is almost always a simpler way of solving such problems using a different approach or algorithm. I would like to be able to give you a precise suggestion, but because I am having so much trouble following the code, I don't really understand how you are trying to solve the problem (or actually what the problem is).

In one comment here I wrote:

> In otherwords, this whole PR changes how Key is interpreted by really using shift/ctrl/alt-masks. Go all the way and get rid of the original (hack, in my opinion) of defining Key.Control<char> as special values. Get rid of that legacy!

It may be that one of the challenges here is you are trying to "hack" around a design error Miguel made in gui.cs. He wrote all this starting with `curses` and Linux in mind. It's what he knew. 

He did not really care about how Windows works. You and I (and a lot of users) do care more about that now. I feel strongly that gui.cs apps, when running on Windows should be able to take full advantage of the keyboard, even if Linux users can't get the same.

In the end, the current design of `ConsoleDriver` relative to keyboard input is based on a 'lowest common denominator` approach where Linux/curses is the base. It was never designed to deal with all the shift key modifiers the way other platforms (e.g. Windows) do. Thus some of the design decisions he took early in are invalid and can be revisited. 

I'm specifically talking about the whole design of `Event.cs`, the `Key` enum, and how the special characters and shift modifiers are encoded. 

If you are going to invest them time in fixing things so shift modifiers work better across platforms, my specific suggestion is to take a look at how `System.Windows.Forms` dealt with it. Then re-design the `ConsoleDriver` interface to have an architecture that intrinsically supports shift modifiers fully. THEN change the Curses driver to implement the appropriate subset of that. 

Am I making sense to you? I'm happy to discuss this further and debate options!

BDisp:
Yes it make sense. This constants is only related with CursesDriver because the lookup of theses keys is different from Windows and there are no sense to integrate in the Keys Enum. Maybe in the constants.cs. The suggestion you made for the Keys Enum is a great chalange but I'm afraid it could break some code because the GUI use as possible a sequential values. I tried before to assign a value to the key F1 as 112 and I experience some behavior issues.
About the rebase I only did a rebase and syncronized to origin. Maybe I should have just done a merge? What you suggest?

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -445,19 +460,65 @@ void ProcessInput (Action<KeyEvent> keyHandler, Action<KeyEvent> keyDownHandler,
...
+					} else if (wch2 == 91) {
+						int [] c = null;
+						while (code == 0) {
+							code = Curses.get_wch (out wch2);
+							if (wch2 > 0) {
+								Array.Resize (ref c, c == null ? 1 : c.Length + 1);
+								c [c.Length - 1] = wch2;
+							}
+						}
+						if (c [0] == 49 && c [1] == 59 && c [2] == 55) { // Ctrl+Alt+(F1 - F4)
```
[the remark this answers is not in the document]
BDisp: Yes we can add some more keys to the Key Enum. I would prefer that too.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -419,8 +419,23 @@ void ProcessInput (Action<KeyEvent> keyHandler, Action<KeyEvent> keyDownHandler,
 					mouseHandler (ToDriverMouse (ev));
 					return;
 				}
-				keyHandler (new KeyEvent (MapCursesKey (wch), keyModifiers));
-				keyUpHandler (new KeyEvent (MapCursesKey (wch), keyModifiers));
+				k = MapCursesKey (wch);
+				if (wch >= 277 && wch <= 288) { // Shift+(F1 - F12)
```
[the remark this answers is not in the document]
BDisp: Well the `wch `sometimes cames with other escapes sequences, sometimes with 4 or 5. I got to do `get_wch` to read all this sequences and decifre the wright key. See the` int [] c` variable. This array doesn't have nothing to do with `Key Enum`. It's result of all my investigate and reserche and testing. So this why they are literal :-)
It's constains all the keys using `Shift+F1 to Shift+F12`.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -419,8 +419,23 @@ void ProcessInput (Action<KeyEvent> keyHandler, Action<KeyEvent> keyDownHandler,
...
+					wch -= 12;
```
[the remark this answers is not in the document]
BDisp: Due to my research I could calculate a constant value to get the pretended key without much code.

Terminal.Gui/Views/TextField.cs
```diff
@@ -299,9 +299,9 @@ public override bool ProcessKey (KeyEvent kb)
 			// Needed for the Elmish Wrapper issue https://github.com/DieselMeister/Terminal.Gui.Elmish/issues/2
 			var oldCursorPos = point;
 
-			switch (kb.Key) {
+			switch (GetModifiersKey (kb)) {
 			case Key.DeleteChar:
-			case Key.ControlD:
+			case Key.ControlD | Key.CtrlMask:
```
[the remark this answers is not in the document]
BDisp: In this case I would suggest to change all the keys from `ControlA `to `ControlZ `into `A `to `Z `and in this case would be
 `case Key.D | Key.CtrlMask:`
What you think about this?

UICatalog/UICatalog.cs
```diff
@@ -154,11 +155,15 @@ private static Scenario GetScenarioToRun ()
 
 			_menu = new MenuBar (new MenuBarItem [] {
 				new MenuBarItem ("_File", new MenuItem [] {
-					new MenuItem ("_Quit", "", () => Application.RequestStop() )
+					new MenuItem ("_Quit", "", () => Application.RequestStop(), null, null, Key.ControlQ)
```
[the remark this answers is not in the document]
BDisp: In this case that doesn't happen because in the end it's translates to` Ctrl+'A'`, but the `Key.CtrlMask` must be assigned to inject into the key.

.gitignore
BDisp: Ups I made something wrong, sorry.

ReactiveExample/FodyWeavers.xml
BDisp: Same, sorry.

ReactiveExample/FodyWeavers.xsd
BDisp: I'll do now, sorry.

Terminal.Gui/Views/TextField.cs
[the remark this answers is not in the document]
BDisp: Another thing which I not agree at all it's the `ControlSpace = 0` where is comment  "`The key code for the user pressing Control-spacebar`". But in my opinion isn't true because space is 32 and not 0. If I press Ctrl+Space I have a CtrlMask+32. I think the intencion was to represent the 0 (NULL Char) and the `ControlSpace `should be changed to "`Null`". The `Unknown `have to stay as is because is returned if no key is found. Of course this will force to change the default value from `ControlSpace `to `Null `instead . What do think about this?

BDisp:
@tig I did a reset with keep changes and did a force push. Now I can't pop the changes. Oh damn!

UICatalog/UICatalog.cs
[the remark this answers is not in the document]
BDisp: Because of this I made the suggestion to change the `Key.Control<char>` to only the` Key.<char>`, e.g. from` ControlA = 1` to `A = 65` and change the verification key in the case of `Ctrl+A` change from:

```
} else if (wch2 - (int)Key.Space >= 'A' && wch2 - (int)Key.Space <= 'Z') {

} else if (wch2 >= (uint)Key.ControlA && wch2 <= (uint)Key.ControlZ) {
```

to:

`} else if (wch2 - (int)Key.Space >= (uint)Key.A && wch2 - (int)Key.Space <= Key.Z) {

} else if (wch2 >= (uint)Key.A - 64 && wch2 <= (uint)Key.Z - 64) {`

What do you think?

README.md
BDisp: Rebase issue, sorry.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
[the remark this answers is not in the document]
BDisp: Here I'm comparing with array elements which cannot be parsing as constants and it need to compared with if statement. At least it's working :-)

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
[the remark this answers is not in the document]
BDisp: `wch2 = c [1] == 53 ? c [1] + 216 : c [1] + 215;`

The 216 and 215 are also values it's need to be added to achieve the desired key. It's only specific for functions keys and only related to the ncurses.

BDisp:
> If you are going to invest them time in fixing things so shift modifiers work better across platforms, my specific suggestion is to take a look at how `System.Windows.Forms` dealt with it. Then re-design the `ConsoleDriver` interface to have an architecture that intrinsically supports shift modifiers fully. THEN change the Curses driver to implement the appropriate subset of that.

I like the idea very much but is not very easy. On `Windows `the `kernel `library has the necessary functions to deal with the modifiers keys but not in `Unix`. We may could use the `X11 `library but I'm not so expert to know if it will have some negative impact. In `Windows `I created the `private Key MapKeyModifiers (ConsoleKeyInfo keyInfo, Key key)` method which get the modifiers keys through the `ConsoleKeyInfo `struct and inject in the key. In `Unix `I created the `KeyModifiers MapKeyModifiers (Key key)` method which set the modifiers keys already pre inject in the keys. As you see they behave in opposite way from each other to achieve the same thing.
So, you can't have a `public abstract Key MapKeyModifiers (ConsoleKeyInfo keyInfo, Key key)` if the `Unix `can't provide a struct like `ConsoleKeyInfo`.

Terminal.Gui/Views/TextField.cs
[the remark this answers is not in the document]
tig: I always thought that `ControlsSpace = 0` looked wrong.

Terminal.Gui/Core/View.cs
```diff
@@ -170,6 +172,28 @@ internal enum Direction {
 		/// </summary>
 		public Rune HotKeySpecifier { get => textFormatter.HotKeySpecifier; set => textFormatter.HotKeySpecifier = value; }
 
+		/// <summary>
+		/// This is the global setting that can be used as a global shortcut to invoke an action if provided.
+		/// </summary>
+		public Key ShortCut {
```
tig, line 178: Minor nit, but  "shortcut" is a word. Thus this should be `Shortcut` not `ShortCut`.

Terminal.Gui/Core/View.cs
```diff
@@ -170,6 +172,28 @@ internal enum Direction {
...
+			get => shortCutHelper.ShortCut;
+			set {
+				if (shortCutHelper.ShortCut != value && (ShortCutHelper.PostShortCutValidation (value) || value == Key.Null)) {
+					shortCutHelper.ShortCut = value;
```
tig, line 182: Minor nit, but  "shortcut" is a word. Thus this should be `shortcutHelper` not `shortCutHelper`.

Terminal.Gui/Core/View.cs
```diff
@@ -170,6 +172,28 @@ internal enum Direction {
...
+				}
+			}
+		}
+
+		/// <summary>
+		/// The keystroke combination used in the <see cref="ShortCut"/> as string.
+		/// </summary>
+		public ustring ShortCutTag => ShortCutHelper.GetShortCutTag (shortCutHelper.ShortCut);
```
tig, line 190: `ShortcutTag`

Terminal.Gui/Core/View.cs
```diff
@@ -170,6 +172,28 @@ internal enum Direction {
...
+
+		/// <summary>
+		/// The action to run if the <see cref="ShortCut"/> is defined.
+		/// </summary>
+		public virtual Action ShortCutAction { get; set; }
```
tig, line 195: `ShortcutAction`

tig (changes requested):
can you fix the casing?

tig:
@BDisp did you see my request to rename `ShortCutHelper.cs` to `ShortcutHelper.cs`?

BDisp:
> 
> 
> @BDisp did you see my request to rename `ShortCutHelper.cs` to `ShortcutHelper.cs`?

Sorry I saw it and thought it had already changed. But I had to change it on the `Github `website because locally git did not recognize the change.

PR #956, opened 2020-10-17
Fixes #933. Updated to work with libncurses 6.2

### LINKED ISSUE #933: Mouse move coordinates not working in Ubuntu 20.04 LTS
The movement event is captured in ncurses but the event flags only return ID=-1. Anyone can confirm this please.

BDisp:
However, this fix does not work properly in WSL with Ubuntu 20.04 but it works perfectly with the graphical version of Ubuntu 20.04 as well as with the graphical version of Linux Mint 20 Ulyana. On the Mac if anyone can help it would be great.


BDisp:
It seems that the service is sometimes down.

`/home/runner/.nuget/packages/pharmacist.msbuild/1.8.1/buildTransitive/netstandard2.0/Pharmacist.MSBuild.targets(48,5): error MSB4018:  ---> System.Net.NetworkException: No connection could be made because the remote host actively refused it. [/home/runner/work/gui.cs/gui.cs/ReactiveExample/ReactiveExample.csproj]`

PR #960, opened 2020-10-19
Fixes #959. CursorPosition with hotkeys.

### LINKED ISSUE #959: A computed layout view should not allowing absolute assignment.
- In UICatalog open Buttons and click on one of both buttons at the bottom. An exception will be firing because those buttons have Width anchored with type DimCombine. So, generics controls should not change that behavior unless in user control specific.
- I also fixed the PositionCursor which should be positioned over the hotkey.
- The radio button is not rendering the initial selection. So, I created a Refresh method that call OnSelectionChanged.

BDisp:


Terminal.Gui/Views/Button.cs
```diff
@@ -145,7 +145,13 @@ internal void Update ()
 				base.Text = ustring.Make (_leftBracket) + " " + text + " " + ustring.Make (_rightBracket);
 
 			int w = base.Text.RuneCount - (base.Text.Contains (HotKeySpecifier) ? 1 : 0);
-			Width = w;
+			try {
+				Width = w;
+#pragma warning disable RCS1075 // Avoid empty catch clause that catches System.Exception.
```
tig, line 150: I don't think DimPos should be throwing an exception in this case. There should be no issue with assigning to `Width` here. I think the right fix is to back out the code that was added to `Width` in Commit 5efa8d53 to throw this exception, or modify it to not throw in this case.

tig (changes requested):

Terminal.Gui/Core/TextFormatter.cs
```diff
@@ -102,6 +102,11 @@ public class TextFormatter {
 		/// </summary>HotKeyTagMask
 		public uint HotKeyTagMask { get; set; } = 0x100000;
 
+		/// <summary>
+		/// Gets the position cursor from <see cref="HotKey"/>. If the <see cref="HotKey"/> is defined, the cursor will be positioned over it.
+		/// </summary>
+		public int PositionCursor { get; set; }
```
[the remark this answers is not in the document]
BDisp: You are right. I'll changing it.

Terminal.Gui/Views/Button.cs
BDisp: This is debatable. Imagine that you have a button anchored in a position that depends on other views but you need to change its caption from time to time. This will stop the button from being anchored and force the user to reconfigure the anchorage again. If it is a custom control the user will have to solve this but with native controls that are part of the `Terminal.Gui` I do not know if this should be avoided so as not to cause future annoyances for the user to wonder why my button is no longer anchored for just changing the caption?

tig: I see your point now. I think.

I need to ponder this some more.

Terminal.Gui/Views/Button.cs
```diff
@@ -145,7 +145,12 @@ internal void Update ()
 				base.Text = ustring.Make (_leftBracket) + " " + text + " " + ustring.Make (_rightBracket);
 
 			int w = base.Text.RuneCount - (base.Text.Contains (HotKeySpecifier) ? 1 : 0);
-			Width = w;
+			try {
```
tig, line 148: Could this be implemented with an `if "DimCombine"` instead of a `try/catch`? Would this require us to add an API to PosDim to be able to test it ahead of time?

BDisp: I made a more elegant solution. Otherwise you get the button undocked and it gets ugly.

BDisp: > 
> 
> Could this be implemented with an `if "DimCombine"` instead of a `try/catch`? Would this require us to add an API to PosDim to be able to test it ahead of time?

For that we will have to make the class public. Can be?

tig: To be clear, I feel pretty strongly that using `try/catch` for anything other than dealing with actual errors is a very bad practice. Can we modify `PosDim` to provide an API such that an `if` statement could be used here?

tig: > > Could this be implemented with an `if "DimCombine"` instead of a `try/catch`? Would this require us to add an API to PosDim to be able to test it ahead of time?
> 
> For that we will have to make the class public. Can be?

To me that would be far preferable to a hack with `try/catch`. Which class are we talking about here?

BDisp: Yes, it is possible. I also preferred this to using `try / catch`. I'll take care of that.

BDisp: This one:

`class DimCombine : Dim {`

BDisp: Done!

Terminal.Gui/Core/PosDim.cs
```diff
@@ -499,7 +499,7 @@ public static Dim Sized (int n)
 			return new DimAbsolute (n);
 		}
 
-		class DimCombine : Dim {
+		public class DimCombine : Dim {
```
tig, line 502: If it's public it needs documentation.

tig (changes requested):
Please add docs to DimCombine

Terminal.Gui/Core/PosDim.cs
BDisp: Sorry. There no need to be public. Changing to `internal `like the `DimAbsolute `is enough. Thanks.

BDisp:
> 
> 
> Please add docs to DimCombine

Sorry. There no need to be public. Changing to `internal `like the `DimAbsolute `is enough. Thanks.

tig (approved):
Well done.

PR #969, opened 2020-10-22
Fixes #968. Button Dim handler and a Dim unit test. Added a ToString to DimCombine.

### LINKED ISSUE #968: It may be more complex, but having more control over Pos / Dim is the best.
A control may or may not impose specific positions and dimensions, but it must always perfectly manipulate its intention and prevent the user from witnessing strange behaviors in anchoring a control and then, by mistake, assign an absolute value and only after much debugging does it discover the bug. Therefore, I advocate the launch of an exception that will immediately alert you to this bug. Of course, for this to be possible, control has to be created in order to avoid it. At the moment, only the `Button `control handles the control's dimension properly. So that you can better understand how this works, I will also submit a unit test to demonstrate what I said. The best way to check that `Pos / Dim` works perfectly is by comparing it to the `Frame `property of `View`.

BDisp:


BDisp:
@tig I don't have any local error. Do you know what is happen?

UnitTests/DimTests.cs
```diff
@@ -317,6 +317,184 @@ public void Dim_Validation_Do_Not_Throws_If_NewValue_Is_DimAbsolute_And_OldValue
...
+				Assert.Equal (47, v1.Frame.Width); // 49-2=47
+				Assert.Equal (89, v1.Frame.Height); // 98-5-2-2=89
+
+
+				Assert.Equal ("Dim.Combine(DimView(side=Width, target=FrameView()({X=49,Y=0,Width=49,Height=5}))-Dim.Absolute(2))", v2.Width.ToString ());
+				Assert.Equal ("Dim.Combine(Dim.Fill(margin=0)-Dim.Absolute(2))", v2.Height.ToString ());
+				Assert.Equal (47, v2.Frame.Width); // 49-2=47
+				Assert.Equal (89, v2.Frame.Height); // 98-5-2-2=89
+
+				Assert.Equal ("Dim.Factor(factor=0,1, remaining=False)", v3.Width.ToString ());
```
tig, line 414: You must specify a format specifier here to ensure that a period `.` is used for decimals. For the tests to run reliably in both local environments and in CI/CD the way #s are formatted has to be explicit. You used `,` instead of `.`.

See Line 170.

`Assert.Equal ($"Dim.Factor(factor={f / 100:0.###}, remaining={false})", dim.ToString ());`

BDisp: I should get that error on my pc. That bad. I will fix that. Thanks

BDisp: This change throws exceptions in all cultures and not just in github.

Terminal.Gui/Core/PosDim.cs
```diff
@@ -388,7 +391,8 @@ public bool IsFromRemaining ()
 
 			public override string ToString ()
 			{
-				return $"Dim.Factor(factor={factor}, remaining={remaining})";
+				var decChar = CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator;
```
tig, line 394: Sorry, this isn't right either. We can't have ToString be hard-coded to ignore user culture info. I know this works, but it's not right.

There has got to be a way to make the TESTS set the culture for the test run, so this code is not necessary. 

Like this: https://stackoverflow.com/questions/13354211/how-to-set-default-culture-info-for-entire-c-sharp-application

tig (changes requested):

Terminal.Gui/Core/PosDim.cs
BDisp: Thank you. That's what I wanted, although the ideal would be for the whole project.

tig (approved):

PR #970, opened 2020-10-22
Fix F# example

Martin4ndersen:
Fix F# example so that it builds.

- Fix "The type 'string' is not compatible with the type 'ustring'"
- Fix "No accessible member or object constructor named 'Button' takes"
- Fix "The member or object constructor 'MenuItem' does not take X argument(s). An overload was found taking 0 arguments."
- Fix "The event 'Key{Down/Press/Up}' has a non-standard type."
- Fix "The event 'Toggled' has a non-standard type."
- Fix "The type 'Label' does not define the field, constructor or member 'TextColor'" by removing .TextColor as it is not used.

### DIFF
--- a/FSharpExample/Program.fs
+++ b/FSharpExample/Program.fs
@@ -82,11 +82,12 @@ type Demo() = class end
             ()
 
     let ShowTextAlignments() =
-        let mutable container = new Dialog(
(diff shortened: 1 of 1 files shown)


tig (approved):
Thank you! Thank you! Thank you!

Issue #949, opened 2020-10-05
Most Glyphs not showing on macOS - E.g. Menu Glyphs and Character Map


BDisp:
It's possible to scroll top and see the old screen. It's a weird and ugly situation. Its happens with the `Terminal `and even `ITerm2`.

![iterm2](https://user-images.githubusercontent.com/13117724/95105441-16a98d80-072f-11eb-8d4d-f9caf418aeb7.gif)

tig:
The scrolling thing happens on Windows too.

Would you please do me the favor of running the Character Map scenario, selecting Box Drawing & Geometric Shapes, take a screenshot and post here? I don't know how to use a mac.

BDisp:
Here it is and I also enabled all encodings in the `Terminal`. `iTerm2 `is the same. Sorry for the delay.

![encodings](https://user-images.githubusercontent.com/13117724/95349013-86965000-08b6-11eb-856e-c9340b2c3423.png)

tig:
Da fuq? That's a pretty bad/deep bug. I wonder what I did wrong? 

As you scroll up/down do ANY glyphs show?

BDisp:
Here are all the encodings.

![mac-terminal](https://user-images.githubusercontent.com/13117724/95359773-6c626f00-08c2-11eb-8749-d82f61204fc8.gif)

tig:
I don't have time to dive into this today or tomorrow. But it seems there's some silly configuration related bug in ConsoleDriver/CursesDriver when running on Mac based on this.

tig:
Hey @mklement0 - you seem like you know what you are doing on a Mac. Can you help us with this?

mklement0:
I also see the symptom, both Terminal and iTerm2, and it seems that any non-ASCII-range character isn't recognized - even a window title such as `"Motörhead"` shows the problem, as does the `"Se iniciará el análisis"` label in the `./Example` project.

I know the Mac primarily from a shell perspective and know nothing about ncurses, so in order to be helpful I would need guidance.

Don't know if it's relevant, but skimming the source code I see `setlocale(LC_ALL, "");` in `binding.cs`, which should use the calling process' locale; from `man setlocale` on a Mac: 

> an argument of "" will determine the name of the new locale taking into account the environment variables LANG and LC_*`

In my case is `en_US.UTF-8`.

BDisp:
[...]

Issue #41, opened 2018-02-03
Must measure output size with Rune.ColumnWidth


migueldeicaza:
Many characters that take up two slots are not properly rendered as we render those two, but do not fully fill the requested region.   Need to use `Rune.ColumnWidth` to compute the current position.

tig:
@migueldeicaza do you have a test case for this?

tig:
FWIW, this is a disaster. I dare anyone to follow the discussions w.r.t. wide unicode chars in Windows Terminal land and successfully follow along... Or have any idea what we should do.

Examples: 

* https://github.com/microsoft/terminal/pull/2928
* https://github.com/microsoft/terminal/issues/900
* https://github.com/microsoft/terminal/issues/5095#issuecomment-619186373

tig:
FWIW, I believe at some point I modified the code (in `ConsoleDriver`? to explicitly ignore Runes with ColumnWidth > 1 and print the default char as a work around of this issue until it can be addressed correctly. In #949, it appears @BDisp discovered that I missed a case.

BDisp:
@migueldeicaza I think maybe it's time to use `System.Text.Rune`.

BDisp:
I submitted a PR at https://github.com/migueldeicaza/NStack/pull/48 to be able to control what kind of rune is being passed to `AddRune`. I hope that @migueldeicaza takes into account to be able to come up with a solution.

migueldeicaza:
The patch from @BDisp is now merged and should be on the new NStack.Core

BDisp:
@migueldeicaza there is another PR (https://github.com/migueldeicaza/NStack/pull/52) that fixes the "Chess Symbols (U+1fa00-1fa0f)". Thanks.

migueldeicaza:
Merged - I will release a new NStack in the morning.

PR #975, opened 2020-10-26
#41 and #949. Unit test to compare the difference between System.Rune and System.Text.Rune.

BDisp:


UnitTests/TextFormatterTests.cs
```diff
@@ -2242,5 +2242,69 @@ public void Reformat_Unicode_Wrap_Spaces_NewLines ()
...
+			Assert.Equal (1, Rune.ColumnWidth (c));
+			Assert.Equal (1, ustring.Make (c).ConsoleWidth);
+			Assert.Equal (1, ustring.Make (c).Length);
+
+			c = new System.Rune (123);
+			Assert.Equal (1, Rune.ColumnWidth (c));
+			Assert.Equal (1, ustring.Make (c).ConsoleWidth);
+			Assert.Equal (1, ustring.Make (c).Length);
+
+			c = new System.Rune (4432);
```
tig, line 2264: Pls use `/u` notation.

tig:
This may sound sacrilegious, but why do we use nstack at all?

BDisp:
I don't know why it gave the error.

BDisp:
> 
> 
> This may sound sacrilegious, but why do we use nstack at all?

Because of the `ustring `type that I think only exists in that library.

tig:
> > This may sound sacrilegious, but why do we use nstack at all?
> 
> Because of the `ustring `type that I think only exists in that library.

I meant to ask, why are we even using `ustring`?

BDisp:
> I meant to ask, why are we even using `ustring`?

I do not know why. It has always existed here. In fact, it was here that I learned of the existence of this type. Only @migueldeicaza can answer that.

BDisp:
> 
> 
> > > This may sound sacrilegious, but why do we use nstack at all?
> > 
> > 
> > Because of the `ustring `type that I think only exists in that library.
> 
> I meant to ask, why are we even using `ustring`?

@tig I think @migueldeicaza will prefer to use `NStack `because `System.Text.Rune` is not available in `net472` and in `netstandard2.0`.

PR #987, opened 2020-10-29
Fixes #986. Comparing with ShortcutHelper.GetModifiersKey in the demo.cs

### LINKED ISSUE #986: In demo.cs pressing any key in the TextField open always the menu.
- To compare the keys pressed always must use the `ShortcutHelper.GetModifiersKey` method to get all the keys combinations because the `shift `key is not injected into the `KeyEvent.Key` and thus returning a wrong comparation.

- I will also include a fix to FSharp project changing `Key.ControlQ` to `Key.Q`.

BDisp:


tig (commented):
LGTM

PR #991, opened 2020-11-04
Fixes #990. Pos and Dim only can be properly used when all the views are totally initialized.

### LINKED ISSUE #990: Adding dialog buttons causes redraw issues
In UICatalog the Dialogs scenario has redraw issues on the bottom button bar when "Add a button" button is pressed

image

version: current latest master, e4f5df6d

reproducible also in `cmd.exe` and release build.

BDisp:


tig (approved):
Very nice!

PR #996, opened 2020-11-09
Fixes #995. Improving TextField to work properly with Unicode characters.

### LINKED ISSUE #995: Most of the TextField functions doesn't work properly with Unicode characters.
- I'm talking about keys navigation, text selection, copy and paste.
- Initially it only started to work with `Ascii `characters and now it it fails with `Unicode `characters.

BDisp:


Terminal.Gui/Views/TextField.cs
```diff
@@ -264,10 +269,9 @@ void Adjust ()
 			if (point < first) {
 				first = point;
 			} else if (first + point - (Frame.Width + offB) == 0 ||
-				  DisplaySize (text, first, point) >= Frame.Width + offB) {
+				  DisplaySize (text, first, point).Item1 >= Frame.Width + offB) {
```
tig, line 272: Should be `DisplaySize (...).Length` in order for the code to be more readable.

Terminal.Gui/Views/TextField.cs
```diff
@@ -147,11 +147,10 @@ public override bool OnLeave (View view)
 				}
 				TextChanged?.Invoke (oldText);
 
-				if (point > text.Count)
-					point = Math.Max (DisplaySize (text, 0) - 1, 0);
+				if (point > text.Count) {
+					point = Math.Max (DisplaySize (text, 0).Item1 - 1, 0);
```
tig, line 151: Should be `DisplaySize (...).Length` in order for code to be more readable.

tig (changes requested):

Terminal.Gui/Views/TextField.cs
```diff
@@ -242,20 +241,26 @@ static int SetCol (int col, int width, int cols)
 			return col;
 		}
 
-		// Returns the size in a range of the string.
-		int DisplaySize (List<Rune> t, int start = -1, int end = -1)
+		// Returns the size and length in a range of the string.
+		(int, int) DisplaySize (List<Rune> t, int start = -1, int end = -1, bool checkNextRune = true)
```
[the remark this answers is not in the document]
BDisp: I agree!

Terminal.Gui/Views/TextField.cs
BDisp: In this case Item1 is `Size `and not `Length`. I'll change to `DisplaySize (...).Size`.

Terminal.Gui/Views/TextField.cs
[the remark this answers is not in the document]
BDisp: @tig it's not possible to use the named items. It's only possible when we instantiate a new one. Please review that.

Terminal.Gui/Views/TextField.cs
BDisp: @tig it's not possible to use the named items. It's only possible when we instantiate a new one. Please review that.

Terminal.Gui/Views/TextField.cs
BDisp: @tig it's not possible to use the named items. It's only possible when we instantiate a new one. Please review that.

Terminal.Gui/Views/TextField.cs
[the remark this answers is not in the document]
tig: I don't understand. If you declare the `DisplaySize` function to return a tuple, and name the items in the tuple (as I showed above), this will all work. Or i'm missing something fundamental. Which could be the case. I swear I've done this before.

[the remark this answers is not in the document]
BDisp: @tig sorry. Yes you are right. I did something wrong. I'll change. Thanks.

tig (approved):

BDisp:
@tig, my only question is whether we should use items named as uppercase or lowercase. Then say something please.

tig:
> @tig, my only question is whether we should use named items beginning with a capital or small letter. Then say something please.

I view method return tuples the same as I view method parameters in this regard: they should be camel case. I erred in my sample and I should have noticed this before accepting this PR. Feel free to do another to fix (`size` and `length`). Not a huge deal.

tig:
> > @tig, my only question is whether we should use named items beginning with a capital or small letter. Then say something please.
> 
> I view method return tuples the same as I view method parameters in this regard: they should be camel case. I erred in my sample and I should have noticed this before accepting this PR. Feel free to do another to fix (`size` and `length`). Not a huge deal.

Oops I thought I already merged this. So no need for a new PR.

BDisp:
Done @tig. Thanks.

BDisp:
@tig don't forget to merge this PR please.

PR #1000, opened 2020-11-13
Fixes #999. Toplevel should only redraw the subviews if !NeedDisplay.IsEmpty or layoutNeeded.

### LINKED ISSUE #999: Toplevel should only redraw the subviews if !NeedDisplay.IsEmpty or layoutNeeded.

BDisp:


tig (approved):
LGTM

PR #1003, opened 2020-11-15
Fixes #1002. Added a AutoSize property to the View class.

### LINKED ISSUE #1002: Label does not expand to accommodate new text
I'm not sure if this is intended behavior or not.

So I have code that initializes several labels
```C#
registerLabels = new Label[8];
for (int i = 0; i < 8; i++) {
	//We need to make the label 30-wide on init because it won't expand
	Label l = new Label(1, i, $"R{i}");
	registerLabels[i] = l;
}
```

I then add these labels to a window and start the program.
```C#
Window w = new Window(new Rect(x, 1, 32, 16), "Registers");
w.Add(registerLabels);
```

While the program is running, I have to update the labels
```C#
public void UpdateRegisterView() {
	for (int i = 0; i < 8; i++) {
		var r = lc3.processing.registers[i];
		var s = $"R{i} {r.ToRegisterString()}";
		registerLabels[i].Text = s;
	}
}
```

So I noticed that the register values apparently don't update. I later found out that the `Text` field does get updated, but the visible label cuts off anything past the length of the initial `Text` value. There's no warning that it does this.

I know that the last time I used this library, the Label showed the entire `Text` on update, regardless of the length of the initial value.

BDisp:


BDisp:
Sorry I forgot to create a new branch for this PR.

PR #1004, opened 2020-11-15
Fixes #1002. Added a AutoSize property to the View class.

### LINKED ISSUE #1002: Label does not expand to accommodate new text

BDisp:


Terminal.Gui/Core/View.cs
```diff
@@ -1899,25 +1900,27 @@ public virtual void LayoutSubviews ()
...
-					}
-				}
+				ResizeView (autoSize);
 				SetNeedsLayout ();
 				SetNeedsDisplay ();
 			}
 		}
 
+		/// <summary>
+		/// Used by <see cref="Text"/> to resize the view's <see cref="Bounds"/> with the <see cref="TextFormatter.Size"/>.
```
tig, line 1910: Can you add documentation explaining that this should only be used with Absolute layout; it will not work with Computed layout?

tig (changes requested):
Good stuff, but it'd be great if the docs made it clear this only works with Absolute layout

Terminal.Gui/Core/View.cs
BDisp: Ok. Done!

tig (approved):

PR #1005, opened 2020-11-15
Fixes #998. Added a cancelable TextChanging event to prevent the TextChanged event being called if the changing is canceled.

### LINKED ISSUE #998: Question: TextChanged Action
Good evening,

I am working on a console with this gui to edit configuration. I am having a really hard time controlling the TextField. In WPF this is really easy for me. I can filter and interact with the main sender and handle the event as well.

 ```
  private void TextboxTest_PreviewTextInput(object sender, TextCompositionEventArgs e)
        {
            TextBox OS = (TextBox)sender;
            if (OS.Text.Length > 4) // Don't allow more than 4 digits
            {
                e.Handled = true;
                return;
            }
            e.Handled = Regex.IsMatch(e.Text, "[^0-9]+"); //Match Digits only
        }
```

On gui.cs I am using an action

```
Label LabelSSHPort = new Label(17, 1, "port: ");
            TextField TextFieldSSHPort = new TextField(24, 1, 5, MainConfig.LoginServersSSH.Qstr("port"));
            TextFieldSSHPort.TextChanged += UpdateSSHPort;

      private static void UpdateSSHPort(ustring e)
        {
            string PortString = e.ToString();
            string result = Regex.Replace(PortString, @"[^\d]", "");
           //Snipped
            if (Int32.TryParse(result, out ParsedPort))
            {
                MainConfig.LoginServersSSH["port"] = ParsedPort;
            }
        }

```

The first issue with the action is it isn't returning the changed text. It sends me the text of the field before the change has been processed. In this case I have a port number. Its port is 8889. If I type anything, The text changed will show me 8889 and not include what I just typed. So I don't know how to process and filter what has changed.

Also since the action is just the text, even if it was including the changed text. It doesn't include the sender like the events. My variables for the menus are not global so I don't have an easy way to access them in my function.

BDisp:


tig (approved):
Nice!

PR #1006, opened 2020-11-15
Removal NameFieldLabel. Change "Open" to "File"

SemperFu:
I noticed the SaveDialog used Open for the filename. 

The NameFieldLabel was hardcoded and ignoring the value passed by the OpenDialog and SaveDialog

Save1

At first I changed it to use the passed fields but it doesn't really make since with Open and Save as the file name. The buttons at the bottom are Open and Save. 

Save2

The top field is the directory and the bottom is the filename it self. I think it would make sense to name it "File" or "Filename" like Directory. 
File1
File2
Word

This matches normal FileDialogs as well. So if this would stay hard coded I just removed the the input fields from the Open and SaveDialogs all together.

### DIFF
--- a/Terminal.Gui/Windows/FileDialog.cs
+++ b/Terminal.Gui/Windows/FileDialog.cs
@@ -479,16 +479,15 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Initializes a new <see cref="FileDialog"/>.
 		/// </summary>
-		public FileDialog () : this (title: string.Empty, prompt: string.Empty, nameFieldLabel: string.Empty, message: string.Empty) { }
+		public FileDialog () : this (title: string.Empty, prompt: string.Empty, message: string.Empty) { }
 
 		/// <summary>
 		/// Initializes a new instance of <see cref="FileDialog"/>
 		/// </summary>
 		/// <param name="title">The title.</param>
 		/// <param name="prompt">The prompt.</param>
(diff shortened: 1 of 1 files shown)


tig (changes requested):
I think this is a breaking API change as you've coded it, right? Existing code that tries to use `nameFieldLabel` will no longer build. 

Did you check to see if this breaks any existing apps?

migueldeicaza:
Good point, I would introduce a new constructor/API with the additional parameter.

Alternatively, this could be a property that is set after it has been created, not something that must be passed during construction.

SemperFu:
Sorry I missed the first response, yes it would be a breaking change for someone who uses filedialog directly but it would not break Save and Open Dialogs. I did leave the ability to set the property after the creation. The problem is you can only have 4 chars based on how the window is built, it doesn't take into account the number of chars to position it. 

I can add modify the pull for an additional constructor this weekend.

tig:
> Sorry I missed the first response, yes it would be a breaking change for someone who uses filedialog directly but it would not break Save and Open Dialogs. I did leave the ability to set the property after the creation. The problem is you can only have 4 chars based on how the window is built, it doesn't take into account the number of chars to position it.
> 
> I can add modify the pull for an additional constructor this weekend.

Hey @SemperFu , did you get  a chance to make the requested changes?

BDisp:
This was already fixed on PR #1157 with the option to the derived class now can choose the both label names.

PR #1014, opened 2020-11-17
Trying fixing #518. Almost functions work on both Windows and Unix with the NetDriver.

BDisp:
- To run this PR is needed passing the -usc argument in the `UICatalog` project or uncomment the `Application.UseSystemConsole = true;` in the `Example `or `UICatalog `projects. On the `UICatalog `trying select "`Use System Console`" from the menu item "`Diagnostics`", which is bad spelled, won't work because a driver must be setted on startup and not after driver was being created. To open the menu press `F9` or `Alt+hotkey`.
- Interestingly, it runs faster on `WSL `than on `CMD`.
- Consumes more `CPU `stress than the `WindowsDriver `or `CursesDriver `may be due the workaround with the `WaitWinChange `loop method to capture window resizing.
- It could be a good alternative to use a common platform for both `OS `with the new `Net5.0` version fusion of the `Net Core` and `Net Framework` in only one tool. But we need to wait a long time yet :-)
- I think that the `CSI (Code Sequence Identifier)` could be a cross reference to achieve most of the common functions cross platform `OS `compatible.
- Control keys, like `Ctrl`, `Alt`, `Shift`, `Capslock `and `Numlock`, aren't detected alone without combinations with other keys.
- In `WSL` `Shift+End` prints "`1;2F`" on a `TextField `instead selecting the characters on the right.
- I think mouse interaction could be achieved with `CSI `code and developers with more knowledge with code sequences would be a very good help.
- I would prefer convert the buffer size with current window size but I couldn't achieve and that why I created the new abstract property `Top `to only display from the current windows top unto the window height. Unfortunately in the `WSL `the `Top `is always equal to `0` no matter where the position of the vertical scroll bar.
- The good news is that all events like timers, events actions methods, thread events, all work without any problems. Try run the `Threading `scenario and you will see that.
- On `WSL `the terminal does not hangs with the Net5.0 unlike using ncurses.
- The `UICatalog `"`Diagnostics`" menu was changed to use all the `Enum ConsoleDriver.DiagnosticFlags` and was removed the "`Use _System Console`" which has nothing to do with these diagnostics.

### DIFF
--- a/Example/Example.csproj
+++ b/Example/Example.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -17,9 +17,9 @@ namespace Terminal.Gui {
 	/// This is the Curses driver for the gui.cs/Terminal framework.
 	/// </summary>
 	internal class CursesDriver : ConsoleDriver {
-#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
 		public override int Cols => Curses.Cols;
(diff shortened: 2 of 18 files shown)


tig:
Almost? What doesn't work?

BDisp:
For now only the mouse. I only used Console properties and functions that work on both.

Issue #255, opened 2019-09-03
Tabular data?


Bartolomeus-649:
Does gui.cs handle tabular data?
* Auto sizing columns?
* line breaks within columns
* truncate column content?
* scroll the table if needed
 and so on?

Anything like this or something similar?

migueldeicaza:
No, but that would be a cool control - a CSV control for ListViews.

```
public class CsvListView : ListView {
    public CsvListView () {}
    public string Separator { get; set; }
    public string Quotes { get; set; }
}
```

tig:
Closing as this is materially addressed with #1015

PR #1015, opened 2020-11-19
New Control 'TableView' for viewing tabular data

tznind:
Hey,

I have been using your awesome framework for a while and wanted to contribute.  I have written a control for #255 .  I have reviewed the contribution guidelines and code of conduct.

I have added a new `Scenario` and tested it in Windows and MXLinux.

- The control works with `System.DataTable`
   - I thought that would be a good way to avoid parsing CSVs as well as supporting viewing results from relational databases etc
- Columns autosize based on the column widths and the widths of the **visible** cells
- Supports horizontal and vertical scrolling including PageUp/Down/Home/End
- User can configure 
  - Maximum cell width
  - How to represent nulls
  - What character to use as a column separator (defaults to space)

Currently when you change the contents of the `Table` or the `ColumnOffset` or `SelectedColumn` you will also have to call `Update` before it will re-render.  This lets the API user drop columns or make many changes to the `System.DataTable` without having to respond to every one.  But it could also be confusing if the user changes the `SelectedRow` and doesn't see the UI immediately update.

image

Let me know if there are inefficiencies in the `Redraw`.  I noticed that the `HexView` (which I based this on) has a local method `SetAttribute` which I kept... although I'm not using it.

I'm unsure about the guidline (any guidance on how to do that would be great please):
> Ensure the UICatalog demo for the new class illustrates both Absolutle Layout and Computed Layout.

Currently the control doesn't explicitly handle newlines in cell values but those seem to be replaced by spaces in DrawString anyway.

### DIFF
--- a/Terminal.Gui/Views/TableView.cs
+++ b/Terminal.Gui/Views/TableView.cs
@@ -6,6 +6,9 @@ using System.Linq;
 
 namespace Terminal.Gui.Views {
 
+	/// <summary>
+	/// Describes how to render a given column in  a <see cref="TableView"/> including <see cref="Alignment"/> and textual representation of cells (e.g. date formats)
+	/// </summary>
 	public class ColumnStyle {
 		
 		/// <summary>
@@ -448,7 +451,7 @@ namespace Terminal.Gui.Views {
 		}
 
 		/// <summary>
-		/// Truncates or pads <paramref name="representation"/> so that it occupies a exactly <paramref name="availableHorizontalSpace"/> using the alignment specified in <paramref name="style"/> (or left if no style is defined)
+		/// Truncates or pads <paramref name="representation"/> so that it occupies a exactly <paramref name="availableHorizontalSpace"/> using the alignment specified in <paramref name="colStyle"/> (or left if no style is defined)
 		/// </summary>
 		/// <param name="originalCellValue">The object in this cell of the <see cref="Table"/></param>
 		/// <param name="representation">The string representation of <paramref name="originalCellValue"/></param>
--- a/Terminal.Gui/Views/TableView.cs
+++ b/Terminal.Gui/Views/TableView.cs
@@ -172,8 +172,15 @@ namespace Terminal.Gui {
 		public int SelectedColumn {
 			get => selectedColumn;
 
-			//try to prevent this being set to an out of bounds column
-			set => selectedColumn = Table == null ? 0 :  Math.Min (Table.Columns.Count - 1, Math.Max (0, value));
+			set {
+				var oldValue = selectedColumn;
+
+				//try to prevent this being set to an out of bounds column
+				selectedColumn = Table == null ? 0 :  Math.Min (Table.Columns.Count - 1, Math.Max (0, value));
+
+				if(oldValue != selectedColumn)
+					OnSelectedCellChanged();
+			} 
 		}
 
 		/// <summary>
@@ -181,7 +188,15 @@ namespace Terminal.Gui {
 		/// </summary>
 		public int SelectedRow {
 			get => selectedRow;
-			set => selectedRow =  Table == null ? 0 : Math.Min (Table.Rows.Count - 1, Math.Max (0, value));
+			set {
+
+				var oldValue = selectedColumn;
+
+				selectedRow =  Table == null ? 0 : Math.Min (Table.Rows.Count - 1, Math.Max (0, value));
+
+				if(oldValue != selectedRow)
+					OnSelectedCellChanged();
+			}
 		}
 
 		/// <summary>
@@ -198,6 +213,11 @@ namespace Terminal.Gui {
 		/// The symbol to add after each cell value and header value to visually seperate values (if not using vertical gridlines)
 		/// </summary>
 		public char SeparatorSymbol { get; set; } = ' ';
+
+		/// <summary>
+		/// This event is raised when the selected cell in the table changes.
+		/// </summary>
+		public event Action<EventArgs> SelectedCellChanged;
 
 		/// <summary>
 		/// Initialzies a <see cref="TableView"/> class using <see cref="LayoutStyle.Computed"/> layout. 
@@ -690,6 +710,14 @@ namespace Terminal.Gui {
 			if (SelectedRow < RowOffset) {
 				RowOffset = SelectedRow;
 			}
+		}
+
+		/// <summary>
+		/// Invokes the <see cref="SelectedCellChanged"/> event
+		/// </summary>
+		protected virtual void OnSelectedCellChanged()
+		{
+			SelectedCellChanged?.Invoke(new EventArgs());
 		}
 
 		/// <summary>
--- a/Terminal.Gui/Views/TableView.cs
+++ b/Terminal.Gui/Views/TableView.cs
@@ -579,6 +579,22 @@ namespace Terminal.Gui {
 			return true;
 		}
 
+		/// <summary>
+		/// Positions the cursor in the area of the screen in which the start of the active cell is rendered.  Calls base implementation if active cell is not visible due to scrolling or table is loaded etc
+		/// </summary>
+		public override void PositionCursor()
+		{
+			if(Table == null) {
+				base.PositionCursor();
+				return;
+			}
+				
+			var screenPoint = CellToScreen(SelectedColumn,SelectedRow);
+			
+			if(screenPoint  != null)
+				Move(screenPoint.Value.X,screenPoint.Value.Y);
+		}
+
 		///<inheritdoc/>
 		public override bool MouseEvent (MouseEvent me)
 		{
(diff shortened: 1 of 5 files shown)


BDisp:
It looks good, thanks.



> I'm unsure about the guidline (any guidance on how to do that would be great please):
> 
> > Ensure the UICatalog demo for the new class illustrates both Absolutle Layout and Computed Layout.

See the `Button` `Update `method on how to change the `Width`. If it isn't null or an absolute value you must use the necessary code to change his `Width `and `Height` based on their type, unless that a control have a specific `Width `or `Height`, like the case of the `Button` `Height `is always equal to `1` where you must save the state of the `LayoutStyle `before changing his value and restore to the original `LayoutStyle `value. If a view have it's `Layout Style.Computed` the position and size depends on other views and changing to an absolute value causes the views stop to be anchored.

BDisp:
Don't forget to add documentation to the "`Missing XML comment`" warnings.

BDisp:
If you aren't use the `SetAttribute `you can remove because already exist a `ColorScheme `property in the base class `View`.

tznind:
Thanks, now that I look again at `Redraw` I think I am confusing `bounds` and `Frame`.  I think all my size calculations (what columns to render where) should be done solely on `Frame` is that right?

BDisp:
> Thanks, I will add comments and tidy up pointless variables. Now that I look again at `Redraw` I think I am confusing `bounds` and `Frame`. I think all my size calculations (what columns to render where) should be done solely on `Frame` is that right?

`Bounds `is related to the position and size of own view. `Frame `is related to the position and size in the screen. Only the size is equal on both. So in general it's only needed to deal with the `Bounds `property from `0` to `Bounds.Width` and `0` to `Bounds.Height` and the `View `class will deal with the convertion from view position to screen position. But there are some situations where you need to deal with the `Frame `property to check some positions and size against the windows size to take some actions needed.

tznind:
> 
> 
> > Thanks, I will add comments and tidy up pointless variables. Now that I look again at `Redraw` I think I am confusing `bounds` and `Frame`. I think all my size calculations (what columns to render where) should be done solely on `Frame` is that right?
> 
> `Bounds `is related to the position and size of own view. `Frame `is related to the position and size in the screen. Only the size is equal on both. So in general it's only needed to deal with the `Bounds `property from `0` to `Bounds.Width` and `0` to `Bounds.Height` and the `View `class will deal with the convertion from view position to screen position. But there are some situations where you need to deal with the `Frame `property to check some positions and size against the windows size to take some actions needed.

Ok Cool thanks for that explanation, then I will leave it how it is for now.

tig:
I'll try to take a close look at this asap because I think it's gonna be awesome. I will see if I can re-write `out-consolegridview` to use this instead of the hacky way it simulates a table view today!

migueldeicaza:
Hello,

Is System.Data.DataTable available on .NET Core and .NET Standard?   

One thing that might be good to do with this view is to support the rendering similar to what Spectre.Console has, allowing users to set borders for example:

https://www.hanselman.com/blog/spectreconsole-lets-you-make-beautiful-console-apps-with-net-core

tznind:
> Hello,
> 
> Is System.Data.DataTable available on .NET Core and .NET Standard?
> 

Hi, thanks for taking the time to look at this PR.  Yes it was in dotnet core 1.0 (Dotnet Standard 2.0 onwards)

https://docs.microsoft.com/en-us/dotnet/api/system.data.datatable?view=net-5.0#applies-to

> One thing that might be good to do with this view is to support the rendering similar to what Spectre.Console has, allowing users to set borders for example:
> 
> https://www.hanselman.com/blog/spectreconsole-lets-you-make-beautiful-console-apps-with-net-core

Thanks for pointing me there they look awesome.  I have added table border lines and toggles for them in UICatalog.

image

BDisp:
Great job, thanks. Perhaps it may be implemented cells alignment.

tznind:
> Great job, thanks. Perhaps it may be implemented cells alignment.

Thanks, I have added a `ColumnStyle` class so that some columns can have different widths, alignment,formats  etc.

image

See method `SetDemoTableStyles()` in UICatalog for example of how to configure them.  I also added a menu item to the Scenario that clears the styles.

migueldeicaza:
Wow, this is looking fantastic!

In the DoubleCol screenshot, I imagine that this can be aligned right, and that I can also specify the formatting options?

tznind:
> 
> 
> Wow, this is looking fantastic!

Thanks!

> In the DoubleCol screenshot, I imagine that this can be aligned right, and that I can also specify the formatting options?

Yes, you can either specify a fixed alignment for the col:

```csharp
Alignment = TextAlignment.Right,
```

Or you can assign alignment based on the values (e.g. negative values aligned right, positive values aligned left)
```csharp
AlignmentGetter = (v)=>v is double d ? d < 0 ? TextAlignment.Right : TextAlignment.Left:
								// not a double e.g. could be DBNull.Value
								TextAlignment.Left
```

For formatting theres a delegate too... although that might be overkill, a simple string property `Format` might be better if I can work out a way to do that.

```csharp
var dateFormatStyle = new ColumnStyle() {
				Alignment = TextAlignment.Right,
				RepresentationGetter = (v)=> v is DateTime d ? d.ToString("yyyy-MM-dd"):v.ToString()
};
```

You can also specify maximum and minimum widths for cols in the `ColumnStyle` class

dynamoRando:
Just wanted to say that this is great and have started used it. Appreciate the contribution! I have the class manually added to my code, looking forward to when this is merged!

Have a great day!

tznind:
I have added mouse support to table view

- Clicking a cell selects the cell
- Mouse wheel scrolls up/down 

Not sure if there is a way to tell if the Shift key is down in which case Shift Mouse Wheel could do horizontal scrolling.

table-view-mouse

BDisp:
Very great job, many thanks. You can see the `ScrollView `sample where it deals with `WheeledLeft `and `WheeledRight `flags.
It's possible to select to entire line? It will be nicer. It also be nicer on double click will edit the cell like F3 does.

tznind:
FullRowSelect makes sense but would make horizontal scrolling with keyboard alone a challenge.  Currently when you move the selected cell using the arrow keys the scroll area updates when it moves off the current scroll area (like a spreadsheet does).

With full row selection though there is no horizontal 'current cell'... I guess the simplest way would be to make full row select still include a single 'active cell' sort of like this:

image

BDisp:
That's would be nice too. Thanks.

tznind:
Ok have added FullRowSelect property (defaults to false).  I implemented `PositionCursor` so the selected cell always has the cursor in it.  This should ensure the user can tell which cell is the 'active' one within the row even when the whole row is selected.

image

BDisp:
Fantastic you are awesome. Thanks.

tznind:
I have added multi selection (Shift + arrow/home/end key):

multi-select

Is there a way to tell if the Shift/Ctrl key are down during mouse activities e.g. click one cell and ctrl click another.

BDisp:
> Is there a way to tell if the Shift/Ctrl key are down during mouse activities e.g. click one cell and ctrl click another.

You have to check for the `MouseFlags.ButtonShift` and `MouseFlags.ButtonCtrl`. Unfortunately in `Linux `doesn't not work as well as in `Windows`.

BDisp:
I forgot to mention that I dealed with `MouseFlags.ButtonShift` and `MouseFlags.ButtonCtrl` a long time ago in the `FileDialog.cs` at the `MouseEvent `method on `DirListView `class.

tig:
Like TreeView, I feel like we need a good collection of use-cases coded up before merging a big new control like this.

The more there are, the more confidence we all will have that something fundamental has not been missed in the design/API.

What are the current samples that have been written?

tznind:
Thanks for the guidance @tig

I have added a simple CSV editor use case to the Scenarios.  Given there are entire libraries around parsing / saving CSV data I've not tried to handle newlines, escaping etc.  I'm assuming the idea is primarily demo the control and how to do common tasks with it.

csvExample

I've also added an article to the docs on common use cases and how to get a DataTable for them:

https://github.com/tznind/gui.cs/blob/table-view/docfx/articles/tableview.md

Let me know if this is the right way to go and I can expand on the docs to cover common tasks, limitations etc.

BDisp:
Impressive. Is it possible to move a column or row to another location?

tznind:
> Impressive. Is it possible to move a column or row to another location?

Thanks, yes you can do most things with `DataTable` since it is a core class of C# - it is pretty versatile.  I've added moving rows and sorting to the example.  All these operations (sort, reorder) change the `System.DataTable` data directly.

The only 'view' level manipulations I've implemented so far are the `Format` and `MaxLength` etc fields in `TableStyle`.  You can see if you put a Format on a column it only affects the presentation layer not the value stored (so you can change the Format later on if you want).

migueldeicaza:
This is outstanding!   Thanks for this beautiful work!

tig:
I *finally* got a chance to test and play with this. Thank you!!!!

dynamoRando:
Awesome stuff!

lzanoni-stm:
Is there a way to remove the selected style when the focus leaves the TableView  (and put it back when entering)?

lzanoni-stm:
Thanks for the blazing fast reply and example :)

Sorry, I probably haven't used the right words: what I'd like is that, when losing focus, the currently selected cell becomes normal (without the highlight), without altering the rest of the table.

tznind:
Good point.  I have a prototype fix in #1776  please can you write this as a new issue so that it is easier to find and not mixed in with this old PR?

PR #1021, opened 2020-11-23
Fixes #1020. NetDriver does not deal well with unicode characters yet.

### LINKED ISSUE #1020: NetDriver does not deal well with unicode characters yet.
netdriver-unicode

BDisp:
- In this fix I included a new `Enum `property that is `HeightSize `which allows to use the window height or the buffer height  size.
- I also splited the `WindowsDriver `main loop class allowing `DI` (Dependency Injection) and thus works like the `NetDriver `and the `CursesDriver`.
- I changed the `CharacterMap `scenario to allows to toggle to `DRAW_CONTENT `or `BASE_DRAW_CONTENT `more easily.
- In the `CharacterMap `scenario with the `Consolas `font only the "`Chess Symbols (U+1fa00-1fa0f)`" fail with unicode, but I already submit a PR (https://github.com/migueldeicaza/NStack/pull/52) that fixes this.
- The `Rune.ColumnWidth` method will fail with others not true fonts because it calculates the column width based on a range of unicode characters tested with true font and the same range will fail with other fonts types, like "`MS Gothic`", with the same range but with diferents sizes. I hope if it has another way to solves this one day.
- I didn't use the` OperatingSystem.IsWindows` because it isn't available on `net472 `and `netstandard2.0`.
- The menu can be accessed in `Windows `pressing `Ctrl+Space` and in `Unix `as it's usual by pressing `Alt+Space`. In `Windows `this combination  open the window menu.

### DIFF
--- a/Terminal.Gui/Core/ConsoleDriver.cs
+++ b/Terminal.Gui/Core/ConsoleDriver.cs
@@ -589,12 +589,10 @@ namespace Terminal.Gui {
 		/// <returns></returns>
 		public static Rune MakePrintable (Rune c)
 		{
-			if (c <= 0x1F) {
-				// ASCII (C0) control characters. 
-				return new Rune (c + 0x2400);
-			} else if (c >= 0x80 && c <= 0x9F) {
+			if (c <= 0x1F || (c >= 0x80 && c <= 0x9F)) {
+				// ASCII (C0) control characters.
 				// C1 control characters (https://www.aivosto.com/articles/control-characters.html#c1)
-				return new Rune (0x25a1); // U+25A1, WHITE SQUARE, □: 
+				return new Rune (c + 0x2400);
 			} else {
 				return c;
 			}
--- a/Terminal.Gui/Core/ConsoleDriver.cs
+++ b/Terminal.Gui/Core/ConsoleDriver.cs
@@ -458,6 +458,20 @@ namespace Terminal.Gui {
 		public static Dictionary<string, ColorScheme> ColorSchemes { get; }
 	}
 
+	/// <summary>
+	/// The visible height should be used on the window.
+	/// </summary>
+	public enum HeightSize {
+		/// <summary>
+		/// Only window height will be visible not allowing scroll.
+		/// </summary>
+		WindowHeight,
+		/// <summary>
+		/// All buffer height will be visible allowing scroll.
+		/// </summary>
+		BufferHeight
+	}
+
 	///// <summary>
 	///// Special characters that can be drawn with 
 	///// </summary>
@@ -545,6 +559,11 @@ namespace Terminal.Gui {
 		/// The current top in the terminal.
 		/// </summary>
 		public abstract int Top { get; }
+
+		/// <summary>
+		/// The current <see cref="HeightSize"/> used in the terminal.
+		/// </summary>
+		public abstract HeightSize HeightSize { get; set; }
 
 		/// <summary>
 		/// Initializes the driver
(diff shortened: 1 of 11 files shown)


Terminal.Gui/Core/ConsoleDriver.cs
```diff
@@ -458,6 +458,20 @@ static void SetColorScheme (ColorScheme colorScheme, [CallerMemberName] string c
 		public static Dictionary<string, ColorScheme> ColorSchemes { get; }
 	}
 
+	/// <summary>
+	/// The visible height should be used on the window.
+	/// </summary>
+	public enum HeightSize {
```
tig, line 464: I'm having trouble understanding the naming here. `HeightSize` doesn't see right. Would a more descriptive name be 'TerminalHeight` or `ConsoleHeight`? Or `VisibleHeight` (since the doc reads `the *visible height*`)?

Actually...

Why is this an enum and not just a boolean? Do you believe you'll have more than 2 values in the future (behind Window and Buffer)?

Why not simplify this by making this 

```
/// if false height is measured by the window height and thus no scrolling. If true then height is measured by the buffer height, enabling scrolling
public bool HeightAsBuffer {get;set;}
```

?

Or, why not name it around the scrolling behavior (because I think that's what matters)?

tig (changes requested):
I think the naming of `HeightSize` (and the fact it is an enum) is confusing and overly complex. Please try to simplify by just making it a bool and renaming it to something more logical.

Terminal.Gui/Core/ConsoleDriver.cs
BDisp: I think `public bool HeightAsBuffer {get;set;}` is fine to me.

Terminal.Gui/Core/View.cs
```diff
@@ -1328,12 +1331,31 @@ public virtual void Redraw (Rect bounds)
...
 
+			if (IsCurrentTop || this == Application.Top) {
+				if (!NeedDisplay.IsEmpty || LayoutNeeded) {
+					Driver.SetAttribute (Colors.TopLevel.Normal);
+
+					// This is the Application.Top. Clear just the region we're being asked to redraw 
+					// (the bounds passed to us).
+					Clear (bounds);
+					Driver.SetAttribute (Colors.Base.Normal);
+					((Toplevel)this).PositionToplevels ();
```
tig, line 1342: I really don't like how this breaks the class hierarchy. A base class (`View`) should not, generally, have knowledge of a derived class (`TopLevel`). 

Do you think this is good practice?

BDisp: For a toplevel may be it is not so bad. If was the Window that is derived from Toplevel that is badly. But if you want I can revert that. So say something.

BDisp: Reverted. Thanks.

BDisp:
I am facing a weirdo behavior with the `NetDriver `rendering. If I run from debug in the `Visual Studio 2019` or run on the `WSL`, the views show all right. But if I use the `cmd.exe`, `PowerShell `or `ConEmu `the view moves outside the top causing an misaligned behavior. I created a `.NET Framework 472` project of the `Example `and that behavior also exist. So, I think this is causing by `.NET Framework` that is used by the `Windows `on that terminals. If someone know what is going on, I appreciate any help on this. Thanks.

BDisp:
Think I found the culprit for this damn behavior. When writing in the last column the `System.Console `insert a newline and thus, if it is in the last row, push everything upside one line. With the `Net Core` and `Net5.0` that don't happen and I don't know if I avoid writing in the last column, which seems ugly and dirty, or let this behavior like it is in the `.Net Framework`.

tig (approved):

Issue #913, opened 2020-09-28
Someone build a TreeView please


tig:
There have been several requests for a tree view for Terminal.Gui. 

E.g. see this twitter convo between @SteveL-MSFT, @tig, and @migueldeicaza.
https://twitter.com/Steve_MSFT/status/1310360781024706560

There's no reason this has to be a built-in part of `Terminal.Gui` though. Someone could build it as an independent package.

tig:
Closing. thank you @tznind !

PR #1022, opened 2020-11-25
Tree view

tznind:
#913

Hey.  I have created a tree view based on delegates.  The user defines `ChildrenGetter` method (and optionally `CanExpandGetter` for performance) then adds root objects.  I copied the getter pattern from object list view (http://objectlistview.sourceforge.net/cs/index.html).  I also emailed the author of that package to check that he was fine with me reusing it and he has confirmed it's not a problem (I can forward/attach the email chain if needed).

image

For the example in UICatalog I used the file system since it is a nice tree structure that goes deep, looks impressive and has mixed object Types `FileInfo` and `DirectoryInfo` with minimal amount of setup code:

```csharp
// As a shortcut to enumerating half the file system, tell tree that all directories are expandable (even if they turn out to be empty later on)
_treeView.CanExpandGetter = (o)=>o is DirectoryInfo;

// Determines how to compute children of any given branch
_treeView.ChildrenGetter = GetChildren;

...

private IEnumerable<object> GetChildren(object model)
{
	// If it is a directory it's children are all contained files and dirs
	if(model is DirectoryInfo d) {
		try {
			return d.GetFileSystemInfos()
				//show directories first
				.OrderBy(a=>a is DirectoryInfo ? 0:1)
				.ThenBy(b=>b.Name);
		}
		catch(SystemException ex) {
			// if there is an error enumerating a directory we can put the Exception in the tree too if we want <3
			return new []{ex};
		}
	}

	return new object[0];
}
```

Let me know if this isn't the route you want to go down or if you would rather it were in a seperate package.


tig:
This appears to kick ass. I have added it to my backlog of things to dive deep into. Not sure it's ready to merge yet. I'd like to know there are three or four fairly different use cases before adopting a control like this directly into the library. 

I have been thinking of how to enhance or augment `out-consolegridview` to be able to browse PowerShell object hierarchies. This might (or might not) really work for this.

tznind:
I've started adding a few bits and bobs (events, tests, refresh etc).  Let me know if this is generating spam and I can hold off (I don't want to make any faux pas!).  I am using a copy of the code in a GPL app I'm developing so working out the kinks as I go.

migueldeicaza:
I think that the tree should have an option to render the actual tree elements, we have the characters available, so it should be present.

I think it is worth thinking about the data model for the tree, the delegate is an interesting idea, but a bit unusual.   

Lastly, yes, we would need an explicit statement from the original author that he is willing to relinese the code under the MIT license, and his name and copyright under the MIT license needs to be attached to the patch, as this currently stands, this code should not be merged due to the GPLv3 license.

tznind:
> 
> I think that the tree should have an option to render the actual tree elements, we have the characters available, so it should be present.

Thanks for taking the time to look at the PR, by 'actual tree elements' do you mean the lines between nodes? Happy to put this in should be relatively easy :)

image

> I think it is worth thinking about the data model for the tree, the delegate is an interesting idea, but a bit unusual.

Happy to take guidance on this or look into making it support both delegates and more traditional winforms style node creation (`treeView1.Nodes.Add(new TreeNode("Text for new node"));`)
 
> Lastly, yes, we would need an explicit statement from the original author that he is willing to relinese the code under the MIT license, and his name and copyright under the MIT license needs to be attached to the patch, as this currently stands, this code should not be merged due to the GPLv3 license.


<details>
  <summary>This is my current correspondence with the original author (Phillip Piper) - phillip.piper@gmail.com</summary>
  
Hi Thomas,

Thanks for getting in touch.

That is fine. Go ahead with my blessing, and may you have great success :) 

Regards,
Phillip
-----------------------------------------------------------------
Phillip Piper                             phillip.piper@gmail.com
A man's life does not consist in the abundance of his possessions


On Mon, 23 Nov 2020 at 23:27, Thomas Nind (Staff) <T.Z.Nind@dundee.ac.uk> wrote:
Hey,

I have been using Object List View for several years and am a big fan.  Recently I got involved in writing some Terminal (Console) user interface controls for the MIT licensed, open source framework Terminal.Gui (https://github.com/migueldeicaza/gui.cs)

I have implemented a Console based tree view and based it on the same design pattern you have used (ChildrenGetterDelegate, CanExpandGetterDelegate etc).  

The source code for this fork is:
https://github.com/tznind/gui.cs/compare/master...tznind:tree-view

I have credited your framework in the source code.  I am just writing to check that you are ok with me creating this derivative work and whether it is ok to use it under the MIT license?  If not I can create a separate package and license it under GPLv3 as an extension package of Terminal.Gui.

Thanks,

Thomas Nind
</details>

Based on his positive response I don't think it would be a problem to get a more formal statement if needed.

Terminal.Gui/Views/TreeView.cs
```diff
@@ -0,0 +1,703 @@
+// This code is based on http://objectlistview.sourceforge.net (GPLv3 tree/list controls by phillip.piper@gmail.com)
```
tig, line 1: Please change this comment to explicitly note that Phillip has licensed this code under the MIT lic and not GPLv3.

tznind: Have updated, let me know if you want me to adjust the phrasing.

tznind:
I have added an option `ShowBranchLines` and a simpler example to UICatalog

image

I have also updated the default child getter delegate to support more traditional node addition:

```csharp
var tree = new TreeView();

var root = new TreeNode("Root");
root.Children.Add(new TreeNode("Leaf1"));
root.Children.Add(new TreeNode("Leaf2"));

tree.AddObject(root);
```

The new default delegate doesn't change any of the previous functionality.  The user can still define their own delegate and supply any object types they want into the tree at runtime as nodes are expanded based on the delegate.  

I think it makes sense to have both options, a dynamic one where you don't want to have to build the entire tree before showing it and a more traditional one based on explicit nodes.

I could update the tree to use generics if that would be useful (for where all tree nodes have a shared base class Type)?  That would allow defining `TreeView<TreeNode>` or `TreeView<FileSystemInfo>` or just `TreeView<object>`.

BDisp:
A generic constructor (`TreeView<T>`) would be nice too :-)

migueldeicaza:
I like where this is going, some additional bits of feedback:

- I think you can remove the [+] and [-] symbols.
- I like what you have done with the "storage" api, would love to see an interface-based approach that can supply the data on demand.
- I think that we should show the whole line as selected like this in mc:

image

tznind:
@migueldeicaza thanks for the further feedback.

Per your advice (let me know if I have misunderstood it), I have removed the delegate (`CanExpandGetter`) in place of a new interface `ITreeBuilder`.  There is an abstract implementation and 2 out of the box implementations (one for `ITreeNode` and one for Delegates).

The old method is still alive but now looks like this:

```csharp
// Creates a tree with the default builder (expects tree to be filled with `TreeNode` objects)
var tree1 = new TreeView();

// Same as above (TreeNodeBuilder is the default builder when using blank constructor)
var tree2 = new TreeView(new TreeNodeBuilder());

// Creates a tree that expects Cars and Factories to be in the tree and explains that with delegates
var tree3 = new TreeView(new DelegateTreeBuilder((s)=> s is Factory f ? f.Cars: null));

// Creates a tree that asks the users custom class what the children are for any given object
var tree4 = new TreeView(new MyCarFactoryBuilder());
```

Implementing `ITreeBuilder` looks like this:

```csharp
public class MyCarFactoryBuilder : TreeBuilder {
	public MyCarFactoryBuilder():base(false /*does not support CanExpand*/)
	{
	}
	public override IEnumerable<object> GetChildren (object model)
	{
		if(model is Factory f)
			return f.Cars;
		
		return Enumerable.Empty<object>();
	}
}
```

I have updated the selection to cover the full width of the `Bounds` and spaced out the branch lines a bit.  I removed the [-] symbol and have made the [+] optional (using `ShowExpandableSymbol`) since it is the only way for the user to know whether a branch can be expanded without moving to it and trying to expand.  For example to tell the difference between folders and files (see below).

image

I've not made TreeView into a generic class yet.  It could avoid a lot of casting but trees often have mixed types (e.g. `FileInfo` and `DirectoryInfo`) even when they have a shared base class (`FileSystemInfo`) so not sure about whether it will make things easier or harder for end users of the API.

BDisp:
You could create a generic constructor and let end users deals with it if they have methods that can be overridden.
Another suggestion is if it's possible to have the plus signal with a configurable color or an inverted color to stay highlighted, or is it a silly suggestion?

BDisp:
@tznind sorry for I answer on the other place because I thought was you are asking how to print the plus sign.

`"@tznind of course you can. You need to deal with the Rune.ColumnWidth to check if the char occupies two columns and consider it as only one column, on the Redraw method. See what I did in the TextField.  But I was talking about only change the Attribute color when you print the '+' char :-)"`

My apologies again.

BDisp:
@tznind I think is better to have the minus sign instead the horizontal line because it overlaps over the character to the right.

BDisp:
@tznind another great feature would be add mouse support too :-)

tznind:
Ive made Expand/Collapse symbols configurable and with color options

image

BDisp:
> 
> 
> @tznind I think is better to have the minus sign instead the horizontal line because it overlaps over the character to the right.

Forget this because I was using the `KaiTi `font to test a user issue  and I forgot the restore back to the default, sorry.

BDisp:
> Is there a way to figure out the background color of the control e.g.
> 
> `new Terminal.Gui.Attribute(Color.Green,Win.ColorScheme.Normal.Background)`
> (Background property is private in this Attribute)

@tznind  you must rebase your branch to get that feature.

BDisp:
Perhaps you could use the checked feature menu to get the menu options checked or unchecked. See the `Diagnostics `menu.

BDisp:
@tznind you must set a default `ColorScheme `and when the object is being created set the focus for the default object instead  waiting the a key press.

BDisp:
The `ColorScheme `exception only happens if we choose `"Clear Objects"` and creating another `Tree `instance. For e.g. choosing `"File Tree"`, `"Clear Objects"` and `"Simple Tree"` or choosing `"Simple Tree"`, `"Clear Objects"` and `"File Tree"`.

tznind:
> 
> @tznind you must set a default `ColorScheme `and when the object is being created set the focus for the default object instead waiting for a key press.

Not sure what you mean... Do you mean I need a ColorScheme for ExpandableSymbolColor not just `Attribute` or something else?

I noticed that when rapidly removing and adding the trees I was getting various Exceptions in drawing.  I changed the Scenario to get rid of switching and instead added both trees to the UI one above the other (this also fixed a unit test around all controls being disposed in Scenarios).  But I appreciate this just avoided the issue rather than fixing it!

BDisp:
`Attribute `it's enough but the problem perhaps is due that when disposing the  trees nodes the `TreeView `Attribute is also disposed and when you create another different instance of tree nodes the `Attribute `is null in the `Draw `method. So you must ensure that the `Attribute `is not null before being drawing.

BDisp:
About the focus is setting the focused for the first object of the new instance finished created.

BDisp:
I forgot to tell that normally we define the a default value for the `Colors  `that a `View` (`TopLevel, Base, Dialog, Menu or Error`) could have and the user might change the color layout as desired. So implicitly setting colors prevent the end user could costumizing colors.

tznind:
Ok, I have updated to use color scheme instead of explicit colors.  This should ensure that when turned on the `ColorExpandSymbol` setting never looks too odd.  I also added an `InvertExpandSymbolColors` option as you mentioned that earlier.

Both options are off by default.

In the scenario the Settings properly use checked states for feedback on what is turned on/off.

tznind:
I have added mouse support:

- Scroll wheel scrolls view up/down without changing the current selection
- When expansion symbols are on click it will expand/collapse without changing selection
- Left clicking (Button1) on a line will select it
- Left clicking on a line which is already selected will toggle expand/collapse (assuming it is not a leaf).

tree-view-mouse

BDisp:
Very great job. Thanks. Only some observations:
If it was showned a large item before it isn't clearing the remaining item at right. So the SetNeededDisplay have to clear the entire width before showing the news items.
The menu will be more appeal if adding separators between normal item, checked and radio buttons. See the "Diagnostics" menu example.

tznind:
> Very great job. Thanks. Only some observations:

Thanks

> If it was showned a large item before it isn't clearing the remaining item at right. So the SetNeededDisplay have to clear the entire width before showing the news items.

Is this in the video gif or something you are getting when you run it locally? I do try to fill up the available width when rendering rows but it's possible my calculation is wrong or it is some character set related issue?

https://github.com/migueldeicaza/gui.cs/blob/4795f321bf7be902e95617ec33dd6a76ccf2ca14/Terminal.Gui/Views/TreeView.cs#L902-L905

Can you make a screenshot of the issue so I can try to reproduce and fix it?

> The menu will be more appeal if adding separators between normal item, checked and radio buttons. See the "Diagnostics" menu example.

I have added these, is it just `null` in the menu items array? seems to work (https://github.com/migueldeicaza/gui.cs/pull/1022/commits/4795f321bf7be902e95617ec33dd6a76ccf2ca14)

BDisp:
It was locally. Here is the video. Yes null is the trick, thanks.

treeview-remaining

tznind:
Thanks, looks like I am spilling long strings over the edge of the control.  The top tree is deliberately thinner and should only be filling 40 width.

https://github.com/migueldeicaza/gui.cs/pull/1022/files#diff-870237d4231a038f4e832f07728c8555c5c9a89cbb6d7505d5740b190c5377bbR72-R77

BDisp:
I haven't seen already if you are dealing with unicode characters, which will also beyond the frame width if they have a `ColumnWidth `geater than 1.
Another thing you have to deal is with the PositionCursor method to put the cursor at the beginner of the selected item.

BDisp:
You can toggle the `File Tree` to fill the `Window `frame and the `Node Tree` with the width of the `File Tree`, since the `Node Tree` is smaller.

tznind:
> 
> 
> I haven't seen already if you are dealing with unicode characters, which will also redraw beyond the frame width if they have a `ColumnWidth `greater than 1. See the `TextField `and the `TextView`.

Thanks, what is the difference between `Rune.ColumnWidth` and `Rune.RuneLen`?  Does this look along the right lines:

https://github.com/migueldeicaza/gui.cs/pull/1022/commits/aef698ac1f76c3669f781deea95694bf9bb97e0e

BDisp:
> Thanks, what is the difference between `Rune.ColumnWidth` and `Rune.RuneLen`? Does this look along the right lines:

You can see here https://github.com/migueldeicaza/nstack.
Summing up:
`Rune.ColumnWidth` is the number of column positions of a wide-character code. This is used to measure runes as displayed by text-based terminals.The width in columns, 0 if the argument is the null character, -1 if the value is not printable, otherwise the number of columns that the rune occupies.
`Rune.RuneLen` is the number of bytes required to encode the rune. The length, or -1 if the rune is not a valid value to encode in `UTF-8`. So a rune which have a `Rune.ColumnWidth` equal to 1 could be have a `Rune.RuneLen` from 1 up to 4 bytes. I use this for copy and paste in the `TextField`.

tig:
This is just fantastic work.

For such an important control, I think we need to ensure we have at least three use-cases coded up and working before we merge. 

I was hoping to find the time to implement a PowerShell object browser with this, complimenting `out-consolegridview`. E.g. see this https://social.technet.microsoft.com/wiki/contents/articles/28651.how-to-create-and-use-an-object-browser-in-powershell.aspx. I have, sadly, not found the time. But maybe this weekend ;-)

What are the current 'samples' that illustrate the power of this control?

tznind:
Thanks for taking the time to look over the PR.  

I think the most powerful demonstration of the control is in the TreeViewFileSystem Scenario in UICatalog.  This shows how the tree 'builds itself' as you expand/collapse nodes so that you don't have to enumerate the entire file system before you show your control.

I have added some simpler examples including describing how to use the more traditional 'build it up front' approach where you just declare `TreeNode` instances:

https://github.com/migueldeicaza/gui.cs/blob/bbfb6b28c0363cf9f56f6632ddaf35348ee8ed11/docfx/articles/treeview.md

Was this the kind of thing you were thinking of? I can add another Scenario which is more about adding/removing branches and refreshing etc.

Alternatively I can beef up TreeViewFileSystem perhaps to display info about the currently selected file and have a button to refresh the selected folder etc?

BDisp:
Wow fantastic.  @tznind does `TreeView `and the `TableView `provides public properties for the `Top `and `Left `(or another name you wish) with read and write access, to allowing using with the new `ScrollBarView `version? Also providing information readonly about the number of lines and colums to allow interact with the `ScrollBarView`. Of course if the view doesn't scroll horizontally the left and number of colums is unecessary.

tznind:
Thanks for spotting that @BDisp

I have added `ContentHeight` (readonly) and implemented the new `ScrollBarView` in the Scenario example `TreeViewFileSystem`.   The scroll property in `TreeView` is called `ScrollOffset` (although happy to change that if there is a convention).

Currently it doesn't support horizontal scrolling but it is something I could look into.

BDisp:
The name is indifferent as long as they are understandable. I think it would be interesting to be able to scroll horizontally so that we can fully view the item. Sort of like the vertical but dealing with the columns. If you are going to implement it, then I would suggest changing the `ScrollOffset `to, for example, `ScrollHeightOffset `or whatever you think is more appropriate. Thank you very much.

BDisp:
@tznind `ContentHeight `is not passing the correct value when clicking `+/-`. It only passes the correct value when it is rolling. Therefore, before redrawn the TreeView the `ContentHeight `value must be the expanded value when clicking `+` and the colapsed value when clicking `-`.

tznind:
Thanks for spotting that issue.  I have changed the ContentHeight / Width properties to be computed when asked for rather than cached.

I have added horizontal scrolling capability and the relevant scrollbar.  But I seem to be loosing the rendering of the last line of my control and its like my control has lost focus too since pressing down again wont scroll down the last bit of the control:

overdrawn

Any ideas?

BDisp:
Yes I know what it's happens. Like on the scroll bar implementation in the `ComboBox `I had to add one more in the height size to get the selected item visible.

BDisp:
Sorry, I forgot to tell you that that happens also to `TextView `and the `ListView `when scrolling by pressing a key, because they are coded to go to the last line unlike the mouse. I think the best solution is creating in the host something that if it is hosted must scroll one more line.

BDisp:
In the `TextView `and `ListView `I allowed the top and the left go to the lines count minus one and columns count minus one, respectively, but I didn't the same for key press. That why theree are differences between scrolling with mouse and with key. Things to improve yet.

BDisp:
The reason you don't see this happening is because the `ScrollBarView `has the `KeepContentAlwaysInViewport `property as true. If you change it to false you will see that the `TextView `and `ListView `scrolls beyond the vertical and horizontal limits.

tznind:
Thanks for the guidance.  I have added an optional flag `Style.LeaveLastRow` which will ensure the selection never goes to the last line e.g. when  using keyboard navigation to change the selection.

TreeView has 2 ways of changing the scroll offset:

- Directly e.g. `tree.ScrollOffsetVertical += 10`
- Indirectly when the selected object changes e.g. `AdjustSelection(10)`

When directly editing the scroll offset e.g. when the mouse wheel moves then the selected object does not matter and it scrolls like you would expect.

Changing the selected object (e.g. by pressing up/down) will indirectly change the scroll offset (if it is not already visible) to ensure that the new selection is always within the visible scroll area.  When using `LeaveLastRow` this will not consider the last row as 'visible' when doing this calculation.

BDisp:
I'm happy to be able to help you. I really had thought of something like that but for both orientations. For example, in `TextView` using the mouse, I can scroll to the limit passed by the vertical and horizontal sizes, but not by using the `Up / Down` or `Page Up / Page Down` key. But in `TextView` it needs both a row and an additional column so that we can continue editing with the content visible on the screen. Therefore, the hosts will have to provide these additional properties. Thank you for putting this into practice.

tznind:
I have added another scenario as requested.  For this one I used Reflection to enumerate all Types, Methods etc as they are expanded.  You can see how some nodes look expandable but turn out not to be when expanded that is because of the CanExpand delegate (which is optional) and allows the system to dive into very deep tree structures and only discover children as needed.

classexplorer

Theres some things this use case has highlighted to me like:

- Ordering
- Preserving order/selection during refresh (see when you toggle public/private) in view

So I will look into streamlining the control a bit for that.

tznind:
interactive

I have added a third use case using the `TreeView` to add, rename, delete nodes interactively.

In order to make using the control as similar to winforms as possible I have added `Tag` field to TreeNode so that if desired you can just manage TreeNodes and still reference your own Model objects through that (without having to implement `ITreeBuilder` or `ITreeNode` etc).

This brings the number of use cases to 3:

- File System Viewer
- Class Explorer (Reflection based exploration of public/private members of loaded assemblies)
- Interactive (Using control like you would in winforms just adding TreeNode instances as and when required)

tznind:
I have added `AllowLetterBasedNavigation` setting that replicates windows tree behavior when you press a letter or digit e.g. 'N' (cycles selection to next branch beginning with N):

letterBasedNavigation

migueldeicaza:
This looks fantastic!

Terminal.Gui/Views/TreeView.cs
```diff
@@ -0,0 +1,1836 @@
+// This code is based on http://objectlistview.sourceforge.net (GPLv3 tree/list controls by phillip.piper@gmail.com).  Phillip has explicitly granted permission for his design and code to be used in this library under the MIT license.
```
tig, line 1: Totally minor, but can you make this comment more readable by making it fit multiple lines (so horiz scrolling isn't needed)?

Terminal.Gui/Views/TreeView.cs
```diff
@@ -0,0 +1,1836 @@
...
+		/// </summary>
+		/// <value></value>
+		public bool LeaveLastRow {get;set;}
+
+	}
+	
+	/// <summary>
+	/// Hierarchical tree view with expandable branches.  Branch objects are dynamically determined when expanded using a user defined <see cref="ITreeBuilder{T}"/>
+	/// </summary>
+	public class TreeView<T> : View, ITreeView where T:class
```
tig, line 292: Why `where T: class` and not `where T: TreeNode`?  I am not 100% up to speed on generic type constraints yet, but it seems that you are missing an opportunity to ensure type safety. 


What am I missing?

Terminal.Gui/Views/TreeView.cs
```diff
@@ -0,0 +1,1836 @@
...
+		/// <param name="oldValue"></param>
+		/// <param name="newValue"></param>
+		public SelectionChangedEventArgs(TreeView<T> tree, T oldValue, T newValue)
+		{
+			Tree = tree;
+			OldValue = oldValue;
+			NewValue = newValue;
+		}
+	}
+}
```
tig, line 1836: You have my permission to break this class into multiple files. This is a lot of code to navigate in one file. 

Consider which if any of the classes might be generic and useful for purposes other than `TreeView`.

tig (changes requested):
I'm still diving into this, but figured I'd provide thee feedback/questions I have so far.

This is really fantastic and I want to get it merged in as soon as possible.

Please add a relnotes update for this in `Terminal.Gui.csproj`. Thanks.

tig (changes requested):
I don't see any code in Scenarios that exercises `ObjectActivated` `ObjectActivationKey`.

How do we know it works? 

I know it does because I added this to `ClassExplorer`:


image

Terminal.Gui/Views/TreeView.cs
tznind: Thanks for diving into this so deeply!

The generic `TreeView<T>` supports any Type (e.g. `FileSystemInfo`) but requires that you specify an `ITreeBuilder` that shows how the objects relate to one another (how to determine children).

The non generic `TreeView` supports only `ITreeNode` objects and has a default `ITreeBuilder` already set up on it (`TreeNodeBuilder`).

Internally the TreeView stores the state of the tree as `Branch<T>`.  At the moment this is a private class and not visible to the API user.

The purpose of the regular `TreeView` is to mimic as closely as possible the WinForms tree view (create all nodes up front and then add them to the tree).

The purpose of the generic `TreeView<T>` is two fold
- To support hosting Typed objects e.g. if all your tree objects are `FileSystemInfo` then that is the Type that arrives in events (without having to change your class definition e.g. to implement an interface)
- To support late loading of children (e.g. until the user clicks expand on a node).  This is essential when fetching children may be relatively expensive e.g. listing database tables/columns or contacting a REST API.

When children can change over time (e.g. on a file system) the tree view can update itself to get a new authorative answer from the `ITreeBuilder` through methods like:

- RebuildTree
- RefreshObject

I have created an article (included in the PR) that describes from a user perspective how to create either type of tree:

docfx/articles/treeview.md

Terminal.Gui/Views/TreeView.cs
```diff
@@ -0,0 +1,1836 @@
...
+					AdjustSelection(0);
+					SetNeedsDisplay();
+				}
+			}
+		}
+
+		/// <summary>
+		/// Changes the <see cref="SelectedObject"/> to the first root object and resets the <see cref="ScrollOffsetVertical"/> to 0
+		/// </summary>
+		public void GoToFirst()
```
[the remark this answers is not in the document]
tznind: I've added a new method `GoTo(T toSelect)` in https://github.com/migueldeicaza/gui.cs/pull/1022/commits/db322b27ef40d0cd881440be3573781f1d9964cd to make life easier on users.

Terminal.Gui/Views/TreeView.cs
tznind: I've split it up into `TreeBuilder.cs`, `TreeStyle.cs`, `TreeNode.cs` and `TreeView.cs`.  Currently the interfaces and implementations are in the same file.  Let me know if you want me to go further and put every class/interface in it's own file.

Currently `Branch<T>` is not public so I have left it in `TreeView.cs` but I can move it out too if required.

tig (approved):

PR #1031, opened 2020-12-04
Fixes #1030. Getting colors from the Attributes of a ColorScheme.

### LINKED ISSUE #1030: Getting colors from the Attributes of a ColorScheme
So in my app, I wanted to make a function that would dynamically change the foreground color of one of the Attributes of a View's ColorScheme while keeping the background color the same as in the View's SuperView. Unfortunately, you cannot modifiy only one of the colors and need to pass both foreground and background colors to the Attribute via Application.Driver.MakeAttribute.

That wouldn't be an issue if I could just get the colors from the Attributes of the ColorScheme that the SuperView in question uses and pass the background color as an argument in Application.Driver.MakeAttribute. But as it is right now, there's no way to extract the colors from an already existing Attribute (or at least I couldn't find it).

Any advice on how to solve this?

BDisp:


tig (approved):
nice

PR #1047, opened 2020-12-22
Fixes #93. Audit TextView just like we did TextField to ensure proper treatment of Unicode

### LINKED ISSUE #93: Audit TextView just like we did TextField to ensure proper treatment of Unicode

BDisp:


Terminal.Gui/Views/TextView.cs
```diff
@@ -817,25 +849,43 @@ void Adjust ()
...
+			}
+			return (w, h);
 		}
 
 		/// <summary>
 		/// Will scroll the <see cref="TextView"/> to display the specified row at the top
 		/// </summary>
 		/// <param name="row">Row that should be displayed at the top, if the value is negative it will be reset to zero</param>
-		public void ScrollTo (int row)
+		public void ScrollToRow (int row)
```
tig, line 869: This is a breaking API change. Do you have any data on how many clients have used this API?

Options for addressing:
- Update the relnotes to indicate a breaking change.
- Ensure API docs indicate the API changed
- Keep the old API in place, but document it as deprecated and point to new API

tig (changes requested):

Terminal.Gui/Views/TextView.cs
BDisp: Ok I fixed it without breaking anything. Thanks.

tig (approved):

PR #1063, opened 2020-12-28
Added unhandled exception handling in Application.Run

HermanSchoenfeld:


Terminal.Gui/Core/Application.cs
```diff
@@ -682,21 +682,41 @@ public static void Run<T> () where T : Toplevel, new()
...
+					RunLoop (runToken);
+					End (runToken);
+				} 
+				catch (Exception error) 
+				{
+					if (errorHandler == null)
+						throw;
+					resume = errorHandler (error);
+				}
+				if (resume)
```
tig, line 708: Very debatable, but I think this code would be more clear if it was

```
var resume = true;
while (resume)
{
   try
   {
       resume = false;
       ...
   }
   catch
   {
       if (errorHandler == null)
       {
             throw;
       }
       resume = errorHandler(error);
   }
}
```

At the very least, please put brackets around the `throw;` per our style guidelines (always use enclosing brackets for `if`).

tig (changes requested):
This is Very Nice!

Please see my comment.

Terminal.Gui/Core/Application.cs
HermanSchoenfeld: Done, thanks for feedback.

tig (approved):

Issue #1065, opened 2020-12-28
UI Catalog is not filling console (at least on Windows)


tig:
VS 2019 debug:
![image](https://user-images.githubusercontent.com/585482/103229943-fbdb5500-48e8-11eb-8452-e389cd956784.png)

Cmd.exe
![image](https://user-images.githubusercontent.com/585482/103229853-ca628980-48e8-11eb-9104-eaaaff418886.png)

conemu:
![image](https://user-images.githubusercontent.com/585482/103229990-17466000-48e9-11eb-99c8-0e76ac843f60.png)

Windows Terminal:
(hangs)

BDisp:
In `cmd.exe` and `ConEmu `I confirm. The strange thing is happening to you in the VS 2019 debug which is the one I use the most and in my is filling well. What code page do you use? I use 65001 (UTF-8). Unfortunately, WindowPlacement does not have column and row dimensions and I have to make the conversion. The others consoles may being calculating other dimensions values.
Maybe is better using Application.HeightAsBuffer=true as default?

tig:
I haven't run a new build in a while. This used to work fine. We need to figure out what broke and revert it.

The worst part is the app doesn't show at all under Windows Terminal. 

![](https://i.imgur.com/AH5cNy3.gif)

tig:
I'm really confused now. Why did you change things so the system console was not default?

I just saw this code 

![image](https://user-images.githubusercontent.com/585482/103246550-b898da80-4918-11eb-8a87-f946fa3203e2.png)

If I run UICatalog with -usc It works fine AND *MOUSE WORKS*!
![](https://i.imgur.com/hfNVX6k.gif)

tig:
Now I can't get the incorrect size thing to reproduce. Ugh.

BDisp:
I didn't make the use system console as default because it works slower than `WindowsDriver `and `CursesDriver`, but at least the `NetDriver `works better on others consoles with exception on ConEmu.

BDisp:
@tig if you use the arg -usc it use the `NetDriver `on all SO and the mouse works like on the Linux using sequence escapes.

tig:
> I didn't make the use system console as default because it works slower than `WindowsDriver `and `CursesDriver`, but at least the `NetDriver `works better on others consoles with exception on ConEmu.

I confused myself. Nevermind. I was thinking System Console == WindowsDriver/CursesDriver.

[...]

PR #1066, opened 2020-12-29
A provisional solution to the #1065 issue. Setting HeightAsBuffer as true by default.

BDisp:


BDisp:
@tig I think this PR improves `WindowsDriver `a little more than it is right now. Please check when you have time. Thanks.

tig:
I just tested this and I don't think it is right.

At some point Terminal.Gui with WidnowsDriver sized correctly and worked fine on all consoles I used on Windows (cmd.exe, VS debug, Windows Terminal, and ConEmu). Somewhere along the line a change was introduced which broke this. 

Trying to calculate the size of the console by asking for the console font used is going to be very error-prone. The current PR doesn't work correctly. 

Windows Terminal:
(sometimes it is right, but only if the window is sized just right)
image

cmd.exe:
image

Conemu:
image

VS Debug:
image

What was the change that broke this? @BDisp do you know precisely what changed that broke this?

BDisp:
> @BDisp can you back that commit out? What else will break if you do?

@tig this is still not in perfect working order but I think it is better than it is currently. Removing this commit will stop the console's resizing and the vertical scroll bar will always appear on the console. In my opinion we will have to take the font size into account when calculating the size although it can sometimes be erroneous, but it will be better than using fixed constants. If someone more knowledgeable about the `Windows Native APIs` could help, that would be great. In my opinion if we want to have the possibility of having the console resized without always appearing the vertical scroll bar, we could merge this PR and wait for a solution, because the console can work as it used to do if we set `HeightAsBuffer` to true.

tig:
> > @BDisp can you back that commit out? What else will break if you do?
> 
> @tig this is still not in perfect working order but I think it is better than it is currently. Removing this commit will stop the console's resizing and the vertical scroll bar will always appear on the console. In my opinion we will have to take the font size into account when calculating the size although it can sometimes be erroneous, but it will be better than using fixed constants. If someone more knowledgeable about the `Windows Native APIs` could help, that would be great. In my opinion if we want to have the possibility of having the console resized without always appearing the vertical scroll bar, we could merge this PR and wait for a solution, because the console can work as it used to do if we set `HeightAsBuffer` to true.

Every Windows console calculates font sizes differently. There is no way to ask the various consoles how they calculate font size. Thus it will be impossible to make Terminal.Gui always work on all consoles this way. I think living with the vertical scroll bar is a far better solution. 

Previously resizing the console horizontally (shrinking or growing) worked well. 

Vertical resizing worked for growing but not shrinking, except on ConEmu which worked both ways. This means it is possible for consoles to do the right thing. We may never get VS debug or cmd.exe to do the right thing, but I'm willing to bet we can convince the Windows Terminal team to do what Conemu does.

tig:
Thank you!

PR #1075, opened 2021-01-06
Fixes #1071. Toplevel.GetTopLevelSubviews (). Changed from HashSet<View> to IList<View>.

### LINKED ISSUE #1071: Toplevel.GetTopLevelSubviews ()
This method is intended to return a list in the order it was found (I am guessing from the requirement to get an "isForward") parameter.   This is called during the focus processing systems, and if the value is true, the items are reversed.

HashSet does not seem to guarantee the order in which the elements were added.

Pinging, as this was part of a bug fix, and not sure what the intent was, but this does look strange

BDisp:


tig (commented):
Do we have unit tests that prove that this won't break anything?

BDisp:
Unfortunately the feature to navigate through all `Application.Top`'s subviews is not working on `Unix `because there are no way to get if the `Control `and `Shift `keys are pressed together with the `Tab `key.

BDisp:
Here how `Ctrl+Tab` and `Shift+Ctrl+Tab` works on the `WindowsAndFrameViews `scenario:

toplevels-navigation

tig:
Nice work!

PR #1090, opened 2021-01-18
Add copyable instructions for .NET Core users

pcgeek86:
Adding a simple copy/paste command to install Terminal.Gui into .NET Core projects


tig (commented):
Thank you!

Issue #905, opened 2020-09-26
Cursor blinks on listviews etc...


tig:
I do not think the cursor should blink on UI elements that already show selected status.

I thought this had been fixed/changed but it looks like a recent PR changed it back.

Does anyone disagree that the blinking cursor should only show in edit boxes?

BDisp:
I agree but I think it isn't possible hiding the cursor but only hacking with the same background color.

tig:
IIRC, there's a driver API to hide/show cursor...

Issue #280, opened 2019-10-26
ListView renders cursor


kzu:
Even if there's nothing to type in a listview :)

![Cursor](https://user-images.githubusercontent.com/169707/67626520-6216a480-f81a-11e9-9d89-d218183ef14c.gif)

BDisp:
I really don't know if the cursor can be hidden in the `Gui`, without overshadowing the color.

tig:
Fixed somewhere along the line.

PR #1102, opened 2021-01-28
Cursor shape and visibility

G-Freart:
Hello,

This pull request contains a fix for issues #905 and #280

It has been tested under Linux (CentOS 8 Stream, Ubuntu 20.04) in Gnome Terminal and also in real TTY.
It works also under Windows. 

Under XTerm, the amount of cursor possibilities are more important than under TTY. The documentation 
related to the CursorVisibility enumeration describe it.

In order to test the different possibility, a dynamic menu will appear within the Editor Scenario.

I was not able to make it working with System.Console. 
May be due the set of customization this code already have. 
Even, if it was working which is not the case, the 'native' method work only under Windows.

As i don't have a MacOS i could not test this PR under this Operating Systems.

Best Regards,
Gilles Freart

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -700,10 +700,23 @@ namespace Terminal.Gui {
 			// We are setting Invisible as default so we could ignore XTerm DECSUSR setting
 			//
 			switch (Curses.curs_set (0)) {
-				case 0:		currentCursorVisibility = initialCursorVisibility = CursorVisibility.Invisible;	break;
-				case 1:		currentCursorVisibility = initialCursorVisibility = CursorVisibility.Underline;	Curses.curs_set (1); break;
-				case 2:		currentCursorVisibility = initialCursorVisibility = CursorVisibility.Box;		Curses.curs_set (2); break;
-				default:	currentCursorVisibility = initialCursorVisibility = null;						break;
+				case 0:		
+					currentCursorVisibility = initialCursorVisibility = CursorVisibility.Invisible;	
+					break;
+
+				case 1:		
+					currentCursorVisibility = initialCursorVisibility = CursorVisibility.Underline;	
+					Curses.curs_set (1); 
+					break;
+
+				case 2:		
+					currentCursorVisibility = initialCursorVisibility = CursorVisibility.Box;		
+					Curses.curs_set (2); 
+					break;
+
+				default:	
+					currentCursorVisibility = initialCursorVisibility = null;						
+					break;
 			}
 
 			Curses.raw ();
@@ -891,9 +904,8 @@ namespace Terminal.Gui {
 		{
 			visibility = CursorVisibility.Invisible;
 
-			if (!currentCursorVisibility.HasValue) {
+			if (!currentCursorVisibility.HasValue)
 				return false;
-			}
 
 			visibility = currentCursorVisibility.Value;
 
@@ -909,8 +921,7 @@ namespace Terminal.Gui {
 
 			Curses.curs_set (((int) visibility >> 16) & 0x000000FF);
 
-			if (visibility != CursorVisibility.Invisible)
-			{
+			if (visibility != CursorVisibility.Invisible) {
 				Console.Out.Write ("\x1b[{0} q", ((int) visibility >> 24) & 0xFF);
 				Console.Out.Flush ();
 			}
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -886,6 +886,7 @@ namespace Terminal.Gui {
 			return currentAttribute;
 		}
 
+		/// <inheritdoc/>
 		public override bool GetCursorVisibility (out CursorVisibility visibility)
 		{
 			visibility = CursorVisibility.Invisible;
@@ -899,6 +900,7 @@ namespace Terminal.Gui {
 			return true;
 		}
 
+		/// <inheritdoc/>
 		public override bool SetCursorVisibility (CursorVisibility visibility)
 		{
 			if (initialCursorVisibility.HasValue == false) {
@@ -918,6 +920,7 @@ namespace Terminal.Gui {
 			return true;
 		}
 
+		/// <inheritdoc/>
 		public override bool EnsureCursorVisibility ()
 		{
 			return false;
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -915,9 +915,8 @@ namespace Terminal.Gui {
 		/// <inheritdoc/>
 		public override bool SetCursorVisibility (CursorVisibility visibility)
 		{
-			if (initialCursorVisibility.HasValue == false) {
+			if (initialCursorVisibility.HasValue == false) 
 				return false;
-			}
 
 			Curses.curs_set (((int) visibility >> 16) & 0x000000FF);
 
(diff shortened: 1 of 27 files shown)


BDisp:
Really good job. Thanks. I think that selecting the entire character initially can cause the impression that the capslock is on and confuse the user. Wouldn't it be possible to put a narrower width if the capslock is turned off?
In the case of `NetDriver`, perhaps the best solution would be through `Virtual Terminal Sequences` but he is not yet ready to process it and would be grateful for all the possible help for the possibility of improving it. Unfortunately, `System.Console` still does not work the same way on both systems (`Windows` and `Unix`) and I had to activate the escape sequences by creating the console with native `Windows APIs` and some of the `System.Console APIs` are not compatible with `Unix`. As a last resort, for Windows a possible implementation of cursor visibility would perhaps be the solution you implemented on `WindowsDriver` but for `Unix` it would have to be escape sequences and some native Unix library, other than `NCurses`. But if there is a common solution it would be better. Any idea?

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -689,6 +695,17 @@ public override void Init (Action terminalResized)
...
+
+			// 
+			// We are setting Invisible as default so we could ignore XTerm DECSUSR setting
+			//
+			switch (Curses.curs_set (0)) {
+				case 0:		currentCursorVisibility = initialCursorVisibility = CursorVisibility.Invisible;	break;
+				case 1:		currentCursorVisibility = initialCursorVisibility = CursorVisibility.Underline;	Curses.curs_set (1); break;
+				case 2:		currentCursorVisibility = initialCursorVisibility = CursorVisibility.Box;		Curses.curs_set (2); break;
+				default:	currentCursorVisibility = initialCursorVisibility = null;						break;
+			}
```
migueldeicaza, line 707: Let us follow the coding style, statements should be in different lines, and we should remove the additional space

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -868,6 +885,46 @@ public override Attribute GetAttribute ()
 		{
 			return currentAttribute;
 		}
+
+		/// <inheritdoc/>
+		public override bool GetCursorVisibility (out CursorVisibility visibility)
+		{
+			visibility = CursorVisibility.Invisible;
+
+			if (!currentCursorVisibility.HasValue) {
```
migueldeicaza, line 894: Style - generally, in this codebase I do not use braces for single statements.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -868,6 +885,46 @@ public override Attribute GetAttribute ()
...
+				return false;
+			}
+
+			Curses.curs_set (((int) visibility >> 16) & 0x000000FF);
+
+			if (visibility != CursorVisibility.Invisible)
+			{
+				Console.Out.Write ("\x1b[{0} q", ((int) visibility >> 24) & 0xFF);
+				Console.Out.Flush ();
+			}
```
migueldeicaza, line 916: Braces should open on the same line as the "if"

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -100,8 +107,76 @@ public bool SetCursorPosition (Coord position)
...
+			if (initialCursorVisibility.HasValue == false && GetCursorVisibility (out CursorVisibility visibility)) {
+				initialCursorVisibility = visibility;
+			}
+		}
+
+		public bool GetCursorVisibility (out CursorVisibility visibility)
+		{
+			if (!GetConsoleCursorInfo (ScreenBuffer, out ConsoleCursorInfo info)) {
+
+				var err = Marshal.GetLastWin32Error ();
```
migueldeicaza, line 121: Let us eliminate white space lines

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -100,8 +107,76 @@ public bool SetCursorPosition (Coord position)
...
+
+				visibility = Gui.CursorVisibility.Default;
+
+				return false;
+			}
+
+			if (!info.bVisible)        visibility = CursorVisibility.Invisible;
+			else if (info.dwSize > 50) visibility = CursorVisibility.Box;
+			else                       visibility = CursorVisibility.Underline;
+
```
migueldeicaza, line 135: I see what is going on here, but let us stick to the coding style we use here, each`visibility = XX` needs to go on its own line.

Terminal.Gui/Views/TextField.cs
```diff
@@ -866,6 +866,33 @@ public virtual TextChangingEventArgs OnTextChanging (ustring newText)
...
 		}
+
+
+		private CursorVisibility wishedCursorVisibility = CursorVisibility.BoxFix;
+
+		/// <summary>
+		/// Get / Set the wished cursor when the field is focused
+		/// </summary>
+		public CursorVisibility WishedCursorVisibility 
+		{
```
migueldeicaza, line 877: Let us replace "Wished" with "Desired"

UICatalog/Scenarios/Editor.cs
```diff
@@ -23,21 +24,43 @@ public override void Init (Toplevel top, ColorScheme colorScheme)
...
-					new MenuItem ("_Quit", "", () => Quit()),
-				}),
-				new MenuBarItem ("_Edit", new MenuItem [] {
-					new MenuItem ("_Copy", "", () => Copy()),
-					new MenuItem ("C_ut", "", () => Cut()),
-					new MenuItem ("_Paste", "", () => Paste())
-				}),
-				new MenuBarItem ("_ScrollBarView", CreateKeepChecked ())
-			});
+			List<MenuBarItem> menuBarItems = new List<MenuBarItem> (
```
migueldeicaza, line 27: Gratuitous reformatting, please keep the original style

migueldeicaza (changes requested):

G-Freart:
> Really good job. Thanks. I think that selecting the entire character initially can cause the impression that the capslock is on and confuse the user. Wouldn't it be possible to put a narrower width if the capslock is turned off?

Thanks

I have search after a way to known the capslock state but i found nothing direct . What i have seen under linux is to connect to the X11 if any display is available and get the led status as we could get using the shell command 'xset -q' but it's far from a nice solution as we could have user connected using ssh and without X11 connected !

I have change the cursor displayed to the CursorVisibility.Default so under xterm, it will use the setinngs set by the user, and if not under xterm, it will use the underline. Under windows, the cursor will be underline as not under an xterm.

G-Freart:
Hello Miguel,

Don't hesitate to tell me if they are anything else or if more re-styling must be done.

Best Regards,
Gilles Freart

BDisp:
I was never able to make capslock work on `Linux`. He really needs a Display to work. But if you managed to make it work, having a `Display` is better than nothing. An example I can give about the uncertainty that capslock may or may not be activated, is in the `TextField` where if it is activated, the character under the cursor is printed with the inverted color looking like the cursor blinking, but without blinking. In my opinion if the `capslock` is off, it is better to leave the cursor underlined, in cases where its visibility is true.

BDisp (approved):

G-Freart:
Hello @migueldeicaza ,

Something is missing ?

Best Regards,
Gilles Freart

Bonny40:
Ich bin neu hier 
Hallo zusammen 
Ich brauche hilf ich habe Probleme mit mein Handy ich glaube Hecker ist hinter mir her

tlemmons:
I am not sure if this is related.  I cannot see the cursor in a dialog->text field unless it is on a character.  Being in empty space (after any characters) it is invisible.  I have pulled current but no change.  Is there some setting I am missing for this?

PR #1137, opened 2021-03-15
New Control: Tabview

tznind:
Hey.  I have created a new control that replicates tabs (like browsers, notepad++ etc).

I have added 2 scenarios

 - TabViewExample.cs demonstrates a range of tabs (100 tabs)
 - Notepad.cs demonstrates creating/removing tabs dynamically

tabs

### DIFF
--- a/Terminal.Gui/Views/TabView.cs
+++ b/Terminal.Gui/Views/TabView.cs
@@ -1,0 +1,192 @@
+using NStack;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// A single tab in a <see cref="TabView"/>
+	/// </summary>
+	public class Tab {
+		private ustring text;
+
+		/// <summary>
+		/// The text to display in a <see cref="TabView"/>
+		/// </summary>
+		/// <value></value>
+		public ustring Text { get => text ?? "Unamed"; set => text = value; }
+
+		/// <summary>
+		/// The control to display when the tab is selected
+		/// </summary>
+		/// <value></value>
+		public View View { get; set; }
+
+		/// <summary>
+		/// Creates a new unamed tab with no controls inside
+		/// </summary>
+		public Tab ()
+		{
+
+		}
+
+		/// <summary>
+		/// Creates a new tab with the given text hosting a view
+		/// </summary>
+		/// <param name="text"></param>
+		/// <param name="view"></param>
+		public Tab (string text, View view)
+		{
+			this.Text = text;
+			this.View = view;
+		}
+	}
+
+	/// <summary>
+	/// Describes render stylistic selections of a <see cref="TabView"/>
+	/// </summary>
+	public class TabStyle{
+
+        /// <summary>
+        /// True to show the top lip of tabs.  False to directly begin with tab text during 
+        /// rendering.  When true header line occupies 3 pixels, when false only 2.
+        /// Defaults to true.
+        /// 
+        /// <para>When <see cref="TabsOnBottom"/> is enabled this instead applies to the
+        ///  bottomost line of the control</para>
+        /// </summary> 
+        public bool ShowHeaderOverline {get;set;} = true;
+
+        /// <summary>
+        /// True to render tabs at the bottom of the view instead of the top
+        /// </summary>
+        public bool TabsOnBottom {get;set;} = false;
+
+    }
+
+    /// <summary>
+    /// Control that hosts multiple sub views, presenting a single one at once
+    /// </summary>
+    public class TabView : View
+    {
+        /// <summary>
+        /// All tabs currently hosted by the control, after making changes call <see cref="View.SetNeedsDisplay()"/>
+        /// </summary>
+        /// <value></value>
+        public List<Tab> Tabs {get;set;} = new List<Tab>();
+
+        /// <summary>
+        /// The currently selected member of <see cref="Tabs"/> chosen by the user
+        /// </summary>
+        /// <value></value>
+        public Tab SelectedTab {get;set;}
+
+        /// <summary>
+        /// Render choices for how to display tabs
+        /// </summary>
+        /// <value></value>
+        public TabStyle Style {get;set;} = new TabStyle();  
+
+		///<inheritdoc/>
+		public override void Redraw (Rect bounds)
+		{
+            Move(0,0);
+
+            var tabLocations = MeasureTabs().ToArray();
+            var width = bounds.Width;
+
+            int currentLine = 0;
+
+            if(Style.ShowHeaderOverline){
+                RenderOverline(tabLocations,width);
+                currentLine++;
+            }
+
+            Move(0,currentLine);
+
+            foreach(var toRender in tabLocations){
+
+                Move(toRender.X,currentLine);
+                Driver.AddStr(toRender.Tab.Text);
+            }
+        }
+
+		private void RenderOverline (TabToRender[] tabLocations, int width)
+		{
+            // Renders the top line of selected tab like 
+            //          
+
+            Move(0,0);
+
+            var selected = tabLocations.FirstOrDefault(t=>t.IsSelected);
+
+            // Clear out everything
+            Driver.AddStr(new string(' ',width));
+
+            // Nothing is selected... odd but we are done
+            if(selected == null)
+            {
+                return;
+            }
+
+
+            Move(0,selected.X-1);
+            Driver.AddRune(Driver.ULCorner);
+
+            for(int i= 0; i < selected.Width ;i++){
+
+                if(selected.X+i > width)
+                {
+                    // we ran out of space horizontally
+                    return ;
+                }
+
+                Driver.AddRune(Driver.HLine);
+            }
+
+            // Add the end of the selected tab
+            Driver.AddRune(Driver.URCorner);		    
+                
+		}
+
+		/// <summary>
+		/// Returns which tabs to render at each x location
+		/// </summary>
+		/// <returns></returns>
+		private IEnumerable<TabToRender> MeasureTabs()
+		{
+            int i=1;
+            var toReturn = new Dictionary<int,Tab>();
+
+            foreach(var tab in Tabs)
+            {
+                var tabTextWidth = tab.Text.Sum(c=>Rune.ColumnWidth(c));
+                yield return new TabToRender(i,tab,Equals(SelectedTab,tab),tabTextWidth);
+                i += tabTextWidth + 1;
+            }
+		}
+
+        private class TabToRender{
+            public int X {get;set;}
+            public Tab Tab {get;set;}
+
+            /// <summary>
+            /// True if the tab that is being rendered is the selected one
+            /// </summary>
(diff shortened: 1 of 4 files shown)


tig:
AWESOME!!!

BDisp:
@tznind to avoid the `ValidatePosDim` error and force changes on `Pos` or `Dim`, it's necessary to set the Layout to `Absolute` first. See the `Update` method in the Button.cs. I made this to avoid break the `Layout.Computed` inadvertently.

tznind:
> @tznind to avoid the `ValidatePosDim` error and force changes on `Pos` or `Dim`, it's necessary to set the Layout to `Absolute` first. See the `Update` method in the Button.cs. I made this to avoid break the `Layout.Computed` inadvertently.

Cool, I thought it was something like that.

I can change it to 
```
tabsBar.LayoutStyle = LayoutStyle.Absolute;
tabsBar.Y = 0;
tabsBar.LayoutStyle = LayoutStyle.Computed;
```

or I can leave it how it is (and remove comment about `ValidatePosDim`)?
```
tabs.Y = Pos.Percent(0);
```

Let me know which you would prefer or if there is a better approach.

BDisp:
You can maintain the comment about `ValidatePosDim` to alert others. I should alert in the exception to change from `Computed` to `Absolute` if it is the case. May be later if it another change is needed.
You can do in the way you like more. In the `Button.cs` I save the Layout to a variable and restore it again. If you think the `Percent` works well, leave it as it is. You are doing a great job. Thanks.

tznind:
Cool, thanks for explaining, it makes sense.  I will leave it as `Percent.Pos` then.

TabView has 2 private subviews.  One for the line of the tabs (text and tab outline) and one for the content area (inside the current tab frame).

The `ApplyStyleChanges` method (which is where `Pos.Percent` code is) all relates to these private views so shouldn't run into any state based issues imposed by the API user (i.e. the layout of the TabView itself should not matter).

UICatalog/Scenarios/TabViewExample.cs
```diff
@@ -0,0 +1,210 @@
...
+			var menu = new MenuBar (new MenuBarItem [] {
+				new MenuBarItem ("_File", new MenuItem [] {
+
+					new MenuItem ("_Add Blank Tab", "", () => AddBlankTab()),
+
+					new MenuItem ("_Clear SelectedTab", "", () => tabView.SelectedTab=null),
+					new MenuItem ("_Quit", "", () => Quit()),
+				}),
+				new MenuBarItem ("_View", new MenuItem [] {
+					miShowTopLine = new MenuItem ("_Show Top Line", "", () => ShowTopLine()){
```
tig, line 38: Note how you use different words to describe this flag in UI ("Show Top Line") vs. code (`ShowHeaderOverline). Please consider how this would all be simpler if you used consistent terminology. I've always found that "what makes sense to the end-user will make sense to the programmer" (not the other way around).

tig:
I pulled it down locally and played. Lovely!!!!

Some Feedback:

* The Notepad sample when loaded by default looks confusing. There's just a horizontal line across the screen and it looks like a bug/painting error. This is because you are not following the "New" codepath at  startup.  
  * Related to this, I think a default 'empty' Tab, with no name should still LOOK like a tab. Right?
  * I really find that when "Show Top Line" is off (which Notepad uses by default) looks confusing. I understand how it 'wastes' a row, but I'd be surprised if many apps actually use it. Consider just removing this style initially in the spirit of removing  complexity from "Minimally Loveable Product" (MLP) release.


More in review feedback.

UICatalog/Scenarios/TabViewExample.cs
```diff
@@ -0,0 +1,210 @@
...
+			var frameRight = new FrameView ("About") {
+				X = Pos.Right (tabView),
+				Y = 0,
+				Width = Dim.Fill (),
+				Height = Dim.Fill (),
+			};
+
+
+			frameRight.Add (new TextView () {
+				Text = @"This demos the tabs control
```
tig, line 90: When I run this I see 

```
This demos the tabs contro
Switch between tabs using cursor key
```
(the `l` in `control` is not drawn)

Terminal.Gui/Views/TabView.cs
```diff
@@ -0,0 +1,794 @@
...
+				// if there are mmore tabs to the right not visible
+				if (tabLocations.LastOrDefault ()?.Tab != host.Tabs.LastOrDefault ()) {
+					Move (width - 1, y);
+
+					// indicate that
+					Driver.AddRune (Driver.RightArrow);
+				}
+			}
+
+			public override bool MouseEvent (MouseEvent me)
```
tig, line 729: I can't get the left/right arrows to scroll tabs in the `Tab View` sample with mouse (in cmd.exe, which should work).

UICatalog/Scenarios/TabViewExample.cs
```diff
@@ -0,0 +1,210 @@
...
+			var frameBelow = new FrameView ("Bottom Frame") {
+				X = 0,
+				Y = Pos.Bottom (tabView),
+				Width = tabView.Width,
+				Height = Dim.Fill (),
+			};
+
+
+			frameBelow.Add (new TextView () {
+				Text = @"This frame exists to check you can still tab here
```
tig, line 111: Same here (`her`) is displayed insetead of (`here`).

Terminal.Gui/Views/TabView.cs
```diff
@@ -0,0 +1,794 @@
...
+	}
+
+	/// <summary>
+	/// Describes render stylistic selections of a <see cref="TabView"/>
+	/// </summary>
+	public class TabStyle {
+
+		/// <summary>
+		/// True to show the top lip of tabs.  False to directly begin with tab text during 
+		/// rendering.  When true header line occupies 3 pixels, when false only 2.
```
tig, line 54: We use the terms `rows` and `cols` or `columns`, not pixels. Please be consistent.

Terminal.Gui/Views/TabView.cs
```diff
@@ -0,0 +1,794 @@
...
+		/// Defaults to true.
+		/// 
+		/// <para>When <see cref="TabsOnBottom"/> is enabled this instead applies to the
+		///  bottomost line of the control</para>
```
tig, line 58: bottommost (two m)

Terminal.Gui/Views/TabView.cs
```diff
@@ -0,0 +1,794 @@
...
+		/// <summary>
+		/// When there are too many tabs to render, this indicates the first
+		/// tab to render on the screen.
+		/// </summary>
+		/// <value></value>
+		public int TabScrollOffset { get; set; }
+
+		/// <summary>
+		/// The maximum number of characters to render in a Tab header.  This prevents one long tab 
+		/// from pushing out all the others. Values less than 0 are ignored
```
tig, line 113: If values less than 0 are ignored, why not make the type a `uint`?

Terminal.Gui/Views/TabView.cs
```diff
@@ -0,0 +1,794 @@
...
+		}
+
+
+		/// <summary>
+		/// Removes the given <paramref name="tab"/> from <see cref="Tabs"/>.
+		/// Optionally disposes the tabs hosted <see cref="Tab.View"/>>
+		/// </summary>
+		/// <param name="tab"></param>
+		/// <param name="dispose">True to dispose of the tabs control</param>
+		public void RemoveTab (Tab tab, bool dispose)
```
tig, line 451: I wonder (I have no idea if my intution is correct, so please debate) if it would just be simpler for you to have the documentation say "caller is responsible for disposing any Tabs added". 

You don't normally see APIs on .NET classes that include a `dispose` option.

UICatalog/Scenarios/Notepad.cs
```diff
@@ -0,0 +1,242 @@
...
+		{
+			Win.Title = this.GetName ();
+			Win.Y = 1; // menu
+			Win.Height = Dim.Fill (1); // status bar
+			Top.LayoutSubviews ();
+
+			var menu = new MenuBar (new MenuBarItem [] {
+				new MenuBarItem ("_File", new MenuItem [] {
+					new MenuItem ("_New", "", () => New()),
+					new MenuItem ("_Open", "", () => Open()),
```
tig, line 31: When I choose this two bad things happen:

1. The `Ok` button in the fIle open dialog doesn't say `Ok`
2. If I ht `tab` 5 times the app crashes.

I think you're doing something wrong in your File Open dialog calling code. I can't reproduce this using the `Editor` scenario.

tig: Note "Save As" seems to work fine.

tig (changes requested):
Please add unit tests.

I've added some additional review comments in the code. most of them are just suggestions for your consideration. 

This is really amazing work and I'm excited to be able to merge it in!!!! Thank you!!!!

UICatalog/Scenarios/Notepad.cs
tznind: Think it was calling the blank constructor that was causing this problem.  I resolved it by using the strings constructor.

BDisp: @tznind please check if it will also throws exception in the `UICatalog` `All Views Tester`.

UICatalog/Scenarios/TabViewExample.cs
tznind: Think this was down to using `@""` notation with newlines in source.  I fixed by changing to regular string `"some\ntext"`

UICatalog/Scenarios/Notepad.cs
tznind: Ok have figured out what I did wrong.  I used `FileDialog` instead of `OpenDialog`.  But it does seem that there is a bug when using `FileDialog()` blank constructor that results in the Open button having no text and crashing when Tabbed to.  Do you want me to raise a seperate issue for that?

tznind:
Thanks for the indepth review! I think I have fixed all the issues you raised.  Let me know if I missed any or misinterpreted any.

**When there are no Tabs**
Now when there are no Tabs the TabView its behaviour is:

- If `ShowBorder` is on then the border is rendered but the rest is blank
- If `ShowBorder` is off then control is completely blank

I thought about adding a blank tab in the case when there are no tabs but thought that would be confusing and not consistent with the other views like ListView which is just blank when there are no items.

I have added a call to `New()` at the start of the Notepad Scenario so it looks better when it first appears (consistent with notepad++ etc)

**ShowTopLine**

I have removed the setting of `ShowTopLine=false`  from the `Notepad` scenario.  

The default value is true for `ShowTopLine`. 

I think there are still some corner cases where it could be useful to cut down on the amount of vertical space taken up by TabView (e.g. when you want to show 2 tab views one above the other).  Let me know if you really don't want it in and I can delete the code for it.

UICatalog/Scenarios/Notepad.cs
tig: Yes, I think you should.

tig (approved):

tig:
I chatted with @migueldeicaza and we both agree this should be merged! Well done!

PR #1147, opened 2021-03-20
Fixes #134. TextView: Add line wrapping.

### LINKED ISSUE #134: TextView: Add line wrapping
currently text flows outside the frame. Add an option to wrap text horizontally.

BDisp:
- Still need to update the cursor position when resizing. - DONE
- Still need to update the changes made in the text. - DONE
- And some help would be welcome :-) - Always

textview-line-wrapping

### DIFF
--- a/Terminal.Gui/Core/Application.cs
+++ b/Terminal.Gui/Core/Application.cs
@@ -702,8 +702,13 @@ namespace Terminal.Gui {
 			var resume = true;
 			while (resume)
 			{
-				try
-				{
+#if DEBUG
+				resume = false;
+				var runToken = Begin (view);
+				RunLoop (runToken);
+				End (runToken);
+#else
+				try {
 					resume = false;
 					var runToken = Begin (view);
 					RunLoop (runToken);
@@ -717,6 +722,7 @@ namespace Terminal.Gui {
 					}
 					resume = errorHandler(error);
 				}
+#endif
 			}
 		}
 
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1471,6 +1471,8 @@ namespace Terminal.Gui {
 			KeyPress?.Invoke (args);
 			if (args.Handled)
 				return true;
+			if (Focused?.ProcessKey (keyEvent) == true)
+				return true;
 			if (subviews == null || subviews.Count == 0)
 				return false;
 			foreach (var view in subviews)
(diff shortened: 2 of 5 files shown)


tig:
Is this a draft? 

* Still need to update the changes made in the text.

This seems to be a pretty bad experience.

BDisp:
> Is this a draft?

Sorry, I forgot to set as draft.



> This seems to be a pretty bad experience.

I just wanted to convey my intention of what I want to do, but I accept the reprimand :-)

tig:
> > Is this a draft?
> 
> Sorry, I forgot to set as draft.
> 
> > This seems to be a pretty bad experience.
> 
> I just wanted to convey my intention of what I want to do, but I accept the reprimand :-)

It was not a reprimand! You are doing fantastic work. I was just stating an opinion about the experience, not the person.

BDisp:
See `Ctrl+K` (`kill-to-end`) and `Ctrl+Y` (`yank`) in action. In the video I cut (`Ctrl+K`) all the text from the current cursor position and paste (`Ctrl+Y`) on the same cursor position.  If you move the cursor or do anything else, the clipboard contents stops appending.

textview-kill-to-end+yank

BDisp:
It also can be done from the bottom to top.

textview-kill-to-end+yank-2

BDisp:
There also features to select (`Ctrl+Space`), cut (`Ctrl+W`) and paste (`Ctrl+Y`).

BDisp:
Well, I think it's all done. Please give feedback of any issues or suggestions. I maintained the original `TextView` hotkeys which it isn't the same as `TextField`, to not break anything.

textview-copy-cut-paste

Terminal.Gui/Core/Application.cs
```diff
@@ -702,8 +702,13 @@ public static void Run (Toplevel view, Func<Exception, bool> errorHandler = null
 			var resume = true;
 			while (resume)
 			{
-				try
-				{
+#if DEBUG
```
tig, line 705: I think it's generally a bad idea to have DEBUG work differently than RELEASE.

Is this something that you needed for when YOU were debugging this, or something that is absolutely required anytime ANYONE is running DEBUG?

tig (approved):
This is beyond impressive.

I'm going to pull it down and test the heck out of it before merging though; there's a lot of code here!

Have you considered what unit tests might be able to verify behavior? There is so much logic involved that I fear this could become a "bug farm" over time without solid unit tests.

Terminal.Gui/Core/Application.cs
BDisp: For me is for anytime ANYONE is running DEBUG because is much faster to catch errors in the local where they occurred. Without this forces me to waste much more time to debug. This is the main enter and there no needs to catch here all the exceptions in DEBUG mode.

BDisp:
> Have you considered what unit tests might be able to verify behavior? There is so much logic involved that I fear this could become a "bug farm" over time without solid unit tests.

Yes, I considered in unit tests but I can not guarantee it for when, but I assure you that I have solved some bugs that existed and is better than before.

Issue #1152, opened 2021-03-24
All Views Tester breaks on TreeView generic constructor


tznind:
Hi guys, sorry I didn't spot this before but the `All Views Tester` in `UICatalog`  reflection code is tripping up on the generic version of TreeView.

```
{"Cannot create an instance of Terminal.Gui.TreeView`1[T] because Type.ContainsGenericParameters is true."}

This exception was originally thrown at this call stack:
    [External Code]
    UICatalog.AllViewsTester.CreateClass(System.Type) in AllViewsTester.cs
    UICatalog.AllViewsTester.Setup.AnonymousMethod__22_8(Terminal.Gui.ListViewItemEventArgs) in AllViewsTester.cs
    Terminal.Gui.ListView.OnSelectedChanged() in ListView.cs
    Terminal.Gui.ListView.MoveDown() in ListView.cs
    Terminal.Gui.ListView.ProcessKey(Terminal.Gui.KeyEvent) in ListView.cs
    Terminal.Gui.View.ProcessKey(Terminal.Gui.KeyEvent) in View.cs
    Terminal.Gui.View.ProcessKey(Terminal.Gui.KeyEvent) in View.cs
    Terminal.Gui.Toplevel.ProcessKey(Terminal.Gui.KeyEvent) in Toplevel.cs
    Terminal.Gui.View.ProcessKey(Terminal.Gui.KeyEvent) in View.cs
```

I will try to get a PR in to fix it asap.

Steps to repro:

1. Run All Views Tester
2. Scroll down to TreeView

tznind:
Fixed by #1153

PR #1153, opened 2021-03-24
Fix for  #1152 (Generic TreeView`1 breaks upon selection in All Views Tester) 

tznind:
Fix for #1152 

This PR makes `CreateClass ` method in the 'All Views Tester' Scenario handle generic types.

Now when asked to create `MyClass<T>` (which is impossible) it will create an instance of `MyClass<object>`

### DIFF
--- a/UICatalog/Scenarios/AllViewsTester.cs
+++ b/UICatalog/Scenarios/AllViewsTester.cs
@@ -344,6 +344,20 @@ namespace UICatalog {
 
 		View CreateClass (Type type)
 		{
+			// If we are to create a generic Type
+			if (type.IsGenericType) {
+
+				// For each of the <T> arguments
+				List<Type> typeArguments = new List<Type> ();
+
+				// use <object>
+				foreach (var arg in type.GetGenericArguments ()) {
+					typeArguments.Add (typeof (object));
+				}
+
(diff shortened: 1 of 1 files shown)


tig:
Why are there two treeviews?

image

tznind:
> Why are there two treeviews?
> 
> image

The first is `TreeView`.  The second is `TreeView<T>` (Generic version).  The C# ToString on a Type returns the name of the type followed by a backtick and a number (if it is generic).  The number is the number of generic templates it has e.g. `Dictionary<T1,T2>` would be Dictionary backtick 2.

More info on the difference between `TreeView` and `TreeView<T>` and example use cases is in:

https://github.com/migueldeicaza/gui.cs/blob/master/docfx/articles/treeview.md

tig:
> > Why are there two treeviews?
> > image
> 
> The first is `TreeView`. The second is `TreeView<T>` (Generic version). The C# ToString on a Type returns the name of the type followed by a backtick and a number (if it is generic). The number is the number of generic templates it has e.g. `Dictionary<T1,T2>` would be Dictionary backtick 2.

I did not know that. Wild. I guess I should change the code in the all views tester to look for that backtick and replace it with the appropriate `<T1>`. I wonder if anyone's written a ToString extension method that already does that?

PR #1156, opened 2021-03-27
Fixes #1155. MoveForward/MoveBackward not bound on Text controls.

### LINKED ISSUE #1155: MoveForward/MoveBackward not bound on Text controls
Alt-F, and Alt-B for forward-word and backward-word do not seem to be hooked up anymore in both controls.

On TextView, it seems like backward has the wrong semantics, I suspect that when the code was moved from indexes to col/row manipulation we lost one line of code that did a "i++" after the two loops.  The equivalent would be to remember the last col/row before MoveBack is attempted and then restore that value.

BDisp:
textview-textfield-forw-back

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1471,6 +1471,8 @@ namespace Terminal.Gui {
 			KeyPress?.Invoke (args);
 			if (args.Handled)
 				return true;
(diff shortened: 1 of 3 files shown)


BDisp:
This PR was already merged into the PR #1147 because I needed these features there.

PR #1170, opened 2021-04-02
Fixes #1169. Forcing line feed on cs files.

### LINKED ISSUE #1169: Everytime I do a Ctrl+K+D(F) format the c# end up with mixed lines feed.

BDisp:


tig:
Why is this needed?

BDisp:
Force us to change to line feed when comparing the changes before commit.

BDisp:
For a long time this did not happen in the VS2019, because it respected the end of line attached to the file. Recently the VS2019 is following .editorconfig file guidelines and as the files are stored in GitHub as LF the VS2019, after some formatting, changes to CRLF + LF mixed lines.

tznind:
Does VS not use the local git settings for diffs?

https://docs.github.com/en/github/getting-started-with-github/configuring-git-to-handle-line-endings

Apologies if I have misunderstood

BDisp:
Thanks @tznind. I already have in my `.gitconfig` under user folder this setting:
```
[core]
	autocrlf = true
```
I' don't have any problems on checkout and on commits. I have only problems when I use `Ctrl+K+D` or `Ctrl+K+F` to format the `.cs` file which end up with mixed lines and appearing in the git diff. I have to toggle to `LF` or ignore in the git diff, but I prefer to save the file with `LF`.
I think this starting happen with some VS2019 update.

tig:
@BDisp what version of VS2019 are you running?

I'm running `Version 16.9.0 Preview 4.0` and i do not see the same problem. I note that there is a  `Version 16.10.0 Preview 1.0` available but I have not upgraded.

BDisp:
@tig, I only use releases versions which is the `VS2019 Community 16.9.3` version.

tig:
@BDisp I hate accepting PRs that I do not understand. 

I do not understand why this is needed. I DO understand that you saw a problem on your machine and that this fixes that problem. But I do not understand what CAUSED that problem, and why it is not impacting anyone else. 

I never heard back from the VS team on if they changed behavior.

BDisp:
Do not worry about this @tig :-)
It's not the first time this happens and then some VS2019 update resolves.

PR #1178, opened 2021-04-03
Fixes #1177. Now is possible to copy or cut on the TextView with the …

### LINKED ISSUE #1177: On Linux releasing the mouse after a previous pressing and moving, causes a button click event.
Because of this, it isn't possible to copy or cut on the `TextView`.

BDisp:
…mouse.


tig (approved):
Love!

PR #1192, opened 2021-04-09
Add Vertical Alignment and Text Direction + UICatalog Demo

jmperricone:
image


jmperricone:
I used the wrong user to commit. I will pull with the correct one. Sry.

PR #1201, opened 2021-04-10
Feature/graphs

### LINKED ISSUE #1045: New control: Charts
From the go library term UI it might be nice to bring the sparkline, plot, bar chart and line chart controls:

https://github.com/gizak/termui

tznind:
Hey I have created a basic graph rendering control.

It works by having a 2D 'graph space' using `decimal` and a CellSize for mapping how to project screen coordinates into rectangles in the graph space.

New series implement `ISeries`  which is a single method

```csharp
/// <summary>
/// Return the rune that should be drawn on the screen (if any)
/// for the current position in the control
/// </summary>
/// <param name="graphSpace">Projection of the screen location into the chart graph space</param>
GraphCellToRender GetCellValueIfAny (RectangleD graphSpace);
```
I started out with float instead of decimal but ran into issues with rounding where a given point in graph space would overlap into 2 different cells on the UI.  Those all disappeared when I switched to `decimal` which is the exact units Type (see test `TestNoOverlappingCells`).

graphview

Let me know if this is too out there or better in a separate library.

I can split out the classes int separate files too if needed.

Fixes #1045


BDisp:
That is fantastic. Thanks.

tznind:
Marking this as draft as I have found it is very slow for large numbers of plot points, need to tinker with it a bit.

tig:
Once again... AMAZING WORK!!!!!!

tznind:
Ok I have added the ability for an `ISeries` to (optionally) draw all at once instead of having to return each cell 1 at a time.

I've added `SeriesDrawMode` to `ISeries` and an abstract base class.  That makes it easier for API user to understand the two draw modes.  There are also tests that cover this (see `FakeSeries`).

- AllAtOnce : More efficient but requires user to use methods like `GraphSpaceToScreen`
- CellByCell: Simpler and easier to implement for the API user

I think this PR is ready for review now when you have the time.

```csharp
/// <summary>
/// Describes a series of data that can be rendered into a <see cref="GraphView"/>>
/// </summary>
public interface ISeries {

	/// <summary>
	/// Indicates which method is used for drawing
	/// </summary>
	SeriesDrawMode DrawMode {get; }

	/// <summary>
	/// Draw method when <see cref="DrawMode"/> is <see cref="SeriesDrawMode.AllAtOnce"/>.
	/// Do custom drawing as needed to fill relevant areas of the graph
	/// </summary>
	/// <param name="graph">Graph series is to be drawn onto</param>
	/// <param name="driver"></param>
	/// <param name="bounds">Visible area of the graph in Console Screen units (excluding margins)</param>
	/// <param name="graphBounds">Visible area of the graph in Graph space units</param>
	void DrawSeries (GraphView graph, ConsoleDriver driver, Rect bounds, RectangleD graphBounds);

	/// <summary>
	/// Draw method when <see cref="DrawMode"/> is <see cref="SeriesDrawMode.CellByCell"/>.
	/// Return the rune that should be drawn on the screen (if any)
	/// for the current position in the control
	/// </summary>
	/// <param name="graphSpace">Projection of the screen location into the chart graph space</param>
	GraphCellToRender GetCellValueIfAny (RectangleD graphSpace);
}
```

Terminal.Gui/Views/GraphView.cs
```diff
@@ -0,0 +1,1913 @@
...
+			Orientation = orientation;
+			ScreenLocation = screen;
+			GraphSpace = graphSpace;
+		}
+	}
+
+	/// <summary>
+	/// Rectangle class based on the exact floating point primative 'decimal'
+	/// </summary>
+	public class RectangleD {
```
tig, line 1582: Did you code these floating point graphic primiatives yourself or copy the code from e.g. `System.Drawing's RectangleF`?

We should avoid re-implementing something that others have done before (and tested). So I'm hoping you say you copied the code. e.g. from https://docs.microsoft.com/en-us/dotnet/api/system.drawing.rectanglef?view=net-5.0

Why `decimal` v. `float`?

As you have implemented this, these classes are at the top `Terminal.Gui` class hierarchy. This means we are explicitly documenting them such that they can be used by others OUTSIDE of the use of `GraphView`. I think we need to very carefully consider whether this is a good idea or not.

If it IS a good idea then the bar for exposing them as `public` is MUCH higher and we should have superb unit tests for every aspect of the API. 

If it is NOT a good idea then we should consider how we would either NOT make them public or ensure they are contained with a namespace that identifies them as part of `GraphView`.

Terminal.Gui/Views/GraphView.cs
```diff
@@ -0,0 +1,1913 @@
...
+			return Math.Min (Math.Max ((int)graph.MarginLeft, origin.X), bounds.Width - 1);
+		}
+	}
+
+
+	/// <summary>
+	/// A location on an axis of a <see cref="GraphView"/> that may
+	/// or may not have a label associated with it
+	/// </summary>
+	public class AxisIncrementToRender {
```
tig, line 1536: Example of a basic class that is now exposed as `public` for all of `terminal.gui`. But it has no unit tests and it is not clear if it would ever be useful outside of GraphView. 

Please consider both making it private and adding unit tests.

Terminal.Gui/Views/GraphView.cs
```diff
@@ -0,0 +1,1913 @@
...
+			// float the X axis so that it accurately represents the origin of the graph
+			// but anchor it to top/bottom if the origin is offscreen
+			return Math.Min (Math.Max (0, origin.Y), bounds.Height - ((int)graph.MarginBottom + 1));
+		}
+	}
+
+	/// <summary>
+	/// The vertical (i.e. Y axis) of a <see cref="GraphView"/>
+	/// </summary>
+	public class VerticalAxis : Axis {
```
tig, line 1371: Another example of a building-block class that should be constrainted to GraphView somehow.

Terminal.Gui/Views/GraphView.cs
```diff
@@ -0,0 +1,1913 @@
...
+
+				return new GraphCellToRender (Fill.Rune, customColor);
+			}
+		}
+	}
+
+	/// <summary>
+	/// Sequence of lines to connect points e.g. of a <see cref="ScatterSeries"/>
+	/// </summary>
+	public class PathAnnotation : IAnnotation {
```
tig, line 1020: I'm being repetitive here (and going from bottom up, so sorry about that).

If some `Terminal.Gui` user encounters this class in the object browser or documentation they will have no clue what it is for. It may distract them from their primary goal. Simillarly, someone LOOKING for this functionality in relation to `GraphView` will not know it is part of `GraphView`.

There are many ways of addressing this. 

- Rename the class to include `GraphView` (ugh)
- Create a namespace as a container
- Refactor some  other way

What other solutions are there? (I'm sincerely asking).

tig (changes requested):
My biggest feedback is around how this implementation exposes half-a-dozen or so auxiliary classes to all of `Terminal.Gui` needlessly.

I don't think we can merge this until the questions I ask are resolved.

This is amazing work and I am eager to merge it, so don't get me wrong!

Terminal.Gui/Views/GraphView.cs
migueldeicaza: It could just get its own namespace.

That means that only those using graphics would bring the additional types.

migueldeicaza:
This is a glorious set of controls!

Simply incredible!

Terminal.Gui/Views/GraphView.cs
BDisp: I think changing the namespace to `namespace Terminal.Gui.GraphView` and put inside a folder named `GraphView`  may be a solution.

tznind:
Thanks for the indepth review, probably best to start with the use of `decimal` instead of `float`

I did start out with using `float` but ran into issues where points would occasionally double render where the graph location was at the boundary of a console cell.

This test replicates the issue.  We partition the console space 50x30 into graph space units of 0.1 x 0.25 but find that the resulting rectangles overlap with one another (at x=7 y=0 on my computer).  I think this is because of float rounding:

https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html

The result is that an axis tick or scatter series point can wind up appearing twice in two adjacent cells.

```csharp
[Fact]
public void TestOverlappingRectangleF ()
{
	int width = 50;
	int height = 30;

	var size = new PointF (0.1f, 0.25f);

	List<RectangleF> otherRects = new List<RectangleF> ();

	List<System.Drawing.Point> overlappingPoints = new List<System.Drawing.Point> ();

	for (int x = 0; x < width; x++) {
		for (int y = 0; y < height; y++) {
			
			var newRect = new RectangleF (x * size.X, y * size.Y,size.X,size.Y);

			Assert.Empty (otherRects.Where (r => r.IntersectsWith (newRect)));
			otherRects.Add (newRect);
		}
	}
}
```

It is tied up to how you can't use `floata == floatb` and instead have to use `Math.Abs(floata - floatb) < 0.000000001`

See:
https://www.jetbrains.com/help/resharper/CompareOfFloatsByEqualityOperator.html

I am open to ideas on how to deal with this issue as I agree it feels janky.

Note that the equivellent test with decimal does not show this issue:

```csharp

[Fact]
public void TestOverlappingRectangleD ()
{
	int width = 50;
	int height = 30;

	var size = new PointD (0.1M, 0.25M);

	List<RectangleD> otherRects = new List<RectangleD> ();

	List<PointD> overlappingPoints = new List<PointD> ();

	for (int x = 0; x < width; x++) {
		for (int y = 0; y < height; y++) {

			var newRect = new RectangleD (x * size.X, y * size.Y, size.X, size.Y);
			Assert.Empty (otherRects.Where (r => r.IntersectsWith (newRect)));
			otherRects.Add (newRect);
		}
	}
}
```

Terminal.Gui/Views/GraphView.cs
tznind: Thanks, I will move them to a sub namespace and folder and look to see if any could be made private.

Terminal.Gui/Views/GraphView.cs
tznind: Yes the code for RectangleD is just RectangleF with a find/replace on `float` for `double` http://www.dotnetframework.org/default.aspx/DotNET/DotNET/8@0/untmp/whidbey/REDBITS/ndp/fx/src/CommonUI/System/Drawing/Advanced/RectangleF@cs/1/RectangleF@cs 

Since dotnet is open source already I didn't see a problem with copying it.  I've put a comment in the main thread about why I switched from `float` to `decimal` and am open to exploring that choice.

Terminal.Gui/Views/GraphView.cs
BDisp: Since it is to stay on a different namespace/folder and if these classes needs to be public for derived classes or whatever, I think they can stay as they are.

BDisp:
This work:

`var newRect = new RectangleF ((int)(x * size.X), (int)(y * size.Y), (int)size.X, (int)size.Y);`

Terminal.Gui/Views/GraphView.cs
BDisp: I now saw the structure and I think the `Terminal.Gui.Graphs` should be inside the `Core` folder with a `Graph .cs` file with all the APIs that belong to him. The `GraphView.cs` file could be stay on the root of the `Views` folder with only the APIs of the view. What do you think?

tznind:
> var newRect = new RectangleF ((int)(x * size.X), (int)(y * size.Y), (int)size.X, (int)size.Y);

So that will round all the RectangleF to integer which results in exact overlaps instead (which IntersectsWith returns false for) 

See:
```csharp
[Fact]
public void TestOverlappingRectangleF ()
{
	int width = 50;
	int height = 30;

	var size = new PointF (0.1f, 0.25f);

	List<RectangleF> otherRects = new List<RectangleF> ();

	List<System.Drawing.Point> overlappingPoints = new List<System.Drawing.Point> ();

	for (int x = 0; x < width; x++) {
		for (int y = 0; y < height; y++) {

			var newRect = new RectangleF ((int)(x * size.X), (int)(y * size.Y), (int)size.X, (int)size.Y);

			// I added this line to show that the returned rectangles are identical when casting to int
			Assert.Empty (otherRects.Where (r => r.Equals (newRect)));
			Assert.Empty (otherRects.Where (r => r.IntersectsWith (newRect)));
			otherRects.Add (newRect);
		}
	}
}
```

The GraphView control maps integer screen coordinates (e.g. 50,30) into 'graph space'.  For example the user might want a graph
that has an x axis that goes from 0-1 in increments of 0.001.

That is what CellSize does, it lets the user say "Each console cell translates to 0.001 of space on my data".  The Zoom function in the Scenario just adjusts the CellSize so that more of the x/y axis fits into each cell of the console.

If you zoom out far enough points in the scatter graph will disapear, that is because the cell size is large enough that multiple points render into the same cell of the UI.  A Series could deal with that by changing the color (like a heat map) but that's really another issue.

image

To simplify this for the API user (e.g. when implementing a new Series / Annotation) there are the following methods:

```csharp
/// Returns the section of the graph that is represented by the given
/// screen position
public RectangleD ScreenToGraphSpace (int col, int row)

/// Calculates the screen location for a given point in graph space.
public Point GraphSpaceToScreen (PointD location)
```

I am trying to make sure that the mapping between graph space and screen space is deterministic and not subject to rounding (i.e. no adjacent screen cells ever return overlapping rectangles).

The `decimal` class is exact while the `float` (and `double`) are subject to rounding:
https://stackoverflow.com/questions/618535/difference-between-decimal-float-and-double-in-net

BDisp:
Although the decimal type is more precise is too much slower, so you were having this problem of slowness. I think you will have to manipulate these values so that they are within a certain tolerance. Microsoft will never want to change for decimal because they need the floats for scientific purposes :-)

tznind:
Ok I can have another look at making `float` work... it might be possible to keep the decimal aspect internal to the class and round things on the way out or rounding these probalematic boundaries in a consistent way.

tznind:
Ok I have managed to get `float` working properly.  I have scrapped the `SeriesDrawMode` and simplified the API for `ISeries` and `IAnnotation` implementers.

I've also tidied up elements that were drawing into the margin (still inside the control but into the area reserved for title/axis labels) 

I've added a lot of tests for `BarSeries` and `Axis` etc.  The classes are designed for inheritance (so you can customise how bars render) so implmenenting `FakeBarSeries` etc in tests was quite easy.  I went with Fake as a prefix since there are a lot of other tester classes with that prefix.

Finally I have split the big file up and put everything except `GraphView` into the namespace `Graphs` (and the files into Core folder).  This is as suggested by @BDisp

Terminal.Gui/Core/Graphs/Annotations.cs
```diff
@@ -0,0 +1,325 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
```
tig, line 3: I have a feeling having this project depend on `System.Drawing` is a bad idea:

image

from https://docs.microsoft.com/en-us/dotnet/api/system.drawing?view=net-5.0

I'm pretty sure the classes you are actually using (e.g. `RectangleF`) are fine, but there's a risk.

I think we should ask @migueldeicaza  for an opinion on which option is best:

0) Add the `System.Drawing.Primitives` nuget package, which includes all of these primitives. It continues to use the same namespace. https://www.nuget.org/packages/System.Drawing.Primitives/4.3.0?_src=template
1) Just use `System.Drawing` and continue to be careful not to pull in classes that actually have a dependency on GDI+.
2) Fork the classes we want from `System.Drawing` into `/Terminal.Gui/Types`
3) Take a dependency on another floating point library with similar functions.

My reading indicates option 0 is the best, but I don't fully understand what DLLs end up getting used in the end and if this actually provides an alternative to using the full `System.Drawing` assemblies.

I read through this https://github.com/dotnet/runtime/issues/21980 where @migueldeicaza participated and the answer is still not clear to me.

tig (changes requested):
@migueldeicaza - can you please make a recommendation. Graphs needs `PointF` and `RectangleF`.  What option mentioned in the code review is the best?

Terminal.Gui/Core/Graphs/Annotations.cs
migueldeicaza: These are good concerns and challenges.

What we have done is copy the types from System.Drawing into the code (see Terminal.Gui/Types/Size.cs., Rectangle, Point for example).   So we should do the same in this case.

tznind: Thanks for the guidance.

I have copied the code from https://github.com/dotnet/corefx/tree/master/src/System.Drawing.Primitives/src/System/Drawing into the Types dir.  

I made the following changes

- Changed namespace to `Terminal.Gui` 
- Updated cref in comments to refer to `Terminal.Gui` version
- Removed `readonly` keyword from the private properties (which required csharp 8)
- Implement GetHashCode (I just copied the definition from Rectangle/Point/Size - adding ).

migueldeicaza: Excellent!

tig (approved):

tig:
@tznind Would you please make some sort of minor change to this PR and do a new push just to kick off the CI/CD. I want to see the code coverage report the new YAML generates before I accept this.

Thanks.

Terminal.Gui/Core/Graphs/Annotations.cs
```diff
@@ -0,0 +1,324 @@
...
+				linesDrawn++;
+
+				// Legend has run out of space
+				if (linesDrawn >= availableHeight) {
+					break;
+				}
+			}
+		}
+
+		private string TruncateOrPad (string text, int width, TextAlignment alignment)
```
tig, line 202: I'd like to see a unit test for this function.

Terminal.Gui/Core/Graphs/Annotations.cs
```diff
@@ -0,0 +1,324 @@
...
+		public void AddEntry (GraphCellToRender graphCellToRender, string text)
+		{
+			entries.Add (Tuple.Create (graphCellToRender, text));
+		}
+	}
+
+	/// <summary>
+	/// Sequence of lines to connect points e.g. of a <see cref="ScatterSeries"/>
+	/// </summary>
+	public class PathAnnotation : IAnnotation {
```
tig, line 249: This has enough complexity to justify some unit tests.

Terminal.Gui/Core/Graphs/Series.cs
```diff
@@ -0,0 +1,323 @@
...
+
+		}
+
+	}
+
+
+	/// <summary>
+	/// Collection of <see cref="BarSeries"/> in which bars are clustered by category
+	/// </summary>
+	public class MultiBarSeries : ISeries {
```
tig, line 63: Please add some unit tests. We have 0% coverage for this class.

tig (changes requested):
In the spirt of constantly trying to raise the bar on our unit test code coverage, I'd like you to add a bit more before we merge this.

Use Fine Code Coverage to view all of your classes and note where your numbers are low (< 70-ish%). Look for code with obvious complexity and see if you can think up some good unit tests that would stress it.

I've made a few specific recommendations in my code review.

Thanks @tznind !!!!

tig:
Also, can you please fix the XML doc errors? I think there may be a way to do it without having to add docs to the original files, but I'm not sure. It may just be easier to add some documentation.

UnitTests/GraphViewTests.cs
```diff
@@ -0,0 +1,1013 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Terminal.Gui;
+using Xunit;
+using Terminal.Gui.Graphs;
+using Point = Terminal.Gui.Point;
+using Attribute = Terminal.Gui.Attribute;
+using System.Text;
+
+namespace UnitTests {
```
tig, line 11: Somehow this got changed back from `Terminal.Gui.Views`. Please change the namespace to `Terminal.Gui.Views`.

tig (changes requested):
If you can fix the unit test namespace thing real quick, I think we can still get this into 1.0.0 ;-)

Terminal.Gui/Core/Graphs/Annotations.cs
```diff
@@ -0,0 +1,325 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+
+namespace Terminal.Gui.Graphs {
```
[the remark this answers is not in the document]
tznind: I did consider it but `IAnnotation` and `ISeries` and the other classes are set up for inheritance e.g. to allow API users to declare their own custom series later on (e.g. stacked bar charts, area charts etc).  So having to prefix every reference to any of these interfaces/instances with `GraphView.` would produce quite a clutter.

Having separate files for each of these classes makes review and future PRs easier and reduces chances of merge conflicts.  If they were nested classes every file would  have to start with `partial class GraphView` which would add complexity and uncertainty to anyone reviewing the code.

If you think I made the wrong call let me know and I can change it.

UnitTests/GraphViewTests.cs
```diff
@@ -8,7 +8,7 @@
 using Attribute = Terminal.Gui.Attribute;
 using System.Text;
 
-namespace UnitTests {
+namespace Terminal.Gui {
```
BDisp: I think it should be `Terminal.Gui.Views` now.

UnitTests/TabViewTests.cs
```diff
@@ -7,7 +7,7 @@
 using Xunit;
 using System.Globalization;
 
-namespace UnitTests {
+namespace Terminal.Gui {
```
BDisp: I think it should be `Terminal.Gui.Views` now.

UnitTests/TableViewTests.cs
```diff
@@ -7,7 +7,7 @@
 using Xunit;
 using System.Globalization;
 
-namespace UnitTests {
+namespace Terminal.Gui {
```
BDisp: I think it should be `Terminal.Gui.Views` now.

UnitTests/TreeViewTests.cs
```diff
@@ -6,7 +6,7 @@
 using Terminal.Gui;
 using Xunit;
 
-namespace UnitTests {
+namespace Terminal.Gui {
```
BDisp: I think it should be `Terminal.Gui.Views` now.

UnitTests/GraphViewTests.cs
tznind: Fixed.  Sorry I was behind on `main` branch a bit so hadn't seen the namespace changes in tests so updated Tab/Tree too (hence d7e33c4) but have merged latest main branch and fixed it correctly this time in c8f197c

BDisp (approved):

tznind:
I think I have addressed all the points @tig .  Let me know if I missed any.  

I have added a public property `Contents` to `FakeDriver` (single line addition) which allows tests to evaluate whether the 'console' matches expectations after a render.  I think this makes for more understandable tests for these very graphical (and not interactive) components like annotations and series.  Let me know if there is some hidden danger in this or it is the wrong approach:

```csharp
[Fact]
public void PathAnnotation_Box()
{
	var gv = GraphViewTests.GetGraph ();

	var path = new PathAnnotation ();
	path.Points.Add (new PointF (1, 1));
	path.Points.Add (new PointF (1, 3));
	path.Points.Add (new PointF (6, 3));
	path.Points.Add (new PointF (6, 1));

	// list the starting point again so that it draws a complete square
	// (otherwise it will miss out the last line along the bottom)
	path.Points.Add (new PointF (1, 1));

	gv.Annotations.Add (path);
	gv.Redraw (gv.Bounds);

	var expected =
@"
 │......
 ┤.    .
 ┤......
0┼┬┬┬┬┬┬┬┬
 0    5";

	GraphViewTests.AssertDriverContentsAre (expected);

}
```

BDisp:
That's nice @tznind. I already had thought on something like this to see the `Redraw `output. Many thanks.

tig:
This is a fantastic addition to the project. Thank you for your investment in Unit Testing too!

Delighted to be able to merge this in as part of 1.0.0!!!

tznind:
Thanks for your efforts reviewing too definetly pushed me to get float working properly and is a better feature as a result.

PR #1203, opened 2021-04-10
Fixes #1202. CheckBox now deals with a functional '_' underscore hotkey.

### LINKED ISSUE #1202: CheckBox fail using only upper letter in hotkeys which don't work. Should be the '_' char.

BDisp:


tig (approved):
Groovy. Thanks!

Issue #1204, opened 2021-04-12
Jalaali(Persian) Date input support


inamvar:
It's good to support  Jalaali(Persian) date.   User can input date text in Jalaali  and  DateFiled  should parse it to Georgian date. it's very useful feature  for whom using  Jalaali calendar system(it's  about 140 millions people).

tig:
I know nothing about Persian date formats. 

Is there no built-in capability in .NET for handling them? In reviewing the PR for this (#1205) I see you've added a specific flag to the API. On the surface, this appears clunky to me. 

Are you sure there's no way to add this support in a way that would ALSO add support for any other important but less common date input formats?

What we want to avoid is having the API need to support more and more flags like this over time. If Persian is the only outlier that requires this then I'm all for the solution you've provided!

inamvar:
> I know nothing about Persian date formats.
> 
> Is there no built-in capability in .NET for handling them? In reviewing the PR for this (#1205) I see you've added a specific flag to the API. On the surface, this appears clunky to me.
> 
> Are you sure there's no way to add this support in a way that would ALSO add support for any other important but less common date input formats?
> 
> What we want to avoid is having the API need to support more and more flags like this over time. If Persian is the only outlier that requires this then I'm all for the solution you've provided!

 You can learn about Persian date [here](https://en.wikipedia.org/wiki/Iranian_calendars). more that 150 millions people are using this calendar. As I know there is no built-in input for that in .NET and just has a [PersianCalendar](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.persiancalendar?view=net-5.0) class  that converts `Georgian` date into `Persian` date but no input control is available. I agree that is not a good way to set a flag for each type of format. I can change it to new class and call it 'PersianDateField' but it needs to be derived from `DateField` class and override some of it's methods. the problem is `DateField` methods are not `virtual` and can not be overridden.  do you have any idea to ignore flag parameter?

tig:
Got it. 

I imagine over time there may be more controls that take date input/display dates...

I wonder if we should add a flag to a base class, e.g. `View` that gets set once to indicate Persian date formatting is in use. This could also be set automatically if it's possible to detect the user's computer-wide settings?

Now that I mention it, are you sure there's no way to detect this is a user's preference?

inamvar:
> Got it.
> 
> I imagine over time there may be more controls that take date input/display dates...
> 
[...]

PR #1205, opened 2021-04-12
Jalaali (Persian) Date input support

inamvar:
resolving  issue #1204

### DIFF
--- a/Terminal.Gui/Views/DateField.cs
+++ b/Terminal.Gui/Views/DateField.cs
@@ -5,10 +5,10 @@
 //
 // Licensed under the MIT license
 //
+using NStack;
 using System;
 using System.Globalization;
 using System.Linq;
-using NStack;
 
 namespace Terminal.Gui {
 	/// <summary>
@@ -25,6 +25,7 @@ namespace Terminal.Gui {
 		string sepChar;
 		string longFormat;
 		string shortFormat;
+		bool isJalaali;
 
 		int FieldLen { get { return isShort ? shortFieldLen : longFieldLen; } }
 		string Format { get { return isShort ? shortFormat : longFormat; } }
@@ -47,9 +48,11 @@ namespace Terminal.Gui {
 		/// <param name="y">The y coordinate.</param>
 		/// <param name="date">Initial date contents.</param>
 		/// <param name="isShort">If true, shows only two digits for the year.</param>
-		public DateField (int x, int y, DateTime date, bool isShort = false) : base (x, y, isShort ? 10 : 12, "")
+		/// <param name="isJalaali">If true, parse will convert jalaali input fo georgian date</param>
+		public DateField (int x, int y, DateTime date, bool isShort = false, bool isJalaali = false) : base (x, y, isShort ? 10 : 12, "")
 		{
 			this.isShort = isShort;
+			this.isJalaali = isJalaali;
 			Initialize (date);
 		}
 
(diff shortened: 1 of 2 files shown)


Terminal.Gui/Views/DateField.cs
```diff
@@ -47,9 +48,11 @@ public class DateField : TextField {
 		/// <param name="y">The y coordinate.</param>
 		/// <param name="date">Initial date contents.</param>
 		/// <param name="isShort">If true, shows only two digits for the year.</param>
-		public DateField (int x, int y, DateTime date, bool isShort = false) : base (x, y, isShort ? 10 : 12, "")
+		/// <param name="isJalaali">If true, parse will convert jalaali input fo georgian date</param>
+		public DateField (int x, int y, DateTime date, bool isShort = false, bool isJalaali = false) : base (x, y, isShort ? 10 : 12, "")
```
tig, line 52: I propose that you move `bool isJalaali` out of `DateField` into `Application` and make it an app wide setting.

That way we avoid cluttering constructors with more parameters and we allow for additional controls that need Jalaali support.

What do you think of that?

tig (changes requested):

Terminal.Gui/Views/DateField.cs
inamvar: it's good idea, you mean I move `isJalaali' as a static field  into `Application.cs` and use it inside the `DateField` class like this: 
```
if(Application.IsJallali == true){
....
}
```
is that  right?

tig: Yes.

inamvar: > Yes.

what if we need to use both Georgian and Jalaali inputs date in an application at the same time? in some situations we have a form that needs to enter multiple date inputs in different formats.

tznind:
Is this something that could be done with inheritance?  By making `GetShortFormat` and maybe the `Date` property virtual (perhaps combined with making `shortFormat` and `longFormat` public (or readonly/protected)?

Then users could create their own custom date formats and calendars types? Could even have a Jalaali version of DateField (either in core or Scenarios) to show how to implement.

inamvar:
> Is this something that could be done with inheritance? By making `GetShortFormat` and maybe the `Date` property virtual (perhaps combined with making `shortFormat` and `longFormat` public (or readonly/protected)?
> 
> Then users could create their own custom date formats and calendars types? Could even have a Jalaali version of DateField (either in core or Scenarios) to show how to implement.

Sounds good! It needs to change some methods identifiers in `DateField` class  and set them as `virtual` methods  to be able to override them in the inherited class.

Terminal.Gui/Views/DateField.cs
tig: Good point.

tig:
> > Is this something that could be done with inheritance? By making `GetShortFormat` and maybe the `Date` property virtual (perhaps combined with making `shortFormat` and `longFormat` public (or readonly/protected)?
> > Then users could create their own custom date formats and calendars types? Could even have a Jalaali version of DateField (either in core or Scenarios) to show how to implement.
> 
> Sounds good! It needs to change some methods identifiers in `DateField` class and set them as `virtual` methods to be able to override them in the inherited class.

this seems like a good approach.

Terminal.Gui/Views/DateField.cs
tig, line 52: A few unit tests that prove the various constructors work as they should would be good.

Terminal.Gui/Views/DateField.cs
```diff
@@ -377,6 +410,28 @@ public virtual void OnDateChanged (DateTimeEventArgs<DateTime> args)
 		{
 			DateChanged?.Invoke (args);
 		}
+
+		string ToJalaaliString (string format, DateTime date)
```
tig, line 414: You should have a unit test for this.

tig (changes requested):
@inamvar, 

We currently have zero unit tests for `DateField`. Via other tests, it has about 53% code coverage.

I'd like to ask you to add a `DateFieldTest.cs` module with some tests that would get this closer to 80%. Would you please do this as part of this PR?

Examples: 
- the constructors
- functions like `GetLongFormat`, 'GetDate', etc...
- anywhere you see any sort of logic that maniuplates stuff

Thanks!!!!

tig:
Hi @inamvar - We'd love to integrate this PR. Will you have a chance soon to address the changes requested?

tig:
@inamvar - are you still interested in helping with this?

tig:
Too stale. Closing.

PR #1212, opened 2021-04-15
Fixes #1211. Added support to TextView for word based operations Ctrl+Del and Ctrl+Backspace.

### LINKED ISSUE #1211: TextView support for word based operations Ctrl+Del and Ctrl+Backspace
I'm loving the improvements to TextView including Ctrl + left/right moving to next/previous word.

But it would be nice if this also extended to del and backspace (if not too hard).

textViewCtrl
_Deleting to end/begining of word with Ctrl+Del / Ctrl+Backspace in Kate text editor_

BDisp:


BDisp:
I don't know why it fails.

BDisp:
In my PC this passes `Assert.Equal ("This is the first line.\r\n", _textView.Text);`, but it should be `"This is the first line.\nThis is the second line."`. I'll see better tomorow. I don't know why I have to put the `'\r'` to it pass in my PC and naturally on the `Github` it does not passed. Any help please.

tznind:
> 
> 
> In my PC this passes `Assert.Equal ("This is the first line.\r\n", _textView.Text);`, but it should be `"This is the first line.\nThis is the second line."`. I'll see better tomorow. I don't know why I have to put the `'\r'` to it pass in my PC and naturally on the `Github` it does not passed. Any help please.

Windows will be using `\r\n` while linux will be using `\n`.  You can probably change your test to 

```csharp
Assert.Equal ("This is the first line."+Environment.NewLine , _textView.Text);
```

BDisp:
Thank you very much for your help @tznind .

BDisp:
Strange behavior with unit tests which sometimes don't throws any error under debug mode and only throws when running without debug from the `Test Explorer` window. Is there any explanation to this?

tig:
> Strange behavior with unit tests which sometimes don't throws any error under debug mode and only throws when running without debug from the `Test Explorer` window. Is there any explanation to this?

I pulled the PR down locally and I can't repro this.

BDisp:
There are no errors now. It was the last before. Try to delete the last commit ant test please. Thanks.

Issue #23, opened 2018-01-11
Control: masked text input


migueldeicaza:

jmperricone:
This was implemented in #1230 and #1261.

PR #1215, opened 2021-04-17
Feature #23 Feature TextMaskField

jmperricone:
mask

Features:
- Mask
  - Mask char can by changed. Default : ( _ ).
  - Charset
    - 0 : Number (Required)
    - 9 : Number
    - L : Letter (Required)
    - ? : Letter
    - A: Alphanumeric (Required)
    - a: Alphanumeric
  - Each character has an associated regex. When the field is generated, it joins everything together and compiles it.
  - On input use the regex of the char to accept the input, and after input use the joined regex to validate the field.

- Regex
  - Pattern matching, validating on input.

- TextAlignment
  - Left
  - Centered
  - Right
  
- Events
  - TextChanged
  - OnValidating

- Keys
  - Left
  - Right
  - Home
  - End
  - Delete
  - Backspace

Issue #23 

**Hope you like it!**

### DIFF
--- a/Terminal.Gui/Views/TextMaskField.cs
+++ b/Terminal.Gui/Views/TextMaskField.cs
@@ -17,16 +17,6 @@ namespace Terminal.Gui {
 
 	/// <summary>
 	/// TextField with Mask
-	/// 
-	/// Mask:
-	///	000-9000
-	/// Input:
-	///	123- 456
-	/// Output:
-	///	123456
-	///	123 456
-	///	123-456
-	///	123- 456
 	/// </summary>
 	public class TextMaskField : View {
 
@@ -102,7 +92,7 @@ namespace Terminal.Gui {
 
 		void Initialize ()
 		{
-			this.Width = masktype == TextMaskType.Mask ? text.Count : 20;
+			this.Width = masktype == TextMaskType.Mask ? text.Count : 20; // default width.
 			this.Height = 1;
 			this.CanFocus = true;
 		}
@@ -179,7 +169,7 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
-		/// 
+		///  
 		/// </summary>
 		public Rune EmptyRune { get; set; } = ' ';
 
@@ -189,7 +179,7 @@ namespace Terminal.Gui {
 		public Rune InputRune { get; set; } = '_';
 
 		/// <summary>
-		/// If char is not set, shows the mask. If this is set to false it shows <see cref="InputRune"/>.
+		/// If set, shows the mask. If this is set to false it shows <see cref="InputRune"/>.
 		/// </summary>
 		public bool ShowMaskOnEmpty { get; set; } = false;
 
@@ -276,7 +266,7 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
-		/// Moves cursor to the left. In mask mode, skip until it finds the an editable char.
+		/// Moves cursor to the left. In mask mode, skip until it finds an editable char.
 		/// </summary>
 		/// <returns>True if moved.</returns>
 		bool DecCursorPosition ()
--- a/Terminal.Gui/Views/TextMaskField.cs
+++ b/Terminal.Gui/Views/TextMaskField.cs
@@ -1,0 +1,542 @@
+﻿//
+// TextMaskField.cs: single-line text editor with Emacs keybindings
+//
+// Authors:
+//   Miguel de Icaza (miguel@gnome.org)
+//
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+using System.Linq;
+using System.Text.RegularExpressions;
+using NStack;
+
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// TextField with Mask
+	/// 
+	/// Mask:
+	///	000-9000
+	/// Input:
+	///	123- 456
+	/// Output:
+	///	123456
+	///	123 456
+	///	123-456
+	///	123- 456
+	/// </summary>
+	public class TextMaskField : View {
+
+		/// <summary>
+		/// Mask Type 
+		/// </summary>
+		public enum TextMaskType {
+			/// <summary>
+			/// Validates input with <see cref="MaskRunes"/>.
+			/// </summary>
+			Mask,
+			/// <summary>
+			/// Validetes input with a Regex Pattern.
+			/// </summary>
+			Regex
+		}
+
+		Regex regex;
+		List<Rune> text;
+		List<Rune> mask;
+		TextMaskType masktype;
+		bool error = false;
+
+		/// <summary>
+		/// When masktype == regex and text is right aligned, cursorPosition goes from 1 to x, leaving 0 for insert.
+		/// </summary>
+		int cursorPosition;
+
+		/// <summary>
+		/// Raised when the <see cref="Text"/> of the <see cref="TextMaskField"/> changes.
+		/// </summary>
+		public event Action<ustring> TextChanged;
+
+		/// <summary>
+		/// Raised when the validation of the <see cref="TextMaskField"/> occurs.
+		/// </summary>
+		public event Action<bool> OnValidation;
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="TextMaskField"/> class using <see cref="LayoutStyle.Computed"/> positioning.
+		/// </summary>
+		protected TextMaskField ()
+		{
+			Initialize ();
+		}
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="TextMaskField"/> class using <see cref="LayoutStyle.Computed"/> positioning.
+		/// </summary>
+		/// <param name="mask">Mask</param>
+		public TextMaskField (ustring mask) : this (mask, ustring.Empty) { }
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="TextMaskField"/> class using <see cref="LayoutStyle.Computed"/> positioning.
+		/// </summary>
+		/// <param name="mask">Mask or Regex Pattern</param>
+		/// <param name="masktype">Mask or Regex</param>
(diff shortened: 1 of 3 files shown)


BDisp:
That's really great. Thanks.

jmperricone:
mask2

Added **bool RegexValidateInput** .  
Setting it to false validates after input.. allowing any input.

Use **bool IsValid** to check.

BDisp:
Amazing.

UnitTests/TextMaskFieldTests.cs
```diff
@@ -0,0 +1,75 @@
...
+
+		[Fact]
+		public void Validate_Regex_Field_RegexValidateInput_False ()
+		{
+			// Allow input. Validates field.
+			_textMaskField_Regex.RegexValidateInput = false;
+
+			_textMaskField_Regex.ProcessKey (new KeyEvent ((Key)(int)'E', new KeyModifiers ()));
+			Assert.Equal ("E", _textMaskField_Regex.Text);
+			Assert.False (_textMaskField_Regex.IsValid);
```
tig, line 24: Seems to me a few more tests to exhaustively prove you are filtering bad (dangerous?) regex might be in order.

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+		/// Valid mask characters and regex validator.
+		/// </summary>
+		protected Dictionary<Rune, string> MaskRunes { get; set; } = new Dictionary<Rune, string> {
+			['0'] = "[0-9]",
+			['9'] = "[0-9\\s]?",
+			['L'] = "[a-zA-Z]",
+			['?'] = "[a-zA-Z\\s]?",
+			['A'] = "[0-9a-zA-Z]",
+			['a'] = "[0-9a-zA-Z\\s]?"
+		};
```
tig, line 464: What if I wanted a field that only accepted non-alpha and non-numeric characters? e.g. `@#$%^&*`.

I suppose I could use a regex?

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+		/// </summary>
+		public bool ShowMaskOnEmpty { get; set; } = false;
+
+		/// <summary>
+		/// Validates input on MaskType.Regex.
+		/// </summary>
+		public bool RegexValidateInput { get; set; } = true;
+
+		///inheritdoc/>
+		public override void PositionCursor ()
```
tig, line 192: It is not immediately obvious what you are doing here. If it were me, I'd creates some unit tests that clearly show intent (and ensure intent doesn't break).

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+			for (int i = 0; i < margins.right; i++) {
+				Driver.AddRune (' ');
+			}
+		}
+
+		/// <summary>
+		/// Moves cursor to the left. In mask mode, skip until it finds an editable char.
+		/// </summary>
+		/// <returns>True if moved.</returns>
+		bool DecCursorPosition ()
```
tig, line 277: Add unit test?

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+				break;
+			}
+			return moved;
+		}
+
+		/// <summary>
+		/// Moves cursor to the right. In mask mode, skip until it finds an editable char.
+		/// </summary>
+		/// <returns>True if moved.</returns>
+		bool IncCursorPosition ()
```
tig, line 307: Add unit test?

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+				if (cursorPosition < text.Count) {
+					moved = true;
+					cursorPosition += 1;
+				}
+				break;
+			}
+			return moved;
+		}
+
+		void BackspaceKeyHandler ()
```
tig, line 333: Add unit test?

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+					if (cursorPosition > 0) {
+						text.RemoveAt (--cursorPosition);
+						TextChanged?.Invoke (Text);
+					}
+				}
+				break;
+			}
+		}
+
+		void DeleteKeyHandler ()
```
tig, line 357: Add unit test?

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+
+		/// <summary>
+		/// This characters must be escaped in the regex pattern. 
+		/// </summary>
+		static readonly List<Rune> RegexSpecialsChars = new List<Rune> { '.', '$', '^', '{', '[', '(', '|', ')', '*', '+', '?', '\\' };
+
+		/// <summary>
+		/// Compiles the regex pattern for validation. <see cref="Validate(List{Rune})"/>
+		/// </summary>
+		void CompileMask ()
```
tig, line 474: If it were me, I'd add a big unit test for this and challenge myself to feed it at least a dozen test cases. Some with just random jibberish. Some with intentional boundary cases. Etc... A lot could go wrong here with someone mucking with this code someday.

UnitTests/TextMaskFieldTests.cs
```diff
@@ -0,0 +1,75 @@
+﻿using Xunit;
```
tig, line 1: Kudos for these tests.

tig (changes requested):
This is an AWESOME new capability. Thank you so much for doing it and contributing. 

Most of my comments are just nits, you've done a great job keeping this clean and smiple. It does just want it needs to and no more. I really appreciate that you provided some obvious "no code" built-in functionality AND provided the ultra-powerful regex capability. Well done.

The `Terminal.Gui` project is still woefully weak in unit tests and I'm on mission to push folks to change that. I think you've done the bare minimum here and want to challenge you to raise the bar as a demonstration to other contributors (including me!) of how it should really be done. Are you up for it?

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+
+		/// <summary>
+		/// Raised when the validation of the <see cref="TextMaskField"/> occurs.
+		/// </summary>
+		public event Action<bool> OnValidation;
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="TextMaskField"/> class using <see cref="LayoutStyle.Computed"/> positioning.
+		/// </summary>
+		protected TextMaskField ()
```
tig, line 61: I believe this needs to be public. We require all `View` classes to have a parameterless public constructor.

Like this

image

Without this `AllVeiwsTester` crashes.

I think all unit test suites should have tests for each constructor, FWIW.

tig (changes requested):
Crashes `AllViewTester`

tig:
I changed the Scenario code to this and it did not work as I expected:

image

And the app crashes if you attempt to type after doing this. 

I then did this and, again, was surprised with the result:
image

I think it's debatable whether setting the `Text` property should work exactly as if the user typed the contents, or if the caller has to "pre-compiled" the result.

Likewise, getting the value of the `Text` property COULD return exactly what the customer typed or the "compiled" result.

Can you explain why you chose to do it they way you did? I think there are probably valid reasons for doing it either way but I'm not convinced which should be default/primary. 

I note that .NET's version of this takes the approach you took: 

https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.maskedtextbox?view=net-5.0

FWIW, I also noticed that .NET provides a nice primitive that we COULD leverage (not saying we should, just ensure you are aware of it).

https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.maskedtextprovider?view=net-5.0

Terminal.Gui/Views/TextMaskField.cs
```diff
@@ -0,0 +1,535 @@
...
+				SetupText ();
+
+				SetNeedsDisplay ();
+			}
+		}
+
+		void SetupText ()
+		{
+			if (text != null) {
+				// TODO: Implement logic to match non-perfect inputs (mask == text)
```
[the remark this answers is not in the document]
jmperricone: ### What I meant by TODO was more of a proposal to improve if necessary.

Type Mask:
When an initial value is provided, it is validated with the mask 1:1.
Example:
```
        mask = "(000-9000)" // 0 (required) , 9 (optional)
```

Some cases (of course there are many more cases)
```
	input1 = "(123-4567)" // OK
	input2 = "(123- 567)" // OK
	input3 = "(123-456)"  // FAIL
	input4 = "123- 456"   // FAIL
	input5 = "123-456"    // FAIL
	input6 = "123456"     // FAIL
	input7 = "1234567"    // FAIL
```

This is the basic (current) implementation. 
	
```
       if (Validate (text) == false) {
           text = new List<Rune> ();
       }
```


I think that some of those cases for example should be valid and others not, but I find it difficult to find a validator that is logical for everyone and not only for the one who implemented it.

Now, the chances of validating something wrong as correct is zero with the current implementation.

jmperricone:
Hi tig

I saw that net provider and many more implementation but I thought about implementing this by my self.
Of course I can try to plug that provider if you think its a better option.

tig:
> Hi tig
> 
> I saw that net provider and many more implementation but I thought about implementing this by my self.
> Of course I can try to plug that provider if you think its a better option.

I think your implementation is awesome. I just wanted to make sure you were aware of it.

jmperricone:
Hi Tig, I reimplemented the whole thing around the same base. How do you want to do?
I can commit to this pull or make a new pull request ?

jmperricone:
Ok!

tig:
Closing. Dupe of #1230

PR #1218, opened 2021-04-18
Fix link and styling of Toplevel

ryandle:
1. Remove the extra `(` causing the link not to render.
2. Use backtick to style the Toplevel link similar to other links in the doc.


tig (approved):
Thank you!

PR #1230, opened 2021-04-20
Feature TextValidateField

jmperricone:
mask3

Hi. I explain what I did.

```csharp
public class TextValidateField<T> : View where T : ITextValidateProvider
```
We create the view specifying a provider that implements `ITextValidateProvider`.

```csharp
public interface ITextValidateProvider {
			bool Fixed { get; }
                        int Cursor(int pos);
			int CursorLeft (int pos);
			int CursorRight (int pos);
			int CursorStart ();
			int CursorEnd ();
			bool Delete (int pos);
			bool InsertAt (char ch, int pos);
			bool IsValid { get; }
			ustring Text { get; set; }
			ustring Mask { get; set; }
		}
```

I made 3 providers.

```csharp
public class NetMaskedTextProvider : ITextValidateProvider {} // .Net MaskedTextProvider  👍 
public class TextMaskProvider : ITextValidateProvider {} // My version of a Masked Input using Regex.
public class TextRegexProvider : ITextValidateProvider {} // Validates with Regex
```
Example use with NetMaskedTextProvider.
[Microsoft MaskedTextProvider Documentation](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.maskedtextbox.mask?view=net-5.0) Here are all the masking characters.
```csharp
var netProviderField = new TextValidateField<NetMaskedTextProvider> ("999 000 LLL >LLL AAA aaa") {
        X = 0
        Y = 0
};
Win.Add (netProviderField);
```
I have not done any tests yet, I want to know if we are on track.
Let me know what you think.
Round 2.


tig:
This is so much fun. What an awesome idea to make it extensible like this.

I'll take a closer look as soon as I get a chance!

tig:
Suggest you mark this as a draft PR until you have unit tests etc....

Ok?

jmperricone:
Hi, I want to add mouse tests. I need the width to be setted, to test with text alignment center and right.
Can this be done ?

jmperricone:
@tig  What do you think if we take out the mask provider I made, because it doesn't make sense to have both now.
Maybe save it as an example somewhere.

tig:
> @tig What do you think if we take out the mask provider I made, because it doesn't make sense to have both now.
> Maybe save it as an example somewhere.

I will leave the decision on that to you, sir!

UnitTests/TextValidateFieldTests.cs
```diff
@@ -0,0 +1,561 @@
...
+			var mask = "";
+			for (int i = 0; i < 255; i++) {
+				mask += (char)i;
+			}
+
+			try {
+				var field = new TextValidateField<TextRegexProvider> (mask) {
+					Width = 20
+				};
+			} catch (RegexParseException ex) {
```
tig, line 441: Please fix all build warnings.

UnitTests/TextValidateFieldTests.cs
```diff
@@ -0,0 +1,561 @@
+﻿using System.Diagnostics;
+using System.Text.RegularExpressions;
+using Terminal.Gui.TextValidateProviders;
+
+using Xunit;
+
+namespace Terminal.Gui {
```
tig, line 7: Change namespace to `Terminal.Gui.Views`.

UnitTests/TextValidateFieldTests.cs
```diff
@@ -0,0 +1,561 @@
+﻿using System.Diagnostics;
```
tig, line 1: Nice job!!

NetMaskedTextProvider - 85.4% Code Coverage
TextMaskProvider - 57.3%
TextRegExProvider - 81.9%
TextValidateField - 85.4%

Really impressive work! Thank you.

tig (changes requested):
Just a few nits that I'd like you to address before I merge... 

This is all really great work!

tig (approved):

tig:
Woot! Merging!

PR #1237, opened 2021-04-20
Fixes #1234. Setting Handled to true in the KeyPress event avoids ProcessKey from running.

### LINKED ISSUE #1234: Setting Handled to true in the KeyPress event don't avoids ProcessKey from running.
Try running this unit test and you will see that the text is changed in the `TextField` view because the `KeyPress` event won't be called.

```cs
[Fact]
public void KeyPress_Handled_To_True_Prevents_Changes ()
{
	Application.Init (new FakeDriver (), new FakeMainLoop (() => FakeConsole.ReadKey (true)));

	Console.MockKeyPresses.Push (new ConsoleKeyInfo ('N', ConsoleKey.N, false, false, false));

	var top = Application.Top;

	var text = new TextField ("");
	text.KeyPress += (e) => {
		e.Handled = true;
		Assert.True (e.Handled);
		Assert.Equal (Key.N, e.KeyEvent.Key);
	};
	top.Add (text);

	Application.Iteration += () => {
		Console.MockKeyPresses.Push (new ConsoleKeyInfo ('N', ConsoleKey.N, false, false, false));
		Assert.Equal ("", text.Text);

		Application.RequestStop ();
	};

	Application.Run ();
}

```

BDisp:


BDisp:
On my PC I don't have no errors now. Please someone can check this?

BDisp:
Fixed, see here why https://github.com/migueldeicaza/gui.cs/discussions/1229#discussioncomment-639779

PR #1242, opened 2021-04-21
Fixes #1241. Added SendKeys feature to the ConsoleDriver.

### LINKED ISSUE #1241: SendKeys feature to the ConsoleDriver would be nice.

BDisp:


BDisp:
Well, these unit tests are already giving up my head. They are all well on `Windows` and do not pass on `Linux`. I do not even know how to solve it, damn it.

BDisp:
Please verify this. If you run all tests at once it will fail. If you run the tests by file there is no errors as you can as can see. If you run in debug mode also does not give any error. I do not know how to work in this way. It looks like a bug.

unit-tests-behavior

BDisp:
Sometimes this still giving unit tests errors, but I think that because almost the tests are calling in parallel `Application.Run()`, `Application.RequestStop()` and `Application.Shutdown()` which is causing erroneous values.

tig:
> Sometimes this still giving unit tests errors, but I think that because almost the tests are calling in parallel `Application.Run()`, `Application.RequestStop()` and `Application.Shutdown()` which is causing erroneous values.

xUnit is setup so that our tests DO NOT run in parallel:

```c#
// Since Application is a singleton we can't run tests in parallel
[assembly: CollectionBehavior (DisableTestParallelization = true)]
```

I am sure this is working correctly, because if I change it to `false` many tests fail (as expected, because `Application` is a singleton).

However, the order in which the tests within each class run is not clear. It is possible to force an ordering (https://docs.microsoft.com/en-us/dotnet/core/testing/order-unit-tests?pivots=xunit).

Way back when I implemented the unit tests I tried to be very careful to ensure that, as long as `Run` was always coupled with a `Shutdown`, the state of things would be reset such that subsequent `Run/Shutdown` iterations would work as though it was the first run. I even implemented tests to ensure this didn't break (see `ApplicationTests`). 

Somewhere along the line since then, I believe changes have been introduced that undid this. Or I made a mistake.

Regardless, I think what we should do now is to deeply inspect the behavior to ensure that calling `Shutdown` fully cleans up.

BDisp:
@tig I realized that when I was forced to debug from the root of all the tests to detect the cause of the error and had to put up with all the Assert.Throws. Meanwhile I picked up some errors in other tests and when I checked the call stack, another test had already been started, so it was running in parallel.

tig:
> @tig I realized that when I was forced to debug from the root of all the tests to detect the cause of the error and had to put up with all the Assert.Throws. Meanwhile I picked up some errors in other tests and when I checked the call stack, another test had already been started, so it was running in parallel.

I wonder if I don't have it configured correctly, because I don't doubt you. 

I'll go read the docs again...

BDisp:
Do not worry about it @tig. It could have been some instability of Visual Studio itself. Now, fortunately passed :-)

BDisp:
Confirmed @tig. Now all the tests pass without fail, thank you.

BDisp:
@tig I'll change `Key.Null` to `'\0'` (null string). Is there any problem of changing `FakeConsole` `ReadKey` for `'\0'` instead of `'~'`?
I merged your PR. Is I better deleting it from this my PR to submit a new commit or maintain the merged?

BDisp:
The reason why I changed the `ReadKey` is that it fail on cursor position and text value by always entering the `'~'` char, without needed.

tig:
I'll look later.

-cek
Via mobile; expect brevity, typos, and dangerous driving.

BDisp:
In my opinion, I think you should submit another PR with the `ApplicationTests` fix, without renaming the namespace, because it is a nightmare to run the tests :-)

BDisp:
Sending a `Ctrl+Q` to quit.

WindowsDriver:
sendkeys-windowsdriver
NetDriver on Windows:
sendkeys-netdriver-windows
CursesDriver:
sendkeys-cursesdriver
NetDriver on Linux:
sendkeys-netdriver-linux

UnitTests/ConsoleDriverTests.cs
```diff
@@ -66,5 +68,164 @@ public void SetColors_Changes_Colors ()
...
+		public void FakeDriver_Always_Sends_ConsoleKey_Oem3_Without_Using_MockKeyPresses ()
+		{
+			Application.Init (new FakeDriver (), new FakeMainLoop (() => FakeConsole.ReadKey (true)));
+
+			var top = Application.Top;
+			var view = new View ();
+			var oem3 = 0;
+
+			view.KeyPress += (e) => {
+				Assert.Equal ('~', (uint)e.KeyEvent.Key);
```
tig, line 82: Did you miss this?

Terminal.Gui/ConsoleDrivers/FakeDriver/FakeConsole.cs
```diff
@@ -798,7 +798,7 @@ public static ConsoleKeyInfo ReadKey (bool intercept)
 			if (MockKeyPresses.Count > 0) {
 				return MockKeyPresses.Pop();
 			} else {
-				return new ConsoleKeyInfo ('~', ConsoleKey.Oem3, false,false,false);
```
tig, line 801: IIRC, I choose `~` over `\0` because it can be typed and will never be mistaken for a string terminator.

I think going with `\0` will likely work fine, but I don't understand why you think you need to change it. If you do change it, you need to change the `FakeDriver` tests as well (see other comment below) because they use `~`.

tig (changes requested):
This is pretty cool.

See my commetns.

Terminal.Gui/ConsoleDrivers/FakeDriver/FakeConsole.cs
BDisp: I want to iterate to keys events and and capture all that is being sending by `MockKeyPresses` or by `SendKeys`. With the `'\0'` the events won't be called and I can capture all with cleaning way.

UnitTests/ConsoleDriverTests.cs
BDisp: Yes, due the constant errors, I forgot to check this, sorry. I searched for '~' in the `UnitTests` and found none.

BDisp:
On my PC all unit tests pass.

unit-tests

tig:
I'm close to being able to merge #1235, which may fix this.

tig:
I'm a bit confused. Is this a duplicate PR? 

If not, can you describe the status?

BDisp:
> 
> 
> I'm a bit confused. Is this a duplicate PR?
> 
> If not, can you describe the status?

This PR was never merged I think.

PR #1243, opened 2021-04-22
Nested types for TabView and TableView

### LINKED ISSUE #1239: Use nested types for ColumnStyle, TableStyle, TabStyle, CellActivatedEventArgs, etc...
Like so:

image

Note this will be a breaking change, but I'm ok with it if we do it before we launch 1.0.

@tznind

tznind:
Fixes #1239


tig:
@tznind can you see why we have a merge conflict here?

tznind:
Looks like conflict comes from:

b29240f362b29b05e8f5e32a8b347433cf35a1c1

The Code Coverage pull request has a change to TableView.  Just looks like moving it's location int he file though nothing more:

https://github.com/migueldeicaza/gui.cs/pull/1235/files#diff-cabc06df9590a035dadf256277de6facda529d77c24898509033704c9f32c43bL146

Moving the class resulted in 2 copies of it, I have resolved conflict by deleting the extra copy

PR #1249, opened 2021-04-26
Fixes #1210.  Added AllowsReturn, AllowsTab and Multiline into the TextView.

### LINKED ISSUE #1210: How Tab and Shift+Tab keys should be performed on TextView?
- In my opinion `Tab` should insert the `char 9` and the `Shift+Tab` should retreat the `char 9` to the previous column, because on a multiline editor is normal to press the tab key to insert spaces and and Shift+Tab to remove spaces. So, if it goes to the next view, forces us to go to the `TextView` again causing stressing.
- Another solution is replacing the tabs to spaces, but if the original text has tabs it should be maintained. Perhaps an option to convert to spaces or vice-versa.
- For navigation `Ctrl+Tab` should go to the next view and `Ctrl+Shift+Tab` should go to the previous view.
- Please, give your opinions to makes this definitively clear. Thanks in advance.

### LINKED ISSUE #1234: Setting Handled to true in the KeyPress event don't avoids ProcessKey from running.

BDisp:
Fixes #1234


BDisp:
I had to merge the PR #1242 because there are unit test that only work with them.

BDisp:
`TextView` can be used as single line text box as well.

textview-tab

BDisp:
I also had to create a new word wrap API for the `preserveTrailingSpaces` to deal with tab looking for the text forward, because the backward way complicate that.

tig (approved):
I read through this and didn't find anything obvious, but the vastness of it makes me nervous.

I appreciate that you added some more unit tests. However, since all of the `*Driver` classes zero or close to zero code coverage we should all continue to be nervous. At some point someone should write a test suite for the drivers.

tig:
@BDisp after merging this, `main` no longer builds for me.

image

What did I do wrong?

BDisp:
> 
> What did I do wrong?

Because you didn't merged the `ConsoleDriver` `SendKeys` feature before.

tig:
it's working now. Thanks.

BDisp:
> I read through this and didn't find anything obvious, but the vastness of it makes me nervous.

The main reason why I did a new word wrapper api was that when it found a white space he go out from the while statement and I needed to find tabs before that.

tig:
Yea, I see that.

TextView is one of the more complex pieces of code in the project. It only has ~55% code coverage. 

If we can get it above 80% I think we'll all feel better. For example `WrapModel` is a very complex function that has no tests.

PR #1254, opened 2021-04-27
Fixes #931. Fixed the limit 25 lines issue.

### LINKED ISSUE #931: Linux terminal hangs after quit the GUI on Net5.0

BDisp:


tig (changes requested):
Can you please add some comments in the code that explain all of this?

Issue #1266, opened 2021-04-28
Ctrl+Tab not registering in xterm


tznind:
Is there a key combination that gets you out of TextView when AcceptsTab is true (default)?

I am testing in Linux and finding that in `Notepad` and `Tab View` Scenarios,  once you tab into a TextView you can't get out (without using mouse).

No combination of Ctrl/Alt/Shift + Tab are moving focus.

tig:
I don't think so.

BDisp:
That was working before, damn. I' ll check.

tznind:
> That was working before, damn. I' ll check.

Could be OS specific (have not tried on Windows).  I am using xterm-256color

BDisp:
In `Windows` is working. Is `Ctrl+Tab` for forwarding and `Ctrl+Shift+Tab` for backwarding.

tznind:
Hmn looking at `Scenario Keys` it seems like its something to do with my console.  Ctrl+Key works but Ctrl+Tab just sends Tab:

![ctrlTab](https://user-images.githubusercontent.com/31306100/116453497-de96a300-a856-11eb-8953-f86c75a6e05d.gif)
_Pressing Ctrl+D followed by Ctrl+Tab (which just registers as Tab)_

BDisp:
Yes, in WSL the tab is not caught anymore with the Tab key too.

BDisp:
Please try with the -usc for the NetDriver. It don't work too.

BDisp:
Oh damn, it always open with CursesDriver and not NetDriver.

BDisp:
I found that the ResetState is setting the UseSystemConsole to false,

BDisp:
@tig on `NetDriver` it's working but it's need to comment the `UseSystemConsole = false;` in the `ResetState` method

BDisp:
> Hmn looking at `Scenario Keys` it seems like its something to do with my console. Ctrl+Key works but Ctrl+Tab just sends Tab:

No it's not with your terminal. Is that because `nCurses` always only send `Tab` or `Shift+Tab`. This is a bad notice.

BDisp:
@tznind what terminal do you use? Mine is not printing the last key press as your is in red. It's blank.

BDisp:
I found this:
https://superuser.com/questions/787280/ctrltab-is-not-working-in-vim-with-gnome-terminal

BDisp:
> 
> 
[...]

PR #1271, opened 2021-04-29
TextView CursorLeft/CursorRight change focus when at start/end

tznind:
This PR makes CursorLeft / CursorRight (without modifier keys) change the focus out of TextView.  It does this by returning false from ProcessKey (i.e. no longer claim to handle CursorLeft when cursor is at 0,0).  If you test it out I think it feels quite natural and improves the usability of the control for both Windows and Linux users without changing any existing functionality (Ctrl+Tab taking you out of control)

This PR relates to #1266 where it is difficult to get focus out of TextView without reaching for the mouse on old/linux terminals (due to control keys not playing nicely with Tab key).

cursor-textview-focusshift

_Interacting with TextView using only the cursor keys (and typing)_

### DIFF
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -1944,6 +1944,25 @@ namespace Terminal.Gui {
 			int restCount;
 			List<Rune> rest;
 
+			// if the user presses Left (without any control keys) and they are at the start of the text
+			if(kb.Key == Key.CursorLeft && currentColumn == 0 && currentRow == 0) {
+				// do not respond (this lets the key press fall through to navigation system - which usually changes focus backward)
+				return false;
+			}
+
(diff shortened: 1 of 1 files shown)


BDisp:
Great. How about also using the CursorUp (if at 0) and CursorDown (if at bottom), only for `TextView` and `Window`?

tznind:
That sounds sensible (consistency is good).  I have looked at `Toplevel.ProcessKey` but feel a bit out of my depth there.  Are you able to make a PR for that? or copy the code from this one and make both the changes (and I can close this PR)?

BDisp:
No problem, I thought better and cursor up/down could lead to the previous and next view without the user need. Your PR is very good as is and I will try a fix for the windows navigation for Linux, later. Thanks.

tig (approved):
Nice!

Issue #48, opened 2018-02-16
Truecolors (direct RGB colors) support


XVilka:
# Terminal Colors

There exists common confusion about terminal colors. This is what we have right now:

- Plain ASCII
- ANSI escape codes: 16 color codes with bold/italic and background
- 256 color palette: 216 colors + 16 ANSI + 24 gray (colors are 24-bit)
- 24-bit true color: "888" colors (aka 16 milion)

```bash
printf "\x1b[${bg};2;${red};${green};${blue}m\n"
```

The 256-color palette is configured at start and is a 666-cube of colors,
each of them defined as a 24-bit (888 rgb) color.

This means that current support can only display 256 different colors in the
terminal while "true color" means that you can display 16 million different
colors at the same time.

Truecolor escape codes do not use a color palette. They just specify the
color itself.

This is a good test case:

```bash
printf "\x1b[38;2;255;100;0mTRUECOLOR\x1b[0m\n"
```

- or
  https://raw.githubusercontent.com/JohnMorales/dotfiles/master/colors/24-bit-color.sh
- or http://github.com/robertknight/konsole/tree/master/tests/color-spaces.pl
- or https://git.gnome.org/browse/vte/tree/perf/img.sh
- or just run this:

```sh
awk 'BEGIN{
    s="/\\/\\/\\/\\/\\"; s=s s s s s s s s;
    for (colnum = 0; colnum<77; colnum++) {
        r = 255-(colnum*255/76);
        g = (colnum*510/76);
        b = (colnum*255/76);
        if (g>255) g = 510-g;
        printf "\033[48;2;%d;%d;%dm", r,g,b;
        printf "\033[38;2;%d;%d;%dm", 255-r,255-g,255-b;
        printf "%s\033[0m", substr(s,colnum+1,1);
    }
    printf "\n";
}'
```

Keep in mind that it is possible to use both ';' and ':' as Control Sequence
Introducer delimiters.

According to Wikipedia[1], this behavior is only supported by xterm and konsole.

[1] https://en.wikipedia.org/wiki/ANSI_color

Since
[ncurses-6.0-20180121](http://lists.gnu.org/archive/html/bug-ncurses/2018-01/msg00045.html),
terminfo began to support the 24-bit True Color capability under the name of
"RGB". You need to use the "setaf" and "setab" commands to set the foreground
and background respectively.

# True Color Detection

There will be no reliable way to detect the "RGB" flag until the new release of
[...]

PR #1275, opened 2021-04-30
Fixes #219. Adds Underline and Blink

### LINKED ISSUE #219: Attributes: Support Bright/Bold, Dim, Underline, Blink, Reverse, Hidden, and Strike-through
Once https://github.com/gui-cs/Terminal.Gui/issues/2797 is done, we'll be read to add these:

- [ ] Bright/Bold
- [ ] Dim
- [ ] Underline
- [ ] Blink
- [ ] Reverse
- [ ] Hidden,
- [ ] Strike-through

jmperricone:
UNDERLINE

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -102,18 +102,28 @@ namespace Terminal.Gui {
 			//Console.Out.Flush ();
 
 			//Set cursor key to cursor.
-			Console.Out.Write("\x1b[?1l");
-			Console.Out.Flush();
+			Console.Out.Write ("\x1b[?1l");
+			Console.Out.Flush ();
 		}
 
 		public override void UpdateScreen () => window.redrawwin ();
 
 		int currentAttribute;
+		bool currentUnderlineAttribute;
+
+		static int A_UNDERLINE = 0x20000;
 
 		public override void SetAttribute (Attribute c)
 		{
 			currentAttribute = c.Value;
+			currentUnderlineAttribute = c.UnderLine;
 			Curses.attrset (currentAttribute);
+
+			if (c.UnderLine) {
+				Curses.attron (A_UNDERLINE);
+			} else {
+				Curses.attroff (A_UNDERLINE);
+			}
 		}
 
 		public Curses.Window window;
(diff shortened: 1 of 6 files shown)


jmperricone:
It works in WindowsDriver, CursesDriver and NetDriver.

The only thing that seems off to me is that I did not make ```Attribute.Value``` contain this value as well.

### **If someone is using this value assuming it only have color,  it would be very bad.**

NetDriver sets the biggest number.
```csharp
value = ((((int) f) & 0xffff) << 16) | (((int) b) & 0xffff)
```
with ConsoleColor.White = 15

```
value == 0xF000F
```

using something like 0x100000 for underline would be possible if I'm not wrong.


Tell me what you think. Is set this as a draft.

BDisp:
Good work. Any day we will have the #48 RGB working :-)
You can play with that limit values, you'll see very interesting. You can achieve a greater colors values. Only it'll need to change the default for each of `ColorScheme`. Otherwise you'll have identical colors.
It's possible to underline only words too?

BDisp:
I am pleased for my `CSI (Code Sequence Introducer)` implementation on the `NetDriver`. Too bad it's too slow on `Windows`.

jmperricone:
> I just realized that it will only work for windows, when using the netdriver. (@jmperricone)

Now that I read this, it seems to me that it is not clear.

It works with `CursesDriver` and the `WindowsDriver`.

What I meant is that with the `NetDriver`, it only works on Windows. This is what I did:

```csharp
void SetUnderline (bool state)
{
	redrawUnderline = state;
	if (IsWinPlatform == true) {
		var OutputHandle = NetWinVTConsole.GetStdHandle (NetWinVTConsole.STD_OUTPUT_HANDLE);
		var success = NetWinVTConsole.GetConsoleScreenBufferInfo (OutputHandle, out var csbi);
		if (success) {
			if (state == true) {
				csbi.wAttributes |= 0x8000;
			} else {
				csbi.wAttributes &= 0x7FFF;
			}
			NetWinVTConsole.SetConsoleTextAttribute (OutputHandle, csbi.wAttributes);
		}
	}
}
```

BDisp:
> What I meant is that with the NetDriver, it only works on Windows.

You don't need to use` Windows API` to work on `Windows`. The only reason I use `Windows API` was because the `System.Console` don't recognize `escape sequences` and I needed to use `ENABLE_VIRTUAL_TERMINAL_PROCESSING` for that. To do that, you need to use `escape sequences` `"\x1B[4m" to underline` and `"\x1B[0m" to reset`, both valid to `Windows` and `Linux`.

BDisp:
But is better to add the underline bitwise to the contents attribute than increase the contents length to save the underline.

BDisp:
@jmperricone, when you have the time please check my PR https://github.com/jmperricone/gui.cs/pull/1, please. Thanks.

jmperricone:
Hi, thanks for the pull #1 @BDisp 

About the bitwise underline, I thought about that:

> If someone is using this value assuming it only have color, it would be very bad.
https://github.com/migueldeicaza/gui.cs/pull/1275#issuecomment-829734583

Because `Driver.GetAttribute()` is
```csharp
public override Attribute GetAttribute ()
{
	return currentAttribute;
}
...
public static implicit operator Attribute (int v) => new Attribute (v);
...
```
Then Attribute.Value contains that

If that is not a problem, is ok!

BDisp:
> If that is not a problem, is ok!

It's okay because if Attribute.Underline is true it's possible to extract from the value.

jmperricone:
Should be that added to the `Attribute.Value` Property Get(). For example if someone have code like:
```csharp
if (attr.Value > 100) {
   ...
}
```
Having Underline set, will always return true.

BDisp:
> Having Underline set, will always return true.

Is true, but the user may check if `Underline` property is true and so the value isn't only colors, taking the desired action. But, normally, the user may take the colors by checking in the `Background` and `Foreground` colors.

tig:

tig:
This work should be merged into #2612. Closing.

PR #1283, opened 2021-05-05
Fixes #1266. Using Ctrl+PageUp/PageDown to allowing navigate through windows.

### LINKED ISSUE #1266: Ctrl+Tab not registering in xterm

BDisp:


BDisp:
@tznind with this PR, we do not need `TextView` to be at the beginning or at the end. What do you think?

tznind:
> 
> 
> @tznind with this PR, we do not need `TextView` to be at the beginning or at the end. What do you think?

Not sure I understand the question... do you mean 'should Ctrl+PageUp and Ctrl+PageDown fall through TextView ProcessKey (as unprocessed) all the time or only when the cursor is at beginning/end'?  

I think it would be sensible for it to work anywhere.  Its quite an obscure shortcut so if we are saying that it is an alternative (e.g. for linux users) to Ctrl+Tab then we should have it behave in the same way (i.e. work regardless of where cursor is in control).

BDisp:
> Not sure I understand the question... do you mean 'should Ctrl+PageUp and Ctrl+PageDown fall through TextView ProcessKey (as unprocessed) all the time or only when the cursor is at beginning/end'?

All the time, because `TextView` only use `PageUp/PageDown` alone or with the `Shift` key.



> I think it would be sensible for it to work anywhere. Its quite an obscure shortcut so if we are saying that it is an alternative (e.g. for linux users) to Ctrl+Tab then we should have it behave in the same way (i.e. work regardless of where cursor is in control).

I think we should consider it as a alternative to the `Ctrl+Tab` and `Shift+Ctrl+Tab`, because we can´t handle with that combination in Linux.

BDisp:
But the `TextView` should maintain the two alternatives to navigate forwards and backwards.

tig:
I can't tell if this is good to go or not. Please mark it draft if you don't think it's ready. Otherwise let me know and I'll CR it and merge.

BDisp:
To me I think it is good, but I would like the @tznind  opinion too, if he could too.

tznind:
> To me I think it is good, but I would like the @tznind opinion too, if he could too.

I am happy with this PR.

Only point I can contribute is that Ctrl+PageUp/Down does get swallowed by Firefox (and possibly other browsers) so if you are using a terminal via a browser (e.g. Guacamole or Chrome Remote Desktop) it still might not help.  But there are always going to be environmental issues with shortcuts in the terminal so I'm not too worried.

The only other approach I can think of would be to expose the navigation key to be user configurable (e.g. as a public field).  That is the approach I made with `ObjectActivationKey` and prevents this being a breaking change of users have rigged up Ctrl+PageUp to something else (e.g. as a status bar shortcut).

BDisp:
Thanks @tznind. I'll try to implement a configurable `Toplevel` key navigation.

tig (approved):

Issue #1286, opened 2021-05-06
TableView - MaxWidth is not honored on last column?


tig:
![image](https://user-images.githubusercontent.com/585482/117346969-4e132080-ae5d-11eb-9764-1deab39e360a.png)

```c#
var tableView = new TableView () {
	X = 0,
	Y = 0,
	Width = Dim.Fill (),
	Height = Dim.Fill (1),
};
var tableStyle = new TableView.TableStyle ();
tableStyle.AlwaysShowHeaders = true;
tableStyle.ShowHorizontalHeaderUnderline = true;
tableView.Style = tableStyle;

Win.Add (tableView);

var dt = new DataTable ();
tableView.Table = dt;

dt.Columns.Add (new DataColumn ("Scheme", typeof (string)));
tableView.Style.ColumnStyles.Add (tableView.Table.Columns ["Scheme"],
	new TableView.ColumnStyle () { MinWidth = 10, MaxWidth = 10, Alignment = TextAlignment.Left });

var attrs = typeof(ColorScheme).GetProperties ();
foreach (var prop in attrs) {
	dt.Columns.Add (new DataColumn (prop.Name, typeof (string)));
	tableView.Style.ColumnStyles.Add (tableView.Table.Columns [prop.Name], 
		new TableView.ColumnStyle () { MinWidth = 15, MaxWidth = 15, Alignment = TextAlignment.Left });
}
```

IMO, this code should result in the last most column being only 15 cells wide.

@tznind

tznind:
Yes, currently the rendered table always fills the full bounds of the `TableView` control.  This makes the last **visible** column always fill up to the edge of the control.  It is important to consider that the visible column(s) may not be all of them, for example:

If you have many columns e.g: 10 width; 10 width; 10 width; 50 width;  

And your control is width 40 then the first 3 columns would be rendered only and the third column would be increased to 20 width (because there is not enough space for the 50 width one).  Otherwise the table would appear to grow and shrink as you scrolled right.

We could detect whether horizontal scrolling comes into play or not and shrink the table if it doesn't (as is your example above)?

tznind:
We would also need to float the scroll bar (if vertical scrolling was happening) otherwise it would look wierd just hanging off to the right in thin air:

![image](https://user-images.githubusercontent.com/31306100/117376369-e7c3e780-aec8-11eb-8451-b7fec315a7bf.png)

BDisp:
[...]

PR #1289, opened 2021-05-07
TableView - adds last column dividing line

tznind:
There are a couple of issues I need to iron out (see demo) and a unit test to add but is this a suitable fix for #1286 ?

table-last-column

### DIFF
--- a/Terminal.Gui/Views/TableView.cs
+++ b/Terminal.Gui/Views/TableView.cs
@@ -302,7 +302,8 @@ namespace Terminal.Gui {
 						rune = Driver.TopTee;
 					}
 					// if the next console column is the lastcolumns end
-					else if (columnsToRender.Any (r => r.IsVeryLast && r.X + r.Width == c)) {
+					else if ( Style.EnforceMaxWidthOnLastColumn &&
+						 columnsToRender.Any (r => r.IsVeryLast && r.X + r.Width == c)) {
 						rune = Driver.TopTee;
 					} else if(c == availableWidth -1){
 						rune = Driver.URCorner;
@@ -337,7 +338,7 @@ namespace Terminal.Gui {
 				
 				Driver.AddStr(TruncateOrPad(colName,colName,current.Width ,colStyle));
 
-				if (current.IsVeryLast) {
+				if (Style.EnforceMaxWidthOnLastColumn && current.IsVeryLast) {
 					RenderSeparator (current.X + current.Width, row, true);
 				}
 			}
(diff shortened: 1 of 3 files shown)


BDisp:
But with an option if the last column width is not set, then it fill all the available width.

tig:
This meets my needs.

tznind:
Ok cool I will try to work out the issues and make it only behave this way when there is a `MaxLength` set but it might take me a few days to get to this.

tznind:
So MaxWidth is an `int` so always has a value (it also has a configurable default).  Rather than change this or impose some arbitrary logic I have just added a style setting `EnforceMaxWidthOnLastColumn`:

I have also added tests.  In the interests of 'no surprises' backwards compatibility I have left it as default false (do not have the last column).  If you think this should instead default to true let me know and I can change that.

enforceMaxWidth
_Toggling EnforceMaxWidthOnLastColumn in the TableEditor scenario in UICatalog_

BDisp:
How about if `MaxWidth` is equal to `-1` then the column is hidden, if equal to `0` (zero), automatically sets the `EnforceMaxWidthOnLastColumn` to true.

tznind:
So the way the table view works is:

- Take y rows of data (scroll offset + height - space taken by headers)
- Calculate width of first column (maximum width of the content of those rows)
   - Enforce MinWidth
   - Enforce MaxWidth
- If there is space for the column add it to the list of columns that will be rendered

This has several advantages
- As you scroll down a large table the columns auto resize to allow the content to fit
- We don't have to resolve the entire table into text in order to render it

So I'm not sure `MaxWidth` is exactly related to this last column feature.  Given that with small content the last column is going to be less width than `MaxWidth` anyway.

Maybe It would be better to rename this Style setting `Style.ExpandLastColumn`? (certainly that is a better name than `EnforceMaxWidthOnLastColumn` now I think about it).

tznind:
> Just a question, can it have a hidden column?
> 
> Thanks for your efforts.

Not in the `TableView` API.  But you can use `System.DataTable.RemoveColumn(DataColumn col)` to take a column out and you can add it back in again with `System.DataTable.AddColumn(DataColumn col)`.  So it would be very easy for the API user to keep a list of removed ones and simulate hiding/showing columns.

BDisp:
Understood. Many thanks @tznind :-)

BDisp:
> Well, that depends. If `EnforceMaxWidthOnLastColumn` can be used by any` Style`, it can also cover other styles.

Sorry @tznind, my bad. `EnforceMaxWidthOnLastColumn` already belong to `Style`. I failed that.

tig (approved):
Love.

PR #1292, opened 2021-05-08
Fixes #1291. Combining two PosAbsolute or two DimAbsolute result on a PosAbsolute or DimAbsolute.

### LINKED ISSUE #1291: Views - View.Height += 1  does not have the expected behavior
Before I use ```view.Height = view.Subviews.Count;```, I had it in another way where I did ```view.Height += 1;```
I remove a lot of other views with Dim and Pos to other views, thinking maybe was circular reference or something.
But no, the problem is just that line .... ```view.Height += 1;```

If you try this, it gets exponentially slower.

This is beacuse ```Dim.Combine```. After 3 times, 
```view.Height == {Dim.Combine(Dim.Combine(Dim.Combine(Dim.Absolute(0)+Dim.Absolute(1))+Dim.Absolute(1))+Dim.Absolute(1))}```

**Is this the expected behavior for this case ???** 

Maybe if it's the same instance, combine should behave differently, because I think that ```view.Height += x``` is something very basic to have such a behavior.

```csharp
public override void Setup ()
{
  var view = new View () { X = 0, Y = 0, Width = 20, Height = 0 };
  var field = new TextField () { X = 0, Y = Pos.Bottom (view), Width = 20 };
  
  field.KeyDown += (k) => {
	  if (k.KeyEvent.Key == Key.Enter) {
  
		  var label = new Label (field.Text) { X = 0, Y = view.Subviews.Count, Width = 20 };
		  
                  field.Text = "";
  
		  view.Add (label);
  
                  // works
		  view.Height = view.Subviews.Count;
                  
                  // don't try this at home
		  // view.Height += 1;
	  }
  };
  
  Win.Add (view);
  Win.Add (field);
}
```

BDisp:


tig (approved):
Nicely done.

PR #1294, opened 2021-05-10
Feature Slider

jmperricone:
Slider


BDisp:
Great work. I like it very much. Thanks.

tig:
1. I can't change the "Range" (the bottom-most demo slider) to be bigger/smaller once it's been set the first time. Should I be able to? I can change "LeftRange" and "RightRange" as expected.
1. Also, have you considered making the hit target for double click a bit broader, to include the dashes on each side of a ball? 
1. The Slider Options Editor is confusing. What does it do?
1. I assume I can somehow make the balls be at the very ends of a slider if I wanted that?
1. Horiz-sizing goofs up the cursor:

jmperricone:
Over an setted option with ctrl pressed you move the option.

jmperricone:
With the range type, if you hit enter in the center of the range, should move the left one to the right, or the right one to the left ?

tig:
> With the range type, if you hit enter in the center of the range, should move the left one to the right, or the right one to the left ?

Whichever one it is closer to. If it's a tie: both.

jmperricone:
You dont like it by pressing Ctrl and the arrows ?

jmperricone:
Ahh, you want that Enter makes something too, to be consitent.
Maybe once you press enter, the next arrow pressed moves the respective end.

tznind:
Loving this!  But for me the Range slider is confusing.

My first instinct was to move to the first radio button then hold shift to move to the end one (like selecting text in a text editor).  When that didn't work I wanted to hit Enter on the first and then scroll to the end and hit Enter again to complete the selection.

tig:
I'm eggcited to get this into main. What's the latest news @jmperricone ?

jmperricone:
> Loving this! But for me also the Range slider is confusing.
> 
> My first instinct was to move to the first radio button then hold shift to move to the end one (like selecting text in a text editor). When that didn't work I wanted to hit Enter on the first and then scroll to the end and hit Enter again to complete the selection.
> 
> Something like (when no selection):
> 
>     * First Enter creates a selection
> 
>     * Arrow cursors extend the selection (like if you mouse down in text and drag left/right)
> 
>     * Second Enter ends selection at the release point (or clears the selection if second Enter is in the same place as the first).
> 
> 
> When selection is already in place
> 
>     * First Enter Clears selection and starts a new selection process (as above).
> 
> 
> If focus leaves control during selection
> 
>     * Treat it like the second Enter
> 
> 
> Also I've noticed the cursor disappears when tabbing into the slider when the focus is on a 'checked' radio button (see below). Maybe it's just my terminal but it left me confused as to where the focus was and whether I was in the control:
> 
> tabbing _Tab navigation, first to a slider with no selection then to one with a fixed selection_

I did the selection mode as you proposed for the ranges.

> Also I've noticed the cursor disappears when tabbing into the slider when the focus is on a 'checked' radio button.

After pfff time, i realize that windows, for cursor blinking, invernts the actual colors of the char, but on curses, it inverts fg with bg colors, and I setted both the same color, because in windows it doesnt matter.

tznind:
Nice!  Thanks for taking the time to respond to the feedback.  These are really great.  Only issue I had testing this updated PR was when doing selection right to left on the 'Range' slider (i.e. first Enter was higher than the second) - when I did that the fill was not rendered:

image

It looks like its a drawing issue only as the ends clear correctly when making a new selection just as if user had done low to high normally.

jmperricone:
Hi, is there a way to know when the View is Added  ? Or maybe other method. What I need is to use the Colorscheme of the parent View, so the slider has "transparent" background and use the same fore color, unless we set the Attributes for the chars ( Set, Option, Space, Legend, Header). Each of these Runes, have Attributes to customize. But on initialization, Colorscheme is null.

What I am doing now is setting for example LegendAttribute = new Attribute(-1). This is invalid, but when I draw the Legend I check this, and if Attribute.Value == -1 , I use Colorscheme.Normal ( now it works because the View is already added, and use the parent Colorscheme).

Perhaps there is another approach that you are already using for this case.

tznind:
Thanks for reaching out for advice and taking the effort to make sure the feature is nice at the code level as well as the functionality.

> Perhaps there is another approach that you are already using for this case.

You should definitely use `ColorScheme` of your `View` not the parent.  In most cases it will be the same anyway.  

If you want to allow the user to change the color of stuff then I would use `Attribute?`.  Like so:

```csharp
///<summary>
/// Overrides the default color of the Option radio buttons
///</summary>
public Attribute? OptionColor {get;set;}
```

Then in your draw code you can use:
```
var drawColor = OptionColor ?? ColorScheme.Foreground;
```

BDisp:
@jmperricone there is the following available methods and events:

- (`View`) `Initialized` - This may not being sufficient because the parent view can be changed after the initialization.
- (`View`) `OnAdded` and `Added` - This may not being sufficient because the parent view can be changed after the initialization.
- (`Toplevel`) `OnLoaded` and `Loaded` - This may not being sufficient because the parent view can be changed after the initialization.
- (`View`) `OnDrawContent` and `DrawContent` - This may be the better solution but it needs to check if the `ColorScheme` was changed in the parent view and only update if it was changed.

Sometimes it may needs a combination of two methods/events. See the `StatusBar` view.

jmperricone:
Hi, @tznind  @BDisp @tig , thanks for the help. There are a couple of things that I have to improve and others to add, but I would like you to try it. I put the scenario at the beginning of the catalog.

jmperricone:
@tznind 
> ///<summary>
> /// Overrides the default color of the Option radio buttons
> ///</summary>
> public Attribute? OptionColor {get;set;}

> var drawColor = OptionColor ?? ColorScheme.Normal.Foreground;

> ///<summary>
> ///  Contains options for customizing the colors and symbols used in the slider
> ///</summary>
> public SliderStyle Style {get;set};
> 


I have it like that the style class. I replaced de Attribute to Attribute?.
```csharp
public class SliderChar {
    public Rune Rune { get; set; }
    public Attribute? Attribute { get; set; }
}
```
Then I have 
```csharp
public class SliderStyle {
    public SliderChar OptionChar {get;set;}
    public SliderChar SetChar {get;set;}
    ...
}
```
Then before drawing I declare:
```csharp
var normalScheme = style.SetChar.Attribute ?? ColorScheme?.Normal ?? Application.Current.ColorScheme.Normal;
```
This will ensure it finds a Colorscheme right ?

tznind:
Very nice 

> This will ensure it finds a Colorscheme right ?

In general I think most Views draw code assumes that `ColorScheme` is not null since `View` code returns SuperView ColorScheme if it doesn't have its own custom one:

https://github.com/migueldeicaza/gui.cs/blob/d84ab39ec3d38c48f18e4fa9f560fc997055c69e/Terminal.Gui/Core/View.cs#L1318-L1323

I've tested out the changes and they are great! But I did notice a couple of usability things.  It is possible to clear the selection in multi slider but not single slider.  This is because:

- In multi select mode a second Enter will unselect (so it is possible to unselect all options)
- In single select mode a second Enter will not unselect 

I think it would be nice to be able to clear the slider selection in single with a second Enter.  I wondered also about having the `Del` key clear selection too (for range sliders) but that might be alkward and its not like you can clear a radio button in Windows/Linux so may not be needed.

Do the red X symbols indicate invalid settings?

image

We probably shouldn't have 2 classes called `Orientation` in our library especially when they are identical. Current one is in the namespace `Terminal.Gui.Graphs` so I guess we could either:
- Move `Orientation` to `Terminal.Gui` namespace (mild breaking change).
- Reference it where it currently is by adding `using Terminal.Gui.Graphs` to the slider code (bit ugly)

Second option makes it annoying for new users since they have to add that extra line when changing orientation so I would say the first option is nicer but @tig or others could inform this?

BDisp:
Very cool @jmperricone, thanks.
I subscribe all the @tznind observations. All sliders layouts must allow to clear and be empty value. The auto innerspacing isn't formatting well after selecting another value and selecting auto again. I confirm the legend color issue in `Windows`.
Since `Orientation` library can be common to various scenarios, I think it should be move to the `Terminal.Gui.Core`.

jmperricone:
@tznind 

> I've tested out the changes and they are great! But I did notice a couple of usability things. It is possible to clear the selection in multi slider but not single slider.

Todo: For `Single`, `LeftRange` and `RightRange` sliders, it will work like the `Multiple`, by pressing over the option.
Todo: For `Range`, it has to be something else, like pressing `Esc` when selecting, or `Del` when focus.


> We probably shouldn't have 2 classes called Orientation in our library especially when they are identical. Current one is in the namespace Terminal.Gui.Graphs so I guess we could either:

Yes, I thought of that when I did it and added a note to it. Let me know.

```csharp
// Note(jmperricone): I named this Orientation instead of SliderOrientation, because I use it for the Slider and for Legends.
//Graph has also an Orientation class.
public enum Orientation {
	Horizontal,
	Vertical
}
```


> Do the red X symbols indicate invalid settings?

I always do this to see the entire area of `​​View`. So it means that the `Width` is +1.
```csharp
// Debug
Driver.SetAttribute (new Attribute (Color.White, Color.Red));
for (int y = 0; y < bounds.Height; y++) {
    for (int x = 0; x < bounds.Width; x++) {
        MoveAndAdd (x, y, 'X');
    }
}
```
Todo: When the `Slider` is `Vertical`, it is miscalculating the `Width`.


> Finally I noticed that when changing Slider focus the Legend changes color (which i cool) but the colors in the Scenario (Dark Grey on Black) were showing up as Black on Black on PowerShell for me (but it worked fine on my Linux laptop). This made me think the headers were vanishing!

I did not understand the problem well. I am using Linux at the moment, but will test it on Windows.
When I draw the options, if the option is focused, I change it to Coloscheme.Focus, to make sure it blinks.

```csharp
// Note(jmperricone): Maybe only for curses, windows inverts actual colors, while curses inverts bg with fg.
if (Application.Driver is CursesDriver) {
    if (currentOption == i && HasFocus) {
	Driver.SetAttribute (ColorScheme.Focus);
    }
}
```
Related to this right ?
> Also I've noticed the cursor disappears when tabbing into the slider when the focus is on a 'checked' radio button (see below). Maybe it's just my terminal but it left me confused as to where the focus was and whether I was in the control:


Todo: Make the blinking option (focused option) the same on Linux and Windows

jmperricone:
Hi @BDisp, thank you.

> Very cool @jmperricone, thanks.
I subscribe all the @tznind observations. All sliders layouts must allow to clear and be empty value. The auto innerspacing isn't formatting well after selecting another value and selecting auto again. I confirm the legend color issue in Windows.
Since Orientation library can be common to various scenarios, I think it should be move to the Terminal.Gui.Core.

I have to work on this. `Innerspacing` is the space between options. The way `Innerspacing` is calculated depends on this enum:
```csharp
public enum SliderLayout
{
    Auto, // Default. I do the math to find the `Innerspacing`, then I set the `Width` and `Height`,
    Layout, // Innerspacing based on `Width` when `Horizontal` and based on the `Height` when it is `Vertical`.
    Custom // If you change the `Innerspacing` property, layout is set to `Custom`, so it does not recalculate.
}
```
With `SliderLayout.Layout` set, if the slider does not fit, it will explode.
Do you think this option is necessary?

Todo: Make `SliderLayout.Layout` work.

`Layout` already exists. We need another name, or, If it's okay with you, I do it like this:
```csharp
public class Slider
{
    ...
    public SliderLayout SliderLayout {get; set;}
    ...
}
```

> Since Orientation library can be common to various scenarios, I think it should be move to the Terminal.Gui.Core.

okay for me.

BDisp:
> With SliderLayout.Layout set, if the slider does not fit, it will explode.
> Do you think this option is necessary?

No, I think is equipared to the `Custom`, so if it will only have two options, maybe is better using a bool AutoSpacing.


> Layout already exists. We need another name, or, If it's okay with you, I do it like this:

In the case you still maintaining the `SliderLayout` enum then `public SliderLayout SliderLayout {get; set;}` is valid.

jmperricone:
I think there are 3 cases.
For `Auto = true`, we use the limits. Dim.Fill() for example.

When `Auto = false`, this case could happen.

1) The view is created and `Innerspacing` is calculated.
2) We change `Innerspacing` to `3` for example.
3) We add a new Option to the Slider, that with `3`, does not work well.

In this case we can no longer go back to the automatic to adjust to the best.

Maybe we can add a method to do this. 
In this case we have 
```csharp
public int Innerspacing {get;set;}
public bool AutoSpacing {get;set;}
public void AdjustBest()
```

I think i prefer the enum over a method.

BDisp:
> I called Auto for finding the best value, and Layout, to use the View Dim.

I understand, thanks for explanation. In the case the `Auto` layout is selected again and `AutoSpacing = true`, then the `Innerspacing` must be calculated again based on the width and the `public void AdjustBest()` is a good choise.

BDisp:
But is better to add the underline bitwise to the contents attribute than increase the contents length to save the underline.

tig:
This is such a great PR. What's it going to take to get it finished so we can incorporate it??

@jmperricone ??

tznind:
Thanks for sticking with this, its going to be a great addition when its finished 🚀

tig:

tig:
I need this for the v2 Color Picker (TrueColor) control!

@jmperricone you still interested in finishing this?

tig:
Please continue this discussion on https://github.com/gui-cs/Terminal.Gui/pull/2901 where I've ported @jmperricone's awesome code to v2 and started trying to finish it.

Closing this PR since we will not be putting this in v1.

PR #1308, opened 2021-05-16
Fixes #1307 MainLoop timeouts duplicate keys error.

### LINKED ISSUE #1307: MainLoop timeouts giving duplicate key error on Mac.
This only happens on `Mac` and not on `Windows` and `Linux`. Happens because the time ticks on `Mac` system may work different from the others systems. By coincidence, can happen two threads are adding scheduled time at the same time, resulting in duplicate keys..

timeout-error
unit-test-error

BDisp:
Here is a prove that is triggered on a Mac system.

timeouts-duplicate-keys-mac

### DIFF
--- a/Terminal.Gui/Core/MainLoop.cs
+++ b/Terminal.Gui/Core/MainLoop.cs
@@ -171,8 +171,9 @@ namespace Terminal.Gui {
 			long now = DateTime.UtcNow.Ticks;
 			var copy = timeouts;
 			timeouts = new SortedList<long, Timeout> ();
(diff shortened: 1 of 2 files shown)


Terminal.Gui/Core/MainLoop.cs
```diff
@@ -31,7 +31,7 @@ public interface IMainLoopDriver {
 		bool EventsPending (bool wait);
 
 		/// <summary>
-		/// The interation function.
+		/// The iteration function.
```
FreddieMercurial, line 34: nice catch :)

PR #1315, opened 2021-05-20
Fixes #983. Improving clipboard with interaction with the OS.

### LINKED ISSUE #983: Resizing terminal doesn't seem to work on linux
Been a while since I've worked on my console version of my app, and noticed lots of stuff has changed since I was last here. One of the new issues I noticed (on the latest commit - I just pulled master before making this) is that on Linux the terminal size doesn't cause the application to refresh the interface, and it's stuck in the original size, which makes it nearly unusable.

image

Right screenshot is on Ubuntu 20.04.

I've also had some reports where users mouse over things and it causes the application to detect clicks, or inputs of some kind. One user had it simulate 12 clicks on a download button - but I don't think this is due to this gui framework, I've noticed linux just kind of thinks mouseover is some sort of input, e.g. i have terminals that are 'press any key to continue' (like when a console app ends in the debugger) and mousing over them immediately closes them. I don't use Linux that often so not sure if this is intended behavior of a terminal on Linux?

BDisp:
I also added more features to the `FakeDriver` and `FakeConsole`.


tig (approved):
Brilliant!

PR #1319, opened 2021-05-22
Fixes #1318. Ensures that the OS clipboard is always sets.

### LINKED ISSUE #1318: The setting for the OS clipboard on Linux is not fully guaranteed.
I detected this when I was trying to do unit tests to getting the clipboard content from the OS and also setting the OS clipboard.

BDisp:
A clipboard trust unit test would to be using native OS functions to check it's really working.
- For `Windows` there is the `CLIP` function, but unfortunately it only places a copy of the text on to the `Windows` clipboard and don't gets from. So, the better solution was using the `PowerShell` which already come with the OS, with the `Set-Clipboard` and `Get-Clipboard`
- For `Mac` the choice was the `pbcopy` and `pbpaste`,  which also comes with the OS.
- For `Linux` there is nothing that come with the OS. I even thought about installing the `PowerShell`, but it also uses the `xclip` for could do the job :-). So, I decided to use the same `xclip` without installing nothing more. It was then that I discovered that something was not functioning well, because I did not specify that the clipboard configuration of the OS would be the cliboard itself because it also exists `Primary` and `Secondary`.


BDisp:
To test in `Linux` it is needed to install `xclip`.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -1071,52 +1071,78 @@ public string GetClipboardData ()
...
+			// }
+
 			try {
-				// BashRunner.Run ($"cat {tempFileName} | xsel -i --clipboard");
-				BashRunner.Run ($"cat {tempFileName} | xclip -selection clipboard");
-			} finally {
-				System.IO.File.Delete (tempFileName);
+				BashRunner.Run ("xclip -selection clipboard -i", false, text);
+			} catch (Exception ex) {
+				throw new Exception (ex.Message);
```
tznind, line 1086: Should this be `NotSupportedException` so that there is a standard exception type across all the drivers? Either way you probably want to expose the ex as an InnerException e.g. `throw new NotSupportedException("Write to clipboard failed",ex);`

BDisp: Thanks. In your opinion what's the best solution?

tznind: Not sure how feasible this is but I would try to extend `IClipboard` to the following (see below).  If it was possible to implement this then testing should become much easier as you can just return true from test if `IsSupported` is false.

```csharp
/// <summary>
/// Definition to interact with the OS clipboard.
/// </summary>
public interface IClipboard {

    /// <summary>
    /// Returns true if the environmental dependencies are in place to interact with the OS clipboard
    /// </summary>
    /// <value></value>
    bool IsSupported {get;}

    /// <summary>
    /// Sets the operation system clipboard.
    /// </summary>
    /// <param name="text"></param>
    /// <exception cref="NotSupportedException">Thrown if it was not possible to set the clipboard contents</exception>
    void SetClipboardData (string text);

    /// <summary>
    /// Sets the operation system clipboard if possible.
    /// </summary>
    /// <param name="text"></param>
    /// <returns>True if the clipboard content was set successfully</returns>
    bool TrySetClipboardData(string text);

    /// <summary>
    /// Gets the current contents of the OS clipboard
    /// </summary>
    /// <returns></returns>
    /// <exception cref="NotSupportedException">Thrown if it was not possible to read the clipboard contents</exception>
    string GetClipboardData ();

    /// <summary>
    /// Returns true if it was possible to read the OS clipboard
    /// </summary>
    /// <param name="result">Clipboard contents read</param>
    /// <returns></returns>
    bool TryGetClipboardData(out string result);
}
```

I think regardless of implementation `GetClipboardData` and `SetClipboardData` should only throw `NotSupportedException`.  You could enforce this rule with a shared abstract base implementation e.g.

```csharp
public abstract class ClipboardBase : IClipboard {
    public abstract bool IsSupported {get;}

    public string GetClipboardData ()
    {
        try
        {
            return GetClipboardDataImpl();
        }catch(Exception ex)
        {
            throw new NotSupportedException("Failed to read clipboard",ex);
        }
    }

    protected abstract string GetClipboardDataImpl ();

    public void SetClipboardData (string text)
    {	
        try
        {
            SetClipboardDataImpl(text);
        }catch(Exception ex)
        {
            throw new NotSupportedException("Failed to write to clipboard",ex);
        }
    }

    protected abstract void SetClipboardDataImpl (string text);

    public bool TryGetClipboardData (out string result)
    {
        // Don't even try to read because environment is not set up
        if(!IsSupported){
            result = null;
            return false;
        }


        try
        {
            result =  GetClipboardDataImpl();
            return true;
        }catch(Exception)
        {
            result = null;
            return false;
        }
    }

    public bool TrySetClipboardData (string text)
    {
        // Don't even try to set because environment is not set up
        if(!IsSupported){
            return false;
        }

        try
        {
            TrySetClipboardData(text);
            return true;
        }catch(Exception ex)
        {
            return false;
        }
    }
}
```

BDisp: This is challenging but I really like implementation. You know, if I find it difficult to implement it I'll ask you to implement :-). But I will try to put it into practice and still try a new implementation for the WSL, which is not worth installing the XCLIP because it will never work. I'm having some ideas and with your help, I think I'll get it right. Thanks.

BDisp: > Let me know if you think this is over engineering!

Well, it is not very easy :-). What should I use in each driver, the `IClipboard` or the `ClipboardBase`? I think there is no sense to use both. Can you please provide some link to learn more about this implementation. Thanks.

tznind: > 
> 
> > Let me know if you think this is over engineering!
> 
> Well, it is not very easy :-). What should I use in each driver, the `IClipboard` or the `ClipboardBase`? I think there is no sense to use both. Can you please provide some link to learn more about this implementation. Thanks.

I would have `ConsoleDriver` still use `IClipboard` but make `WindowsClipboard`, `MacOSXClipboard` and `CursesClipboard` implement the abstract base.  Its usually better to reference the interface.  The abstract base is there to ensure consistency between implementations (e.g. the type of exception thrown and how to implement Try) as well as to cut out duplication between the clipboard implementations.

If you run into difficulties I can help out :)

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -1083,7 +1107,7 @@ public void SetClipboardData (string text)
 			try {
 				BashRunner.Run ("xclip -selection clipboard -i", false, text);
 			} catch (Exception ex) {
-				throw new Exception (ex.Message);
+				throw new NotSupportedException ("Write to clipboard failed", ex);
```
tznind, line 1115: I think with the `ClipboardBase` abstract you don't need to have this try/catch block as there is already one in the base classes `SetClipboardData`

BDisp:
I'm having some trouble with the permissions issue, because I don't know to to test it on my own, as I have all with permissions. Suggestions?

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
BDisp: Done. Thanks @tznind. If you see anything else not very well implemented, please let me know.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
BDisp: If something is wrong please let me know. Thanks @tznind.

tig:
Is this ready to merge?

BDisp:
There is only a issue on `WSL` after pasting. Does anyone know why after pasting on `WSL` the text becomes corrupt? The cursor on the `Editor` in `UICatalog` can navigate over the `Window` left border. This also happened on Linux before I change the `CursorDriver`'s code but in the `WSL` still persists this issue. But I think it's enabled to merge and then trying to understand this later.

wsl-pasting-issue

BDisp:
Well, this is only happen if I run `WSL` with the command `"dotnet ./UICatalog.dll"` or `"dotnet run"`. If I run from `"Visual Studio Code"` with the integrated terminal this doesn't happens. Damn.

BDisp:
The output corrupted on the `WSL` doesn't has nothing to do with clipboard, but with `Process` API, I think. See this video which `WSL` and also `Windows Terminal` corrupt the terminal when opening the browser. On `Windows`, `Linux` and `Mac`  is working fine.

wsl-process-output-issue

tig:
Is this PR ready to merge?

BDisp:
Yes

tig:
Ok, I could not tell from the comments above. Thanks.

BDisp:
The comments above are only related with WSL issues and nothing with the clipboard of this PR. But I should being more clear. Sorry. Thanks.

tig:
> The comments above are only related with WSL issues and nothing with the clipboard of this PR. But I should being more clear. Sorry. Thanks.

Yes, it was confusing. Thanks for clarifying.

Issue #985, opened 2020-10-28
How to close a toplevel dialog and then change the toplevel view in the same thread(?)


Mgamerz:
Been having this issue for a while but one of my linux users is unable to use my app cause when a modal dialog closes (essentially a 'checking' dialog), it never transitions to a new view. I've known about this issue for a while but it has not been a blocking issue, but it appears to now be. I can reproduce this issue.

![image](https://user-images.githubusercontent.com/2738836/97500868-fa67cd80-1935-11eb-8620-06d2906a9dbd.png)

In the above picture, I have the completion event of a long running task (this pic is after I tested a few methods - the MainIteration is not there originally). During this task, I have a MessageBox.Query() open with no buttons that says 'Performing precheck'. When this process ends, the dialog is to close, and then either an error shows up, or a new UI controller (essentially full screen window) is swapped to.

On Windows, I've never seen any issues with this. On Linux I'm having lots of problems where the dialog disappears but the UI never changes. Debugging shows my swap code is called but the actual Run() event on my new UIController (which extends Window) is never fired. After more debugging, I found the the calls to Application.RequestStop() just request the top level stops, but the toplevel doesn't change until... some time later, I guess.

How am I supposed to wait until all of these actions have been performed? In WinForms, if I recall, you could do Application.DoEvents() to let pump the UI queue. I see there is a `MainLoop.MainIteration()` in Application.cs that seems like it should do the stuff I'm looking for. There's a comment on it: `		///     while (main.EvensPending ()) MainIteration ();`

When I use this, the MainIteration(); call is never invoked. Which makes me think there are no events pending. When calling Application.RequestStop() all it does is make the current TopLevel.Running = false. This has no setter logic so there is no way (I think) that an event pending is set. So this while method call would never work, I think.

You can see how I used it in the above picture, I just removed the while loop to try and force it to do a loop so the next TopLevel would be done after the current one is popped off the stack. This worked... sometimes(?). 

I'm not really sure how I am supposed to wait until the latest TopLevel has been updated. This makes using dialogs difficult because I don't really know when I can safely call Application.RequestStop() and Application.Run(). I'm trying to make sure all of these calls are invoked through MainLoop, but I'm not sure of the thread safety of this library (in WPF trying to UI things on the wrong thread would throw an exception. It might be useful to have that here so I don't accidentally update the UI on the wrong thread?).

BDisp:
For several toplevels see the `Example` project in `demo.cs` the following:

[...]

PR #1331, opened 2021-06-03
Toplevel improvement as a subviews container without frame borders.

BDisp:
- Added a persisting multi Toplevel `BackgroundWorker` threading to demonstrate as requested in the issue on #985.
- For those that don't like to use the `Window` because has the frame border, the `Toplevel` was improved to provide identical funcionaties without frame border.
- All toplevels are draggable now.

With this PR in the same console it's allows 3 types  of Toplevels:
1.  Non persistent toplevel always recreating the `Application.Top`.
2.  A single persistent toplevel maintained his state. (See SingleBackgroundWorker.cs scenario)
3.  Multi persistent toplevels representing Mdi childes of the Mdi container, allowing multi instances of a class with the windows navigation feature between them. (See BackgroundWorkerCollection.cs scenario)
 
mdi-container


tig (changes requested):
Can you please break the parts of this PR that are distinct into separate PRs?

tig:
My deepest apologies for taking so long on this (and others). I've been pretty busy. But I will get through all the PRs asap!

PR #1336, opened 2021-06-13
Expose TextView color methods as protected virtual (allows custom colors)

tznind:
This PR changes the following private methods in TextView to protected and adds parameters for the line being rendered and the index of the character that the color is for.

* ColorNormal
* ColorSelection
* ColorUsed

I've added a scenario that overrides the methods to do SQL syntax highlighting:

syntax-hightlighting

I thought about adding an interface and exposing the functionality that way e.g. `ITextViewColorProvider` but that would lead to a larger footprint... Both options have advantages.  It would be pretty easy to go that route instead if you think that would be better approach.

I can add some unit tests to confirm that the protected methods are called with the right values too if your ok with this approach?


tig (approved):
I think the way you've done this is spot on. 

Love!

PR #1343, opened 2021-06-18
Fixes #1341. Now if AutoSize is true the Bounds size is always updated by  using the Dim.Fill or the Dim.Absolute.

### LINKED ISSUE #1341: Label text does not update when Width=Dim.Fill()
When creating a new `Label` with a `Width=Dim.Fill()`  changes to the text do not affect what gets rendered on the screen.

I think this reproduces the issue.  Interestingly setting `Width=100` does not have the same issue.

```csharp
[Theory]
[InlineData (true)]
[InlineData (false)]
public void LabelChangeText_RendersCorrectly(bool useFill)
{
	var driver = new FakeDriver ();
	Application.Init (driver, new FakeMainLoop (() => FakeConsole.ReadKey (true)));
	driver.Init (() => { });

	// create a wide window
	var mount = new View () {
		Width = 100,
		Height = 100
	};

	try {
		// Create a label with a short text 
		var lbl1 = new Label ("ff");

		// Specify that the label should be very wide
		if (useFill) {
			lbl1.Width = Dim.Fill ();
		} else {
			lbl1.Width = 100;
		}

		//put label into view
		mount.Add (lbl1);

		// render view
		lbl1.ColorScheme = new ColorScheme ();
		Assert.Equal (1, lbl1.Height);
		mount.Redraw (mount.Bounds);

		// should have the initial text
		GraphViewTests.AssertDriverContentsAre ("ff", null);

		// change the text and redraw
		lbl1.Text = "ff1234";
		mount.Redraw (mount.Bounds);

		// should have the new text rendered
		GraphViewTests.AssertDriverContentsAre ("ff1234", null);


	} finally {
		Application.Shutdown ();
	}
}
```

image

BDisp:


tig (approved):
Nice!

PR #1359, opened 2021-07-01
Fixes #1358. Attribute.Foreground / Attribute.Background now working with CursesDriver

### LINKED ISSUE #1358: Attribute.Foreground / Attribute.Background not working with CursesDriver
I have noticed that when creating an `Attribute` with the `CursesDriver` and then trying to invert it the colors are not working correctly.  I have tracked this down to the `Foreground` and `Background` properties being wrong on `Attributes` created by the `CursesDriver` (only).  

The other drivers do not seem to be affected:

```csharp
[Theory]
[InlineData("Terminal.Gui.CursesDriver")]
[InlineData("Terminal.Gui.NetDriver")]
[InlineData("Terminal.Gui.FakeDriver")]
public void TestInvertColors(string driverType)
{
    var type = typeof(View).Assembly.GetType(driverType);
    var driver = Activator.CreateInstance(type,true);

    var d = ((ConsoleDriver)driver);
    d.Init(()=>{});

    try
    {
                var color = d.MakeAttribute(Color.Red,Color.Blue);

                Assert.Equal(Color.Red,color.Foreground);
                Assert.Equal(Color.Blue,color.Background);
    }
    finally{
                d.End();
    }

}
```

**Note:** I think this test needs to be run on the command line and on linux to work so I don't suggest we add it to the codebase

```bash
[xUnit.net 00:00:00.79]     Terminal.Gui.ConsoleDrivers.ConsoleDriverTests.TestInvertColors(driverType: "Terminal.Gui.CursesDriver") [FAIL]
  Failed Terminal.Gui.ConsoleDrivers.ConsoleDriverTests.TestInvertColors(driverType: "Terminal.Gui.CursesDriver") [25 ms]
  Error Message:
   Assert.Equal() Failure
Expected: Red
Actual:   Black
  Stack Trace:
     at Terminal.Gui.ConsoleDrivers.ConsoleDriverTests.TestInvertColors(String driverType) in /home/thomas/gui.cs/UnitTests/ConsoleDriverTests.cs:line 434

Failed!  - Failed:     1, Passed:     2, Skipped:     0, Total:     3, Duration: 73 ms - /home/thomas/gui.cs/UnitTests/bin/Debug/net5.0/UnitTests.dll (net5.0)

```

BDisp:
To allowing some test on some specifics `ConsoleDriver` features, the initialization should be done in the Init method (`ConsoleDriver`) and in the `Setup` (`IMainLoopDriver`), instead in the constructors.

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -137,9 +137,10 @@ namespace Terminal.Gui {
 			return new Attribute (
 				//value: Curses.ColorPair (last_color_pair),
 				value: Curses.ColorPair (v),
-				foreground: (Color)foreground,
-				background: (Color)background);
-
+				//foreground: (Color)foreground,
+				foreground: MapCursesColor (foreground),
+				//background: (Color)background);
+				background: MapCursesColor (background));
 		}
 
 		int [,] colorPairs = new int [16, 16];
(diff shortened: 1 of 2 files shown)


BDisp:
`282 The active test run was aborted. Reason: Test host process crashed : Error opening terminal: unknown.`

I don't know what going here.

tznind:
> `282 The active test run was aborted. Reason: Test host process crashed : Error opening terminal: unknown.`
> 
> I don't know what going here.

Looks like there are a lot of failed calls to `xclip` right before the crash message (although that could be coincidence).  The issue manifested in 9abcdec right? What exactly does that commit do? Is it possible that that commit makes some tests use the 'real' driver when previously they used `FakeDriver`? 

The error message seems to crop up a lot in real life (i.e. outside of CI) and be associated with the variable $TERM [not being defined](https://askubuntu.com/questions/905027/crontab-error-opening-terminal-unknown) (also [here](https://askubuntu.com/questions/1091553/how-do-i-fix-error-opening-terminal-unknown-on-ubuntu-server))

BDisp:
In the last commit I didn't see any error caused by the `xclip`, only in the previous.
I run all the unit tests on `Windows`, `Linux` and `Mac` without any problems.

Normally I need to export `TERM` environment variable on `Docker` images. Maybe the server is not prepared to run a real terminal in `Linux`.


> Is that the first time any test has constructed that class?

Yes, is the first time and maybe the environment is not properly prepared to run a terminal in the server image. But I'll need some help for this, please. Thanks.

tznind:
Ok I have enabled GitHub Actions (disabled by default in forks) and have created a minimum repro of the issue:

https://github.com/tznind/gui.cs/pull/19

I'll do a bit of exploring but I've also noticed that `Console.WriteLine` throws `NotImplementedException` in the CI container so it might be that it's simply not designed to run a real console session as you say.

BDisp:
> I'll do a bit of exploring but I've also noticed that `Console.WriteLine` throws `NotImplementedException` in the CI container so it might be that it's simply not designed to run a real console session as you say.

With the `WindowsDriver` and `NetDriver`, using windows console, you can't create an handle to a real console session. This the reason why I removed this initialization from the drivers constructors to the Init method. So we must avoid using Console functions.
The `CursesDriver` is the only that give no error creating a handle to a window and we need to call The `Init` and the `End` methods, because ncurses need to load the libraries to deal with the colors functions.

tznind:
Yup, seems as soon as you call `Init` ncurses crashes hard.  It might be that we just cant test this functionality this way and scrap the test?

BDisp:
It looks like it will have to be. Do you think I should also discard all changes to the drivers?

tznind:
> 
> 
> It looks like it will have to be. Do you think I should also discard all changes to the drivers?

Its up to you, I think the important thing is to fix the Foreground/Background color issue itself.  

I think changing Constructor/Init logic is something else and if it was **only** done to delay the 'error creating a handle' exception then we probably don't need/want that behaviour change.

tznind:
Great stuff!  I had a little play around with your InvertColors scenario to make it cooler.  If you don't like it you can ignore it:

```csharp
List<Label> labels = new List<Label> ();
var foreColors = Enum.GetValues (typeof (Color)).Cast<Color> ().ToArray ();
for (int y = 0; y < foreColors.Length; y++) {

	var fore = foreColors [y];
	var back = foreColors [(y + 1) % foreColors.Length];
	var color = Application.Driver.MakeAttribute (fore, back);

	var label = new Label ($"{fore} on {back}") {
		ColorScheme = new ColorScheme (),
		Y = y
	};
	label.ColorScheme.Normal = color;
	Win.Add (label);
	labels.Add (label);
}

var button = new Button ("Invert color!") {
	X = Pos.Center (),
	Y = foreColors.Length+1,
};
button.Clicked += () => {

	foreach(var label in labels) {
		var color = label.ColorScheme.Normal;
		color = Application.Driver.MakeAttribute (color.Background, color.Foreground);

		label.ColorScheme.Normal = color;
		label.Text = $"{color.Foreground} on {color.Background}";
		label.SetNeedsDisplay ();

	}
};
```

image

tig (approved):
Nice!

PR #1360, opened 2021-07-06
Prevents application crash if OS clipboard is not supported.

BDisp:
Can be tested on a real `Linux`, not on `WSL`, if `xclip` isn't installed.


tig (approved):
LGTM

PR #1364, opened 2021-07-11
Fixes #1314. TextView now exposes file exceptions from callers.

### LINKED ISSUE #1314: TextView: Hides file exceptions from callers
`TextView`, or more correctly the helper class `TextModel` does the following:

```c#
public bool LoadFile (string file)
{
	if (file == null)
		throw new ArgumentNullException (nameof (file));
	try {
		FilePath = file;
		var stream = File.OpenRead (file);
	} catch {
		return false;
	}
	LoadStream (File.OpenRead (file));
	return true;
}
```

In other words, it prevents callers from knowing WHY a file failed to open because it eats all Exceptions.

This is not cool, IMO. 

I propose to fix this by modifying `TextModel` to not have the try/catch block. 

Related to #487 (in that TextView is actually not very well designed).

BDisp:
Fixes #1314

### DIFF
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -38,14 +38,8 @@ namespace Terminal.Gui {
 
 		public bool LoadFile (string file)
 		{
-			if (file == null)
-				throw new ArgumentNullException (nameof (file));
-			try {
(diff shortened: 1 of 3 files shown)


tig (approved):
Thank you!!!!

PR #1366, opened 2021-07-13
Fixes #1078. ColorScheme setter now calls SetNeedsDisplay.

### LINKED ISSUE #1078: View.ColorScheme setter should likely call setNeedsDisplay ()
So it triggers a re-display.

BDisp:


tznind:
This has the potential to create infinite loops if a user has changed ColorScheme during drawing? e.g. in an override of `Redraw`.

Not saying that it doesn't make sense but could be a runtime breaking change in that situation?

BDisp:
If you look at the #1078, you will see that I don't agree at all on this, but I think you might are right. But as long it was a @migueldeicaza request I ending to make it. I can change the PR to only change the property if the value is different from the old one.

tznind:
d24b7ef is a great solution, very nice.

tig (approved):
Thanks for the deep dive on this both of you!

PR #1374, opened 2021-07-17
Adds Lineview

### LINKED ISSUE #1367: I want to draw horizontal line (HLine)
I want to draw horizontal line (U+2500), just like the border of `Window`/`FrameView`, I use following code (shortened for brevity):

```C#
    private Label _topSeparatorLabel;

    public Window Win { get; private set; }

    public void Init( Toplevel top )
    {
      Win = new Window( "Application Name Setup" )
      {
        X = 0,
        Y = 0,
        Width = Dim.Fill(),
        Height = Dim.Fill() - 1
      };
    }

    public void Setup()
    {
      _topSeparatorLabel = new Label()
      {
        X = 0,
        Y = 5,
        Width = Dim.Width( Win ) - 2,
        Height = 1
      };
      _topSeparatorLabel.DrawContent += TopSeparatorLabel_DrawContent;

      Win.Add( _topSeparatorLabel );
    }

    private void TopSeparatorLabel_DrawContent( Rect obj )
    {
      _topSeparatorLabel.Text = new string( '\u2500', _topSeparatorLabel.Bounds.Width );
    }
```

The idea is I need some kind of `Line` control which I can put anywhere to separate content in `Window`.
Since there is none, I think to simply use `Label` and give its `Text` with repeated value of `\u2500`.

Screenshot 2021-07-15 200151

Is this a good approach, or there will be a problem here?
Thanks in advance.

tznind:
A simple line view as requested, Closes #1367 

image

I reused `Orientation` which is in the `Terminal.Gui.Graphs` namespace which could be mildly annoying but I think it is more readable than using a `bool`.

### DIFF
--- a/Terminal.Gui/Views/LineView.cs
+++ b/Terminal.Gui/Views/LineView.cs
@@ -1,0 +1,106 @@
+﻿using System;
+using Terminal.Gui.Graphs;
+
+namespace Terminal.Gui.Views {
+	
+	/// <summary>
+	/// A straight line control either horizontal or vertical
+	/// </summary>
+	public class LineView : View {
+
+		/// <summary>
+		/// The rune to display at the start of the line (left end of horizontal line or top end of vertical)
+		/// If not specified then <see cref="LineRune"/> is used
+		/// </summary>
+		public Rune? StartingAnchor { get; set; }
+
+		/// <summary>
+		/// The rune to display at the end of the line (right end of horizontal line or bottom end of vertical).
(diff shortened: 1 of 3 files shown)


BDisp:
I think you might know, but if you want to automatically close the issue after your PR being merged, you need to mention in the PR or in a commit the following:
`Fixes #issueID` or `Closes #issueID`

ichan-akira:
Looks perfect!

tig (approved):
A nice little addition!

PR #1379, opened 2021-07-18
Fixes #546. Enhancement ProgressBar.

### LINKED ISSUE #546: ProgressBar: Enhancement ideas
- Support `Height` > 1
- Support a `Text` property that would be centered. Clients could do  `pc.Text = "${pc.Fraction:F0}%"`. 
- Support a `SegmentCharacter` property that would allow caller to specify the character drawn for each segment (currently hard coded to `Driver.Stipple`).
- Add a marquee mode (like https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.progressbarstyle?view=netcore-3.1). `Text` would scroll in marquee fashion.
- It should render more detail depending on the height. 1 line, current design; 2 progress plus percentage; 3 framed; 4 frame with progress; 6 frame wothnprogress padded (via #52).

BDisp:
progressbar-enhancement

### DIFF
--- a/Terminal.Gui/Core/ConsoleDriver.cs
+++ b/Terminal.Gui/Core/ConsoleDriver.cs
@@ -1155,14 +1155,14 @@ namespace Terminal.Gui {
 		public Rune RightBracket = ']';
 
 		/// <summary>
-		/// On Segment indicator for meter views (e.g. <see cref="ProgressBar"/>.
+		/// Blocks Segment indicator for meter views (e.g. <see cref="ProgressBar"/>.
 		/// </summary>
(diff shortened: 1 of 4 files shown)


BDisp:
I already ran again on `Windows`, `Linux `and `Mac `without no errors. The error on server is the same that only happened on `Mac`, but that was already fixed. It's possible forcing the github run the test again?

BDisp:
Now all unit tests have passed. There should be a way to re-run the drive tests on the server.

tig (approved):
This is FANTASTIC!!!

PR #1406, opened 2021-07-31
Autocomplete for TextView

### LINKED ISSUE #1405: Expose public property in TextView for model.GetLine(int line).
Usage would be for allowing user to get how long a particular row/line is. 
Or is there another way?

tznind:
I have been tinkering with how to add autocomplete like they have in Notepad++ and command line apps like [mycli](https://www.mycli.net/).

autocomplete

This is a work in progress and I was wanting some feedback on the design.

So far I have made only 2 changes to TextView

- Make existing method `List<Rune> GetCurrentLine()` public
- Add public method `InsertText(string)`

The Autocomplete class knows how to render into a `TextView` and how to generate suggestions from the current cursor position.

Currently the glue between Autocomplete and TextView is only in the Scenario UICatalog `SyntaxHighlighting.cs` specifically in the class `SqlTextView` which inherits from `TextView`.

I would like Autocomplete to work with TextView events instead of requiring inheritance though.  Is there an 'after render' event that allows drawing over a View after it has rendered, I tried DrawContent event but that seems to happen before the TextView is drawn and hence the changes are immediately overwritten?

Fixes #1405


BDisp:
> I would like Autocomplete to work with TextView events instead of requiring inheritance though. Is there an 'after render' event that allows drawing over a View after it has rendered, I tried DrawContent event but that seems to happen before the TextView is drawn and hence the changes are immediately overwritten?

Indeed the `OnDrawContent` should be invoked after a view was been drawn, as is on the `Redraw` method, before `if (subviews != null)`. In the foreach loop I put it before the `view.Redraw` because I needed it to take information for the `ScrollViewBar`, before the view is redrawn. My suggestion is create a new `Redraw` event that will be invoked before the view is redraws, maybe called `OnBeforeDrawContent`. Must be put at the beginning of the `Redraw` method, before the `var clipRect = new Rect (Point.Empty, frame.Size);`. So in the foreach loop will be looks like the following:

```cs
if (view.Visible && view.Frame.Width > 0 && view.Frame.Height > 0) {
	view.OnBeforeDrawContent (view.Bounds);
	view.Redraw (view.Bounds);
	view.OnDrawContent (view.Bounds);
}

```

tznind:
Ok just so I understand are you saying:

- OnDrawContent currently triggers before view drawing
- This is to allow ScrollViewBar to work
- There is no event for after view drawning yet

If this is correct then maybe we could tackle this in a seperate PR to avoid complicating things?

BDisp:
> 
> Ok just so I understand are you saying:
> 
>     * OnDrawContent currently triggers before view drawing
Only here: https://github.com/migueldeicaza/gui.cs/blob/52a5fccdc412367405790dd811ebf5410c1c9a39/Terminal.Gui/Core/View.cs#L1370.
But here it is invoked after the current view has been drawn: https://github.com/migueldeicaza/gui.cs/blob/52a5fccdc412367405790dd811ebf5410c1c9a39/Terminal.Gui/Core/View.cs#L1358
Maybe this should move to the end, after the https://github.com/migueldeicaza/gui.cs/blob/52a5fccdc412367405790dd811ebf5410c1c9a39/Terminal.Gui/Core/View.cs#L1380
> 
>     * This is to allow ScrollViewBar to work
Yes, but if the ScrollViewBar could be a extension property that could be sets in any view, it would be fantastic.
> 
>     * There is no event for after view drawning yet
Yes there is, in the https://github.com/migueldeicaza/gui.cs/blob/52a5fccdc412367405790dd811ebf5410c1c9a39/Terminal.Gui/Core/View.cs#L1358
It's only invoked after this: https://github.com/migueldeicaza/gui.cs/blob/52a5fccdc412367405790dd811ebf5410c1c9a39/Terminal.Gui/Core/View.cs#L1354

> 
> If this is correct then maybe we could tackle this in a seperate PR to avoid complicating things?
>

If you don't need for this PR, it's ok. In the case you need and if it's in a separate PR, then you need to merge into this PR.

tznind:
Hmn I am finding Autocomplete quite closely tied to TextView.  I wanted to make it possible to deploy in other controls and so was trying to drive it by events but I don't think that is going to work out.  Both 'reading the current word' and 'accepting an insertion' are tied to TextView and that is 90% of the class.

So @BDisp  how would you feel about me just making `Autocomplete` a member of the main `TextView` class?  So user experience would be:

```csharp
var tv = new TextView()
// enables autocomplete
tv.Autocomplete.AllSuggestions = new List<string>("word1","word2","word3", ...);
```

If I go down that route I can revert making the delete / insert text methods public too which might make it easier to refactor things later (once a method is public you can't change it without causing breaking changes).

BDisp:
If the Autocomplete cannot be used by others views, I think you can make a TextView member. If you will make all that methods public, you may make them virtual too. So on derived class, users can change it behaviors. Feel free to do the best off.

tznind:
Ok I have added scrolling and integrated autocomplete with TextView main code.  I've enabled autocomplete as an option in the Editor scenario too.  I've given a `ColorScheme` of cyan,black,white since it needs to be different from the TextView to be distinguishable and the user can change the scheme if needed.

I've tested with scrolling and ensuring that it doesn't over-spill boundaries etc.  Can you experiment and let me know if I missed anything?

autocomplete2

BDisp:
It's looks fantastic. I'm finishing a new big PR that changes around 35 files. I will analyze better as soon as I finish. Maybe there will be a lot of conflicts, but together we can fix all.

tznind:
Just reviewing my own code.  Do you think I should change the default ColorScheme to `Colors.Menu` instead of the cyan.  It would mean users that have adjusted the root color schemes would get the expected behavior for the Autocomplete.  That would make it look like below.

image

BDisp:
I agree. Looks good to me.

BDisp:
@tznind can you take a look at the huge PR #1385, when you have the time, please?

BDisp:
I already reviewed your PR and I only have one question. `Autocomplete` always run the `Redraw` and `ProcessKey` methods but once it checks the `Suggestions.Count == 0` he go out fast. So my question is if this is more time consumer or checking if the `Autocomplete` is null on the `TextView` for the case the user don't want to use it. But I think is enought fast.

BDisp:
Could you the same to the `TextField` since it uses the `TextModel` too, please?

tznind:
> Could you the same to the `TextField` since it uses the `TextModel` too, please?

At the moment the Autocomplete is configured to not over-spill the control borders.  Since TextField is often 1 height something would need to change to support it?

If the autocomplete can spill over control borders then we need to worry about the draw order of other Views in the parent container as well as how to stale renders.  It could get complicated.  At the moment Autocomplete is not a `View` just an effect that is rendered afterwards on top of the existing `View`.

I could still add it if you like and it just wouldn't work unless user has given more vertical space for the `TextField`?

BDisp:
> I could still add it if you like and it just wouldn't work unless user has given more vertical space for the `TextField`?

I think we should not change the height of the `TextField`. So it's best to forget my suggestion. The solution for this would be a genre of context menu view for when possible. Thanks.

tig:
@tznind is this ready for review?

tznind:
> 
> @tznind is this ready for review?

Yes please!

tig (approved):
Sorry this took so long. This is wonderful.

PR #1419, opened 2021-08-15
Correctly mark the .NET Framework reference assemblies as private.

teo-tsirpanis:
And use a built-in MSBuild item to apply the `InternalsVisibleToAttribute` (available since at least the .NET 5 SDK which is already required by `global.json`).


BDisp:
@teo-tsirpanis I tried this configuration `<InternalsVisibleTo Include="UnitTests" />` before without success. I think the `PrivateAssets="all"` did the trick. It's much better now. Thanks.

PR #1420, opened 2021-08-16
Fixes CursesDriver mouse and reset issue.

### LINKED ISSUE #1428: Exiting the application from Linux causes the terminal input to disappear afterward
Related: #931 and #1114

## Steps to reproduce:
1. Install .NET 5.0 on Linux.
2. Clone the repo. I used commit 618714cc69ac6e0c1641076bd3c535c32cfe78fd, which is latest as of when this issue was written.
3. Build and open UICatalog. 

     - Alternatively, use [this minimal repro](https://gist.github.com/breadbyte/b2cf55fbe90b5a2dcfc2c89e174709f4) with a reference to the Terminal.Gui project within the repo.

4. Exit UICatalog.
     - To exit the minimal repro, select the textbox at the bottom of the screen and press F1.
5. Type reset. Nothing will show up.
6. Press Enter. the console should have reset and typed characters will now appear.

## Expected outcome (commit https://github.com/migueldeicaza/gui.cs/commit/666912c046f8af1212b48df83e699f7b8bb73e83 on netcoreapp3.1)
4MRAHV0azj

## Actual outcome (commit https://github.com/migueldeicaza/gui.cs/commit/d90072151982a2499d3cf2df4776bb89a2fbb6f0 onwards on net5.0)
bsLiAQAWy9

Tested on the following machines:
`Linux localhost 5.11.0-25-generic #27-Ubuntu SMP Fri Jul 9 23:06:29 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux` (Ubuntu 21.04)
`Linux localhost 4.19.0-17-amd64 #1 SMP Debian 4.19.194-3 (2021-07-18) x86_64 GNU/Linux` (Debian 10)

I took a few minutes to run `git bisect` and returned with commit https://github.com/migueldeicaza/gui.cs/commit/d90072151982a2499d3cf2df4776bb89a2fbb6f0
I don't think this commit is useful, however, since it just moves the build target from netcoreapp3.1 to .net 5.0.

Using the workaround listed in #931 to toggle `Application.UseSystemConsole` does indeed fix this issue, but at the cost of usability (the application becomes way too slow to be reasonably usable).

BDisp:
### This PR fixes the following issues:

1. Open the `Editor` scenario and do a double or triple click on some text. The text will not selecting because there is a bug on a double or triple click, that is causing an extra click event and, thus, deselecting the text.
2. Only with `WSL`, open the `Mouse` scenario and moving the mouse you will see the `ReportMousePosition` event. Now press `Esc` to exit and open the `Mouse` scenario again. You won't see the `ReportMousePosition` again. On a real `Linux` this doesn't happen.
3. As discussed here https://github.com/migueldeicaza/gui.cs/issues/945#issuecomment-708669985, running on `CursesDriver`, when exiting the application, the terminal isn't restored, forcing to typing "`reset`" to restore. This was one of the most hard issue I had to fix, although the solution be so easy.

Fixes #1428

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -298,6 +298,7 @@ namespace Terminal.Gui {
 					mouseFlag = MouseFlags.Button3DoubleClicked;
 					break;
 				}
+				cancelButtonClicked = true;
 
 			} else if (buttonPressedCount == 3
 			       && (cev.ButtonState == Curses.Event.Button1Pressed
@@ -328,7 +329,7 @@ namespace Terminal.Gui {
 
 			} else if (((cev.ButtonState == Curses.Event.Button1Pressed || cev.ButtonState == Curses.Event.Button2Pressed ||
 				cev.ButtonState == Curses.Event.Button3Pressed) && lastMouseButtonPressed == null) ||
-				isButtonPressed && cev.ButtonState == Curses.Event.ReportMousePosition) {
+				isButtonPressed && lastMouseButtonPressed != null && cev.ButtonState == Curses.Event.ReportMousePosition) {
 
 				mouseFlag = MapCursesButton (cev.ButtonState);
 				if (cev.ButtonState != Curses.Event.ReportMousePosition)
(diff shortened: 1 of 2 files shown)


tig (approved):
nice work!

PR #1439, opened 2021-08-25
Fixes #1438. Backspace not redrawing screen under some situations.

### LINKED ISSUE #1438: Backspace not redrawing screen under some situations
I have a strange bug with the Backspace key when deleting the first character in a line (i.e. merging with previous line).  The change happens but the screen does not update.  I can reproduce it using this code below on `main` branch.  I wonder if it is something to do with the fact that the Window w is not a child of TopLevel or something?

Repro:

- Type some text
- Use backspace (works)
- Go to begining of a line and press backspace (no change visible on screen)
- Press another key (e.g. space) suddenly the screen and model update and changes reflect the backspace taking effect

I don't see the issue in Editor Scenario etc so thing it must be specifically related to the use of a Window @BDisp?

image
_Pressing backspace here will not update the screen until next (non cursor) keypress_

```csharp
Application.Init ();

var w = new Window () {
	Width = Dim.Fill (),
	Height = Dim.Fill ()
};

w.Add (new TextView () {
	Width = Dim.Fill (),
	Height = Dim.Fill (),
});

Application.Run (w);

Application.Shutdown ();
```

BDisp:


BDisp:
Oh damn, I made this changes in the main branch. Is there any workaround without delete this PR, like change the branch name?

BDisp:
@tznind  correct the link to this location PR #1441, because I was mistaken in the branch. Sorry.

PR #1440, opened 2021-08-25
feature: Don't pull Terminal.Gui from NuGet, use ReactiveMarbles.ObservableEvents

worldbeater:
This replaces [Pharmacist](http://github.com/reactiveui/pharmacist) with [ReactiveMarbles.ObservableEvents](https://github.com/reactivemarbles/observableevents) which is a successor for Pharmacist.
Based on https://github.com/migueldeicaza/gui.cs/issues/1132#issuecomment-903699483


BDisp:
I think is better you change the project to `net5.0` version too.

PR #1447, opened 2021-08-27
Fixes #1446. Added more features to the Border and Toplevel focus.

### LINKED ISSUE #1446: Border improvement request.
> 
> 
> Massively impressive.
> 
> I do note that
> 
>     * hit testing on shadows isn't quite right (I would expect when clicking on a shadow for it to NOT activate the view with the shadow, but any window under it).
> 
>     * views are only brought to front if their top line is clicked
> 
> 
> I'm not sure either of these are actually part of his PR and they certainly are not severe enough to not approve an merge.
> 
> Well done @BDisp !

BDisp:
border-improvement


BDisp:
Before:
border-before

After the fix:
border-after-fix

tig (approved):
Wow, tons of work here! Good stuff.

Issue #1435, opened 2021-08-24
VS Code Terminal eats Ctrl-Q, making it impossible to use in Terminal.Gui apps


tig:

BDisp:
In the settings choose external terminal for the  Workspace.

BDisp:
Another better solution is change the shortcut to Ctrl+Alt+Q in the Preferences->Keyboard Shortcuts. Seems it isn't possible configure it in the devcontainer.json, unless a custom extension is created, see https://stackoverflow.com/a/64343493/4872289.

![vscode-keybindings](https://user-images.githubusercontent.com/13117724/131352170-a8c3a862-92f6-4dd8-80c6-5a8d46fc610a.png)

tznind:
It would be nice to replace the code `case Key.Q | Key.CtrlMask:` with a user configurable key e.g. `Application.QuitKey`

BDisp:
I agree, but what should be the default value for the `Application.QuitKey`?

tznind:
> 
> I agree, but what should be the default value for the `Application.QuitKey`?

Probably for backwards compatibility it should be the current value  (Ctrl+Q).

BDisp:
Do you think is better to call `AlternateQuitKey`, like `AlternateForwardKey` and `AlternateBackwardKey`?

tznind:
Hmn so there are 2 routes we could go with this:

- AlternateQuitKey` would make sense if you want Ctrl+Q to always work regardless and this is a secondary key the user can configure.
- `QuitKey` would make sense if we are just pulling that constant value up to an (api) user configurable level.
 
There could be situations where a user wants Ctrl+Q to do something custom and therefore want to disable the current behavior (option 2)

It would be possible to implement both ofcourse (Add both `QuitKey` - defaulting to Ctrl+Q And `AlternateQuitKey`).

BDisp:
Well, thinking better the `AlternateForwardKey` and `AlternateBackwardKey` properties were create to add one more keys options, but maintain the original keys. With the `QuitKey` we want to give to users the possibility to changing the default key and thus, makes no sense to have alternate key. So, I think is better using the `Application.QuitKey` as you had first suggested.

tznind:
Makes sense to me :)

BDisp:
Maybe this can be closed, because it isn't a bug but only a conflict keys and there was already implemented a configurable `QuitKey` property.

tig:
i agree. closing

PR #1450, opened 2021-08-30
Added Application.QuitKey property to allow change the quitting application key.

BDisp:
As discussed here #1435.


BDisp:
Well, the `SendKeys` method is causing the `KeyUp_Event` test fail because the `Application.Top.KeyUp` event maybe is caught the keys send by `SendKeys`. I replaced to `ProcessKey` but also fails.

PR #1472, opened 2021-09-28
ComboBox cursonDownKey nullref fix

En3Tho:
A small fix for https://github.com/migueldeicaza/gui.cs/issues/1471
Added null guard to fix null ref when pressing keyDown inside combobox
Improved an error message when view cannot be found

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1870,16 +1870,19 @@ namespace Terminal.Gui {
 				}
 			}
 
-			if (edges.Any () && edges.First ().From != Application.Top) {
-				if (!ReferenceEquals (edges.First ().From, edges.First ().To)) {
-					throw new InvalidOperationException ($"TopologicalSort (for Pos/Dim) cannot find {edges.First ().From}. Did you forget to add it to {this}?");
-				} else {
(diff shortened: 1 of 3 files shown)


En3Tho:
@tznind @BDisp Is this suitable fix for the problem or should it be done another way?

tznind:
Thanks! @tig handles PR review but this looks fine to me.  Are you able to add a unit test?

En3Tho:
Okay, So I've found a few bugs with a unit test. Please look at them. Hontesly can we just make search set not null all the time? Is one allocation worth it?

BDisp:
I think is better avoiding null list. It must be an empty list on these cases.

BDisp:
Another solution to make possible to insert ComboBox on a smaller superview is to add the ListView to the superview's Superview, instead added to the ComboBox.

tznind:
I agree with both of you, empty list is good.  Lets solve the simple null reference issue first then look at a different PR for bigger changes?

BDisp:
@En3Tho don't forget that the searchset is a new source based on filtering what is tipping. If TextField is empty then all the source elements are shown.

En3Tho:
Yeah, but if you set TextField initially, then it is not actually empty and this behaviour skips searchset initialization resulting in a null ref

tig (approved):
LGTM!

PR #1478, opened 2021-10-06
Fixes #1475. Selection ending with a white space error.

### LINKED ISSUE #1475: View crashes when trying to get seleciton when last char is " "
How to reproduce:
Set last char of TextField (or ComboBox or maybe other views) text to " " then double click at next position.

Reproduceable both on main and nuget version.
PrepareSelection method passes invalid parameters to String.SubString resulting an exception.

As I've understood, selection tries to ignore whitespace but there are some quircks, for example:

" abc " 
 ^ - if double click is called on first char then full string is selected

" abc " 
        ^ - if double click is called on last char "abc " is selected

" abc " 
         ^ - if double click is called on next after last char position then exception is thrown

It is easily reproduceable when the only char is " ". The code below then has start = 1 and length = 2 which results in exception. So I guess start = 1 is ok, but length of 2 is not. Considering text itself has 1 char only.
image

BDisp:
Added mouse triple click feature to the `TextView` to select full line.

text-selection

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1602,7 +1602,7 @@ namespace Terminal.Gui {
 		/// </summary>
 		public event Action<KeyEventEventArgs> KeyDown;
 
-		/// <param name="keyEvent">Contains the details about the key that produced the event.</param>
+		/// <inheritdoc/>
 		public override bool OnKeyDown (KeyEvent keyEvent)
 		{
 			if (!Enabled) {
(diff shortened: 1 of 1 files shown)


En3Tho:
This looks awesome. I've also tested the copy-paste in Powershell in it works perfectly. I believe pasting is totally consistent now.  Tripple click feature is super useful too :)

BDisp:
> 
> 
> @BDisp @tig What prevents this from being merged? Can I help somehow?

I'm only a contributor and not depends on me.

En3Tho:
I see. @migueldeicaza Can you please merge this? This adds useful features and fixes a bug in powershell pasting.

PR #1482, opened 2021-10-09
Nullable Reference Types (NRT) migration

En3Tho:
Please look at this. Migrating to latest C# is not a problem (it's always been this way - it's just that certain features require either special type or runtime support, lang-only features have always worked well).

So C# 9 just requires LangVersion attribute set to 9. Somefeatures require either manual code copy-pasting or a nuget.

For NRT I've utilized https://github.com/manuelroemer/Nullable package

I don't have all the context about gui.cs sadly, but this is not meant to be "the great refactoring", morel ike a poco.
There are many questions we need to decide when migrating to NRT - mainly what really can be null and should be validated and what is not.

NRT really brings out the fact that there are lots of null landmines, sadly.

I would really like to bring out "Empty" versions of type like Dim.Empty, View.Empty etc but the problem is that they are not readonly. So I guess we have to stick with nulls. But at least I believe we have to make some things null-clear.

@BDisp @tig @tznind

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -18,6 +18,8 @@ using System.Diagnostics;
 using System.Linq;
 using NStack;
 
+#nullable enable
+
 namespace Terminal.Gui {
 	/// <summary>
 	/// Determines the LayoutStyle for a view, if Absolute, during LayoutSubviews, the
@@ -119,8 +121,7 @@ namespace Terminal.Gui {
 		}
 
 		// container == SuperView
-		View container = null;
-		View focused = null;
+        View? focused;
 		Direction focusDirection;
 		bool autoSize;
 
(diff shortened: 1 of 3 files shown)


BDisp:
Good stuff, thanks. Please, do a full format with `Ctrl+K+D`, if you are using `VS2019`, to accomplish the `.editorconfig`.
The annoying is the `Roslynator` warnings. He should ignoring them if the `#nullable enable` directive is set.
It must be compatible with `.Net.Framework` to work with `Mono`, as @migueldeicaza need.
But I've to agree with @tznind. It could break with some users codes. Class type needs to be `nullable` and not `empty` to take some desired actions.
But, we need to considering that all `unit tests` have passed and the `UICatalog` run fine.

BDisp:
@En3Tho I made changes to fix the warnings. See here https://github.com/BDisp/gui.cs/tree/csharp-9-and-nrt.

En3Tho:
Well, for now I don't see a problem with breaking signatures as migrating to NRT will most probably cause warning in projects which have NRT enabled. But it's up to people if they choose to use it. Exceptions are different thing - throwing argument null exceptions in null cases will be a breaking change. But I believe it's better to throw ArgNull that NullRef

BDisp:
@En3Tho `Pos` and `Dim` for me are mutable, because they depend from theirs derived classes. It can be `null`, `PosFactor`, `Percent`, `PosAnchorEnd`, `PosAnchorEnd`, `PosCenter`, etc.
`Empty` isn't the better way to differentiate between a default instance class and a class not instantiated at all (`null`).
For example the `Rect` struct has a `Empty` field that default all values to zero and a `IsEmpty` method which returns true if `(X == 0) && (Y == 0) && (Width == 0) && (Height == 0)`. So if we create a `new Rect(0, 0, 0, 0)`, it was instantiated but is still `Empty`. If we want to check if the `Rect` object was provided on a parameter method or not, we can't and must accept the `Empty`. But this is a struct and acceptable. For a class we need frequently check if the object was created or is null, since `Empty` only will indicate it have the defaults values and not if it was been created or not.

BDisp:
> Do you guys think we need a separate nrt branch or it is better to just merge to main bit by bit?

It's better to wait for @tig saying something.

migueldeicaza:
Generally the issue is that we don't have a way of determining where gui.ve users are, on Framework or .net core.

I personally still use framework on Mac as it has a faster launcher and debugger.

I am willing to move to core, but not if we end up dropping users.  We would need a plan for that.

En3Tho:
I understand your concerns but it's important to know that some of newer C# features like NRT do not require modern runtime. Some of them so and some don't. NRT requires only fee special attributes to be present.

BDisp:
Although `NRT` (Nullable reference types) is very Interesting, the most needed for using `C# 8.0` or greater is the ability to have many others features, using statements like `protected`, `virtual`, `protected set`, `private set`, etc... on `Interface` to have much more control of the code. But if we only using `netstandard2.1` or `net6.0` the `.NET Framework` users will don't have access to `Terminal.Gui` API. I think today is possible to use `WinForms`, `ASP.NET`, `WPF`, etc... with `Net5.0` or greater. The only problem is migrate to the newer version without big problems. A survey to investigate this will be good, but I think the change will be inevitable some day.

tig:
I've tagged this PR for the v2.0 milestone.

tig:

tig:
Closing as I believe we've agreed to not migrate completely to NRT in v2, but use nullable sparingly.

PR #1508, opened 2021-11-10
Allows ListView trigger the Enter and Leave events.

### LINKED ISSUE #1507: Leave event doesn't execute on ListView
When I add a `View.Leave` event handler to a `ListView` control it never executes, but the same works for `TextField`. Do I need to enable this event somehow, or is this a bug?

I'm using version 1.3.1 from NuGet.

See the below example:

```csharp
using Terminal.Gui;

namespace GuiListViewBug
{
    public class Program
    {
        public static void Main()
        {
            Application.Init();
            
            var listView = new ListView(new [] { 1, 2, 3 }) { Width = Dim.Fill(), Height = 3 };
            var textField = new TextField { Y = 4, Width = Dim.Fill() };
            var label = new Label { Y = 6, Width = Dim.Fill(), Height = Dim.Fill() };

            listView.Leave += _ => label.Text += "Left List View\n";
            textField.Leave += _ => label.Text += "Left Text Field\n";
            
            Application.Top.Add(listView, textField, label);
            Application.Run(Application.Top);
        }
    }
}

```

## Expected behavior:
When I press <kbd>tab</kbd> 6 times in a row it should display:
```
Left List View
Left Text Field
Left List View
Left Text Field
Left List View
Left Text Field
```

## Actual behavior:
Only the `TextField`'s handler executes so it shows:
```
Left Text Field
Left Text Field
Left Text Field
```

BDisp:
Fixes #1507.

### DIFF
--- a/Terminal.Gui/Views/Button.cs
+++ b/Terminal.Gui/Views/Button.cs
@@ -256,7 +256,7 @@ namespace Terminal.Gui {
 		///<inheritdoc/>
 		public override void PositionCursor ()
 		{
-			if (HotKey == Key.Unknown) {
+			if (HotKey == Key.Unknown && text != "") {
 				for (int i = 0; i < base.Text.RuneCount; i++) {
 					if (base.Text [i] == text [0]) {
 						Move (i, 0);
--- a/Terminal.Gui/Views/ComboBox.cs
+++ b/Terminal.Gui/Views/ComboBox.cs
@@ -238,7 +238,7 @@ namespace Terminal.Gui {
 
 			search.CursorPosition = search.Text.RuneCount;
 
-			return true;
+			return base.OnEnter (view);
 		}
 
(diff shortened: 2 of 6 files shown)


BDisp:
With the `Enter` and `Leave` events unit test, I detected some bugs which I already fixed.
@tznind and @jmperricone if you don't agree with the changes to make successfully the unit test, please let me know to remove them. Thanks.

tznind:
> With the `Enter` and `Leave` events unit test, I detected some bugs which I already fixed. @tznind and @jmperricone if you don't agree with the changes to make successfully the unit test, please let me know to remove them. Thanks.

Looks good to me! More unit tests are always good :) and the boundary check (Math.Max) is nice addition for TabView.

tig (approved):
LGTM! nice.

PR #1542, opened 2021-12-13
Support multi-line items in ListView

migueldeicaza-msft:
This patch allows ListViews to have items that take more than one line.

I have not extensively tested this, but wanted to get the code out so some folks asking for how to do it could reference this.


migueldeicaza-msft:
Ok, that will teach me to use an old checkout.

I was building against an old `master` checkout, rather than `main`.

migueldeicaza-msft:
Ok nuke this

PR #1548, opened 2021-12-18
Adding a KeyBindings feature

### LINKED ISSUE #1544: Question: vim-like bindings
Hi, love this library for its simplicity, while it is still quite powerful.

However, one thing I am having trouble with is the seeming limitations in regards to keybindings.
I'd like to change the defaults (arrow keys, tab, etc.) and use alpha bindings without mods (e. g. use hjkl for navigation).
Would this require a fork and some rewrite, or is there somewhere I can hook in and change stuff like this?

I use tools like https://github.com/jesseduffield/lazygit and https://github.com/jesseduffield/lazydocker daily, and love how the navigation works.

BDisp:
Starting a key bindings implementation...

### DIFF
--- a/Terminal.Gui/Core/KeyBindings.cs
+++ b/Terminal.Gui/Core/KeyBindings.cs
@@ -10,17 +10,42 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Initializes a new key binding for a view. If <see cref="View"/> is used will applied to all views.
 		/// </summary>
-		/// <param name="view">The view.</param>
+		/// <param name="view">The view type.</param>
 		/// <param name="inKey">The input key pressed.</param>
 		/// <param name="outKey">The desired output key.</param>
 		/// <param name="description">The description.</param>
 		/// <param name="enabled"><c>true</c> if enabled, <c>false</c>otherwise.</param>
 		public KeyBinding (Type view, Key inKey, Key outKey, string description = "", bool enabled = true)
+			: this (inKey, outKey, description, enabled)
 		{
 			if (view == null) {
 				throw new ArgumentNullException ("View cannot be null.", nameof (view));
 			}
+			if (!typeof (View).IsAssignableFrom (view)) {
+				throw new ArgumentException ("Type is not assignable from View.", nameof (view));
+			}
 			View = view.Name;
+		}
+
+		/// <summary>
+		/// Initializes a new key binding for a view. If <see cref="View"/> is used will applied to all views.
+		/// </summary>
+		/// <param name="view">The view.</param>
+		/// <param name="inKey">The input key pressed.</param>
+		/// <param name="outKey">The desired output key.</param>
+		/// <param name="description">The description.</param>
+		/// <param name="enabled"><c>true</c> if enabled, <c>false</c>otherwise.</param>
+		public KeyBinding (string view, Key inKey, Key outKey, string description = "", bool enabled = true)
+			: this (inKey, outKey, description, enabled)
+		{
+			if (string.IsNullOrEmpty (view)) {
+				throw new ArgumentNullException ("View cannot be null or empty.", nameof (view));
+			}
+			View = view;
+		}
+
+		private KeyBinding (Key inKey, Key outKey, string description, bool enabled)
+		{
 			InKey = inKey;
 			OutKey = outKey;
 			Description = description;
@@ -30,7 +55,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Get the view name.
 		/// </summary>
-		public string View { get; }
+		public string View { get; internal set; }
 		/// <summary>
 		/// Get or sets the input key pressed.
 		/// </summary>
@@ -92,12 +117,25 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Initializes a instance with the necessary settings.
 		/// </summary>
-		/// <param name="view">The view.</param>
+		/// <param name="view">The view type.</param>
 		/// <param name="inKey">The input key pressed.</param>
 		/// <param name="outKey">The desired output key.</param>
 		/// <param name="description">The description.</param>
 		/// <param name="enabled"><c>true</c> if enabled, <c>false</c>otherwise.</param>
 		public KeyBindings (Type view, Key inKey, Key outKey, string description = "", bool enabled = true)
+		{
+			AddKey (view, inKey, outKey, description, enabled);
+		}
+
+		/// <summary>
+		/// Initializes a instance with the necessary settings.
+		/// </summary>
+		/// <param name="view">The view.</param>
+		/// <param name="inKey">The input key pressed.</param>
+		/// <param name="outKey">The desired output key.</param>
+		/// <param name="description">The description.</param>
+		/// <param name="enabled"><c>true</c> if enabled, <c>false</c>otherwise.</param>
+		public KeyBindings (string view, Key inKey, Key outKey, string description = "", bool enabled = true)
 		{
 			AddKey (view, inKey, outKey, description, enabled);
 		}
@@ -144,7 +182,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Add a new <see cref="KeyBinding"/> to the <see cref="Keys"/> collection.
 		/// </summary>
-		/// <param name="view">The view.</param>
+		/// <param name="view">The view type.</param>
 		/// <param name="inKey">The input key pressed.</param>
 		/// <param name="outKey">The desired output key.</param>
 		/// <param name="description">The description.</param>
@@ -155,9 +193,29 @@ namespace Terminal.Gui {
 			if (view == null) {
 				throw new ArgumentNullException ("View cannot be null.", nameof (view));
 			}
+			if (!typeof (View).IsAssignableFrom (view)) {
+				throw new ArgumentException ("Type is not assignable from View.", nameof (view));
+			}
+			return AddKey (view.Name, inKey, outKey, description, enabled);
+		}
+
+		/// <summary>
+		/// Add a new <see cref="KeyBinding"/> to the <see cref="Keys"/> collection.
+		/// </summary>
+		/// <param name="view">The view.</param>
+		/// <param name="inKey">The input key pressed.</param>
+		/// <param name="outKey">The desired output key.</param>
+		/// <param name="description">The description.</param>
+		/// <param name="enabled"><c>true</c> if enabled, <c>false</c>otherwise.</param>
+		/// <returns><c>true</c> if the key binding was successfully added, <c>false</c>otherwise.</returns>
+		public bool AddKey (string view, Key inKey, Key outKey, string description = "", bool enabled = true)
+		{
+			if (view == null) {
+				throw new ArgumentNullException ("View cannot be null.", nameof (view));
+			}
 			var kb = new KeyBinding (view, inKey, outKey, description, enabled);
-			var viewName = view.Name;
-			if (!Views.Keys.Contains (viewName)) {
+			var viewName = view;
+			if (!Views.ContainsKey (viewName)) {
 				Views.Add (viewName, true);
 			}
 			if (Keys.Contains (kb, new KeyBindingEqualityComparer ())) {
@@ -174,8 +232,7 @@ namespace Terminal.Gui {
 		/// <returns><c>true</c> if the key binding was successfully added, <c>false</c>otherwise.</returns>
 		public bool AddKey (KeyBinding keyBinding)
 		{
-			Type view = GetInstance (keyBinding.View).GetType ();
-			return AddKey (view, keyBinding.InKey, keyBinding.OutKey, keyBinding.Description, keyBinding.Enabled);
+			return AddKey (keyBinding.View, keyBinding.InKey, keyBinding.OutKey, keyBinding.Description, keyBinding.Enabled);
 		}
 
 		/// <summary>
@@ -187,21 +244,10 @@ namespace Terminal.Gui {
 		/// <returns><c>true</c> if the key binding was successfully removed, <c>false</c>otherwise.</returns>
 		public bool RemoveKey (Type view, Key inKey, Key outKey)
 		{
-			var kb = new KeyBinding (view, inKey, outKey);
-			var viewName = view.Name;
-			var idx = Keys.FindIndex (x => x.View == viewName && x.InKey == inKey && x.OutKey == outKey);
-			if (idx == -1) {
-				return false;
+			if (view == null) {
+				throw new ArgumentNullException ("View cannot be null.", nameof (view));
 			}
-			var count = Keys.Count;
-			Keys.RemoveAt (idx);
-			if (Keys.FirstOrDefault (x => x.View == viewName) == null) {
-				Views.Remove (viewName);
-			}
-			if (count == Keys.Count) {
-				return false;
-			}
-			return true;
+			return RemoveKey (view.Name, inKey, outKey);
 		}
 
 		/// <summary>
@@ -216,8 +262,20 @@ namespace Terminal.Gui {
 			if (view == null) {
 				throw new ArgumentNullException ("View cannot be null.", nameof (view));
 			}
-			Type v = GetInstance (view).GetType ();
-			return RemoveKey (v, inKey, outKey);
+			var kb = new KeyBinding (view, inKey, outKey);
+			var idx = Keys.FindIndex (x => x.View == view && x.InKey == inKey && x.OutKey == outKey);
+			if (idx == -1) {
+				return false;
+			}
+			var count = Keys.Count;
+			Keys.RemoveAt (idx);
+			if (Keys.FirstOrDefault (x => x.View == view) == null) {
+				Views.Remove (view);
+			}
+			if (count == Keys.Count) {
+				return false;
+			}
+			return true;
 		}
 
 		/// <summary>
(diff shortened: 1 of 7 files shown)


Terminal.Gui/Core/KeyBindings.cs
```diff
@@ -0,0 +1,370 @@
...
+			kb.OutKey = to.OutKey;
+			kb.Description = to.Description;
+			kb.Enabled = to.Enabled;
+			if (new KeyBindingEqualityComparer().ReferenceEquals (kb, to)) {
+				return true;
+			}
+			return false;
+		}
+
+		private View GetInstance (string strFullyQualifiedName)
```
tznind, line 360: I don't think this method (`View GetInstance(string)`) will work for user defined Views since they will not be in `(View).Assembly`.  

This method seems to only be used to get the `Type` anyway (e.g. `GetInstance (from.View).GetType ().Name`).  Is there a reason not to just do everything at the `string` level instead of with `Type`.  The overloads that take a `Type` could stay but just call `.Name` on those asap and do everything with the class name?

Terminal.Gui/Core/KeyBindings.cs
```diff
@@ -0,0 +1,370 @@
...
+
+		/// <summary>
+		/// Get or sets the enable key to enter insert mode leaving the keys bindings.
+		/// </summary>
+		public Key EnableKey { get; set; } = Key.Enter;
+
+		/// <summary>
+		/// Gets or sets the disable key to leave insert mode returning to the keys bindings.
+		/// </summary>
+		public Key DisableKey { get; set; } = Key.Esc;
```
tznind, line 137: I don't think we should set this out of the box.  Disabling keybindings on Enter/Esc would typically be unexpected for most programmers encountering this feature (the keybinding feature).  I would keep `bool Enabled` but leave it up to the API user to decide whether to use the feature and how to enable/disable it (what shortcut or menu item etc)

Terminal.Gui/Core/KeyBindings.cs
```diff
@@ -0,0 +1,370 @@
...
+		/// <param name="keyBinding">The key binding.</param>
+		public KeyBindings (KeyBinding keyBinding)
+		{
+			AddKey (keyBinding);
+		}
+
+		/// <summary>
+		/// The views name with the enabled status.
+		/// </summary>
+		public List<(string View, bool Enabled)> Views { get; private set; } = new List<(string View, bool Enabled)> ();
```
tznind, line 117: It might simplify the code if this was a `Dictionary<string,bool>`?

tznind:
I've been looking at how keybindings work in other apps just for comparison.  Typically it seems that the keys are bound to `functions`.  So keybindings are not a translation layer `Key.J => Key.CursorLeft` but rather a mapping `Key.J => MoveCursorLeft` which comes with a default.  I know it would be more work to make such a system but worth discussing I think.

image
_In visual studio you bind the keys to specific functions_

image
_In Scintilla (library behind notepad++) there is an enum with all the commands the app supports_

Terminal.Gui/Core/KeyBindings.cs
BDisp: As you can see I store the views as string. I would like to pass as string but I have no warranty that the view is based in the `View` class. Imagine that a user derives from a `Dialog` view and how I have to check if the derived class has the base class `View`? Look that the `Dialog` is derived from `Window`, which is derived from `Toplevel`, which is derived from `View`. I don't want allow store a key binding to a view that is not a valid `View` class. What is the better approach?

Terminal.Gui/Core/KeyBindings.cs
BDisp: I agree, thanks.

Terminal.Gui/Core/KeyBindings.cs
BDisp: Do you mean set the `bool Enabled = true` and get rid of the `Key EnableKey\Key DisableKey`?

tznind: Yes :) I think that makes sense do you?

Terminal.Gui/Core/KeyBindings.cs
tznind: > I don't want allow store a key binding to a view that is not a valid `View` class. [...]

I think it is fair to trust that the API user can provide a `Type` or `nameof(MyViewType)` without checking ourselves that that is definetly derrived from `View`.  The worst case scenario is that they end up with a keybinding that does nothing right? And they should figure out what they have done wrong quite quickly when testing their code (assuming the method xmldoc documents this expectation).

BDisp:
Yes, worth discussing. But I think this won't work for every scenarios because many of the functions are inside the switch statement without calling any function. The use of reflection to get all the functions is very heavy but will not reach all of them. In that applications you have mentioned, there a database with all the enumeration methods and linked to a configurable key bindings. With my PR was not to introduce a key bindings for the `Terminal.Gui` views but provide a configurable keys bindings that user can do what he wants.

Terminal.Gui/Core/KeyBindings.cs
BDisp: More and less :-) The `bool Enabled = true` has no problem if there are no any key bindings setting but will force the `ProcessKey` in the `View` class to do a search in the key bindings unnecessarily, so the reason to setting as false.
For the `Key EnableKey\Key DisableKey` imagine if the user keyboard has malfunction on the `Enter` or `Esc` key and he wants to change these keys temporarily until to repair.

tznind: Ah sorry I didn't mean that it should always be `Enter`/`Esc`

What I meant was to remove the functionality where a key (e.g. Enter) toggles the keybindings `Enabled` and instead leave it up to the API user to set that up themselves (e.g. with a status bar item or menu shortcut).

Since alternate keybindings have to be setup by the programmer in their program.  Disabling those keybindings should be quite a rare thing to do and shouldn't be tied to a commonly used button like `Enter`/`Esc`.

Even in the case of the vim keybindings you only want to disable the keybindings (`hjkl`) under certain scenarios not every time Enter is pressed and it might vary from control to control.

BDisp: But I'm using this functionality in the `ProcessKey` method on the `View` class. Sorry I forgot to remove the tied `Enter` and `Esc` from the `View` class. See the `UICatalog` sample where I binding the same keys for the `View` (which is a global binding) and for the `TextView`. In the `ListView` you can navigate with the `hjkl`, pressing `Enter` will leave the key binding. Try open the `Editor` and you can still navigate with `hjkl`. If the user want to use other keys to enter and leave key binding, he only need to change the `EnableKey` and `DisableKey` keys. In the `ProcessKey` method I changed the `Focused` to `MostFocused` property because is the view that has priority to run.

BDisp:
@tznind you can do all the things like the images above with the `KeyBindings` class. Within a `ListView` on the left side you can show all the views names from the `Views` property and within other `ListView` on the right side you can show all the `KeyBinding.Description` from the `Keys` property, related with the selected view.
I think this is the way how they works. All the functions have shortcut keys and initially the `InKey` and `OutKey` are the same. With the above editors allow the user to change the `InKey` of the related function description but is always tied to the `OutKey`. What could be done is adding a `Action` property to the `KeyBinding` class, which can be used if no `OutKey` is provided, but how and where you can run it? It only can run a function within the current instance or a public function but is tied with a shortcut key, which is managed in the `ProcessHotKey` in the `View` class. It can only be executed when the view that desires this action is achieved, but it only be achieved through a `KeyEvent`. What we can do in this case is also adding a new `Action` type field to the `KeyEvent` class and if it isn't null then execute the action instead the `Key`. But it's worth? Is not enough using only shortcuts for this? Therefore, all functions must be available for both keys and mouse. One useful functionality that can be done is a class that can read all the keys that are used by each view and automatically create a `KeyBindings` object. This can be done?
Your opinion and suggestions are very important to me. Thanks.

tznind:
This is where I was thinking about going with KeyBindings.  It would be quite a lot of work but I think we would end up with an easier API to use, maintain and test.

The first step would be to identify a few very common operations that most views already support (e.g. scrolling and moving caret) and implement it for those.


```csharp
/// <summary>
/// All keybindable operations supported by Terminal.Gui views.  Concepts should be kept
/// broad enough to apply to all views e.g. LineScrollDown.  Not all <see cref="View"/> implement
/// all <see cref="Command"/>
/// </summary>
public enum Command {

	/// <summary>
	/// When bound to keys performs the standard platform behavior.
	/// </summary>
	Default = 0,

	/// <summary>
	/// Moves the caret down one line.
	/// </summary>
	LineDown,

	/// <summary>
	/// Scrolls down one line.
	/// </summary>
	LineScrollDown,

	/// <summary>
	/// Moves the caret down one page.
	/// </summary>
	PageDown,

	/// <summary>
	/// Extends the selection down one page.
	/// </summary>
	PageDownExtend,
}
```

Setting up the keybindings in the view constructor

```csharp

/// <summary>
/// The keys that can be pressed in this control to trigger the given <see cref="Action"/>
/// </summary>
public Dictionary<Key, Command> KeyBindings = new Dictionary<Key, Command> ();


/// <summary>
/// Translation of <see cref="Command"/> to which method must be called to achieve it
/// </summary>
protected Dictionary<Command, Action> CommandBindings = new Dictionary<Command,Action> ();

void Initialize ()
{
	CanFocus = true;
	Used = true;

	SetupKeyBinding (Key.PageDown | Key.AltMask, Command.LineScrollDown, () => ScrollDown (1));
	SetupKeyBinding (Key.PageDown | Key.ShiftMask, Command.PageDownExtend, () => PageDown (true));
	SetupKeyBinding (Key.PageDown, Command.PageDown, () => PageDown (false));
	SetupKeyBinding (Key.CursorDown, Command.LineDown, ()=> MoveDown());
}

private void SetupKeyBinding (Key key, Command command, Action action)
{
	KeyBindings.Add (key, command);
	CommandBindings.Add (command,action );
}
```

Then we can call the relevant methods as normal in ProcessKey:

```csharp
///<inheritdoc/>
public override bool ProcessKey (KeyEvent kb)
{
	if (!CanFocus) {
		return true;
	}

	// if the key pressed maps to a command
	if(KeyBindings.ContainsKey(kb.Key)) {
		
		// carry out the command
		CommandBindings[KeyBindings[kb.Key]]();
		return true;
	}
```

I think this approach would also naturally streamline the API e.g. all views (that support it) would end up with a (probably public) method `MovePageDown` instead of the code being sometimes embedded as multiple lines in the case statements.

I can see how the current KeyIn->KeyOut system in this PR is easier to implement given the current flow of code in `ProcessKey` but I think it is worth discussing doing something more structured like this?

tig:
Just to encourage the thinking and debating you two are doing, which is EXCELLENT...

Some assertions:

- We have one customer how has requested some very specific functionality (@MikaelElkiaer in #1544). His request is well-articulated and easy to understand. Most importantly, it is scoped very narrowly. 
- The existing keyboard processing in `Terminal.Gui` is complex and convoluted. Without meaning to offend anyone (esp. since I 'helped'), The convolution and complexity derives from the desire for this library to work cross-platform as well as Miguel's ("Emacs and the Midnight Commander") and other's (mine; "Windows, Turbo Pascal") biases.  I don't think it's something we should be proud of, and I don't think anyone would assert it is well thought through and "engineered". 
- We have few bugs or feature requests regarding how key bindings work. In other-words: What's there now works fine, even though it's complex and convoluted. I postulated above that being able to give `Editor.cs` different key bindings would be cool, but all I'm really doing there is IMAGINING what a real user might want. 
- The existing keyboard processing code is littered across `Application.cs`, `View.cs`, `Event.cs`, `ConsoleDriver.cs`, and all the drivers. etc... While it's not definitive, but illustrative, Fine Code Coverage identifies keyboard handling code in 4 of the top 10 Risk Hotspots where more unit tests are needed:

image

Is the approach @BDisp is taking in this PR is too broad (600+ lines of new code)? Is it just adding more complexity on top of existing complexity, just to satisfy @mikalElkiaer's scoped request (which appears to ALMOST be addressed via @tznind's sample code)?

If the answer to those questions is YES, then perhaps we should do a PR that surgically addresses #1544 now and then use what has been learned building this PR to start to really engineer a much more elegant key binding system for a future release.

BDisp:
My opinion is start this as new base to make a really key bindings as @tznind  addressed, if he agree of course.

tznind:
> My opinion is start this as new base to make a really key bindings as @tznind addressed, if he agree of course.

I agree yes 👍 I can start writting a keybinding feature based on the above over the christmas period.  I'll try to get the basics down and then we can work together to apply it consistently to the views?

BDisp:
I agree. Thanks. In your suggestion we have to use the code in each view that override the Processkey instead only on View?

tznind:
> In your suggestion we have to use the code in each view that override the Processkey instead only on View?

We can put the code that checks the keybindings in `View` then just call base as the first step in the override e.g.

```csharp
if(base.Processkey(e))
    return;
```

BDisp:
> Unless you want to have a bash ofcourse thats fine too.

No, we will going to have a bash together :-)
I may made some changes to prepare this to the commands implementation and integration with views. Be comfortable to send PR's to my branch if you want and I'll merge as needed.

BDisp:
First proposal is to create a `KeyBindings` folder to accommodate all the classes. What do you think?

tznind:
That makes sense, are you starting a new fresh branch?

BDisp:
No, I'm using the same. I removed the link to the issue #1544 as fix. Does I do it wrong?

tznind:
Hey I have been thinking about this a lot and the best approach to take writting this.    I think we should start with a fresh branch off `main` and refactor the `ProcessKey` methods of all the views so that all they do (as much as is possible) is call sub methods (as much as is possible).   

This would set us up well for mapping keybindings later on.  I don't think we want to move all the current branching into bindable actions until we have a chance to simplify it and make it more modular.

For example at the moment if you want to scroll down 1 in `TextView` you need:

```csharp
var desiredTop = textView.TopRow + 1;

// move the cursor to ensure it is on the screen after we scroll (prevents textview refusing to scroll down)
textView.CursorPosition = new Point(textView.CursorPosition.X, Math.Max(textView.CursorPosition.Y, desiredTop));

textView.ScrollTo(desiredTop);
```

I'd like that just to be:
```csharp
textView.ScrollDown (1);
```

This is already how [ListView's ProcessKey](https://github.com/migueldeicaza/gui.cs/blob/d84ab39ec3d38c48f18e4fa9f560fc997055c69e/Terminal.Gui/Views/ListView.cs#L370-L411) is set up (although I get that it is a simpler control).

I looked at `TreeView` to see how it would go https://github.com/tznind/gui.cs/commit/53702fc6c262c0960d070b7aa9a362a96c89ca6d

The advantages of these (even before considering keybindings) are:

- Tests can call the appropriate methods (e.g. `PageDown`) instead of relying on SendKeys
- API users can tailor behaviour more easily when subclassing views (less hidden private logic)
- API users can more easily use the existing Key events system to order more modular activities.

Do you think this is possible for `TextView`? I know there is a lot of state based logic in there?

BDisp:
> Setting up the keybindings in the view constructor

The disadvantage of this is to have to implement the keybindings in each view. It is a huge effort and a huge dependency that should only be managed by the base class `View`. The biggest problem in the base class `View` is to declare an `Action` in the application that only exists in each view, giving error in the module where these actions is being assigned. I'm just venting :-)

tznind:
At the moment in this PR the Keybindings are still being held by `Application` and handled using the view Type name. 

I was thinking that treating keybindings as a regular instance property on View.  So each view has a member of the Keybindings class as an instance property and that instance only contains mappings for the current view (not all views and certainly not static).

I will update my keybinding branch (which is currently branched directly from `main`) to show what I mean and we can talk more about the advantages/drawbacks.

One rule I think for preventing `action` approach getting out of hand is to commit ourselves to only having single method invokes with constant parameters (see below).  This will ensure we don't end up having to try and debug these lamdas.

Good:
```csharp
SetupKeyBinding (Key.PageDown, Command.PageDown, () => PageDown (false));
```

Not Good:
```csharp
SetupKeyBinding (Key.PageDown, Command.PageDown, () => Scroll (Bounds));
```

BDisp:
> At the moment in this PR the Keybindings are still being held by `Application` and handled using the view Type name.

I agree that it should be an instance initialized in each view so that actions can work. Really the name of the view type will not be necessary because we will have an instance of each view. But anyway you recognize that `KeyBindings` will always be related to views, right?

We will have to organize and select what views that each of us will update, so we will not be changing the same views doubling the work of each one. Thanks for the good work.

BDisp:
Okay, I already managed to do the actions working, using instances.

You realize that when using the actions we will be avoiding the `KeyPress` event to being invoked?

tznind:
> You realize that when using the actions we will be avoiding the `KeyPress` event to being invoked?

Why is that? Is that the current behavior of code in the `ProcessKey` method? we want to make sure this feature is backwards compatible and doesn't compromise the existing events.

BDisp:
Ups sorry, it's because I'm handling in the `View` class and it must be handled on the `ProcessKey` of each view.

Please see the changes made on the PR and say what you think about. Thanks.

BDisp:
@tznind my doubt is if it's better to remove the `OutKey` property from the `KeyBinding` class.
The unit test error is due the `TextView` not be in edit mode. Should we give up of this feature and leave it as user development as you made in your example (https://github.com/tznind/TerminalGuiVimlikeBindingsExample), using KeyPress?

tznind:
> I do not refer to KeyBindings but rather to the functionality of entering and leaving edit mode.

Yes I agree.  I think this is a very specific thing so having it user implemented rather than core library function is a good idea.

BDisp:
Good. If you want I can abandon this PR and you can continue with your. I will going to start to simplify the `ProcessKey` of the views with methods.

tznind:
> my doubt is if it's better to remove the OutKey property from the KeyBinding class.

Sorry do you mean you are doubtful about removing it (i.e. you don't want to loose it)? or that you think removing it is a good idea?

tznind:
> Good. If you want I can abandon this PR and you can continue with your. I will going to start to simplify the `ProcessKey` of the views with methods.

Thanks, I think thats a good idea.  Having elegant simple `ProcessKey` methods will be key to getting this to work.  

I will do a bit of refactoring on the keybinding bit of my PR then work on a Scenario and tests and look at TableView, TreeView and GraphView.

BDisp:
> Sorry do you mean you are doubtful about removing it (i.e. you don't want to loose it)? or that you think removing it is a good idea?

I meaning since it's to use the `Action` type there is no need to use the `OutKey` property. But sincerely, I think your implementation is good and you may submit a PR of yours and I close this one. In meanwhile, I'll changing the `ProcessKey` methods of the views accordingly.


> Thanks, I think thats a good idea. Having elegant simple `ProcessKey` methods will be key to getting this to work. We can take inspiration from this PR and cherry pick any bits that we want to keep.

Copy that :-)


> I will do a bit of refactoring on the keybinding bit of my PR then work on a Scenario and tests and look at TableView, TreeView and GraphView.

Good work then. If I need some help on the others views I'll tell something. Thanks.

PR #1556, opened 2021-12-21
Adds Key Binding support. Also refactors Autocomplete and Undo/Redo.

tznind:
This is a collaboration between myself and @BDisp to enable user configurable keybindings.  The use cases are:

- Programmer wants to remove existing keybindings and replace with their own e.g. vim keys
- Programmer wants to programatically list existing keybindings and corresponding commands e.g. on a UI for application user to configure

This is a big PR with a lot of lines of code changed.  But I think that not only does it implement the keybinding feature but it also improves the code readability for many controls like `TextView`.  The act of deconstructing all the `switch` statements and replacing them with simple methods like `PageUp`, `PageDown` etc has made it much easier to read and maintain.  Having simpler methods will also make many of the controls easier to inherit from and more versatile in general.

The PR also includes a refactor (non breaking) of Autocomplete to decouple it from `TextView` and allow it to be used in other controls (with a bit of work).  This was necessary to support keybinding in autocomplete with the same design as the other views.

rebinding
_Changing up/down item selection to use W and S_

This PR replaces https://github.com/migueldeicaza/gui.cs/pull/1548


BDisp:
@tznind I thought you would use the class `KeyBindings` as in the example that initially you presented (https://github.com/tznind/gui.cs/commit/391d06c86d557caea8051090866dd8f26649968e). This version imply to always create a `Command` for all actions when a `Key` and an `Action` with the `Command.Default` will be enough. In my opinion, commands will be needed to be created if they are used in many views.
`public List<KeyBinding> KeyBindings { get; set; } = new List<KeyBinding> ();` will be more easier and faster to create keys bindings.

tznind:
Yup, I changed from the `KeyBindings` class to using private Dictionary for a few reasons.  But first lets review before and after the change:

Before Change:
```csharp
KeyBindings.Add (new KeyBinding (Key.CursorUp,Command.LineUp,()=>MoveUp()));
KeyBindings.Add (new KeyBinding (Key.P | Key.CtrlMask, Command.LineUp, () => MoveUp ()));
KeyBindings.Add (new KeyBinding (Key.Null, Command.LineScrollDown, ()=>ScrollDown(1))); // notice that there are no bound keys for this Command
```

After Change:
```csharp
AddCommand (Command.LineUp, () => MoveUp ());
AddCommand (Command.LineScrollUp, () => ScrollUp (1)); // notice that there are no bound keys for this Command

AddKeyBinding (Key.CursorUp,Command.LineUp);
AddKeyBinding (Key.P | Key.CtrlMask, Command.LineUp);
```

There are several reasons I made the change:

### Planning for future changes while ensuring backwards compatibility
The first implementation (`KeyBindings`) had the public class `KeyBindings` with a public property `Action` which really prevents changing the design later on.  The new version has `KeyBindings` being private and exposes the public methods:

- `public void AddKeyBinding (Key key, Command command)`
- `protected void AddCommand (Command command, Action action)`

Having these 2 entrypoints to the system makes changing the backend without breaking changes easier in future.

### Keybindings -> Action is a 0..m -> 1 operation

There is only one implementation (`Action`) for each command e.g. MoveUp while there can be multiple key combinations that are bound to that action.  In the old method this was handled with:

```csharp
KeyBindings.Add (new KeyBinding (Key.CursorUp,Command.LineUp,()=>MoveUp()));
KeyBindings.Add (new KeyBinding (Key.P | Key.CtrlMask, Command.LineUp, () => MoveUp ()));
```
_Both Up and Ctl+P move the caret up_

But this introduces duplication of the delegate `MoveUp()` which is ugly.

Worse previously was the case where you want to make an action bindable but not provide a key out of the box:

```csharp
KeyBindings.Add (new KeyBinding (Key.Null, Command.LineScrollDown, ()=>ScrollDown(1)));
```
_Before when we wanted to say 'you can scroll down in this UI but theres no explicit key for it until you bind one yourself_

The new system splits the code up into 2 stages

1. Define what common concepts (`Commands`) the control can do (and what `Action` they map to)
2. Define what initial keybindings exist to access the `Commands`

```csharp
// What common Commands does the view support
AddCommand (Command.LineUp, () => MoveUp ());
AddCommand (Command.LineScrollUp, () => ScrollUp (1));

// 2 keybindings for moving up
AddKeyBinding (Key.CursorUp,Command.LineUp);
AddKeyBinding (Key.P | Key.CtrlMask, Command.LineUp);

// initially keybindings for scroll up
```

### Avoiding state based classes / cloning

When considering the case of adding a new key to an existing binding.  With the previous version we would have to do something like

```csharp
var newBind = myView.KeyBindings.Where(k=>k.Command ==  Command.LineUp).Clone();
newBind.Key = Key.J;
myView.KeyBindings.Add(newBind);
```
_Old version of adding secondary bindings_

With the new system it is just:
```csharp
myView.AddKeyBinding (Key.J, Command.LineUp)
```

### Easier to do destructive operations e.g. Clear

Previously if we remove a binding we loose the associated `Action` (what method to call).  For example if we use `Remove` or `Clear` on the keybindings the knowledge of what methods to call is gone.  We would instead have to set them all to `Key.Null` or something.

With the new method we can clear and recreate the `KeyBinding` collection as often as we want as the delegates are safely retained in `CommandImplementations`

### Summary
So while it is a little more code to have to write to establish the mapping in each view I think the benefits outweigh the drawbacks.

I still need to think about how to elegantly handle when the user tries to write a keybinding for a `Command` that is not supported on that `View` but I think that an answer to that problem will emerge.  For now I have a `NotSupportedException` but might move it to when the keybinding is established or just ignore invalid bindings  (I'm not sure).

```csharp
myLabel.AddKeyBinding (Key.J, Command.LineUp)
```
_The label command has no caret or selection so there is no concept of moving it somewhere_

BDisp:
This was a very clarified explanation, thanks. So, I agree that each command must have a valid key, other than `Key.Null`. A command can have multiple keys mapped to it. But a command can only have one action to avoid duplicates delegates. These are very good arguments and I subscribe entirely.

> I still need to think about how to elegantly handle when the user tries to write a keybinding for a `Command` that is not supported on that `View` but I think that an answer to that problem will emerge. For now I have a `NotSupportedException` but might move it to when the keybinding is established or just ignore invalid bindings (I'm not sure).
> 
> ```cs
> myLabel.AddKeyBinding (Key.J, Command.LineUp)
> ```
> 
> _The label command has no caret or selection so there is no concept of moving it somewhere_

I think is better return `false` allowing another view upwards that eventually would need to use it.

BDisp:
@tznind, I think typing characters cannot be handled by the binding keys or you have some idea for this?

Terminal.Gui/Views/Button.cs
```diff
@@ -215,14 +233,36 @@ public override bool ProcessKey (KeyEvent kb)
...
+			if (InvokeKeybindings (kb))
 				return true;
-			}
+
 			return base.ProcessKey (kb);
 		}
 
+		bool ExecuteHotKey (KeyEvent ke)
+		{
+			if (ke.IsAlt) {
```
tznind, line 244: @BDisp why do we need this here when the keybinding for it is:

```csharp
AddKeyBinding (Key.AltMask | HotKey, Command.ExecuteHotKey);
```

I think this is the only place we currently check the KeyEvent of the `Func<KeyEvent, bool>`.  If we can eliminate it then we can make it `Func<bool>` which I think will be more maintainable and useful later on?

tznind:
@BDisp 

I'm looking at Button.cs and see that we are invoking the keybindings in Hot,Cold and ProcessKey.  This means that:

-  Any new keybindings added to Button (or subclasses) are going to be processed at HotKey time (which could be unexpected).
- `CommandImplementations` delegates may be called up to 3 times (e.g. if returning false) which could also be unexpected

I can see the reason is so that the Button's HotKey can be wrapped into the keybinding system (and that is also why there is a new event `HotKeyChanged`).

What do you think about reverting the HotKey system to where it was before (i.e. not rebindable) and only invoking keybindings in ` bool ProcessKey (KeyEvent kb)`.  

We already know there are some things that are not rebindable e.g. typing text or that are already rebindable through other systems e.g. `TreeView.ObjectActivationKey` and `MenuBarItem.Shortcut`.  I think we could make a decision that for simplicity and backwards compatibility we should not to try and integrate those things into Keybindings (for now at least)?

Current code in Button.cs:
```csharp
///<inheritdoc/>
public override bool ProcessHotKey (KeyEvent kb)
{
	if (!Enabled) {
		return false;
	}

	if (InvokeKeybindings (kb))
		return true;

	return false;
}

///<inheritdoc/>
public override bool ProcessColdKey (KeyEvent kb)
{
	if (!Enabled || !IsDefault) {
		return false;
	}

	if (InvokeKeybindings (kb))
		return true;

	return false;
}

///<inheritdoc/>
public override bool ProcessKey (KeyEvent kb)
{
	if (!Enabled) {
		return false;
	}

	if (InvokeKeybindings (kb))
		return true;

	return base.ProcessKey (kb);
}
```

BDisp:
> What do you think about reverting the HotKey system to where it was before (i.e. not rebindable) and only invoking keybindings in ` bool ProcessKey (KeyEvent kb)`.

I agree. We mustn't use the key bindings on `KeyDown/KeyUp` and `HotKey/ColdKey` events. We should only use them on the `ProcessKey` event. It's more than proved that the `Func<bool>` is needed because it may return `true` or `false` from the command.


> I think we could make a decision that for simplicity and backwards compatibility we should not to try and integrate those things into Keybindings (for now at least)?

I agree too.


> If you agree or want to see what it would look like I can make a commit changing it and you can review and if you don't like it we can always revert it back to the way it is now?

Please do the necessary reverts and changes you think is needed directly to your PR. I trust you :-)

Terminal.Gui/Views/Button.cs
BDisp: Agree, go forward. Thanks.

tznind:
GraphView had the following functions (scroll up/down/left/right by 1 unit) and with Ctrl held (scroll up/down/left/right by 5 units).  I've currently bound these to 

- LineScrollUp
- LineScrollDown
- CharRight
- CharLeft

@BDisp should we rename `LineScrollUp` (and `LineScrollDown`) to just `ScrollUp`?

Is there any issue adding `ScrollLeft` and `ScrollRight` do you think?

BDisp:
Yes, I think it's more simple. The only thing we must care is using the prefix "`Move`" if the cursor moves too and using the prefix "`Scroll`" if the cursor doesn't moves. Do you agree?

BDisp:
Another things I did was changing from `void` to `bool` methods to return only `true`. Do you I must revert that and return `true` only in the call command, like you did?

tznind:
> Yes, I think it's more simple. The only thing we must care is using the prefix "Move" if the cursor moves too and using the prefix "Scroll" if the cursor doesn't moves. Do you agree?

Yup I agree :) 

> Another things I did was changing from void to bool methods to return only true. Do you I must revert that and return true only in the call command, like you did?

I don't think it matters.  Its a bit odd to have a method that always returns true.  Especially if it is like 'PageDown', if I saw such a method I would assume it would return true if the scroll position changed and false if it was already at the bottom (for example) but really its likely to always just return `true`.  

But its not a big deal so I wouldn't change the code you have written already.  I think that either in the delegate or in the method is fine.

BDisp:
> I don't think it matters. Its a bit odd to have a method that always returns true. Especially if it is like 'PageDown', if I saw such a method I would assume it would return true if the scroll position changed and false if it was already at the bottom (for example) but really its likely to always just return `true`.

Yes, it must always return true even if the scroll position has changed or not, because if returns false another view can uses the key and it may have an unexpected behavior.



> But its not a big deal so I wouldn't change the code you have written already. I think that either in the delegate or in the method is fine.

Ok, I think so too. Now a method that returns a bool can be used like a void method and there is no problem.

BDisp:
> > Yes, I think it's more simple. The only thing we must care is using the prefix "Move" if the cursor moves too and using the prefix "Scroll" if the cursor doesn't moves. Do you agree?
> 
> Yup I agree :)

About the "`Move`" prefix I was referring on the the methods only and not on the Command enum, because by default all the navigations are supposed to move the cursor. About the "`Scroll`" it must be included in both.

BDisp:
@tznind I will let the `TableView` and `TabView` for you do the changes, right?

tznind:
> @tznind I will let the `TableView` and `TabView` for you do the changes, right?

Sure, I will try to find some time soon to do them.  I think they shouldn't be too hard.

tznind:
@BDisp 

What do you think about:

merge `LeftItem` and `CharLeft` to:
merge `RightItem` and `CharRight` to:

```csharp
/// <summary>
/// Moves the selection left one by the minimum increment supported by the view e.g. single character, cell, item etc.
/// </summary>
Left,

/// <summary>
/// Moves the selection right one by the minimum increment supported by the view e.g. single character, cell, item etc.
/// </summary>
Right,
```

Also I think we should have `ScrollLeft` and `ScrollRight` since we have `LineScrollUp` and `LineScrollDown`

BDisp:
I agree with only `Left` and `Right`, so this way it allows any type.

> Also I think we should have `ScrollLeft` and `ScrollRight` since we have `LineScrollUp` and `LineScrollDown`

In this case I think we should use the same approach, changing the `LineScrollUp` and `LineScrollDown` to `ScrollUp` and `ScrollDown`. If you want differentiate `Line`, then we must also use  `ColumnScrollLeft` and `ColumnScrollRight`.

BDisp:
Have you ever thought of some scenario to use this feature?

BDisp:
@tznind, should we change the `Ctrl+V` to paste in `TextView` as in `TextField` or keep as it is?

tznind:
> Have you ever thought of some scenario to use this feature, when everything is ready, of course.?

I've had some ideas.  Maybe a small sample app Scenario with all our controls in it and a keybinding/rebinding menu in it.  Similar to the one visual studio.  But I've not started anything.

> @tznind, should we change the Ctrl+V to paste in TextView as in TextField or keep as it is?

What is it currently? I think it should probably be the same in both controls.

BDisp:
> What is it currently? I think it should probably be the same in both controls.

In `TextView` the pasting is bonded to `Ctrl+Y` and the `PageDown` is bounded to `Ctrl+V`.

tznind:
> > What is it currently? I think it should probably be the same in both controls.
> 
> In `TextView` the pasting is bonded to `Ctrl+Y` and the `PageDown` is bounded to `Ctrl+V`.

Then lets leave it how it is and see if there is a consensus on whether to make the change.  At the moment this PR is non breaking (I think) so lets try and keep it that way for now.

tznind:
@BDisp I have made the changes we discussed to scroll and left/right (removing the word Char and merging Item and Char together).  And I've added ScrollLeft / ScrollRight.  I also renamed `CharKillToStartLine` to `CutToStartLine` since it's description talks about the clipboard and it seems to function like the Cut operation in windows.  Let me know if you prefer the old name.

```csharp
/// <summary>
/// Deletes and copies to the clipboard the characters from the current position to the start of the line.
/// </summary>
```
see: 718a06c8992010fa4bce1e559971d528bd6be537

tznind:
@BDisp what does `Command.InsertChar` do? it just seems to call this method.  The comment on the command seems to imply it inserts a keystroke which is confusing.  Also though it is bount to the `Insert` keyboard button which normally wouldn't insert any text anyway?

```csharp
void InsertChar ()
{
	Used = !Used;
	SetNeedsDisplay ();
}
```

BDisp:
The name actually isn't right. It would be more correct to call it `ToggleInsertChar` or just `ToggleInsert`, because if the `Used` is true the typed text is appended and moving the existing text on the right if it exists. If the `Used` is false the typed text replaces the existing text in the cursor position. Do whatever you think is better please. Thanks.

tznind:
> > Also though it is bount to the Insert keyboard button which normally wouldn't insert any text anyway?
> 
> No he won't insert anything. Only toggle the `Used` value.
> 
> The name actually isn't right. It would be more correct to call it `ToggleInsertChar` or just `ToggleInsert`, because if the `Used` is true the typed text is appended and moving the existing text on the right if it exists. If the `Used` is false the typed text replaces the existing text in the cursor position. Do whatever you think is better please. Thanks.

Cool, I have renamed it in d972321350770fffe999a46ec0656a4b972a1847 and added commands for enable/disable too although theres no associated key configured for that out of the box.

BDisp:
I think that the `EnableOverwrite` and the `DisableOverwrite` wouldn't be necessary, since they are related with the `ToggleOverwriteMode`, but they may stay and waiting to see if they are really needed.

tznind:
I was thinking again about a Scenario for keybinding and was thinking of just adding a menu item to the root of the app for changing keybindings for the whole program (similar to the ColorScheme changing menu item).

I was looking at how to implement that and I think it should be possible by tapping the `Added` event and tracking all the views as they are created so that new views get the correct keybindings (if the user has made changes).  Obviously we would only start this process if they opened the menu and made changes to the defaults.

BDisp:
That is great @tznind. I'm finishing the harder view to do the changes, the `TextView`. If you think there is other views that need to be implementing, please say. Thanks.

tznind:
> If you think there is other views that need to be implementing, please say. Thanks.

I was looking through the repo and did notice that `Autocomplete` should probably also support keybinding.  But its not a `View` so will need to have a think about how to do that (or skip it for now).

BDisp:
> I was looking through the repo and did notice that `Autocomplete` should probably also support keybinding. But its not a `View` so will need to have a think about how to do that (or skip it for now).

At least in the `TextView` I'm considering using the keybinding for the `Autocomplete`, but not inside the `ProcessKey` of the `Autocomplete`. I let it to you do it, or skip it for now.

BDisp:
Sorry I couldn't include the `Autocomplete` in the keybindings. I think this is finished.

BDisp:
@tznind we must rethink about the return value for the `InvokeKeybindings` method. The return values `true` or `false` must be immediately returned to the caller, instead of continuing execution until end. This way it's more faster. So, it only continuing executing forward if the `InvokeKeybindings` return null, because there is no key binding associated or if a key binding returns null.

tznind:
@BDisp I have added the Keybindings menu item to UICatalog as we discussed.

tznind:
I've marked this ready for review @tig and updated the original post in the PR to describe the changes.  Happy to talk through any of the design choices made in it.  I don't think there are any breaking changes. 

Big thanks to @BDisp who did more than half the work on this :)

BDisp:
To position a view at the bottom or at the right inside a view is always better to use the `Pos.AnchorEnd(margin)`. To position related at the bottom or at the right outside the view is better to use the `Pos.Bottom(View)` or the `Pos.Right(View)`. The problem using the `Pos.Bottom` or the `Pos.Right` to position a view inside another view we need to take an account of the border length or the child border length and the `Pos.AnchorEnd` does that job.

tznind:
Bit of a strange one @BDisp.  fdf43e7 failed in `Run_All_Scenarios`.  I added a log write line in 16a3dfa and the test run passed.  The test also passes locally for me in both Windows and Linux.  So I guess there must be a random corner case where that test fails.  Maybe test execution order or a race condition inside one of the Scenarios when run by that test.  Have you seen this before or know about it?

At least if it happens again the logging code should show which Scenario was being tested at the time it got an `abort`.

BDisp:
> Have you seen this before or know about it?

Yes, often. Normally when this happens it's run locally without errors and when submitting to the server I receive an email reporting the failure. The scenarios at starting must not perform anything that will prevent respond to a request stop from application to not exceeds the count limit. But I think that was a sporadic error because the commit you have done doesn't suppresses it in case the bug is still there. Otherwise, the error still kept.

tig:
This PR looks awesome. I'm gonna try to find some time today to dive deep into it.

Do you guys both agree that it's ready to merge?

BDisp:
To me it's good enough to be merged without any breaking.

BDisp:
Added `HistoryText` class to allow undoing and redoing changes on `TextView` and on the `TextField`. See here [BDisp:keybindings-continuation](https://github.com/tznind/gui.cs/pull/80)

historytext

tznind:
Hey @tig , @BDisp has done some great work on undo/redo and refactoring `HistoryText` (see https://github.com/tznind/gui.cs/pull/80).  Do you want that combined with this PR or seperate? I don't want to add more code if you have already started reviewing it...

UICatalog/UICatalog.cs
```diff
@@ -331,6 +332,22 @@ private static Scenario GetScenarioToRun ()
...
 		}
+		private static MenuItem[] CreateKeybindings()
+		{
+
+			List<MenuItem> menuItems = new List<MenuItem> ();
+			var item = new MenuItem ();
+			item.Title = "Keybindings";
+			item.Action += () => {
+				var dlg = new KeyBindingsDialog ();
+				Application.Run (dlg);
```
tig, line 343: If I use this dialog to change `QuitTopLevel` to something other than CTRL-Q, UI Catalog doesn't update it's UI.

Wouldn't it be good to implement that as test that all this works end to end?

tig (approved):
I think this is all fantastic. There's TONS in here and "Keybindings" as the name of the PR does not do it justice.

BDisp:
@tznind can you change the title of this pr without the Undo/Redo please? This feature is in #1590. Given it was already merged I don't know if it's possible. Thanks.

PR #1561, opened 2021-12-25
Fixes #1559 - Added DesiredCursorVisibility to TreeView

### LINKED ISSUE #1559: Should TreeView hide the cursor the same way ListView does?
When I create an app with just a `TreeView` the cursor is visible, it follows the selection, and it does not look good.

`ListView` calls `SetCursorVisibility(CursorVisibility.Invisible)` in its `OnEnter` handler.
Should `TreeView` do the same?

tznind:
Fixes #1559


tig (approved):
LGTM

PR #1574, opened 2022-01-13
Use custom resolver for libcoreclr.so also for .NET 6.0

henrikx:
Running into the same issue as in #889, I noticed the custom resolver code was not included for .NET 6.0 . This patch fixes the `System.DllNotFoundException` for single-file mode also in .NET 6.0

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
@@ -258,7 +258,7 @@ namespace Unix.Terminal {
 		/// </summary>
 		static class CoreCLR
 		{
-#if NET6_0 || NET5_0
+#if NET6_0
 			// Custom resolver to support true single-file apps
 			// (those which run directly from bundle; in-memory).
 			//     -1 on Unix means self-referencing binary (libcoreclr.so)
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
@@ -258,7 +258,7 @@ namespace Unix.Terminal {
 		/// </summary>
 		static class CoreCLR
 		{
-#if NET5_0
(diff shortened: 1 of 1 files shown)


BDisp (approved):

BDisp:
@henrikx I'm still having the `System.DllNotFoundException` error even with the `.NET 5.0`. How can I run it successful?

henrikx:
@BDisp I am not sure about .NET 5.0, but this patch definitely fixed it for .NET 6.0

BDisp:
But I'm still having the same error with the `.Net 6.0`. If I run my app from the `/net6.0/linux-x64/` with `dotnet ./myapp.dll` it runs without errors. But if I run it from `/net6.0/linux-x64/publish/` with `./net6.0/linux-x64/publish/myapp` it gaves me the `System.DllNotFoundException` error.

BDisp:
@ henrikx, sorry, it working now. The reason for not working with `.Net 5.0`, was because the `Terminal.Gui` project doesn't include the `.Net 5.0`

Include this change in the `Terminal.Gui` on your PR:

`<TargetFrameworks>net472;netstandard2.0;net5.0;net6.0</TargetFrameworks>`

BDisp (commented):
May be is better to change to `#if NET6_0` only and exclude the `NET5_0`.

henrikx:
@BDisp Please check now. I have implemented the suggested changes.

BDisp:
> @BDisp Please check now. I have implemented the suggested changes.

That is great @henrikx, thanks. Perhaps in the future, be better to keep the last two versions of the `.NET` in the projects to supports the two latest versions.

BDisp (approved):

henrikx:
> Perhaps in the future, be better to keep the last two versions of the `.NET` in the projects to supports the two latest versions.

Not sure what you mean by this or in which projects you mean. This PR is good for merge now, right?

BDisp:
> Not sure what you mean by this or in which projects you mean. This PR is good for merge now, right?

Forget what I have said before, please. The most important is to change this every time the `Terminal.Gui` project is updated with the last `.NET` version. Your PR is good to merge but you have to wait for the approval and merging by @tig, because I'm only a contributor.

PR #1578, opened 2022-01-26
Fixes #1577 - Add Support for Colored ListView Items

### LINKED ISSUE #1577: ListView items color
Is there a way to change background/text colors of each item of a ListView ?

aurelienverbeke:
We can now choose the color (background and foreground/text) we want to display each item.

You need to set a delegate to `ListView.RowRender(ListViewRowEventArgs obj)`, and set the `Attribute rowAttribute` property of the `obj`.
You can get the index of the row displayed with the `int Row` property of the `obj`.
This event is called for each row displayed.

Fixes #1577

### DIFF
--- a/Terminal.Gui/Views/ListView.cs
+++ b/Terminal.Gui/Views/ListView.cs
@@ -347,6 +347,12 @@ namespace Terminal.Gui {
 					for (int c = 0; c < f.Width; c++)
 						Driver.AddRune (' ');
 				} else {
+					var rowEventArgs = new ListViewRowEventArgs (item);
+					OnRowRender (rowEventArgs);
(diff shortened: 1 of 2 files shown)


BDisp (approved):

BDisp:
You have to wait for @tig to merge this.

aurelienverbeke:
I know, I just wanted to rename my branch in order to it corresponds better to the modifications...

BDisp:
That's not possible. It's the same as deleting this PR and open another, but not worth it.

tig (approved):
Nice work. Thank you for your contribution!

If you are feeling extra generous, enhancing the unit tests to ensure your new code doesn't break in the future would be much appreciated.

PR #1593, opened 2022-02-16
Fixes TableView not activating after keybinding refactor

tznind:
Fixes a bug in TableView activation introduced in the keybindings PR


tig (approved):
LGTM

PR #1599, opened 2022-02-17
Adding unit test for the ListView RowRender event.

BDisp:
As requested in https://github.com/migueldeicaza/gui.cs/pull/1578#pullrequestreview-874093482.


tig (approved):
Nice!

PR #1603, opened 2022-02-19
Fixes the auto-size text bug.

### LINKED ISSUE #1602: Auto-size text is not drawn correctly.
Text is placed on the previous frame. The previous frame must be cleared by the superview and the new text must be place in the current frame.

text-bug

Another issue:
auto-size-text-button-bug

BDisp:
Fixes #1602.

text-fix


BDisp:
There is still an issue, because the superview must be responsible for clearing the previous frame.

auto-size-still-having-issue

tig (approved):

PR #1606, opened 2022-02-19
Fixes ComboBox text and list if the height is equal to one.

### LINKED ISSUE #1605: ComboBox has a strange behavior on the Unicode scenario.
The letter 'g' is hidden. Sometimes the key F4 doesn't expand the dropdown list.

combobox-bug

BDisp:
`End` and `Home` keys now navigate on text if `IsShow` is false.

Fixes #1605.


BDisp:
Well I give up. I don't know why this failure is happening.

BDisp:
With the Dev Container on VSCode I also don't get this failure error. It only happens on the github server.

tig:
In the test results:

```
[xUnit.net 00:00:02.30]     Terminal.Gui.Core.MainLoopTests.AddTimer_Duplicate_Keys_Not_Allowed [FAIL]
  Failed Terminal.Gui.Core.MainLoopTests.AddTimer_Duplicate_Keys_Not_Allowed [13 ms]
  Error Message:
   System.InvalidOperationException : Collection was modified after the enumerator was instantiated.
```

Looking at the code in `MainLoop.cs` I'm wondering if we're missing a lock within `RunTimers`:

image

We lock `timeouts` in `AddTimer()`, but not here. 

image

The `foreach` in `RunTimers()` iterates over `copy` which is a reference to the `SortedList<long, Timeout>` that was created earlier and referred to by `MainLoop.timouts`.  Even though the unit tests are set to run single-threaded, the apparently don't always do that. Thus some other test code is calling something (`AddTimer()`?) that is modifying that `SortedList<long, Timeout>` instance. I think.

So I think there is a re-entrancy bug here. I also think it's very frustrating that the unit tests get run concurrently within the github environment.

tig:
The other place `timeouts` is modified is here:

image

BDisp:
May be the mainloop timeouts are not disposed correctly. But the CharacterMap scenario failure is more stranger.

Terminal.Gui/Core/MainLoop.cs
```diff
@@ -162,7 +162,9 @@ public bool RemoveTimeout (object token)
 			var idx = timeouts.IndexOfValue (token as Timeout);
 			if (idx == -1)
 				return false;
-			timeouts.RemoveAt (idx);
+			lock (timeouts) {
```
tznind, line 165: Should this lock also cover the `IndexOfValue` call? If multithreading is a possibility.

BDisp: I was not already giving failing but I think you're right. Done.

tig (approved):

Issue #1611, opened 2022-02-21
Issues with Application.Top being null if dispose is called on the window object


FreddieMercurial:
https://github.com/migueldeicaza/gui.cs/blob/1557f7a71e3dd4a0163d08375e2788dd687a7cba/Terminal.Gui/Core/View.cs#L366

![image](https://user-images.githubusercontent.com/3766240/155020618-a533c3a2-aac3-4bc6-ba19-467ae419c4af.png)

This line does it the correct way
https://github.com/migueldeicaza/gui.cs/blob/1557f7a71e3dd4a0163d08375e2788dd687a7cba/Terminal.Gui/Core/View.cs#L359

BDisp:
I did not understand the comparison you did. It seems you are disposing an object that was already disposed.

FreddieMercurial:
_window was not itself previously disposed- but encounters the error I pointed out trying to read Application.Top.Focused when .Top is null

tznind:
Thanks for identifying an issue and the fix.  Are you able to write a unit test that shows the issue? Something like:

```csharp
[Fact]
[AutoInitShutdown]
public void WindowDispose_FocusProblem()
{
    // TODO: Add other statements here to reproduce your issue
    new Window().Dispose();
}
```

BDisp:
Thanks for reporting this. I already found a bug where a view that has focus and setting `CanFocus` to false is setting `HasFocus` to `false` and to `true` again. I'll submit a PR with a fix.
Nonetheless, I couldn't reproduce the disposing focus issue. Can you provide more information about how you get the exception, or test with the fix I'll provide to check if you still getting the exception again.

FreddieMercurial:
Sorry for being slow on the response! Glad it wasn't PEBKAC after all!
Will give it a run!

tig:
Closing as not-repo.

PR #1612, opened 2022-02-22
Fixes the CanFocus equal to false from setting HasFocus to true again.

BDisp:
By checking this issue #1611 I found this bug, but not yet the disposing issue.

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -356,18 +356,16 @@ namespace Terminal.Gui {
 					}
 					TabStop = value;
 
-					if (!value && Application.Top?.Focused == this) {
-						Application.Top.focused = null;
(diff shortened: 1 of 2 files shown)


FreddieMercurial:
I know you folks were already way on top of this, but I just realized you'd all responded and I'd forgotten about the thread and never expected such a quick and hearty response!
You're awesome. I'll do a quick test too and +1 that other ticket in the next moments here.

FreddieMercurial:
Replicated on e3b62cff38a8ed0a460a26d73951710aea348bbd:
image

Passes afterwards:
Is there an Assert.DoesNotThrow type thing in your test framework?
image

FreddieMercurial:
https://github.com/migueldeicaza/gui.cs/pull/1623

PR #1614, opened 2022-02-22
Fixes the ProcessHotKey event.

### LINKED ISSUE #1613: The StatusBar ProcessHotKey is not working on the Example project by pressing Ctrl+Q.
This is caused because the `ProcessKey` is passed first for the focused view, which could be a toplevel, for example, instead of calling the most focused view.

BDisp:
Running the unit test without the fix will fail.

Fixes #1613.


tig:
@BDisp please resolve the conflicts por favor.

PR #1615, opened 2022-02-24
Adds a popup ContextMenu feature. Implements ContextMenu for TextField.

BDisp:
The default key is Shift+F10 and the default mouse flag is the right click.

contextmenu


tznind:
That looks great!

BDisp:
Thanks. I'll improve in the case where it's tied with a view, like the `TextField` is, to popup above the view if there is no necessary space bellow.

Terminal.Gui/Core/Window.cs
```diff
@@ -97,9 +97,9 @@ public override void OnCanFocusChanged ()
 				base.OnCanFocusChanged ();
 			}
 
-			public override bool MouseEvent (MouseEvent mouseEvent)
+			public override bool OnMouseEvent (MouseEvent mouseEvent)
```
tig, line 100: Isn't renaming a public member a breaking change?

Terminal.Gui/Views/Menu.cs
```diff
@@ -998,9 +1007,10 @@ void Selected (MenuItem item)
 		public event Action<MenuItem> MenuOpened;
 
 		/// <summary>
-		/// Raised when a menu is closing.
+		/// Raised when a menu is closing passing <see langword="true"/>
+		/// if it's closing a sub-menu, <see langword="false"/> otherwise.
 		/// </summary>
-		public event Action MenuClosing;
+		public event Action<bool> MenuClosing;
```
tig, line 1013: Isn't changing the signature of an event a breaking change?

Terminal.Gui/Views/TextField.cs
```diff
@@ -854,6 +901,51 @@ int WordBackward (int p)
 			return -1;
 		}
 
+		void ShowContextMenu ()
+		{
+			ContextMenu = new ContextMenu (this,
+				new MenuBarItem (new MenuItem [] {
+					new MenuItem ("_Select All", "", () => SelectAll (), null, null, GetKeyFromCommand (Command.SelectAll)),
```
tig, line 908: By adding `ContextMenu` support to `TextField` this way you've raised an interesting question:

- Should built-in strings within the `Terminal.Gui` library be in English or stored in a way that is easily localizable?

Where else within the library do we have English text like this?

UnitTests/ContextMenuTests.cs
```diff
@@ -0,0 +1,425 @@
+﻿using Xunit;
+using Xunit.Abstractions;
+
+namespace Terminal.Gui.Views {
+	public class ContextMenuTests {
```
tig, line 5: ❤️❤️❤️❤️❤️

UnitTests/MenuTests.cs
```diff
@@ -119,7 +119,7 @@ public void MenuOpening_MenuOpened_MenuClosing_Events ()
 				e.Action ();
 				Assert.Equal ("Copy", miAction);
 			};
-			menu.MenuClosing += () => {
+			menu.MenuClosing += (_) => {
```
tig, line 122: The fact you had to change this test, proves it is a breaking change!

tig (changes requested):
What would it take to :

1) Implement this amazing new feature without making breaking changes to the existing library API?
2) Not introduce more English text built into the library?

UnitTests/MenuTests.cs
BDisp: That's was a recent event implementation and I think it will not having a bigger impact.

UnitTests/ContextMenuTests.cs
BDisp: Ops I will change it to Core.

Terminal.Gui/Views/TextField.cs
BDisp: Well that's will be a challenge.

Terminal.Gui/Views/Menu.cs
BDisp: That's was a recent event implementation and I think it will not having a bigger impact.

Terminal.Gui/Core/Window.cs
BDisp: OnMouseEvent already exists and it will also call the MouseEvent and I didn't needed to change any unit test because of this change.

Terminal.Gui/Views/TextField.cs
tznind: resx files would normally be how you set the library up for translations.  It uses the running systems culture to pick which resx file resource to use for the given language.  But you would need the translations.

https://docs.microsoft.com/en-us/dotnet/core/extensions/localization#resource-files

The only place I can think of currently where we use hard coded English is `OpenDialog`

Terminal.Gui/Views/Menu.cs
tznind: If we are changing this to take a parameter should we make it `MenuClosingEventArgs` or `MenuEventArgs` or something? that would future proof it for later on if we want to add more info about the menu event that is happening e.g. a cancel property etc.

Terminal.Gui/Core/Window.cs
BDisp: This goes unnoticed for users and does not influence their functioning at all, because it's implemented on the private `ContentView` sub-class of the `Window` class.

UnitTests/ContextMenuTests.cs
BDisp: Done!

Terminal.Gui/Views/Menu.cs
BDisp: Done! https://github.com/migueldeicaza/gui.cs/pull/1615/commits/a2351b818105a7e45a3791befb382116298289c0

UnitTests/MenuTests.cs
BDisp: Done! https://github.com/migueldeicaza/gui.cs/pull/1615/commits/a2351b818105a7e45a3791befb382116298289c0
Improvements are more advantageous.

BDisp:
I implemented localization for the `TextField` and the `FileDialog` for the `en-US` (default) and `pt-PT` languages.

contextmenu-localization

Terminal.Gui/Views/TextField.cs
BDisp: Done! https://github.com/migueldeicaza/gui.cs/pull/1615/commits/b867ab62fe080cb4f53f002a40772183369f022a

BDisp:
Several popups layouts.

contextmenu-layouts

tig:
> I implemented localization for the `TextField` and the `FileDialog` for the `en-US` (default) and `pt-PT` languages.
> 
> contextmenu-localization

The addition of the localization support for FileDialog should be a separate PR. Can you please submit that as a separate PR first. I'll review and merge that, then I will merge the ContextMenu PR.

BDisp:
> The addition of the localization support for FileDialog should be a separate PR. Can you please submit that as a separate PR first. I'll review and merge that, then I will merge the ContextMenu PR.

Done @tig. I only did a Cherry-Pick of the https://github.com/migueldeicaza/gui.cs/pull/1625/commits/689ce6fca65a92f6bb7804b5d42ee20cbb41ff3c on this PR.

PR #1622, opened 2022-03-01
Changed Button Text property to use override instead of new.

### LINKED ISSUE #1619: Use of `new` on Text property of Button
**Describe the bug**
Button has `new` on its Text property.  This means that if you maintain a reference to it as a `View` (e.g. keeping a list of views in your class) and assign its `.Text` property then it is not properly preserved for use later on if you reference it as its explicit Type again.

This could be problematic in quite a few situations especially if using reflection.  Is it possible to make `Button` work without the `new` keyword or make the Text property virtual and use `override` instead of `new` in Button if inheritence is the only way to make it work?

```csharp
[Fact]
public void TestAssignTextToButton ()
{
	View b = new Button ();
	b.Text = "heya";
	Assert.Equal ("heya", b.Text);

	// fails here
	Assert.Equal ("heya", ((Button)b).Text);
}
```

image

BDisp:
Fixes #1619.

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -186,12 +186,12 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Gets or sets the HotKey defined for this view. A user pressing HotKey on the keyboard while this view has focus will cause the Clicked event to fire.
 		/// </summary>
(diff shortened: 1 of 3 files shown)


tig (approved):
Thanks.

PR #1625, opened 2022-03-05
Implementing localization for en-US and pt-PT to the FileDialog.

BDisp:
As requested by @tig here https://github.com/migueldeicaza/gui.cs/pull/1615#issuecomment-1059649833.


UnitTests/ApplicationTests.cs
```diff
@@ -1279,5 +1279,12 @@ public void EnsuresTopOnFront_CanFocus_False_By_Keyboard_And_Mouse ()
 			win2.MouseEvent (new MouseEvent () { Flags = MouseFlags.Button1Released });
 			Assert.Null (Toplevel.dragPosition);
 		}
+
+		[Fact, AutoInitShutdown]
+		public void GetSupportedCultures_Method ()
```
tig, line 1284: I'm curious: Is there a need to have a unit test that verifies if a culture is supported, it defines the full set of required strings? E.g. if we accidentally forget to define the PT translation of some word in the future, will something in the build or tests fail?

tig (approved):
Thank you SO much for doing this.

UnitTests/ApplicationTests.cs
BDisp: No it doesn't fail. This test only returns the number of assembly satellites.

tznind: I think if a string is missing from a language then it will fallback on the default language specified in the assembly.  Therefore you would get a menu of words all in native language and one in English because it was missing from the native resx.

https://stackoverflow.com/questions/2272744/localization-net-fallback-language-when-using-resourcemanager

There must be a pretty simple way of determining if there are missing translations though.

BDisp: > I think if a string is missing from a language then it will fallback on the default language specified in the assembly. Therefore you would get a menu of words all in native language and one in English because it was missing from the native resx.

Yes, is true, but I recommend to make it consistent for all languages.
The unit test don't check that, only gets all the `.resources.dll` files that exists in the execution assembly:

`\gui.cs\Terminal.Gui\bin\Debug\net6.0\pt-PT\Terminal.Gui.resources.dll`
`\gui.cs\Terminal.Gui\bin\Debug\net472\pt-PT\Terminal.Gui.resources.dll`
`\gui.cs\Terminal.Gui\bin\Debug\netstandard2.0\pt-PT\Terminal.Gui.resources.dll`


> No it doesn't fail. This test only returns the number of assembly satellites.

I like to use the free extension `ResXResourceManager` (https://github.com/dotnet/ResXResourceManager) which allow to add/edit/remove all the strings for all languages at the same time. Therefore, is very easy to create a new `.resx` file and translate the strings for that new language.

PR #1628, opened 2022-03-07
Adds true color support

### LINKED ISSUE #48: Truecolors (direct RGB colors) support

veeman:
Fixes #48 

Added true color support.
Based on the #103 pull request implementation.

As in this implantation the `struct Attribute` changed to `class Attribute`.
Also there are now `TrueColorAttribute` and `TrueColor` definitions wich can be used as dropin classes to provide the new feature.

The abstract `ConsoleDriver` has two more properties which determinates the true color support state of the driver:
```c#
bool SupportsTrueColorOutput { get; }
bool UseTrueColor { get; set; }
```

The drivers `WindowsDriver` and `NetDriver` are extended to render the new true colors.
Also the `WindowsDriver` is optimized to do the rendering faster.
_The `CursesDriver` is not touched yet and would run only in the basic color mode._

Moreover i added an exampel scenario and some basic unit test for this feature.

Currently only tested on Windows 10.

Any feedback is appreciated.


BDisp:
I'm loving your work on this, thanks. I think for `CursesDriver` you will need to use `can_change_color` and `init_color`. See https://stackoverflow.com/questions/25091638/other-colors-in-ncurses and https://invisible-island.net/ncurses/man/curs_color.3x.html#h3-can_change_color.

NoUDerp:
I'm using this to render simple canvases as a window. I made a modification for a custom constructor for TrueColorAttribute to allow providing the linux 256 color (calculated on a GPU) to avoid calculating it on the CPU, for flexibility. I'm taking quite a GC beating over the high frequency allocations of TrueColorAttribute being a reference type when I update the frame (it's video). The best solution I came up with was making the base Attributes Foreground, Background, and Value into read/write, but requires I run a modified library.

public TrueColorAttribute(TrueColor foreground, Color foreground_color4, TrueColor background, Color background_color4)
			: base(foreground_color4, background_color4)
{
			TrueColorForeground = foreground;
			TrueColorBackground = background;
}

veeman:
I looked into the ncurses library.
They provide for true color support the methods init_color and init_extended_color.
However this methods are used to define indexed rgb values which in turn can then be used to identify the color. 

With the current implementation, we need some sort of mapping to these indexed colors and keep track of them.

Currently this sounds very inefficient to me. So I will not do this. 

The true collor support works just fine. The NetDriver also works on Linux and Mac.


In my experience, the bottleneck for rendering various things is in the drivers using StringBuilder 
that are instantiated on each run and the entire screen is reprinted.

The whole Drivers could be optimized further, but this should be another task.

tig:
This PR is pretty darn exciting. I'd love to get it merged as soon as it is ready. 

@veeman could you update us on your ability/plans to complete this awesome work?

Thanks!

FreddieMercurial:
Awesome, I still have to keep my promise to make HACC support Spectre.Console and it supports true color so this will be hot as snot!

Terminal.Gui/Core/ConsoleDriver.cs
```diff
@@ -82,6 +82,86 @@ public enum Color {
 		White
 	}
 
+	/// <summary>
+	/// 24bit color. Support translate it to 4bit(Windows) and 8bit(Linux) color
```
tig, line 86: Change summary to:

```
/// Represents a 24bit color. Supports translating to 4bit (Windows) and 8bit (Linux) colors.
```

Terminal.Gui/Core/ConsoleDriver.cs
```diff
@@ -82,6 +82,86 @@ public enum Color {
...
+		/// </summary>
+		public int Blue { get; }
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="TrueColor"/> struct.
+		/// </summary>
+		/// <param name="red"></param>
+		/// <param name="green"></param>
+		/// <param name="blue"></param>
+		public TrueColor (int red, int green, int blue)
```
tig, line 108: Please apply correct source code formatting to all files in this PR (hit Ctrl-K-D in VS).

Terminal.Gui/Core/ConsoleDriver.cs
```diff
@@ -90,7 +170,7 @@ public enum Color {
 	///   scenarios, they encode both the foreground and the background color and are used in the <see cref="ColorScheme"/>
 	///   class to define color schemes that can be used in your application.
 	/// </remarks>
-	public struct Attribute {
+	public class Attribute : IEquatable<Attribute> {
```
tig, line 173: Technically this is a breaking API change. This PR must be marked as "BREAKING CHANGE" and we need to determine whether this will actually break existing code or not.

Terminal.Gui/Core/ConsoleDriver.cs
```diff
@@ -191,6 +277,44 @@ public static Attribute Get ()
...
+		/// <summary>
+		/// Default empty attribute.
+		/// </summary>
+		public static readonly Attribute Default = new Attribute ();
+	}
+
+	/// <summary>
+	/// Defines a true color attribute.
+	/// </summary>
+	public class TrueColorAttribute : Attribute {
```
tig, line 290: I'm confused as to why this class is needed? Why not just put this into `Attribute` and make it so true color is the default/assumed, supporting non-truecolor as a fallback?

When would someone use this class?

Terminal.Gui/Core/ConsoleDriver.cs
```diff
@@ -246,109 +370,109 @@ Attribute SetAttribute (Attribute attribute, [CallerMemberName] string callerMem
 			case "TopLevel":
```
tig, line 370: This whole switch statement scares me. So much brute-force code! It makes me wonder if there's a way of re-factoring it as part of this PR to use reflection or something to radically simplify it?

Terminal.Gui/Core/ConsoleDriver.cs
```diff
@@ -668,6 +792,23 @@ public abstract class ConsoleDriver {
 		// The format is rows, columns and 3 values on the last column: Rune, Attribute and Dirty Flag
 		internal abstract int [,,] Contents { get; }
 
+
+		/// <summary>
+		/// Determinates if the current console driver supports TrueColor output-
+		/// </summary>
+		public virtual bool SupportsTrueColorOutput { get => false; }
```
tig, line 799: I'm confused as to why there's ALSO a method in `CursesDriver.cs` called `CanColorTermTrueColor`. Why not just have `CursesDriver` implement an override of this method?

I think this method should be named `SupportsTrueColor` for tersness.

tig (changes requested):
This PR is a big deal. And great.

1. I believe this is a breaking change and that there's probably no way to NOT make it a breaking change. Thus this PR will likely define Terminal.Gui v2.x. 
2. The ColorPicker and other Scenarios that demo colors should be updated to use TrueColor support. I suggest that be done as separate PR's that are dependent on this one. 
3.  I've asked for some changes throughout. I did not run this code and my review was fairly superficial. We'll need many more eyes on this given the magnitude of the changes.

Terminal.Gui/Core/ConsoleDriver.cs
BDisp: This must be abstract in the ConsoleDriver.cs to force to be implemented in all drivers.

tig:
Marked for the v2.0 milestone.

Terminal.Gui/Core/ConsoleDriver.cs
veeman: We have to distinguish between basic colors and extended colors. This is necessary to decide which escape code format to use later. The easiest way to do this is to change the color attribute struct to a class.

Terminal.Gui/Core/ConsoleDriver.cs
veeman: As describe before we need to decide which color escape sequence we should use. To keep backward compatibility i just introduced a new TrueColorAttribute class which holds the rgb values, like in the other PR.

This method has also he advantage that you only need to create new color escape sequences when the instance reference changes. However i can not evaluate if this has further side effects.

Furthermore there is no easy way to merge *basic* color and *truecolor* attribute class because any platform defines its own basic color range. The later must be unified first.

Terminal.Gui/Core/ConsoleDriver.cs
veeman: The `CursesDriver` currently does not implement the TrueCollorAttribute feature.
On *unix, there is no *easy* way to determinate if the current terminal supports color output. 
Despite *ncruses* provides a check for color support, it tells only if the library is compiled with color output support.

The unofficial community way is to check the environment variable `COLORTERM` for some magic values. This does the static `CursesDriver.CanColorTermTrueColor` function. This variable is also set under WSL, so i just retrive this information in the `NetDriver`.

The virtual property `SupportsTrueColorOutput` tells if the used driver and the used terminal environment supports color output.

The property `UseTrueColor` is used to decide which mode should be used. If `SupportsTrueColorOutput` is set then `UseTrueColor` is also set, but can be uesed to enter baisc color rendering.

veeman:
I merged this PR on the latest devel branch.

The ` WindowsDriver` and `NetDriver` is feature complete however should be performance tested.

Attached an image of the *True Colors* ui catalogue scenario:
screen

The `CursesDriver` **can currently not be extended** (i have no plan how to do this) with the current attributes architecture to provide color output. *ncurses* needs pre initialized indexed color values which is not compatible with the current implementation.

BDisp:
> The `CursesDriver` **can currently not be extended** (i have no plan how to do this) with the current attributes architecture to provide color output. _ncurses_ needs pre initialized indexed color values which is not compatible with the current implementation.

Great work thanks. Do you have any idea how to achieve that?

Terminal.Gui/Core/ConsoleDriver.cs
tznind: Would it be possible to write a mapping algorithm for true color to Console Color?

An approximation would be fine, I'm guessing someone has already done this elsewhere.

If we could translate between the two then we could just add the properties to `Attribute` and add a new true color constructor.  So the pseudocode would be something like:

```csharp
public class Attribute
{
	/// <summary>
	/// Red color component.
	/// </summary>
	public int Red { get; }
	/// <summary>
	/// Green color component.
	/// </summary>
	public int Green { get; }
	/// <summary>
	/// Blue color component.
	/// </summary>
	public int Blue { get; }

	public Color Foreground { get; }
	public Color Background { get; }

	public Attribute (Color foreground, Color background)
	{
		Foreground = foreground;
		Background = background;
		var m = MapToTrueColor (foreground, background);
		Red = m [0];
		Green = m [1];
		Blue = m [2];
	}
	public Attribute (int red, int green, int blue)
	{
		Red = red;
		Green = green;
		Blue = blue;

		var b = MapToBasicColor (red,green,blue);			

		Foreground = b[0];
		Background = b[1];
	}

    [...]
}
```

That would allow clients to mix and match true color whenever they wanted.  And even have an app that supports true color but falls back on basic colors when deployed to an older machine (with zero effort on the API consumers behalf)?

veeman: Yeah, something like that could work. Maybe we can track the index by the lifetime of the Attribute class and use a database approach to select a free index by incrementing a static _next-index_ value.

tznind: Apparently [this is the RGB breakdown](https://stackoverflow.com/a/36674533/4824531).  I might pull your branch and see if a 'transparent' and backwards compatible approach can be adopted following this approach.

|Name    |     R |   G |   B|
|--------------|---|----|----|
|Black        |00 | 00  | 00   |
|DarkBlue     |00 | 00  | 80   |
|DarkGreen    |00 | 80  | 00   |
|DarkCyan     |00 | 80  | 80   |
|DarkRed      |80 | 00  | 00   |
|DarkMagenta  |80 | 00  | 80   |
|DarkYellow   |80 | 80  | 00   |
|DarkGray     |80 | 80  | 80   |
|Blue         |00 | 00  | FF   |
|Green        |00 | FF  | 00   |
|Cyan         |00 | FF  | FF   |
|Red          |FF | 00  | 00   |
|Magenta      |FF | 00  | FF   |
|Yellow       |FF | FF  | 00   |
|Gray         |C0 | C0  | C0   |
|White        |FF | FF  | FF   |

tznind: @veeman I've started working on this approach in https://github.com/tznind/gui.cs/tree/true-color 

Once its ready (and assuming its viable) I will open a PR to your branch.

tznind:
@veeman I've created a PR here https://github.com/veeman/gui.cs/pull/2 with some improvements and changes.  Including this new Scenario for opening images:

Note: You can see that when toggling true/non-true color the exact RGB values rendered on screen change.  I need to dig into why this is but I think its because each console will render `Color.X` to a slightly different RGB value (which might prove a problem).

sauron

tig:
We need to retarget this PR at the `v2_develop` branch.

tig:
Done! See #2612 

(and #2729)

PR #1634, opened 2022-03-10
Feature to display the sub-menus on a single frame instead multiple frames.

BDisp:
The hotkey access was also improved, now pressing `Alt+HotKey` will immediately open the associated menu bar and then pressing a `HotKey` will run an action or open a sub-menu.

UseSubMenusSingleFrame

### DIFF
--- a/Example/demo.cs
+++ b/Example/demo.cs
@@ -119,10 +119,10 @@ static class Demo {
 		int i = 0;
 		string txt = "Hello world, how are you doing today?";
 		container.Add (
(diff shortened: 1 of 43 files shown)


BDisp:
Also implemented in the context menu.

UseSubMenusSingleFrame-contextmenu

BDisp:
Context menu now gets the right view to screen when it have a host by not overlapping it.

contextmenu-view-to-screen

tig:
My apologies if I merged these in the wrong order causing the conflict. I tried to do it in the order they were listed; not sure what I did wront.

PR #1635, opened 2022-03-10
Rootkeyevent

tznind:
Application has a global event `RootMouseEvent`.   This PR adds a corresponding `RootKeyEvent` for keypress events.

### DIFF
--- a/Terminal.Gui/Core/Application.cs
+++ b/Terminal.Gui/Core/Application.cs
@@ -403,6 +403,9 @@ namespace Terminal.Gui {
 
 		static void ProcessKeyEvent (KeyEvent ke)
 		{
+			if(RootKeyEvent?.Invoke(ke) ?? false) {
+				return;
+			}
 
 			var chain = toplevels.ToList ();
 			foreach (var topLevel in chain) {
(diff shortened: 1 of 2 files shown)


BDisp:
You have to swap the arguments because of the warnings.

```cs
		[Fact]
		[AutoInitShutdown]
		public void Test_RootKeyEvent_Cancel ()
		{
			Application.RootKeyEvent += SuppressKey;

			var tf = new TextField ();

			Application.Top.Add (tf);

			Application.Driver.SendKeys ('a', ConsoleKey.A, false, false, false);
			Assert.Equal ("a", tf.Text.ToString ());

			// SuppressKey suppresses the 'j' key
			Application.Driver.SendKeys ('j', ConsoleKey.A, false, false, false);
			Assert.Equal ("a", tf.Text.ToString ());

			Application.RootKeyEvent -= SuppressKey;

			// Now that the delegate has been removed we can type j again
			Application.Driver.SendKeys ('j', ConsoleKey.A, false, false, false);
			Assert.Equal ("aj", tf.Text.ToString ());
		}

		private bool SuppressKey (KeyEvent arg)
		{
			if (arg.KeyValue == 'j')
				return true;

			return false;
		}

```

tznind:
Thanks, fixed.

Terminal.Gui/Core/Application.cs
```diff
@@ -588,6 +591,15 @@ public static void UngrabMouse ()
 		/// </summary>
 		public static Action<MouseEvent> RootMouseEvent;
 
+		/// <summary>
+		/// <para>
+		/// Called for new KeyPress events before any processing is performed or
+		/// views evaluate.  Use for global key handling and/or debugging.
```
tig, line 597: Thank you for describing how it can be used here!

tig (approved):
LGTM!

PR #1639, opened 2022-03-13
Fixed TableView when redrawing a System.DataTable that has no columns

tznind:
A table with no columns is now treated like no table at all (i.e. when `Table` is null).  This fixes a 'Sequence Contains no elements' Exception that would otherwise be thrown when redrawing (see unit test without this fix).


tig (approved):
Nice catch and THANK YOU FOR THE UNIT TESTS!!!!

Issue #1524, opened 2021-12-02
FileSystemWatcher not disposed


XorZy:
### Problem

In `FileDialog.cs` a new instance of FileSystemWatcher is created every time _Reload (ustring value = null)_ is called.
The previous instance is, however, not disposed properly.
```cs
internal bool Reload (ustring value = null)
{
	bool valid = false;
	try {
		dirInfo = new DirectoryInfo (value == null ? directory.ToString () : value.ToString ());
		watcher = new FileSystemWatcher (dirInfo.FullName);
```
It is therefore possible to get this exception if the method is called too often (usually more than 1024 times for Linux), which can happen very quickly. 
```
System.IO.IOException' in System.IO.FileSystem.Watcher.dll: 'The configured user limit (X) on the number of inotify instances has been reached, or the per-process limit on the number of open file descriptors has been reached.'
```
Once the limit is reached, the exception is thrown continually and it becomes impossible to browse directories.

### Suggested change

```cs
+watcher?.Dispose();
watcher = new FileSystemWatcher (dirInfo.FullName);
```

### Suggestion

Also, it would be wise to override Dipose(bool)
```cs
 private bool _disposedValue;
 protected override void Dispose(bool disposing)
    {
        if (!_disposedValue)
        {
            if (disposing)
            {
                watcher?.Dispose();
            }

            _disposedValue = true;
        }

        // Call base class implementation.
        base.Dispose(disposing);
    }
```

tig:
@XorZy thanks for finding this. Since you've already suggested the fix, would you be willing to submit a PR?

PR #1651, opened 2022-03-25
Dispose `FileSystemWatcher` instance before creating another in `FileDialog`

tznind:
There was a fix for #1524 which disposed of the `FileSystemWatcher` in the Dispose method of the view `FileDialog`.  However the issue remains because a new `FileSystemWatcher` is allocated every time `internal bool Reload (ustring value = null)` is called.  This means that while browsing around folders there are new instances constantly being allocated and only the last one is tracked.

This PR ensures that before allocating a new `FileSystemWatcher` the old one (if any) is disposed.


tznind:
For me on linux this bug causes the dialog to freeze when opening directories where there is a lot of the file system access going on e.g. when Visual Studio Code has the directory open.  I can't browse more than 1/2 subfolders before it freezes.

tig (approved):

PR #1664, opened 2022-04-16
Japanese localization

StefanGreve:
Note: I can't use the `_` to highlight the key binding since the words are in Japanese which use a different character set. By convention, this indication is hinted at by placing the character in brackets on the right side of the word, e.g.

```
コピー (C)
```


BDisp (approved):

BDisp:
Thanks for your contribution. Due to your PR I found a bug in the `MenuItem` `Width` calculation (see my issue #1665 and my PR #1666).

tig:
Thank you for this!

Issue #1661, opened 2022-04-11
Missing Japanese Language Support


StefanGreve:
**Describe the bug**
The TUI views are completely broken if the text for labels etc. are in Japanese. It's still possible to use the controls (such as buttons), so this "only" affects the rendering in the terminal. To be clear: In my development environment, Japanese characters are still displayed correctly in the console. Its the border drawing and view alignment that's broken (under these specific circumstances).

**To Reproduce**

To display Japanese characters in the terminal on a machine using a English culture, I added these lines to my PowerShell profile:

```powershell
chcp 932 | Out-Null
$PSDefaultParameterValues['*:Encoding'] = 'utf8'
```

I use Windows Terminal and Microsoft's [Cascadia Mono font](https://github.com/microsoft/cascadia-code) which supports rendering Japanese characters.

**Expected behavior**
I didn't really expect this to work seeing how tricky Japanese font faces can be in a terminal-based environment, so this would be a nice-to-have feature and a heads-up for everyone looking to write a TUI with Japanese localization support.

**Screenshots**
Login view with slight modifications from the reactive sample code and Japanese text. 

<img width="574" alt="image" src="https://user-images.githubusercontent.com/32100623/162828650-2fd57e05-0310-4042-83c1-ef80a73a93c5.png">

**Desktop (please complete the following information):**
 - OS: Windows 
 - Version: 10.0.19044

**Additional context**
As mentioned previously, the code to reproduce this error can be obtained by replacing the label content with Japanese text using the the `ReactiveUI` demo project. This problem potentially also affects all other major Asian Languages (CJK), would be interesting to know if something similar also happens with other difficult languages such as Arabic or Hebrew.

StefanGreve:
Wow thanks a bunch, guess I grew a little to dependent on Windows Terminal that it didn't cross my mind to try a different terminal emulator. Hope your PR goes through!

StefanGreve:
This might also be interesting to other Japanese developers:

The output looks good in Developer PowerShell for VS 2022 if you add

```csharp
public static void ChangeCulture()
{
    Thread.CurrentThread.CurrentUICulture = new CultureInfo(Configuration.Language);
    Thread.CurrentThread.CurrentCulture = CultureInfo.CreateSpecificCulture(Configuration.Language);

    // You only need this line in .NET Core; install System.Text.CodePages from the Nuget Manager
    Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);

[...]

PR #1666, opened 2022-04-16
Fixes the MenuIem Width for wider runes

### LINKED ISSUE #1665: The MenuItem Width is not returning the right value
To calculate wider runes we need to use the `Rune.ColumnWidth` method. The `RuneCount` only return, as the name indicate, the number of runes.

imagem

BDisp:
Fixes #1665

imagem

### DIFF
--- a/Terminal.Gui/Views/Menu.cs
+++ b/Terminal.Gui/Views/Menu.cs
@@ -135,7 +135,7 @@ namespace Terminal.Gui {
 			return CanExecute == null ? true : CanExecute ();
 		}
 
-		internal int Width => Title.RuneCount + Help.RuneCount + 1 + 2 +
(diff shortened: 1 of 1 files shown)


StefanGreve:
Good work! But the result as depicted on the image is still not quite right. In fact, it's such a common mistake with regards to Japanese localizations, that there even is a dedicated [website](https://heistak.github.io/your-code-displays-japanese-wrong/) about it 😄

Here's your screenshot where I highlighted the differences (actual and expected characters):

image

As mentioned in issue #1661, you have to define the encoding explicitly. Now this raises the question if this burden should be placed on the library or the user to ensure that the correct encoding is used in the application.

BDisp:
Thanks for the information. I really did not defined the encoding and also I used the `NSimSun` (Chinese) font type.
I think the user must have to ensure the correct encoding.

PR #1668, opened 2022-04-17
Add ColorPicker Control.

jocelynnatali:
I purpose a ColorPicker Control.
image

You can set or get the color with SelectedColor.
`foregroundColorPicker = new ColorPicker (1, 1, "Foreground color");`
`foregroundColorPicker.SelectedColor = Color.White;`

You can set or get the color with the cursor position.
`foregroundColorPicker.SelectedColor = new Point(2, 1);`

An event is fired when the color is changer.
`foregroundColorPicker.ColorChanged += ForegroundColor_ColorChanged;`

The control works with keyboard and mouse. It manage 16 colors.

I added a scenario to try out.
image

Enjoy :-)

Here the discussion about this control
https://github.com/migueldeicaza/gui.cs/discussions/1663


BDisp:
Very cool thanks. I only suggest that you select the fore/back colors at loading by reading the superview colors.

jocelynnatali:
If you mean the default value for the SelectedColor property, I think 0 is better because this color is not necessarily linked to the component himself, but if you mean another value maybe a didn't understood 🙂

BDisp:
> If you mean the default value for the SelectedColor property, I think 0 is better because this color is not necessarily linked to the component himself, but if you mean another value maybe a didn't understood 🙂

I mean get the color of the label container, `lablel.Superview.ColorScheme`, get the `Base` color and set the `SelectedColor` with that values.

jocelynnatali:
Done.

BDisp:
> Done.

You did a force push that didn't bring no changes. Something went wrong.

jocelynnatali:
I rebased from the main branch and push forced without modifications. If it's blocking I can do a new branch and PR.

BDisp:
> I rebased from the main branch and push forced without modifications. If it's blocking I can do a new branch and PR.

Ah you did no modifications. So the pending requests for Keybinding and initial SelectedColor were still no submitted. You don't need to rebase because there is no conflicts.
If you need help about Keybinding see the ListView to help the implementation.
On the view constructor:
https://github.com/migueldeicaza/gui.cs/blob/c0bd0e0a4a341689de70ce24c98623d8a6beaa93/Terminal.Gui/Views/ListView.cs#L321-L351

and on the ProcessKey:
https://github.com/migueldeicaza/gui.cs/blob/c0bd0e0a4a341689de70ce24c98623d8a6beaa93/Terminal.Gui/Views/ListView.cs#L413-L423

Terminal.Gui/Views/ColorPicker.cs
```diff
@@ -0,0 +1,194 @@
+﻿//
+// ColorPicker.cs: ColorPicker control
+//
+// Authors:
```
tig, line 4: Please remove the Authors section. We use github to track who contributed what and prefer to not have the information duplicated within the source.

tig (changes requested):
This PR looks great, but it's not clear to me if it's complete or not. If it is not complete, please mark it as a draft. 

Also, please double-check to ensure it follows the formatting rules by doing an CTRL-K-D in VS.

UICatalog/Scenarios/ColorPicker.cs
```diff
@@ -0,0 +1,100 @@
+﻿using Terminal.Gui;
+using System;
+
+namespace UICatalog.Scenarios {
+	[ScenarioMetadata (Name: "Color Pickers", Description: "ColorPicker demo.")]
```
tig, line 5: Since the name of the control is "Color Picker", the Name of the scenario should be "Color Picker Demo" not "Color Pickers".

jocelynnatali:
Thanks for all yours comments. I thinks fixes everythings. Do you see other things to improve ?

Terminal.Gui/Views/ColorPicker.cs
jocelynnatali: Done

UICatalog/Scenarios/ColorPicker.cs
jocelynnatali: Done

BDisp (approved):

BDisp:
> Thanks for all yours comments. I think I fixed everything. Do you see other things to improve ?

It seems all right, thanks. I think you may mark as ready.

tig (changes requested):
Nice work!

I suspect users will often want this implemented as a dialog.

How hard would it be for you to add a sample showing it in a dialog to the `ColorPicker.cs` UICat Scenario?

And a question: Someday soon we'll support True Color. Can you articulate how this control will (or won't) scale when that happens? It's ok to just say "I have no idea" by the way.

See: https://github.com/migueldeicaza/gui.cs/issues/48

tig (changes requested):
One other thing: Please add to this PR an edit to `./README.md` that adds this control to the list of views. You'll also need to regen the docs (see `./docfx/README.md`).

Terminal.Gui/Views/ColorPicker.cs
```diff
@@ -0,0 +1,194 @@
...
+						Driver.AddRune (' ');
+					}
+
+					index++;
+				}
+			}
+		}
+
+		///<inheritdoc/>
+		public override bool ProcessKey (KeyEvent kb)
```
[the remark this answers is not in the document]
jocelynnatali: Yes, it was a good thing to do.

Terminal.Gui/Views/ColorPicker.cs
```diff
@@ -0,0 +1,238 @@
...
+
+			set {
+				selectColorIndex = (int)value;
+				ColorChanged?.Invoke ();
+				SetNeedsDisplay ();
+			}
+		}
+
+		/// <summary>
+		/// Initializes a new instance of <see cref="CheckBox"/> based on the given text, using <see cref="LayoutStyle.Computed"/> layout.
```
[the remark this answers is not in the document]
jocelynnatali: Yes copy/paste :-). I updated all the comments. Thanks.

jocelynnatali:
> Nice work!
> 
> I suspect users will often want this implemented as a dialog.
> 
> How hard would it be for you to add a sample showing it in a dialog to the `ColorPicker.cs` UICat Scenario?
> 
> And a question: Someday soon we'll support True Color. Can you articulate how this control will (or won't) scale when that happens? It's ok to just say "I have no idea" by the way.
> 
> See: #48

Yes, it's a good idea. I will try to add a ColorPicker Dialog in the next days. Maybe I could do the a Terminal.Gui Window like OpenDialog and the associated scenario.

I don't know yet how to design the true color ColorPicker but I thinking about that. I will propose something in the next days.

jocelynnatali:
> One other thing: Please add to this PR an edit to `./README.md` that adds this control to the list of views. You'll also need to regen the docs (see `./docfx/README.md`).

Ahhh yes. The README.md and the docs are updated. I used the last stable version of docfx. Is it right ?

jocelynnatali:
I think we can finish this PR and I'll do another one for the ColorPickerDialog and the associated scenario. What do you think about that ?

tig:
> I think we can finish this PR and I'll do another one for the ColorPickerDialog and the associated scenario. What do you think about that ?

I concur.

Terminal.Gui/Views/ColorPicker.cs
```diff
@@ -0,0 +1,234 @@
...
+				selectColorIndex = (int)value;
+				ColorChanged?.Invoke ();
+				SetNeedsDisplay ();
+			}
+		}
+
+		/// <summary>
+		/// Initializes a new instance of <see cref="ColorPicker"/>.
+		/// </summary>
+		public ColorPicker () : this (0, 0, "Color Picker")
```
tig, line 76: Please add a constructor that takes just a Title (you'll see why as I add more comments).

UICatalog/Scenarios/ColorPicker.cs
```diff
@@ -0,0 +1,100 @@
...
+			// Forground ColorPicker.
+			foregroundColorPicker = new ColorPicker (1, 1, "Foreground color");
+			foregroundColorPicker.ColorChanged += ForegroundColor_ColorChanged;
+			Win.Add (foregroundColorPicker);
+
+			foregroundColorLabel = new Label (1, 10, string.Empty);
+			Win.Add (foregroundColorLabel);
+
+			// Background ColorPicker.
+			backgroundColorPicker = new ColorPicker (85, 1, "Background color");
```
tig, line 50: I noticed that if my terminal is not default width, the controls in the Scenario get clipped etc...

I guessed you had not tested using Relative Layout.

I was right. The control is completely broken when using Relative Layout.

Please:

1) Change the Scenario to demonstrate that both Absolute and Relative Layout work.
2) Fix whatever is wrong that prevents #1 from currently working ;-).

Here's code that SHOULD align the color picker to the right of the window:

```csharp
// Background ColorPicker.
backgroundColorPicker = new ColorPicker ();
backgroundColorPicker.Text = "Background Color Pickeroo";
backgroundColorPicker.X = Pos.AnchorEnd () - (Pos.Right (backgroundColorPicker) - Pos.Left (backgroundColorPicker));
backgroundColorPicker.Y = 1;
backgroundColorPicker.ColorChanged += BackgroundColor_ColorChanged;
Win.Add (backgroundColorPicker);
```

tig (changes requested):
Relative positioning is not working.

Support parameterless constructor (implies Relative Layout).

Support a constructor that just takes Title (implies Relative Layout).

tig:
FWIW, I just added this wiki: https://github.com/migueldeicaza/gui.cs/wiki/Checklist-for-creating-new-Views-and-Controls

Terminal.Gui/Views/ColorPicker.cs
jocelynnatali: Yes, something was wrong. Thanks. I added this constructor.

UICatalog/Scenarios/ColorPicker.cs
jocelynnatali: The component manager now the relative layout, but I have a problem this borders. If I set a border, the size of the component doesn't count the border. Is it normal ? So the relative layout not working well.
Can you explain me, please what I have to do with high-intensity colors ?

BDisp: Yes it's normal.. If you want to use borders on your view then you must have two options. With the `PanelView` you have the possibility of having your view growing with the borders or maintaining always the same size by shrinking inside, dependent on the `UsePanelFrame` property. Without the `PanelView` your view can only be growing within the borders, unless it's implemented as `Window` or `FrameView`. If you want to measure the total borders size you can use the `GetSumThickness` method of the `Border` class. Please see the `Borders` scenario to learn about the both implementations and differences.
With the high-intensity colors I have no idea yet.

tig: > Can you explain me, please what I have to do with high-intensity colors ?

I had typed `#1` and github expanded that to Issue # 1, which is not what I meant.  I was referring to my 1st point which was `Change the Scenario to demonstrate that both Absolute and Relative Layout work.`.  My apologies for not noticing github had expanded `#1` like this and confusing you!

I see you've already updated the scenario to use relative positioning, so that part is good!

jocelynnatali: > Yes it's normal.. If you want to use borders on your view then you must have two options. With the `PanelView` you have the possibility of having your view growing with the borders or maintaining always the same size by shrinking inside, dependent on the `UsePanelFrame` property. Without the `PanelView` your view can only be growing within the borders, unless it's implemented as `Window` or `FrameView`. If you want to measure the total borders size you can use the `GetSumThickness` method of the `Border` class. Please see the `Borders` scenario to learn about the both implementations and differences. With the high-intensity colors I have no idea yet.

Thank you for explaining. At the end I think it's better without borders by default and the user can choose to add or not depending if he wants use the control in a (Dialog, PanelView) or not. In fact, the relative layout works well without borders. With border maybe it's normal to taking the borders width in account when positioning the control in the page. Is it right ? I saw the same behaviour with other controls.

jocelynnatali: > > Can you explain me, please what I have to do with high-intensity colors ?
> 
> I had typed `#1` and github expanded that to Issue # 1, which is not what I meant. I was referring to my 1st point which was `Change the Scenario to demonstrate that both Absolute and Relative Layout work.`. My apologies for not noticing github had expanded `#1` like this and confusing you!
> 
> I see you've already updated the scenario to use relative positioning, so that part is good!

Ahhh ok. No problem :-)

jocelynnatali:
The tests are working on my computer but fails on the build. Did I miss something in the control events or initialisation ?

BDisp:
> The tests are working on my computer but fails on the build. Did I miss something in the control events or initialisation ?

Do not worry about this error because it is in `Run_All_Scenarios()` test that lately is always failing. I think the process will be busy in another process that causes a certain delay in its execution.

tig:
> > The tests are working on my computer but fails on the build. Did I miss something in the control events or initialisation ?
> 
> Do not worry about this error because it is in `Run_All_Scenarios()` test that lately is always failing. I think the process will be busy in another process that causes a certain delay in its execution.

See #1712

jocelynnatali:
Thank you. Except the tests. Do you see other things to do on my PR ?

BDisp:
Maybe this can be marked as ready, I think.

tig (approved):
Thanks a ton for this great addition!

jocelynnatali:
Coool :-) Thank you all three for your help.

jocelynnatali:
Ah ok. Yes, I see. Good idea. Thanks. I do the changes.

Télécharger Outlook pour Android<https://aka.ms/AAb9ysg>

PR #1677, opened 2022-04-20
Fixed erroneous namespace on LineView to match other views

### LINKED ISSUE #1653: LineView namespace
**Describe the bug**
Not sure if this is a 'bug' exactly but more of an oversight.

Currently the `LineView` class is in the namespace `Terminal.Gui.Views`.  It is the only thing in that namespace.  Its annoying and untidy but fixing it would be breaking for anyone that is using it (they would need to remove the `using Terminal.Gui.Views` directive since it no longer would exist.

Just raising this issue for awareness rather than saying we definitely should fix it.  Also go get a feel for other peoples opinions on this (whether to fix).

tznind:
Fixes  #1653

### DIFF
--- a/Terminal.Gui/Views/LineView.cs
+++ b/Terminal.Gui/Views/LineView.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using Terminal.Gui.Graphs;
 
-namespace Terminal.Gui.Views {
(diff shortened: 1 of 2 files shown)


tig (approved):
Thanks for taking care of this.

PR #1684, opened 2022-04-29
Add timeout thread safety

tznind:
Fixes  #1679

I think this should fix the above issue.  There are 2 problems addressed by this PR.  

There was locking on the `timeouts` field (i.e. `lock (timeouts)`).  But since this variable is reassigned to a new list in `RunTimers` the reference changes so locking will not be consistent.  I have fixed this by adding an unchanging object `timeoutsLockToken` and locking on that instead.

RunTimers iterates through a `copy` of `timeouts` but there was no protection of the period in which you take the copy but haven't yet assigned a new list for callers to AddTimeout to add new timeouts to which could result in various errors.  I've added an extra lock around that section.  Initially I put it around the whole `RunTimers` but I think that is excessive and the solution implemented works.

I've added a test which is quite conservative (100 threads x 100 timeout callbacks of 1 second x 5 seconds runtime) which shouldn't stress any machines running it much.

Issue #1679, opened 2022-04-26
MainLoop.RunTimers() method throws System.ArgumentException after few hours.


kacperpikacz:
Can someone explain me why I'm getting this error after around 8 hours of running? 
Terminal.Gui version 1.4

`AmiAI - System.ArgumentException: An item with the same key has already been added. Key: 637865266986268040 (Parameter "key")`
                                                                                                                                at System.Collections.Generic.SortedList"2.Add(TKey key, TValue value)
                at Terminal.Gui.MainLoop.RunTimers()
                                                       at Terminal.Gui.MainLoop.MainIteration()
                                                                                                  at Terminal.Gui.Application.RunLoop(RunState state, Boolean wait)
                                                                                                                                                                      at Terminal.Gui.Application.Run(Toplevel view, Func"2 errorHandler)
                                                   at Terminal.Gui.Application.Run(Func"2 errorHandler)
                                                                                                          at Pai.Ama.Run()
                                                                                                                                at Pai.Program.Main(String[] args)

tig:
We probably can't help debug this without knowing what your code does.

tznind:
Hi, do you get this error consistently (after long periods of execution) or was this a one off? What operating system are you using? 

One thing I can advise if you are using a lot of timeouts is to add them using `Application.MainLoop.Invoke` e.g. 

```csharp
Application.MainLoop.Invoke(()=> {
		Application.MainLoop.AddTimeout (TimeSpan.FromSeconds(1), (s) => { /*your method here*/ });
});
```

I was stress testing the timeout system with the following multi threaded 'test' (warning: will eat your CPU and RAM if you run it).

```csharp
[Fact,AutoInitShutdown]
public void TestAddManyTimeouts()
{
	for(int i=0;i<1000;i++) {
		Task.Run (() => {
			Task.Delay (100).Wait();
			while (true) {
				
				Application.MainLoop.Invoke(()=> {
					Application.MainLoop.AddTimeout (TimeSpan.Zero, (s) => { return true; });
				});
			}
		});
	}
	
	Application.MainLoop.Run ();
}
```

The test runs fine with the MainLoopInvoke bit but if you remove it then you get a veriety of failures including:

```
  Message: 
    System.ArgumentException : An item with the same key has already been added. Key: 637865486955401012 (Parameter 'key')

  Stack Trace: 
    SortedList`2.Add(TKey key, TValue value)
[...]

PR #1692, opened 2022-05-04
Fixed RunTimers not checking for key collisions in list when adding to timeouts

tznind:
So we got the race condition sorted in #1684 but there is another issue with `AddTimeout`.  

There is code in `void AddTimeout (TimeSpan time, Timeout timeout)` which looks for key collisions in the `timeouts` dictionary:

https://github.com/migueldeicaza/gui.cs/blob/2ab69a16d3f4ef082c8e822981f7b01e2d35a9e3/Terminal.Gui/Core/MainLoop.cs#L121-L127

 However there is no check for key collisions in `RunTimers` which has this code only below.

https://github.com/migueldeicaza/gui.cs/blob/2ab69a16d3f4ef082c8e822981f7b01e2d35a9e3/Terminal.Gui/Core/MainLoop.cs#L193-L195

If a new timeout is added at time k by AddTimeout at the same time that RunTimers is iterating and the AddTimeout sneaks in first then RunTimers will fail with 'key already exists' error.  This I think is the issue that @kacperpikacz is getting in #1679.

I have added a new method `NudgeToUniqueKey` which performs this check and should ensure no two timeouts are set for the same time.  

Given it takes @kacperpikacz  ~8 hours of runtime for this error to manifest I don't think there is an easy way to unit test the fix.

### DIFF
--- a/Terminal.Gui/Core/MainLoop.cs
+++ b/Terminal.Gui/Core/MainLoop.cs
@@ -120,10 +120,7 @@ namespace Terminal.Gui {
 		{
 			lock (timeoutsLockToken) {
 				var k = (DateTime.UtcNow + time).Ticks;
-				while (timeouts.ContainsKey (k)) {
-					k = (DateTime.UtcNow + time).Ticks;
-				}
-				timeouts.Add (k, timeout);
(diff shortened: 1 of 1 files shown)


tig (approved):
Nicely done!

PR #1703, opened 2022-05-08
Adds support for custom colors in TreeView

### LINKED ISSUE #1702: TreeView and ListView: Per Node ColorTheme
I want to use custom Colors per TreeNode.

https://github.com/migueldeicaza/gui.cs/blob/main/Terminal.Gui/Views/TreeView.cs#L456
I'd like to specify the ColorScheme per TreeNode.

Because all required members in Tree.Redraw are private and, Branch<T> is internal and there's no way to specify a custom Branch<T> type like a CreateBranchInstance, it's impossible to use a custom ColorTheme per Node.

**Additional context**
I have a Command-Runner, and need a (compact) way to differentiate between a Command-Node and a Folder-Node without wasting space for a prefix text or something else.

It would be great if this is possible for ListView as well.

tznind:
Fixes #1702

### DIFF
--- a/Terminal.Gui/Core/Trees/Branch.cs
+++ b/Terminal.Gui/Core/Trees/Branch.cs
@@ -159,9 +159,27 @@ namespace Terminal.Gui.Trees {
 				availableWidth -= lineBody.Length;
 			}
 
-			//reset the line color if it was changed for rendering expansion symbol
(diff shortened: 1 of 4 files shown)


tig (approved):
Great stuff! Way to listen to our customers!

PR #1707, opened 2022-05-10
Avoid TextView throwing when WordWrap is true on constructor initialization.

BDisp:


tig (approved):
Excellent attention to detail. Thank you.

PR #1708, opened 2022-05-10
Avoid UpdateScreen running twice on Application.Refresh because Driver.Refresh already do.

BDisp:
On `Application.Refresh` is needed to clear the buffer and that is what `UpdateOffScreen` does.


tig (approved):
LGTM

PR #1710, opened 2022-05-10
Fixes remaining wide runes render issues.

### LINKED ISSUE #1709: Wide runes still have some issues on render.
wide-runes-issue
wide-runes-menu-issue

BDisp:
Fixes #1709.

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -70,7 +70,9 @@ namespace Terminal.Gui {
 					contents [crow, ccol - 1, 0] = (int)(uint)' ';
 					Curses.move (crow, ccol);
 					Curses.attrset (curAtttib);
-				} else if (runeWidth < 2 && ccol < Cols - 1 && Rune.ColumnWidth ((char)contents [crow, ccol, 0]) > 1) {
+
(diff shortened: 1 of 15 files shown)


tig:
> @tig, @tznind I propose to remove the `AlwaysSetPosition` from the `NetDriver` since I think nobody use it and it don't make no sence. It was created to prevent the cursor from exceeding the total columns on the right side in wide runes, but has the disadvantage of overlapping wide runes, printing only half of the rune. With this my PR I think this will no longer be problematic. Then say your opinion, please.

Since `NetDriver.AlwaysSetPosition` is a `public` it is technically an API and thus this would technically be a breaking change. 

However, if I read this correctly, `NetDriver.AlwaysSetPosition` is ONLY defined on `NetDriver` and not the `ConsoleDriver` base class. It is used within `Application`, but scoped via `if (Driver is NetDriver)`. Thus it's NOT an API.

So, IMO, deleting `NetDriver.AlwaysSetPosition` is NOT a breaking change. 

LIkewise, `Application.AlwaysSetPosition` is `public`, so deleting it is technically a breaking change. But the documentation comment clearly says (at least to me) that this property is only used by `NetDriver`. 

So, I don't think removing `Application.AlwaysSetPosition` is a breaking change.

However, I don't see you deleting it in this PR.  Will you do another PR to remove it or add it to this one?

BDisp:
Thanks. I can remove in this pr, if you don't mind.

tig (approved):
Great!

PR #1725, opened 2022-05-15
Fixes nullref on unclean shutdown

WouterTinus:
First of all thanks for this lovely framework! I've been playing around with it for a couple of days now and I'm seriously considering picking it up for my project.

I ran into crashes in the WindowsDriver on my system while playing around with both my own project and the UICatalog example. It only happens when I forcibly close the terminal window. Because the crash happens while the process is being killed, the executable remains locked by Visual Studio, making it difficult to rebuild again after.

The debugger catches the exception, but it gets killed itself soon after. Luckily it was easy enough to see where it went wrong. I'm guessing that the `ReadConsoleInput` function takes a `ThreadAbortException` or something similar and hence returns `null`. But the `WindowsInputHandler` doesn't do a null check and simply enqueues it. Then in another (still running) thread the `null` is dequeued by `MainIteration` and attempts to access the 1st item, causing a `NullReferenceException`.

I've proposed a patch that not only checks for possible `null` but also verifies that length of the array is at least 1. The latter check might be superfluous but closes another possible crash vector. I've also considering doing the null check at the enqueue side of things, but I didn't see any immediate benefit or downside from doing it in either place, so I kept it at the consuming side so that it's also safe from other future producers.

- Visual Studio 2022 (v17.2)
- Windows 11 (v22000.675)
- Windows Terminal (v1.12.10983.0)

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
@@ -1892,8 +1892,11 @@ namespace Terminal.Gui {
 		void IMainLoopDriver.MainIteration ()
 		{
 			while (resultQueue.Count > 0) {
-				var inputEvent = resultQueue.Dequeue () [0];
-				ProcessInput?.Invoke (inputEvent);
+				var inputRecords = resultQueue.Dequeue ();
+				if (inputRecords != null && inputRecords.Length > 0) {
+					var inputEvent = inputRecords [0];
+					ProcessInput?.Invoke (inputEvent);
+				}
 			}
 			if (winChanged) {
 				winChanged = false;


BDisp:
Good catch, thank. Perhaps other drivers may also have this issue.
Since you use the same configuration I have, I would like to ask you if you have problems with attaching the `Terminal.Gui` running on `WSL` to the `Visual Studio 2022`, please? This issue started after updating the VS2022 to the 17.2.0 version.
Running `WSL` from VS2022 it run without errors, despite not opening with a GUI terminal.
wsl
output

By attach from running on `WSL` to the VS2022, I get this error:
`The program '[836] dotnet' has exited with code 4294967295 (0xffffffff).`
running-on-wsl
attaching
error-dialog
output-debug-error

WouterTinus:
Hi @BDisp, I tried to test it from Ubuntu 20.04.3 LTS in WSL. I started the application with the command `dotnet run --project UICatalog.csproj --` And then from Visual Studio it looked like this:

image

I picked the `UICatalog` one and the debugger nicely attaches (breakpoints, watch etc. working fine). Only extra thing I had to do was to install the `.NET Debugging with WSL` component from the Visual Studio Installer. Perhaps it's worth a shot to uninstall/reinstall that one?

tig (approved):
Nice find and nice tight fix. 

Thank you!

BDisp:
> Hi @BDisp, I tried to test it from Ubuntu 20.04.3 LTS in WSL. I started the application with the command `dotnet run --project UICatalog.csproj --` And then from Visual Studio it looked like this:

Sorry for not tell how I do it. Normally I compile in the VS2022 and in `WSL` on the folder where is the repo I run:
`dotnet ./UICatalog/bin/Debug/net6.0/UICatalog.dll`


> I picked the `UICatalog` one and the debugger nicely attaches (breakpoints, watch etc. working fine). Only extra thing I had to do was to install the `.NET Debugging with WSL` component from the Visual Studio Installer. Perhaps it's worth a shot to uninstall/reinstall that one?

I already have that installed and I'm using it a very long time.


> Edit: I also installed the latest .NET SDK (6.0.300) in WSL, not just the runtime. Didn't try with runtime only though. But that might be another factor I suppose.

I also have the NET SDK (6.0.300).
Installing the SDK also install the runtime.

Thanks for your help.

BDisp:
I already fixed the wsl issue by uninstalling wsl and the distros. After the reinstall all work well now.
I went into this situation after trying to set debugger for c++ following this steps https://docs.microsoft.com/pt-pt/cpp/linux/connect-to-your-remote-linux-computer?view=msvc-170&viewFallbackFrom=vs-2019.

PR #1728, opened 2022-05-17
Fixes GetMaxLengthforWidth on Button.

BDisp:
Without the fix the unit `GetMaxLengthForWidth_On_Button` test will fail.


tig (approved):
LGTM

PR #1734, opened 2022-05-18
Reverts Clear method to use AddRune instead of AddStr.

BDisp:
When I changed this I made some confusion that `AddRune` wasn't moving the column, but is the place where it's moving really, sorry. Perhaps because in the `NetDriver` it wasn't moving if it's outside of the clip bounds, but that was also fixed.


tig (approved):
LGTM

Issue #1748, opened 2022-05-27
TableView long cells get pushed of screen requiring horizontal scrolling


tznind:
**Describe the bug**
Currently TableView wants to display full columns only.  It respects MaxWidth of columns (i.e. it won't make a column longer than that) and it has MinWidth which prevents columns being rendered too small.

When deciding whether or not to show a column it looks at the size of the content and sees if it will all fit on the screen (up to the MaxWidth).  This means you can get a lot of empty space in the table as a long column is pushed off the screen because it doesn't quite fit.

![tableviewscrolling](https://user-images.githubusercontent.com/31306100/170651274-f821db3f-c408-4f8a-a65a-5ad99c63aa33.gif)

**To Reproduce**
1. Create a table with 3 columns
2. Put short strings in first 2 columns
3. Put very long strings in last column
4. Scroll around

Also I think there needs to be a scroll arrow indicator (like with tab view) that shows there are more columns right/left

tig:
@tznind, #1756 didn't have "Fixes #" in the body and thus this didn't get closed automatically.

Can you confirm it can be closed?

tznind:
> @tznind, #1756 didn't have "Fixes #" in the body and thus this didn't get closed automatically.
> 
> Can you confirm it can be closed?

I'd like to look a bit more into the problem of long columns if that's ok.  The PR adds indicators that show when columns are pushed off (due to the console size) but I think we can do more to help this use case (long columns).  

Current API controls that tell TableView about column layout are:

Firstly measure the header text (if visible) and visible rows to determine content width.  This is the size required to render table (if theres not enough it gets pushed off screen).  Then API user can also set:

- MaxWidth (don't make column bigger than this)
- MinWidth (don't make column smaller than this)
- ExpandLastColumn (allow last column in table to go bigger than MaxWidth to fill render space) 

But for long columns you might want to say:
_Make this column fill as much space as possible (bounded by Max/Min)_

That would let long free text columns be more flexible than the current absolute values allow.'  

I might just start with a deep dive write-up of the layout logic.  More docs is always helpful.

tig:
FWIW, this work is really important: 

I plan on enhancing `Out-ConsoleGridView` to use `TableView`...  ;-)

PR #1756, opened 2022-05-28
Add horizontal scroll idicators to TableView

tznind:
This PR helps with #1748 by at least indicating to the user when there are more column(s) they can't see (because they don't fit on the screen)

Requires `tableView.Style.ShowHorizontalHeaderUnderline = true;` (the default)

tableviewscrollindicators

### DIFF
--- a/UnitTests/TableViewTests.cs
+++ b/UnitTests/TableViewTests.cs
@@ -778,6 +778,85 @@ namespace Terminal.Gui.Views {
 			Application.Shutdown ();
 		}
 
+
+		[Fact]
+		public void ScrollIndicators ()
+		{
+			GraphViewTests.InitFakeDriver ();
+
+			var tableView = new TableView ();
+			tableView.ColorScheme = Colors.TopLevel;
+
+			// 3 columns are visibile
+			tableView.Bounds = new Rect (0, 0, 7, 5);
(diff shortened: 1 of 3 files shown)


UnitTests/TableViewTests.cs
```diff
@@ -778,6 +778,85 @@ public void ScrollRight_WithoutSmoothScrolling ()
...
+			// select last visible column
+			tableView.SelectedColumn = 2; // column C
+
+			tableView.Redraw (tableView.Bounds);
+
+			// user can only scroll right so sees right indicator
+			// Because first column in table is A
+			string expected = 
+				@"
+│A│B│C│
```
tig, line 818: Just curious: How exhaustive are the unit tests for tableview with wide runes?

tig (approved):
LGTM

UnitTests/TableViewTests.cs
tznind: I don't think there are any tests covering it.  But `CalculateMaxCellWidth` does use `col.ColumnName.Sum (c => Rune.ColumnWidth (c));` so it should work.  

I am happy to write some though.  Always good to have more tests.

tznind: @tig  I have added a PR #1761 that should ensure any unicode issues are easily diagnosable

PR #1761, opened 2022-05-30
Added unicode testing code to TableEditor

### LINKED ISSUE #152: how to use listview item selection
hi i have this ui i want user select something in listview and item write in textfield how i can do this?
```
 static void Main(string[] args)
        {
            Application.Init();
            var top = Application.Top;

            // Creates the top-level window to show
            var win = new Window(new Rect(0, 1, top.Frame.Width, top.Frame.Height - 1), "UrlShortener v" + Assembly.GetExecutingAssembly().GetName().Version);
            top.Add(win);

            // Add some controls
            win.Add(
               
                new Label(3, 2, "Choose Service: "),
                new ListView(new Rect(20, 2, 16, 4), new string[]
                {
                    "Opizo",
                    "Bitly",
                }),
                new TextField(20, 6, 40, ""),
                new Button(3, 14, "Ok"),
                new Button(10, 14, "Cancel"),
            Application.Run();
        }

```

tznind:
Fixes #152

This PR adds a new table to TableEditor scenario that includes all unicode symbols in the ranges defined here:

https://jrgraphix.net/r/Unicode/  

I've designed it so its easy to add more later if there are other ranges that you want to test.  Hitting Enter shows the unicode uint value of the cell and the name of the range it is classified as.

unicodetable

### DIFF
--- a/UICatalog/Scenarios/TableEditor.cs
+++ b/UICatalog/Scenarios/TableEditor.cs
@@ -48,6 +48,7 @@ namespace UICatalog.Scenarios {
 				new MenuBarItem ("_File", new MenuItem [] {
 					new MenuItem ("_OpenBigExample", "", () => OpenExample(true)),
 					new MenuItem ("_OpenSmallExample", "", () => OpenExample(false)),
+					new MenuItem ("OpenCharacter_Map","",()=>OpenUnicodeMap()),
 					new MenuItem ("_CloseExample", "", () => CloseExample()),
 					new MenuItem ("_Quit", "", () => Quit()),
 				}),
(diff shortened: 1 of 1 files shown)


tig:
Did you happen to look at the ranges I used in Character Map?

tig:
Screwing around with this (which is an awesome stress test btw) I think I found a rendering bug.

image



Note when I use keyboard I can determine the glyph causing the issue:

image

(Can you show the code in hex in the dialog?)

Interestingly, CharMap has no issue in this range.

image

tig:
Not worth holding merging this, but check this out:



Character Map has a bit of a mis-render with those glyphs (around 0x202A, 0x202B, 0x202C) too, FWIW.

image

tig (approved):
This is awesome.

tznind:
> (Can you show the code in hex in the dialog?)

The hex code is the title of the menu box e.g. 202A.  But I agree it should be formated `0x0202A`.

Its interesting behaviour there.  I see from the [unicode docs](https://unicode-explorer.com/c/202A) that 202A is LEFT-TO-RIGHT EMBEDDING

> The "LEFT-TO-RIGHT EMBEDDING" directional formatting character is the classical Unicode method of explicit bidirectional formatting, and as of Unicode 6.3, is being discouraged in favor of [LEFT-TO-RIGHT ISOLATE](https://unicode-explorer.com/c/2066). An "embedding" signals that a piece of text is to be treated as directionally distinct. The text within the scope of the embedding formatting characters is not independent of the surrounding text. Also, characters within an embedding can affect the ordering of characters outside. **Unicode 6.3 recognized that directional embeddings usually have too strong an effect on their surroundings and are thus unnecessarily difficult to use.**

_From docs (emphasis mine)_

tznind:
> Did you happen to look at the ranges I used in Character Map?

I've added those ranges in #1762

PR #1764, opened 2022-05-31
Fixes #1763. Allowing read console inputs before idle handlers.

### LINKED ISSUE #1763: Continuous button pressed process not stopping at button released.
To test run the `CharacterMap` and press the left mouse button on the scroll arrow and the scrolling won't stop, because the task of the `ProcessContinuousButtonPressedAsync` method is adding a idle handler to the mainloop and the subsequent inputs aren't being read, as the button released.

BDisp:


tig (approved):
Good catch and fix. 

The CharMap scenario is now usable, but my rendering code is still too slow.

PR #1765, opened 2022-06-02
Fixes some scrolling performance issues

### LINKED ISSUE #1743: UICatalog: Character Map scenario scrolling is unusably slow
Using the scroll bars in Character Map is unusable because rendering is taking too long. 

Need to look into making it more usable.

BDisp:


BDisp:
> I pulled this down locally and using the mouse wheel is slightly better. But I still think CharMap renders too slowly. Compare how quickly you can scroll (with keyboard up/down or pageup/pagedown) the CharacterMap in TableEditor.

For mouse work like pageup/pagedown you must stay with the button pressed between above or bellow the scroll arrows and the bar. If you press on the scrollbar and moving it's much faster. With the wheel it scrolls 1 line at a time, the same as pressing on the scroll arrows, so very much slower for the length of the runes data.

mouse-scroll

PR #1770, opened 2022-06-04
Exposes APIs to support upcoming Web console feature

FreddieMercurial:
Our team, and especially @BDisp have been working quite hard and have produced a web console driver for Gui.cs. Our driver needs visibility into some internal variables and we've also found/fixed a couple typos and bugs.
ezgif com-gif-maker
For https://github.com/migueldeicaza/gui.cs/issues/1642

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -30,7 +30,7 @@ namespace Terminal.Gui {
 		IClipboard clipboard;
 		int [,,] contents;
 
-		internal override int [,,] Contents => contents;
+		public override int [,,] Contents => contents;
 
 		// Current row, and current col, tracked by Move/AddRune only
 		int ccol, crow;
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -491,7 +491,7 @@ namespace Terminal.Gui {
 					Flags = mouseFlag
 				};
 
-				var view = Application.wantContinuousButtonPressedView;
+				var view = Application.WantContinuousButtonPressedView;
 				if (view == null)
 					break;
 				if (isButtonPressed && lastMouseButtonPressed != null && (mouseFlag & MouseFlags.ReportMousePosition) == 0) {
--- a/Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs
@@ -35,7 +35,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Assists with testing, the format is rows, columns and 3 values on the last column: Rune, Attribute and Dirty Flag
 		/// </summary>
-		internal override int [,,] Contents => contents;
+		public override int [,,] Contents => contents;
 
 		//void UpdateOffscreen ()
 		//{
--- a/Terminal.Gui/ConsoleDrivers/NetDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/NetDriver.cs
@@ -1187,7 +1187,7 @@ namespace Terminal.Gui {
 		public NetWinVTConsole NetWinConsole { get; }
 		public bool IsWinPlatform { get; }
 		public override IClipboard Clipboard { get; }
-		internal override int [,,] Contents => contents;
+		public override int [,,] Contents => contents;
 
 		int largestWindowHeight;
 
--- a/Terminal.Gui/ConsoleDrivers/NetDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/NetDriver.cs
@@ -947,7 +947,7 @@ namespace Terminal.Gui {
 			await Task.Delay (200);
 			while (isButtonPressed) {
 				await Task.Delay (100);
-				var view = Application.wantContinuousButtonPressedView;
+				var view = Application.WantContinuousButtonPressedView;
 				if (view == null) {
 					break;
 				}
--- a/Terminal.Gui/Core/Application.cs
+++ b/Terminal.Gui/Core/Application.cs
@@ -106,6 +106,11 @@ namespace Terminal.Gui {
 		public static Toplevel Current { get; private set; }
 
 		/// <summary>
+		/// The current <see cref="View"/> object that wants continuous mouse button pressed events.
+		/// </summary>
+		public static View WantContinuousButtonPressedView { get; private set; }
+
+		/// <summary>
 		/// The current <see cref="ConsoleDriver.HeightAsBuffer"/> used in the terminal.
 		/// </summary>
 		public static bool HeightAsBuffer {
@@ -603,7 +608,6 @@ namespace Terminal.Gui {
 		/// </summary>
 		public static Func<KeyEvent, bool> RootKeyEvent;
 
-		internal static View wantContinuousButtonPressedView;
 		static View lastMouseOwnerView;
 
 		static void ProcessMouseEvent (MouseEvent me)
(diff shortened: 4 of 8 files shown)


Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -491,7 +491,7 @@ async Task ProcessContinuousButtonPressedAsync (Curses.MouseEvent cev, MouseFlag
 					Flags = mouseFlag
 				};
 
-				var view = Application.wantContinuousButtonPressedView;
+				var view = Application.WantContinuousButtonPressedView;
```
FreddieMercurial, line 494: Capitalization

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -30,7 +30,7 @@ internal class CursesDriver : ConsoleDriver {
 		IClipboard clipboard;
 		int [,,] contents;
 
-		internal override int [,,] Contents => contents;
+		public override int [,,] Contents => contents;
```
FreddieMercurial, line 33: We need access to the buffer

Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs
```diff
@@ -35,7 +35,7 @@ public class FakeDriver : ConsoleDriver {
 		/// <summary>
 		/// Assists with testing, the format is rows, columns and 3 values on the last column: Rune, Attribute and Dirty Flag
 		/// </summary>
-		internal override int [,,] Contents => contents;
+		public override int [,,] Contents => contents;
```
FreddieMercurial, line 38: Another visibility

Terminal.Gui/ConsoleDrivers/NetDriver.cs
```diff
@@ -947,7 +947,7 @@ async Task ProcessContinuousButtonPressedAsync ()
 			await Task.Delay (200);
 			while (isButtonPressed) {
 				await Task.Delay (100);
-				var view = Application.wantContinuousButtonPressedView;
+				var view = Application.WantContinuousButtonPressedView;
```
FreddieMercurial, line 950: Capitalization

Terminal.Gui/ConsoleDrivers/NetDriver.cs
```diff
@@ -1187,7 +1187,7 @@ internal class NetDriver : ConsoleDriver {
 		public NetWinVTConsole NetWinConsole { get; }
 		public bool IsWinPlatform { get; }
 		public override IClipboard Clipboard { get; }
-		internal override int [,,] Contents => contents;
+		public override int [,,] Contents => contents;
```
FreddieMercurial, line 1190: Visibility

Terminal.Gui/Core/Application.cs
```diff
@@ -385,7 +411,7 @@ protected virtual void Dispose (bool disposing)
 
 		static void ProcessKeyEvent (KeyEvent ke)
 		{
-			if(RootKeyEvent?.Invoke(ke) ?? false) {
+			if (RootKeyEvent?.Invoke (ke) ?? false) {
```
FreddieMercurial, line 414: the space after invoke may be accidental?

BDisp: No it's how .editorconfig is configured.

Terminal.Gui/Core/Application.cs
```diff
@@ -210,6 +215,24 @@ static void OnQuitKeyChanged (Key oldKey)
 		/// </summary>
 		public static bool IsMouseDisabled { get; set; }
 
+		/// <summary>
+		/// By default this is false which will continue executing the <see cref="RunLoop(RunState, bool)"/> method,
```
tig, line 219: Poor English phrasing. I think you mean:

"Set to true to cause the RunLoop method to exit after the first iterations. Set to false (the default) to cause the RunLoop to continue running until `Application.RequestStop()` is called."

Terminal.Gui/Core/Application.cs
```diff
@@ -952,51 +977,65 @@ public static void RunLoop (RunState state, bool wait = true)
...
-					}
-				} else if (!wait) {
+				if (ExitRunLoopAfterFirstIteration && !firstIteration)
 					return;
+				RunIteration (ref state, wait, ref firstIteration);
+			}
+		}
+
+		/// <summary>
+		/// Run the <see cref="RunLoop(RunState, bool)"/> iteration.
```
tig, line 987: I think the cref is potentiall confusing. Because what you really mean is "Run one iteration of the MainLoop".

Terminal.Gui/Core/Application.cs
```diff
@@ -952,51 +977,65 @@ public static void RunLoop (RunState state, bool wait = true)
...
+		/// </summary>
+		/// <param name="state">The state returned by the Begin method.</param>
+		/// <param name="wait">If will execute the runloop waiting for events.</param>
+		/// <param name="firstIteration">If it's the first run loop iteration.</param>
+		public static void RunIteration (ref RunState state, bool wait, ref bool firstIteration)
```
tig, line 992: Would it be more clear if this was named `RunLoopIteration` or `RunMainLoopIteration`?

tig (changes requested):
Very interesting stuff guys! 

See my comments.

Terminal.Gui/Core/MainLoop.cs
```diff
@@ -45,21 +45,46 @@ public interface IMainLoopDriver {
...
 		}
 
 		internal SortedList<long, Timeout> timeouts = new SortedList<long, Timeout> ();
 		object timeoutsLockToken = new object ();
 		internal List<Func<bool>> idleHandlers = new List<Func<bool>> ();
 
+		/// <summary>
+		/// Gets the list of all timeouts.
+		/// </summary>
+		public SortedList<long, Timeout> Timeouts => timeouts;
```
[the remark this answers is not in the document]
BDisp: @tig is this enough?

```cs
		/// <summary>
		/// Gets the list of all timeouts sorted by the <see cref="TimeSpan"/> time ticks./>.
		/// A shorter limit time can be added at the end, but it will be called before an
		///  earlier addition that has a longer limit.
		/// </summary>
		public SortedList<long, Timeout> Timeouts => timeouts;
```

Terminal.Gui/Core/MainLoop.cs
```diff
@@ -45,21 +45,46 @@ public interface IMainLoopDriver {
...
 		/// The current IMainLoopDriver in use.
 		/// </summary>
 		/// <value>The driver.</value>
 		public IMainLoopDriver Driver { get; }
 
+		/// <summary>
+		/// Invoked when a new timeout is added to be used on the case
+		/// if <see cref="Application.ExitRunLoopAfterFirstIteration"/> is true,
+		/// </summary>
+		public event Action<long> TimeoutAdded;
```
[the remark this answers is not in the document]
BDisp: There is need to have a `TimeoutRemoved` event because the `MainLoop` class will handle this itself. We only need a `TimeoutAdded` event for `WebConsole` being notified because `SynchronizationContext` doesn't work on a web app.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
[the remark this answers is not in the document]
BDisp: Well, when the `Contents` was only needed basically for unit tests, this was irrelevant. But when the `Border` class was created, it was needed for to write the `3D Offset`, then it passed to be mandatory for all drivers and thus it must be `abstract` or at least `virtual`. The refactor indeed suggest to convert it to a method, but `dotnet` has many only `getter` properties or with `private/internal setters` and doesn't represent any break to design pattern. Since it's already used as is, I suggest maintain as `getter` property, if you don't mind.

Terminal.Gui/Core/Application.cs
BDisp: Done thanks.

Terminal.Gui/Core/Application.cs
BDisp: Done thanks.

BDisp:
Now is the `Editor` scenario that is giving failed test, OMG.

editor-scenario

It's not supposed the `Build` should be run after the `Code scanning`?

code-scanning

tig:
> It's not supposed the `Build` should be run after the `Code scanning`?
> 

They are independent workflows. Theoretically they run in parallel.

PR #1771, opened 2022-06-06
Fixes #1769. Supports a minimum view size for non-automatic size views.

### LINKED ISSUE #1769: Label: If created with 'string.Empty' in constructor, fails to render
**Describe the bug**
If `Label` is created with `new Label(string.Empty)` it fails to ever render. In some cases it causes an exception in `TextFormatter` (`throw new InvalidOperationException ("Size must be set before accessing Lines");`).

**To Reproduce**
Modify `Generic.cs`:

```csharp
using Terminal.Gui;

namespace UICatalog.Scenarios {
	[ScenarioMetadata (Name: "Generic", Description: "Generic sample - A template for creating new Scenarios")]
	public class MyScenario : Scenario {
		public override void Setup ()
		{
			// Put your scenario code here, e.g.
			var button = new Button ("Press me!") {
				X = 1,
				Y = 1,
			};

			var label = new Label (string.Empty) {
				X = 1,
				Y = 2,
				ColorScheme = Colors.Error,
				Width = 50
			};
			Win.Add (label);

			button.Clicked += () => {
				label.Text = "Hello";
			};
			Win.Add (button);
		}
	}
}
````

**Expected behavior**
Change thee above constructor to 
```csharp
		var label = new Label (" ") {
````

To see it operate correctly.

I found this upgrading `Out-ConsoleGridView` from 1.0 to 1.6. A temporary fix in OCGV is to use `" "` instead of `string.Empty`.

BDisp:


BDisp:
I think that this failed tests has to do with the sequence of the tests. `Build Terminal.Gui` has started before the `Code scanning` is ended.

tig (approved):
Very nice work.

PR #1779, opened 2022-06-10
Fixes #1739. Setting menu UseKeysUpDownAsKeysLeftRight as false by default.

### LINKED ISSUE #1739: Sub menus auto expand on down arrow
**Describe the bug**

When pressing arrow key down in a menu, any sub menus automatically expand and draw focus from the main menu.  The only way for the user to continue down the main list is to press then LEFT arrow to get back into the main tree. 

This is not the way keyboard navigation works in most GUIs e.g. Visual Studio:

sub-menus
_Comparing keyboard navigation of menus that have submenus in Visual Studio vs Terminal.Gui_

**To Reproduce**
1. Create menu with multiple submenus
2. Press down arrow multiple times

**Expected behavior**
It should be possible to keep scrolling down a menu using only up/down.  Submenus should only expand on pressing RIGHT arrow (when using keyboard only navigation).

BDisp:


tig (approved):
Looks great. Nice work.

Issue #1768, opened 2022-06-02
Windows Terminal: Resizing can cause display to get corrupted


tig:
I've seen this before, but never captured it. 

![aCYOb97 1](https://user-images.githubusercontent.com/585482/173106977-26a938c7-68db-4d83-9af5-a676a1e1f578.gif)

alexeevvlad:
Same problem when using Bitvise SSH Client 8.53 xterm-256color terminal on Linux server.
I try bind function Application.Refresh(); for manual refresh screen, but no effect. How function i can use for manual update all content for new terminal size? How about Application.Driver.UpdateScreen()?

BDisp:
I can't reproducing this on resizing. The only display corrupted I see is when I scroll up the buffer height. I didn't find yet a way to force the `Windows Terminal` to set the size equal to the `Terminal.Gui` console.

![windows-terminal-resize-behavior](https://user-images.githubusercontent.com/13117724/173117541-05b26147-6bee-47e2-868f-1a90a82af840.gif)

tig:
I occasionally get this effect when starting a UI Catalog debug session:

<img width="545" alt="image" src="https://user-images.githubusercontent.com/585482/173134762-84972f5c-b864-4ea2-a281-9e07ec23d294.png">

BDisp:
> I occasionally get this effect when starting a UI Catalog debug session:

But it's paused in the debugger, right?

tig:
No.

-cek
Sent via mobile. Expect brevity, typos, and dangerous driving.
________________________________

BDisp:
I already fixed the resize issue for all drivers. Nonetheless with the `NetDriver` on `Terminal.Gui.Window` resizing I'm having this issue that I'm still investigating. @tig can you confirm if the issue you reported is with the `NetDriver`, please.

![netdriver-behavior](https://user-images.githubusercontent.com/13117724/173154094-8435074e-7537-4dba-bb64-d0cf33c51453.gif)

tig:
This is netdriver:

![6bsIjur 1](https://user-images.githubusercontent.com/585482/173173882-9e3a2990-398c-42ca-a8f0-61fa507a3252.gif)

The issue I reported with the screen appearing corrupted when the app starts is not easily reproducable. It happens only occasionally.

BDisp:
Is this still happen?

tig:
I don't remember. I'll pay attention to it next few days.

tig:
[...]

PR #1788, opened 2022-06-11
Fixes Windows Terminal initial resize on CursesDriver related with #1768.

BDisp:
It wasn't clearing at `CursesDriver` initialization.


BDisp:
Should `ResizeScreen` method be made abstract?

BDisp:
> Should `ResizeScreen` method be made abstract?

@tig, @tznind what your opinion about this?

BDisp:
CursesDriver was the one that used this method before and now it's using I thinking is better made as abstract and override in each derived class.

BDisp:
> I don't see a problem with doing that. What is your concern?

I'll do it then. Thanks.

tig (approved):
Nice.

PR #1790, opened 2022-06-11
Simplified Run_All_Scenarios test

### LINKED ISSUE #1712: Intermittent Github Action build failures
Happens frequently. 

image

Could it be the unit tests you added in this PR @tznind?

https://github.com/migueldeicaza/gui.cs/pull/1684

tznind:
Fixes #1712

This PR changes the test to use `FakeDriver` and simply Init, Setup, Run each scenario.  The callback closes the scenario after 200ms.

I think this achieves the goal of the test which is to confirm that scenarios do not crash on load?

### DIFF
--- a/UnitTests/ScenarioTests.cs
+++ b/UnitTests/ScenarioTests.cs
@@ -37,10 +37,14 @@ namespace Terminal.Gui {
 			return FakeConsole.MockKeyPresses.Count;
 		}
 
+
 		/// <summary>
+		/// <para>
 		/// This runs through all Scenarios defined in UI Catalog, calling Init, Setup, and Run.
-		/// It puts a Ctrl-Q in the input queue so the Scenario immediately exits. 
-		/// Should find any egregious regressions.
+		/// </para>
+		/// <para>
+		/// Should find any Scenarios which crash on load or do not respond to <see cref="Application.RequestStop()"/>.
+		/// </para>
 		/// </summary>
 		[Fact]
 		public void Run_All_Scenarios ()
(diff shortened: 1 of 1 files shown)


tznind:
@BDisp do you think this covers enough of the intended purpose of the original test?

BDisp:
Not all, when @tig did this test was to check the immediate response at the first call to the `Application.RequestStop` method, without the need abort it.

> I think this achieves the goal of the test which is to confirm that scenarios do not crash on load?

This gave me the idea to check if the instance is successfully created and I test for a exception. After this change I get failure on the `TestAddManyTimeouts` unit test and I thought if this was interfering with the `Run_All_Scenarios`, so I locked the `Application.Top`. But I haven't certain is this won't failed again.

tig:
Way back we had some regressions in Scenarios that prevented them from exiting. That was one of the reasons, IIRC, I added the ctrl-q stuff. But it was a long time ago.

When I built the "All Views Tester" Scenario it found a lot of bugs. I felt that trying to simulate manually running through each scenario as one could do with All Views Tester would make good stress tests. 

However, now that we have much better unit test coverage I think its ok to simplify this, especially given the pain it's causing in the CI/CD process (if it is in fact the cause).

BDisp:
@tznind I think that better you resolve the conflicts for @tig merge this, because I got the unit test failure again.

tig:
@tznind can you resolve conflicts?

tznind:
I have resolved conflicts.  CI failed first time checking out NuGet packages but I have hit re-run and now passing.  Error was:

```
/home/runner/work/gui.cs/gui.cs/.dotnet/sdk/6.0.100/NuGet.targets(130,5): error :   Response status code does not indicate success: 502 (Bad Gateway). [/home/runner/work/gui.cs/gui.cs/Terminal.sln]
```

PR #1792, opened 2022-06-13
Added Pos/Dim Function feature to automate layout.

### LINKED ISSUE #1791: View should not turn off AutoSize (in setting Width and/or Height)
**Describe the bug**
In https://github.com/migueldeicaza/gui.cs/pull/1685 from @BDisp, there's this commit:

```
Commit a82ac45843d6153f10d7c545721d4aeabb75266d
Author: BDisp <bd.bdisp@gmail.com>
Date: Friday, April 29, 2022 5:12 PM
Parent: 933a1bda

Fixes AutoSize, Window Tile and added option to Border for fill or not.
```

https://github.com/migueldeicaza/gui.cs/pull/1685/files#diff-4a82080ff9e808d1867c94cdd6ac106755bd80a81509546e43f154aed1fe90deR572

```csharp
		/// <summary>
		/// Gets or sets the width of the view. Only used the <see cref="LayoutStyle"/> is <see cref="LayoutStyle.Computed"/>.
		/// </summary>
		/// <value>The width.</value>
		/// <remarks>
		/// If <see cref="LayoutStyle"/> is <see cref="LayoutStyle.Absolute"/> changing this property has no effect and its value is indeterminate. 
		/// </remarks>
		public Dim Width {
			get => width;
			set {
				if (!ValidatePosDim (width, value)) {
					throw new ArgumentException ();
				}

				width = value;
				if (autoSize && value.Anchor (0) != TextFormatter.Size.Width) {
					autoSize = false;
				}
				SetNeedsLayout ();
				if (width is Dim.DimAbsolute) {
					frame = new Rect (frame.X, frame.Y, width.Anchor (0), frame.Height);
				}
				TextFormatter.Size = frame.Size;
				SetNeedsDisplay (frame);
			}
		}
```

See how `AutoSize` is set to false? This breaks code where a `Button` (or any View) has been set to autosize and then has it's text changed and the view has been moved (e.g. in the new `Dialog` code for justifying buttons). 

I found this in coding up Wizard. I noticed that when I changed the text of the `NextFinished` button from `Next...` to `Finished` (which is longer) the button did not resize. A workaround is to re-set `AutoSize` to true when changing the text, but this should not be needed (and will not work in other scenarios).

This Unit Test should be added once this bug is fixed:

```csharp
		[Fact, AutoInitShutdown]
		public void AutoSize_Stays_True ()
		{
			var btn = new Button () {
				X = Pos.Center (),
				Y = Pos.Center (),
				Text = "Say Hello 你",
				AutoSize = true
			};

			var win = new Window () {
				Width = Dim.Fill (),
				Height = Dim.Fill (),
				Title = "Test Demo 你"
			};
			win.Add (btn);
			Application.Top.Add (win);

			Assert.True (btn.AutoSize);

			var expected = @"
┌ Test Demo 你 ──────────────┐
│      [ Say Hello 你 ]      │
";

			Assert.True (btn.AutoSize);
			btn.Width = Dim.Fill ();
			Assert.True (btn.AutoSize);
		}
```

@BDisp, can you find another way to solve whatever you fixed by doing this?

BDisp:
It's enough setting once and the value is always calculate when changed.

### DIFF
--- a/Terminal.Gui/Core/PosDim.cs
+++ b/Terminal.Gui/Core/PosDim.cs
@@ -36,6 +36,7 @@ namespace Terminal.Gui {
 			return 0;
 		}
 
+		// Helper class to provide dynamic value by the execution of a function that returns an integer.
 		internal class PosFunc : Pos {
 			Func<int> function;
 
@@ -60,7 +61,7 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
-		/// Creates an Func <see cref="Pos"/> from the specified function value.
+		/// Creates a "PosFunc" from the specified function.
 		/// </summary>
 		/// <param name="function">The function to be executed.</param>
 		/// <returns>The <see cref="Pos"/> returned from the function.</returns>
@@ -425,6 +426,7 @@ namespace Terminal.Gui {
 			return 0;
 		}
 
+		// Helper class to provide dynamic value by the execution of a function that returns an integer.
 		internal class DimFunc : Dim {
 			Func<int> function;
 
@@ -449,7 +451,7 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
-		/// Creates an Func <see cref="Dim"/> from the specified function value.
+		/// Creates a "DimFunc" from the specified function.
 		/// </summary>
 		/// <param name="function">The function to be executed.</param>
 		/// <returns>The <see cref="Dim"/> returned from the function.</returns>
--- a/Terminal.Gui/Core/PosDim.cs
+++ b/Terminal.Gui/Core/PosDim.cs
@@ -167,22 +167,24 @@ namespace Terminal.Gui {
 
 		internal class PosAbsolute : Pos {
 			int n;
+			Func<int> function;
 			public PosAbsolute (int n) { this.n = n; }
+			public PosAbsolute (Func<int> function) { this.function = function; }
 
 			public override string ToString ()
 			{
-				return $"Pos.Absolute({n})";
+				return $"Pos.Absolute({(function == null ? n : function ())})";
 			}
 
 			internal override int Anchor (int width)
 			{
-				return n;
+				return function == null ? n : function ();
 			}
 
-			public override int GetHashCode () => n.GetHashCode ();
+			public override int GetHashCode () => function == null ? n.GetHashCode () : function ().GetHashCode ();
 
-			public override bool Equals (object other) => other is PosAbsolute abs && abs.n == n;
-
+			public override bool Equals (object other) => other is PosAbsolute abs
+				&& (function == null ? abs.n == n : abs.function () == function ());
 		}
 
 		/// <summary>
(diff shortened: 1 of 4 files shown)


Terminal.Gui/Core/PosDim.cs
```diff
@@ -167,22 +167,24 @@ public static Pos Center ()
 
 		internal class PosAbsolute : Pos {
 			int n;
+			Func<int> function;
 			public PosAbsolute (int n) { this.n = n; }
+			public PosAbsolute (Func<int> function) { this.function = function; }
```
tznind, line 172: Please could this be a new class e.g. `PosFunc` instead of being part of `PosAbsolute`?  Currently `PosAbsolute` is an absolute unchanging position while others e.g. `PosRelative` are changing.  I think it would be nice to keep these roles if possible.

I know these are private classes but I think we should still follow the single responsibility principle if possible.

BDisp: Thanks for your review. I totally agree with you. I'll create the `PosFunc` class. If the `PosRelative` is more appropriate please let me know.

Terminal.Gui/Core/PosDim.cs
```diff
@@ -36,6 +36,39 @@ internal virtual int Anchor (int width)
...
+				return $"Pos.PosFunc({function ()})";
+			}
+
+			public override int GetHashCode () => function.GetHashCode ();
+
+			public override bool Equals (object other) => other is PosFunc f && f.function () == function ();
+		}
+
+		/// <summary>
+		/// Creates an Func <see cref="Pos"/> from the specified function value.
```
tig, line 63: "Creates a "PosFunc" from the specified function"

Terminal.Gui/Core/PosDim.cs
```diff
@@ -36,6 +36,39 @@ internal virtual int Anchor (int width)
 			return 0;
 		}
 
+		internal class PosFunc : Pos {
```
tig, line 39: Can you provide some documentation describing the purpose of this class?

Terminal.Gui/Core/PosDim.cs
```diff
@@ -375,6 +425,39 @@ internal virtual int Anchor (int width)
 			return 0;
 		}
 
+		internal class DimFunc : Dim {
```
tig, line 428: Same... needs conceptual documentation; a sentence or two that explains WHY someone would use this class.

Terminal.Gui/Core/PosDim.cs
```diff
@@ -375,6 +425,39 @@ internal virtual int Anchor (int width)
...
+				return $"Dim.DimFunc({function ()})";
+			}
+
+			public override int GetHashCode () => function.GetHashCode ();
+
+			public override bool Equals (object other) => other is DimFunc f && f.function () == function ();
+		}
+
+		/// <summary>
+		/// Creates an Func <see cref="Dim"/> from the specified function value.
```
tig, line 452: "Creates a Dim..."

tig (changes requested):
I haven't had a chance to test this yet, but I like the concept!

Terminal.Gui/Core/PosDim.cs
BDisp: Done

Terminal.Gui/Core/PosDim.cs
BDisp: Done

tig:
I may be missing something, but this PR does not actually fix #1791 - After checking this PR out locally and running Wizard Scenario, the Finished button is still not sized correctly:

image

tig (approved):
Thanks.

PR #1803, opened 2022-06-13
Wizard and Dialog fix.

BDisp:


tig (approved):
LGTM. Can you explain what it does though?

BDisp:
> LGTM. Can you explain what it does though?

It was capturing the vertical scroll bar and not the other horizontal.
In Dialog the left justify wasn't work well because of the Accept button, which I think you used to test a wizard example.

PR #1804, opened 2022-06-13
Fixes #1796. If TextView is ReadOnly tab and return are ignored. Also fixes Button click via spacebar

### LINKED ISSUE #1796: Button: Spacebar does not always cause a button clicked event
To reproduce see this step in the Wizards Scenario:

image

Tab around the dialog, and note that if "Back" or "Next..." have focus, hitting the space bar does a button click.

However, when "Press Me" has focus, spacebar doesn't do anything.

BDisp:


tig (approved):
Excellent. Thanks.

PR #1807, opened 2022-06-13
Fixes #1797. AutoSize is true by default on CheckBox.

### LINKED ISSUE #1797: CheckBox: AutoSize = true causes control to not be visible
This unit test should be added to `CheckBoxTests.cs`:

```csharp

		[Fact, AutoInitShutdown]
		public void AutoSize_StaysVisible ()
		{

			var checkBox = new CheckBox () {
				X = 1,
				Y = Pos.Center (),
				Text = "Check this out 你"
			};
			var win = new Window () {
				Width = Dim.Fill (),
				Height = Dim.Fill (),
				Title = "Test Demo 你"
			};
			win.Add (checkBox);
			Application.Top.Add (win);

			Assert.False (checkBox.IsInitialized);

			var runstate = Application.Begin (Application.Top);

			Assert.True (checkBox.IsInitialized);
			Assert.Equal ("Check this out 你", checkBox.Text);

			var expected = @"
┌ Test Demo 你 ──────────────┐
│ ╴ Check this out 你        │
";

			// Positive test
			var pos = GraphViewTests.AssertDriverContentsWithFrameAre (expected, output);

			// Negative test
			checkBox.AutoSize = true;
			bool first = false;
			Application.RunMainLoopIteration (ref runstate, true, ref first);

		}
```

Output as of now:

```
 Terminal.Gui.Views.CheckboxTests.AutoSize_StaysVisible
   Source: CheckboxTests.cs line 72
   Duration: 106 ms

  Message: 
    Assert.Equal() Failure
                                      ↓ (pos 63)
    Expected: ···                │\n│ ╴ Check this out 你        │\n│            ···
    Actual:   ···                │\n│                            │\n│           ···
                                      ↑ (pos 63)

  Stack Trace: 
    GraphViewTests.AssertDriverContentsWithFrameAre(String expectedLook, ITestOutputHelper output) line 208
    CheckboxTests.AutoSize_StaysVisible() line 112

  Standard Output: 
    Expected:
    ┌ Test Demo 你 ──────────────┐
    │ ╴ Check this out 你        │
    But Was:
    ┌ Test Demo 你 ──────────────┐
    │ ╴ Check this out 你        │
    Expected:
    ┌ Test Demo 你 ──────────────┐
    │ ╴ Check this out 你        │
    But Was:
    ┌ Test Demo 你 ──────────────┐
```

BDisp:
Fixes #1797

### DIFF
--- a/Terminal.Gui/Core/TextFormatter.cs
+++ b/Terminal.Gui/Core/TextFormatter.cs
@@ -136,7 +136,7 @@ namespace Terminal.Gui {
 			set {
 				text = value;
 
-				if (text.RuneCount > 0 && (Size.Width == 0 || Size.Height == 0 || Size.Width != text.ConsoleWidth)) {
+				if (text != null && text.RuneCount > 0 && (Size.Width == 0 || Size.Height == 0 || Size.Width != text.ConsoleWidth)) {
 					// Provide a default size (width = length of longest line, height = 1)
 					// TODO: It might makes more sense for the default to be width = length of first line?
 					Size = new Size (TextFormatter.MaxWidth (Text, int.MaxValue), 1);
@@ -154,9 +154,15 @@ namespace Terminal.Gui {
 		/// </summary>
 		public bool AutoSize { get; set; }
 
+		/// <summary>
+		/// Get or sets a flag if the wrapped text will keep the trailing spaces or
+		/// the trailing spaces will be trimmed (default).
+		/// </summary>
+		public bool PreserveTrailingSpaces { get; set; }
+
 		// TODO: Add Vertical Text Alignment
 		/// <summary>
-		/// Controls the horizontal text-alignment property. 
+		/// Controls the horizontal text-alignment property.
 		/// </summary>
 		/// <value>The text alignment.</value>
 		public TextAlignment Alignment {
@@ -327,18 +333,16 @@ namespace Terminal.Gui {
 					if (IsVerticalDirection (textDirection)) {
 						var colsWidth = GetSumMaxCharWidth (shown_text, 0, 1);
 						lines = Format (shown_text, Size.Height, textVerticalAlignment == VerticalTextAlignment.Justified, Size.Width > colsWidth,
-							false, 0, textDirection);
+							PreserveTrailingSpaces, 0, textDirection);
 						if (!AutoSize) {
 							colsWidth = GetMaxColsForWidth (lines, Size.Width);
 							if (lines.Count > colsWidth) {
-								for (int i = colsWidth; i < lines.Count; i++) {
-									lines.Remove (lines [i]);
-								}
+								lines.RemoveRange (colsWidth, lines.Count - colsWidth);
 							}
 						}
 					} else {
 						lines = Format (shown_text, Size.Width, textAlignment == TextAlignment.Justified, Size.Height > 1,
-							false, 0, textDirection);
+							PreserveTrailingSpaces, 0, textDirection);
 						if (!AutoSize && lines.Count > Size.Height) {
 							lines.RemoveRange (Size.Height, lines.Count - Size.Height);
 						}
@@ -468,7 +472,7 @@ namespace Terminal.Gui {
 			var runes = StripCRLF (text).ToRuneList ();
 			if (!preserveTrailingSpaces) {
 				if (IsHorizontalDirection (textDirection)) {
-					while ((end = start + GetMaxLengthForWidth (runes.GetRange (start, runes.Count - start), width)) < runes.Count) {
+					while ((end = start + Math.Max (GetMaxLengthForWidth (runes.GetRange (start, runes.Count - start), width), 1)) < runes.Count) {
 						while (runes [end] != ' ' && end > start)
 							end--;
 						if (end == start)
@@ -494,16 +498,25 @@ namespace Terminal.Gui {
 				}
 			} else {
 				while ((end = start) < runes.Count) {
-					end = GetNextWhiteSpace (start, width);
+					end = GetNextWhiteSpace (start, width, out bool incomplete);
+					if (end == 0 && incomplete) {
+						start = text.RuneCount;
+						break;
+					}
 					lines.Add (ustring.Make (runes.GetRange (start, end - start)));
 					start = end;
+					if (incomplete) {
+						start = text.RuneCount;
+						break;
+					}
 				}
 			}
 
-			int GetNextWhiteSpace (int from, int cWidth, int cLength = 0)
+			int GetNextWhiteSpace (int from, int cWidth, out bool incomplete, int cLength = 0)
 			{
 				var to = from;
 				var length = cLength;
+				incomplete = false;
 
 				while (length < cWidth && to < runes.Count) {
 					var rune = runes [to];
@@ -512,13 +525,19 @@ namespace Terminal.Gui {
 					} else {
 						length++;
 					}
+					if (length > cWidth) {
+						if (to >= runes.Count || (length > 1 && cWidth <= 1)) {
+							incomplete = true;
+						}
+						return to;
+					}
 					if (rune == ' ') {
 						if (length == cWidth) {
 							return to + 1;
 						} else if (length > cWidth) {
 							return to;
 						} else {
-							return GetNextWhiteSpace (to + 1, cWidth, length);
+							return GetNextWhiteSpace (to + 1, cWidth, out incomplete, length);
 						}
 					} else if (rune == '\t') {
 						length += tabWidth + 1;
@@ -527,7 +546,7 @@ namespace Terminal.Gui {
 						} else if (length > cWidth && tabWidth > cWidth) {
 							return to;
 						} else {
-							return GetNextWhiteSpace (to + 1, cWidth, length);
+							return GetNextWhiteSpace (to + 1, cWidth, out incomplete, length);
 						}
 					}
 					to++;
@@ -693,9 +712,6 @@ namespace Terminal.Gui {
 		{
 			if (width < 0) {
 				throw new ArgumentOutOfRangeException ("width cannot be negative");
-			}
-			if (preserveTrailingSpaces && !wordWrap) {
-				throw new ArgumentException ("if 'preserveTrailingSpaces' is true, then 'wordWrap' must be true either.");
 			}
 			List<ustring> lineResult = new List<ustring> ();
 
@@ -1093,10 +1109,10 @@ namespace Terminal.Gui {
 				break;
 			}
 
-			for (int line = 0; line < linesFormated.Count; line++) {
-				var isVertical = IsVerticalDirection (textDirection);
+			var isVertical = IsVerticalDirection (textDirection);
 
-				if ((isVertical && (line > bounds.Width)) || (!isVertical && (line > bounds.Height)))
+			for (int line = 0; line < linesFormated.Count; line++) {
+				if ((isVertical && line > bounds.Width) || (!isVertical && line > bounds.Height))
 					continue;
 
 				var runes = lines [line].ToRunes ();
--- a/Terminal.Gui/Core/TextFormatter.cs
+++ b/Terminal.Gui/Core/TextFormatter.cs
@@ -934,7 +934,7 @@ namespace Terminal.Gui {
 				w = mw;
 				h = ml;
 			} else {
-				int vw = 0;
+				int vw = 1, cw = 1;
 				int vh = 0;
 
 				int rows = 0;
@@ -945,14 +945,13 @@ namespace Terminal.Gui {
 							vh = rows;
 						}
 						rows = 0;
+						cw = 1;
 					} else if (rune != '\r') {
 						rows++;
 						var rw = Rune.ColumnWidth (rune);
-						if (rw < 0) {
-							rw++;
-						}
-						if (rw > vw) {
-							vw = rw;
+						if (cw < rw) {
+							cw = rw;
+							vw++;
 						}
 					}
 				}
--- a/Terminal.Gui/Core/TextFormatter.cs
+++ b/Terminal.Gui/Core/TextFormatter.cs
@@ -155,12 +155,13 @@ namespace Terminal.Gui {
 		public bool AutoSize { get; set; }
 
 		/// <summary>
-		/// Get or sets a flag if the wrapped text will keep the trailing spaces or
-		/// the trailing spaces will be trimmed (default).
+		/// Gets or sets a flag that determines whether <see cref="Text"/> will have trailing spaces preserved
+		/// or not when <see cref="WordWrap"/> is enabled. If `true` any trailing spaces will be trimmed when
+		/// either the <see cref="Text"/> property is changed or when <see cref="WordWrap"/> is set to `true`.
+		/// The default is `false`.
 		/// </summary>
 		public bool PreserveTrailingSpaces { get; set; }
 
-		// TODO: Add Vertical Text Alignment
 		/// <summary>
 		/// Controls the horizontal text-alignment property.
 		/// </summary>
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -827,7 +827,7 @@ namespace Terminal.Gui {
 			var _y = y is Pos.PosAbsolute ? y.Anchor (0) : frame.Y;
 
 			if (AutoSize) {
-				var s = CalculateAutoSize ();
+				var s = GetAutoSize ();
 				var w = width is Dim.DimAbsolute && width.Anchor (0) > s.Width ? width.Anchor (0) : s.Width;
 				var h = height is Dim.DimAbsolute && height.Anchor (0) > s.Height ? height.Anchor (0) : s.Height;
 				frame = new Rect (new Point (_x, _y), new Size (w, h));
@@ -2118,7 +2118,7 @@ namespace Terminal.Gui {
 			var s = Size.Empty;
 
 			if (AutoSize) {
-				s = CalculateAutoSize ();
+				s = GetAutoSize ();
 			}
 
 			if (x is Pos.PosCenter) {
@@ -2401,10 +2401,12 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
-		/// Used by <see cref="Text"/> to resize the view's <see cref="Bounds"/> with the <see cref="TextFormatter.Size"/>.
-		/// Setting <see cref="AutoSize"/> to true only work if the <see cref="Width"/> and <see cref="Height"/> are null or
-		///   <see cref="LayoutStyle.Absolute"/> values and doesn't work with <see cref="LayoutStyle.Computed"/> layout
-		///   and <see cref="ForceValidatePosDim"/> to avoid breaking the <see cref="Pos"/> and <see cref="Dim"/> settings.
+		/// Gets or sets a flag that determines whether the View will be automatically resized to fit the <see cref="Text"/>.
+		/// The default is `false`. Set to `true` to turn on AutoSize. If <see cref="AutoSize"/> is `true` the <see cref="Width"/>
+		/// and <see cref="Height"/> will always be used if the text size is lower. If the text size is higher the bounds will
+		/// be resized to fit it.
+		/// In addition, if <see cref="ForceValidatePosDim"/> is `true` the new values of <see cref="Width"/> and
+		/// <see cref="Height"/> must be of the same types of the existing one to avoid breaking the <see cref="Dim"/> settings.
 		/// </summary>
 		public virtual bool AutoSize {
 			get => autoSize;
@@ -2421,8 +2423,10 @@ namespace Terminal.Gui {
 		}
 
 		/// <summary>
-		/// Get or sets a flag if the wrapped text will keep the trailing spaces or
-		/// the trailing spaces will be trimmed (default).
+		/// Gets or sets a flag that determines whether <see cref="TextFormatter.Text"/> will have trailing spaces preserved
+		/// or not when <see cref="TextFormatter.WordWrap"/> is enabled. If `true` any trailing spaces will be trimmed when
+		/// either the <see cref="Text"/> property is changed or when <see cref="TextFormatter.WordWrap"/> is set to `true`.
+		/// The default is `false`.
 		/// </summary>
 		public virtual bool PreserveTrailingSpaces {
 			get => TextFormatter.PreserveTrailingSpaces;
@@ -2496,7 +2500,7 @@ namespace Terminal.Gui {
 		public virtual bool IsInitialized { get; set; }
 
 		/// <summary>
-		/// Gets information if the view was already added to the superview.
+		/// Gets information if the view was already added to the <see cref="SuperView"/>.
 		/// </summary>
 		public bool IsAdded { get; private set; }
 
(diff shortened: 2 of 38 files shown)


tig:
Is this really a Draft?

BDisp:
> Is this really a Draft?

Yes, I did big changes to this but not yet finished, perhaps tomorrow.

tig (changes requested):
I wonder if this can be simplified by failing with an exception when AutoSize is invalid instead of changing it to false?

Terminal.Gui/Core/View.cs
```diff
@@ -568,17 +568,19 @@ void SetTabIndex ()
...
 
 				width = value;
-				if (autoSize && value.Anchor (0) != TextFormatter.Size.Width
-					- (TextFormatter.IsHorizontalDirection (TextDirection)
-					&& TextFormatter.Text.Contains (HotKeySpecifier) ? 1 : 0)) {
+
+				var isValidNewAutSize = autoSize ? IsValidAutoSizeWidth (width) : false;
+
+				if (autoSize && !isValidNewAutSize) {
+					TextFormatter.AutoSize = false;
```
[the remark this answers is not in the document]
BDisp: I agree with you, it will allow to users pay more attention of that and learning more by throwing an exception. Thanks.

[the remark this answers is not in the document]
tznind: Just for reference the `AutoSize` in winforms says

> When using [AutoSize](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.form.autosize?view=windowsdesktop-6.0), [...] the current value of the [Size](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.size?view=windowsdesktop-6.0) property is ignored.

If we are throwing then we should make sure we are not creating a breaking change.

Maybe I'm not fully following but if before you could do 
```
myView.AutoSize = true;
myView.Width = 10;
```
Then it shouldn't suddenly start throwing Exception.  Especially if it doesn't throw when you do:

```
myView.Width = 10;
myView.AutoSize = true;
```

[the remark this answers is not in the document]
BDisp: It only throw the exception when the `IsInitialized` is true,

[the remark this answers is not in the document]
tznind: Ok, that seems more fair.  In that case I think it should be [InvalidOperationException](https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-6.0) with a message like "Width cannot be modified when when AutoSize is enabled"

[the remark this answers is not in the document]
BDisp: I already did like this:
`throw new InvalidOperationException ("Must set AutoSize to false before set the Width.");`

If not right please say.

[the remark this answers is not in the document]
BDisp: > Maybe I'm not fully following but if before you could do
> 
> ```
> myView.AutoSize = true;
> myView.Width = 10;
> ```
> 
> Then it shouldn't suddenly start throwing Exception. Especially if it doesn't throw when you do:
> 
> ```
> myView.Width = 10;
> myView.AutoSize = true;
> ```

@tznind setting the width and the height with absolute value that fit the auto size feature, it will stay true. But if you setting with values that doesn't fit the auto size when the `IsInitialized` is processed it set to false So, expressions like below that are set before the `IsInitialized` is processed will not throw even if don't fit the auto size. I have no choice to avoid to silent set to false on the `IsInitialized`, if it don't fit the auto size, otherwise it will throw the exception. I call your attention that the order which the object initializers are called are essential to avoid wrong values. Before by silent set to false  this wasn't an issue, so there must be coherence on how to order the objects. If you know how to bypass the order let me know, please.

`var myView = new View("Hi") { Width = 10,  AutoSize = true }; ` AutoSize will be true and the width will be 2.

`var myView = new View("Hi") { AutoSize = true, Width = 10 }; ` AutoSize will be false and width will be 10.

`var myView = new View("Hi") { Width = Dim.Fill(),  AutoSize = true };` AutoSize will be false and the width will be the parent width.

`Button` and `Checkbox` has the AutoSize enabled because the absolute values fit the auto size values.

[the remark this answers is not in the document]
tznind:
```
var myView = new View("Hi") { Width = 10, AutoSize = true };
```
To me this reads as 'set size to 10 but auto resize it if the text grows above that.  It is a bit annoying that the order of initializer matters.  Is there no way to store all the values then apply them at Initialize time in a consistent order?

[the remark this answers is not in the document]
BDisp: I also would like to know. It must be consistent at design time too. It's a good idea, considering a minimum size for auto size.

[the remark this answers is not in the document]
BDisp: But that minimum value will not be preserved after it grows, right?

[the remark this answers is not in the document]
BDisp: Well this is one more difficulty level. I don't know if it worth.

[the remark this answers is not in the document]
BDisp: > ```
> var myView = new View("Hi") { Width = 10, AutoSize = true };
> ```
> 
> 
>     
>   
> 
> To me this reads as 'set size to 10 but auto resize it if the text grows above that'.

@tznind, we must clarify about `AutoSize` does. I don't agree that the behavior you mentioned should be handle by the `View` but through a derived class, like `Button` and `Checkbox` do. Otherwise it will throw an exception if you decide at runtime to change the Width=20 with the text kept as "Hi" and auto size enabled. To do this we must first disable the auto size. To kept a minimum width a derived class must implement that property and deal with that. But perhaps I'm wrong about that.
`Dim` types other than `DimAbsolute` will always be used all the respective area and auto size must be disabled.
`DimAbsolute`can handled with auto size enable if his size fit the TextFormatter.Text, otherwise it must be disabled.
The `View` class can handle generic behaviors and not with special specific behaviors.
Since you are developing the great repo `TerminalGuiDesigner` you probably have more opinion about this.

tig:
Before we spend more time writing code on this, let's step back and make sure we're all in agreement (or not) on the fundamentals of AutoSize.

I really, really, like the WinForms definition @tznind mentioned.: If AutoSize == true, Height & Width are ignored.

I really, really, do not like the idea of a behavior property like `AutoSize` being changed under the covers. 

Even though I suggested it, upon further reflection, I'm leery of throwing an exception in the error case (mostly because it does feel like a potential breaking change). 

Let's try to write the documentation for AutoSize as though we were first designing it:

```
Determines whether the View will automatically change its size (`.Height` and `.Width`) to fit the `.Text`. If `AutoSize` is `true`, changes to `Height` and `Width` are ignored, and retrieving their values will be indeterminate (use `Bounds` to determine the Views's size instead). For some Views (e.g. `Label` and `Button`), `AutoSize` is `true` by default. In others (e.g. `Window`) it is `false` by default. 
```

Is this the definition you both agree with? If not, please re-word it to match your definition. Let's keep editing this documentation until we belive it fully specifies the behavior that will work.

BDisp:
> If AutoSize == true, Height & Width are ignored.

@tig, please clarify me on this. You mean if the `AutoSize` is `true` and `Width` and `Height` are `Dim.Fill` they are changed to `DimAbsolute` to fit the text or they be kept as `Dim.Fill`?

tznind:
> Is this the definition you both agree with? 

That definition works for me.

BDisp:
> > If AutoSize == true, Height & Width are ignored.
> 
> @tig, please clarify me on this. You mean if the `AutoSize` is `true` and `Width` and `Height` are `Dim.Fill` they are changed to `DimAbsolute` to fit the text or they be kept as `Dim.Fill`?

Ok I already get that if `AutoSize` is `true`, changes to `Height` and `Width` are ignored, but they really are changed if they are `Dim.Fill` and changing to `Dim.Absolute`. No problem with deal if `AutoSize` is `true` the frame is not resized with `Width` and `Height`, but I need to know if the bellow is to be ignore too?
```cs
                                if (!ValidatePosDim (width, value)) {
					throw new ArgumentException ();
				}
```

BDisp:
There is also a major modification in many unit tests.

BDisp:
I understand you both also have doubt. One thing I want to achieve is the object state are the same in both design time as well as in runtime.

tznind:
So based on the description @tig  wrote, would the pseudocode for Width/Height make sense like this: 

```csharp
  private Dim width;
  public Dim Width {
      get
      {
          if (AutoSize)
          {
              return CalculateAutoSizeWidth();
          }
      
          return width;
      }
      set
      {
          width = value;
      }
  }
```

> Otherwise @tznind will deal with difficulties on design time :-)

Don't worry about that too much.  So far I haven't run into any of these 'validate' style Exceptions that I can't work around e.g.

https://github.com/tznind/TerminalGuiDesigner/blob/8a4b4020aef27b8f320076ba32a62141deaa5072/src/ToCode/Property.cs#L85-L92

BDisp:
Thanks. I deduct that `CalculateAutoSizeWidth` method returns a `DimAbsolute`, right? So, if `AutoSize` is `false` the `ValidatePosDim (width, value)` is not needed anymore? I was thinking changing the `IsInitialized` setter to private, because it's the `Application` who handle it, but that way you will not could set it. If I can set the same object state on both, design and rutnime, that will not be a problem.

tznind:
> Thanks. I deduct that `CalculateAutoSizeWidth` method returns a `DimAbsolute`, right? So, if `AutoSize` is `false` the `ValidatePosDim (width, value)` is not needed anymore? 

I think I follow you and that you are right:  If we follow this approach it should be possible for user to set any kind of Dim without the need to validate.

> I was thinking changing the `IsInitialized` setter to private, because it's the `Application` who handle it, but that way you will not could set it. If I can set the same object state on both, design and rutnime, that will not be a problem.

I already have a lot of Reflection to access private/internal methods (e.g. [PosExtensions](https://github.com/tznind/TerminalGuiDesigner/blob/main/src/PosExtensions.cs) so I wouldn't worry about me being able to access/not access it in TerminalGuiDesigner.  But we probably don't want to make something private if it is already released as public (without a good reason) since that is a breaking change.

BDisp:
> ```cs
>           if (AutoSize)
>           {
>               return CalculateAutoSizeWidth();
>           }
> ```

This is also a trap if almost all the code use the `width` field to get his value instead of using the `Width` property, which return different value if `AutoSize` is true. What the best workaround for this?

BDisp:
@tig I almost have all this ready. Even when auto-size is enabled the `Width` and `Height` properties are used, if the text length is lower. The `Bounds` are only changed if the text grows. If the text is lower the `Width` and `Height` are used. But I'm dealing with a problem that is the `Width` and `Height` are being set on the `View.Initialize`, even if they aren't passed to the constructor. In this case I can't manage the values that was origin passed or not by the user if the `TextDirection` is changed. So my proposal is the `Width` and `Height` be null at the `View` constructor and only be set if the user explicitly set them. So, all the magical feature of the enabled auto-size is handled only by the `Bounds` or `Frame`. @tig, @tznind please say your opinion about this.

BDisp:
I'm also thinking about reverting throwing exceptions on negative values for the `Rect.Width/Rect.Height` and for the `Size.Width/Size.Height`, because it's possible to have negative values for the cases of horizontally and vertically inverted frames.

Terminal.Gui/Core/View.cs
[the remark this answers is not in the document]
BDisp: I end up to use the @tznind suggestion. If the user set the view `Width` equal to 10 and the `Text` length is smaller than 10, then the `Width` of 10 will be used. If `AutoSize` is true and the text grows above 10 the size will be adjust to the new auto-size length but the `Width` of 10 will always be kept. If the user doesn't set the view `Width`, so is null and the `Text` length is equal to 10, then the `Width` will be set to 10 in the `OnAdded` method. If `AutoSize` is true and the text grows above 10 the size will be adjust to the new auto-size length but the `Width` of 10 will always be kept. It's vital to know at design time if was the user who set the `Pos (X,Y)` and `Dim (Width,Height)` at object initializers. It can't be in the `Initialize` method of the base class nor on the `Initialize` method of the derived class. Object initializers I meant with objects that are set after the constructor execution, like this:
```cs
var view = new View() {
    TextDirection = TextDirection.TopBottom_LeftRight,
    AutoSize = true
}
```
If the `Pos/Dim` were set on the constructor initialization without the user action, with a text length equal to 10 the `Width` will also be sets. When the `TextDirection` is set with the old code the `Width` will keep 10 and the `Height` will be 10. This is wrong because the user didn't set the `Width` equal to 10. With the new code, the `Width` will be equal to auto-size length and the `Height` will be equal to 10. I wanted something that giving the same object state at design time and runtime. The `IsInitialized` only occur on `Application.Begin`, a little later and the `OnAdded` method occur early. The better solution will be some magical event will be fired on the base class when all the objects initializers are done on the derived class. But I think there is nothing that can handle that. The only care we must have is by views that hold views child's where the `Pos/Dim` aren't yet set, like the `Border` and `PanelView`, but they was fixed.

Terminal.Gui/Core/TextFormatter.cs
```diff
@@ -154,9 +154,15 @@ public class TextFormatter {
 		/// </summary>
 		public bool AutoSize { get; set; }
 
+		/// <summary>
```
tig, line 157:
```
Gets or sets a flag that determines whether <see cref="Text"/> will have trailing spaces preserved or not when <see cref="WordWrap"/> is enbled. If `true` any trailing spaces will be trimmed when either the <see cref="Text"/> property is changed or when <see cref="WordWrap"/> is set to `true`. The default is `false`.
```
This makes it clear to the user a) that this controls something that changes the contents of the control, and b) when it does so.

Terminal.Gui/Core/View.cs
```diff
@@ -181,15 +181,30 @@ internal enum Direction {
 		/// </summary>
 		public event Action<Key> HotKeyChanged;
 
+		Key hotKey = Key.Null;
```
tig, line 184: Why is the is `HotKey` stuff part of THIS PR? This PR is titled 'AutoSize is true by default on CheckBox`. I don't understand what hotkeys have to do with that. This PR changes 38 files!

Please try not to conflate multiple, separate/independent fixes into one PR. It will make it much harder to identify which PR did what in the future.

Terminal.Gui/Core/View.cs
```diff
@@ -2303,28 +2391,20 @@ void CollectAll (View from, ref HashSet<View> nNodes, ref HashSet<(View, View)>
...
-					new Size (Math.Max (frame.Width, prevSize.Width), Math.Max (frame.Height, prevSize.Height))));
+				text = value;
+				SetHotKey ();
+				UpdateTextFormatterText ();
+				ProcessResizeView ();
 			}
 		}
 
 		/// <summary>
 		/// Used by <see cref="Text"/> to resize the view's <see cref="Bounds"/> with the <see cref="TextFormatter.Size"/>.
```
tig, line 2404: The English of this documentation is not well-formed, and is thus hard to understand. I think it should say this:

```
Gets or sets a flag that determines whether the View will be automatically resized to fit the <see cref="Text"/>.  The default is `false`.  Set to `true` to turn on AutoSize. For <see cref="AutoSize"/> to work <see cref="Width"/> and <see cref="Height"/> must be null or a valid <see cref="LayoutStyle.Absolute"/> values. In addition, AutoSize only works with <see cref="LayoutStyle.Computed"/> layout,
```

Right?

Terminal.Gui/Core/View.cs
```diff
@@ -2334,8 +2414,22 @@ void CollectAll (View from, ref HashSet<View> nNodes, ref HashSet<(View, View)>
...
-					SetNeedsLayout ();
-					SetNeedsDisplay ();
+					UpdateTextFormatterText ();
+					ProcessResizeView ();
+				}
+			}
+		}
+
+		/// <summary>
+		/// Get or sets a flag if the wrapped text will keep the trailing spaces or
```
tig, line 2424: See my other comment where I rewrote the text fo this. (Shouldn't `<inheritdoc/>` be used?)

Terminal.Gui/Core/View.cs
```diff
@@ -2372,32 +2467,38 @@ void CollectAll (View from, ref HashSet<View> nNodes, ref HashSet<(View, View)>
...
-				isInitialized = value;
-				SetMinWidthHeight ();
-			}
-		}
+		public virtual bool IsInitialized { get; set; }
+
+		/// <summary>
+		/// Gets information if the view was already added to the superview.
+		/// </summary>
+		public bool IsAdded { get; private set; }
```
tig, line 2501: Why `IsAdded` and not `IsAddedToSuperView`?

Terminal.Gui/Core/View.cs
```diff
@@ -2501,12 +2608,108 @@ bool SetWidthHeight (Rect nBounds)
...
 				Bounds = new Rect (Bounds.X, Bounds.Y, canSizeW ? rW : Bounds.Width, canSizeH ? rH : Bounds.Height);
-				TextFormatter.Size = Bounds.Size;
+				TextFormatter.Size = GetBoundsTextFormatterSize ();
 			}
 
 			return aSize;
 		}
 
+		/// <summary>
+		/// Calculates the size to fit all text if <see cref="AutoSize"/> is true.
```
tig, line 2618: This function returns a `Size`. I think it should be named `GetAutoSize`.

Why are all 3 functions needed? `CalculateAutoSize`, `CalcuulateAutoSizeWidth`, `CalculateAutoSizeHeight`?

Why not just `Size`?

Terminal.Gui/Core/View.cs
```diff
@@ -2501,12 +2608,108 @@ bool SetWidthHeight (Rect nBounds)
...
+		{
+			return new Size (TextFormatter.Size.Width - GetHotKeySpecifierLength (),
+			    TextFormatter.Size.Height - GetHotKeySpecifierLength (false));
+		}
+
+		/// <summary>
+		/// Gets the text formatter size from a <see cref="Bounds"/> size.
+		/// </summary>
+		/// <returns>The text formatter size more the <see cref="TextFormatter.HotKeySpecifier"/> length.</returns>
+		public Size GetBoundsTextFormatterSize ()
```
tig, line 2704: Does this really need to be public?

Terminal.Gui/Core/View.cs
```diff
@@ -2501,12 +2608,108 @@ bool SetWidthHeight (Rect nBounds)
...
+				    TextFormatter.Text?.Contains (HotKeySpecifier) == true
+				    ? Math.Max (Rune.ColumnWidth (HotKeySpecifier), 0) : 0;
+			}
+		}
+
+		/// <summary>
+		/// Gets the bounds size from a <see cref="TextFormatter.Size"/>.
+		/// </summary>
+		/// <returns>The bounds size minus the <see cref="TextFormatter.HotKeySpecifier"/> length.</returns>
+		public Size GetTextFormatterBoundsSize ()
```
tig, line 2694: Does this really need to be public?

UICatalog/Scenarios/AutoSizeAndDirectionText.cs
```diff
@@ -7,18 +7,23 @@ public class AutoSizeAndDirectionText : Scenario {
 		public override void Setup ()
 		{
 			var text = "Hello World";
+			//var text = "Hello World 你";
```
tig, line 10: ?

tig (changes requested):

Terminal.Gui/Core/TextFormatter.cs
BDisp: I agree, thanks. It will looks like this:

```cs
		/// <summary>
		/// Gets or sets a flag that determines whether <see cref="Text"/> will have trailing spaces preserved
		/// or not when <see cref="WordWrap"/> is enabled. If `true` any trailing spaces will be trimmed when
		/// either the <see cref="Text"/> property is changed or when <see cref="WordWrap"/> is set to `true`.
		/// The default is `false`.
		/// </summary>
```

Terminal.Gui/Core/View.cs
BDisp: Thanks for pointing this. That not true anymore. It will look like this:

```cs
		/// <summary>
		/// Gets or sets a flag that determines whether the View will be automatically resized to fit the <see cref="Text"/>.
		/// The default is `false`. Set to `true` to turn on AutoSize. If <see cref="AutoSize"/> is `true` the <see cref="Width"/>
		/// and <see cref="Height"/> will always be used if the text size is lower. If the text size is higher the bounds will
		/// be resized to fit it.
		/// In addition, if <see cref="ForceValidatePosDim"/> is `true` the new values of <see cref="Width"/> and
		/// <see cref="Height"/> must be of the same types of the existing one to avoid breaking the <see cref="Dim"/> settings.
		/// </summary>
```

Terminal.Gui/Core/View.cs
BDisp: > See my other comment where I rewrote the text fo this. (Shouldn't `<inheritdoc/>` be used?)

I don't agree because the `Responder` class doesn't inherits the `TextFormatter` class. The user would like to see a documentation pointing to the `TextFormatter.Text` and the `TextFormatter.WordWrap`. It will be like this:


```cs
		/// <summary>
		/// Gets or sets a flag that determines whether <see cref="TextFormatter.Text"/> will have trailing spaces preserved
		/// or not when <see cref="TextFormatter.WordWrap"/> is enabled. If `true` any trailing spaces will be trimmed when
		/// either the <see cref="Text"/> property is changed or when <see cref="TextFormatter.WordWrap"/> is set to `true`.
		/// The default is `false`.
		/// </summary>
```

Terminal.Gui/Core/View.cs
BDisp: Because in this logic would also have to change the following names:

`Add` to `AddToSuperView`
`Added` to `AddedToSuperView`
`OnAdded` to `OnAddedToSuperView`

But I changed the documentation to:
```cs
		/// <summary>
		/// Gets information if the view was already added to the <see cref="SuperView"/>.
		/// </summary>
```

Terminal.Gui/Core/View.cs
BDisp: > This function returns a `Size`. I think it should be named `GetAutoSize`.

Not all the code that returns something has to have the prefix `Get`, but I agree with the suggestion and I'll rename it.



> Why are all 3 functions needed? `CalculateAutoSize`, `CalcuulateAutoSizeWidth`, `CalculateAutoSizeHeight`?
> 
> Why not just `Size`?

You are right, I forgot to delete the other two. Thanks.

Terminal.Gui/Core/View.cs
BDisp: > Does this really need to be public?

Yes, it should be public because it may be needed in some layout calculation. See the in the `ViewTests.cs`.

UICatalog/Scenarios/AutoSizeAndDirectionText.cs
BDisp: I tested with wide runes and I commented to avoid rewrite all the time, but I'll automate that.

Terminal.Gui/Core/View.cs
BDisp: I tried to create a PR with the `View` and `TextFormatter` and I get a lot of unit tests errors. To fix this it I also will end up with a big PR like this one and not worth to submit. So resolving this.

tig (approved):
This is really fantastic work. 

In the future, let's really try to create more surgical PRs that focus on specific fixes vs. big 38 file PRs. Ok?

BDisp:
> This is really fantastic work.

Thanks.



> In the future, let's really try to create more surgical PRs that focus on specific fixes vs. big 38 file PRs. Ok?

Okay, I promise I go, but believe in this case it was inevitable to have many altered files because the changes required many changes at the test units level.

PR #1824, opened 2022-06-16
Fixes #1812. AutoSize stays true with empty text.

### LINKED ISSUE #1812: Button: AutoSize gets reset to false in constructor if Text == empty
```csharp
		[Fact, AutoInitShutdown]
		public void AutoSize_Stays_True_With_EmptyText ()
		{
			var btn = new Button () {
				X = Pos.Center (),
				Y = Pos.Center (),
				AutoSize = true
			};

			var win = new Window () {
				Width = Dim.Fill (),
				Height = Dim.Fill (),
				Title = "Test Demo 你"
			};
			win.Add (btn);
			Application.Top.Add (win);

			Assert.True (btn.AutoSize);

			btn.Text = "Say Hello 你";

			Assert.True (btn.AutoSize);

			var expected = @"
┌ Test Demo 你 ──────────────┐
│      [ Say Hello 你 ]      │
";

		}
```

This new test currently fails on the first `Assert.True (btn.AutoSize);`.

BDisp:


Terminal.Gui/Core/View.cs
```diff
@@ -568,17 +568,19 @@ void SetTabIndex ()
...
 
 				width = value;
-				if (autoSize && value.Anchor (0) != TextFormatter.Size.Width
-					- (TextFormatter.IsHorizontalDirection (TextDirection)
-					&& TextFormatter.Text.Contains (HotKeySpecifier) ? 1 : 0)) {
+
+				var isValidNewAutSize = autoSize ? IsValidAutoSizeWidth (width) : false;
+
+				if (autoSize && !isValidNewAutSize) {
+					TextFormatter.AutoSize = false;
```
tig, line 575: Same comment/question I have for #1797: Should this throw an exception instead of changing the property?

tig (changes requested):

Terminal.Gui/Core/View.cs
BDisp: Sure I'll do the changes. Thanks.

BDisp:
@tig, @tznind I'm thinking to allow the `Button` height to be greater than one, if `AutoSize` is disabled, respecting the `Height` value. What do both think about this?

tig:
> @tig, @tznind I'm thinking to allow the `Button` height to be greater than one, if `AutoSize` is disabled, respecting the `Height` value. What do both think about this?

Just so I'm clear: You are suggesting a new feature for Button - "Buttons can now be taller than one row."

What will a 2 row button look like?

What will a 3 row button look like?

What will a 4 row button look like?

BDisp:
> Will `button.TextDirection` have any impact?

For now no.



> Will `button.TextAlignment` have any impact?

Probably yes, may be like this.
imagem


> Will the "newline in button" code in UI Catalog "Buttons" behave differently than it does now?

For now is like this:
imagem


> At one point I had thoughts that Button COULD really just a superclass of Label; one that accented the label with button decorations. My thinking was we could reduce a lot of duplicate code in the library. This was why I wrote the "Labels as Buttons" scenario, which now illustrates how they DON'T currently behave the same.

For now is like this:
imagem


> I'm just asking questions here, BTW. I do not know the right thing to do. But it seems worth exploring whether we could combine Label/Button to remove a bunch of duplicate code.

I agree. What differ from each other is the `TextFormatter.Text` from the `View.Text`. See the changes to the `Checkbox` that is the same.

tig (approved):

PR #1838, opened 2022-06-26
Add example of auto scrolling TextView

tznind:
I've opened this PR to explore the best way to enable an 'auto scrolling' log style view.  The two problems I encountered implementing the scenario were:

- When updating the `Text` the cursor position is reset to 0
- There doesn't seem to be a way to scroll to easily reposition the cursor (e.g. to the first Column of the last Row) 

chat

I am a little nervous about my changes to `TextView` so am creating this as a draft.  I did think that instead of adding `void SetText(ustring value, bool resetCursorPosition=true)`  we could have a bool flag e.g. `PreserveSelectionWhenChangingText`.  Also I'm not sure if this could result in unstable behaviour if clearing text etc (could result in invalid selections).

Or if there is a way to achieve this same effect without making any changes to `TextView` that would be ideal.

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -334,6 +334,22 @@ namespace Terminal.Gui {
 		bool oldCanFocus;
 		int oldTabIndex;
 
+
+		/// <summary>
+		/// Performs the requested command.  Returns true if
+		/// the command is of a Type supported by the view
+		/// and the command was successful
+		/// </summary>
+		public bool DoCommand(Command cmd)
+		{
+			if (!CommandImplementations.ContainsKey(cmd))
+			{
+				return false;
+			}
+
+			return CommandImplementations[cmd]() ?? false;
+		}
+
 		/// <inheritdoc/>
 		public override bool CanFocus {
 			get => base.CanFocus;
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -336,9 +336,9 @@ namespace Terminal.Gui {
 
 
 		/// <summary>
-		/// Performs the requested command.  Returns true if
+		/// Performs the requested command. Returns true if
 		/// the command is supported by the view and the 
-		/// command was successful
+		/// command was successful.
 		/// </summary>
 		public bool DoCommand(Command cmd)
 		{
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -337,8 +337,8 @@ namespace Terminal.Gui {
 
 		/// <summary>
 		/// Performs the requested command.  Returns true if
-		/// the command is of a Type supported by the view
-		/// and the command was successful
+		/// the command is supported by the view and the 
+		/// command was successful
 		/// </summary>
 		public bool DoCommand(Command cmd)
 		{
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -1456,10 +1456,22 @@ namespace Terminal.Gui {
 			}
 		}
 
+
+		/// <summary>
+		/// Appends <paramref name="toAppend"/> to <see cref="Text"/> optional 
+		/// preserving the current selection
+		/// </summary>
+		/// <param name="resetCursorPosition">true to clear currently selected text (if any) and jump to 0,0.  False to retain current selection</param>
+		/// <param name="toAppend">The text to append to the end of <see cref="Text"/></param>
+		public void AppendText(ustring toAppend, bool resetCursorPosition)
+		{
+			SetText(Text + toAppend,resetCursorPosition);
+		}
+
 		/// <summary>
 		/// Updates the text displayed to the provided value.
 		/// </summary>
-		public void SetText(ustring value, bool resetCursorPosition=true)
+		private void SetText(ustring value, bool resetCursorPosition=true)
 		{
 			if (resetCursorPosition)
 			{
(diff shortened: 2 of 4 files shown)


BDisp (approved):

BDisp:
> Or if there is a way to achieve this same effect without making any changes to `TextView` that would be ideal.

Anyway there has to be any changes in `TextView`. Perhaps another option is to reposition the cursor after `ResetPosition`, but I think your solution is preferable. Like always, good stuff :-)

tznind:
Ok cool, I will stick with this approach then but leave as Draft for now as I want to add some unit tests to make sure theres no errors crept in where e.g. SetText and preserve selection results in illegal selection.

BDisp:
> Ok cool, I will stick with this approach then but leave as Draft for now as I want to add some unit tests to make sure theres no errors crept in where e.g. SetText and preserve selection results in illegal selection.

I'm convinced that the selected text preceding the text added are preserved, but more some unit tests are always better.


> Also expect preserving the selection but prepending the text (as opposed to adding after the current selection point) results in the selection changing but we can probably live with that.

As the text are always added at the end, probably the selected text also will be preserved, but if not I thinking we can live with that :-)

tznind:
So I originally made wrote this with `SetText` as public.  This lets the user preserve the selection regardless of how they are changing the Text.  However when writing tests I realized two things:

- Preserving the selection when prepending doesn't make much sense as although the selected indexes are the same the text underneath has changed
- When setting the text to something shorter or `""` it is quite hard to work out if the selected indexes are still valid (since they track row/col/length not indexes into the string directly.

So I have instead made `SetText` private and added a new method `AppendText` which has the option to preserve selection.  This still lets AutoScrollingText scenario work and avoids all of the above issues.

We can always revisit this later and even if we do make SetText public later on it is still nice enough to have the `AppendText` method there in the public API.

tznind:
I have made `MoveStartOfLine` public in this PR too.  Should we also make other methods like `MoveToEndOfLine` public for consistency?

Or another option would be to make use of the keybinding system with a method on `View` like:

```csharp
/// <summary>
/// Performs the requested command.  Returns true if
/// the command is of a Type supported by the view
/// and the command was successful
/// </summary>
public bool DoCommand(Command cmd)
{
	if (!CommandImplementations.ContainsKey(cmd))
	{
		return false;
	}

	return CommandImplementations[cmd]() ?? false;
}
```

This would give the user quite a bit more power to automate stuff in the views?

Terminal.Gui/Core/View.cs
```diff
@@ -334,6 +334,22 @@ void SetTabIndex ()
 		bool oldCanFocus;
 		int oldTabIndex;
 
+
+		/// <summary>
+		/// Performs the requested command.  Returns true if
+		/// the command is supported by the view and the 
+		/// command was successful
```
tig, line 341: Please don't use double spaces after periods.

Missing period at end.

Terminal.Gui/Views/TextView.cs
```diff
@@ -3278,6 +3302,9 @@ void MoveEndOfLine ()
 			DoNeededAction ();
 		}
 
+		/// <summary>
+		/// Moves the cursor to the start of the current line
```
tig, line 3306: Make a sentence by adding a period at end, please.

UICatalog/Scenarios/AutoScrollingChat.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.Text.RegularExpressions;
+using Terminal.Gui.Graphs;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace UICatalog.Scenarios {
+
+	[ScenarioMetadata (Name: "Auto Scrolling Text", Description: "Demonstrates async update and auto scrolling.")]
+	[ScenarioCategory ("Controls")]
+	[ScenarioCategory ("Text and Formatting")]
```
tig, line 19: Add TextView scenario category?

UICatalog/Scenarios/AutoScrollingChat.cs
```diff
@@ -0,0 +1,137 @@
...
+
+
+		public override void Setup ()
+		{
+			Win.Title = this.GetName();
+			textView = new TextView {
+				Width = Dim.Fill(),
+				Height = Dim.Fill(2),
+				ReadOnly = true
+			};
```
tig, line 36: The Scenario has a usability issue if mouse is used: There's no scroll bar. 

Similarly, there's no visual indication that I can scroll (because there's no scrollbarview).

Can you update the demo to include a `ScrollBarView`?

Related: #1838

tig (changes requested):
See comments.

Also, no matter what I do, I can't get selection to work.  The selection is cleared automatically. 

https://i.imgur.com/i8Pq1Dq.gif

Finally, what happens if wordwrap is on?

BDisp:
> Also, no matter what I do, I can't get selection to work. The selection is cleared automatically.
> 
> https://i.imgur.com/i8Pq1Dq.gif

This was already fixed by the PR #1840.
Maybe @tznind could merge the fix into this PR.

tig:
> > Also, no matter what I do, I can't get selection to work. The selection is cleared automatically.
> > https://i.imgur.com/i8Pq1Dq.gif
> 
> This was already fixed by the PR #1840. Maybe @tznind could merge the fix into this PR.

My bad for not merging main before testing his PR. I've just verified you are correct.

tznind:
Thanks for the review, theres quite a lot to respond to.  I will try to find some time in the next week or two to address the issues you raised.

tig:
@tznind I'm marking this as Draft so I don't keep checking it ;-).

PR #1843, opened 2022-06-27
Fix TableView multi selections extending to -1 indexes

tznind:
Noticed this while looking at another issue.  If you select cells with Shift+Arrow key and push the selection off the left of the table or up off the top you get index out of bounds when indexing the underlying Table.  This only affects the `MultiSelectedRegions` that are tracked and not the main selected cell (which correctly remains bounded to the valid cells)

This PR fixes the problem and adds unit tests for pushing a multiselect off the left/right/top/bottom of a TableView.

### DIFF
--- a/UnitTests/GraphViewTests.cs
+++ b/UnitTests/GraphViewTests.cs
@@ -52,12 +52,27 @@ namespace Terminal.Gui.Views {
 
 	public class GraphViewTests {
 
+		static string LastInitFakeDriver;
 
 		public static FakeDriver InitFakeDriver ()
 		{
 			var driver = new FakeDriver ();
-			Application.Init (driver, new FakeMainLoop (() => FakeConsole.ReadKey (true)));
+			try {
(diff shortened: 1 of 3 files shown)


UnitTests/GraphViewTests.cs
```diff
@@ -52,12 +52,27 @@ public override void DrawAxisLabel (GraphView graph, int screenPosition, string
 
 	public class GraphViewTests {
 
+		static string LastInitFakeDriver;
 
 		public static FakeDriver InitFakeDriver ()
 		{
 			var driver = new FakeDriver ();
-			Application.Init (driver, new FakeMainLoop (() => FakeConsole.ReadKey (true)));
+			try {
```
tig, line 60: I dig all this. At some point I'd like to extract these test-helper-code things out of `GraphViewTests` into a test library.

tig (approved):

PR #1857, opened 2022-07-18
Fixes #1855. Window and Frame content view without the margin frame.

### LINKED ISSUE #1855: Window always has padding because of ContentView
**Describe the bug**
The Window class contains a ContentView, referenced as 'Client'. This view is positioned at X = Y = 1. 

This means that even if you set Window.Border.BorderThickness to 0, there's still a 1 row/column pad. 

I think this can be worked around by changing dimensions of Client, but it's annoying. 

This may just be a doc issue. I'm filing it so I don't forget.

BDisp:
window-frame-margin

### DIFF
--- a/Terminal.Gui/Core/Border.cs
+++ b/Terminal.Gui/Core/Border.cs
@@ -165,7 +165,6 @@ namespace Terminal.Gui {
 			{
 				ColorScheme = Colors.TopLevel;
 				Text = title ?? "";
-				Frame = frame;
 				if (border == null) {
 					Border = new Border () {
 						BorderStyle = BorderStyle.Single,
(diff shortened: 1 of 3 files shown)


tig:
This is still drawing the `Title`. Run `WizardAsView` to see what I mean.

image

BDisp:
While DrawMarginFrame is true the title is always showing.

tig:
> While DrawMarginFrame is true the title is always showing.

I think that's a bug. I think the Title is part of the "MarginFrame".

BDisp:
> I think that's a bug. I think the Title is part of the "MarginFrame".

Of course it's. If the DrawMarginFrame is false the title won't be showing up. On the other hand if the Text isn't null it will always showing because only three views has control to this.

PR #1877, opened 2022-07-21
Fixes #1867. Use Undo and Redo commands with WordWrap enabled.

### LINKED ISSUE #1867: Use Undo and Redo commands with WordWrap enabled
No, it's just a feature, which in this case would be to be able to use Undo and Redo commands with WordWrap activated in a TextView.

BDisp:


tig (approved):
Nice.

Terminal.Gui/Views/TextView.cs
```diff
@@ -2211,12 +2223,19 @@ void ResetContinuousFind ()
...
 
+		string currentCaller;
+
 		/// <summary>
 		/// Restore from original model.
 		/// </summary>
-		void SetWrapModel ()
+		void SetWrapModel ([CallerMemberName] string caller = null)
 		{
+			if (currentCaller != null)
```
tznind, line 2233: I'm confused won't this always be true and therefore change this method to do nothing? ( due to the `[CallerMemberName]` attribute).

BDisp: No, it isn't always true. At compile time all the calls to the `SetWrapModel` and `UpdateWrapModel` will be replaced by the method name where it's calling due to the `[CallerMemberName]` attribute. For example, in the `DeleteCharLeft` method the call to the `SetWrapModel ();` will be replaced by `SetWrapModel ("DeleteCharLeft");` and the call to the `UpdateWrapModel ();` will be replaced by `UpdateWrapModel ("DeleteCharLeft");`

tznind: But your check says if it is **not null** return...

tznind: Oh sorry I misread that as `if (caller != null)` not `currentCaller` sorry.

BDisp: Ah that is intentional, because I only want the `SetWrapModel` run once if `currentCaller` is null and only run the `UpdateWrapModel` if the caller method are the same.

BDisp: This is to avoid to unwrap again if it's already unwrapped.

tznind: That makes sense, I just misread it 🚀

Issue #1820, opened 2022-06-16
TextView: Word Wrapping is not correct


tig:
```csharp

var tv = new TextView();
tv.WordWrap = true;
tv.ReadOnly = true;
tv.Text = "This is the help text for the Second Step.\n\nPress the button demo changing the Title.\n\nIf First Name is empty the step will prevent moving to the next step.";
```

<img width="183" alt="image" src="https://user-images.githubusercontent.com/585482/173963744-ce45e6a1-be65-4662-b405-af987e800930.png">

The spaces at the beginning of the lines in the third sentence should not be there.

BDisp:
`TextView` preserve trailing spaces.

tig:
> `TextView` preserve trailing spaces.

I had forgotten all about that...

Changing `TextView` to set that to false, fixes it:

<img width="402" alt="image" src="https://user-images.githubusercontent.com/585482/173965401-14d1adf5-04f8-45b8-8580-ec12d4e8c69d.png">

<img width="196" alt="image" src="https://user-images.githubusercontent.com/585482/173965292-24fe3bcf-e7c0-4c3f-b848-98a204be1fb7.png">

Can anyone think of a reason why this should not be fixed in `TextView`?

BDisp:
You must ensure that there are blank spaces at the end and if it does not fit the end, it should be placed at the beginning of the next line.

tig:
> You must ensure that there are blank spaces at the end and if it does not fit the end, it should be placed at the beginning of the next line.

Sorry, I'm not sure what you mean by this.

BDisp:
> Sorry, I'm not sure what you mean by this.

In a `Label` the spaces are removed in the end of the line but in the `TextView` it's need to copy and past. So the spaces are necessary.

tig:
I still don't understand.

My question is simple: If we change that single line of code in `TextView` such that `preseveTrailingSpces` is false instead of true, will it break anything else?

BDisp:
> My question is simple: If we change that single line of code in `TextView` such that `preseveTrailingSpces` is false instead of true, will it break anything else?

Yes, because if you select two lines the last word of the first line will be joined with the first word of the next line. Yes, because if you select two lines the last word of the first line will be joined with the first word of the next line.

BDisp:
[...]

PR #1886, opened 2022-07-24
Fixes #1867. TextView: Fixes wordwrap on LoadFile and fixes extra ending space when deleting backwards.

### LINKED ISSUE #1867: Use Undo and Redo commands with WordWrap enabled

BDisp:
Fixes #1867.


UnitTests/TextViewTests.cs
```diff
@@ -1962,19 +1968,74 @@ public void WordWrap_WrapModel_Output ()
...
-This is 
-the 
-first 
-line.
-This is 
-the 
-second 
+This is
+the first
+ line.
```
tig, line 1973: Per #1820, we all agree that TextView is not RENDERING word-wrapped text correctly. The correct rendering is in the deleted text here where the space was removed from the ` line.` line.

Right?

So, once the RENDERING behavior is fixed, this unit test will need to change.

I'd like you to rename these unit test methods to be clear on what is actually being tested.

In this case, the unit test is testing RENDERING.

There are (should be?) other unit tests that test that the INTERNAL REPRESENTATION of the `Text` is correct.

tig (changes requested):

UnitTests/TextViewTests.cs
BDisp: The space is removed after set ReadOnly to true. But I'm seeing you didn't push all the code changes that I submitted to your pr branch. I'll submit a new pr to the gui-cs.

BDisp:
Merging one of them will also merging the other.

UnitTests/TextViewTests.cs
tig: > The space is removed after set ReadOnly to true. But I'm seeing you didn't push all the code changes that I submitted to your pr branch. I'll submit a new pr to the gui-cs.

You are not understanding me. 

The `ReadOnly` fix is a temporary HACK that will fix the way TextView *RENDERS* when WordWrap is enabled.

This DOES NOT FINALLY FIX #1820.  #1820 will ONLY be fixed correctly once we do a massive change to TextView that separates RENDERING of `Text` from how `Text` is stored internally. 

In THIS PR (#1867) I am requesting that you change the unit test NAMES to make it clear that what they are testing is just the RENDERING.

tig (changes requested):

UnitTests/TextViewTests.cs
BDisp: @tig I hope have done what you asked. I'm a bit confused :-)

BDisp:
@tig, is there any problem with this pr?

PR #1918, opened 2022-07-28
Fixes #1917. Sometimes Clipboard.IsSupported doesn't return the correct value.

### LINKED ISSUE #1917: Sometimes the Clipboard.IsSupported property doesn't return the correct value.
The culprit is the clipboard hasn't been totally initialized and thus the `Clipboard.IsSupported` must always get from the `Application.Driver.Clipboard.IsSupported`.

BDisp:


tig (approved):
Thank you. I assume this will fix the intermittent unit test run failures related to the clipboard?

PR #1929, opened 2022-07-30
Fixes #1925. Preserve trailing spaces on word wrap must be refactored…

### LINKED ISSUE #1925: Preserve trailing spaces on word wrap must be refactored on TextFormatter.
But you need to agree that there are cases where a line starting with a space is inevitable.

```cs
			maxWidth = 8;
			expectedClippedWidth = 8;
			wrappedLines = TextFormatter.WordWrap (text, maxWidth, true);
			Assert.True (expectedClippedWidth >= wrappedLines.Max (l => l.RuneCount));
			Assert.True (expectedClippedWidth >= wrappedLines.Max (l => l.ConsoleWidth));
			Assert.Equal ("A ", wrappedLines [0].ToString ());
			Assert.Equal ("sentence", wrappedLines [1].ToString ());
			Assert.Equal (" has ", wrappedLines [2].ToString ());
			Assert.Equal ("words.", wrappedLines [^1].ToString ());
			Assert.True (wrappedLines.Count == 4);

```

The `sentence` word has already a width equal to 8 and can't moved to the next line, causing an infinite loop :-) and thus the `has` word must start with the existing space.
I also will remove from the `TextView` all the references to the `preserveTrailingSpaces == !ReadOnly` and the methods added to tweak this, because it must always have to be true to allow track all the necessary methods. So, @tig I hope you understanding this, please.

BDisp:
Thanks to @tig for insisting on this.

### DIFF
--- a/Terminal.Gui/Core/TextFormatter.cs
+++ b/Terminal.Gui/Core/TextFormatter.cs
@@ -515,6 +515,7 @@ namespace Terminal.Gui {
 
 			int GetNextWhiteSpace (int from, int cWidth, out bool incomplete, int cLength = 0)
 			{
+				var lastFrom = from;
 				var to = from;
 				var length = cLength;
 				incomplete = false;
@@ -552,8 +553,10 @@ namespace Terminal.Gui {
 					}
 					to++;
 				}
-				if (cLength > 0 && to < runes.Count && runes [to] != ' ') {
+				if (cLength > 0 && to < runes.Count && runes [to] != ' ' && runes [to] != '\t') {
 					return from;
+				} else if (cLength > 0 && to < runes.Count && (runes [to] == ' ' || runes [to] == '\t')) {
+					return lastFrom;
 				} else {
 					return to;
 				}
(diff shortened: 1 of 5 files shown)


tig:
@BDisp are you ok with me merging this now so that it will go in v1.7.2 that I plan on releasing today?

tig (approved):

BDisp:
> @BDisp are you ok with me merging this now so that it will go in v1.7.2 that I plan on releasing today?

Of course I'm with you. I thought you don't agree with these changes. In `TextView`, the best solution is not to mix preservation spaces on the right with not preserving, because controlling the selected text is not possible in all situations.

tig:
The immeidate issue I had has been resolved via `ReadOnly == true && WordWrap == true`. 

I think we should consider a re-factor/rewrite of all text handling as part of v2.0 instead of continuing to hack at what's in there.

Are you ok with me closing this PR and opening a new Issue where we can focus on that?

BDisp:
> The immeidate issue I had has been resolved via ReadOnly == true && WordWrap == true.

But with this pr you have the same behavior like you did, with the difference keep trailing space, to allowing consistence with the selected text no mater ReadOnly is enabled or not with the WordWrap enabled.

> I think we should consider a re-factor/rewrite of all text handling as part of v2.0 instead of continuing to hack at what's in there.

I agree with a re-factor/rewrite as part part of v2.0 in the way you want on the TextView, by hacking. This pr has changes that correct some bug and preserve trailing spaces more correctly.

> Are you ok with me closing this PR and opening a new Issue where we can focus on that?

Not totally because this PR has important changes to the word wrap API when preserve trailing spaces is enabled.

PR #1964, opened 2022-08-25
Fixes #1963. Only remove one character on backspace when wordwrap is on

### LINKED ISSUE #1963: Backspace key deletes multiple characters when editing a TextView with WordWrap = true
While editing a TextView that has wordwrap turned on. Pressing the backspace key anywhere other than the last line in the window will delete 2 characters. If the backspace key deletes the last line in the file, an out of range exception will be generated trying to remove a non existent line.

1. Run the UI Catalog
2. Select Dialogs -> Editor
3. Enable word wrap in the Format menu
4. Move the cursor into the middle of the first 'hello'
5. Press backspace

**Expected behavior**
only one character should be deleted

hippiehunter:
Fixes #1963 check if word wrap is enabled before removing the character from the model. If its enabled, use the wrap manager instead.

### DIFF
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -3760,9 +3760,12 @@ namespace Terminal.Gui {
 
 				historyText.Add (new List<List<Rune>> () { new List<Rune> (currentLine) }, CursorPosition);
 
-				currentLine.RemoveAt (currentColumn - 1);
-				if (wordWrap && wrapManager.RemoveAt (currentRow, currentColumn - 1)) {
-					wrapNeeded = true;
+				if (wordWrap) {
+					if (wrapManager.RemoveAt (currentRow, currentColumn - 1)) {
+						wrapNeeded = true;
+					}
+				} else {
+					currentLine.RemoveAt (currentColumn - 1);
 				}
 				currentColumn--;
 
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -1613,12 +1613,7 @@ namespace Terminal.Gui {
 					return ustring.Empty;
 				}
 
-				SetWrapModel ();
-				var sel = GetRegion ();
-				UpdateWrapModel ();
-				Adjust ();
-
-				return sel;
+				return GetSelectedRegion ();
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


Terminal.Gui/Views/TextView.cs
```diff
@@ -3760,9 +3760,12 @@ bool DeleteTextBackwards ()
 
 				historyText.Add (new List<List<Rune>> () { new List<Rune> (currentLine) }, CursorPosition);
 
-				currentLine.RemoveAt (currentColumn - 1);
-				if (wordWrap && wrapManager.RemoveAt (currentRow, currentColumn - 1)) {
-					wrapNeeded = true;
+				if (wordWrap) {
```
tig, line 3763: Please hit CTRL-K, CTRL-D in VS2022 to reformat the source to match the coding standards.

tig (changes requested):
Excellent find and fix. Thank you.

Once you fix the source formatting I'll merge.

Terminal.Gui/Views/TextView.cs
hippiehunter: I applied the reformat but this is the result. Is there something specific that isn't matching?

tig: My bad. I got caught off by how the diff was showing. Sorry about that.

tig (approved):

tig:
@hippiehunter, per @BDisp can you please also fix the Delete key use-case as part of this PR?

Many thanks for helping with this!

BDisp:
@tig this bug is much more wide than this pr is intended for. Because now all the actions on the `TextView` unwrap and re-wrap, if word wrap is enabled, the management of the wrapped lines, like remove, add, is not needed anymore. So, this pr will not fix all the issues. Do you want I submit a pr to the @hippiehunter's repo or should I submit a new pr?

BDisp:
@tig here is my fix https://github.com/BDisp/Terminal.Gui/tree/textview-word-wrap-delete-backwards-forwards-fix. @hippiehunter can merge into his pr or if you want I can create a new pr with my branch. Thanks.

BDisp (approved):

hippiehunter:
Is there anything else you need me to do on this PR?

BDisp:
> Is there anything else you need me to do on this PR?

For me, since you already merged my fix I think is all OK and I already approved. I appreciate your help for found this issue, but the fix implied more some changes. The management for remove and add runes on wrapped lines aren't need anymore and that why it was deleting twice when word wrap is enabled We must wait for @tig to merge or saying something. Thanks.

PR #1966, opened 2022-08-25
Fixes #1962 - Change KeyBindings to allow chaining commands

### LINKED ISSUE #1962: Support "Auto Move To Next Item" in ListView
`ListView` supports multi-select; SPACE toggles (@tznind I don't see a sample of `TableView` demonstrating multi-select).

In `Out-ConsoleGridView` a user has requested that when in multi-select mode (`-OutputMode Multiple`) pressing SPACE to select will toggle the item and move to the next row: https://github.com/PowerShell/GraphicalTools/issues/148

I don't see how to implement this in a generic/extensible way.

Ideally, I'd be able to (in `Out-ConsoleGridView`) define a new `Command` be added named `ToggleCheckedWithKey`:

```csharp
...
            if (!_applicationData.MoveToNextOnSelect)
            {
                var cmd = new Command();
                _listView.AddCommand(cmd, () => { if (_listView.MarkUnmarkRow ()) return _listView.MoveDown (); else return false;  });
                _listView.AddKeyBinding(Key.Space, cmd);
            }

            win.Add(_listView);
...
```

But `Command` is not `public`, so this won't work. 

We could modify `ListView`:

```csharp

// change this:
			AddCommand (Command.ToggleChecked, () => MarkUnmarkRow ());

// to this:

			AddCommand (Command.ToggleChecked, () => {
				if (MarkUnmarkRow ()) {
					if (_moveNextOnToggleWithKey) {
						return MoveDown ();
					}
					return true;
				} else {
					return false;
				}
			});
```

But this means that if `Command.ToggleChecked` is invoked in a way OTHER than the SPACE key the `MoveDown` will happen, which is not what we want. Also, we have all this `Command` and `KeyBinding` functionality, and it seems crazy to have to add yet-another-option to `ListView`.

What are the best ideas for addressing this?

tznind:
Fixes #1962 - Allow specifying more than 1 `Command` to order when using `AddKeyBinding`

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1751,11 +1751,19 @@ namespace Terminal.Gui {
 		/// </para>
 		/// <para>If the key is already bound to a different <see cref="Command"/> it will be
 		/// rebound to this one</para>
+		/// <remarks>Commands are only ever applied to the current <see cref="View"/>(i.e. this feature
+		/// cannot be used to switch focus to another view and perform multiple commands there)</remarks>
 		/// </summary>
 		/// <param name="key"></param>
-		/// <param name="command"></param>
+		/// <param name="command">The command(s) to run on the <see cref="View"/> when <paramref name="key"/> is pressed.
+		/// When specifying multiple, all commands will be applied in sequence.  The bound <paramref name="key"/> strike
+		/// will be consumed if any took effect.</param>
 		public void AddKeyBinding (Key key, params Command [] command)
 		{
+			if (command.Length == 0) {
+				throw new ArgumentException ("At least one command must be specified", nameof (command));
+			}
+
 			if (KeyBindings.ContainsKey (key)) {
 				KeyBindings [key] = command;
 			} else {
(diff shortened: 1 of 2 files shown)


tznind:
Needs more tests but I think that this is what 'command chaining' would look like.  

I've changed `AddKeyBinding` from taking `Command` to taking `params Command[]` rather than adding overloads.  This means all existing code written will be compatible but might not be binary compatible (e.g. dll binding redirects to new version with code compiled against old version)

It also means a user could specify no `Command` objects at all (params are optional) which would be wierd... probably needs a test but isn't going to break the world.

BDisp:
> Needs more tests (hence Draft) but I think that this is what 'command chaining' would look like.
> 
I like your solution for this.

> I've changed `AddKeyBinding` from taking `Command` to taking `params Command[]` rather than adding overloads. This means all existing code written will be compatible but might not be binary compatible (e.g. dll binding redirects to new version with code compiled against old version)
> 
No big deal, compiling again solves it.

> It also means a user could specify no `Command` objects at all (params are optional) which would be wierd... probably needs a test but isn't going to break the world.

I think is better to throw an exception if the command is null or was not passed no command at all.

PR #1970, opened 2022-08-26
Fixes #1969. ESC on CursesDriver take to long to being processed.

### LINKED ISSUE #1969: ESC on CursesDriver take to long to being processed.

BDisp:


tznind:
> Note that xterm mouse events are built up from character sequences received from the xterm. If your application makes heavy use of multiple-clicking, you may wish to lengthen this default value because the timeout applies to the composed multi-click event as well as the individual clicks.

Not at my Linux comp so can't check quickly but we should make sure this setting doesn't affect multiple-clicking (i.e. double clicking)

Does setting an environment variable for `ESCDELAY` still override this setting or will the new code force it to 1ms?

BDisp:
I read that but it isn't impact the mouse click and double click. We can't rely on the environment variable ESCDELAY  because it can be overridden by another app. So, is preferable to use the set_escdelay method. But I agree we must test on more Linux system.

BDisp:
@tznind, please see this https://invisible-island.net/ncurses/man/curs_threads.3x.html:

> The ESCDELAY and TABSIZE global variables are modified by some applications. To modify them in any configuration, use  the set_escdelay or set_tabsize functions. Other global variables are not modifiable.

tig (approved):

PR #1971, opened 2022-08-27
Fixes #1800. TextView now uses the same colors as TextField.

### LINKED ISSUE #1800: TextView: Default background color has usability issues
In Wizards Scenario, I have what is essentially a modal dialog box with a TextView filling the client area:

image

Note how it's actually very hard to tell that this is a non-read-only/editible TextView. 

I think `TextView` should use the same colors as `TextField`:

image

BDisp:
Fixes #1800.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -1936,7 +1936,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Sets the driver to the default color for the control where no text is being rendered.  Defaults to <see cref="ColorScheme.Normal"/>.
 		/// </summary>
-		protected virtual void ColorNormal ()
+		protected virtual void SetNormalColor ()
 		{
 			Driver.SetAttribute (GetNormalColor ());
 		}
@@ -1948,7 +1948,7 @@ namespace Terminal.Gui {
 		/// </summary>
 		/// <param name="line"></param>
 		/// <param name="idx"></param>
-		protected virtual void ColorNormal (List<Rune> line, int idx)
+		protected virtual void SetNormalColor (List<Rune> line, int idx)
 		{
 			Driver.SetAttribute (GetNormalColor ());
 		}
@@ -1960,7 +1960,7 @@ namespace Terminal.Gui {
 		/// </summary>
 		/// <param name="line"></param>
 		/// <param name="idx"></param>
-		protected virtual void ColorSelection (List<Rune> line, int idx)
+		protected virtual void SetSelectionColor (List<Rune> line, int idx)
 		{
 			Driver.SetAttribute (new Attribute (ColorScheme.Focus.Background, ColorScheme.Focus.Foreground));
 		}
@@ -1972,7 +1972,7 @@ namespace Terminal.Gui {
 		/// </summary>
 		/// <param name="line"></param>
 		/// <param name="idx"></param>
-		protected virtual void ColorReadOnly (List<Rune> line, int idx)
+		protected virtual void SetReadOnlyColor (List<Rune> line, int idx)
 		{
 			Attribute attribute;
 			if (ColorScheme.Disabled.Foreground == ColorScheme.Focus.Background) {
@@ -1990,7 +1990,7 @@ namespace Terminal.Gui {
 		/// </summary>
 		/// <param name="line"></param>
 		/// <param name="idx"></param>
-		protected virtual void ColorUsed (List<Rune> line, int idx)
+		protected virtual void SetUsedColor (List<Rune> line, int idx)
 		{
 			Driver.SetAttribute (ColorScheme.HotFocus);
 		}
@@ -2382,7 +2382,7 @@ namespace Terminal.Gui {
 		///<inheritdoc/>
 		public override void Redraw (Rect bounds)
 		{
-			ColorNormal ();
+			SetNormalColor ();
 
 			var offB = OffSetBackground ();
 			int right = Frame.Width + offB.width + RightOffset;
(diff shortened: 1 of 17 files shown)


BDisp:
This commit https://github.com/gui-cs/Terminal.Gui/pull/1971/commits/3308ab4d1054a45f3a7479cadaf2e20ffb2cd0cc will be needed after the pr #1972 is merged.

Terminal.Gui/Views/TextView.cs
```diff
@@ -1962,7 +1962,25 @@ protected virtual void ColorNormal (List<Rune> line, int idx)
...
+		}
+
+		/// <summary>
+		/// Sets the <see cref="View.Driver"/> to an appropriate color for rendering the given <paramref name="idx"/> of the
+		/// current <paramref name="line"/>.  Override to provide custom coloring by calling <see cref="ConsoleDriver.SetAttribute(Attribute)"/>
+		/// Defaults to <see cref="ColorScheme.Focus"/>.
+		/// </summary>
+		/// <param name="line"></param>
+		/// <param name="idx"></param>
+		protected virtual void ColorReadOnly (List<Rune> line, int idx)
```
tig, line 1975: Should this be named `SetReadOnlyColor` instead?

Maybe since they are all currently `protected` we use this PR to fix all these methods to use correct naming?

tig, line 1975: Why does this take a `List<Rune> line` and `int idx` pram, if they are not used?

Terminal.Gui/Windows/Wizard.cs
```diff
@@ -203,11 +203,6 @@ public WizardStep (ustring title)
 
 				base.Add (contentView);
 
-				helpTextView.ColorScheme = new ColorScheme () {
```
tig, line 206: Why is this change part of this PR?

UICatalog/Scenarios/Wizards.cs
```diff
@@ -217,6 +217,7 @@ void Top_Loaded ()
 						Height = Dim.Fill (1),
 						WordWrap = true,
 						AllowsTab = false,
+						ColorScheme = Colors.Base
```
tig, line 220: I think changes to Wizard should be in a separate PR.

tig (changes requested):
Please see my comments.

Terminal.Gui/Views/TextView.cs
BDisp: I think @tznind only provide them for overridden them, see the `SyntaxHighlighting` scenario.

BDisp: I agree. I will change them.

Terminal.Gui/Windows/Wizard.cs
BDisp: Because it only have the appropriate effect with this PR.

UICatalog/Scenarios/Wizards.cs
BDisp: Because it only have the appropriate effect with this PR. But if you insist I'll do a new PR with this PR merged. I'll wait for your feedback.

Terminal.Gui/Windows/Wizard.cs
BDisp: Yes it does, because the Wizard use the `TextView` and you changed the `ColorScheme` to looks better with the old `TextView` `ColorScheme`. With this PR, the changes some of their colors will affect how the `Wizards` scenario will display with this PR.

Terminal.Gui/Views/TextView.cs
tznind: So this method is declared on the `TextView` and is part of the render loop.  It is called for each index in the line.  It allows the user to alter the color set on the `Driver` based on arbitrary requirements (e.g. syntax highlighting).  

So for this reason it is not called `SetNormalColor` because it doesn't change the Normal color of the view.  It only changes the Color that is being used for that part of the current line as it is being rendered. 

I guess you could call it `SetDriverToNormalColor` though? 

It was implemented in this PR: https://github.com/gui-cs/Terminal.Gui/pull/1336

BDisp: I think the changes @tig requested and which I already did is enough. The explanation for the `line` and `idx` parameters not being used but handled on overridden methods are clarified.

BDisp: Done.

Terminal.Gui/Windows/Wizard.cs
tig: Please answer this question: 

Did you make these changes to fix Wizard, or to fix a broader problem in TextView and TextField?

BDisp: @tig I did to fix `Wizard` after fixing a broader problem in `TextView` and `TextField`.

tig: > @tig I did to fix `Wizard` after fixing a broader problem in `TextView` and `TextField`.

I'm still not getting my thoughts to you clearly. My apologies.

I do not think this PR should include changes to Wizard. I think it should be two PRs, fixing two Issues:

1) TextView/TextField use different colors.
2) Wizard no longer needs to set specific colors, because 1) has been fixed.

BDisp: > 2\. Wizard no longer needs to set specific colors, because 1) has been fixed.

Done #2012 teacher @tig :-)

tig (approved):

arakis:
What is the "right" way to get the old colors / Behavior?

Old Behavior: Blue Background, Gray Text.
Now Gray Background, Darkgray Text.

tznind:
> Old Behavior: Blue Background, Gray Text. Now Gray Background, Darkgray Text.
> 
> In my case, I have a border/panel around the view, it looks much more natural to have the background blue.

You can set the colors by creating a `ColorScheme` and assigning it to the `View`.  Make sure to only declare after you have called `Application.Init ();`

For example:
```csharp
using Terminal.Gui;
using Attribute = Terminal.Gui.Attribute;

Application.Init();
var win = new Window("Example App (Ctrl+Q to quit)");

ColorScheme myscheme = new ColorScheme
{
    // Color when NOT tabbed into
    Normal = new Attribute(
        Color.Red,Color.Black),

    // Selected text color when NOT tabbed into
    HotNormal = new Attribute(
        Color.Black,Color.Red),

    // Color when tabbed into
    Focus = new Attribute(
        Color.Red,Color.Black),
    
    // Selected text color when tabbed into
    HotFocus = new Attribute(
        Color.Black,Color.Red),

    // When disabled
    Disabled = new Attribute(Color.Gray,Color.Black),
    
};
win.Add(
    new TextView()
    {
        Width = Dim.Fill(),
        Height = Dim.Fill() -1,
        ColorScheme = myscheme,
        AllowsTab = false,
    }
);
win.Add(
    new TextField()
    {
        Width = 10,
        Y = Pos.AnchorEnd(1)
    }
);

Application.Run(win);
Application.Shutdown();

```
redtext

PR #1983, opened 2022-09-03
Fixes #820. Added HideDropdownListOnClick property.

### LINKED ISSUE #820: Review ComboBox 'autoHide' design
Background: ComboBox was initially written with one use-case in mind (hosted in a dialog), @BDisp then helped extend it to a more generalized control. Of which there are now two render modes. 1. List hidden until user types or hits drop down arrow, and 2. List is always displayed. I aimed to auto-detect which mode should be used which led to the introduction of the Added() Removed() actions.

I believe the simple answer to this is to not try and auto-detect which mode to use and make the mode propety 'autoHide' public so the user can decided. And in doing so, change the name and invert the logic
<pre>
from:
    autoHide = true

to:
    AlwaysShow = false
</pre>

In summary, the complexity is not down to this being a composite control, it's down to my own making :)

BDisp:
Fixes #820 - To not break the initial auto-hide behavior I decided to create a new flag `HideDropdownListOnClick` which will hide the dropdown list on click, avoiding the double click, instead of made a `AutoHide` public property. If this property is false the current behavior is valid.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/ComboBox.cs
+++ b/Terminal.Gui/Views/ComboBox.cs
@@ -8,6 +8,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Threading.Tasks;
 using NStack;
 
 namespace Terminal.Gui {
@@ -107,6 +108,9 @@ namespace Terminal.Gui {
 			SetSource (source);
 		}
 
+		bool itemPressed;
+		bool itemClicked;
+
 		private void Initialize ()
 		{
 			if (Bounds.Height < minimumHeight && (Height == null || Height is Dim.DimAbsolute)) {
@@ -132,9 +136,45 @@ namespace Terminal.Gui {
 			};
 
 			listview.SelectedItemChanged += (ListViewItemEventArgs e) => {
-
+				if (itemClicked) {
+					itemPressed = false;
+					itemClicked = false;
+					SelectedItem = e.Item;
+					if (isShow) {
+						isShow = false;
+						HideList ();
+					}
+				}
 				if (searchset.Count > 0) {
 					SetValue (searchset [listview.SelectedItem]);
+				}
+			};
+
+			listview.MouseClick += e => {
+				if (isShow && HideDropdownListOnClick && e.MouseEvent.Flags == MouseFlags.Button1Pressed) {
+					itemPressed = true;
+				} else if (itemPressed && isShow && HideDropdownListOnClick && e.MouseEvent.Flags == MouseFlags.Button1Clicked) {
+					if (itemClicked && SelectedItem == listview.SelectedItem) {
+						DoHideList ();
+						e.Handled = true;
+					} else if (itemPressed) {
+						itemClicked = true;
+						Task.Run (async () => {
+							await Task.Delay (100);
+							if (itemClicked && SelectedItem == listview.SelectedItem) {
+								DoHideList ();
+								Application.MainLoop.Invoke (() => SetNeedsDisplay ());
+							}
+						});
+					}
+				}
+
+				void DoHideList ()
+				{
+					itemPressed = false;
+					itemClicked = false;
+					isShow = false;
+					HideList ();
 				}
 			};
 
@@ -235,6 +275,11 @@ namespace Terminal.Gui {
 				}
 			}
 		}
+
+		/// <summary>
+		/// Gets or sets if the drop-down list can be hide with a button click event.
+		/// </summary>
+		public bool HideDropdownListOnClick { get; set; }
 
 		///<inheritdoc/>
 		public override bool MouseEvent (MouseEvent me)
--- a/Terminal.Gui/Views/ComboBox.cs
+++ b/Terminal.Gui/Views/ComboBox.cs
@@ -17,6 +17,36 @@ namespace Terminal.Gui {
 	/// </summary>
 	public class ComboBox : View {
 
+		private class ComboListView : ListView {
+			public bool HideDropdownListOnClick { get; set; }
+
+			public ComboListView (bool hideDropdownListOnClick)
+			{
+				HideDropdownListOnClick = hideDropdownListOnClick;
+			}
+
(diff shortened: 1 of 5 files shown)


Terminal.Gui/Views/ComboBox.cs
```diff
@@ -132,12 +136,49 @@ private void Initialize ()
...
+				if (isShow && HideDropdownListOnClick && e.MouseEvent.Flags == MouseFlags.Button1Pressed) {
+					itemPressed = true;
+				} else if (itemPressed && isShow && HideDropdownListOnClick && e.MouseEvent.Flags == MouseFlags.Button1Clicked) {
+					if (itemClicked && SelectedItem == listview.SelectedItem) {
+						DoHideList ();
+						e.Handled = true;
+					} else if (itemPressed) {
+						itemClicked = true;
+						Task.Run (async () => {
+							await Task.Delay (100);
```
tznind, line 164: Why is there a delay in closing on select?

BDisp: That was really trickily. The listview `MouseClick` event is invoked before the listview `MouseEvent` is processed. So, I can't rely in the condition `SelectedItem == listview.SelectedItem` itself, because the listview doesn't already processed the change and that why I needed the flags `itemPressed` and `itemClicked`. If another item is selected the `SelectedItemChanged` event will be invoked resetting the flags and hiding the dropdown list. If we click on the same selected item the `SelectedItemChanged` event will not be invoked and we won't know what is the response. That is the reason for the delay to check if the `itemClicked` is true (already processed by the `SelectedItemChanged` event) or false (already processed by the listview `MouseEvent` but without notifying any response).

tznind: > The listview MouseClick event is invoked before the listview MouseEvent is processed. [...], because the listview doesn't already processed the change 

What about subclassing `ListBox` as a private class e.g. `ComboListBox`.  Then you could override the On Mouse Event method and call `base` method then the custom code to finish up?

Might be more reliable than introducing threading into the equation?

BDisp: I also though about that but I was afraid if it was breaking anything. I'm still doing more unit test because mixing mouse with keyboard doesn't behave as only with mouse. After I have all unit test to guarantee all is fine, maybe I'll do that. Converting to draft to until finishing.

BDisp: I already did what you ask, thanks for the help. Please say something that you think I have to change.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -101,7 +132,20 @@ public ComboBox (ustring text) : base ()
...
 			search = new TextField ("") { Width = rect.Width };
-			listview = new ListView (rect, source) { LayoutStyle = LayoutStyle.Computed, ColorScheme = Colors.Base };
+			listview = new ComboListView (rect, source, HideDropdownListOnClick) { LayoutStyle = LayoutStyle.Computed, ColorScheme = Colors.Base };
+
+			Initialize ();
+			SetSource (source);
+		}
+
+		/// <summary>
+		/// Initialize with the source.
```
tig, line 142: If a user chooses to NOT use this constructor, but uses the parameterless constructore and sets the `Source` property instead, will the ComboBox work the same?

Generally, parameterless constructors are preferred but this PR seems to be biased towards having people use constructors with params.

tig (changes requested):

Terminal.Gui/Views/ComboBox.cs
BDisp: > If a user chooses to NOT use this constructor, but uses the parameterless constructore and sets the `Source` property instead, will the ComboBox work the same?

Yes it will work the same without any issue because this new constructor isn't used.


> Generally, parameterless constructors are preferred but this PR seems to be biased towards having people use constructors with params.

This doesn't affect the normal execution of the view. I only created it to be consistent with `ListView(IList source)` constructor which the `ComboBox` already had but with the `Rect` parameter. So, it's obvious the this `Computed` constructor is great.

tig: > > If a user chooses to NOT use this constructor, but uses the parameterless constructore and sets the `Source` property instead, will the ComboBox work the same?
> 
> Yes it will work the same without any issue because this new constructor isn't used.

Sorry,  but my question was not as clear as I meant it:

If a user chooses to NOT use this constructor, but uses the parameterless constructor and sets the `Source` property instead, will the ComboBox work the SAME AS IF THE USER used the new constructor?

BDisp: No, it won't, because the `Source` property only accepts a `IListDataSource`. For `IList` type the `Source` property only can be set by the `SetSource` method. The new constructor is more to work directly to set the `ListView` source by using the same `IList` source parameter.

BDisp:
Current item always highlighted with the `HideDropdownListOnClick=true`.

combobox-feature

Terminal.Gui/Views/ComboBox.cs
BDisp: > No, it won't, because the `Source` property only accepts a `IListDataSource`. For `IList` type the `Source` property only can be set by the `SetSource` method.

My bad. I forgot t mention it can be set by calling the `ListWrapper` constructor directly:
`listView.Source = new ListWrapper (null);`

I prefer this approach because we can have a null source but still having a non null `Source` instance.

BDisp: > If a user chooses to NOT use this constructor, but uses the parameterless constructor and sets the `Source` property instead, will the ComboBox work the SAME AS IF THE USER used the new constructor?

Sorry for my wrong response before. Yes it will if the `Source` is initialize using the `ListWrapper` constructor but with the only difference that if it's set with the `SetSource` method with a null value the `IListDataSource` instance will not be created and using the `ListWrapper` constructor with a null parameter a `IListDataSource` instance will be created with a null source.

PR #1997, opened 2022-09-07
Fixes #1994. BREAKING CHANGE. Ensure only a single IdleHandlers list can exist.

### LINKED ISSUE #1994: MainLoop dispatcher can lose idle handlers (BREAKING CHANGE).
**Describe the bug**
The MainLoop idleHandlers variable can lose functions added to it. This is catastrophic if using async/await as you might await and never be continued

**To Reproduce**
1. Go to my fork: https://github.com/montekarlos/Terminal.Gui
2. Check out bug_demo and run the hacked StandaloneExample: https://github.com/montekarlos/Terminal.Gui/tree/develop/StandaloneExample
3. Within a few cycles of the background task it should hang. If you break you will see at least one completed Task that is stuck (nothing i s continuing it)

**Possible Fix**
1. Go to my fork: https://github.com/montekarlos/Terminal.Gui
2. Check out bug_fix and run the hacked StandaloneExample: https://github.com/montekarlos/Terminal.Gui/tree/develop/StandaloneExample
3. The background tasks should never hang now

montekarlos:
Fixes #1994 - Add unit test that demonstrates the leaking of idle handlers

### DIFF
--- a/Terminal.Gui/Core/MainLoop.cs
+++ b/Terminal.Gui/Core/MainLoop.cs
@@ -77,9 +77,15 @@ namespace Terminal.Gui {
 		public SortedList<long, Timeout> Timeouts => timeouts;
 
 		/// <summary>
-		/// Gets the list of all idle handlers.
+		/// Gets a copy of the list of all idle handlers.
 		/// </summary>
-		public ReadOnlyCollection<Func<bool>> IdleHandlers => idleHandlers.AsReadOnly();
+		public ReadOnlyCollection<Func<bool>> IdleHandlers {
+			get {
+				lock (idleHandlersLock) {
+					return new List<Func<bool>> (idleHandlers).AsReadOnly ();
+				}
+			}
+		}
 
 		/// <summary>
 		/// The current IMainLoopDriver in use.
--- a/Terminal.Gui/Core/MainLoop.cs
+++ b/Terminal.Gui/Core/MainLoop.cs
@@ -61,6 +61,11 @@ namespace Terminal.Gui {
 
 		internal SortedList<long, Timeout> timeouts = new SortedList<long, Timeout> ();
 		object timeoutsLockToken = new object ();
+
+		/// <summary>
+		/// The idle handlers and lock that must be held while manipulating them
+		/// </summary>
+		object idleHandlersLock = new object ();
 		internal List<Func<bool>> idleHandlers = new List<Func<bool>> ();
 
 		/// <summary>
@@ -73,7 +78,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Gets the list of all idle handlers.
 		/// </summary>
-		public List<Func<bool>> IdleHandlers => idleHandlers;
+		public List<Func<bool>> IdleHandlers => new List<Func<bool>>(idleHandlers);
 
 		/// <summary>
 		/// The current IMainLoopDriver in use.
@@ -123,7 +128,7 @@ namespace Terminal.Gui {
 		/// <param name="idleHandler">Token that can be used to remove the idle handler with <see cref="RemoveIdle(Func{bool})"/> .</param>
 		public Func<bool> AddIdle (Func<bool> idleHandler)
 		{
-			lock (idleHandlers) {
+			lock (idleHandlersLock) {
 				idleHandlers.Add (idleHandler);
 			}
 
@@ -139,7 +144,7 @@ namespace Terminal.Gui {
 		///  This method also returns <c>false</c> if the idle handler is not found.
 		public bool RemoveIdle (Func<bool> token)
 		{
-			lock (token)
+			lock (idleHandlersLock)
 				return idleHandlers.Remove (token);
 		}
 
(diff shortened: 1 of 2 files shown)


BDisp:
Thanks @montekarlos  for your help. I confirm this unit test fails with the current `develop` branch and don't fails with your https://github.com/montekarlos/Terminal.Gui/tree/bug_fix branch.
I think you may merge your `bug-fix` branch into this PR after fix a typo by change on the comment `thead` to `thread`.
I also recommend to add the `readonly` keyword on these two declarations:

`readonly object timeoutsLockToken = new object ();`
`readonly object idleHandlersLock = new object ();`

UnitTests/MainLoopTests.cs
```diff
@@ -505,5 +506,77 @@ public void Wakeup ()
 		// - wait = false
 
 		// TODO: Add IMainLoop tests
+
+		volatile static int tbCounter = 0;
+		volatile static int counter = 0;
+		volatile static int waiting = 0;
+		volatile static int completed = 0;
+		static void Launch (Random r, TextField tf)
```
BDisp, line 514: Prefix with the `private` keyword, please.

UnitTests/MainLoopTests.cs
```diff
@@ -505,5 +506,77 @@ public void Wakeup ()
...
+		{
+			Thread.Sleep (r.Next (2, 4));
+			Application.MainLoop.Invoke (() => {
+				tf.Text = $"index{r.Next ()}";
+				Interlocked.Increment (ref tbCounter);
+			});
+		}
+
+		public void RunTest (Random r, TextField tf, int numPasses, int numIncrements, int pollMs)
```
BDisp, line 523: Add the `static` keyword after `private` keyword, please.

BDisp (changes requested):

montekarlos:
Ok I've tidied up the unit test and fix

Terminal.Gui/Core/MainLoop.cs
```diff
@@ -73,7 +78,7 @@ public sealed class Timeout {
 		/// <summary>
 		/// Gets the list of all idle handlers.
 		/// </summary>
-		public List<Func<bool>> IdleHandlers => idleHandlers;
+		public List<Func<bool>> IdleHandlers => new List<Func<bool>>(idleHandlers);
```
tznind, line 81: `IdleHandlers` is a public property that was introduced in 68b40a93f11335651ee8877f812ee27b0b9f64f1.

This PR changes the property to return a new list every time which means it is now effectively immutable (which is probably a good thing).  But it doesn't look immutable to an API user.  Should we change this to:

```csharp
/// <summary>
/// Gets the list of all idle handlers.
/// </summary>
public ReadOnlyCollection<Func<bool>> IdleHandlers => idleHandlers.AsReadOnly();
```

Its a breaking change but quite minor one and to a new bit of code recently introduced.  Also I think this is only going to be used by WebConsole so far.

BDisp: I agree. It must be read only and return the idlehandlers field and not a new list.

BDisp: How about write a unit test trying changing the public `IdleHandlers` field to be sure that the private `idleHandlers` field doesn't changes at all outside the Terminal.Gui namespace?.

montekarlos: I've made that changes as suggested by tznind - that makes sense to me.

BDisp: I'm not sure how we would create a unit test for this as it's now a compile time error to attempt to change IdleHandlers. Could you elaborate further?

BDisp: > BDisp: I'm not sure how we would create a unit test for this as it's now a compile time error to attempt to change IdleHandlers. Could you elaborate further?

Sorry @montekarlos, you are right it already throws at compile time. Forget that and thanks for your important contribution.

montekarlos: @BDisp @tznind Hey is there anything else I need to do with this PR?

BDisp: I don't think so. We just have to wait for @tig to have time available to check out.

tznind: Just reading up on the docs on `AsReadOnly()` and noticed this:

> However, if changes are made to the underlying [List<T>](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1?view=net-6.0) object, the read-only collection reflects those changes.
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.asreadonly?view=net-6.0

So I think we should actually have `new List<Func<bool>>(idleHandlers).AsReadOnly();`.  That way it will be properly immutable.  Also I wondered if we should have `lock (idleHandlersLock) ` around it too?

So it would be:

```csharp
/// <summary>
/// Gets the list of all idle handlers.
/// </summary>
public ReadOnlyCollection<Func<bool>> IdleHandlers {
	get {
		lock (idleHandlersLock) {
			return new List<Func<bool>> (idleHandlers).AsReadOnly();
		}
	}
}
```

montekarlos: Agreed

tig:
Updated title to reflect this is a breaking change to the public API.

I agree the chance this change will impact existing code is low, and the positive impact of it is high. So it is worth merging even if it is a breaking change.

tig (approved):
Excellent find, excellent sluthing, excellent collaboration, excellent fix. Thank you all!

PR #2020, opened 2022-09-16
Fixes #2026. Added ClearOnVisibleFalse to flag if the view must be cleared or not.

### LINKED ISSUE #1984: Setting Label.Visible to false does not hide the Label
**Describe the bug**
In ReactiveExample, binding `Label.Visible` to a boolean does not hide the label when the boolean turns false.

**To Reproduce**
1. Clone the repo
2. In Visual Studio, open `Terminal.sln`, set `ReactiveExample` as Startup project
3. In `ReactiveExample` -> `LoginView.cs` make the following change (replace setting `ColorScheme` with setting `Visible`):

image

4. Run and fill the input fields, then clear the input fields

**Expected behavior**
I expect the label to disappear when both input fields contain values This does not happen, it just stays, showing the same old text.

(My goal is to hide the validation label when the input fields are filled.)

**Screenshots**

**Modified app (as in above screenshot)**
Starting the app all is well, the label is showing ("Please, enter..."):

image

After entering text the label should disappear, but instead still shows the "old" text ("Please, enter..."):

image

**UNmodified app (as freshly pulled)**
Just to check that the bindings etc. are all set up I reverted back to the unmodified version. Here is the app when starting, showing the validation text (here in red since the code for changing the `ColorScheme` is active):

image

The validation text changes properly when having filled both input fields ("The input is valid"):

image

And it changes back when clearing the input fields:

image

Am I wrong in my expectation that hiding the label should be possible as well? If yes, how else would I do it?

 - OS: Windows 11
 - Version 1.7.2
 - .NET 6 console app

**Additional context**
-

### LINKED ISSUE #2026: Create a a flag to set if the view must be cleared or not on Visible as false.
I'm reopen this because it's causing some bad behaviors on situations where we don't want to clear the view at all. I appreciate suggestions for this. On example where I don't want to clear is on a `ScrollBarView` when set Visible to false I don't want the area been cleaned if the Redraw method is overridden.

I think the best solution is creating a `ClearOnVisibleFalse` property with a default value as `true`.

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/1984#issuecomment-1249336229_

BDisp:
Fixes #2026- Added a flag to set if a view can be cleared if the `Visible` property is false, the default is `true`.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1820,7 +1820,7 @@ namespace Terminal.Gui {
 		/// <param name="command"></param>
 		public void ClearKeybinding (params Command [] command)
 		{
-			foreach (var kvp in KeyBindings.Where (kvp => kvp.Value.SequenceEqual (command)).ToArray()) {
+			foreach (var kvp in KeyBindings.Where (kvp => kvp.Value.SequenceEqual (command)).ToArray ()) {
 				KeyBindings.Remove (kvp.Key);
 			}
 		}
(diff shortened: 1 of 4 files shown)


tig (approved):
LGTM

PR #2032, opened 2022-09-17
Fixes #2008 - attempt to unwrap `ConsoleKey.Packet` in WindowsDriver

### LINKED ISSUE #2008: Shift+key denied by text field as "control character" on remote systems
**Describe the bug**
I cannot type any character with Shift modifier - no uppercase chars, no colon, no nothing. This does not happen on all systems.

**To Reproduce**
1. Pull the Terminal.GUI
2. Start the unmodified UICatalog project
3. Put the cursor in any text input field
4. Type Shift+a - nothing happens
5. Type a - a is typed

**Expected behavior**
I expect A to be typed when pressing Shift+a

**Screenshots**
It seems like the key is discarded as "control character". So the `InsertText (kb)` is never reached.

image

This does not happen on all systems, but on the one I have the debugger running ^^ Either the check should ignore the Shift modifier, or in an earlier step a conversion of Shift+a to uppercase A fails. Not sure how it is supposed to work.

_(Note: removing the numlock modifier makes no difference.)_

Here is the corresponding `KeyEventRecord`:
image

 - OS: Windows 11, English UI language, German keyboard language, through RDP session
 - Version: relatively recent dev version: https://github.com/gui-cs/Terminal.Gui/commit/3866e65919ec67c1159162fb271f5f99fcf3a169

**Additional context**

Call stack:

```
>	Terminal.Gui.dll!Terminal.Gui.TextField.ProcessKey(Terminal.Gui.KeyEvent kb) Line 565	C#
 	Terminal.Gui.dll!Terminal.Gui.View.ProcessHotKey(Terminal.Gui.KeyEvent keyEvent) Line 1881	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.ProcessKeyEvent(Terminal.Gui.KeyEvent ke) Line 431	C#
 	Terminal.Gui.dll!Terminal.Gui.WindowsDriver.ProcessInput(Terminal.Gui.WindowsConsole.InputRecord inputEvent) Line 859	C#
 	Terminal.Gui.dll!Terminal.Gui.WindowsDriver.PrepareToRun.AnonymousMethod__34_0(Terminal.Gui.WindowsConsole.InputRecord e) Line 759	C#
 	Terminal.Gui.dll!Terminal.Gui.WindowsMainLoop.Terminal.Gui.IMainLoopDriver.MainIteration() Line 1881	C#
 	Terminal.Gui.dll!Terminal.Gui.MainLoop.MainIteration() Line 295	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.RunMainLoopIteration(ref Terminal.Gui.Application.RunState state, bool wait, ref bool firstIteration) Line 1014	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.RunLoop(Terminal.Gui.Application.RunState state, bool wait) Line 996	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.Run(Terminal.Gui.Toplevel view, System.Func<System.Exception, bool> errorHandler) Line 1172	C#
 	UICatalog.dll!UICatalog.Scenario.Run() Line 200	C#
 	UICatalog.dll!UICatalog.UICatalogApp.Main(string[] args) Line 111	C#
```

tznind:
Fixes #2008 - When using some remoting technologies some keyboard input is supplied as `ConsoleKey.Packet`.

## Pull Request checklist:

- [ x I have checked my code and corrected any poor grammar or misspellings

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
@@ -1244,6 +1244,9 @@ namespace Terminal.Gui {
 				keyModifiers.Scrolllock = scrolllock;
 
 			var ConsoleKeyInfo = new ConsoleKeyInfo (keyEvent.UnicodeChar, (ConsoleKey)keyEvent.wVirtualKeyCode, shift, alt, control);
+
+			ConsoleKeyInfo = RemapPacketKey(ConsoleKeyInfo);
+
 			return new WindowsConsole.ConsoleKeyInfoEx (ConsoleKeyInfo, capslock, numlock);
 		}
 
@@ -1756,6 +1759,40 @@ namespace Terminal.Gui {
 
 		public override void CookMouse ()
 		{
+		}
+
+
+		/// <summary>
+		/// Handles case when the 'key' providied is ConsoleKey.Packet by
+		/// returning a new <see cref="ConsoleKeyInfo"/> where key is remapped
+		/// by parsing the Unicode char data of the <see cref="ConsoleKeyInfo"/>
+		/// </summary>
+		internal static ConsoleKeyInfo RemapPacketKey (ConsoleKeyInfo original)
+		{
+			// If the key struck was virtual
+			if(original.Key == ConsoleKey.Packet)
+			{
+				// Try to parse the unicode key e.g. 'A' into a value in ConsoleKey
+				// so that other parts of the program consider it as a regular button
+				// press
+				ConsoleKey remappedkey;
+				if(Enum.TryParse<ConsoleKey>(
+					// have to turn e.g. 'a' to something parseable as
+					// an enum value (i.e. Upper it)
+					original.KeyChar.ToString().ToUpper(), 
+					out remappedkey))
+				{
+					return new ConsoleKeyInfo(
+						original.KeyChar,
+						remappedkey,
+						original.Modifiers.HasFlag(ConsoleModifiers.Shift),
+						original.Modifiers.HasFlag(ConsoleModifiers.Alt),
+						original.Modifiers.HasFlag(ConsoleModifiers.Control)
+					);
+				}
+			}
+
+			return original;
 		}
 		#endregion
 	}
--- a/UnitTests/ConsoleDriverTests.cs
+++ b/UnitTests/ConsoleDriverTests.cs
@@ -608,5 +608,29 @@ namespace Terminal.Gui.ConsoleDrivers {
 			Application.Run (win);
 			Application.Shutdown ();
 		}
+
+		/// <summary>
+		/// Sometimes when using remoting tools EventKeyRecord sends 'virtual keystrokes'.
+		/// These are indicated with the wVirtualKeyCode of 231.  When we see this code
+		/// then we need to look to the unicode character (UnicodeChar) instead of the key
+		/// when telling the rest of the framework what button was pressed. For full details
+		/// see: https://github.com/gui-cs/Terminal.Gui/issues/2008
+		/// </summary>
+		[Theory]
+		[InlineData('A',true,false,false,ConsoleKey.A)]
+		[InlineData('z',false,false,false,ConsoleKey.Z)]
+		public void TestVKPacket(char unicodeCharacter,bool shift, bool alt, bool control, ConsoleKey expectedRemapping)
+		{
+			var before = new ConsoleKeyInfo(unicodeCharacter,ConsoleKey.Packet,shift,alt,control);
+			var after = WindowsDriver.RemapPacketKey(before);
+
+			Assert.Equal(before.KeyChar, after.KeyChar);
+
+			// The thing we are really interested in, did we correctly convert
+			// the input ConsoleKey.Packet to the correct physical key
+			Assert.Equal(expectedRemapping,after.Key);
+
+			Assert.Equal(after.Modifiers,before.Modifiers);
+		}
 	}
 }


tznind:
@heinrich-ulbricht please are you able to test my PR to see if this has any effect on your Shift+char issue with `TextView` / `TextField` etc.

I have gone to the point where the keystrokes first appear in `WindowsDriver` and made a conservative edit that tries to remap the Packet virtual key so the virtual input keystrokes should now look identical to regular keystrokes from then on. 

Since I don't have a machine that sends these PACKET entries I can't test it myself.  But I have written a couple of unit tests and if successful I can write some more to make sure nothing goes wrong when trying to remap wierd characters e.g. 'newline' or 'escape' etc.  

I've left it as draft for now since I want to see that
- It definitely fixes your problem
- There are no corner cases where the remapping will map to an invalid/different Enum.

For the remapping I followed [this thread on StackOverflow](https://stackoverflow.com/questions/11495104/is-there-a-conversion-method-that-takes-a-char-and-produces-a-consolekey).  But adjusted it to correctly handle lower case too.

Test is written to use `InlineData` so its easy to expand for the strange symbols etc

heinrich-ulbricht:
Thank you, will do this weekend!

heinrich-ulbricht:
@tznind Looks promising. It works for A, B etc. But not for : and other non-alphabet characters. For those the `Enum.TryParse` fails.

BDisp:
> @tznind Looks promising. It works for A, B etc. But not for : and other non-alphabet characters. For those the `Enum.TryParse` fails.

May be for them is not necessary to be mapped and come with the correct key code? So only try parse from A to Z?

heinrich-ulbricht:
Had a look at `NetDriver` and what the `ConsoleKeyInfo` there looks like after pressing Shift+'.' (== ':'). The KeyChar there is set, no modifier:
```
-		consoleKeyInfo	{System.ConsoleKeyInfo}	System.ConsoleKeyInfo
		Key	0	System.ConsoleKey
		KeyChar	58 ':'	char
		Modifiers	0	System.ConsoleModifiers
```

BDisp:
The `NetDriver` works in a different way. It read all the keys come from a virtual terminal and also read characters sequences. The `Shift` key is irrelevant from `A to Z`, because we want to receive the letters from `A to Z` or from `a to z` and we don't need to check if it has the shift modifier or not. Also for others keys that are return by pressing the `shift+char` the keyboard driver also send the right key. We only need to check for the shift key if it's used with more keys, like `Ctrl and Alt`, or it's used with another char that isn't modified with the shift key, like `Shift+CursorLeft` and so on.
But this keys come encoded with the `PACKET` key and need to be decode, but perhaps for the keys from `A to Z` and not for others that depend of the keyboard language specific.

heinrich-ulbricht:
Some tests:

Ctrl
```
-		original	{System.ConsoleKeyInfo}	System.ConsoleKeyInfo
		Key	17	System.ConsoleKey
		KeyChar	0 '\0'	char
		Modifiers	Control	System.ConsoleModifiers
```

Ctrl+Shift
```
-		original	{System.ConsoleKeyInfo}	System.ConsoleKeyInfo
		Key	16	System.ConsoleKey
		KeyChar	0 '\0'	char
		Modifiers	Shift | Control	System.ConsoleModifiers
```

Ctrl+Shift+A
```
-		original	{System.ConsoleKeyInfo}	System.ConsoleKeyInfo
		Key	A	System.ConsoleKey
		KeyChar	1 '\u0001'	char
		Modifiers	Shift | Control	System.ConsoleModifiers
```

tznind:
> @tznind Looks promising. It works for A, B etc. But not for : and other non-alphabet characters. For those the `Enum.TryParse` fails.

That's great @heinrich-ulbricht thanks for testing.  I suspected that `TryParse` was a bit hacky.  I'll just write a switch statement for the whole `ConsoleKey` Enum (at least where theres a compatible letter key and skip A-Z since they work with TryParse).  There are only ~100 entries so shouldn't be to onerous.

I suspect for the foreign language keys there might still be issues.  But since you can now type the US letters, this approach seems sensible to start with.

Your comment is very helpful where it shows what is and isn't working, I can add those to the unit test I created.

heinrich-ulbricht:
@tznind I have URL input fields so as long as one can type URLs it's fine for me :zany_face:

tznind:
@heinrich-ulbricht  please can you see how it is behaving now.  

I have mapped every key on my US layout keyboard and added the novel characters that appear in UK keyboard (since I also have one of those to test with).  

I'm not sure what to do with characters that appear in different keys on different keyboard layouts (e.g. `"` and `@` are on different keys depending on layout).

I didn't seem to see all the `Oem` keys so there might be some additional mappings that could be added.  And there are clearly some keys that produce the same symbol e.g. `.` which appears on both the `>` key and the numerical keypad.  I think we get a flag for that though so maybe that could be looked into (or it may not really matter which is used).

It seems that some of the code I am writting is similar to `public Key MapKey (WindowsConsole.ConsoleKeyInfoEx keyInfoEx)` so I will probably also dig deeper to make sure I'm not rewritting duplicate code.

heinrich-ulbricht:
@tznind It works better now, I can input URLs ^^ Pressed a few more keys, single and double quotes are missing:
```
-		original	{System.ConsoleKeyInfo}	System.ConsoleKeyInfo
		Key	Packet	System.ConsoleKey
		KeyChar	34 '"'	char
		Modifiers	Shift	System.ConsoleModifiers
```
```
-		original	{System.ConsoleKeyInfo}	System.ConsoleKeyInfo
		Key	Packet	System.ConsoleKey
		KeyChar	39 '\''	char
		Modifiers	Shift	System.ConsoleModifiers
```

Somehow this approach doesn't feel right, though. I read that the packet key type was introduced to _not_ have to deal with different keyboard layouts when sending remote key presses around. To "just pass the actual key through" without going through multiple transformations. It would be nice if we could do that. But this again somehow feels like it might require a bigger architectural change.

As an immediate workaround it definitely solves the (my) problem.

tznind:
> Somehow this approach doesn't feel right, though. I read that the packet key type was introduced to _not_ have to deal with different keyboard layouts when sending remote key presses around.
_[heinrich-ulbricht](https://github.com/heinrich-ulbricht)_

I agree this definetly feels a bit sketchy.  I'm very much open to talking more about this though!

### Terminal.Gui keyboard events

All the `View` in `Terminal.Gui` deal with `KeyEvent` which is a wrapper for `Key` and `KeyModifiers`

### The Key Enum
`Key` is an `Enum` declared and defined in Terminal.Gui

> The Key enumeration contains special encoding for some keys, but can also encode all the unicode values that can be passed.   

This means that any `char` can be stored in references to `Key` **even if they don't have an explicit declared entry**.  There is [a ticket already open in discussion](https://github.com/gui-cs/Terminal.Gui/discussions/1998) to add many of the missing values.

Because `Key` is using the Unicode/ASCII numbers for entries it doesn't care about keyboard layout

https://github.com/gui-cs/Terminal.Gui/blob/558b2b793f700dec5672536b934300b86148bfe7/Terminal.Gui/Core/Event.cs#L139-L147

The `KeyEvent` class is explicitly set up with the expectation that unmapped `Key` values are used e.g.:


https://github.com/gui-cs/Terminal.Gui/blob/558b2b793f700dec5672536b934300b86148bfe7/Terminal.Gui/Core/Event.cs#L487-L492

### ConsoleKey
`ConsoleKeyInfo` only appears in the implementations of `ConsoleDriver` (i.e. `NetDriver` and `WindowsDriver`) and not in the rest of the `Terminal.Gui` codebase which is good!

### Summary
So the goal here is to correctly convert `ConsoleKeyInfo.Packet` + `char` (unicode) into the correct `Key` enum entry.  This involves 2 things

- If the char is `\b` or `\t` or `\u001b` (escape)  or `\r` then we should turn it into the correct `Key` entry (unless those already match on `uint` value).
- If the char is any printable character we should just cast it to `Key` and not worry that it doesn't have an explicit entry

I think my going via `ConsoleKeyInfo` was a mistake since it seems tied to keyboard layouts and probably over complicating things.  Especially since 99% of keystrokes should just be castable to `Key` as `uint`.

heinrich-ulbricht:
@tznind Well. This just works :D

I can type anything in combination with Shift now. No exceptions.
Home, End, Cursors, Escape, Tab, Backspace also work and don't use the Packet type.

Here are some key events taken from `ToConsoleKeyInfoEx`, just say if you need more:

Home
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	0 '\0'	char
		bKeyDown	true	bool
		dwControlKeyState	EnhancedKey	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	36	ushort
		wVirtualScanCode	71	ushort
```
End
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	0 '\0'	char
		bKeyDown	true	bool
		dwControlKeyState	EnhancedKey	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	35	ushort
		wVirtualScanCode	79	ushort
```
Right arrow
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	0 '\0'	char
		bKeyDown	true	bool
		dwControlKeyState	EnhancedKey	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	39	ushort
		wVirtualScanCode	77	ushort
```
ESC
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	27 '\u001b'	char
		bKeyDown	true	bool
		dwControlKeyState	0	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	27	ushort
		wVirtualScanCode	1	ushort
```
Shift
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	0 '\0'	char
		bKeyDown	true	bool
		dwControlKeyState	ShiftPressed	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	16	ushort
		wVirtualScanCode	42	ushort
```
Shift + a pressed
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	65 'A'	char
		bKeyDown	true	bool
		dwControlKeyState	ShiftPressed	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	231	ushort
		wVirtualScanCode	0	ushort
```
Shift + Tab pressed
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	9 '\t'	char
		bKeyDown	true	bool
		dwControlKeyState	ShiftPressed	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	9	ushort
		wVirtualScanCode	15	ushort
```
Return
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	13 '\r'	char
		bKeyDown	true	bool
		dwControlKeyState	0	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	13	ushort
		wVirtualScanCode	28	ushort
```
Ctrl + Home pressed
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	0 '\0'	char
		bKeyDown	true	bool
		dwControlKeyState	LeftControlPressed | EnhancedKey	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	36	ushort
		wVirtualScanCode	71	ushort
```
Ctrl + a pressed
```
-		keyEvent	{Terminal.Gui.WindowsConsole.KeyEventRecord}	Terminal.Gui.WindowsConsole.KeyEventRecord
		UnicodeChar	1 '\u0001'	char
		bKeyDown	true	bool
		dwControlKeyState	LeftControlPressed	Terminal.Gui.WindowsConsole.ControlKeyState
		wRepeatCount	1	ushort
		wVirtualKeyCode	65	ushort
		wVirtualScanCode	30	ushort
```

By the way - is Ctrl+A (to select all text) supposed to work in input fields? Because it doesn't. Just something I noticed a while back, not sure if this is connected to our current issue. Just want to mention it just in case.

Thanks for diving into this. This solution feels a lot better.

BDisp:
> By the way - is Ctrl+A (to select all text) supposed to work in input fields? Because it doesn't. Just something I noticed a while back, not sure if this is connected to our current issue. Just want to mention it just in case. Shift+Home and Shift+End do work.
> (Note: Is this a unicode character there for Ctrl+a that needs special handling?)

On `TextField` and `TextView` to select all text is by pressing` Ctrl+T`.

heinrich-ulbricht:
@BDisp You are fast. Just this moment I pressed the right mouse key and the context menu told me this :) Strange combo, my Windows background strongly expects Ctrl+A ^^

BDisp:
`Ctrl+A` is for `Command.LeftHome` on `TextField` and for `Command.StartOfLine` on `TextView`.

tznind:
So it looks like the packet key is not used in any of these cases (below) for you.  So the following code I have in this PR will probably never be hit so could be removed in your case.  But I don't know if other software or environments might use PACKET for any of these.  @tig let me know if you want me to remove this code or leave it in (or anyone else who has a view!).

```csharp
case '\t':
	result = original.Modifiers == ConsoleModifiers.Shift ? Key.BackTab : Key.Tab;
	return true;
case '\u001b':
	result = Key.Esc;
	return true;
case '\b':
	result = Key.Backspace;
	return true;
case '\n':
case '\r':
	result = Key.Enter;
	return true;
```

@heinrich-ulbricht  one last thing I thought I'd mention incase you hadn't seen -  You can change the keybinding for SelectAll with:

```csharp
TextField tv = new TextField ();
tv.AddKeyBinding (Key.A | Key.CtrlMask, Command.SelectAll);
```

tig:
> @tig let me know if you want me to remove this code or leave it in (or anyone else who has a view!).
> 

Nuke it. ;-)

tig (approved):

tznind:
Ok I have nuked the switch.  @heinrich-ulbricht if you have the time please to do a final check that I haven't missed something (i.e. that it still works for all teh typing).

BDisp:
> Somehow this approach doesn't feel right, though. I read that the packet key type was introduced to _not_ have to deal with different keyboard layouts when sending remote key presses around. To "just pass the actual key through" without going through multiple transformations. It would be nice if we could do that. But this again somehow feels like it might require a bigger architectural change. **Edit**: Thought about that. If the final console output IS dependent on those enums then there is no choice.
> 
> As an immediate workaround it definitely solves the (my) problem.

This justification intrigued me and reflected if they created this `Packet` Key was precisely so there was no conflict with any keyboard layout. As this situation only happened to `WindowsDriver` so I thought the problem was just like this driver handles `Shift` Key.

So I  thinking commenting this two lines on the `WindowsDriver` fix that problem because the `Shift` key doesn't need to the Key at this moment but only int the modifiers.

https://github.com/gui-cs/Terminal.Gui/blob/deef59e877b7889c59b317eb0dccc03465e0f8ad/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs#L1377-L1378

I also commented on these two on the `FakeDriver` and all the current unit tests passed.

https://github.com/gui-cs/Terminal.Gui/blob/deef59e877b7889c59b317eb0dccc03465e0f8ad/Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs#L363-L364

@heinrich-ulbricht since only you are using that remote tool can you do this changes to the develop branch and test if the Shift key is processed well in all situations, please? Thanks.

BDisp:
I change the unit to this and it passed.
```cs

public void TestVKPacket (char unicodeCharacter, Key expectedRemapping)
{
	var before = new ConsoleKeyInfo (unicodeCharacter, ConsoleKey.Packet, false, false, false);
	var top = Application.Top;

	top.KeyPress += (e) => {
		var after = e.KeyEvent.Key;
		Assert.Equal (before.KeyChar, (char)after);
		Assert.Equal (expectedRemapping, after);
	};

	Application.Begin (top);

	Application.Driver.SendKeys (unicodeCharacter, ConsoleKey.Packet, false, false, false);
}

```

BDisp:
I'm pretty sure the culprit of the issue was as demonstrated in the previous comment.

heinrich-ulbricht:
So you mean changong the Shift handling could be alternative solution?

And for the record my test setting:
* Fedora + Remmina for RDP
* any Windows VM on Azure to RDP to
* UICatalog from the current repo running on the Windows machine

That's it, not sooo esoteric. But yes, I seem to be the only one doing this with Terminal.Gui :D

I'll check.

BDisp:
> So you mean changing the Shift handling could be alternative solution?
> 

Yes. In `Windows Froms` normally we check for a `KeyChar` with the keys modifiers, but `Terminal.Gui` is very tied with the shifted keys, like `Key.A | Key.CtrlMask`, instead of `KeyChar == 'A' && Modifiers.Control`. I tried to change that behavior before but I'm started having many conflicts and errors and I gave up.

> And for the record my test setting:
> 
>     * Fedora + Remmina for RDP
> 
>     * any Windows VM on Azure to RDP to
> 
>     * UICatalog from the current repo running on the Windows machine
> 
> 
> That's it, not sooo esoteric. But yes, I seem to be the only one doing this with Terminal.Gui :D
> 
Yes i think you are the only one doing that :-) I don't have `Azure` and have no idea how to setup something like that.

> I'll check.

Thanks.

BDisp:
Instead of commenting the lines related above is better do the follow:

Changing this:
https://github.com/gui-cs/Terminal.Gui/blob/deef59e877b7889c59b317eb0dccc03465e0f8ad/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs#L1367-L1369

To this:
```cs
	if (keyInfo.KeyChar != 0) {
		if (keyInfo.Key == ConsoleKey.Packet) {
			return (Key)((uint)keyInfo.KeyChar);
		} else {
			return MapKeyModifiers (keyInfo, (Key)((uint)keyInfo.KeyChar));
		}
	}
```

and changing this:
https://github.com/gui-cs/Terminal.Gui/blob/deef59e877b7889c59b317eb0dccc03465e0f8ad/Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs#L351-L353

to this:
```cs
	if (keyInfo.KeyChar != 0) {
		if (keyInfo.Key == ConsoleKey.Packet) {
			return (Key)((uint)keyInfo.KeyChar);
		} else {
			return MapKeyModifiers (keyInfo, (Key)((uint)keyInfo.KeyChar));
		}
	}

```

BDisp:
Another better solution and as this issue only affect the `Packet` key used with the `Shift` key it's more accurate only not shifted the `Shift` key with the `ConsoleKeyInfo.KeyChar` if the `ConsoleKeyInfo.Key` is the `ConsoleKey.Packet`.

So, in this lines:
https://github.com/gui-cs/Terminal.Gui/blob/deef59e877b7889c59b317eb0dccc03465e0f8ad/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs#L1377-L1378

Change to this:
```cs
	if (keyInfo.Key != ConsoleKey.Packet && (keyInfo.Modifiers & ConsoleModifiers.Shift) != 0)
		keyMod = Key.ShiftMask;

```

In this lines:
https://github.com/gui-cs/Terminal.Gui/blob/deef59e877b7889c59b317eb0dccc03465e0f8ad/Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs#L363-L364

Change to this:
```cs
	if (keyInfo.Key != ConsoleKey.Packet && (keyInfo.Modifiers & ConsoleModifiers.Shift) != 0)
		keyMod = Key.ShiftMask;

```

tznind:
Sounds like you have a simpler solution than this PR @BDisp.  It looks good but I'm sort of loosing track of all the edits your suggesting.

Do you want to create a new PR with your changes? I'm happy to go with adjusting the existing logic instead of my additional logic if that is more correct.

BDisp:
But we have to be sure if my solution is more accurate. Please, change your unit test to this and confirm if it must pass.

```cs
[Theory]
[InlineData ('A', false, false, false, Key.A)]
[InlineData ('A', true, false, false, Key.A)]
[InlineData ('A', true, true, false, Key.A | Key.AltMask)]
[InlineData ('A', true, true, true, Key.A | Key.AltMask | Key.CtrlMask)]
[InlineData ('z', false, false, false, Key.z)]
[InlineData ('z', true, false, false, Key.z)]
[InlineData ('z', true, true, false, Key.z | Key.AltMask)]
[InlineData ('z', true, true, true, Key.z | Key.AltMask | Key.CtrlMask)]
[InlineData ('=', false, false, false, (Key)'=')]
[InlineData ('=', true, false, false, (Key)'=')]
[InlineData ('=', true, true, false, (Key)'=' | Key.AltMask)]
[InlineData ('=', true, true, true, (Key)'=' | Key.AltMask | Key.CtrlMask)]
[InlineData ('英', false, false, false, (Key)'英')]
[InlineData ('英', true, false, false, (Key)'英')]
[InlineData ('英', true, true, false, (Key)'英' | Key.AltMask)]
[InlineData ('英', true, true, true, (Key)'英' | Key.AltMask | Key.CtrlMask)]
[InlineData ('+', false, false, false, (Key)'+')]
[InlineData ('+', true, false, false, (Key)'+')]
[InlineData ('+', true, true, false, (Key)'+' | Key.AltMask)]
[InlineData ('+', true, true, true, (Key)'+' | Key.AltMask | Key.CtrlMask)]
[InlineData ('\0', false, false, false, (Key)'\0')]
[InlineData ('\0', true, false, false, (Key)'\0')]
[InlineData ('\0', true, true, false, (Key)'\0' | Key.AltMask)]
[InlineData ('\0', true, true, true, (Key)'\0' | Key.AltMask | Key.CtrlMask)]
public void TestVKPacket (char unicodeCharacter, bool shift, bool alt, bool control, Key expectedRemapping)
{
	var before = new ConsoleKeyInfo (unicodeCharacter, ConsoleKey.Packet, shift, alt, control);
	Assert.True (WindowsDriver.TryRemapPacketKey (before, out var after));

	// The thing we are really interested in, did we correctly convert
	// the input ConsoleKey.Packet to the correct physical key
	Assert.Equal (expectedRemapping, after);
}

```

BDisp:
The `WindowsDriver` is the only driver that really processes the key down and the key up events, by check the `bKeyDown` field. The others drivers processes that by simulation. Only on `WindowsDriver` is possible to get the modifiers keys directly by the library, the others drivers are dependent of the result from the `Terminal.Gui.Key`. So the modifiers keys in the `WindowsDriver` are very reliable.

tznind:
I added `static Key ApplyModifiers (Key c, ConsoleModifiers modifiers)` which applies the Alt and Shift masks to the `Key` but it looked a lot like the existing method `Key MapKeyModifiers (ConsoleKeyInfo keyInfo, Key key)`.  However if I use `MapKeyModifiers` instead of the method I added the new tests fail.

BDisp:
> I've updated my code and added the test cases you provided. But I'm a little confused:
> 
> ```
> [InlineData ('z', true, true, true, Key.z | Key.AltMask | Key.CtrlMask)]
> ```
> 
> Why is this expecting CtrlMask and AltMask but not ShiftMask? I get that shift is already modelled with `Key.A` vs `Key.a` but then it seems confusing why there is a `Key.ShiftMask` at all?

Thanks, for point me this. I was forget that. The below code doesn't process the `MapKeyModifiers` method at all. But if you use the lower cases the `MapKeyModifiers` is processed. You can ask, but a Shift+'lower case' always return a upper case. Yes, it's right. that why there is a `ShortcutHelper.GetModifiersKey` method which allows getting a shortcut like `AddKeyBinding ((Key)((int)'B' + Key.ShiftMask | Key.AltMask), Command.WordLeftExtend);` to execute the `Command.WordLeftExtend` on the `TextField.`. So, in the lower cases the Shift key shouldn't be added on the Terminal.Gui.Key and be managed like a upper case.
https://github.com/gui-cs/Terminal.Gui/blob/deef59e877b7889c59b317eb0dccc03465e0f8ad/Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs#L313-L329

So, my code, which is using the develop branch and not this PR, to make the things more cleared, will be the same for the `WindowsDriver` and the `FakeDriver`. Although they are duplicate is better maintain each one separated.

```cs
private Key MapKeyModifiers (ConsoleKeyInfo keyInfo, Key key)
{
	Key keyMod = new Key ();
	if (CanShiftBeAdded (keyInfo))
		keyMod = Key.ShiftMask;
	if ((keyInfo.Modifiers & ConsoleModifiers.Control) != 0)
		keyMod |= Key.CtrlMask;
	if ((keyInfo.Modifiers & ConsoleModifiers.Alt) != 0)
		keyMod |= Key.AltMask;

	return keyMod != Key.Null ? keyMod | key : key;
}

private bool CanShiftBeAdded (ConsoleKeyInfo keyInfo)
{
	if ((keyInfo.Modifiers & ConsoleModifiers.Shift) == 0) {
		return false;
	}
	if (keyInfo.Key == ConsoleKey.Packet) {
		var ckiChar = keyInfo.KeyChar;
		if (char.IsLetter (ckiChar)) {
			return false;
		}
	}
	return true;
}

```

In the `ConsoleDriverTests`:
```cs
[Theory, AutoInitShutdown]
[InlineData ('A', false, false, false, Key.A)]
[InlineData ('A', true, false, false, Key.A)]
[InlineData ('A', true, true, false, Key.A | Key.AltMask)]
[InlineData ('A', true, true, true, Key.A | Key.AltMask | Key.CtrlMask)]
[InlineData ('z', false, false, false, Key.z)]
[InlineData ('z', true, false, false, Key.z)]
[InlineData ('z', true, true, false, Key.z | Key.AltMask)]
[InlineData ('z', true, true, true, Key.z | Key.AltMask | Key.CtrlMask)]
[InlineData ('=', false, false, false, (Key)'=')]
[InlineData ('=', true, false, false, (Key)'=' | Key.ShiftMask)]
[InlineData ('=', true, true, false, (Key)'=' | Key.ShiftMask | Key.AltMask)]
[InlineData ('=', true, true, true, (Key)'=' | Key.ShiftMask | Key.AltMask | Key.CtrlMask)]
[InlineData ('英', false, false, false, (Key)'英')]
[InlineData ('英', true, false, false, (Key)'英')]
[InlineData ('英', true, true, false, (Key)'英' | Key.AltMask)]
[InlineData ('英', true, true, true, (Key)'英' | Key.AltMask | Key.CtrlMask)]
[InlineData ('+', false, false, false, (Key)'+')]
[InlineData ('+', true, false, false, (Key)'+' | Key.ShiftMask)]
[InlineData ('+', true, true, false, (Key)'+' | Key.ShiftMask | Key.AltMask)]
[InlineData ('+', true, true, true, (Key)'+' | Key.ShiftMask | Key.AltMask | Key.CtrlMask)]
[InlineData ('0', false, false, false, Key.D0)]
[InlineData ('0', true, false, false, Key.D0 | Key.ShiftMask)]
[InlineData ('0', true, true, false, Key.D0 | Key.ShiftMask | Key.AltMask)]
[InlineData ('0', true, true, true, Key.D0 | Key.ShiftMask | Key.AltMask | Key.CtrlMask)]
[InlineData ('\0', false, false, false, (Key)'\0')]
[InlineData ('\0', true, false, false, (Key)'\0' | Key.ShiftMask)]
[InlineData ('\0', true, true, false, (Key)'\0' | Key.ShiftMask | Key.AltMask)]
[InlineData ('\0', true, true, true, (Key)'\0' | Key.ShiftMask | Key.AltMask | Key.CtrlMask)]
public void TestVKPacket (char unicodeCharacter, bool shift, bool alt, bool control, Key expectedRemapping)
{
	var before = new ConsoleKeyInfo (unicodeCharacter, ConsoleKey.Packet, shift, alt, control);
	var top = Application.Top;

	top.KeyPress += (e) => {
		var after = e.KeyEvent.Key;
		Assert.Equal (before.KeyChar, (char)after);
		Assert.Equal (expectedRemapping, after);
	};

	Application.Begin (top);

	Application.Driver.SendKeys (unicodeCharacter, ConsoleKey.Packet, shift, alt, control);
}

```

BDisp:
> Sounds like you have a simpler solution than this PR @BDisp. It looks good but I'm sort of loosing track of all the edits your suggesting.
> 
> Do you want to create a new PR with your changes? I'm happy to go with adjusting the existing logic instead of my additional logic if that is more correct.

@tznind do you still want I submit a PR to your branch? I only do it if you agree that mine follow the current logic and is more accurate. If you have doubt about my solution I prefer you go forward. Thanks.

tznind:
> [...]
> @tznind do you still want I submit a PR to your branch? I only do it if you agree that mine follow the current logic and is more accurate. If you have doubt about my solution I prefer you go forward. Thanks.

Yes please! I think you know this area of the library better than me and I'm happy to follow your lead on this :)

BDisp:
> Yes please! I think you know this area of the library better than me and I'm happy to follow your lead on this :) Or if it is easier I can close this PR and you can open a new one.

Thanks. I'll try to change this PR and submit to your branch.

BDisp:
Done. It really good if @heinrich-ulbricht test first with others chars than letters where using the shift also change the output char.

heinrich-ulbricht:
I'll check the next day or two. I hope to not confuse any versions when testing ^^

BDisp:
> I'll check the next day or two. I hope to not confuse any versions when testing ^^

You only need to checkout this PR :-) Thanks.

heinrich-ulbricht:
@BDisp I checked out the pull request. The uppercase characters (Shift+a == A) still work, but the special characters (e.g. Shift+1 == !) do not work anymore.

BDisp:
> @BDisp I checked out the pull request. The uppercase characters (Shift+a == A) still work, but the special characters (e.g. Shift+1 == !) do not work anymore.

Thanks for reporting. I'll check.

BDisp:
@heinrich-ulbricht I submited the PR https://github.com/tznind/gui.cs/pull/126 where I excluded the shift key when `IsLetterOrDigit`, `IsSymbol` and `IsPunctuation`. I don't know if the unit test will fail with others keyboard layouts. Maybe it's necessary to exclude some more but I haven't a way to test it. If it's not very difficult and is free can you explain as clear as possible how setup one, please.

heinrich-ulbricht:
@BDisp Sorry, what should I describe? My setup?

BDisp:
Yes, if possible,of course.

BDisp:
Submitted the PR https://github.com/tznind/gui.cs/pull/128.
Shortcuts keys are to be used with a physical keyboard and one like `Ctrl+Shift+'a'` will never be achieve because the driver will always return the uppercase `'A'`, so only this will work `Ctrl+Shift+A`. The same for the symbols e.g. `Ctrl+Shift+'!'` will also never be achieved because the driver will return always `Ctrl+Shift+1`.

I'm trying to create a `Packet` simulator scenario on the `UICatalog` which will use the [SendInput function](https://learn.microsoft.com/en-US/windows/win32/api/winuser/nf-winuser-sendinput?redirectedfrom=MSDN).

heinrich-ulbricht:
Ah here my setup:
* Source maching: Fedora Linux, current version
* Remote machine: Windows 11, Azure VM
* RDP software: Remmina
* German Keyboard
* Both the local and remote machine use the German keyboard layout

That's it.

tig:
Once the merge conflicts are resolved, is this PR ready to be merged?

BDisp:
I'm doing big changes to this and I ask to @tznind if he can be wait untill I finish it. The decoded key must be provide they're mapped to key and not after.

BDisp:
@tznind I submitted the PR https://github.com/tznind/gui.cs/pull/128 to your branch. I already merged the develop branch because of the resulting conflicts. All drivers, except the `CursesDriver` are prepared to receive Packet keys. The `CursesDriver` only simulate through the `SendKeys` method. The reason is the keys are only caught in the `ProcessInput` method with the `Curses.get_wch` method, while the others drivers the keys are passed on the parameter. I really don't know if `ncurses` can handle the virtial packet keys.
I also created a `VkeyPacketSimulator` scenario which supports all the drivers. Typing in the first`TextView` will write in the second `TextView`.

tznind:
This is awesome work thanks!  I have merged the PR and will get to testing it / looking in depth asap.  

Do you consider this now as good to go? shall I mark it no longer Draft?

BDisp:
> This is awesome work thanks! I have merged the PR and will get to testing it / looking in depth asap.
> 
Thanks.
> Do you consider this now as good to go? shall I mark it no longer Draft?

I forgot to delete the `CanShiftBeAdded` from the `FakeDriver`. I'll submit a new PR to fix it. After you merge you can mark as ready, please.

BDisp:
Done. Thanks for your waiting.

PR #2079, opened 2022-10-10
Fixes #2078: Enable collection initializer syntax for View

### LINKED ISSUE #2078: Enable collection initializer syntax for View
Would be nice to add support for the more structural/declarative style allowed by 
collection initializers when creating views, such as:

```csharp
	var view = new View {
		new View {
			new View()
		},
		new View()
	};
```

The only reason this currently doesn't work is that `View` does not implement `IEnumerable`:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1922	Cannot initialize type 'View' with a collection initializer because it does not implement 'System.Collections.IEnumerable'	UnitTests	...Terminal.Gui\UnitTests\ViewTests.cs	122	Active
```

The view could implement that interface explicitly like so:

```csharp
IEnumerator IEnumerable.GetEnumerator () => Subviews.GetEnumerator ();
```

And that alone would enable this usage, since the class already provides an `Add(View)` method, which is 
the other thing the compiler looks for in this initialization pattern.

kzu:
Adds the missing IEnumerable interface implementation so that collection initializers can be used to more declaratively create view structures.

Fixes #2078

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -111,7 +111,7 @@ namespace Terminal.Gui {
 	///    frames for the vies that use <see cref="LayoutStyle.Computed"/>.
 	/// </para>
 	/// </remarks>
-	public partial class View : Responder, ISupportInitializeNotification {
+	public partial class View : Responder, IEnumerable, ISupportInitializeNotification {
 
 		internal enum Direction {
 			Forward,
@@ -3039,5 +3039,7 @@ namespace Terminal.Gui {
 
 			return top;
 		}
+
+		IEnumerator IEnumerable.GetEnumerator () => Subviews.GetEnumerator ();
 	}
 }
--- a/UnitTests/ScrollBarViewTests.cs
+++ b/UnitTests/ScrollBarViewTests.cs
@@ -282,8 +282,12 @@ namespace Terminal.Gui.Views {
 			AddHandlers ();
 
 			Assert.NotNull (_scrollBar.OtherScrollBarView);
-			Assert.NotEqual (_scrollBar, _scrollBar.OtherScrollBarView);
-			Assert.Equal (_scrollBar.OtherScrollBarView.OtherScrollBarView, _scrollBar);
+			Assert.NotSame (_scrollBar, _scrollBar.OtherScrollBarView);
+			Assert.False (_scrollBar == _scrollBar.OtherScrollBarView);
+			Assert.True (_scrollBar != _scrollBar.OtherScrollBarView);
+			Assert.Same (_scrollBar.OtherScrollBarView.OtherScrollBarView, _scrollBar);
+			Assert.True (_scrollBar.OtherScrollBarView.OtherScrollBarView == _scrollBar);
+			Assert.False (_scrollBar.OtherScrollBarView.OtherScrollBarView != _scrollBar);
 		}
 
 		[Fact]
--- a/UnitTests/ViewTests.cs
+++ b/UnitTests/ViewTests.cs
@@ -117,6 +117,20 @@ namespace Terminal.Gui.Core {
 		}
 
 		[Fact]
+		public void New_CollectionInitialized ()
+		{
+			var view = new View {
+				new View {
+					new View()
+				},
+				new View()
+			};
+
+			Assert.Equal (2, view.Subviews.Count);
+			Assert.Equal (1, view.Subviews [0].Subviews.Count);
+		}
+
+		[Fact]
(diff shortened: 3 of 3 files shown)


BDisp:
I'm now remembering the View had a `IEnumerable` before but @tig removed it because wasn't `PowerShell` compatible. Many users uses `Terminal.Gui` in `PowerShell `implementations and we must have the certain sure that this PR won't be incompatible. If I'm wrong with what I ended to talk i beg for excuse.

kzu:
Bummer. I don't even know what's the deal with powershell. Is the idea that a script/commandlet would return a View and it would magically be rendered?

kzu:
If the choice is between collection initializers and ReactiveUI, I'd vote for the former, TBH 😉

kzu:
I don't understand the assert, TBH. But changing from NotEqual > NotSame and Equal > Same, fixes it 🤷

BDisp:
I don't think so, Many times we have to compare if two views are the same by using the `==`/`!=` operator and I think the `IEnumerable` will break that, but I perhaps wrong.

BDisp:
> I don't understand the assert, TBH. But changing from NotEqual > NotSame and Equal > Same, fixes it 🤷

Sorry you are right about changing from  `NotEqual` > `NotSame` because it isn't failing. It seems that the `Equal` and `NotEqual` verify if both parameters are the same object type and not if they are the same instance.

> I don't think so, Many times we have to compare if two views are the same by using the ==/!= operator and I think the IEnumerable will break that, but I perhaps wrong.

Sorry for not test this before I commented. The `==` and `!=` works also fine with this PR.

If I was you I reopen the PR again with the follow change in the failing unit test and waiting for @tig opinion.

```cs
[Fact]
[InitShutdown]
public void OtherScrollBarView_Not_Null ()
{
	Hosting_A_View_To_A_ScrollBarView ();

	AddHandlers ();

	Assert.NotNull (_scrollBar.OtherScrollBarView);
	Assert.NotSame (_scrollBar, _scrollBar.OtherScrollBarView);
	Assert.False (_scrollBar == _scrollBar.OtherScrollBarView);
	Assert.True (_scrollBar != _scrollBar.OtherScrollBarView);
	Assert.Equal (_scrollBar.OtherScrollBarView.OtherScrollBarView, _scrollBar);
	Assert.True (_scrollBar.OtherScrollBarView.OtherScrollBarView == _scrollBar);
	Assert.False (_scrollBar.OtherScrollBarView.OtherScrollBarView != _scrollBar);
}

```

BDisp:
> Worst case, I can just make my own base class for View, but I'd rather not :)

I hope @tig ask for the `ReactiveUI` guys if the `IEnumerable` implementation doesn't affecting anymore their library which that is much more evolved and thus not prevent for this PR to be merged.

BDisp:
@tig is for this scenario [#2078](https://github.com/gui-cs/Terminal.Gui/issues/2078#issue-1402552556) that @kzu wants.

tznind:
If `IEnumerable` is problematic we could add a `params` constructor:

```csharp

public View (params View[] subviews) : this (text: string.Empty, direction: TextDirection.LeftRight_TopBottom) { 
	foreach(var s in subviews)
	{
		Add(s);
	}
}
```

This would allow for something very similar to what you need (see below).

```csharp
[Fact]
public void Blah(){
		var view = new View (
			new View (
				new View()
			),
			new View()
		);
}
```

tznind:
@kzu would the `params` approach work for you? It feels like a smaller change than making `View` into `IEnumerable`.

To me the main reasons not to make `View` into `IEnumerable` are:

- It causes issues with Powershell
- It would be a breaking change for anyone writing a `View` subclass with their own `IEnumerable` implementation (e.g. a view thats shows errors and is enumerable over those error objects).
- Other UI frameworks do not have enumerable root objects (e.g. `Control` is not enumerable in WinForms)

I think each reason is quite weak on it's own but together they present a good argument for not making this change.  Does the `params` solution (above) do everything you need? If not then what are we missing on?

kzu:
I still prefer the object initializer syntax, but understand the constraint. Params should be OK, but I'll most likely define my base class nevertheless to bring it to my own usage :). So, closing.

Thanks for the responsiveness!

tznind:
Thanks for being understanding :) good luck and let us know how it goes! Or if you have any further issues.

Just incase you weren't aware and are having to write a lot of boilerplate code, there is i a visual designer tool that might make your life a bit easier: https://github.com/gui-cs/TerminalGuiDesigner

PR #2097, opened 2022-10-18
Fixes  #2096 - `Application.RootMouseEvent` now allows for cancellation

### LINKED ISSUE #2096: Allow `Application.RootMouseEvent` to cancel events
There are 2 global 'pre' handlers that fire before events get passed to any Views.  These are:

- Application.RootKeyEvent
- Application.RootMouseEvent

`RootKeyEvent` is a `Func<KeyEvent,bool>` which allows callback to return true to cancel the event.  But `RootMouseEvent` is `Action<MouseEvent>` and there is no `Handled` property on `MouseEvent` for cancelling (or any other way I could see to cancel).

My primary use case for cancellation is in TerminalGuiDesigner where you don't want right clicking in controls such as TextField to throw up the context menu - Instead the designer wants to show its own context menu and suppress right click events).

Also for consistency with RootKeyEvent.

Also handy if events are being misreported in an environment e.g. like #1617 .   A developer could add a global event handler that turned off the mouse completely by just returning `true`.

 Change `RootMouseEvent` to `Func<MouseEvent,bool>` so that it works like keyboard one.

The code change to achieve this is quite small but it is a breaking change for anyone subscribing to this event.  The current docs for this event are:

https://github.com/gui-cs/Terminal.Gui/blob/32e7b63b77655ff07dc3a9dce6074e77d0590fa3/Terminal.Gui/Core/Application.cs#L641-L644

So I don't think there will be that many users of this bit of the system so the breakage is limited.

tznind:
Fixes  #2096 - `Application.RootMouseEvent` now allows for cancellation

## Pull Request checklist:


Terminal.Gui/Core/Application.cs
```diff
@@ -639,9 +639,13 @@ static void OnUnGrabbedMouse (View view)
...
 		/// <summary>
-		/// Merely a debugging aid to see the raw mouse events
+		/// <para>
+		/// Called for new MouseEvent events before any processing is performed or
+		/// views evaluate.  Use for global mouse handling and/or debugging.
+		/// </para>
+		/// <para>Return true to suppress the MouseEvent event</para>
 		/// </summary>
-		public static Action<MouseEvent> RootMouseEvent;
+		public static Func<MouseEvent, bool> RootMouseEvent;
```
tig, line 648: Unfortunately, it does appear folks are using this API. 
https://cs.github.com/?scopeName=All+repos&scope=&q=%22RootMouseEvent%22+AND+%22Terminal.GUI%22+AND+NOT+repo%3Agui-cs%2FTerminal.Gui

e.g. (There are dozens):

image

tig (changes requested):
I think this needs to be fixed in a way that isn't a breaking change. Sad, but reality.

Terminal.Gui/Core/Application.cs
tznind: Ok that's fair.  I guess we could have a `RootMouseEventCancellable` or a `RootMouseEvent2`?

Alternatively we could add the `Handled` to `MouseEvent`.  I have looked into it and this is my understanding:

### MouseEvent
There are 3 public structs all called `MouseEvent` but the one dealt with by the `RootMouseEvent` event is `Terminal.Gui.MouseEvent`:

https://github.com/gui-cs/Terminal.Gui/blob/da34dedaa5eddf489a7a725b0ec8ece3546be433/Terminal.Gui/Core/Event.cs#L711

- `MouseEvent` is a `struct`
- `MouseEventArgs` is a class that wraps `MouseEvent` and adds `Handled` property and inherits from `EventArgs`

https://github.com/gui-cs/Terminal.Gui/blob/df7285954210c2d64ce98e5f4043263b4c86f60a/Terminal.Gui/Core/View.cs#L2745-L2763

We could add `Handled` to `MouseEvent` struct and redirect the one in `MouseEventArgs` to point to it's wrapped instance?

tig:
Closing, given #2100

PR #2107, opened 2022-10-19
Makes Notepad look better and fixes bugs

tig:
This PR modifies the Notepad Scenario in UI Catalog to be more frugal with screen real estate (and look better IMO). 

It also fixes some bugs in updating the tab names on saving etc...

aYG3458 1


tznind:
Nice, this is a big improvement!

PR #2117, opened 2022-10-21
Fixes #2109. MenuBar has extra space on left. Refactors MenuTests. Better Menu API docs.

### LINKED ISSUE #2109: MenuBar has an extra space on left. Seems odd.
image

This has long bugged me. I probably wrote the code that did it. 

Don't remember why, but I think it should be nuked. 

And why are there TWO spaces after the `?`?

tig:
Fixes #2109 

Also revamps most of the `MenuTests` to be less dependent on the formatting of `expected` within the source file, and thus less fragile.

### DIFF
--- a/Terminal.Gui/Core/ContextMenu.cs
+++ b/Terminal.Gui/Core/ContextMenu.cs
@@ -116,7 +116,7 @@ namespace Terminal.Gui {
 				Y = position.Y,
 				Width = 0,
 				Height = 0,
-				UseSubMenusSingleFrame = UseSubMenusSingleFrame
+				UseSubMenusSingleFrame = UseSubMenusSingleFrame // BUGBUG: This line of code does nothing
 			};
 
 			menuBar.isContextMenuLoading = true;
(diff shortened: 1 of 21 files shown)


tig:
@BDisp - please review this! Thanks.

Terminal.Gui/Core/ContextMenu.cs
```diff
@@ -110,13 +110,13 @@ public void Show ()
...
 			}
-
+			
 			menuBar = new MenuBar (new [] { MenuItems }) {
 				X = position.X,
 				Y = position.Y,
 				Width = 0,
 				Height = 0,
-				UseSubMenusSingleFrame = UseSubMenusSingleFrame
+				UseSubMenusSingleFrame = this.UseSubMenusSingleFrame
```
BDisp, line 119: `this` "Name can be simplified.". See https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0003-ide0009.

tig: Can you be more clear on what you are saying/suggesting? 

In my opinion, just because a "Name can be simplified" doesn't mean it SHOULD be simplified. I find code like line 119 to be frustrating to parse. The `this.` makes it clear it's not a bug (as I original thought) but intentional.

BDisp: Sorry @tig, I didn't say it was a bug but only inform the VS2022 advise for potential fixes. Forget please, thanks.

tig:
Also fixes #2126

PR #2123, opened 2022-10-21
Fixes #2120. TreeView: Hide cursor in single select mode by default, add cursor toggle to TreeViewFileSystem Scenario

### LINKED ISSUE #2120: TreeView: Cursor caret should not be visible
V1c1hcw 1

We've been slowly getting rid of the flashing caret on non-input Views. TreeView still does it.

tznind:
Fixes #2120- Add cursor toggle to TreeViewFileSystem Scenario

@tig regarding issue #2120, there is already a feature for this (`DesiredCursorVisibility`).  The default state of this is `Default` (i.e. it has a cursor).  This PR adds a menu item to the scenario that lets user turn cursor on and off.  

If you would like the default to be `Invisible` let me know and I can update this PR.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/TreeView.cs
+++ b/Terminal.Gui/Views/TreeView.cs
@@ -219,19 +219,23 @@ namespace Terminal.Gui {
 		/// <value></value>
 		public AspectGetterDelegate<T> AspectGetter { get; set; } = (o) => o.ToString () ?? "";
 
-		CursorVisibility desiredCursorVisibility = CursorVisibility.Default;
+		CursorVisibility desiredCursorVisibility = CursorVisibility.Invisible;
 
 		/// <summary>
-		/// Get / Set the wished cursor when the tree is focused
+		/// Get / Set the wished cursor when the tree is focused.
+		/// Only applies when <see cref="MultiSelect"/> is true.
+		/// Defaults to <see cref="CursorVisibility.Invisible"/>
 		/// </summary>
 		public CursorVisibility DesiredCursorVisibility {
-			get => desiredCursorVisibility;
+			get { 
+				return MultiSelect ? desiredCursorVisibility : CursorVisibility.Invisible;
+			}
 			set {
-				if (desiredCursorVisibility != value && HasFocus) {
-					Application.Driver.SetCursorVisibility (value);
-				}
-
 				desiredCursorVisibility = value;
+
+				if (desiredCursorVisibility != value && HasFocus) {
+					Application.Driver.SetCursorVisibility (DesiredCursorVisibility);
+				}
 			}
 		}
 
--- a/UICatalog/Scenarios/TreeViewFileSystem.cs
+++ b/UICatalog/Scenarios/TreeViewFileSystem.cs
@@ -25,6 +25,7 @@ namespace UICatalog.Scenarios {
 		private MenuItem miFullPaths;
 		private MenuItem miLeaveLastRow;
 		private MenuItem miCustomColors;
+		private MenuItem miCursor;
 		private Terminal.Gui.Attribute green;
 		private Terminal.Gui.Attribute red;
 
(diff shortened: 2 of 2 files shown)


tig:
What is the customer scenario where a cursor is useful on Treeview?

tznind:
> What is the customer scenario where a cursor is useful on Treeview?

Primary use case is during multi selection to determine where the active row is.  Even when multiple rows are selected there is still a 'primary' row which is the one tracked by `SelectedObject` and the one which the selection will be adjusted from if user uses Shift up/down.

You can see a similar effect in Visual Studio Code, but in this case the active row has a bright border.

activerowtree

I'm happy to go either way on this though.  Setting the default to `Invisible` could be a good change for new users.  The setting was originally added in response to #1559

tig:
> > What is the customer scenario where a cursor is useful on Treeview?
> 
> Primary use case is during multi selection to determine where the active row is. Even when multiple rows are selected there is still a 'primary' row which is the one tracked by `SelectedObject` and the one which the selection will be adjusted from if user uses Shift up/down.
> 
> You can see a similar effect in Visual Studio Code, but in this case the active row has a bright border.
> 
> activerowtree
> 
> I'm happy to go either way on this though. Setting the default to `Invisible` could be a good change for new users. The setting was originally added in response to #1559

Is there another scenario beyond multi-select where it's useful? Perhaps the right thing to do is only enable it when multi-select is active?

tznind:
> Is there another scenario beyond multi-select where it's useful? 

I can't really think another use case other than stylistic choice.  Also because there is a public `DesiredCursorVisibility` property then removing it would be a breaking change.  Changing its default value might also be considered breaking but is much more minor.

> Perhaps the right thing to do is only enable it when multi-select is active?

As I was describing the use case I did think of this but I dislike 'magic' properties (where changing one public property `MultiSelect` auto changes another `DesiredCursorVisibility`).  Especially you don't want :

```csharp
tv.DesiredCursorVisibility = CursorVisibility.Invisible;
tv.MultiSelect = true;
```
To behave differently from:
```csharp
tv.MultiSelect = true;
tv.DesiredCursorVisibility = CursorVisibility.Invisible;
```

Also relevant to this discussion is that MultiSelect defaults to true.

Let me know what you want to do, options seem to be:
- Leave unchanged
- Change `DesiredCursorVisibility` default to Invisible
- Remove `DesiredCursorVisibility` and render cursor based solely on `MultiSelect`
- Something else

BDisp:
@tznind see the `TextView` or the `TextField` where are conditions to show or hide the cursor. There is also a backup field where the cursor is restored (visible) accordingly with the user settings. What @tig suggest is only show the cursor if multi select is enabled and hide otherwise.

tznind:
> There is also a backup field where the cursor is restored (visible) accordingly with the user settings. What @tig suggest is only show the cursor if multi select is enabled and hide otherwise.

Ok I think I understand... do you mean that `DesiredCursorVisibility` becomes a setting that only applies when multi select is on? And if single select is on then the `DesiredCursorVisibility` value is ignored and the cursor is suppressed?

BDisp:
> Ok I think I understand... do you mean that `DesiredCursorVisibility` becomes a setting that only applies when multi select is on? And if single select is on then the `DesiredCursorVisibility` value is ignored and the cursor is suppressed?

Right, but take an account the the user may be configured the `DesiredCursorVisibility` to other than the default. Then you must use a private backup key to save the user setting before the change and use that backup key to restore it. For e.g. the user may be set a block visible cursor. Before you set to invisible you have to save the user setting in that field. Before restore to visible you have to set to the block cursor and not the default.

tznind:
Wouldn't it work just to change the `get` property:

```csharp
/// <summary>
/// Get / Set the wished cursor when the tree is focused.
/// Only applies when <see cref="MultiSelect"/> is true.
/// </summary>
public CursorVisibility DesiredCursorVisibility {
	get { 
		return MultiSelect ? desiredCursorVisibility : CursorVisibility.Invisible;
	}
[...]
```

BDisp:
Probably yes.

BDisp:
@tznind don't forget that call `Application.Driver.SetCursorVisibility (returnedValue)` or `Driver.SetCursorVisibility (returnedValue)` after you get the value to effectively apply the changes.

tznind:
Ok I made the following changes of behavior:

TreeViewFileSystem now has menu items for toggling:
- Cursor (on/off)
- MultiSelect (on/off)

The changes to TreeView are:
- Cursor now defaults to off (`Invisible`)
- Cursor stores but ignores any changes to `DesiredCursorVisibility` while MultiSelect is off and always uses `Invisible` in this state.

Cursor visibility is set in the `Enter` event of the control and the `set` on `DesiredCursorVisibility`.

tig (approved):

PR #2136, opened 2022-10-25
Fixes #2135. Character Map scenario doesn't show the content view.

### LINKED ISSUE #2135: Character Map scenario doesn't show the content view.
After the #2104 this issue comes up. In reality, if a `View` is dirty, it should be allowed to be cleared before redraw the sub views as reported by #2094. To avoids this issue all the views that inherit from the `View` class and overrides the `Redraw` method, which have a `View` container, must inherit a private nested `View` class, like `private class ContentView : View`. The reason why the `Character Map` scenario doesn't show the content view is because it doesn't use it and it's redrawn on the `DrawContent` event with the hack call to the `OnDrawContent` on the `SrollView.Redraw`. Then the call to the `contentView.Redraw (contentView.Frame)` cause it cleared causing the issue. The `CharMap` class should be inherit from View and use the `ScrollBarView` and not be inherit from `ScrollView`. But I won't change any of that.
Another bug that I also checked is the `ScrollView`'s `AutoHideScrollBars`, `ShowHorizontalScrollIndicator` and `ShowVerticalScrollIndicator` doesn't update the vertical and horizontal `ScrollBarView`.

imagem

BDisp:
Fixes #2135 - Only to remember that if a container view isn't to intended to be cleared on the `Redraw` method of the `View` class, then create a nested private class inherit from the `View` class.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/ScrollView.cs
+++ b/Terminal.Gui/Views/ScrollView.cs
@@ -29,7 +29,13 @@ namespace Terminal.Gui {
 	/// </para>
 	/// </remarks>
 	public class ScrollView : View {
-		View contentView = null;
+		private class ContentView : View {
+			public ContentView (Rect frame) : base (frame)
+			{
+			}
+		}
+
+		ContentView contentView;
 		ScrollBarView vertical, horizontal;
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


tig (approved):
Nice find.

PR #2152, opened 2022-11-01
Fixes #2140. OnSelectedChange is being called from OnEnter when the selection didn't actually change.

### LINKED ISSUE #2140: ListView: OnSelectedChanged is called when it didn't actually change (OnEnter)
image

Because of this, there's no way for an override to know if the selection REALLY changed. 

In UI Catalog, this means that whenever the focus returns to the Category ListView, the Secenario ListView gets regenerated, causing IT to lose what Scenario was previously selected:

00ibvTI 1

I believe the call to `OnSelectedChanged` in `OnEnter` should be removed.

@BDisp - what do you think?

tig:
Fixes #2140. 

OnSelectedChange is being called from OnEnter when the selection didn't actually change.

The fix broke `ComboBox` - this test: `HideDropdownListOnClick_True_Highlight_Current_Item ()`

image

> System.Exception : Colors used did not match expected at row 1 and col 0 (indexes start at 0).  Color index used was Black,Black but test expected Black,Black (these are indexes into the expectedColors array)

Which is confusing to me...

### DIFF
--- a/Terminal.Gui/Core/CollectionNavigator.cs
+++ b/Terminal.Gui/Core/CollectionNavigator.cs
@@ -1,0 +1,244 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Terminal.Gui {
+	/// <summary>
+	/// Navigates a collection of items using keystrokes. The keystrokes are used to build a search string. 
+	/// The <see cref="SearchString"/> is used to find the next item in the collection that matches the search string
+	/// when <see cref="GetNextMatchingItem(int, char)"/> is called.
+	/// <para>
+	/// If the user types keystrokes that can't be found in the collection, 
+	/// the search string is cleared and the next item is found that starts with the last keystroke.
+	/// </para>
+	/// <para>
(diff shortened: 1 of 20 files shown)


BDisp:
Ok @tig .

tznind:
> Black,Black but test expected Black,Black 

Black was not black enough! :goth: 

Maybe we should change this test to do 'by value' rather than 'by reference' comparison on those.  Unless it is that the error message is bugged itself.

That said it could just be an issue with the test itself:
image

BDisp:
I can't see it right now, but it seems that the background is cleared after it was redraw.

BDisp:
The extra 0 is the down arrow. I already realized that the errors messages are bugged. The only valid information I get is the row, col and if the color already exists in the variable. But I think that it should inform what the color is in the test and what color it should were be.

BDisp:
The fix is here https://github.com/gui-cs/Terminal.Gui/issues/2156#issuecomment-1299371831.

BDisp:
@tig I submitted a PR in your branch https://github.com/tig/Terminal.Gui/pull/4.

PR #2155, opened 2022-11-01
TreeView: Now highlights only model text

### LINKED ISSUE #2151: TreeView: Mouse wheel scrolls infinitely beyond bottom of items
1PH3RZ7 1

1) Run `Class Explorer` in UI Catalog.
2) Using mouse wheel scroll down. Keep going.

Expected: Scrolling would stop either here:
image

or here (possibly up for debate; ListView does this):
image

(ListView for reference):
image

tig:
This PR:

- Adds new `Style` named `HighlightModelTextOnly` - false by default
- Updates `Class Explorer` Scenario to have a menu for testing new Style
- Fixes #2151 

V7eyDAo 1

See: https://github.com/gui-cs/Terminal.Gui/discussions/2154

### DIFF
--- a/Terminal.Gui/Core/Trees/Branch.cs
+++ b/Terminal.Gui/Core/Trees/Branch.cs
@@ -89,10 +89,9 @@ namespace Terminal.Gui.Trees {
 		public virtual void Draw (ConsoleDriver driver, ColorScheme colorScheme, int y, int availableWidth)
 		{
 			// true if the current line of the tree is the selected one and control has focus
-			bool isSelected = tree.IsSelected (Model) && tree.HasFocus;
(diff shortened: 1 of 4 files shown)


tig:
@tznind please review!

tznind:
Looks great, thanks for this and for making it toggleable <3.

Only issue I can see is that the highlight color does not match ListView... seems to be Blue on White instead of Black on White.  Seems to be a regression vs main gui.cs branch.

tznind:
Also I think we should add the Style setting in here too.  That Scenario has all the toggle able options so should have the new one too.

There is a bit of redundancy in Class Explorer and TreeViewFileSystem.  I think later on we could look to share more code between them or scrap one/merge.

treestyle

PR #2158, opened 2022-11-02
Fixes #2156 - BREAKING CHANGE - Fixes IsOverridden and makes it internal

### LINKED ISSUE #2156: `View.IsOverridden` doesn't work
I wrote these unit tests (ViewTests.cs) to prove it.

```csharp
		[Fact, AutoInitShutdown]
		public void IsOverridden_False_IfNotOverriden ()
		{
			var view = new DerivedView () { Text = "DerivedView does not override MouseEvent", Width = 10, Height = 10 };
			
			Assert.False (view.IsOverridden (view, "MouseEvent"));
		}

		[Fact, AutoInitShutdown]
		public void IsOverridden_True_IfOverriden ()
		{
			var view = new Button () { Text = "Button overrides MouseEvent", Width = 10, Height = 10 };

			Assert.True (view.IsOverridden (view, "MouseEvent"));
		}
```

image

@BDisp - Where'd you get this code from? 

```csharp
		/// <summary>
		/// Check if the <paramref name="method"/> is overridden in the <paramref name="view"/>.
		/// </summary>
		/// <param name="view">The view.</param>
		/// <param name="method">The method name.</param>
		/// <returns><see langword="true"/> if it's overridden, <see langword="false"/> otherwise.</returns>
		public bool IsOverridden (View view, string method)
		{
			Type t = view.GetType ();
			MethodInfo m = t.GetMethod (method);

			return (m.DeclaringType == t || m.ReflectedType == t) && m.GetBaseDefinition ().DeclaringType == typeof (Responder);
		}
```

tig:
Fixes #2156

BREAKING CHANGE - 

This PR 
- Fixes IsOverridden(), which actually didn't work 
- Moves IsOverridden to Responder making it both static and internal.
- Adds unit tests proving it works correctly now

I used [cs.github.com](https://cs.github.com/?scopeName=All+repos&scope=&q=%22IsOverridden%22+AND+%22Terminal.GUI%22+language%3AC%23) to gain confidence that no-one outside of core code uses this function. 

If someone wants to use it, they can code it up themselves... It is very bad OO practice to do something like this and we should not encourage it. In v2 we should figure out how to restructure things so it can be completely removed.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -13,7 +13,7 @@ jobs:
       uses: actions/checkout@v2
 
     - name: Setup .NET Core
-      uses: actions/setup-dotnet@v3.0.2
+      uses: actions/setup-dotnet@v3.0.3
       with:
         dotnet-version: 6.0.100
     
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -15,7 +15,7 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: Setup .NET Core
-      uses: actions/setup-dotnet@v3.0.2
+      uses: actions/setup-dotnet@v3.0.3
       with:
         dotnet-version: 6.0.100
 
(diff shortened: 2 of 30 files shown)


tig:
Since other PRs are dependent on this, I'm merging it now. We can back it out if someone has problems with how I did it.

Issue #2163, opened 2022-11-03
The last scenario executed is not kept selected on exit.


BDisp:
@tig just a reminder so you don't forget to correct this, please.

tig:
I did fix it. And then I some how unfixed it AGAIN.

I sometimes wonder... ;-)

tig:
Fixed by #2164

PR #2164, opened 2022-11-03
Fixes #2165 - ListView: OnSelectedChanged is called when it didn't actually change (again)

tig:
Fixes #2136 and #2140 

Reapplied fixes from 26eabdda6c178329d8cd391e1b0200383b4a1e06 which where somehow reverted from #2152


tig:
Fixes #2163 (had a typo (2136)) before.

PR #2165, opened 2022-11-03
Fixes #2162. BackgroundWorker Scenario broke

### LINKED ISSUE #2162: BackgroundWorker Collection scenario is broken.
Some changes in the `UICatalog` caused it to broken,

tig:
Fixes #2162 

Beyond fixing the `BackgroundWorkerCollection` scenario, this PR enhances the API docs for `Run<T> (Func<Exception, bool> errorHandler = null) where T : Toplevel, new()` to note that `Init` is not needed if it's not already been called.

### DIFF
--- a/Terminal.Gui/Core/Application.cs
+++ b/Terminal.Gui/Core/Application.cs
@@ -39,6 +39,7 @@ namespace Terminal.Gui {
 	/// };
 	/// Application.Top.Add(win);
 	/// Application.Run();
+	/// Application.Shutdown();
 	/// </code>
 	/// </example>
 	/// <remarks>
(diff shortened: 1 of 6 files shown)


tig:
I can repro the double menu thing by 

1) Starting Background Worker
2) Clicking View->Worker App
3) Dragging the Worker Collection Log

image

tig:
Also, when running BGW, the cursor on the menu bar is initially visible:

1ZMT1dB 1

This doesn't happen in other Scenarios

BDisp:
> This doesn't happen in other Scenarios

Because the menu bar has the focus. But before the cursor was hidden.

PR #2166, opened 2022-11-03
Fixes 2144. Add half-height bottom shadow for Effect3D.

### LINKED ISSUE #2144: Norton Commander style buttons: Add half-height bottom shadow for Effect3D
Feature Request I guess

I'd like to add buttons like Norton Commander used back in the day. They had a half-height shadow at the bottom. But currently with Terminal.GUI it only seems to be possible to add a full-height shadow at the bottom. This is kind of fat for a small button.

Here's what it should look like:
image

I looked at all the border samples in UICatalog but did not see such a feature. It might be impossible. Or is there some rune to configure? I'm not so deep in the code.

None.

**Additional context**
-

BDisp:
Fixes #2144 - Add half shadow when view height is equal to one to the `Border` class. Added animation 3D effects to the `Button` and `Panel`.

## Pull Request checklist:

animated-3d-shadow


tig:
This is BEAUTIFUL!

Curious if you tested with an assortment of fonts? Do those block glyphs exist in most fonts?

BDisp:
> This is BEAUTIFUL!
> 
> Curious if you tested with an assortment of fonts? Do those block glyphs exist in most fonts?

Thanks. No, I haven't test with others fonts. I also don't know if exist in most fonts, but is probable. @tznind in this https://github.com/gui-cs/Terminal.Gui/issues/2144#issuecomment-1297139569 comment describe some more glyphs.

I converted this to draft because I'll allow the half shadow by set a `UseHalfEffect3D` bool property and not only if the height is equal to one.
The @tznind suggestion about `ShadowStyler` may be used to add different glyphs for each style. I don't know.

tig:
> > This is BEAUTIFUL!
> > Curious if you tested with an assortment of fonts? Do those block glyphs exist in most fonts?
> 
> Thanks. No, I haven't test with others fonts. I also don't know if exist in most fonts, but is probable. @tznind in this [#2144 (comment)](https://github.com/gui-cs/Terminal.Gui/issues/2144#issuecomment-1297139569) comment describe some more glyphs.
> 
> I converted this to draft because I'll allow the half shadow by set a `UseHalfEffect3D` bool property and not only if the height is equal to one. The @tznind suggestion about `ShadowStyler` may be used to add different glyphs for each style. I don't know.

I pulled this down and played with tons of fonts. They all worked great!

tznind:
> I pulled this down and played with tons of fonts. They all worked great!

From my experimentation and reading, the half block elements seem pretty widely supported (probably about as supported as HLine etc).  There are some other block sizes like 'three quarter block'  etc that are less supported but we are not using them so it should be ok!

tig:
I modified `Button.cs` to set `UseEffect3DAnimation = true` by default.

Shows some issues:

1) Buttons on bottom row of a view look broken:

image

2) The weird extra horizontal bar at top-right really shows up:

image

I'm going to submit a PR to this PR with:

1) Changes to UI Catalog that enable setting this new style globally for testing purposes
2) Proposed changes to `Dialog` and `MessageBox` that do the right thing (mostly) if the setting is true

I'd LOVE for us to figure out how to make this style the DEFAULT without breaking apps. This will be very hard.

One suggestion in that direction is to have the `Height` and `Width` of `Button` be `+1` if the `UseEffect3DAnimation` is `true`. I tried this quickly and it broke a lot of things, but it may be made to work!

nG8FpRH 1

BDisp:
Yes, initially I set as default but I've seen that the views must have the necessary space to accommodate it and I reverted that.
I haven't seen the horizontal weird bar in my computer and I don't know why.

tig:
This is odd. 

If `UseEffect3DAnimation` is true, moving buttons after created doesn't work anymore!

Off:
utZr9Do 1

On:
Kk0sNuN 1

BDisp:
Damn, I think that buttons are changing the Pos.X in each click and the Button is changing to the original position. How can I resolve this out?

tig:
> Damn, I think that buttons are changing the Pos.X in each click and the Button is changing to the original position. How can I resolve this out?

Instead of moving the button in `OnClicked`, instead set a flag that causes it to paint differently in `Redraw`?

BDisp:
> Instead of moving the button in `OnClicked`, instead set a flag that causes it to paint differently in `Redraw`?

No, because the Button is redraw by the base View class. I think the better option is when is returning from the mainloop if the X is different to the x backed field then the new X must be used, capiche :-)

tig:
The more I think about it, the more I'm convinced the architecture you chose for this isn't going to scale:

- You are treating the effect like the 3Deffect for `Window`. In that case, there's nothing wrong with the 'shadow' impacting other views.
- But on Button
   - a) A user action (click) causes the button to move the right 
   - b) Buttons are often at the very bottom of a view
- You keep the `Frame` of `Button` the same size (by default only 1 row high, and "text width" wide)

I wonder if a better solution would be to make buttons with this style be bigger by one row and column, and have the effect be drawn within the button itself.

tig:
Another issue:

Right now, when a user clicks, you "move" the button 1 column to the right (or left). But the "shadow" implies the button will move 1 column DOWN as well as RIGHT.  This is how "3D buttons" in Turbo Pascal worked.

BDisp:
> I wonder if a better solution would be to make buttons with this style be bigger by one row and column, and have the effect be drawn within the button itself.

Is doable, but how to surround by the [ ] in multi line. Label can be used as button as well. But we can't reduce the 3d effect more than now.

tig:
Another:

The animation should happen on `ButtonPressed` (presses down) and `ButtonReleased` (rises up), not `ButtonClicked` (presses down/rises up).

tig:
> > I wonder if a better solution would be to make buttons with this style be bigger by one row and column, and have the effect be drawn within the button itself.
> 
> Is doable, but how to surround by the [ ] in multi line. Label can be used as button as well. But we can't reduce the 3d effect more than now.

IMO, if the shadow is there, and the button background color is different than the host view, the `[ ]` can be nuked (thank you!). (Just like TurboPascal).

tig:
Actually TP just moves the button to the right (and hides the shadow) during `ButtonPressed`:

r0OzhIa 1

But note that the "clicked" event does NOT happen until after `ButtonReleased` happens with the mouse still over the button.

BDisp:
> Another:
> 
> The animation should happen on `ButtonPressed` (presses down) and `ButtonReleased` (rises up), not `ButtonClicked` (presses down/rises up).

Well personally I like the option when I press down on a button and I regret then I releases the button outside the control to not execute it. Then my propose is on `ButtonPressed` (presses down) and `ButtonReleased` (rises up). If a `ButtonClicked` is raised then executes the action. What do you thinks?

BDisp:
> Actually TP just moves the button to the right (and hides the shadow) during `ButtonPressed`:

The move direction depend on the 3d effect offset, no big deal. The default is move to the right and the user could choose what he want.



> But note that the "clicked" event does NOT happen until after `ButtonReleased` happens with the mouse still over the button.

Yes, the `ButtonReleased` rises up but it's only executed if the mouse is over the button.

tznind:
> Yes, the `ButtonReleased` rises up but it's only executed if the mouse is over the button.

Yes this is a bit of a challenge.  The `Button` doesn't get the released event if it is outside the buttons area.

We could hook the root mouse event.

```
// on mouse down += TrackMouse;

// TrackMouse (handler)
//{
//   // detect any reports from mouse where button is not down
//  // on mouse down -= TrackMouse  
//  // flip mouse to not down
// //  trigger Click event
//}
```

But it feels a bit hacky

BDisp:
> > Yes, the `ButtonReleased` rises up but it's only executed if the mouse is over the button.
> 
> Yes this is a bit of a challenge. The `Button` doesn't get the released event if it is outside the buttons area.
> 
> We could hook the root mouse event.
> 
> ```
> // on mouse down += TrackMouse;
> 
> // TrackMouse (handler)
> //{
> //   // detect any reports from mouse where button is not down
> //  // on mouse down -= TrackMouse  
> //  // flip mouse to not down
> // //  trigger Click event
> //}
> ```
> 
> But it feels a bit hacky

With mousegrabview.

BDisp:
Another thing his how the behavior by keyboard. The Linux doesn't have the key up/down events.

tznind:
Ah is that what mousegrabview does, I had seen that before when looking at menu/mouse code but never deep enough to realize what it actually did.

That's handy I will have to remember that one thanks.

Terminal.Gui/Views/PanelView.cs
```diff
@@ -246,5 +302,31 @@ public override void Redraw (Rect bounds)
...
+		/// <summary>
+		/// Virtual method to invoke the <see cref="Clicked"/> event.
+		/// </summary>
+		public virtual void OnClicked ()
+		{
+			if (UseEffect3DAnimation) {
+				Border.Effect3D = false;
+				var x = X;
+				if (Border.Effect3DOffset.X > 0) {
+					X += 1;
```
tznind, line 315: Just want to echo @tig comment here that we don't want to physically move the Button here.  What if there is another View with `Pos.Right(myButton) + 1` it would judder with the button wouldn't it?  

I think we want a render effect not a position effect.

BDisp: The `Mainloop.Invoke` will not be used now. It will be moved down on button pressed, moved up on button released and only after the animation the action is executed and will not interfere with the what the action does.

BDisp: > I think we want a render effect not a position effect.

Without a position effect there isn't a true animation. I tried by only clear the shadow but the effect it isn't so good. 

> What if there is another View with `Pos.Right(myButton) + 1` it would judder with the button wouldn't it?

It will only be side by side. But I recommend using more space between the controls, of course.

tznind: I agree that it should 'appear to move' just that it's `X` should not change as that could cause problems for other controls' layouts (e.g. if they use relative positioning).

BDisp: Ok I decided not moving the button but only hide and show the shadow. So, this will avoid issue with relative layout positions used by others controls.

BDisp:
Pressing mouse down will clear the shadow and releasing will show the shadow again. The `Clicked` event will only run if the mouse is over the button on releasing. There is still some issue with the glyps with misalignment. I'm using the the char for left and right. The normal color should be different than the superview normal color to be more visible.

animated-3d-shadow

BDisp:
Sometimes I get this error when I merge from develop branch and push to my fork. It's the first time I get in a PR. I already run all the tests on Windows and Linux without errors. So I think is a false fail. Well, sending a D6 (6) with shift, alt and control as false the result only could be D6 and not N.

tznind:
> Sometimes I get this error when I merge from develop branch and push to my fork. It's the first time I get in a PR. I already run all the tests on Windows and Linux without errors. So I think is a false fail. Well, sending a D6 (6) with shift, alt and control as false the result only could be D6 and not N.


I think I am also getting this failure:

```
[xUnit.net 00:01:14.50]     Terminal.Gui.ConsoleDrivers.ConsoleDriverTests.TestVKPacket(unicodeCharacter: '6', shift: False, alt: False, control: False, initialVirtualKey: '6', initialScanCode: 7, expectedRemapping: D6, expectedVirtualKey: '6', expectedScanCode: 7) [FAIL]
  Failed Terminal.Gui.ConsoleDrivers.ConsoleDriverTests.TestVKPacket(unicodeCharacter: '6', shift: False, alt: False, control: False, initialVirtualKey: '6', initialScanCode: 7, expectedRemapping: D6, expectedVirtualKey: '6', expectedScanCode: 7) [13 ms]
  Error Message:
   Assert.Equal() Failure
Expected: D6
Actual:   N
  Stack Trace:
     at Terminal.Gui.ConsoleDrivers.ConsoleDriverTests.<>c__DisplayClass20_0.<TestVKPacket>b__0(KeyEventEventArgs e) in D:\Repos\gui.cs\UnitTests\ConsoleDriverTests.cs:line 682
   at Terminal.Gui.View.ProcessKey(KeyEvent keyEvent) in D:\Repos\gui.cs\Terminal.Gui\Core\View.cs:line 1692
   at Terminal.Gui.Toplevel.ProcessKey(KeyEvent keyEvent) in D:\Repos\gui.cs\Terminal.Gui\Core\Toplevel.cs:line 411
   at Terminal.Gui.Application.ProcessKeyEvent(KeyEvent ke) in D:\Repos\gui.cs\Terminal.Gui\Core\Application.cs:line 438
   at Terminal.Gui.FakeDriver.ProcessInput(ConsoleKeyInfo consoleKey) in D:\Repos\gui.cs\Terminal.Gui\ConsoleDrivers\FakeDriver\FakeDriver.cs:line 441
   at Terminal.Gui.FakeDriver.<PrepareToRun>b__50_0(ConsoleKeyInfo consoleKey) in D:\Repos\gui.cs\Terminal.Gui\ConsoleDrivers\FakeDriver\FakeDriver.cs:line 413
   at Terminal.Gui.FakeMainLoop.Terminal.Gui.IMainLoopDriver.MainIteration() in D:\Repos\gui.cs\Terminal.Gui\ConsoleDrivers\FakeDriver\FakeMainLoop.cs:line 99
   at Terminal.Gui.MainLoop.MainIteration() in D:\Repos\gui.cs\Terminal.Gui\Core\MainLoop.cs:line 307
   at Terminal.Gui.Application.RunMainLoopIteration(RunState& state, Boolean wait, Boolean& firstIteration) in D:\Repos\gui.cs\Terminal.Gui\Core\Application.cs:line 1056
   at Terminal.Gui.Application.RunLoop(RunState state, Boolean wait) in D:\Repos\gui.cs\Terminal.Gui\Core\Application.cs:line 1038
   at Terminal.Gui.Application.Run(Toplevel view, Func`2 errorHandler) in D:\Repos\gui.cs\Terminal.Gui\Core\Application.cs:line 1214
   at Terminal.Gui.Application.Run(Func`2 errorHandler) in D:\Repos\gui.cs\Terminal.Gui\Core\Application.cs:line 1152
   at Terminal.Gui.ConsoleDrivers.ConsoleDriverTests.TestVKPacket(UInt32 unicodeCharacter, Boolean shift, Boolean alt, Boolean control, UInt32 initialVirtualKey, UInt32 initialScanCode, Key expectedRemapping, UInt32 expectedVirtualKey, UInt32 expectedScanCode) in D:\Repos\gui.cs\UnitTests\ConsoleDriverTests.cs:line 696

Failed!  - Failed:     1, Passed:  1045, Skipped:     0, Total:  1046, Duration: 1 m 27 s - UnitTests.dll (net6.0)
```

BDisp:
The problem is I can't reproduce the error on both Windows/Linux to understand what is causing check workflow throwing an exception.

tig:
We need to retarget this PR at the `v2_develop` branch.

tig:
We'll target this at v2. I'm taking the requirements of this into account in my redesign of `View` for v2, so the code in this PR is likely not useful.

PR #2167, opened 2022-11-03
Fixes #520. Application.Init / Shutdown are confused about TopLevels created by Application.Begin and not cleaned up by Application.End

### LINKED ISSUE #520: `Application.Init` / `Shutdown` are confused about `TopLevel`s created by `Application.Begin` and not cleaned up by `Application.End`
In `UI Catalog`, before I start a `Scenario` I call `Application.Shutdown`. You are supposed to book end every call to `Applciation.Begin` with `Shutdown`. 

I just discovered that `Shutdown` does not actually clean up. As a result, the next time `Application.Begin` is called, fields like `topLevels` still have data. This means that the 'old' stuff interferes with the 'new' stuff.

Short-term fix is to uninitialize stuff:

```csharp
		/// <summary>
		/// Shutdown an application initalized with <see cref="Init()"/>
		/// </summary>
		public static void Shutdown ()
		{
			Driver.End ();
			foreach (var t in toplevels) {
				t.Running = false;
			}
			toplevels.Clear ();
			Current = null;
			CurrentView = null;
			Top = null;
			MainLoop = null; 
			_initialized = false;
		}
```

Crux of issue (added Nov 2, 2022 via #2162):
 
Init/Shutdown are supposed to clean up. But in `ResetState`, the loop that disposes of the `topLevels` does nothing because the `Toplevel` that was created by `Init` has not been added to `topLevels` (it gets added by `Begin`).

The right fix (IMO) would be to move all that state into `RunState` and leverage the fact that `RunState` is already `IDisposable`. However, it may not be possible to do this without a breaking change.

tig:
Fixes #520 - Application.Init / Shutdown are confused about TopLevels created by Application.Begin and not cleaned up by Application.End

This PR Is pretty massive:
- Refactoring of UI Catalog to 
  - Remove `Scenario.Top` 
  - Provide better debuggability of the resources used by the main app and Scenarios
  - Additional Dispose checking (for RunState)
  - EVERY Scenario was updated as part of this PR
- Tons of new unit tests for aspects of the Application, esp RunState
- Refactoring of Unit Test to make more manageable (ApplicationTest.cs was out of control)

### DIFF
--- a/Terminal.Gui/Core/Application.cs
+++ b/Terminal.Gui/Core/Application.cs
@@ -329,11 +329,6 @@ namespace Terminal.Gui {
 		/// into a single call. An applciation cam use <see cref="Run{T}(Func{Exception, bool}, ConsoleDriver, IMainLoopDriver)"/> 
 		/// without explicitly calling <see cref="Init(ConsoleDriver, IMainLoopDriver)"/>.
 		/// </para>
-		/// <remarks>
-		/// Note, due to Issue #520, if Init is called and <see cref="Begin(Toplevel)"/> is not called, the <see cref="Toplevel"/>
-		/// created by this function will NOT be Disposed when <see cref="Shutdown"/> is called. Call Dispose() on <see cref="Application.Top"/> 
-		/// before calling <see cref="Shutdown"/> if <see cref="Begin(Toplevel)"/> (and <see cref="End(RunState)"/>) has not been called.
-		/// </remarks>
 		/// <param name="driver">The <see cref="ConsoleDriver"/> to use. If not specified the default driver for the
 		/// platform will be used (see <see cref="WindowsDriver"/>, <see cref="CursesDriver"/>, and <see cref="NetDriver"/>).</param>
 		/// <param name="mainLoopDriver">Specifies the <see cref="MainLoop"/> to use.</param>
@@ -887,8 +882,6 @@ namespace Terminal.Gui {
 			}
 
 			var rs = new RunState (toplevel);
-			// Fixes #520 - Begin should be decoupled from Init
-			//Init ();
 			
 			if (toplevel is ISupportInitializeNotification initializableNotification &&
 			    !initializableNotification.IsInitialized) {
@@ -902,7 +895,7 @@ namespace Terminal.Gui {
 			lock (toplevels) {
 				// If Top was already initialized with Init, and Begin has never been called
 				// Top was not added to the toplevels Stack. It will thus never get disposed.
-				// Clean it up here (fixes #520).
+				// Clean it up here:
 				if (Top != null && toplevel != Top && !toplevels.Contains(Top)) {
 					Top.Dispose ();
 					Top = null;
@@ -1052,7 +1045,6 @@ namespace Terminal.Gui {
 			}
 			toplevels.Clear ();
 			Current = null;
-			// Fixes #520: Dispose Top
 			Top?.Dispose ();
 			Top = null;
 
--- a/Terminal.Gui/Core/Application.cs
+++ b/Terminal.Gui/Core/Application.cs
@@ -39,6 +39,7 @@ namespace Terminal.Gui {
 	/// };
 	/// Application.Top.Add(win);
 	/// Application.Run();
+	/// Application.Shutdown();
 	/// </code>
 	/// </example>
 	/// <remarks>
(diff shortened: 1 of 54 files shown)


UnitTests/ApplicationTests.cs
```diff
@@ -120,31 +253,63 @@ public void Begin_End_Cleana_Up ()
...
+
+			Shutdown ();
+
+			Assert.Null (Application.Top);
+			Assert.Null (Application.MainLoop);
+			Assert.Null (Application.Driver);
+		}
 
+		[Fact]
+		public void Run_T_NoInit_ThrowsInDefaultDriver ()
```
BDisp, line 276: This test will fail on `Linux` because the `CursesDriver` will not throw any exception and stay waiting for some keystroke. Maybe checking the operation system. Or maybe you have to review the `Run<T>`.

tig: I had a feeling... I should have run in WSL. My bad.

tig:
I just did some comment & test cleanup.

We have one issue remaining that I'm aware of before this can be merged: 

The broken ` ToplevelTests.KeyBindings_Command_With_MdiTop() line 489` test that @BDisp  is looking into.

tig:
I know y'all are busy, but this is a massive PR that I'd love merged for the weekend. I don't feel right merging it without other eyes on the code and running it locally. Thanks.

tig:
@BDisp any idea why me merging your 
[Fixes KeyBindings_Command_With_MdiTop unit test.](https://github.com/gui-cs/Terminal.Gui/pull/2167/commits/5472c9469e468e78c40c86629d411fe228783548) PR has broken the unit tests when run on linux? Very odd!

BDisp:
> @BDisp any idea why me merging your [Fixes KeyBindings_Command_With_MdiTop unit test.](https://github.com/gui-cs/Terminal.Gui/pull/2167/commits/5472c9469e468e78c40c86629d411fe228783548) PR has broken the unit tests when run on linux? Very odd!

Because `Assert.True (win1.IsMdiChild);` was failing here by returning false. Given your are now disposing and set to null a `Top` that isn't added to the toplevels stack, then I have to added `Application.Begin (top);` to avoid the error.

tig:
> > @BDisp any idea why me merging your [Fixes KeyBindings_Command_With_MdiTop unit test.](https://github.com/gui-cs/Terminal.Gui/pull/2167/commits/5472c9469e468e78c40c86629d411fe228783548) PR has broken the unit tests when run on linux? Very odd!
> 
> Because `Assert.True (win1.IsMdiChild);` was failing here by returning false. Given your are now disposing and set to null a `Top` that isn't added to the toplevels stack, then I have to added `Application.Begin (top);` to avoid the error.

Sorry, I'm confused. The branch I tested on Debian has this `Application.Begin (top);` fix in it.

BDisp:
> Sorry, I'm confused. The branch I tested on Debian has this `Application.Begin (top);` fix in it.

No problem. That why I don't like to work with separated repos on various OS. I prefer always to work on the same repo on `Windows` and shared on `WSL`. I know it isn't recommended but save me a lot of time because I'm only changing on one local no mater where I'm testing. That way I have all updated.

tig:
> > Sorry, I'm confused. The branch I tested on Debian has this `Application.Begin (top);` fix in it.
> 
> No problem. That why I don't like to work with separated repos on various OS. I prefer always to work on the same repo on `Windows` and shared on `WSL`. I know it isn't recommended but save me a lot of time because I'm only changing on one local no mater where I'm testing. That way I have all updated.

This was not the issue. 

The issue was my stupidity (and poor memory) in the test that verifies calling `Run<T>` without a valid driver and without calling `Init` first, trying to depend on the default driver to assert. Worked on Windows but (as you noted, and I forgot) fails on Linux. 

I now (finally) have the remote WSL Debugging stuff setup on my Windows box and can debug the unit tests on Linux (thanks for your help on that!). 

I fixed this by adding a proper check and exception to `Run<T>`.

Terminal.Gui/Core/Application.cs
```diff
@@ -930,15 +1039,17 @@ static void ResetState ()
...
 			foreach (var t in toplevels) {
 				t.Running = false;
 				t.Dispose ();
 			}
 			toplevels.Clear ();
 			Current = null;
+			Top?.Dispose ();
 			Top = null;
 
+			// BUGBUG: MdiTop is not cleared here, but it should be?
```
BDisp, line 1051: `MdiTop` is a read only property which check if `Top.IsMdiContainer` and return `Top` or null. So it cannot be set. Maybe the comment can be removed.

Issue #86, opened 2018-05-15
Windows: damageRegion support


migueldeicaza:
The following patch adds damage region support:

https://gist.github.com/migueldeicaza/852a6b03b828a045f1fcf79d1ab8b331

migueldeicaza:
Implemented

PR #2183, opened 2022-11-05
Fixes #2182. Fix output of non-BMP Unicode characters in NetDriver.

### LINKED ISSUE #2182: Non-BMP characters display incorrectly with System Console driver
**Describe the bug**
Non-BMP characters display as whatever character is equal to their least-significant 16 bits.

**To Reproduce**
1. Add a non-BMP character such as "𝔹" to a label.
2. See that instead "픹" (or such) is displayed.

**Expected behavior**
The correct character is displayed.

 - OS: Linux
 - using System Console driver

Patch will follow shortly.

cpacejo:
Fixes #2182.  Breaks non-BMP Unicode characters into their surrogate pairs to work properly with System Console.

Widths for the test characters are wrong, but this is due to unrelated issue with NStack [#86](https://github.com/gui-cs/NStack/issues/86).

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/NetDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/NetDriver.cs
@@ -1483,7 +1483,13 @@ namespace Terminal.Gui {
 							output.Append (WriteAttributes (attr));
 						}
 						outputWidth++;
-						output.Append ((char)contents [row, col, 0]);
+						var rune = contents [row, col, 0];
+						char [] spair;
+						if (Rune.DecodeSurrogatePair((uint) rune, out spair)) {
+							output.Append (spair);
+						} else {
+							output.Append ((char)rune);
+						}
 						contents [row, col, 2] = 0;
 					}
 				}
--- a/UICatalog/Scenarios/Unicode.cs
+++ b/UICatalog/Scenarios/Unicode.cs
@@ -97,7 +97,7 @@ namespace UICatalog.Scenarios {
 
 			label = new Label ("RadioGroup:") { X = Pos.X (label), Y = Pos.Bottom (listView) + 1 };
 			Win.Add (label);
-			var radioGroup = new RadioGroup (new ustring [] { "item #1", gitString, "Со_хранить" }, selected: 0) {
+			var radioGroup = new RadioGroup (new ustring [] { "item #1", gitString, "Со_хранить", "𝔽𝕆𝕆𝔹𝔸ℝ" }, selected: 0) {
 				X = 20,
 				Y = Pos.Y (label),
 				Width = Dim.Percent (60),


BDisp:
@cpacejo I appreciate your contribution and for discovering of this issue. Before I make the changes in the NStack repo, the `CharaterMap` scenario in the `UICatalog`, which I ask you to test there your PR, was rendering very badly and with incorrect buffer size. At the time I realize that were a pattern where the bad rendering was happens. The pattern I found is in the range of the changes I have made, which was the least significant 16 bits. The `gethexaformat` or bitwise `c & 0xFFFF` is badly in the `MakePrintable` method. With your issue it seems that it fall into surrogate pairs which the resulting rune by itself is a wide char, but where reading as sequential separated runes result in a single width.
You don't need to do that changes in the `UpdateScreen` method, because the bug is here, which affect all the drivers and in the `AddRune` of each driver it must be added each surrogate pair into each separated column:

https://github.com/gui-cs/Terminal.Gui/blob/b6dceb304844d95373c1d1ef6d02c100a3e1c62e/Terminal.Gui/Core/ConsoleDriver.cs#L697-L707

After I did the changes above and considering theses runes as single char the saga continues, see image bellow. Is the same problem that I was getting before, but now the right char is printed, thanks to you:

imagem

If you want I can submitted a PR to your branch and maybe we could work together in this fix. Let me know, please.

BDisp:
Somebody know how to write surrogate pairs with `Curses.addwstr`?

I already tried with the below but none print anything. Maybe it doesn't supported:
```cs
Curses.addwstr (new string (spair));
Curses.addwstr ($"'{spair [0]}''{spair [1]}'");
Curses.addwstr ($"\'{spair [0]}'\'{spair [1]}'");
Curses.addwstr ($"\\'{spair [0]}'\\'{spair [1]}'");
```

cpacejo:
Thanks for looking into this.

I don't understand `var controlChars = c & 0xFFFF;` -- why should, for example, U+10009 (a Linear B character, "𐀉") be treated the same as U+0009 (TAB)?

> in the AddRune of each driver it must be added each surrogate pair into each separated column

Non-BMP characters aren't necessarily wide -- if you split them into two separate columns, I think that would make those two columns print as only one (narrow) character width on screen, making the line too short.  Just like in the screen shot you show.

I think it makes sense to store the full code point (even if it is non-BMP) in a single column, just like is already done.  The code point just needs to be sent to the console driver in a way that the driver can understand.  `System.Text.StringBuilder` and `System.Console` expect a `char`, which (in .NET) is a [16-bit type which represents a UTF-16 character](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/char), hence the Unicode code point must be encoded as UTF-16 (using surrogate pairs) to be sent to it.

However on Linux in C, `wchar_t` (as used by curses) [is a 32-bit UTF-32 value](http://www.firstobject.com/wchar_t-string-on-linux-osx-windows.htm), which means surrogate pairs cannot be used (they exist only for UTF-16).  I tested just now in a C program, `addwstr` works correctly with UTF-32 characters.  (It is necessary to `setlocale(LC_ALL, "")` first, which I don't understand.)  e.g.:
```
    wchar_t str[] = { 0x41, 0x211d, 0x1d53d, 0x21, 0 };
    addwstr(str);
 ```
displays "Aℝ𝔽!"

I don't know how this translates to the Terminal.Gui wrapper for curses, which seems to use a `string` (UTF-16) argument for addwstr, which doesn't make sense to me.

(I never use the curses driver, as it just results in black and white squares on my screen, which I haven't had time to debug.)

**Comment by BDisp** [comment]
Yes you are right it's bad. I now see it's a hack for wide char. The problem is how to encode a surrogate pair in one column as you said, because in `WindowsDriver` and `CursesDriver` I can't handle on the output as you did in the `NetDriver`. Thanks.

**Comment by BDisp** [comment]
@cpacejo please confirm that the spacing in each char in the image must exist. If yes, then it must be used the superview color to avoid the black square and also be considered as wider, otherwise the line would be shorter and misaligned.

imagem

**Comment by cpacejo** [comment]
No, there should be no spaces between those characters.  As I mentioned in [NStack bug #86](https://github.com/gui-cs/NStack/issues/86), I believe this is occurring because `Rune.ColumnWidth` is treating characters like U+1D539 as U+D539 for example.

**Comment by cpacejo** [comment]
> The problem is how to encode a surrogate pair in one column as you said, because in WindowsDriver and CursesDriver I can't handle on the output as you did in the NetDriver.

Looking at WindowsDriver, it seems it would have to be rewritten somewhat to follow a similar model as NetDriver.  Or modify `WindowsConsole.CharUnion` to be able to store a surrogate pair.  I can take a stab at this, but I don't often have access to a Windows machine to test on so it may take a few days.

**Comment by BDisp** [comment]
So, the bug is in the `ConsoleDriver` on the `MakePrintable` method that should be like this:

```cs
public static Rune MakePrintable (Rune c)
{
	if (c <= 0x1F || (c >= 0X7F && c <= 0x9F)) {
		// ASCII (C0) control characters.
		// C1 control characters (https://www.aivosto.com/articles/control-characters.html#c1)
		return new Rune (c + 0x2400);
	}

	return c;
}

```

tig:
FWIW (and I'm just 🍿 here), I noticed that even VS2020 has issues with glyphs in this range:

image

Y'all are doing important work on this issue. Carry on!

BDisp:
> FWIW (and I'm just 🍿 here), I noticed that even VS2020 has issues with glyphs in this range:
> 
> image
> 
> Y'all are doing important work on this issue. Carry on!

Well I already done many tests and it seems that the U+10000..10FFFF range is considered wider code points. I also got rid the `gethexaformat` method. I only got all aligned if I consider as wider the range above.
But this is going cause the `TextFormatter.HotKeyTagMask` which is equal to  `U+100000`, to transform the hotkeys, fall to this range, falsifying the column width. I have to consider reviewing this.

BDisp:
I already got `WindowsDriver` work with surrogate pairs but it only renders well on the `Windows Host Console`

windowsdriver-console-host

On `Windows Terminal` it isn't render well. I hope someone of the `Windows Terminal` team can really test if surrogate pairs is working well indeed.

windowsdriver-windows-terminal

Now I'm going trying fix the `CursesDriver`.

BDisp:
> Now I'm going trying fix the CursesDriver.

`CursesDriver` still don't show the code points in the `U+10000...U+10FFFF` range. On a `C` program I can print on a console by using ncursesw flag to the compiler but not yet with the `Terminal.Gui`.

With `WindowsDriver` on the `Windows Terminal` the misalignment is always present event I consider the width equal to one, by commenting the range. You can test by using the pr https://github.com/gui-cs/NStack/pull/89.

imagem

imagem

BDisp:
With `CursesDriver` only ℝ (U+211D), which isn't a surrogate pair, appear:

imagem

tig (approved):
LGTM. But note debugging terminal apps in my VS2022 install is currently broken so I didn't test this myself. (I'm waiting on the Terminal team for a fix they want me to try).

Before I merge this can y'all confirm you really believe it's ready to merge? (@BDisp @cpacejo)

BDisp:
> LGTM. But note debugging terminal apps in my VS2022 install is currently broken so I didn't test this myself. (I'm waiting on the Terminal team for a fix they want me to try).
> 
> Before I merge this can y'all confirm you really believe it's ready to merge? (@BDisp @cpacejo)

Only fixes the `NetDriver`, but no harm will be done if you merge it.

tig:
So it does NOT completely close #2182?

BDisp:
> So it does NOT completely close #2182?

Yes it closes because its only mention the System.Console

PR #2198, opened 2022-11-07
Revamped Character Map scenario to be amaze

tig:
I must have something very important to be doing, because I'm clearly writing code as a way of procrastinating.

This PR massively updates Character Map:

- Now supports keyboard and mouse navigation/selection of glyphs.
  - Single click selects
  - Double click shows message box with info
  - Right click provides Copy Glyph and Copy value context menu
  - Home, End, PageUp, PageDown, Up, Down, Right, Left all work
- New "jump to glyph" textfield takes decimal or hex values to jump to a particular glyph
- Copied all ranges from `TableEditor.cs` and changed radio group to a listbox

I've wanted a way of getting a glyph to the clipboard forever; this makes it easy.

⚽⛔➗⏰

### DIFF
--- a/UICatalog/Scenarios/CharacterMap.cs
+++ b/UICatalog/Scenarios/CharacterMap.cs
@@ -1,12 +1,14 @@
 ﻿#define DRAW_CONTENT
 //#define BASE_DRAW_CONTENT
 
+using Microsoft.VisualBasic;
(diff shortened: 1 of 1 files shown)


tznind:
Very nice, Please keep procrastinating!

PR #2203, opened 2022-11-07
Fixes #2195. CsvEditor: Column Type dialog not wide enough.

### LINKED ISSUE #2195: CsvEditor: Column Type dialog not wide enough
image

This might be a bug in `Dialog`?

BDisp:
Fixes #2195 - Changed `MessageBox` in the `CsvEditor` to use layout computed. Also fixes the `MessageBox` width calculation.

## Pull Request checklist:


tig:
@BDisp - while testing this fix, I found a latent bug in Dialog. I think this will fix it:

```cs
		void LayoutStartedHandler ()
		{
			if (buttons.Count == 0 || !IsInitialized) return;

			int shiftLeft = 0;

			int buttonsWidth = GetButtonsWidth ();
			switch (ButtonAlignment) {
			case ButtonAlignments.Center:
				// Center Buttons
				shiftLeft = (Bounds.Width - buttonsWidth - buttons.Count - 2) / 2 + 1;
				for (int i = buttons.Count - 1; i >= 0; i--) {
					Button button = buttons [i];
					shiftLeft += button.Frame.Width + (i == buttons.Count - 1 ? 0 : 1);
					if (shiftLeft < 0) {
						shiftLeft = 0;
					}
					button.X = Pos.AnchorEnd (shiftLeft);
					button.Y = Pos.AnchorEnd (1);
				}
				break;
```

The bug shows if you add significantly more buttons than will fit. E.g. 20:

Can you incorporate this fix into this PR?  The same `shiftLeft < 0` fix should probably be applied to the other alignments.

BDisp:
> The bug shows if you add significantly more buttons than will fit. E.g. 20:

@tig this isn't a bug. Is the normal center alignment. Now the width will be less or equal to the width console. If there are significantly more buttons than will fit, then it will be centered on the available space. Otherwise your fix will be considered as left alignment.

tznind:
> If there are significantly more buttons than will fit, then it will be centered on the available space. Otherwise your fix will be considered as left alignment.

Could an enhancement be using 2+ lines for the `Button` if they do not all fit on one? As part of layout (incase console is resized)

BDisp:
> > If there are significantly more buttons than will fit, then it will be centered on the available space. Otherwise your fix will be considered as left alignment.
> 
> Could an enhancement be using 2+ lines for the `Button` if they do not all fit on one? As part of layout (incase console is resized)

It's a good suggestion, but I think is better to open a new issue.

What can be still done in this PR is improving some more unit tests.

BDisp:
> @BDisp - while testing this fix, I found a latent bug in Dialog. I think this will fix it:
> 
> ```cs
> 		void LayoutStartedHandler ()
> 		{
> 			if (buttons.Count == 0 || !IsInitialized) return;
> 
> 			int shiftLeft = 0;
> 
> 			int buttonsWidth = GetButtonsWidth ();
> 			switch (ButtonAlignment) {
> 			case ButtonAlignments.Center:
> 				// Center Buttons
> 				shiftLeft = (Bounds.Width - buttonsWidth - buttons.Count - 2) / 2 + 1;
> 				for (int i = buttons.Count - 1; i >= 0; i--) {
> 					Button button = buttons [i];
> 					shiftLeft += button.Frame.Width + (i == buttons.Count - 1 ? 0 : 1);
> 					if (shiftLeft < 0) {
> 						shiftLeft = 0;
> 					}
> 					button.X = Pos.AnchorEnd (shiftLeft);
> 					button.Y = Pos.AnchorEnd (1);
> 				}
> 				break;
> ```
> 
> The bug shows if you add significantly more buttons than will fit. E.g. 20:
> 
> Can you incorporate this fix into this PR? The same `shiftLeft < 0` fix should probably be applied to the other alignments.

You are right @tig, sorry, I've confused about the `button.X` which can be negative and I didn't realized the the `shiftLeft` is used in the `Pos.AnchorEnd` margin.
If you don't mind I'll will fix the #2204 into this PR.

BDisp:
The errors related with this PR are already fixed, but still the tests fails due the same reason as #2212.

BDisp:
file-dialog

PR #2215, opened 2022-11-10
Fixes #2212. Editor Scenario: Crashes on Exit.

### LINKED ISSUE #2212: Editor Scenario: Crashes on Exit
lNOeRpZ 1

My VS2022 install has become hooked and I can't debug right now. Sigh.

BDisp:
Fixes #2212 - The `historyText.Clear` must be called to reset `IsDirty` to false. The `LoadStream` method now also handle null terminated for the cases where they must be part of the text or not.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -92,7 +92,7 @@ namespace Terminal.Gui {
 			var line = new List<byte> ();
 			var wasNewLine = false;
 			while ((v = buff.ReadByte ()) != -1) {
-				if ((!AllowsNullTerminated && v == '\0') || v == 13) {
+				if (v == 13) {
 					continue;
 				}
 				if (v == 10) {
@@ -141,11 +141,6 @@ namespace Terminal.Gui {
 		/// The number of text lines in the model
 		/// </summary>
 		public int Count => lines.Count;
-
-		/// <summary>
-		/// Gets or sets if null terminated in a buffer array is considered as part of the <see cref="lines"/>.
-		/// </summary>
-		public bool AllowsNullTerminated { get; set; }
 
 		/// <summary>
 		/// Returns the specified line as a List of Rune
@@ -1798,14 +1793,6 @@ namespace Terminal.Gui {
 		/// Get the <see cref="ContextMenu"/> for this view.
 		/// </summary>
 		public ContextMenu ContextMenu { get; private set; }
-
-		/// <summary>
-		/// Gets or sets if null terminated in a buffer array is considered as part of the <see cref="Text"/>.
-		/// </summary>
-		public bool AllowsNullTerminated {
-			get => model.AllowsNullTerminated;
-			set => model.AllowsNullTerminated = value;
-		}
 
 		int GetSelectedLength ()
 		{
--- a/Terminal.Gui/Views/TextView.cs
+++ b/Terminal.Gui/Views/TextView.cs
@@ -856,7 +856,7 @@ namespace Terminal.Gui {
 			var firstLine = wrappedModelLines.IndexOf (r => r.ModelLine == modelLine);
 			int modelCol = 0;
 
-			for (int i = firstLine; i <= line; i++) {
+			for (int i = firstLine; i <= Math.Min (line, wrappedModelLines.Count - 1); i++) {
 				var wLine = wrappedModelLines [i];
 
 				if (i < line) {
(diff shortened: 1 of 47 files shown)


Terminal.Gui/Views/TextView.cs
```diff
@@ -1794,6 +1799,14 @@ void WrapTextModel ()
 		/// </summary>
 		public ContextMenu ContextMenu { get; private set; }
 
+		/// <summary>
+		/// Gets or sets if null terminated in a buffer array is considered as part of the <see cref="Text"/>.
+		/// </summary>
+		public bool AllowsNullTerminated {
```
tig, line 1805: What is the user-scenario where a developer would set this property to new?

It seems like very obscure edge-case?

tig (changes requested):

Terminal.Gui/Views/TextView.cs
BDisp: See the unit test added. If it's encoded as `Unicode` it add a null terminated in the array byte.

UnitTests/TextViewTests.cs
```diff
@@ -1930,6 +1929,42 @@ public void LoadStream_LF ()
 			Assert.Equal ($"This is the first line.{Environment.NewLine}This is the second line.{Environment.NewLine}", tv.Text);
 		}
 
+		[Fact]
+		public void LoadStream_IsDirty_AllowsNullTerminator ()
+		{
+			var text = "Testing";
```
tig, line 1935: I don't see any tests that show what happens if a `null` is in the middle of the text. Should there be a tests for that?

BDisp: Yes you are right. I'll do it.

Terminal.Gui/Views/TextView.cs
```diff
@@ -142,6 +142,11 @@ public override string ToString ()
 		/// </summary>
 		public int Count => lines.Count;
 
+		/// <summary>
+		/// Gets or sets if null terminated in a buffer array is considered as part of the <see cref="lines"/>.
+		/// </summary>
+		public bool AllowsNullTerminated { get; set; }
```
tig, line 148: It doesn't matter if the `\0` is at the end of the line or not, right? So this has nothing to do with 'NullTerminated', but just `Null`.

Also, I believe the logic you implemented means that if this member is `true` any `\0` characters in the input stream will be preserved (and thus displayed in the editor). If `false` they will be ignored.

Thus, I believe this member should be named `AllowsNulls` instead. 

If I'm correct please rename it to `IgnoreNulls` and update the help text to something like 

> 'Gets or sets if '\0' characters will be preserved in the input stream by the <see cref="LoadStream(Stream stream)"/> method. If set to <see langref="false"/> (the default) any `\0` characters in the input stream will be preserved (and rendered in the TextView). If set to <see langref="true"/> `\0` characters will be stripped from the input stream."

And what will the editor show when `\0` characters are preserved in the input stream? Will it show a U+2400 (␀)?

I may be completely misunderstanding this though, so please correct me if I'm wrong!

tig (changes requested):
See my comments about `AllowsNullTerminated`

Terminal.Gui/Views/TextView.cs
BDisp: @tig I really I'm confused about this. This issue come up because I was getting wrong value on the unit test because of this statement:
```cs
var text = "Testing";
byte [] buff = System.Text.Encoding.Unicode.GetBytes (text);
```
`buff` returned an array length of 14, the double of 7 that is the length if the `text` variable. So, the `GetBytes` method was returning a null terminated for each character and not only at the end of the line. Maybe was my mistake of the using of the `GetBytes` method. If you think it's better not implementing this property at all I have no problem by remove it.
If you want to preserve I'm now with a doubt if I should named `AllowsNulls` or `IgnoreNulls`.

> And what will the editor show when \0 characters are preserved in the input stream? Will it show a U+2400 (␀)?

Yes it will.

tig: I think you should simplify this fix to fix the crash in the most simple way possible.

BDisp: You are meaning I should discard this `AllowsNullTerminated` or whatever property, right?

BDisp: Done @tig. I removed the `AllowsNullTerminated` property then.

tig (approved):

PR #2250, opened 2022-12-06
Fixes #2249. CharacterMap isn't refreshing well non-BMP code points on scroll.

### LINKED ISSUE #2249: CharacterMap isn't refreshing well non-BMP code points on scroll.
It's showing this:

imagem

But must showing this:

imagem

BDisp:
Fixes #2249 - Also improves the searching glyph by selecting and make visible the related value found in the `ListView`.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/NetDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/NetDriver.cs
@@ -1242,12 +1242,12 @@ namespace Terminal.Gui {
 
 			if (validClip) {
 				if (runeWidth < 2 && ccol > 0
-					&& Rune.ColumnWidth ((char)contents [crow, ccol - 1, 0]) > 1) {
+					&& Rune.ColumnWidth ((Rune)contents [crow, ccol - 1, 0]) > 1) {
 
 					contents [crow, ccol - 1, 0] = (int)(uint)' ';
 
 				} else if (runeWidth < 2 && ccol <= Clip.Right - 1
-					&& Rune.ColumnWidth ((char)contents [crow, ccol, 0]) > 1) {
+					&& Rune.ColumnWidth ((Rune)contents [crow, ccol, 0]) > 1) {
 
 					contents [crow, ccol + 1, 0] = (int)(uint)' ';
 					contents [crow, ccol + 1, 2] = 1;
@@ -1483,10 +1483,9 @@ namespace Terminal.Gui {
 							output.Append (WriteAttributes (attr));
 						}
 						outputWidth++;
-						var rune = contents [row, col, 0];
-						char [] spair;
-						if (Rune.DecodeSurrogatePair((uint) rune, out spair)) {
-							output.Append (spair);
+						Rune rune = contents [row, col, 0];
+						if (Rune.DecodeSurrogatePair (rune, out char [] spair)) {
+							output.Append (spair).Append ('\ufeff');
 						} else {
 							output.Append ((char)rune);
 						}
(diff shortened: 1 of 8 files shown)


BDisp:
Also fixes this;

imagem

imagem

BDisp:
With `WindowsDriver` and with windows console host I can achieve this:

imagem

But with `Windows Terminal` it's all mess up:

imagem

tig (approved):
Nice work.

tig:
> With `WindowsDriver` and with windows console host I can achieve this:
> 
> imagem
> 
> But with `Windows Terminal` it's all mess up:
> 
> imagem

Is there an Issue filed against Windows Terminal on this?

BDisp:
> Is there an Issue filed against Windows Terminal on this?

I don't think there is. I think you may ask them. I don't know if I'm doing something wrong.

tig:
> > Is there an Issue filed against Windows Terminal on this?
> 
> I don't think there is. I think you may ask them. I don't know if I'm doing something wrong.

They are quite responsive. I suggest you file an Issue

https://github.com/microsoft/terminal/issues

DHowett:
FWIW, `U+1D53D MATHEMATICAL DOUBLE-STRUCK CAPITAL F` and friends are all indicated in the Unicode character database as occupying one column (`ea=N Emoji=N EPres=N`).

If I had to guess, it's actually a bug that they're being displayed in _two_ columns in the old Windows console host! 🙂 

(Terminal has broadly better Unicode compatibility than the Console does, because it has a faster release cycle and we've been able to fix it much more quickly.)

BDisp:
Thanks for your answers @DHowett. The problem is if I consider it as one column it also appear misaligned. Where can I get the information about how many columns a code point occupies?

tig:
@BDisp -should this be marked as a draft PR?

tig:
@BDisp I pulled this down locally. I noted this. I'm not sure if this is the same issue you are attempting to fix with this PR, but I thought I'd let you know:

SkqFtdL 1

BDisp:
> @BDisp I pulled this down locally. I noted this. I'm not sure if this is the same issue you are attempting to fix with this PR, but I thought I'd let you know:

Yes it's. Non-BMP cannot be used with the current `WindowsDriver` because it only allow handle it by using two blocks or columns and not all them occupies two columns. I'm working on changing the write API to another type which can handle 32 bits characters.

tig:
> > @BDisp I pulled this down locally. I noted this. I'm not sure if this is the same issue you are attempting to fix with this PR, but I thought I'd let you know:
> 
> Yes it's. Non-BMP cannot be used with the current `WindowsDriver` because it only allow handle it by using two blocks or columns and not all them occupies two columns. I'm working on changing the write API to another type which can handle 32 bits characters.

Got it.

Do you have an ETA for the fix? I'm asking because we need to get a release out. I'm thinking that if you need more time, then I'll do a PR for CharMap that disables those ranges for now. What do you think?

BDisp:
> Do you have an ETA for the fix? I'm asking because we need to get a release out. I'm thinking that if you need more time, then I'll do a PR for CharMap that disables those ranges for now. What do you think?

No, sorry I don't have an ETA for the fix, so is better you do a PR for disable those ranges for now. Thanks.

BDisp:
@tig the fixes included in this PR for `Toplevel` and `ListView` are in #2267 and #2270, so we can leverage them for the new package while the non-BMP fixes aren't ready for the others drivers.

PR #2258, opened 2022-12-23
Fixes #505. Finishes merge of TileView (previously SplitContainer)

### LINKED ISSUE #505: New Control: Split Panel
This may already be possible, but I'm missing it:

I want a control that lets me build paneled UIs like `UI Catalog` more easily:

image

But with arbitrary horizonal and vertical panels. It's a very common console GUI idiom.

I should be able to resize the panels with either the mouse or keyboard (with well defined shortcut key idioms that would be consistent across all `Terminal.Gui` apps). 

TAB should navigate gracefully between panels.

I should be able to add and remove panels dynamically at runtime. 

Panels should have (optional) Titles (like shown above). 

Unlike above, the border between panels should be a single column (or row for horizontal panels).

tznind:
Fixes #505 - Adds new View TileView

TileView can have any number of Tiles.  Each tile is seperated with a line which can be moved.  Tiles can be added and removed adhoc (see `InsertTile`, `RemoveTile`).  

All Tiles must follow the same `Orientation`.  You can achieve grids/ T shapes etc of tiles through nesting.  To make this easier you can use the method `bool TrySplitTile(int idx, int panels, out TileView result)`.

This PR contains the following sub PR which could be merged seperately

- #2299 
- #2281 

### Noteworthy changes

- Makes `ViewToScreen` public.  This is to position ContextMenu when right clicking TabView headers correctly.  I have used this method via reflection in TerminalGuiDesigner and don't really see why `ScreenToView` is public but `ViewToScreen` is private.
- Fix bug in TabView where the 'damage region' property of `bounds` passed to `Redraw` is used for drawing border instead of `Bounds` which is always correct.  This manifested as a missing border when moving splitter lines (which reapeared after moving cursor or clicking... most strange).

### ToDos

- [x] Titles with the new Tile system
- [x] Verify min tile sizes with 3+ tiles
- [x] Respect `Visible` property to hide tiles
   - [x] Test all combinations for 3x2 tile configuration with border
   - [x] Test all combinations for 3x2 tile configuration without border
- [x] Needs Tests
- [x] Needs More Tests
- [x] When switching from horizontal to vertical the SplitterDistance is usually way off (see in video) because consoles are far wider than they are tall.
- [x] Actually respect the Min Panel Size properties
- [x] Max/Min panel sizes should preserve Percent/Absolute state of SplitterDistance (currently resizing below the min results in a change to absolute splitter positioning
- [x] Refactor UICatalogApp to use this as part of the PR

splitter

Modeled after the WinForms / Windows SplitContainer I've used the same name and public property names (Panel1, Panel2, SplitterDistance etc).

Draft for now as there's a few kinks to work out.  Theres a bit of a challenge in supporting both Percent and Absolute splitter distances.  Its mostly working for both but needs some thourough testing and to consider the other `Pos` types users could assign to `SplitterDistance`.

## Out of Scope

- [x] SplitterDistance should handle Percent and Absolute but not other types of `Pos`.  Letting user resize with keyboard/mouse means preserving the PosType is difficult.  Its most feasible if we restrict to Percent and Absolute.  So we should throw `ArgumentException` if user sets `SplitterDistance` to non percent/absolute (at least for time being).

## Pull Request checklist:


tig:
Awesome.

Can you refactor UICatalogApp to use this as part of the PR? That would add another use/test case.

tznind:
@tig I took a look at changing UICatalog to use a split container.  The simplest solution would be just to put the 2 `FrameView` into either side of the parent SplitContainer but this would leave 3 separating lines down the middle (1 for each FrameView and 1 for the SplitterLine.

I looked into how to do it with just 1 line down the middle and have made some progress.

The biggest issue with this approach is that the 'Scenarios' caption has to go somewhere.  It also ideally has to move with the Splitter when dragged (see video).  

I have a quite hacky solution using a `Label` floating over the Title but its pretty janky (see code below).  The Label also vanishes periodically so need to come up with something better.  Might be possible to just change the `Title` on the `FrameView` to have both strings (Categories and Scenarios) seperated by HLine.

But I thought I'd show the route I am going down before I put too much work into changing `UICatalog.cs` .  Was this what you were thinking of?  

I've not committed the change but put the patch diff below so you can give me some feedback on whether you think this is an improvement?

UICatalogWithSplitter


```diff
--- a/UICatalog/UICatalog.cs
+++ b/UICatalog/UICatalog.cs
@@ -1,4 +1,4 @@
-´╗┐using NStack;
+using NStack;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -151,9 +151,9 @@ namespace UICatalog {
                        public MenuItem miIsMouseDisabled;
                        public MenuItem miHeightAsBuffer;

-                       public FrameView LeftPane;
+                       public FrameView ContentPane;
+                       public SplitContainer SplitContainer;
                        public ListView CategoryListView;
-                       public FrameView RightPane;
                        public ListView ScenarioListView;

                        public StatusItem Capslock;
@@ -201,24 +201,29 @@ namespace UICatalog {
                                        }),
                                        new StatusItem(Key.F10, "~F10~ Hide/Show Status Bar", () => {
                                                StatusBar.Visible = !StatusBar.Visible;
-                                               LeftPane.Height = Dim.Fill(StatusBar.Visible ? 1 : 0);
-                                               RightPane.Height = Dim.Fill(StatusBar.Visible ? 1 : 0);
+                                               ContentPane.Height = Dim.Fill(StatusBar.Visible ? 1 : 0);
                                                LayoutSubviews();
                                                SetChildNeedsDisplay();
                                        }),
                                        DriverName,
                                };

-                               LeftPane = new FrameView ("Categories") {
+                               ContentPane = new FrameView ("Categories") {
                                        X = 0,
                                        Y = 1, // for menu
-                                       Width = 25,
+                                       Width = Dim.Fill (),
                                        Height = Dim.Fill (1),
                                        CanFocus = true,
                                        Shortcut = Key.CtrlMask | Key.C
                                };
-                               LeftPane.Title = $"{LeftPane.Title} ({LeftPane.ShortcutTag})";
-                               LeftPane.ShortcutAction = () => LeftPane.SetFocus ();
+                               ContentPane.Title = $"{ContentPane.Title} ({ContentPane.ShortcutTag})";
+                               ContentPane.ShortcutAction = () => ContentPane.SetFocus ();
+
+                               SplitContainer = new SplitContainer {
+                                       Width = Dim.Fill (0),
+                                       Height = Dim.Fill (0),
+                                       SplitterDistance = 25
+                               };

                                CategoryListView = new ListView (_categories) {
                                        X = 0,
@@ -229,21 +234,23 @@ namespace UICatalog {
                                        CanFocus = true,
                                };
                                CategoryListView.OpenSelectedItem += (a) => {
-                                       RightPane.SetFocus ();
+                                       ScenarioListView.SetFocus ();
                                };
                                CategoryListView.SelectedItemChanged += CategoryListView_SelectedChanged;
-                               LeftPane.Add (CategoryListView);
+                               ContentPane.Add (SplitContainer);

-                               RightPane = new FrameView ("Scenarios") {
-                                       X = 25,
+                               SplitContainer.Panel1.Add (CategoryListView);
+
+                               Label lblScenarios = new Label ("Scenarios") {
+                                       X = SplitContainer.SplitterDistance + 1,
                                        Y = 1, // for menu
-                                       Width = Dim.Fill (),
-                                       Height = Dim.Fill (1),
-                                       CanFocus = true,
-                                       Shortcut = Key.CtrlMask | Key.S
                                };
-                               RightPane.Title = $"{RightPane.Title} ({RightPane.ShortcutTag})";
-                               RightPane.ShortcutAction = () => RightPane.SetFocus ();
+                               SplitContainer.SplitterMoved += (s, e) => {
+                                       lblScenarios.X = e.SplitterDistance + 1;
+                                       Application.Top.BringSubviewToFront (lblScenarios);
+                                       lblScenarios.SetNeedsDisplay ();
+                                       Application.Top.SetNeedsDisplay ();
+                               };

                                ScenarioListView = new ListView () {
                                        X = 0,
@@ -255,13 +262,13 @@ namespace UICatalog {
                                };

                                ScenarioListView.OpenSelectedItem += ScenarioListView_OpenSelectedItem;
-                               RightPane.Add (ScenarioListView);
+                               SplitContainer.Panel2.Add (ScenarioListView);

                                KeyDown += KeyDownHandler;
                                Add (MenuBar);
-                               Add (LeftPane);
-                               Add (RightPane);
+                               Add (ContentPane);
                                Add (StatusBar);
+                               Add (lblScenarios);

                                Loaded += LoadedHandler;

@@ -287,7 +294,7 @@ namespace UICatalog {
                                        _isFirstRunning = false;
                                }
                                if (!_isFirstRunning) {
-                                       RightPane.SetFocus ();
+                                       ScenarioListView.SetFocus ();
                                }
                                Loaded -= LoadedHandler;
                        }
```

tznind:
Ok UICatalog is now using the new Splitter pane and it looks good.  Its one commit so if we don't think it is an improvement I can just revert [dd246b7](https://github.com/gui-cs/Terminal.Gui/pull/2258/commits/dd246b74ae63db5fe420cb5c9a348bb8d34049a2)

tig:
Why can't we have a "T" here?

image

UICatalog/UICatalog.cs
```diff
@@ -270,6 +272,21 @@ public UICatalogTopLevel ()
...
+			{ 
+				// Pos.Anchor is internal so we have to use reflection to access it
+				var anchor = typeof (Pos).GetMethod ("Anchor", BindingFlags.Instance | BindingFlags.NonPublic);
+				var splitterWidth = (int)anchor.Invoke (
+					SplitContainer.SplitterDistance, new object [] { SplitContainer.Bounds.Width}
+					);
+
+				var newTitle = $"Categories ({ContentPane.ShortcutTag})";
+				newTitle = newTitle.PadRight (splitterWidth + 1, (char)Driver.HLine);
+				newTitle += "Scenarios";
```
tznind, line 285: > Why can't we have a "T" here?
> 
> image

Should be able to do that with a bit of work.

Currently a `SplitContainer` control is 2 borderless `View` panels and a horizontal line down the middle.

The `Title` (and border) come from a `FrameView`.  I align the Title using this approach here which feels a bit ad-hoc.  I could update this method to have the `Title` use a T at the right horizontal position so that it aligns.  But there would be now corresponding upside down T at the bottom since those Runes come from the `Title`.

Alternatively I could look into providing a version of `SplitContainer` that draws its own (optionally captioned  border).

tig (changes requested):
I have a PR to this PR incoming.  You'll like it!

Terminal.Gui/Views/SplitContainer.cs
```diff
@@ -0,0 +1,550 @@
...
+
+			CanFocus = false;
+		}
+
+		/// <summary>
+		/// The left or top panel of the <see cref="SplitContainer"/>
+		/// (depending on <see cref="Orientation"/>).  Add panel contents
+		/// to this <see cref="View"/> using <see cref="View.Add(View)"/>.
+		/// </summary>
+		public View Panel1 { get; } = new View ();
```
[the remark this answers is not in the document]
tznind: > What if I wanted a SplitContainer with 3 panels?

Easiest way would be to add a split container inside another split container.  Doing that would also let you do more complex stuff like having a square where either side can have separate vertical offsets e.g.

```
|----|-------|
|p1  |p2     |
|    |-------|
|____|p4     |
|p3  |_______|
`````
_Example of nesting a SplitContainer in each Panel1 and Panel2 of a parent SplitContainer_

File: UICatalog/UICatalog.cs [hunk 3]
```diff
@@ -270,6 +272,21 @@ public UICatalogTopLevel ()
 				ScenarioListView.SelectedItem = _cachedScenarioIndex;
 			}
 
+			private void SetTitlePaddingToAlignScenariosHeader ()
+			{ 
+				// Pos.Anchor is internal so we have to use reflection to access it
```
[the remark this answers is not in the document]
tznind: > What if SplitContainer did full borders (like FrameView) for each of it's panes.

I like this idea but I think it must be optional.  

There are cases where you don't want to loose the vertical space taken up by the border (e.g. in FileDialog2) or where it would be unwelcome to always have borders (e.g. when nesting `SplitContainer`s).

tznind:
Thanks for your code @tig its looking way better now! I've fixed the compilation errors in the unit tests and added support for `SplitContainer` detecting when theres no border set (which results in it not using T's).  But there does seem to be some bad layout logic when the split container gets very small (most evident in tests since they use a 11x3 size).  But you can repro it in the UICatalog Scenario too:

Wierder still the splitter line seems to jump around when the height is small and you mouse on and off it (see end of this video):

small-splits

BDisp:
Here my opinion. I think `SplitContainer` must obtain the border configuration of his host. If has no border then a single line must be drawn without T. If it's a single border then it must drawn a single T (left-right or top-bottom). If it's a double border it must drawn double T. In the bellow image the `A` and `B` was created with a single vertical `SplitContainer`. In the `B` panel was created another horizontal `SplitContainer` generating the `C` panel. I think this strategic is more controllable to deal with odd panels.

imagem

tznind:
> I think `SplitContainer` must obtain the border configuration of his host. If has no border then a single line must be drawn without T.

I think we can get the best of both worlds by tapping into the `Border` property.

@tig change means `SplitContainer` inherits from `FrameView` which looks great and works well with Titles.  But if user sets the `BorderStyle` to `None` then we should adapt to that.  I've started that in 
e4cb4b9a676374a7b8b9dc7924fdadde7eb3f16c but few kinks to work out.

tznind:
I've still got 3 failing tests after your changes @tig.  They are all the same and manifest when having horizontal splitter and height 3.  The top panel does not render (or is cleared, or something renders over the top of it.  I've not been able to track down what the issue is.  The positions are all correct according to the Watch window:

image


Any help would be appreciated.

```
    Expected:
    11111111111
    ───────────
    22222222222
    But Was:
    ───────────
    22222222222
```

tznind:
Have added a 3 way split to the SplitContainerExample which makes this more exciting demo.  But it has also revealed an issue with Tab navigation not working.  

Ill dig into it when I have some time but marked as Draft for now.

I can tab fine on the UICatalog home screen but in the example I can only tab into the first split container and not the nested one.

BDisp:
> Have added a 3 way split to the SplitContainerExample which makes this more exciting demo. But it has also revealed an issue with Tab navigation not working.
> 
> Ill dig into it when I have some time but marked as Draft for now.
> 
> I can tab fine on the UICatalog home screen but in the example I can only tab into the first split container and not the nested one.

Thanks for your effort. That great. I think the `SplitContainer` must deal with tabbing. So, if the the last view on the is selected, tabbing will move to the first view of the nested slit container. If the first view of he nested split container is selected, shift tabbing will move to the last view of the first split container.

tznind:
Ok turns out I just needed `CanFocus = true;` on SplitContainer.  Seems to work correctly now

tznind:
Ok turns out I just needed `CanFocus = true;` on SplitContainer.  Seems to work correctly now

UICatalog/Scenarios/SplitContainerExample.cs
```diff
@@ -0,0 +1,149 @@
...
+
+namespace UICatalog.Scenarios {
+	[ScenarioMetadata (Name: "Split Container", Description: "Demonstrates the SplitContainer functionality")]
+	[ScenarioCategory ("Controls")]
+	[ScenarioCategory ("LineView")]
+	public class SplitContainerExample : Scenario {
+
+		private SplitContainer splitContainer;
+
+		private SplitContainer splitContainer2;
```
tig, line 15: Suggest you name this `nestedSplitContainer` or something?

UICatalog/Scenarios/SplitContainerExample.cs
```diff
@@ -0,0 +1,149 @@
...
+			 {
+				Width = Dim.Fill(),
+				Height = Dim.Fill(),
+				Text = GenerateLotsOfText(),
+				AllowsTab = false,
+				WordWrap = true,
+			 });
+
+			splitContainer2.Border.BorderStyle = BorderStyle.None;
+			splitContainer2.Border.DrawMarginFrame = false;
```
tig, line 68: Why is this needed?

UICatalog/Scenarios/SplitContainerExample.cs
```diff
@@ -0,0 +1,149 @@
...
+			Win.Y = 1;
+
+			Win.Add (new Label ("This is a SplitContainer with a minimum panel size of 4. Drag the splitter to resize:"));
+
+			splitContainer = new SplitContainer {
+				Y = 2,
+				X = 2,
+				Width = Dim.Fill () - 2,
+				Height = Dim.Fill () - 1,
+				SplitterDistance = Pos.Percent (50), // TODO: get this to work with drag resizing and percents
```
tig, line 38: Is this comment still relevant?

Terminal.Gui/Views/SplitContainer.cs
```diff
@@ -0,0 +1,554 @@
...
+				});
+
+				AddKeyBinding (Key.CursorRight, Command.Right);
+				AddKeyBinding (Key.CursorLeft, Command.Left);
+				AddKeyBinding (Key.CursorUp, Command.LineUp);
+				AddKeyBinding (Key.CursorDown, Command.LineDown);
+
+				LayoutStarted += (e) => {
+					moveRuneRenderLocation = null;
+					if (Orientation == Orientation.Horizontal) {
```
tig, line 326: This should all be refactored. 

We should add to `ConsoleDriver` a new method, `DrawWindowLine`. `ConsoleDriver.DrawWindowFrame` should be refactored to use `DrawWindowLine`. `DrawWindowLine` should have a parameter `StartingAnchor` and `EndingAnchor`. 

`LineView` should also be refactored to use `DrawWindowLine` (basically we should move the `Redraw` code from `LineView` to `ConsoleDriver.DrawWindowLine`.

UICatalog/Scenarios/SplitContainerExample.cs
```diff
@@ -0,0 +1,149 @@
...
+			splitContainer2.Panels [1].Add (new Label ("Here is a Text box:") { Y = 1 });
+			splitContainer2.Panels [1].Add (new TextView () { Y = 2, Width = Dim.Fill (), Height = Dim.Fill (), AllowsTab = false });
+
+			Win.Add (splitContainer);
+
+			var menu = new MenuBar (new MenuBarItem [] {
+			new MenuBarItem ("_File", new MenuItem [] {
+				new MenuItem ("_Quit", "", () => Quit()),
+			}),
+			new MenuBarItem ("_Options", new MenuItem [] {
```
tig, line 81: We should add another submenu that changes the `Border` the split panels so we can prove all the border styles work.

UICatalog/Scenarios/SplitContainerExample.cs
```diff
@@ -0,0 +1,149 @@
...
+			Label lbl1;
+			splitContainer.Panels [0].Title = "Hello";
+			splitContainer.Panels [0].Add (lbl1 = new Label ("Type Something:") { Y = 0 });
+			splitContainer.Panels [0].Add (new TextField () { Width = Dim.Fill (), Y = 0, X = Pos.Right (lbl1) + 1 });
+
+			Label lbl2;
+			splitContainer.Panels [1].Title = "World";
+			splitContainer.Panels[1].Add(splitContainer2);
+
+			splitContainer2.Panels [0].Add (new TextView ()
```
tig, line 58: The line-ends on the embedded split container are not drawing as Ts correctly. They really should!

tig (changes requested):

Terminal.Gui/Views/SplitContainer.cs
tig: I just had a crazy idea...

What if we just made it so an extended version of `DrawWindowFrame` was able to draw arbitrary nested/tiled panels?

There's got to be a way to describe a frame with infinitely nested sub-panels via some sort of data structure (I too dum to know, but I think I remember MS doing something like this back in Windows 1.x). Then all the logic trying to deal with z-order issues in `Window` and `FrameView` (and now `SplitContainer`) could be moved down to `DrawWindowFrame`?

UICatalog/Scenarios/SplitContainerExample.cs
tig: Nevermind. I see now. 

But i also see a problem: If you want splitContainer2.Panel[1] to have a title this doesn't work.

tig: image

tig: With it true (and `[1].Title` set to "Two"):

image

Terminal.Gui/Views/SplitContainer.cs
```diff
@@ -0,0 +1,554 @@
...
+		/// <summary>
+		/// A panel within a <see cref="SplitterPanel"/>. 
+		/// </summary>
+		public class SplitterPanel : View {
+			Pos minSize = 1;
+
+			/// <summary>
+			/// Gets or sets the minimum size for the panel.
+			/// </summary>
+			public Pos MinSize { get => minSize;
```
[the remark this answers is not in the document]
tznind: I have fixed the implementation of `BoundByMinimumSizes` in [63c3742](https://github.com/gui-cs/Terminal.Gui/pull/2258/commits/63c37420f478d816027909c41f507ca5715f98d8)

The minimum size calculation for panel 2 was off by 2 units when the SplitContainer had a Border.

I've also added test cases for with/without border.

UICatalog/Scenarios/SplitContainerExample.cs
tznind: So I've noticed that `BorderStyle` and `DrawMarginFrame` are tightly coupled.  If you change one but not the other then things stop laying out correctly (at least for this case).

For the tests and example I have stuck to either

**With Border**
```csharp
mySplitContainer.Border.BorderStyle = BorderStyle.Single; // the default
mySplitContainer.Border.DrawMarginFrame = true; // the default
```

**Without Border**
```csharp
mySplitContainer.Border.BorderStyle = BorderStyle.None;
mySplitContainer.Border.DrawMarginFrame = false;
```

If you do that then it does work with subpane titles.  But you get that nasty 'double nested' appearance where there are clearly 2 borders:

image

UICatalog/Scenarios/SplitContainerExample.cs
tznind: Comment rot! Fixed in [675a6b0](https://github.com/gui-cs/Terminal.Gui/pull/2258/commits/675a6b02898b62a4cfbea637f9f6070ab750d3f3)

UICatalog/Scenarios/SplitContainerExample.cs
tznind: I've added the menu items to the Scenario in [61a4544](https://github.com/gui-cs/Terminal.Gui/pull/2258/commits/61a45443c8972a48fa33a673a932d8552d32631b) but it does highlight how `Title` doesn't work properly without `Border` (or at least margin frame).

tznind: Ill resolve this specific conversation since we can now toggle them and we can continue the discussion about layout in the main conversation.

UICatalog/Scenarios/SplitContainerExample.cs
BDisp: > So I've noticed that `BorderStyle` and `DrawMarginFrame` are tightly coupled. If you change one but not the other then things stop laying out correctly (at least for this case).

Yes is true. There is needed to set both, otherwise an empty frame without line is drawn.

UICatalog/Scenarios/SplitContainerExample.cs
tznind: With the current way this is architected it would be difficult for the nested container to draw these line endings.

While it looks like there is a single container with 2 dividing lines there is in fact 2 seperate containers nested inside each other.  The nested container is configured to have no border so it looks like it integrates with the parent.  To draw a 'T' into the parent it would have to draw outside its bounds

image
_Pink rectangle highlights the bounds of the parent container.  Green rectangle highlights the bounds of the nested container_

Things would get more complicated when considering a 2x2 configuration.  We would need a 'crosshair' in the middle but only until one of the two horizontal lines were moved.

image
_This would require a crosshair for the intersecting line_

After the horizontal line (e.g. left) is moved up we would need two Ts again.

image
_After moving the nested container up we would need two T instead of a crosshair_

Terminal.Gui/Views/SplitContainer.cs
tig: Almost like WM_NCPAINT of old (fun fact, I was one of the first Windows developers outside of MS to leverage WM_NCPAINT for fun and profit back in the late '80s). One of my most popular pieces of code that got shared everywhere was named "Itsybits" and enabled windows with tiny title bars.

https://learn.microsoft.com/en-us/windows/win32/gdi/nonclient-area

Just found this:

https://view.officeapps.live.com/op/view.aspx?src=https%3A%2F%2Fraw.githubusercontent.com%2FPaolo-Maffei%2FOpenNT%2F5c5b979ec08c17d3ca2eb70e8aad62d26515d01c%2Fsdktools%2Ftopdesk%2Fapp%2Fitsybits.doc&wdOrigin=BROWSELINK

LMAO!

UICatalog/Scenarios/SplitContainerExample.cs
tig: I went down the "DrawWindowFrame With Panes" hole because of this. 

Another idea is to make `SplitContainer` support arbitrary panes without them being nested.

tznind:
Ok I've done some thinking (dangerous I know) and I think this is doable.

We need 2 systems for this to work:

## Arbitrary Connected Lines Drawer

This would work like a 'back buffer' in in graphics.  It would store all the vector lines that are added to it.  Then user would call 'Finalize' or 'Merge' or something which would create a 2D array in which all the lines are drawn.  Corners would be rendered with the appropriate symbol (T, Crosshair or L) based on the direction of intersecting vectors at that point.

The full final frame would then be used for whatever (e.g. drawn to screen).  This would be pretty easy to test and develop independently and demoable with a kind of 'drawing' Scenario.

## Integrated nesting of SplitContainers

Currently a SplitContainer has 2 panels.  If we change `SplitPanel` to be a binary tree we can do let the user define as many sublevels as they want without having to create additional instances (good!).

In layout logic we can use recursion to pass ever reducing `Bounds` sizes into child branches based on the Splitter position consmed at the current level.

## Estimate
Lots of work! but sounds fun!

tznind:
Ok started exploring this idea here:

https://github.com/tznind/gui.cs/blob/line-drawer/Terminal.Gui/Core/Graphs/StraightLineCanvas.cs

Terminal.Gui/Views/SplitContainer.cs
tznind: Code never dies! Thats so awesome.

BDisp: Also in GitHub https://gist.github.com/tig/3491637.

tig:
> Ok I've done some thinking (dangerous I know) and I think this is doable.
> 
> We need 2 systems for this to work:
> 
> ## Arbitrary Connected Lines Drawer
> This would work like a 'back buffer' in in graphics. It would store all the vector lines that are added to it. Then user would call 'Finalize' or 'Merge' or something which would create a 2D array in which all the lines are drawn. Corners would be rendered with the appropriate symbol (T, Crosshair or L) based on the direction of intersecting vectors at that point.
> 
> The full final frame would then be used for whatever (e.g. drawn to screen). This would be pretty easy to test and develop independently and demoable with a kind of 'drawing' Scenario.
> 
> ## Integrated nesting of SplitContainers
> Currently a SplitContainer has 2 panels. If we change `SplitPanel` to be a binary tree we can do let the user define as many sublevels as they want without having to create additional instances (good!).
> 
> In layout logic we can use recursion to pass ever reducing `Bounds` sizes into child branches based on the Splitter position consmed at the current level.
> 
> ## Estimate
> Lots of work! but sounds fun!

I'm so in love with this thinking I can't stand it.

I wonder if we can do just enough to the current PR to ship it quickly while ensuring we can layer this other stuff in without breaking backwards compat?

tznind:
StraightLineCanvas is comming along:

lines

> I wonder if we can do just enough to the current PR to ship it quickly while ensuring we can layer this other stuff in without breaking backwards compat?

It should be possible but I think i need to work this a bit longer to be sure that the solution will work.  Don't want to rename our panels `Left` and `Right` in anticipation of a binary tree structure then find out its not going to work out.

tig:
You guys are going to be disgusted by this, but for giggles, I asked ChatGPT to write a paneled window manager:

> I want to build a paneling window manager for Terminal.Gui in c#. Please write the code for a c# class that implements a paneled window manager that supports an arbitrary number of panels. The lines separating the panels should be draggable to allow the panels to be resized.

```cs
using System;
using Terminal.Gui;

namespace PanelDemo
{
    public class PanelManager
    {
        private List<Panel> panels;
        private List<Line> lines;
        private Window window;
        private int panelSize;
        private int lineSize;
        private int numPanels;

        public PanelManager(Window window, int numPanels, int panelSize, int lineSize)
        {
            this.window = window;
            this.numPanels = numPanels;
            this.panelSize = panelSize;
            this.lineSize = lineSize;

            panels = new List<Panel>();
            lines = new List<Line>();
            for (int i = 0; i < numPanels; i++)
            {
                Panel panel = new Panel(new Rect(0, 0, panelSize, window.Frame.Height));
                panels.Add(panel);
                window.Add(panel);

                if (i < numPanels - 1)
                {
                    Line line = new Line(new Rect(panelSize, 0, lineSize, window.Frame.Height));
                    lines.Add(line);
                    window.Add(line);
                }
            }
        }

        public void ResizePanel(int panelIndex, int newSize)
        {
            Panel panel = panels[panelIndex];
            int oldSize = panel.Frame.Width;
            panel.Frame = new Rect(panel.Frame.X, panel.Frame.Y, newSize, panel.Frame.Height);
            if (panelIndex < numPanels - 1)
            {
                Line line = lines[panelIndex];
                line.Frame = new Rect(line.Frame.X + newSize - oldSize, line.Frame.Y, line.Frame.Width, line.Frame.Height);
                panels[panelIndex + 1].Frame = new Rect(panels[panelIndex + 1].Frame.X + newSize - oldSize, panels[panelIndex + 1].Frame.Y, panels[panelIndex + 1].Frame.Width - newSize + oldSize, panels[panelIndex + 1].Frame.Height);
            }
            window.Refresh();
        }
    }
}
```

> To use this class, you will need to create an instance of the PanelManager class and pass in the Window object that you want to add the panels to, the number of panels you want, the initial size of each panel, and the size of the lines separating the panels. Then you can use the ResizePanel method to resize a panel by passing in the index of the panel and the new size.

```cs
Window window = new Window("Panel Demo");
PanelManager panelManager = new PanelManager(window, 3, 50, 2);

// Resize the second panel to be 60 units wide
panelManager.ResizePanel(1, 60);

```

tznind:
Starting to feel like a deep rabbit hole!

Not sure this recursive approach is the best.  I'll keep digging and see but does feel a bit like I am working against API rather than with it.

```csharp
class SplitPanel : View

    View Left      
    View Right
    Pos SplitterDistance
    
    
    public virtual void LayoutSubviews ()
    {
        var currentSize = Bounds;
        LayoutSubviews(this, currentSize);
    }
    
    private LayoutSubviews(SplitPanel panel, Rect bounds)
    {
    
        var leftSpace = panel.SplitterDistance.Anchor(currentSize)
        var rightSpace = bounds.Width - leftSpace;
        
        if(panel.Left is SplitPanel ls)
        {
            LayoutSubviews(ls,new Rect(bounds.X,bounds.Y, width:leftSpace, height: bounds.Height));
        }
        else
        {
            panel.Left.X = bounds.X;
            panel.Left.Y = bounds.Y;
            panel.Width = leftSpace;
            panel.Height = bounds.Height;
        }
        
        
        if(panel.Right is SplitPanel rs)
        {
            LayoutSubviews(rs,new Rect(bounds.X+leftSpace,bounds.Y, width:rightSpace, height: bounds.Height));
        }
        else
        {
            panel.Right.X = bounds.X + leftSpace;
            panel.Right.Y = bounds.Y;
            panel.Right.Width = rightSpace;
            panel.Right.Height = bounds.Height;
        }
        
        // TODO: Also position line view
    }
    
    // TODO: Override get child views to return from tree
    
    // TODO: How to suppress child layouts?
    
    // TODO: How to add new child views
    
}
    
```

BDisp:
> Not sure this recursive approach is the best. I'll keep digging and see but does feel a bit like I am working against API rather than with it.

No, what's happening is you are more closer to find the best way to do this.

tznind:
Haha thanks for your confidence :)

tznind:
As requested, bottomless recursion for SplitContainer! Still need to sort out titles and some other bits and bobs.  This PR now includes #2281 PR which is necessary for the line drawing.

recursion-split

tznind:
@BDisp I've got a strange bug where `Label` is not respecting its max bounds during layout in some situations.  Its very wierd.  Manifests in test `TestNestedContainer2LeftAnd1Right_RendersNicely`.  Failing line is where I try to confirm that `Label` Frame width is constrained correctly to the available space (10).  The issue does not seem to manifest when using `TableView` (see video).  Also the `Label` does not render its text the full length (the '1' characters stop at the right place) but they overwrite the right panel (2s).

```csharp
[Fact,AutoInitShutdown]
public void TestNestedContainer2LeftAnd1Right_RendersNicely()
{
	// [...]

	Assert.Equal (10, onesTop.Frame.Width);
	Assert.Equal (10, onesBottom.Frame.Width);

	splitContainer.Redraw (splitContainer.Bounds);

	string looksLike =
@"    
1111111111│22222222
1111111111│22222222
          │
          │
          │
──────────┤
          │
          │
          │
          │";
	TestHelpers.AssertDriverContentsAre (looksLike, output);


}
```

You can see the issue in action here.  It manifests when you set the number of panels to 3 or 7 and use `Label` instead of `TextView`.  If you don't have any ideas, I can try to manifest it using vanilla views.  It may be a side effect of moving a `Label` out of one `View` and putting it into another one? (This happens during splitting to create more panels).

wierd-layout-bug

BDisp:
Quick response, before I can view this. Do you are using AutoSize as true or false?

tznind:
> Quick response, before I can view this. Do you are using AutoSize as true or false?

`AutoSize` is true.  I am creating them using `new Label (new string ('1', 100))`

tznind:
Test passes if switching to TextView 

```diff
@@ -583,10 +583,10 @@ namespace UnitTests {

                        container.IntegratedBorder = withBorder ? BorderStyle.Single : BorderStyle.None;

-                       container.Panel1.Add (new Label (new string ('1', 100)));
-                       container.Panel1.Add (new Label (new string ('1', 100)) { Y = 1});
-                       container.Panel2.Add (new Label (new string ('2', 100)));
-                       container.Panel2.Add (new Label (new string ('2', 100)) { Y = 1});
+                       container.Panel1.Add (new TextView { Text = new string ('1', 100), Height = 1, Width = Dim.Fill()});
+                       container.Panel1.Add (new TextView { Text = new string ('1', 100), Height = 1, Width = Dim.Fill (), Y = 1});
+                       container.Panel2.Add (new TextView { Text = new string ('2', 100), Height = 1, Width = Dim.Fill ()});
+                       container.Panel2.Add (new TextView { Text = new string ('2', 100), Height = 1, Width = Dim.Fill (), Y = 1});

                        container.Panel1MinSize = 0;
                        container.Panel2MinSize = 0;
```

BDisp:
But I think using AutoSize as false and also set the Width=Dim.Fill(). If still not working also set the Height=1.

tznind:
> But I think using AutoSize as false and also set the Width=Dim.Fill(). If still not working also set the Height=1.

Thanks that worked! Is `AutoSize` **supposed** to allow a `Label` to overspill its parents bounds?

```csharp
[Theory, AutoInitShutdown]
[InlineData(true)]
[InlineData (false)]
public void Test_LayoutSizeOfLabel(bool label)
{
	var root = new View (){ Width = 20};

	var v = label == true ?
		new Label (new string ('c', 100)) {
			// Height = 1,          // <- Uncomment to make this test pass
			// AutoSize = false,
			Width = Dim.Fill () } :
		(View)new TextView () { 
			Height = 1,
			Text = new string ('c', 100),
			Width = Dim.Fill () };
	
	root.Add (v);

	Application.Top.Add (root);
	root.LayoutSubviews ();
	root.BeginInit ();
	root.EndInit ();

	root.LayoutSubviews ();
	Assert.Equal (20, v.Frame.Width);
}
```

BDisp:
> Thanks that worked! Is `AutoSize` **supposed** to allow a `Label` to overspill its parents bounds?

If it's using `X=0, Y=0,Width=Dim.Fill(),Height=Dim.Fill()` it must not  overspill its parents bounds. If it's using absolute values that is lower or greater than its parents bounds it's possible. If it's using `X=0, Y=0,Width=Dim.Fill(),Height=Dim.Fill()` and occupies all the  parents bounds, which has border and `DrawMarginFrame` is true, if the parent draw the border after the children, the border may overlap it's children. If overlap mustn't happen then the children must use pos and dim like `X=1, Y=1,Width=Dim.Fill(1),Height=Dim.Fill(1)`. That why the advantage of the content view as a container to avoid adding the necessary margin to the children. The `PanelView` use a similar approach. If `AutoSize` is true and the text is greater than the parent width then word wrap comes into action, but I think it may overlap the parent height, not sure. Please, let me know if a label `AutoSize` is false and `X=0, Y=0,Width=Dim.Fill(),Height=Dim.Fill()` and the label still overspill its parents bounds. Thanks.

BDisp:
Ok, thanks for the test. I see that it's failing in width, I think it's supposed the word wrap comes into action. I'll see this better.

BDisp:
@tznind I already remember about this. Do you remember when I asked if `AutoSize` is true the view `Width` and `Height` must be ignored and only get the text length and you answered if them are set then we must use its size if the text length is lower than the `Width` or `Height` size. So, the logic is correct because the text length is greater than the `Width` in this unit test. At the time we agree if we need the view size setting is respected then we also must se `AutoSize` to false. If you still think this is not perfect please  let me know.

tznind:
Ok, I think I understand.  Lets leave it for now then if behaviour is intended.  Sounds like a balancing act.

tig:
This PR is freaking amazing BTW.

The combination of "SplitContainer" and "LineCanvas" are such fundamental/rich capabilities: going to enable a ton of great apps, me thinks.

I now wonder if `SplitContainer` is the correct name? 

Consider this: The `SplitContainerNesting` Scenario, could be abstracted into the library as a "Tiled View Container". Or just "Tiled Container"? Or "TileView"? 

https://en.wikipedia.org/wiki/Tiling_window_manager

I guess I'm railing (really just brainstorming) against the choice of the word "Container" given the legacy of Terminal.Gui has never really used the term "Container" because "View" is fundamentally a container, thus "container" is superfluous.

tig:
(I'm actually excited to refactor `Wizard` to use all this at some point (esp the left & right panes in `WizardStep` and the area where the buttons go).

tig:
Sorry for all the comments, but I haz questions...

The latest code here, drives a model where tiles are nested (because SplitContainer has Panel1 and Panel2; if you want additional panels, you make Panel1 or Panel2 a container of a sub-view that is also SplitContainer).

If I want an enumeration of ALL Views that are in the hierarchy, I have to enum down a tree.

I have an intuition that there are cases where a developer will want a flat model. e.g. the case where all Views that are contained within the "TileView" are siblings of each other.  

```cs
// Append "!" to the Title of every tile
foreach (var view in TileView) {
   view.Title += "!";
}
```

@tznind, given the current architecture, is this possible?

BDisp:
> Also the `Label` does not render its text the full length (the '1' characters stop at the right place) but they overwrite the right panel (2s). You can see this in the video when you select the `Label` (I made them focusable) - the grey focus line spills into the next container but the text does not.

The case of the grey focus line spills into the next container it's a bug that I will investigate. Thanks for the catch @tznind.

BDisp:
One more observation to emphasize that `AutoSize` as true must grow if the text is bigger than its parents bounds is for the case of alignments. Imagine you have a view with `AutoSize` as true, text alignment as centered and his size is bigger than the parent' size. Forcing the child size the same of the parent's size will result in a erroneous center alignment behavior, but the child never should never drawn outside the its parent bound. Please, correct me if I'm wrong.

BDisp:
@tznind I already submitted the PR #2290 that fixed the clearing bug. Thanks for the catch. But the label size maintain higher than it parent. So, you can't rely on the label size but only on the lower size of both.

tznind:
> I now wonder if `SplitContainer` is the correct name?
> 
> Consider this: The `SplitContainerNesting` Scenario, could be abstracted into the library as a "Tiled View Container". Or just "Tiled Container"? Or "TileView"?
> 
> https://en.wikipedia.org/wiki/Tiling_window_manager
> 
> I guess I'm railing (really just brainstorming) against the choice of the word "Container" given the legacy of Terminal.Gui has never really used the term "Container" because "View" is fundamentally a container, thus "container" is superfluous.
> 
> "SplitView"?
> 
> "PanelView"? I like this since the two main properties of the class are named "Panel1" and "Panel2"?


### Naming

Good point.  I like `SplitView` and I can rename `Panel1` to `View1` and `Panel2` to `View2` if we are sticking with the current implementation.

### API
I have moved some of the nesting code into `SplitContainer` (e.g. see TrySplitPanel1) but I think its worth exploring how in an ideal world the user might want to interact with the API.  

I've detailed below how user would create various simple but common configurations but would like to open floor to how in an ideal world the user would interact with the control to create various panel layouts.

The current implementation expects user to do something like:

```csharp
var root = new SplitContainer { 
		Width = Dim.Fill (),
		Height = Dim.Fill (),
		Orientation = Terminal.Gui.Graphs.Orientation.Vertical,
		SplitterDistance = 30 };

root.TrySplitPanel1 (out SplitContainer lhs);

lhs.Orientation = Terminal.Gui.Graphs.Orientation.Horizontal;
lhs.SplitterDistance = Pos.Percent (25);

lhs.Panel1.Add (myNavigationPane);
lhs.Panel2.Add (myContentTreePane);
root.Panel2.Add (myMainContentView);
```
_Creating a 3 panel UI where left 2 appear ontop of each other_

If you want a 2x2 grid you would do 
```csharp
var root = new SplitContainer { 
		Width = Dim.Fill (),
		Height = Dim.Fill (),
		Orientation = Terminal.Gui.Graphs.Orientation.Vertical,
		SplitterDistance = 30 };

root.TrySplitPanel1 (out SplitContainer lhs);
root.TrySplitPanel2 (out SplitContainer rhs);

rhs.Orientation = Terminal.Gui.Graphs.Orientation.Horizontal;
rhs.Orientation = Terminal.Gui.Graphs.Orientation.Horizontal;

lhs.Panel1.Add (myXPlane);
lhs.Panel2.Add (myYPlane);
rhs.Panel1.Add (myZPlane);
rhs.Panel2.Add (my3DRender);
```

tig:
> "how in an ideal world the user would interact with the control?"

I have been thinking about this. Here's a scenario:

- I'd like the `ConfigurationEditor` scenario I wrote to use SplitView instead of TabView. It would have 6 vertical split views, one for each config file/resource (the resources TextViews would be read only):

```cs
		/// <summary>
		/// Describes the location of the configuration files. The constancts can be
		/// combined (bitwise) to specify multiple locations.
		/// </summary>
		[Flags]
		public enum ConfigLocation {
			/// <summary>
			/// Application configuration found in the users's home directory (<c>~/.tui/appname.config.json</c>) -- Highest precidence 
			/// and the default for <see cref="ConfigurationManager.Save()"/>.
			/// </summary>
			AppHomeDirectory,

			/// <summary>
			/// Global configuration in the directory the app was launched from (<c>./.tui/config.json</c>).
			/// </summary>
			GlobalAppDirectory,

			/// <summary>
			/// Global configuration in the user's home directory (<c>~/.tui/config.json</c>).
			/// </summary>
			GlobalHomeDirectory,

			/// <summary>
			/// Application configuration in the app's resources (<c>appname.Resources.config.json</c>).
			/// </summary>
			AppResources,

			/// <summary>
			/// Application configuration in the directory the app was launched from.
			/// </summary>
			AppDirectory,

			/// <summary>
			/// Global configuration in <c>Terminal.Gui.dll</c>'s resources (<c>Terminal.Gui.Resources.config.json</c>) -- Lowest Precidence.
			/// </summary>
			LibraryResources,

			/// <summary>
			/// This constant is a combination of all locations
			/// </summary>
			All = AppHomeDirectory | GlobalAppDirectory | GlobalHomeDirectory | AppResources | AppDirectory | LibraryResources

		}
```

image

So instead of the above, you'd have a1x6 grid.

I'd expect to be able to create/manage (so I could show/hide locations the user didn't want to see) like this:

```cs
foreach (var configFile in configFiles) {
	var homeDir = $"{Environment.GetFolderPath (Environment.SpecialFolder.UserProfile)}";
	FileInfo fileInfo = new FileInfo (configFile.Replace ("~", homeDir));
	string json;
	json = File.ReadAllText (fileInfo.FullName);

        var textView = new TextView () {
	  X = 0,
	  Y = 0,
	  Width = Dim.Fill (),
	  Height = Dim.Fill (),
   	  Text = initialText
         };

         // Add a panel to the right of the last one

         // Add textView to that panel
}

How would this look with your concept?

Take this to the next level, and enable spitting editor panes arbitrarily like VSCode does:

![66Rshzr 1](https://user-images.githubusercontent.com/585482/213715868-6d754552-6694-4a2d-af26-5ce79c0ca8e1.gif)

**Comment by tznind** [comment]
> Take this to the next level, and enable spitting editor panes arbitrarily like VSCode does:

Ok this is a good target.  I think it is reasonable to write a scenario that allows for this kind of SplitUp/SplitDown etc experience.

I started exploring implementing this and realised that getting a context menu up for tabs is needlessly difficult currently so have created this seperate PR https://github.com/gui-cs/Terminal.Gui/pull/2299

**Comment by tznind** [comment]
Is this the experience you are looking for @tig ?  I've only implemented SplitRight for now and its buggy when you split a panel you've already split (e.g. split right then split the right pane right again).  But just trying to get the visuals down.  Its making me feel like theres scope for a root level view e.g. `DynamicSplitView` that hosts and manages its subcontainers and has the `SplitLeft`, `SplitRight`, `SplitUp`, `SplitDown` methods on it.  Kind of like a facade design pattern.

But if it gets good enough it might even be possible to make `SplitView` internal and only expose the `DynamicSplitView` (naming to be confirmed).

splitting

**Comment by tig** [comment]
> Is this the experience you are looking for @tig ? 

Yep!

**Comment by tznind** [comment]
Getting closer.  Some layout bugs and bad measurements but the pattern is now there.

I've come around to the `Tile` idea.  So a SplitView (probably to be renamed TileView) can have any number of `Tile` (including 1).  All tiles have the same orientation (e.g. all horizontal or all vertical).  

If you want to switch orientation you still need to nest but you can just convert a single `Tile` into a new `SplitView` with the alternate orientation.

This allows all tiles on a plane to resize equally nicely and makes for simpler insertion (see `InsertTile`).

And the best bit is we got rid of `View1` and `View2` stuff.  A `SplitView` has N `Tiles` and N-1 `Splitter lines`

tilesyay

File: Terminal.Gui/Views/TabView.cs [hunk 11]
```diff
@@ -194,8 +194,8 @@ public override void Redraw (Rect bounds)
 				int spaceAtBottom = Math.Max (0, GetTabHeight (false) - 1);
 				int startAtY = Math.Max (0, GetTabHeight (true) - 1);
 
-				DrawFrame (new Rect (0, startAtY, bounds.Width,
-			       Math.Max (bounds.Height - spaceAtBottom - startAtY, 0)), 0, true);
+				DrawFrame (new Rect (0, startAtY, Bounds.Width,
```
tznind, line 197: Should be able to revert this change once https://github.com/gui-cs/Terminal.Gui/pull/2309 is merged.

Terminal.Gui/Core/View.cs
```diff
@@ -1134,7 +1134,7 @@ public void Clear (Rect regionScreen)
 		/// <param name="rcol">Absolute column; screen-relative.</param>
 		/// <param name="rrow">Absolute row; screen-relative.</param>
 		/// <param name="clipped">Whether to clip the result of the ViewToScreen method, if set to <see langword="true"/>, the rcol, rrow values are clamped to the screen (terminal) dimensions (0..TerminalDim-1).</param>
-		internal void ViewToScreen (int col, int row, out int rcol, out int rrow, bool clipped = true)
+		public void ViewToScreen (int col, int row, out int rcol, out int rrow, bool clipped = true)
```
tznind, line 1137: ScreenToView is public already so hopefully its ok to make this also public (@BDisp thoughts?).

Its very handy for drawing stuff to screen directly and translating between coordinate systems (e.g. to draw interconnected borders).

Terminal.Gui/Views/LineView.cs
```diff
@@ -71,6 +71,7 @@ public LineView (Orientation orientation)
 		public override void Redraw (Rect bounds)
 		{
 			base.Redraw (bounds);
+			bounds = Bounds;
```
tznind, line 74: Should be able to revert this change once https://github.com/gui-cs/Terminal.Gui/pull/2309 is merged.

Terminal.Gui/Core/View.cs
BDisp: I agree.

UICatalog/Scenarios/SplitContainerExample.cs
tznind: Fully implemented now, sub split containers all join up nicely now.

UICatalog/Scenarios/SplitContainerExample.cs
tznind: Resolved this because we now use `IntegratedBorder`.  `IntegratedBorder` is unique to `TileView`.  It is drawn **within** the `TileView` client area (whereas a regular border is outside) and searches child elements recursively to pull out sub elements that should be joined up into one big seamless grid and border.

Terminal.Gui/Views/SplitContainer.cs
tznind: Now refactored to use `LineCanvas` so resolving conversation.

tznind:
@tig this is now ready for review.  I've left some comments though.

Specifically there are some lines that exist to work around https://github.com/gui-cs/Terminal.Gui/pull/2309 which can be removed if that is merged first.

Also if its easier to review seperately you could tackle #2281 pull request since this is a supper set of that branch.

tznind:
Found a bug, titles can spill over into other `Tiles`.  Should be possible to stop this.

shot-2023-02-01_20-07-08

tznind:
@tig do you want me to re-target this at v2 branch?

BDisp:
If my opinion is allowed I think if a PR doesn't cause any break changes it may be target the develop branch, because the v2 branch will always be merged with the develop branch before releasing.

tznind:
Cool, makes sense 👍

tig:
@tznind can you resolve merge conflicts?

tznind:
> @tznind can you resolve merge conflicts?

Conflicts fixed @tig and have removed the now superfluous Bounds workarounds.  

I did have to add `SetNeedsDisplay` on a lot of my tests  after merging.  Not a problem since they are redraw heavy tests but might be worth noting for later.  I didn't see any changes in behavior of the Scenarios (i.e. things not updating) so should be fine.

tig:
I wanted to write some code against this before merging (redo of AllViewsTester and ConfigurationEditor) but I now think the right thing to do is merge it now because it's so damn epic and any issues I find are likely on the margins. 

I'll do a deep CR asap!

tig (changes requested):

UICatalog/Scenarios/TileViewNesting.cs
```diff
@@ -0,0 +1,243 @@
+﻿using System;
```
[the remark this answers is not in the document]
BDisp: I think the joints is placed wrongly on the upper side instead of on the middle.

tig:
Here's Wizard as-is today (but with FrameView.DefaultBorder set to have a double style just to make a point).

image

Via `LineView` and `SplitView` it would be awesome if Wizard could be refactored (and other apps easily built) such that:

a) the horiztontal line above the buttons had the correct T connectors with the frame
b) the helptext box on right had a splitter between it and the controls of the step to the left (again, with proper Ts at top and bottom).

Having the correct Ts is one of the reasons I'm so excited about the work on this PR. Apps look SOOOO MUCH BETTER when this is done right. It really helps make Terminal.Gui shine, IMO.

I think you punted on mixed border styles (e.g. you can't mix Single and Double); can you opine on whether you think it's impossible with current SplitView architecture, possible but hard, or easy but will just take time?

tznind:
> I think you punted on mixed border styles (e.g. you can't mix Single and Double); can you opine on whether you think it's impossible with current SplitView architecture, possible but hard, or easy but will just take time?

This is doable and I have a design pattern in mind.  I've started coding up the approach in https://github.com/tznind/gui.cs/tree/line-canvas-style-mixing .  Basically we isolate each intersection type into its own 'resolver' class so it is maintainable and we just handle all the relevant permutations in each of those.

tig:
I suck at git. I just accidentally force pushed an experiment to your splitcontainer branch, @tznind.

I meant to submit a PR to your PR. Ugh.

tig:
@tznind please forgive me for my incompetence. I think the PR I just submitted to your fork (for this PR) fixes the build break.

Anyway, please checkout the experiment i added in that PR (and force pushed). I modified FrameView such that this code magically does all the borders correctly. E.g. is the whole `TIleView` needed if we can go a route like this??

```cs
public override void Setup ()
		{
				new MenuItem ("_Quit", "", () => Application.RequestStop()),
			}) });

			Application.Top.Add (menu);

			var frame = new FrameView () {
				X = 0,
				Y = 1,
				Width = Dim.Fill (),
				Height = Dim.Fill (),
				IgnoreBorderPropertyOnRedraw = true
			};
			frame.Border.BorderStyle = BorderStyle.Double;

			Application.Top.Add (frame);

			var view1 = new FrameView () {
				Title = "View 1",
				Text = "View1 30%/50% Single",
				X = -1,
				Y = -1,
				Width = Dim.Percent (30),
				Height = Dim.Percent (50),
				ColorScheme = Colors.ColorSchemes ["Dialog"],
				Border = new Border () { BorderStyle = BorderStyle.Single }
			};

			frame.Add (view1);

			var view2 = new FrameView () {
				Title = "View 2",
				Text = "View2 right of view1, 30%/Fill Single.",
				X = Pos.Right (view1) - 1,
				Y = -1,
				Width = Dim.Percent (30),
				Height = Dim.Fill (-1),
				ColorScheme = Colors.ColorSchemes ["Error"],
				Border = new Border () { BorderStyle = BorderStyle.Single }
			};

			frame.Add (view2);

			var view3 = new FrameView () {
				Title = "View 3",
				Text = "View3 right of View2 Fill/Fill Single",
				X = Pos.Right (view2) - 1,
				Y = -1,
				Width = Dim.Fill (-1),
				Height = Dim.Fill (-1),
				ColorScheme = Colors.ColorSchemes ["Menu"],
				Border = new Border () { BorderStyle = BorderStyle.Single }
			};

			frame.Add (view3);

			var view4 = new FrameView () {
				Title = "View 4",
				Text = "View4 below View1 30%/5 Single",
				X = -1,
				Y = Pos.Bottom (view1)-1,
				Width = Dim.Percent (30),
				Height = 5,
				ColorScheme = Colors.ColorSchemes ["TopLevel"],
				Border = new Border () { BorderStyle = BorderStyle.Single }
			};

			frame.Add (view4);

			var view5 = new FrameView () {
				Title = "View 5",
				Text = "View5 below View4 view4.Width/5 Double",
				X = -1,
				Y = Pos.Bottom (view4) - 1,
				Width = view4.Width,
				Height = 5,
				ColorScheme = Colors.ColorSchemes ["TopLevel"],
				Border = new Border () { BorderStyle = BorderStyle.Double }
			};

			frame.Add (view5);
		}
```

image

tig:
This was all accopmlished by modifying `Frameview.Redraw` thusly:

```cs
if (!IgnoreBorderPropertyOnRedraw) {
				Driver.SetAttribute (GetNormalColor ());
				//Driver.DrawWindowFrame (scrRect, padding + 1, padding + 1, padding + 1, padding + 1, border: true, fill: false);
				Border.DrawContent (this, false);
				if (HasFocus)
					Driver.SetAttribute (ColorScheme.HotNormal);
				if (Border.DrawMarginFrame)
					Driver.DrawWindowTitle (scrRect, Title, padding.Left, padding.Top, padding.Right, padding.Bottom);
				Driver.SetAttribute (GetNormalColor ());
			} else {
				Driver.SetAttribute (ColorScheme.Normal);

				var lc = new LineCanvas ();

				if (Border?.BorderStyle != BorderStyle.None) {

					lc.AddLine (new Point (0, 0), bounds.Width - 1, Orientation.Horizontal, Border.BorderStyle);
					lc.AddLine (new Point (0, 0), bounds.Height - 1, Orientation.Vertical, Border.BorderStyle);

					lc.AddLine (new Point (bounds.Width - 1, bounds.Height - 1), -bounds.Width + 1, Orientation.Horizontal, Border.BorderStyle);
					lc.AddLine (new Point (bounds.Width - 1, bounds.Height - 1), -bounds.Height + 1, Orientation.Vertical, Border.BorderStyle);
				}

				foreach (var subview in contentView.Subviews) {
					lc.AddLine (new Point (subview.Frame.X+1, subview.Frame.Y+1), subview.Frame.Width - 1, Orientation.Horizontal, subview.Border.BorderStyle);
					lc.AddLine (new Point (subview.Frame.X+1, subview.Frame.Y+1), subview.Frame.Height - 1, Orientation.Vertical, subview.Border.BorderStyle);

					lc.AddLine (new Point (subview.Frame.Width - 1, subview.Frame.Y + subview.Frame.Height), -subview.Frame.Width + 1, Orientation.Horizontal, subview.Border.BorderStyle);
					lc.AddLine (new Point (subview.Frame.X + subview.Frame.Width, subview.Frame.Height - 1), -subview.Frame.Height + 1, Orientation.Vertical, subview.Border.BorderStyle);
				}

				//TODO: Figure out how to draw Titles
				
				// TODO: Figure out how to add splitters
}
```
 Which is hacky, but shows directionally that this idea should work.

To get the Splitters, my thinking is to detect where they're needed in `LayoutSubviews` and add them.

Or, perhaps, not do it automatically, but let the caller (who added the subviews to `FrameView`) add them explicitly.  Noodling on this now.

tig:
Titles were easy:
image

```cs
// FrameView.Redraw...
var lc = new LineCanvas ();

				if (Border?.BorderStyle != BorderStyle.None) {

					lc.AddLine (new Point (0, 0), bounds.Width - 1, Orientation.Horizontal, Border.BorderStyle);
					lc.AddLine (new Point (0, 0), bounds.Height - 1, Orientation.Vertical, Border.BorderStyle);

					lc.AddLine (new Point (bounds.Width - 1, bounds.Height - 1), -bounds.Width + 1, Orientation.Horizontal, Border.BorderStyle);
					lc.AddLine (new Point (bounds.Width - 1, bounds.Height - 1), -bounds.Height + 1, Orientation.Vertical, Border.BorderStyle);
				}

				foreach (var subview in contentView.Subviews) {
					lc.AddLine (new Point (subview.Frame.X + 1, subview.Frame.Y + 1), subview.Frame.Width - 1, Orientation.Horizontal, subview.Border.BorderStyle);
					lc.AddLine (new Point (subview.Frame.X + 1, subview.Frame.Y + 1), subview.Frame.Height - 1, Orientation.Vertical, subview.Border.BorderStyle);

					lc.AddLine (new Point (subview.Frame.X + subview.Frame.Width, subview.Frame.Y + subview.Frame.Height), -subview.Frame.Width + 1, Orientation.Horizontal, subview.Border.BorderStyle);
					lc.AddLine (new Point (subview.Frame.X + subview.Frame.Width, subview.Frame.Y + subview.Frame.Height), -subview.Frame.Height + 1, Orientation.Vertical, subview.Border.BorderStyle);

				}

				Driver.SetAttribute (ColorScheme.Normal);
				lc.Draw (this, bounds);

				foreach (var subview in contentView.Subviews) {
					if (subview is FrameView viewWithTite) {
						var rect = viewWithTite.Frame;
						rect.X = rect.X + 1;
						rect.Y = rect.Y + 2;
						Driver.DrawWindowTitle (rect, viewWithTite.Title, padding.Left, padding.Top, padding.Right, padding.Bottom);
					}
				}
```

tig:
Two thoughts just collided:

1) I don't like the fact that keyboard access to the splitter is via TAB. E.g. in UICatalog it breaks the simplicity of using Tab to toggle between the two panes (I have to double-tab now). Suggestion: Let's define a standard keystroke that activates "sizing" and subsequent presses navigates across possible points where sizing can occur. E.g. `F10`.

2) View should be optionally sizable on all 4 sides. E..g `view.IsSizable = true` -> diamonds show up on all 4 borders. Click/Drag on any border supports resize. Or hit F10 and left border diamond shows foucus. HIt F10 again and it moves to top, etc... 5th press resumes previous focus.

If we did this, then setting up the subviews of a FrameView as I show above, will automatically get splitter functionality (mostly).

I think this concept will work. Pretty excited about it.

tznind:
Love this work on `FrameView`.  Looks really nice and the way the subframes float freely looks very cool!

tznind:
I have fixed the layout bug when horizontal orientation with a border.  I've also added a Ctrl+F10 key shortcut to toggle splitter focusability.

image

tig:
FWIW, I've been massively distracted refactoring View for v2. I have some eggciting results that I hopefully can share soon that will make all of this work you've done be even more magic.

tig (approved):

tig (approved):

tznind:
I'll need to update this to the new method signature of `LineCanvas`, shouldn't take long, should manage thisevening.

tznind:
Think tests are failing due to this bug which manifests when nesting normal (non TileViews too), see repro.

https://github.com/gui-cs/Terminal.Gui/issues/2368

tig (approved):

Terminal.Gui/Views/FrameView.cs
```diff
@@ -85,9 +86,10 @@ class ContentView : View {
 		/// <param name="title">Title.</param>
 		/// <param name="views">Views.</param>
 		/// <param name="border">The <see cref="Border"/>.</param>
-		public FrameView (Rect frame, ustring title = null, View [] views = null, Border border = null) : base (frame)
+		public FrameView (Rect frame, ustring title = null, View [] views = null, Border border = null) //: base (frame)
```
tznind, line 89: `Terminal.Gui.ViewTests.FrameViewTests.Constuctors_Defaults` is failing.

If I remove the comment so it is calling the base constructor again it passes.  I think you made this change @tig, is there a reason not to call base? and if so does it mean the test needs changed? Or should I just remove the comment?

tig (approved):

tig:
@tznind can you please try to get this passing again. the unit tests are failing after accepting your change and updating from `develop`

tznind:
Ok I think the failing tests are due to me merging https://github.com/tznind/gui.cs/pull/158 so I have reverted it.

Was that PR purely stuff already in develop? if so something must have gone wrong with me merging it.

If theres other work in it then I can look into porting it to the this branch.

tznind:
Hmn seems to be an issue with Tab navigation after splitting containers in Notepad Scenario.  You end up getting stuck in a cycle where you can't get across to there other tile no matter how much you hit Tab.  Think I need to look into that and see if theres a 'recompute tabs' method or something.

BDisp:
> Hmn seems to be an issue with Tab navigation after splitting containers in Notepad Scenario. You end up getting stuck in a cycle where you can't get across to there other tile no matter how much you hit Tab. Think I need to look into that and see if theres a 'recompute tabs' method or something.

The better choose is perhaps using the alternate key navigation which default is the `Ctrl+Tab` for forward and `Shift+Ctrl+Tab` for backward.

tznind:
Ok heres a minimum repro (see below).  But the weird thing is that Shift+Tab works! its only Tab that gets stuck looping.

tab-repeatedly
_Pressing Tab repeatedly_

shift-tab-repeatedly
_Pressing shift tab repeatedly_

Repro Code:
```csharp
Application.Init ();

var t = new TileView (1) {Y = 1, Width = Dim.Fill (), Height = Dim.Fill () };
t.Tiles.ElementAt (0).ContentView.Add (new TextField { Width = 10, Height = 1, Text = "111" });
var result = t.InsertTile (1);		
result.ContentView.Add (new TextField { Width = 10, Height = 1, Text = "222" });
var w = new Window ();
w.Add (t);
w.Add (new TextField { Text = "000" ,Width = 10,Height = 1});

Application.Run (w);
Application.Shutdown ();
```

tznind:
Ok I think the problem is this code in OnEnter

```csharp
if (!Tiles.Where (t => t.ContentView.HasFocus).Any ()) {
		Tiles.FirstOrDefault ()?.ContentView.SetFocus ();
	}
```

BDisp:
> Ok I think the problem is this code in OnEnter

Ah right. You have to ensure if there have or not another tile view. If exist and is the last view focused and Tab pressed then must navigate to the first view of the tile view.

tznind:
Ok should be working now.  I have deleted the entire `OnEnter` override and made the `TileView` itself `CanFocus` false (the default).

BDisp:
Yet better, let the View base work :-)

BDisp:
It's working but has a new wrong behavior. The navigation is backwards.

tznind:
> It's working but has a new wrong behavior. The navigation is backwards.
> 
> **Edit:** Never mind. It's the right navigation. So when a new tile is inserted the last being focused. So first "222", "111" and last "000".

Yeah I think it mostly depends what order you split and add to `TileView`.  For example in the `Notepad` example if you SplitLeft then take left pane and **SplitDown** then tab navigation works correctly but feels off since it enters the bottom from the right then goes up:

split-left


But if you SplitLeft then take the left pane and **SplitUp** then tab navigation works and feels right.

split-up

tznind:
Hmn test failures are `System.InvalidOperationException : Cannot set CanFocus to true if the SuperView CanFocus is false!`  

They are for when you make the splitter lines focusable during runtime.

BDisp:
Well if you want to order always by left top corner to down and right bottom corner, then you have to use `BringSubviewForward`, `SendSubviewBackwards`, `BringSubviewToFront` or `SendSubviewBackwards`, accordingly.

tznind:
> Well if you want to order always by left top corner to down and right bottom corner, then you have to use `BringSubviewForward`, `SendSubviewBackwards`, `BringSubviewToFront` or `SendSubviewBackwards`, accordingly.

Yeah don't think I want to do that! feels a bit hacky/risky and I think the current behaviour is ok. 

Might look into adjusting the `tabIndexes` during `InsertTile` or `TrySplitTile` though.

BDisp:
> Hmn test failures are `System.InvalidOperationException : Cannot set CanFocus to true if the SuperView CanFocus is false!`
> 
> They are for when you make the splitter lines focusable during runtime.
> 
At runtime you have to set the superview can focus to true before.

> I guess maybe I need to remove the line views and then add them again as focusable or something
> 
If you are adding at load a line view that has CanFocus as true it will also sets the superview CanFocus to true.

> Strangely it doesn't affect the Scenarios only the tests.
Because the scenarios ha already initialized by the `Application.Begin`.

Let me say some clarification on this. `Toplevel.CanFocus` is false. When a subview with `CanFocus` as true and if toplevel is enabled, then the `Toplevel.CanFocus` is sets to true, but this is done when the subview is adding to the superview. At runtime if you sets the superview `CanFocus` to false it also will sets all the subviews `CanFocus` to false.

tznind:
Cool, so the fix I applied for the `CanFocus` buisness is https://github.com/gui-cs/Terminal.Gui/pull/2258/commits/7929084591c753b2841bd8ce2a38495c95ea54d0 which seems to avoid the throw even if it seems a bit hacky.  And since the `LineView` has no children it shouldn't have any unforseen consequences I think.

BDisp:
> Cool, so the fix I applied for the `CanFocus` buisness is [7929084](https://github.com/gui-cs/Terminal.Gui/commit/7929084591c753b2841bd8ce2a38495c95ea54d0) which seems to avoid the throw even if it seems a bit hacky. And since the `LineView` has no children it shouldn't have any unforseen consequences I think.

Well I don't agree. Even it have no children, but it could can get the focus to manage his position. Imagine you want to move the line through the keyboard. For that it needed to get focus first for being receiving keys events. But if the superview `CanFocus` is false, then you can't also focused the `LineView`. Make sense.

tznind:
> > Cool, so the fix I applied for the `CanFocus` buisness is [7929084](https://github.com/gui-cs/Terminal.Gui/commit/7929084591c753b2841bd8ce2a38495c95ea54d0) which seems to avoid the throw even if it seems a bit hacky. And since the `LineView` has no children it shouldn't have any unforseen consequences I think.
> 
> Well I don't agree. Even it have no children, but it could can get the focus to manage his position. Imagine you want to move the line through the keyboard. For that it needed to get focus first for being receiving keys events. But if the superview `CanFocus` is false, then you can't also focused the `LineView`. Make sense.

Ah there is a keypress (Ctrl+F10 by default) that makes the lines focusable so user can tab to them and move them with cursor keys :).  In fact it is that toggling that was originally throwing the InvalidArgumentException.

BDisp:
> I've also seen that even if `CanFocus` is false you can still focus children. Even the default `View` class has `CanFocus` of false but you can select stuff that you have added to it no problem.

The trick is bellow. It's handled on `View.Add` method. At runtime if the superview `CanFocus` is false you can't add a subview with `CanFocus` as true without first sets the superview `CanFocus` to true. This is the expected the desired behavior. If a superview `CanFocus` was sets to false at runtime, we need to inspect his value before adding a subview with `CanFocus` as true to it.

https://github.com/gui-cs/Terminal.Gui/blob/9cfa78a0330f6ff05c9460d7064fd9167d90d60d/Terminal.Gui/Core/View.cs#L943-L953

tig:
@tznind - same question about v1 v v2 for this.

tznind:
Sure, if it saves work I'm all for it

tznind:
I hadn't realized this was already in v2.  I've fetched and merged v2 into this and now the only changes are the toggle button and tab order fixes.  It also targets the v2 branch now.

tznind:
> Yeah, I needed it for Config Manager so merged it a bit ago.
> 
> Why does this PR have all the config manager stuff in it as well?

This PR is now only +102 / −62 after merging `v2` branch into it (`splitcontainer`) and re targeting the pr to merge back into `v2`.  I thought that was the plan? but didn't realize it had in effect already been merged into v2.

Looks like there are only very recent (unmerged) changes now in this branch that are not still on v2.

The configuration manager stuff is already in v2.

tig (approved):
Great stuff. Thanks for doing the dirty work on this one.

Be aware, however, that I have a crazy idea that `TileView` may not be needed as a separate view my v2 `View` redesign works. We shall see... ;-)

PR #2259, opened 2022-12-28
Fixes #2150. Revamping FileDialog

### LINKED ISSUE #2150: File Dialogs: Re-implement using `TreeView` + `ListView` etc...
image

This would be so much better if there was a TreeView on the left and a ListView on the right.

And a lot of crusty code would be nuked...

tznind:
Fixes #2150.

FileDialog2

**Note:** This branch/PR includes #2258 (SplitContainer)

The goals are:

### Speed

If using a console TUI, entering a file/directory should be as fast as on the shell directly

- Allow immediate direct typing of the full/remaining path
  - no tabbing around the UI
  - no moving cursor around in text box etc
- Support tab autocomplete while typing
- Support up/down cycling autocomplete

### Coloring

Listing files on the console with `ls` gives coloring that shows what is a directory, what is an image, executable etc.  

- Highlight contents at least as well as `ls` command
- Use colors consistent with `ls`

### Navigation (Backwards/Forwards/Up)

After making a navigation into a folder there should be a single key click you can make to take you back.  Once moving back you should be able to use another keystroke to move forward again.

- Hitting `Backspace` in the Table view changes history
- `Shift+Backspace` moves forward again

### More File Info

Being able to sort by FileSize or Date Modified is super handy

- Support hideable/sortable columns for these file details

### Other Places
Most GUI file explorers have either a list or a tree recent/root places (e.g. drives)

- Show relevant `SpecialFolder` and `LogicalDrives` on left in a collapsable pane.

### TODO

- [x] Fix all the bugs
- [x] Make it bulletproof (to permissions, invalid characters etc)
- [x] Cancellable event for picking existing e.g. 'x.doc already exists, Do you want to replace it'.  This allows Save dialogs to function as on Windows/Linux.  If user hits No they can continue to make a different selection.
- [x] Eliminate duplicates in tree view
- [ ] Keyboard controls for column operations (show menu / sort / hide etc)
- [x] Allow delete files/folders
- [x] Allow create folder
- [x] Allow rename files/folders
- [x] Swallow <Enter> on Search (or restart the search - this covers use case where you search then tab to new dir then hit Ctrl+F to focus search again and hit enter to restart the search in the new dir).
- [x] Empower AllowedType e.g. Regex, Default=true etc
- [x] AllowedType should have max length e.g. if user has 20 file extensions listed
- [x] Default contains search implementation should treat "blah csv" as 2 search terms (users can always override with custom interface implementation).
- [ ] Make things controllable from config manager
  - [ ] Including column stuff e.g. minimum size, column default visibility etc.
- [x] Fix Title centering when width is odd number and ensure no leading whitespace in title.
- [x] Multi select
- [x] Select File/Folder
- [x] Filter on type  (e.g. demand user open only a *.csv file)
- [x] Maybe some kind of 'contains' search box
- [x] For **scenario**, don't use Unicode characters for 'icons' on windows where they are unlikely to not be installed.  Instead maybe use `' '` for files and `'\'` for directories
- [x] Dark blue for folders is quite hard to read on Windows, default to monochrome mode or light blue?
- [x] Home/End in TableView should go to first/last row not first/last cell (improves UX when acting like a list)
- [x] Tests
- [x] Moving selection to `..` should set the text entry to the current dir (not last selected file at that level)
- [x] Pressing a letter in the TableView should cycle selection (like with `ListView`).
- [x] Pressing `LEFT` cursor when TileView is expanded and first cell in table is selected should move focus to the tree
- [x] Performance test - go to '/' and search for 'e'. UI should remain responsive even as thousands of files are found.
- [ ] Prevent re-entry e.g. sym link to parent dirs when searching (maybe just don't search them)
- [ ] Show sym links in light blue see: https://stackoverflow.com/a/26473940/4824531

## # Additional Review TODOs
--------------------------
- [x] Show initial sort order
- [x] Revisit Icons so good default behaviour and understandable docs for user configuration
- [x] Split up .cs into new folder for filesystem
- [x] Make spinner reusable
- [x] Improve usability when pushing a column of table off the screen
- [x] Revisit forward/backward/up single character buttons
- [x] get rid of Path chevron
- [x] Respect ColorScheme more
- [ ] Run Ctrl+K,D on all changed files at end

### Reference Examples

Since we are in the domain of terminal use here we should first look at what the basic console offers.  The user can select a directory with autocomplete by typing it directly and hitting tab to complete or enter to confirm.  Some terminals have more or less autocomplete e.g. with [fish](https://github.com/fish-shell/fish-shell) will show full autocomplete suggestions in gray:

image
_Folder selection in the console_

image
_Windows file dialog_

image
_Save/Export dialog in [GIMP](https://www.gimp.org/)_

image
_MacOS file open dialog_

image
_MS-Dos open dialog for text editor_

image
_Terminal.Gui file dialog_

## Pull Request checklist:


tig:
Awesome!

BDisp (commented):
You also can use `ustring.IsNullOrEmpty(Title)`.

tznind:
Have put lots of work in polishing this and would be great to get some feedback.  

The PR is a super set of #2258 (SplitContainer) so probably shouldn't be merged until that is in but other than that it is ready for review.  

Theres probably some other stuff that could be added e.g.

- Refresh (e.g. F5)
- New File/Folder via context menu

tznind:
Marking this as draft while SplitContainer is still WIP

tig:
Gotta admit, I'm kinda digging this new title bar style:

image

I haven't looked at the code yet, but can the alignment be set?

I need to take this idea into account in the new v2 View design!

tznind:
> Gotta admit, I'm kinda digging this new title bar style:
> 
> image
> 
> I haven't looked at the code yet, but can the alignment be set?
> 
> I need to take this idea into account in the new v2 View design!

Thanks! yeah its pretty cool eh? I was using `nmtui` the other day and noticed the effect they had on the title bar so thought I would do something similar on the file dialog.

image

At the moment its just in the `Redraw` method of `FileDialog2` but I think its worth refactoring into a standard option for v2 titles (with alignment as you say ).  This implementation is hard coded to centre align.

Also I am now noticing a bug where the top double line doesn't go all the way to the end in some resolutions.  Think the integer division /2 for centering is wrong when its not divisible by 2.

image

tig:
@tznind how would you react to me suggesting we target this PR for v2 and not v1?

We have already reached the point where v1 (main/devleop) have diverged so much from v1 that automatic merging no longer works. It is becoming herculean (at least for me) to try to merge new stuff from v1 into v2.  

I'd like us all to just stop working on v1 except for critical bugs. What do you think?

tznind:
Sure, that works for me.

If we can get this PR to the point where it is feature equivalent to the
old dialog we could just replace it with the new one? It would be good not
to have to maintain 2 file dialogs.

Happy to put some more work in to fill in any gaps in functionality.

What do you think @BDisp?

BDisp:
I agree.

BDisp:
@tznind this is also used for Save Dialog? Also have check mark for multi-select and how to do it? Sorry for asking.

tig:
> Sure, that works for me. If we can get this PR to the point where it is feature equivalent to the old dialog we could just replace it with the new one? It would be good not to have to maintain 2 file dialogs. Happy to put some more work in to fill in any gaps in functionality. What do you think @BDisp?

Not sure we're aligned. 

This PR is dependent on TileView, which I think we just agreed will only go in v2 now.

BDisp:
> Not sure we're aligned.
> 
> This PR is dependent on TileView, which I think we just agreed will only go in v2 now.

I think @tznind was referring that agreed with moving this to the v2 and replacing the others original file dialog with this to avoiding maintain both on the v2. @tznind don't forget you have to close this one and create another PR to merge in the v2 branch, but don't forgot to merge the v2 first before creating a new branch. Thanks.

tig:
> > Not sure we're aligned.
> > This PR is dependent on TileView, which I think we just agreed will only go in v2 now.
> 
> I think @tznind was referring that agreed with moving this to the v2 and replacing the others original file dialog with this to avoiding maintain both on the v2. @tznind don't forget you have to close this one and create another PR to merge in the v2 branch, but don't forgot to merge the develop branch into the v2 branch first before creating a new branch. Thanks.

Regarding merging `develop` into `v2`: Please see my comment here:

https://github.com/tig/Terminal.Gui/pull/10#issuecomment-1448382665

In short: I think we should STOP trying to merge `develop` into `v2` until we agree on a plan for v2. I suspect the right thing to do is to revert `develop` to the `v1.9` tag and THEN merge it into `v2` for the LAST TIME. From that point forward, any changes to v1 will need to be PORTED to v2 (not merged (unless merging is the obvious simple way to port)).

tznind:
Sorry if I was not clear.  I'm happy to make this v2 only.

You can multi select with shift+Arrow keys.  Or with ctrl+click (currently our API only reports ctrl+click on windows).

I think for keyboard multiselct toggling i will try to make this a feature of TableView

tznind:
Its hotting up! Optional 'icons' and search.

search

Still need to make it async, recursive into subfolders and cancellable.

tznind:
Recursive search is a Go Go

recursive

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+using System.Threading.Tasks;
+using NStack;
+using Terminal.Gui.Trees;
+using static System.Environment;
+using static Terminal.Gui.OpenDialog;
+
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// Modal dialog for selecting files/directories.  Has auto-complete and expandable
```
tig, line 17: Please, please, please stop putting two spaces after periods. I know it's a hard habit to break. 💕

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+	/// navigation pane (Recent, Root drives etc).
+	/// </summary>
+	public class FileDialog2 : Dialog {
```
tig, line 20: Before we merge this to v2 (I know it's still draft), we should replace the old (renaming this to `FileDialog`). Of course, this means ensuring all the existing Scenarios and Unit Tests that use the dialogs get updated/deleted/whateverneeded.

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+
+		/// <summary>
+		/// The maximum number of results that will be collected
+		/// when searching before stopping.
+		/// </summary>
+		/// <remarks>
+		/// This prevents performance issues e.g. when searching
+		/// root of file system for a common letter (e.g. 'e').
+		/// </remarks>
+		public int MaxSearchResults {get;set;} = 10000;
```
tig, line 36: Should we make this configurable? If so, simply add `[SerializableConfigurationProperty (Scope = typeof(SettingsScope))]` and put the default into `./Terminal.Gui/Resources/config.json`.

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// Modal dialog for selecting files/directories.  Has auto-complete and expandable
+	/// navigation pane (Recent, Root drives etc).
+	/// </summary>
+	public class FileDialog2 : Dialog {
+
+		// TODO : expose these somehow for localization without compromising case/switch statements
+		private const string HeaderFilename = "Filename";
```
tig, line 23: What are your feelings on using `_` to prefix private/globl vars?

I find it makes code more readable and avoids errors caused by confusing local vars with globals. It's not part of the coding guidelines for this project, but we are not consistent.

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+		private CaptionedTextField tbFind;
+		private SpinnerLabel spinnerLabel;
+
+
+		/// <summary>
+		/// Initializes a new instance of the <see cref="FileDialog2"/> class.
+		/// </summary>
+		public FileDialog2 ()
+		{
+			const int okWidth = 6;
```
tig, line 100: Why this hackiness? Is there something missing from our computed layout system/button that leads you to using these hard coded values?

tig: It's very common to have dialogs with OK/Cancel. I wonder if we should push this down into `Dialog`?

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+			this.Add (this.btnOk);
+			this.Add (this.btnCancel);
+			this.Add (this.lblUp);
+			this.Add (this.lblBack);
+			this.Add (this.lblForward);
+			this.Add (lblPath);
+			this.Add (this.tbPath);
+			this.Add (this.splitContainer);
+		}
+		/// <inheritdoc/>
```
tig, line 324: Please be consistent in your whitespace (blank lines). E.g. here you have none, a little further down you have 2.

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+		/// Gets or sets a value indicating whether to use Utc dates for date modified.
+		/// Defaults to <see langword="false"/>.
+		/// </summary>
+		public static bool UseUtcDates { get; set; } = false;
+
+		/// <summary>
+		/// Sets a <see cref="ColorScheme"/> to use for directories rows of
+		/// the <see cref="TableView"/>.
+		/// </summary>
+		public ColorScheme ColorSchemeDirectory { private get; set; }
```
tig, line 420: Shoud these all be configurable? For those that are default: In my early ConfigManager stuff, the existing examples of settings that are defaults, I went with `Default<propname>` for naming. Should we do this for these?

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+		public bool AllowsMultipleSelection {
+			get => this.tableView.MultiSelect;
+			set => this.tableView.MultiSelect = value;
+		}
+
+		/// <summary>
+		/// Gets or Sets a value indicating whether different colors
+		/// should be used for different file types/directories.
+		/// </summary>
+		public bool Monochrome { get; set; }
```
tig, line 487: Configurable?

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+			// to streamline user experience and allow direct typing of paths
+			// with zero navigation we start with focus in the text box and any
+			// default/current path fully selected and ready to be overwritten
+			this.tbPath.FocusFirst ();
+			this.tbPath.SelectAll ();
+
+			if (ustring.IsNullOrEmpty (Title)) {
+				switch (OpenMode) {
+				case OpenMode.File:
+					this.Title = $"OPEN {(MustExist ? "EXISTING " : "")}FILE";
```
tig, line 613: Localalizable?

Let's debate `ALLCAPS` vs not for dialog titles. If we agree we like `ALLCAPS` stylistically, let's build `ToUpper` into the new `View.Title`.  And make it configurable!

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+				return this.history.Back ();
+			}
+			if (keyEvent.Key == (Key.ShiftMask | Key.Backspace)) {
+				return this.history.Forward ();
+			}
+
+			return false;
+		}
+
+		private void SetupColorSchemes ()
```
tig, line 816: Move these to ConfigMan. I'll update the docs, but best practice is:

- Put the defaults into `Resources/config.json`
- Don't set default colors in code (use `-1`)
- If reasonable, add a check that throws an exception if a Color is `-1` ("Configuration was not loaded properly").

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+		}
+		private bool IsCompatibleWithOpenMode (string s, out string reason)
+		{
+			reason = null;
+			if (string.IsNullOrWhiteSpace (s)) {
+				return false;
+			}
+
+			if (!this.IsCompatibleWithAllowedExtensions (s)) {
+				reason = "Wrong file type";
```
tig, line 962: Loc.

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+				// This is the same
+				return f.Name.IndexOf (terms, StringComparison.OrdinalIgnoreCase) >= 0;
+			}
+		}
+
+		/// <summary>
+		/// Wrapper for <see cref="FileSystemInfo"/> that contains additional information
+		/// (e.g. <see cref="IsParent"/>) and helper methods.
+		/// </summary>
+		internal class FileSystemInfoStats {
```
tig, line 1295: Did you consider using extension methods instead of a wrapper? I'm not sure it's a good idea, I am just curious.

Terminal.Gui/Windows/FileDialog2.cs
```diff
@@ -0,0 +1,2118 @@
...
+					return directoryInfo.EnumerateDirectories ();
+				} catch (Exception) {
+
+					return Enumerable.Empty<DirectoryInfo> ();
+				}
+			}
+
+		}
+	}
+}
```
tig, line 2118: Dunno about you, but I really don't like giant `.cs` files. I'd much rather us adopt using `partial`.

tznind:
Thanks for the detailed review @tig I will find some time to address the points soon 👍

Terminal.Gui/Windows/FileDialog2.cs
tznind: Definetly needs to be a wrapper.  The advantages are:

- Persist stuff needed for repaint (e.g. human readable size) so it is not constantly recalculated
- Allows for special instances of the class e.g. `..` for navigating up one level

Its internal so we can always change the implementation later on if theres a problem but I feel like this is the best way to go for now.

Terminal.Gui/Windows/FileDialog2.cs
tznind: I will have a look and see if I can get rid of the const.  It is used with Anchor.AtEnd to position at lower right of window.

Could possibly be moved to `Dialog` but theres currently a lot of interaction with the button e.g. the arrow keys `NavigateIf` stuff and they are also positioned at the bottom right (i.e. not centered like would be with regular base class `Dialog`) and on the same row as other controls (Search, toggle tree and feedback).   

Ill do some coding and see where it goes to.

Terminal.Gui/Windows/FileDialog2.cs
tznind: I love this notation for private fields and have always thought it was clearer and helped avoid mistakes.  I don't know why it is not the standard.  Am happy to change to using it.

Terminal.Gui/Windows/FileDialog2.cs
tznind: Ill do a find and replace ;) its subconcious

tznind: Fixed

tznind:
How do you guys feel about dropping the `message` setting as part of the update to `FileDialog2`? Looking at the other example file dialogs (see 'Reference Examples' above) none of them have this arbitrary Label.  

I feel the Title and the text on the button should be enough for good UX.  For example the GIMP dialog has the title "Export Image" and the button title of "Export" makes it clear what the purpose is.

Reasons not to add it to `FileDialog2`:

- It takes up vertical space
- It competes with the Title for user attention
- If users provide a long string or resize the dialog its not going to wrap onto new lines correctly without a lot of extra code

image

BDisp:
No have no objection with dropping the label.

UICatalog/Scenarios/CsvEditor.cs
```diff
@@ -394,7 +394,7 @@ private void Save ()
 		private void Open ()
 		{
 			var ofd = new FileDialog () {
-				AllowedTypes = new List<AllowedType> { new AllowedType("Comma Separated Values (.csv)", ".csv") }
+				AllowedTypes = new List<AllowedType> { new AllowedType("Comma Separated Values", ".csv") }
```
BDisp: I think that is needed, at least before was, because the first parameter show the information to the user and the second parameter will provide the extension to the source. Sorry if I'm wrong.

tig:
Nuke it.

UICatalog/Scenarios/CsvEditor.cs
tznind: I will revisit this before taking this out of Draft.  

I want to look to empower `AllowedType` as well as not break the old API for OpenDialog/SaveDialog.

Ultimately user should be able to say things like:

- Only show/let pick a file called `Dockerfile`

But also I like the simplicity of the old API so that will be kept when going through OpenFile or SaveFile subclasses.

By making `AllowedType` a first class citizen then user can subclass or tailor as needed and many cool things will be possible.

Terminal.Gui/Windows/AllowedType.cs
```diff
@@ -58,12 +66,21 @@ public override string ToString ()
...
+			if (this == Any) {
+				return true;
 			}
 
+			var extension = Path.GetExtension (path);
+
+			// There is a requirement to have a particular extension and we have none
+			if (string.IsNullOrEmpty (extension)) {
+				return false;
+			}
```
BDisp, line 81: And in the case where we want to open or save without extension? What the behavior?

tznind: As a library user you can just implement `IAllowedType` interface yourself.  For example:

```csharp
public class RequireNoExtension : IAllowedType {
		public bool IsAllowed (string path)
		{
			return string.IsNullOrEmpty (Path.GetExtension (path));
		}
		{
			return "Must have no extension";
		}
}
```

BDisp: Ah ok injection here. Good work,, thanks.

tznind: I've been thinking about the 'All Types' option.  This is currently implemented via  `AllowedTypesIsStrict` but I'm debating changing that.

image

### Option 1 - AllowedTypesIsStrict
This is the way it currently is

```csharp
// Must be csv
fd.AllowedTypes = new List<IAllowedType> { new AllowedType("Delimited",".csv",".tsv") };
fd.AllowedTypesIsStrict = true;

// Can be csv or any other
fd.AllowedTypes = new List<IAllowedType> { new AllowedType ("Delimited", ".csv", ".tsv") };
```
### Option 2 - Just make the 'Any' option be another `IAllowedType` the user can add

In this case we would scrap the 'strict' flag and the user would just add the 'any' option explicitly.

```csharp
// Must be csv
fd.AllowedTypes = new List<IAllowedType> { new AllowedType ("Delimited", ".csv", ".tsv") };

// Can be csv or any
fd.AllowedTypes = new List<IAllowedType> { 
    new AllowedType ("Delimited", ".csv", ".tsv") ,
    new AllowedTypeAny()
};
```

BDisp: The option 2, no doubt. Why the user is bother creating additional extension and only going to use the first if `AllowedTypesIsStrict` is true.

tznind:
Got a wierd bug on this branch when I launch a modal dialog (FileDialog) then fire another modal on top (overwrite yes.no).  When I exit the second modal dialog the border remains.  Then if you click or move keyboard or do anything that causes a repaint you get a null reference trying to access `Border` property of `ContentView`

crash-on-2dialogs

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Terminal.Gui
  StackTrace:
   at Terminal.Gui.Border.DrawTitle(View view) in D:\Repos\gui.cs\Terminal.Gui\Core\Border.cs:line 1075


>	Terminal.Gui.dll!Terminal.Gui.Border.DrawTitle(Terminal.Gui.View view) Line 1075	C#
 	Terminal.Gui.dll!Terminal.Gui.Border.DrawChildBorder(Terminal.Gui.Rect frame, bool fill) Line 805	C#
 	Terminal.Gui.dll!Terminal.Gui.Border.DrawContent(Terminal.Gui.View view, bool fill) Line 588	C#
 	Terminal.Gui.dll!Terminal.Gui.Window.Redraw(Terminal.Gui.Rect bounds) Line 311	C#
 	Terminal.Gui.dll!Terminal.Gui.FileDialog.Redraw(Terminal.Gui.Rect bounds) Line 661	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.RunMainLoopIteration(ref Terminal.Gui.Application.RunState state, bool wait, ref bool firstIteration) Line 1243	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.RunLoop(Terminal.Gui.Application.RunState state, bool wait) Line 1190	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.Run(Terminal.Gui.Toplevel view, System.Func<System.Exception, bool> errorHandler) Line 1400	C#
 	UICatalog.dll!UICatalog.Scenarios.FileDialogExamples.SetupHandler.AnonymousMethod__12_0() Line 156	C#
 	Terminal.Gui.dll!Terminal.Gui.Button.OnClicked() Line 239	C#
 	Terminal.Gui.dll!Terminal.Gui.Button.MouseEvent(Terminal.Gui.MouseEvent me) Line 263	C#
 	Terminal.Gui.dll!Terminal.Gui.View.OnMouseEvent(Terminal.Gui.MouseEvent mouseEvent) Line 2983	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.ProcessMouseEvent(Terminal.Gui.MouseEvent me) Line 870	C#
 	Terminal.Gui.dll!Terminal.Gui.WindowsDriver.ProcessInput(Terminal.Gui.WindowsConsole.InputRecord inputEvent) Line 880	C#
 	Terminal.Gui.dll!Terminal.Gui.WindowsDriver.PrepareToRun.AnonymousMethod__34_0(Terminal.Gui.WindowsConsole.InputRecord e) Line 744	C#
 	Terminal.Gui.dll!Terminal.Gui.WindowsMainLoop.Terminal.Gui.IMainLoopDriver.MainIteration() Line 1989	C#
 	Terminal.Gui.dll!Terminal.Gui.MainLoop.MainIteration() Line 308	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.RunMainLoopIteration(ref Terminal.Gui.Application.RunState state, bool wait, ref bool firstIteration) Line 1210	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.RunLoop(Terminal.Gui.Application.RunState state, bool wait) Line 1190	C#
 	Terminal.Gui.dll!Terminal.Gui.Application.Run(Terminal.Gui.Toplevel view, System.Func<System.Exception, bool> errorHandler) Line 1400	C#
 	UICatalog.dll!UICatalog.Scenario.Run() Line 204	C#
 	UICatalog.dll!UICatalog.UICatalogApp.Main(string[] args) Line 124	C#

```

tznind:
I've marked this ready for review.  There are still some tasks in the OP that are unticked but they are polish.  I can address them when I tackle the review (I'm sure there will be changes during review!).

To assist in review here is a class diagram (I have added to repo but can remove if we don't like those):


image

tig (commented):
Starting review now...

Aesthetically, I'm a fan of:

- Excellent use of screen real estate. Tight.
- TItle bar (but some details to be debated)

Not a fan of:

- Colors - doesn't seem to honor colorscheme, and begs for https://github.com/gui-cs/Terminal.Gui/issues/2381
 
image

- ">" Glyph to left of path edit seems superfluous.
- ▲ ◄ are hard to read and when ▲ ◄► (at least with my font) don't look great
image
  I wonder if we can come up with something better?
- `[ << ]` a button to hide/show tree seems crude. I wonder if the `◊` we use in the splitter line could be replaced with a `►` or something else if the splitter is all the way to the left indicating ctrl-f10 is available? Clicking on it with mouse would expand to default size?

Bugs?

- I had one case where filenames disappeared (see screenshot above). Not sure what I did to make them disappear or come back
- What is default sort? I have to click on header to have the "sorted by" symbol to appear.
- "Multiple" doesn't seem to work?
- Turning on "Directory" OpenMode and selecting one caused an exception

image

- I don't see what "Icons" does.

tig:
> image

How'd you generate this?

Terminal.Gui/Windows/DefaultFileOperations.cs
```diff
@@ -0,0 +1,144 @@
...
+                    return true;
+				}
+			} catch (Exception ex) {
+				MessageBox.ErrorQuery (Strings.fdDeleteFailedTitle, ex.Message, "Ok");
+			}
+
+            return false;
+		}
+
+        private bool Prompt(string title,string defaultText, out string result)
```
tig, line 45: Why not use MessageBox here?

tznind: MessageBox doesn't support text input does it?

image

tznind:
> 
> How'd you generate this?

Its a ClassDiagram.  You need the Visual Studio 'Class Designer component' (its in Individual Components in the Visual Studio Installer).  Then its one of the things you can add with the New menu.

Here is an install guide: https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/docs/ide/class-designer/how-to-add-class-diagrams-to-projects.md

I then just screenshotted it for the image.

Terminal.Gui/Windows/FileDialogStyle.cs
```diff
@@ -0,0 +1,216 @@
...
+		/// <see cref="Environment.SpecialFolder"/>.
+		/// </summary>
+		/// <remarks>Must be configured before showing the dialog.</remarks>
+		public FileDialogTreeRootGetter TreeRootGetter { get; set; } = DefaultTreeRootGetter;
+
+
+		public FileDialogStyle ()
+		{
+
+			ColorSchemeDirectory = new ColorScheme {
```
tig, line 148: We gotta make this configurable/pull from OS (#2381)!

tig (changes requested):
I now think of this PR as more of a "Terminal.Gui File Explorer class that exposes FileDialog as one use-case". 

IOW, most of this class hierarchy can be broken out into a new namespace. Instead of putting it under `Window`, put it at the root as `FileServices` (or something). 

What do you think of that?

Regardless, the `File System Explorer` scenario (which is really a treeview demo) could be refactored to use this.

tznind:
> 



> Starting review now...
> 
> Aesthetically, I'm a fan of:
> 
>     * Excellent use of screen real estate. Tight.
> 
>     * TItle bar (but some details to be debated)
> 
> 
> Not a fan of:
> 
>     * Colors - doesn't seem to honor colorscheme, and begs for #2381

Can revisit this.  The idea was to be the exact colors the user has configured for the `ls` command as they would be running `ls` on command line on linux.   But am open to ideas on this feature.  I did find it quite distracting so disabled it myself by default.

It should respect the dialog color scheme I think though you are right.

> 
> image
> 
>     * ">" Glyph to left of path edit seems superfluous.

Agreed!

>     * ▲ ◄ are hard to read and when ▲ ◄► (at least with my font) don't look great
> 

I agree these are a bit sucky atm. 

> image
> I wonder if we can come up with something better? - `[ << ]` a button to hide/show tree seems crude. I wonder if the `◊` we use in the splitter line could be replaced with a `►` or something else if the splitter is all the way to the left indicating ctrl-f10 is available? Clicking on it with mouse would expand to default size?
> 
> Bugs?
> 
>     * I had one case where filenames disappeared (see screenshot above). Not sure what I did to make them disappear or come back

So file name is a column in the table, you can scroll right to see other columns like file size and Modified etc.  If there is not enough space the Filename column will be pushed off the screen.  We can add some kind of indicator for this so that it is clearer how to get it back.

>     * What is default sort? I have to click on header to have the "sorted by" symbol to appear.

I'll look into this, probably its Filename or whatever order the enumerate file system returns objects.  I can ensure that this is appropriately rendered. 
 
>     * "Multiple" doesn't seem to work?

For multi select you must use the table view and either select with Shift+Up/Down for a contiguous block.  Or use `Space` to toggle the selection.  Ctrl+Click should work too as should Shift+Click but I think the windows driver does not report these correctly.  Either way we might find a way to hint to the user that this is how to multi select.

multiselect


>     * Turning on "Directory" OpenMode and selecting one caused an exception

Should now be fixed

>     * I don't see what "Icons" does.

Icons lets you put a string in front of entries.  In the Scenario on Windows it adds a `\`  to folders but on Linux it adds little folder and file unicode characters.  We can adjust exact implementation, I did this to show off the cool unicode characters but realized that most Windows users wouldn't have the font installed.

Terminal.Gui/Windows/DefaultFileOperations.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            result = tf.Text?.ToString();
+
+            return confirm;
+        }
+
+		/// <inheritdoc/>
+		public FileSystemInfo Rename (FileSystemInfo toRename)
+		{
+            // Dont allow renaming C: or D: or / (on linux) etc
```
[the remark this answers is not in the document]
tznind: Yup, I do half my coding on Linux laptop so don't have access to that command.  Will go through the files in the diff and run it them all before merge :)  

I'll add a todo for this and do it at the end.

Terminal.Gui/Windows/DefaultFileOperations.cs
```diff
@@ -0,0 +1,144 @@
...
+                                return newLocation;
+
+                            } else {
+                                var d = ((DirectoryInfo)toRename);
+
+                                var newLocation = new DirectoryInfo(Path.Combine(d.Parent.FullName,newName));
+                                d.MoveTo(newLocation.FullName);
+                                return newLocation;
+                            }
+                    } catch (Exception ex) {
```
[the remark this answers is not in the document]
tznind: The following Exceptions are possible here (Also `ArgumentNullException` and `ArgumentException` but those should be prevented by `IsNullOrWhiteSpace` check).  I generally just use `Exception` for file IO since  there are so many different kinds of excceptions.

[IOException](https://learn.microsoft.com/en-us/dotnet/api/system.io.ioexception?view=net-7.0)

An attempt was made to move a directory to a different volume.

-or-

destDirName already exists.

-or-

You are not authorized to access this path.

-or-

The directory being moved and the destination directory have the same name.
[SecurityException](https://learn.microsoft.com/en-us/dotnet/api/system.security.securityexception?view=net-7.0)

The caller does not have the required permission.
[DirectoryNotFoundException](https://learn.microsoft.com/en-us/dotnet/api/system.io.directorynotfoundexception?view=net-7.0)

The destination directory cannot be found.

Terminal.Gui/Windows/DefaultFileOperations.cs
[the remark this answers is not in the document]
tznind: Ah this one is particularly bad, have fixed it

Terminal.Gui/Windows/DefaultFileOperations.cs
```diff
@@ -0,0 +1,144 @@
...
+		{
+            if(Prompt(Strings.fdNewTitle,"",out var named))
+            {
+                if(!string.IsNullOrWhiteSpace(named))
+                {
+                    try {
+                        var newDir = new DirectoryInfo(Path.Combine(inDirectory.FullName,named));
+                        newDir.Create();
+                        return newDir;
+                    } catch (Exception ex) {
```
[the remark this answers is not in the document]
tznind: Done

tznind:
To elaborate on icons.  This is what it looks like with icons (see below).  Files have the 3 horizontal bars ( ☰ ).  Folders have `\ua909`  ( ꤉ ).  There is a better folder unicode symbol ( 🗀 ) but it is not supported in any of the terminals I tried.

shot-2023-03-24_21-44-38

I think API wise the goal is for the default behavior to look decent and work on as many terminals as possible but also support customization.

It would be great if there was a way to tell what fonts are available but I don't think thats possible.

tznind:
Have updated to use vanilla out of the box coloring of Dialog and made other updates to style/layout:

shot-2023-04-02_20-24-23

tznind:
I have enabled `ShowHorizontalScrollIndicators` and horizontal lines dividing columns.  I think this should help prevent 'missing column'  syndrome (where you loose the Filename column and can't get it back).

I've also updated the up, back and forward buttons to be full `Button` classes (you can now tab to them and they are easier to click on and understand).

There is now a default sort order of Filename.

Sort order now always sorts as follows:

- `..` first
- Directories first (regardless of asc/desc)
- Then by sort column (asc/desc)

shot-2023-04-03_19-37-56
_Running FileDialog in a small resolution terminal having scrolled horizontally in table to see type/modified_

tznind:
Looks like TileView visibility toggling is misbehaving since [f61f31b](https://github.com/gui-cs/Terminal.Gui/pull/2259/commits/f61f31b494431d339e0a08bf2691370ab9233332)

tile-view-visibility

I have removed the custom title drawing code because it was rendering at the wrong location after  [f61f31b](https://github.com/gui-cs/Terminal.Gui/pull/2259/commits/f61f31b494431d339e0a08bf2691370ab9233332).  Not a big loss, its probably best that it uses the same look and feel as the other dialogs anyway, even if it is less pretty now  (notice how it is now left aligned).

I will try to find some time to dig into why tree view isn't popping out like it should anymore.

@tig how do you feel about us taking dependency on [System.IO.Abstractions](https://www.nuget.org/packages/TestableIO.System.IO.Abstractions)? I'm in two minds about it.  

On one hand it gives us the ability to easily write a lot of file based tests without having to write to the real file system (or worry about linux/windows incompatibility of tests). 

But on the other hand it's an extra dependency that doesn't add value to our API users.

The way it would manifest would be something like a `FileDialog()` constructor and an overload `FileDialog(IFileSystem)`.  Likewise all the sub component classes would be more easily testable.

tig:
> @tig how do you feel about us taking dependency on [System.IO.Abstractions](https://www.nuget.org/packages/TestableIO.System.IO.Abstractions)? I'm in two minds about it.
> 
> On one hand it gives us the ability to easily write a lot of file based tests without having to write to the real file system (or worry about linux/windows incompatibility of tests).
> 
> But on the other hand it's an extra dependency that doesn't add value to our API users.
> 
> The way it would manifest would be something like a `FileDialog()` constructor and an overload `FileDialog(IFileSystem)`. Likewise all the sub component classes in the `FileServices` folder would be more easily testable.

I think I'm a fan of the idea. It seems like a well maintained and used project. I'm all for better testability. I think we could learn a lot from them too.

tznind:
I'm loving System.IO.Abstractions.  How awesome is this unit test:

```csharp
[Fact, AutoInitShutdown]
public void DoStuff()
{
	// Arrange
	var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
	{
		{ @"c:\myfile.txt", new MockFileData("Testing is meh.") },
		{ @"c:\demo\jQuery.js", new MockFileData("some js") },
		{ @"c:\demo\image.gif", new MockFileData(new byte[] { 0x12, 0x34, 0x56, 0xd2 }) }
	});

	var fd = new FileDialog(fileSystem){
		Height = 15
	};
	fd.Path = @"c:\demo\";
	Begin(fd);

	fd.Redraw(fd.Bounds);

	string expected = 
	@"
┌┤OPEN├────────────────────────────────────────────────────────────┐
│c:\demo\                                                          │
│                                                                  │
│┌─────────────────┬───────────┬──────────────────────────────────┐│
││Filename (▲)     │Size       │Modified                          ││
│├─────────────────┼───────────┼──────────────────────────────────►│
││\temp            │           │                                  ││
││c:\demo\image.gif│4.00 bytes │31/12/1600 23:59:00               ││
││c:\demo\jQuery.js│7.00 bytes │31/12/1600 23:59:00               ││
││c:\myfile.txt    │15.00 bytes│31/12/1600 23:59:00               ││
│                                                                  │
│                                                                  │
│                                                                  │
│[ ►► ] Enter Search                            [ Cancel ] [ Ok ]  │
└──────────────────────────────────────────────────────────────────┘
";
	TestHelpers.AssertDriverContentsAre (expected, output, true);
}
```

tznind:
Ok sadly it seems that System.Abstractions.IO is still tied to the host OS for path interpretation.  See https://github.com/TestableIO/System.IO.Abstractions/issues/800 

The proposed fix is to Mock `IPath` which sounds complicated and difficult.  And probably actually ends up masking bugs.

So for now I have a basic linux test passing (see below) and its skipped if host OS is not linux.  I started on a windows test too but that's a WIP.

Its still way better than no coverage but will need to have a think how to handle this.

```csharp
[Fact, AutoInitShutdown]
public void TestDirectoryContents_Linux()
{
	if(System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows))
	{
		// Cannot run test except on linux :( 
		// See: https://github.com/TestableIO/System.IO.Abstractions/issues/800
		return;
	}

	// Arrange
	var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>(),"/");
	fileSystem.MockTime(()=>new DateTime(2010,01,01,11,12,43));

	fileSystem.AddFile( @"/myfile.txt", new MockFileData("Testing is meh."){LastWriteTime = new DateTime(2001,01,01,11,12,11)});
	fileSystem.AddFile( @"/demo/jQuery.js", new MockFileData("some js"){LastWriteTime = new DateTime(2001,01,01,11,44,42)});
	fileSystem.AddFile( @"/demo/image.gif", new MockFileData(new byte[] { 0x12, 0x34, 0x56, 0xd2 }){LastWriteTime = new DateTime(2002,01,01,22,42,10)});
	
	var m = (MockDirectoryInfo)fileSystem.DirectoryInfo.New(@"/demo/subfolder");
	m.Create();
	m.LastWriteTime = new DateTime(2002,01,01,22,42,10);
	
	fileSystem.AddFile( @"/demo/subfolder/image2.gif", new MockFileData(new byte[] { 0x12, 0x34, 0x56, 0xd2 }){LastWriteTime = new DateTime(2002,01,01,22,42,10)});

	var fd = new FileDialog(fileSystem){
		Height = 15
	};
	fd.Path = @"/demo/";
	Begin(fd);

	fd.Redraw(fd.Bounds);

	fd.Style.DateFormat = "yyyy-MM-dd hh:mm:ss";

	string expected = 
	@"
┌┤OPEN├────────────────────────────────────────────────────────────┐
│/demo/                                                            │
│[▲]                                                               │
│┌────────────┬──────────┬──────────────────────────────┬─────────┐│
││Filename (▲)│Size      │Modified                      │Type     ││
│├────────────┼──────────┼──────────────────────────────┼─────────┤│
││..          │          │                              │dir      ││
││\subfolder  │          │2002-01-01T22:42:10           │dir      ││
││image.gif   │4.00 bytes│2002-01-01T22:42:10           │.gif     ││
││jQuery.js   │7.00 bytes│2001-01-01T11:44:42           │.js      ││
│                                                                  │
│                                                                  │
│                                                                  │
│[ ►► ] Enter Search                            [ Cancel ] [ Ok ]  │
└──────────────────────────────────────────────────────────────────┘
";
	TestHelpers.AssertDriverContentsAre (expected, output, true);
}
```

tznind:
Putting this back in ready for review.  I think I have addressed most of the feedback.  I still need to look more into configuration manager support and now that we have System.IO.Abstractions I can write some more tests but I think all the functionality is there otherwise.

tznind:
Tickets that would be nice to have to refactor this include:

- https://github.com/gui-cs/Terminal.Gui/issues/2443
- https://github.com/gui-cs/Terminal.Gui/issues/2505

tznind:
Bugs

- Sometimes the Back arrow will take you from "c:\somedir\" back to "c:\somedir".  These duplicates should not happen
- Tab Autocomplete has a case sensitivity issue where if you have a file system:

```csharp
/folder/item1
/folder/item2
```

And you enter a capital I (i.e. `"/folder/I"`) then it will correctly propose "item1" and tab completion will work BUT if youpress Down arrow the focus will move to table view (i.e. it will not cycle suggestions).

tig (approved):

Terminal.Gui/FileServices/DefaultFileOperations.cs
```diff
@@ -0,0 +1,139 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Abstractions;
+using System.Linq;
+using Terminal.Gui.Resources;
+
+namespace Terminal.Gui.FileServices {
```
tig, line 8: I know you said you'd put this in a different namespace, but in #2471 I (rightly or wrongly) made the decision to get rid of all sub-namespaces for the public API. My feeling was it just added more complexity with no real value.

I probably should have asked for debate, but I was trying to move quickly and didn't. 

So, do you think there's a big benefit to having `FileServices` be in a distinct namespace?

Terminal.Gui/Views/FileDialog.cs
```diff
@@ -1,1053 +1,1634 @@
...
-			var msgLines = TextFormatter.MaxLines (message, Driver.Cols - 20);
+			base.Redraw (bounds);
 
-			this.nameDirLabel = new Label (nameDirLabel.IsEmpty ? $"{Strings.fdDirectory}: " : $"{nameDirLabel}: ") {
-				X = 1,
-				Y = 1 + msgLines,
-				AutoSize = true
-			};
+			if (!string.IsNullOrWhiteSpace (feedback)) {
+				var feedbackWidth = feedback.Sum (c => Rune.ColumnWidth (c));
```
tig, line 625: Can you use `View.Margin` for this instead?

Terminal.Gui/Views/FileDialog.cs
```diff
@@ -1,1053 +1,1634 @@
...
+			this.tbPath.SelectAll ();
 
-			KeyPress += (s, e) => {
-				if (e.KeyEvent.Key == Key.Esc) {
-					Cancel ();
-					e.Handled = true;
+			if (ustring.IsNullOrEmpty (Title)) {
+				switch (OpenMode) {
+				case OpenMode.File:
+					this.Title = $"OPEN {(MustExist ? "EXISTING " : "")}FILE";
```
tig, line 720: Eventually these need to be localizable.

Terminal.Gui/Views/FileDialog.cs
```diff
@@ -1,1053 +1,1634 @@
...
+
+			try {
+				this.pushingState = true;
+
+				// push the old state to history
+				if (addCurrentStateToHistory) {
+					this.history.Push (this.State, clearForward);
+				}
+
+				this.tbPath.Autocomplete.ClearSuggestions ();
```
tig, line 1159: Does stuff like this really belong in `Redraw`? It's a yellow-flag to me when I see `Redraw` implementations changing state. Is there another event (`LayoutComplete`?) where this stuff can be done?

I now believe a litmus test for good `View` design (and a test of how good our public API is) is for `View` overrides to not have to override `Redraw`. Took me a long time to get to this point ;-)

UnitTests/FileDialogTests.cs
```diff
@@ -0,0 +1,319 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Abstractions.TestingHelpers;
+using System.Linq;
+using System.Reflection;
+using Terminal.Gui.FileServices;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Terminal.Gui.Core {
```
tig, line 11: Terminal.Gui.Core -> Terminal.Gui.FileServicesTests

UnitTests/FileDialogTests.cs
```diff
@@ -0,0 +1,319 @@
+using System;
```
tig, line 1: Put all of this in `./UnitTests/FileServicesTests/`?

UnitTests/FileDialogTests.cs
```diff
@@ -0,0 +1,319 @@
...
+			fd.Path = @"/demo/";
+			Begin(fd);
+
+			fd.Redraw(fd.Bounds);
+
+			fd.Style.DateFormat = "yyyy-MM-dd hh:mm:ss";
+
+			string expected = 
+			@"
+ ┌┤OPEN├────────────────────────────────────────────────────────────┐
```
tig, line 168: Remove setting of `Title` from any test where you're not explicitly testing Title functionality. We're going to evolve the look and feel of how Title is drawn and this way we won't have to mess with so many unit tests.

tig (changes requested):
This is simply fantastic. I know you're now remembering that the last 10% is the hardest 90% ;-).

Feel free to punt all of my change requests to new Issues we can fix at our leisure except the namespace/file location ones.

Terminal.Gui/Views/FileDialog.cs
tznind: Sure, is there docs on this?  I tried

```csharp
lblFeedback = new Label ("Test") {
		Y = Pos.AnchorEnd(1),
		Width = Dim.Fill()
};
Margin.Thickness = new Thickness(0,0,0,1);
Margin.Add(lblFeedback);
```

Terminal.Gui/Views/FileDialog.cs
tznind: This code is in `PushState` method not `Redraw`.  Redraw currently only has the Feedback drawing and that will be moved to Margin 🚀

Terminal.Gui/FileServices/DefaultFileOperations.cs
tznind: I agree with your change.  I think the less namespaces the better.  I will move everything into the root namespace and rename anything that has ambigious name.  Good naming ensures things cluster anyway even without another namespace e.g. starting all classes with `File...`

UnitTests/FileDialogTests.cs
tznind: Done I've put it in `./UnitTests/FileServices` (no Tests suffix) as the other folders don't end with `Tests`.  But I have set the namespace to `Terminal.Gui.FileServicesTests` to be consistent with the rest of the test project:

image

Terminal.Gui/FileServices/DefaultFileOperations.cs
tznind: Fixed in [380e901](https://github.com/gui-cs/Terminal.Gui/pull/2259/commits/380e901e255a660e394e6e83f3101d9954decb8a)

tznind:
I've updated the class diagram to show all the new classes involved in the dialog:

image

UnitTests/FileDialogTests.cs
tznind: Fixed in [0189a38](https://github.com/gui-cs/Terminal.Gui/pull/2259/commits/0189a3836d6ff1d8d4c7177177665981116a6210)

Terminal.Gui/Views/FileDialog.cs
tznind: Actually @tig I'm not sure I understand the change you are looking for here with Margin.  This is the current user interaction.  

- Error messages appear in center of Dialog
- Any keypress in the Dialog clears the message

If it were in Margin wouldn't it have to be at the top/bottom? Please are you able to provide some pseudocode for what you were thinking of?

I'm open to suggestions on how to improve UX.  I only did it this way to avoid throwing up a modal popup over another popup (e.g. an error popup).  But didn't want to give the user no feedback at all that they were trying to take an action that was being prevented.

file-must-exist

Terminal.Gui/Views/FileDialog.cs
tznind: Fixed in [428c2ac](https://github.com/gui-cs/Terminal.Gui/pull/2259/commits/428c2acb174f9697ae0c3889ab389bce1f9adef0)

Terminal.Gui/Views/FileDialog.cs
tig: > Sure, is there docs on this? My first instinct I tried was variations on:
> 
> ```cs
> lblFeedback = new Label ("Test") {
> 		Y = Pos.AnchorEnd(1),
> 		Width = Dim.Fill()
> };
> Margin.Thickness = new Thickness(0,0,0,1);
> Margin.Add(lblFeedback);
> ```

Adding views to `Frame`s is not yet supported. 

What I was asking about is the logic you have for calculating spacing between the subviews.  

Instead of all of that, it might be possible to just set their `Margin.Thickness` to values that get you the spacing you want.

tig: This is no big deal; and I think I may have misread your code.

tig (approved):

Likarion:
Lets's make FileDialog more customizible for buttons. I suggest adding to the constructor the label for Cancel button and the ability to change the order of the buttons ([ Cancel ][ Open ] -> [ Open ][ Cancel ] -> [ Open ][ Done ]):
`..., ustring prompt, ustring cancel, bool buttonsReverseOrder,...'

tig:
I'm a fan of making it more customizable but not via a constructor. Constructors should be parameter-less.

tznind:
Most of what you describe can be accomplished with the Style class.  But there are improvements that could be made.  I've put what works at the top of this code sample and what we could enable in the bottom (in comments).

```csharp
fd.Style.OkButtonText = "Lets go!";
fd.Style.ModifiedColumnName = "Date";
fd.Title = "Blerg";

// does not exist
// fd.Style.OkCancelText = "Wait no!";

// currently has internal setters (so you cannot do this).  Also needs slight adjustment
// in order of initialization so user changes would take effect
//fd.Style.PathCaption = "Pick a file";
//fd.Style.SearchCaption = "Find yer stuff";
```
image

PR #2264, opened 2023-01-01
Fixes #2261 - Drivers clear the scrollback buffer on resize and exit.

### LINKED ISSUE #2261: Drivers use incorrect ESC codes and clear the scrollback buffer on exit
**Disappearing Scroll bar and scrolling**
I have found out that running Terminal.Gui in PowerShell or Command Prompt disables the scrolling. After one exits the Terminal.Gui window, the scrolling feature gets disabled including the scroll bar.

**To Reproduce**

1. Set up the quick start Terminal.Gui template, don't run the project yet.
https://github.com/gui-cs/Terminal.Gui/#quick-start

2. Use ``dir C:\Windows`` to generate significant output on the terminal, one should be able to scroll across multiple pages.

3. Run the template using ``dotnet run``, the Terminal.Gui window should be displayed. It can be now closed by ``Ctrl+Q``.

4. Upon quitting, the default terminal should be displayed. Use ``dir C:\Windows``  to generate more text, however, this time it is not possible to scroll across several pages of the text. The scroll bar has also vanished now.

**Expected behavior**
It should be able to preserve history and support scrolling text after the Terminal.Gui has exited.

### LINKED ISSUE #2303: HeightAsBuffer is confusingly named
See the comments in PR #2264 

This will be fixed by #2264 (current proposal is `EnableConsoleScrolling`.

tig:
Fixes #2261 
Fixes #2303

It turns out that all drivers were behaving poorly in this regard. 

- WindowsDriver: was emitting `ESC [3J` on both Resize and End. This always clears the scrollback buffer (even if the alternative screen buffer is activated).
- NetDriver: was emitting `ESC [3J`.
- CursesDriver; was emitting `ESC [3J` instead of calling `Curses.refresh()` in resize.

We were using `ESC [ ? 1049 h/l` to activate the alternative buffer in both WindowsDriver and NetDriver. Changed this to `ESC [ ? 1047 h/l` because it is less destructive (doesn't mess with the cursor) and simply works better on Windows Terminal.

Turns out `Console.Out.Flush()` is not needed because the underlying `Console.Out` stream has `AutoFlush` enabled. Removed all calls to simplify code.

Removed old comments and ensured formatting was correct (spaces->tabs) in all driver files.


BDisp:
@tig this only solves when you exit the app, but if buffer as height is false, when you runni

tig:
> @tig this only solves when you exit the app, but if buffer as height is false, when you running the app you can scroll and you shouldn't. Here you needed the 3j. I'll work on it today.

I don't understand this at all. I don't see the app scrolling. This code is only emitted when the app exits, so it should have no impact while the app is running.

BDisp:
> I don't understand this at all. I don't see the app scrolling. This code is only emitted when the app exits, so it should have no impact while the app is running.

I'm talking about this:
scroll-buffer


I think you would want like this:
scroll-buffer-fix

This isn't working well with `Windows Terminal` due to this issue https://github.com/microsoft/terminal/issues/14568.

tig:
Got it. I now understand. Thank you.

For now, until [#5094](https://github.com/microsoft/terminal/issues/5094)  is fixed by the WT team, I think the behavior shown in your first video is preferrable to clearing the buffer and causing what I (and @kingzawar, apparently) think data loss. 

Make sure y'all thumbs-up [#5094](https://github.com/microsoft/terminal/issues/5094) so the Terminal team takes it seriously.

BDisp:
> Got it. I now understand. Thank you.
> 
> For now, until [#5094](https://github.com/microsoft/terminal/issues/5094) is fixed by the WT team, I think the behavior shown in your first video is preferrable to clearing the buffer and causing what I (and @kingzawar, apparently) think data loss.
> 
> Make sure y'all thumbs-up [#5094](https://github.com/microsoft/terminal/issues/5094) so the Terminal team takes it seriously.

The `CSI 8;n;mt` for resize is good when we want to have the buffer size equal to window size. But if you want to have a buffer size greater than the window size, then we can't use it because it'll always set the window size. I couldn't find a escape sequence that set the terminal buffer size.

tig:
> > Got it. I now understand. Thank you.
> > For now, until [#5094](https://github.com/microsoft/terminal/issues/5094) is fixed by the WT team, I think the behavior shown in your first video is preferrable to clearing the buffer and causing what I (and @kingzawar, apparently) think data loss.
> > Make sure y'all thumbs-up [#5094](https://github.com/microsoft/terminal/issues/5094) so the Terminal team takes it seriously.
> 
> The `CSI 8;n;mt` for resize is good when we want to have the buffer size equal to window size. But if you want to have a buffer size greater than the window size, then we can't use it because it'll always set the window size. I couldn't find a escape sequence that set the terminal buffer size.

I think I understand you.

However, what I don't know is how often people want a buffer size that's different than the window size. I suspect it's a very rare use case?

BDisp:
> I think I understand you.
> 
> However, what I don't know is how often people want a buffer size that's different than the window size. I suspect it's a very rare use case?

Not exactly. See the example of the second video where I enabled the `BufferAsHeight` and reduced the window size but I still can get see all the rows by using the vertical scroll.

tig:
> > I think I understand you.
> > However, what I don't know is how often people want a buffer size that's different than the window size. I suspect it's a very rare use case?
> 
> Not exactly. See the example of the second video where I enabled the `BufferAsHeight` and reduced the window size but I still can see all the rows by using the vertical scroll.

But WHY would someone want to enable `HeightAsBuffer`? The API docs are confusing to me. 

I'm sorry if I am being persistent on this, but I sincerely do not understand (and never have) the use case. This is likely just me being slow.

BDisp:
> But WHY would someone want to enable `HeightAsBuffer`? The API docs are confusing to me.
> 
> I'm sorry if I am being persistent on this, but I sincerely do not understand (and never have) the use case. This is likely just me being slow.

It's only an option. Why the author opened issue? Because it was clearing the historic after exit the app. But, why he needs to scroll to see the historic? Well, before I fixed the window resize on the `WindowsDriver`, it only used the buffer size.
I was preparing a fix for this that was by using the escape sequence `CSI ?1049h` on the `Init` method to enable alternative screen buffer and `CSI ?1049l` on the `End` method to disable alternative screen buffer. In practice it save the original screen buffer before start the app and restore the original screen buffer after exit the app. Do you still want I open a PR to this branch or it not worth?

tig:
If one were to re-write the API docs for `HeightAsBuffer` to accurately reflect what it does what would they say?

Application.HeightAsBuffer currently reads as 
```cs
		/// <summary>
		/// The current <see cref="ConsoleDriver.HeightAsBuffer"/> used in the terminal.
		/// </summary>
		public static bool HeightAsBuffer {
```

ConsoleDriver.HeightAsBuffer currently reads as:
```cs
		/// <summary>
		/// If false height is measured by the window height and thus no scrolling.
		/// If true then height is measured by the buffer height, enabling scrolling.
		/// </summary>
		public abstract bool HeightAsBuffer { get; set; }
```

"If set to false (the default), the height of the Terminal.Gui application's top-most, TopLevel will be resized to match the visible console size if the console is resized and the console will not scroll. If set to true, the height of the Terminal.Gui application's top-most, TopLevel, will not resize when the console is resized and the console can be scrolled."

Is that correct? If not, please re-write it so that it is correct.

BDisp:
/// <summary>
		/// If false height is measured by the window height and thus no scrolling.
		/// If true then height is measured by the buffer height, enabling scrolling.
		/// </summary>
		public abstract bool HeightAsBuffer { get; set; }


This definition is perfectly wright. Should be copied to the `Application.HeightAsBuffer` too?

tig:
```cs
		/// <summary>
		/// If false height is measured by the window height and thus no scrolling.
		/// If true then height is measured by the buffer height, enabling scrolling.
		/// </summary>
		public abstract bool HeightAsBuffer { get; set; }

```

This definition is perfectly right. Should be copied to the `Application.HeightAsBuffer` too?

It may be perfectly right, but I find it VERY confusing and unclear. 

- It is not clear what 'height' refers to. Is it the height of the ConsoleDriver `Contents` (e.g. `Rows`)? Or is it the height of `Application.Top`?
- It is not clear what `window` refers to. Is it the height of `Application.Top` or is it the height of the visible console view (or both)?
- It is not clear what `buffer` is. Is it the console's scroll-back buffer? Or is it the buffer for the visible console view (e.g. for `WindowsConsole`, the buffer managed by `SetConsoleOutputWindow`? Or something else? 
- It is not clear that scrolling refers to scrolling the console (and not within the app).
- It is not clear what scrolling has to do with any of this. 

Would it have made more sense to name this property "EnableConsoleScrolling"?

I'm not challenging or disagreeing with you. I seriously do not understand how this is supposed to work.

BDisp:
> ```cs
> 		/// <summary>
> 		/// If false height is measured by the window height and thus no scrolling.
> 		/// If true then height is measured by the buffer height, enabling scrolling.
> 		/// </summary>
> 		public abstract bool HeightAsBuffer { get; set; }
> ```
> 
> This definition is perfectly right. Should be copied to the `Application.HeightAsBuffer` too?
> 
> It may be perfectly right, but I find it VERY confusing and unclear.
> 
>     * It is not clear what 'height' refers to. Is it the height of the ConsoleDriver `Contents` (e.g. `Rows`)? Or is it the height of `Application.Top`?
> 

Refers to the Rows.

>     * It is not clear what `window` refers to. Is it the height of `Application.Top` or is it the height of the visible console view (or both)?
> 

It's the visible console view.

>     * It is not clear what `buffer` is. Is it the console's scroll-back buffer? Or is it the buffer for the visible console view (e.g. for `WindowsConsole`, the buffer managed by `SetConsoleOutputWindow`? Or something else?
> 

It's the Rows length.


>     * It is not clear that scrolling refers to scrolling the console (and not within the app).
> 

Refers scrolling the console within the app.


>     * It is not clear what scrolling has to do with any of this.
> 

Scrolling the console within the app allows view all the available rows.

> 
> Would it have made more sense to name this property "EnableConsoleScrolling"?
> 

Perhaps is better. I leave that up to you.


> I'm not challenging or disagreeing with you. I seriously do not understand how this is supposed to work.

No problem. I hope you now understanding all my concern about this. Thanks.

tig:
> No problem. I hope you now understanding all my concern about this. Thanks.

I would appreciate it if you used your answers above to recraft the API documentation for `HeightAsBuffer` so that it is more clear and complete.

BDisp:
> I would appreciate it if you used your answers above to recraft the API documentation for `HeightAsBuffer` so that it is more clear and complete.

I will in both, `ConsoleDriver` and `Application`. In `WindowsDriver` it's already working as expected but I'm facing some problems with `NetDriver` and `CursesDriver`. I'll continue try to find the solution.

tig:
> > I would appreciate it if you used your answers above to recraft the API documentation for `HeightAsBuffer` so that it is more clear and complete.
> 
> I will in both, `ConsoleDriver` and `Application`. In `WindowsDriver` it's already working as expected but I'm facing some problems with `NetDriver` and `CursesDriver`. I'll continue try to find the solution.

Could you draft it here in a comment first?

BDisp:
> Could you draft it here in a comment first?

Sure and please make the changes needed.

In `ConsoleDriver.cs`:
```cs
		/// <summary>
		/// If <see langword="false"/> height sets to <see cref="Rows"/> is measured by the window height
		/// and thus no scrolling, showing all the available <see cref="Rows"/> contents.
		/// If <see langword="true"/> then height sets to <see cref="Rows"/> is measured by the buffer height,
		/// enabling scrolling within the app to allow showing all the <see cref="Rows"/> contents
		/// in the visible console window height for the current scrolling view-port.
		/// </summary>
		public abstract bool HeightAsBuffer { get; set; }

```

In `Application.cs`:
```cs
		/// <summary>
		/// The current <see cref="ConsoleDriver.HeightAsBuffer"/> used in the terminal.
		/// </summary>
		/// <remarks>
		/// If <see langword="false"/> height sets to <see cref="ConsoleDriver.Rows"/> is measured by the window height
		/// and thus no scrolling, showing all the available <see cref="ConsoleDriver.Rows"/> contents.
		/// If <see langword="true"/> then height sets to <see cref="ConsoleDriver.Rows"/> is measured by the buffer height,
		/// enabling scrolling within the app to allow showing all the <see cref="ConsoleDriver.Rows"/> contents
		/// in the visible console window height for the current scrolling view-port.
		/// </remarks>
		public static bool HeightAsBuffer {

```

tig:
I edited your post. Is it still accurate?

BDisp:
> I edited your post. Is it still accurate?

Only this is incorrect:
```cs
                /// If <see langword="true"/> then height of the Terminal.Gui application <see cref="Rows"/> does not
                /// change when the console is resized (it stays the same value it was set to at application startup). 
```

The height only changes if the current height after resizing is greater than the current `Rows`. So, it may be different than the value it was set to at application startup. Another observation is it depends of when the `HeightAsBuffer` is toggled from false to true. If `HeightAsBuffer` is false and `Rows` is equal to 10, toggling `HeightAsBuffer` to true the minimum `Rows` is 10 on window shrink. If the window is expanded to 20, then the `Rows` will be also equal to 20 and will maintain this will if window is shrink. If `HeightAsBuffer` is toggled to true `Rows` will be equal to window height again.
Please do the changes you think more appropriate.

BDisp:
@tig I sent the PR https://github.com/tig/Terminal.Gui/pull/8 which is currently working on Windows. Also the scroll bar is only hidden when the app is running and `HeightAsBuffer` is `false`, but is restored when returned to the console after exit the app. When the app is running the scroll is only showed on Windows when `HeightAsBuffer` is `true`.

tig:
Hey @BDisp, 

I'm playing with https://github.com/tig/Terminal.Gui/pull/8 merged in to this and regardless of what I set `_heightAsBuffer/_enableConsoleScrolling` in UICatalog to, I get the same behavior: The vertical scroll bar in Windows Terminal is always active, and scrolling works.

The only difference in behavior I see is on resize:

- When `false` UICatalog resizes when the console is made bigger OR smaller vertically 
xOqLiAe 1

- When `true` UI Catalog only resizes when then console is made bigger.
6gtLlYw 1

I *like* the behavior, but I think I confused you or you confused me on creating the new name for the property.

If this is the correct behavior, then the property means "Do Not Shrink Vertically On Console Resize`.

Am I now (finally) thinking of this correctly?

If so, I will update the property name and API docs in this PR to:

```cs
/// <summary>
/// Prevents the application window from shrinking vertically when the terminal is resized smaller. Tracks <see cref="ConsoleDriver.NoVerticalShrink"/>.
/// </summary>
/// <remarks>
/// <para>
/// If <see langword="false"/> (the default) the height of the Terminal.Gui application (<see cref="ConsoleDriver.Rows"/>) 
/// tracks to the height of the visible terminal view when the terminal is resized. In this case 
// all <see cref="ConsoleDriver.Rows"/> will remain visible regardless of the vertical size of the terminal.
/// </para>
/// <para>
/// If <see langword="true"/> then height of the Terminal.Gui application <see cref="ConsoleDriver.Rows"/> only tracks 
/// the visible terminal height when the terminal is made larger (the application will only grow in height, never shrink). 
/// </para>
/// This API was previously named 'HeightAsBuffer` but was renamed to clarify its purpose.
/// </remarks>
public static bool NoVerticalShrink {

```

What do you think?

BDisp:
@tig sorry I didn't test with `Windows Terminal`, but only with `cmd, conhost and powershell`. I removed the 
```cs
			Console.Out.Write ("\x1b[3J");
			Console.Out.Flush ();

```
because it wasn't needed on those consoles, as it was echoing to the console buffer after exit. But with the `Windows Terminal` is needed. But please, believe me, you must use the `"\x1b[3J"`, otherwise the scrolling of the original buffer isn't cleaned. The problem is it isn't allowing resizing the buffer size. Do you remember the issue on the `terminal` repo?
I'll submit a new PR with the fix, but remember that the scrolling will not working on the `Windows Trerminal` until they decide allowing resizing the buffer. So, I don't agree with changing to `NoVerticalShrink` because is a `Windows Trerminal` issue.

BDisp:
@tig  here is the fix https://github.com/tig/Terminal.Gui/pull/9.

BDisp:
The `NetDriver` is also having issue in `Windows Terminal` on resizing causing some lines not be drawing after exceptions. I'll commit a fix in your branch on the same PR.

Before:
net-driver-resize-issue

After:
net-driver-resize-fix

BDisp:
The `Windows Terminal` buffer resize is discussing [here](https://github.com/microsoft/terminal/issues/5094#issuecomment-1399628636).

tig:
> @tig sorry I didn't test with `Windows Terminal`, but only with `cmd, conhost and powershell`. I removed the
> 
> ```cs
> 			Console.Out.Write ("\x1b[3J");
> 			Console.Out.Flush ();
> ```
> 
> because it wasn't needed on those consoles, as it was echoing to the console buffer after exit. But with the `Windows Terminal` is needed. But please, believe me, you must use the `"\x1b[3J"`, otherwise the scrolling of the original buffer isn't cleaned. The problem is `Windows Terminal` isn't allowing resizing the buffer size. Do you remember the issue on the `terminal` repo? I'll submit a new PR with the fix, but remember that the scrolling will not working on the `Windows Trerminal` until they decide allowing resizing the buffer. So, I don't agree with changing to `NoVerticalShrink` because is a `Windows Trerminal` issue.

Because the `Console` that's being written to is now the alternate frame buffer, emitting `[3J` only clears the alternate frame buffer.

Thus the bug reported in #2261 is fixed (in Windows Terminal) by your updates; The scroll buffer is not deleted when the Terminal.Gui app exits.

However, in cmd.exe, when the Terminal.Gui app exits, garbage is printed (WindowsDriver) and the scrollback buffer is nuked:
gkm7Du2 1

With -usc in cmd.exe no garbage is printed, and the scrollback buffer seems ok, but the prompt goes in the wrong place:
U53kNEk 1

Curses under WSL in cmd.exe works as expected:
fhVnHSm 1

In ConEmu, the scrollback buffer is deleted with and without -usc.

So this bug is still not fixed for:

- cmd.exe
- ConEmu

tig:
@BDisp Why are you using `ESC [ ? 1049 h` instead of `ESC [ ? 1047 h`?

I have found Windows Terminal works as expected if `ESC [ ? 1047 h` and `ESC [ ? 1047 l` are used and `ESC [ n J` is NOT used at all!

I also removed all calls to `Console.Out.Flush()` because I couldn't see it having any impact.

N0NAqxV 1

This does not fix cmd.exe or ConEmu though.

BUT: Even with v1.6.4 the backscroll buffer doesn't work in cmd.exe and ConEmu.

BDisp:
> However, in cmd.exe, when the Terminal.Gui app exits, garbage is printed (WindowsDriver) and the scrollback buffer is nuked:

That what I referred when I said that with `cmd, conhost and powershell`, was echoing to the console buffer after exit using `WindowsDriver`. The escape sequence is only recognized on `Windows Terminal`. A workaround is a way that recognize the app is using the `Windows Terminal` and in this case apply the escape sequences. The scrollback buffer is nuked because of the expose reasons and the solution might be some other solution that not use escape sequences but only `Win32 Api`.



> With -usc in cmd.exe no garbage is printed, and the scrollback buffer seems ok, but the prompt goes in the wrong place:

The prompt goes to wrong place on exit if the app did some resizing. I think that because the virtual terminal of the Win32 console and the System.Console aren't processing well the escape sequences, like `CursesDriver` does on `WSL`.

BDisp:
> @BDisp Why are you using `ESC [ ? 1049 h` instead of `ESC [ ? 1047 h`?
> 

Because the `ESC [ ? 1049 h` saves the cursor.

> I have found Windows Terminal works as expected if `ESC [ ? 1047 h` and `ESC [ ? 1047 l` are used and `ESC [ n J` is NOT used at all!
> 
The `ESC [ n J` is only needed if `EnableConsoleScrolling` is false, to clear the buffer of the app and not the original saved .

> I also removed all calls to `Console.Out.Flush()` because I couldn't see it having any impact.

Maybe the` Console.AutoFlush` is enabled and thus the `Console.Out.Flush()` isn't needed.


@tig that isn't the behavior expected by the app. On the app we only want to scroll if `EnableConsoleScrolling` is enabled, through our app and not through the saved buffer. On your video you didn't do any resize. Try do it and you'll the garbage when scrolling while the app is running. The user who open the issue only reported that the original buffer and thus if it's greater than the windows size, the vertical scroll bar. I don't think he, and us, of course, don't want to see other buffer than the one generated by the app.

tig:
> @tig that isn't the behavior expected by the app. On the app we only want to scroll if `EnableConsoleScrolling` is enabled, through our app and not through the saved buffer. On your video you didn't do any resize. Try do it and you'll the garbage when scrolling while the app is running. The user who open the issue only reported that the original buffer disappeared and thus if it's greater than the windows size, also the vertical scroll bar. I don't think he, and us, of course, while app is running, don't want to see other buffer than the one generated by the app.

I disagree.

The way Terminal.Gui worked from v1.6.4 and earlier was as follows:

- The default behavior (HeightAsBuffer = false) was
  - the app would fill the terminal and any resizing would resize the app.
  - If the terminal had scroll bars (some do; some don't) and the user scrolled up, they could see the scrollback buffer
  - Exiting the app would PRESERVE the scrollback buffer
- If HeightAsBuffer = true 
  - the app would fill the terminal and resizing vertically would grow the app, but never shrink it
  - If the terminal had scroll bars (some do; some don't) and the user scrolled up, they could see the scrollback buffer. 
  - Scrolling after changing the size of the terminal would cause garbage
  - Exiting the app would PRESERVE the scrollback buffer (but sometimes with some garbage).

MOST IMPORTANTLY: when the app exited, in EITHER case, the scrollback buffer was NOT cleared.

I assert `HeightAsBuffer` has never really worked well (at least on Windows). 

Only a few users set `HeightAsBuffer` to true. Most do not use it.

Most of our users are on Windows, and Windows Terminal is now the default terminal for Windows. 

With my latest push, I believe things are working properly on Windows Terminal (our most important terminal on Windows) with the exception of Windows Terminal not really honoring the alternative screen buffer (and corrupting the scrollback buffer) when resizing. 

Users do not want the scrollback buffer cleared by a console app running (except the `cls` command).  They also don't want the scrollback buffer corrupted. But if they don't resize the terminal, the current code doesn't clear the scrollback buffer when the app exits.

The fact that conhost/cmd.exe doesn't handle the esc sequences and prints them is annoying. if we can find a way to detect the terminal, that would be great.

The fact that conemu acts weird is annoying, but conemu is just weird, IMO.

Can you show me an example of `EnableConsoleScrolling = true` on any platform that is working correctly?

BDisp:
@tig if you happy with this, no problem. Though `Windows Terminal` is the default terminal on `Windows`, doesn't mean that everything is well done. It's working well on `cmd, conhost and powershell`, despite the echoing `Ansi Escape Codes`, which is better than the `Windows Terminal` garbage does. But for me I'll not talk more about this,

windows-driver-garbage

tig:
@BDisp Can you show me an example of HeightAsBuffer/EnableConsoleScrolling = true on a platform other than Windows/Windows Terminal that is working as you expect?

Here's Debian/xterm/CursesDriver (using `develop`):
UYJei9G 1

It has no effect.

Debian/xterm/NetDriver behaves identically (HeightAsBuffer has no effect).

BDisp:
> @BDisp Can you show me an example of HeightAsBuffer/EnableConsoleScrolling = true on a platform other than Windows/Windows Terminal that is working as you expect?

With `cmd` has the ansi escape code echos but is less garbage than `Windows Terminal`.



> It has no effect.
> 
> Debian/xterm/NetDriver behaves identically (HeightAsBuffer has no effect).

On `WSL` the `CursesDriver` only use the `LINES` equal to windows height and the `NetDriver` also only uses the windows height. But I think there is some ansi escape code that can set the buffer height greater than the windows height.

tig:
> > @BDisp Can you show me an example of HeightAsBuffer/EnableConsoleScrolling = true on a platform other than Windows/Windows Terminal that is working as you expect?
> 
> With `cmd` has the ansi escape code echos but is less garbage than `Windows Terminal`.
> 
> windows-driver-cmd [ windows-driver-cmd ](https://user-images.githubusercontent.com/13117724/214179242-9f55a368-19cf-42f9-a11f-fe27fd446ce1.gif) [ ](https://user-images.githubusercontent.com/13117724/214179242-9f55a368-19cf-42f9-a11f-fe27fd446ce1.gif)
> 
> > It has no effect.
> > Debian/xterm/NetDriver behaves identically (HeightAsBuffer has no effect).
> 
> On `WSL` the `CursesDriver` only use the `LINES` equal to windows height and the `NetDriver` also only uses the windows height. But I think there is some ansi escape code that can set the buffer height greater than the windows height.

So, if I understand you correctly (and this is backed by my tests):

- The "HeightAsBuffer"/"EnableConsoleScrolling" feature only works on Windows.
- When used on Windows Terminal, scrolling is always enabled, the only difference is the app will grow, but never shink vertically.
- When used on cmd.exe, setting it to false enables the scroll bar.
- The setting has no effect on the CursesDriver on Windows or Linux

I've now tried 2 different Linux terminals (xterm and terminator) and I have fixed all issues having to with THIS issue (independent of HeightAsBuffer). Please review my code.

I also took the liberty of fixing the formatting in all the ConsoleDrivers files and removing extraneous comments since so many of the files were already touched.

BDisp:
> So, if I understand you correctly (and this is backed by my tests):
> 
>     * The "HeightAsBuffer"/"EnableConsoleScrolling" feature only works on Windows.
> 
For now yes. My intention is also use a workaround on Unix to add new lines after the windows bottom while the `Rows` is greater than the windows height. I also want to use ansi escape to request the windows size, but this is only working on a real `Linux` and not on Windows and `WSL`.

>     * When used on Windows Terminal, scrolling is always enabled, the only difference is the app will grow, but never shink vertically.
> 
Negative. The scrolling is only enabled if the `EnableConsoleScrolling` property is enabled, otherwise the app will never scroll. If `EnableConsoleScrolling` is disabled and the scroll bar is visible before start the `Terminal.Gui` app, the scroll bar will be disabled when the app start and the scroll bar will be restored after exit the app, making it visible if it was already visible before. If the `EnableConsoleScrolling` is enabled and the scroll bar is visible before start the `Terminal.Gui` app, the scroll bar will be disabled when the app start. If the window is shrunken the scroll bar will be visible and we can scroll between our app, in the case of `UICatalog` will be possible to scroll between the menu bar and the status bar, but never the scrollback buffer while app is running. After exit the app the scroll bar will be visible, if it was visible before, to allow scrollback buffer.

>     * When used on cmd.exe, setting it to false enables the scroll bar.
> 
False, it shouldn't. Only must enable the scroll bar if the `EnableConsoleScrolling` property is enabled. Remember that setting to true the scroll bar isn't show immediately, but when the windows is shrunken. When the window is expanded until the windows height is equal to `Rows`, the scroll bar is invisible and if is shrunken again the scroll bar is visible again.

>     * The setting has no effect on the CursesDriver on Windows or Linux
> 
For now isn't working but as I said above maybe is possible to make it work. The behavior is intended to make it work with all terminal, but sometimes there are some incompatibilities
> 
> I've now tried 2 different Linux terminals (xterm and terminator) and I have fixed all issues having to with THIS issue (independent of HeightAsBuffer). Please review my code.
> 
You have to be more specific and explain what was bad and now it's working. I know that as I said that scrolling wasn't working on `Linux` if `EnableConsoleScrolling` property is enabled.

> I also took the liberty of fixing the formatting in all the ConsoleDrivers files and removing extraneous comments since so many of the files were already touched.

I agree, it was needed but you know I'm not nothing good with docs :-)

PS: When I create the `HeightAsBuffer`, now `EnableConsoleScrolling`, was with the intentions I described above. If you want to give another behavior than that, then I disagree, but it is in your hand decide.

BDisp:
@tig but I admit another scenario for the vertical scroll bar behavior, like an `enum`:

```cs
public enum VerticalScrollBehavior {
	// No scroll at all.
	None,
	// Only allow scrolling into the app.
	AppScrolling,
	// Allow scrolling on all scrollback buffer (original buffer more app buffer)
	// This allows an app write to the console preserving with all scrollback buffer.
	ScrollBackBuffer
}
```

tig:
We need to retarget this PR at the `v2_develop` branch or delete it if it's already been merged via other PRs. I don't remember.

tig:
I'm going to get back to working on this asap. I think this may be worth taking into `develop` / v1, so I've tagged it as such.

Please debate if you don't agree.

tig:
> * WindowsDriver: was emitting `ESC [3J` on both Resize and End. This always clears the scrollback buffer (even if the alternative screen buffer is activated).
> * NetDriver: was emitting `ESC [3J`.
> * CursesDriver; was emitting `ESC [3J` instead of calling `Curses.refresh()` in resize.
> 
> We were using `ESC [ ? 1049 h/l` to activate the alternative buffer in both WindowsDriver and NetDriver. Changed this to `ESC [ ? 1047 h/l` because it is less destructive (doesn't mess with the cursor) and simply works better on Windows Terminal.

I can't remember what we finally decided to do here. 

* NetDriver is still using `[?1049h`; WindowsDriver is using 1047 and I think NetDriver should as well.
* WindowsDrivder is still emittting `[?3J` in ResizeScreen (for EnableconsoleScrolling); I think it should be using `[0J?`.

The only other thing in that was in this PR, that does not appear to be in `v2_develop` is the code added to `UpdateScreen` that prevents writing to the console if SetCursorPos failed. I don't know what that is for.

@bdiisp, do you have an opinion on what we should do with this PR?

BDisp:
If `EnableconsoleScrolling` is enabled the app can scroll through the console (note that is the app scrolling, not the content existing before the app running). The `[?3J` allows all the scroll-back is cleared before update screen the app. But, also before that was only working on `conhost` and not on `Windows Terminal` (because doesn't allows resize the console buffer by code). The` [?1049h` and `?1049l` allows save and restore the original screen buffer. Now I don't know how it's.

tig:
Dupe of #2764

PR #2266, opened 2023-01-02
Fixes #2265 - Keyboard input doesn't work in UnixMainLoop when Idle function is registered

### LINKED ISSUE #2265: Keyboard input doesn't work in UnitMainLoop when Idle function is registered
**Describe the bug**
Keyboard input doesn't work in UnitMainLoop when Idle function is registered. 

**To Reproduce**
Run this application on Linux:
```
using Terminal.Gui;

Application.Init();
Application.Top.Add(new TextField { Width = 30, Height = 1 });

// comment out this line and it works on Linux (but no idle function, of course)
Application.MainLoop.AddIdle(() => true);

Application.Run();
Application.Shutdown();
```

**Expected behavior**
Text field is editable. Ctrl+Q quits app. If you comment out the AddIdle line

- This same code runs fine on Windows 11 native
- It doesn't work on Linux. Tested on native Ubuntu 20.04 and Windows 11 WSL2 Ubuntu 20.04.

AaronLieberman:
Fixes #2265 - Keyboard input doesn't work in UnixMainLoop when Idle function is registered

When running in WSL Ubuntu, I was unable to process input. The issue was that when there's an idle function registered with the main loop, UnixMainLoop doesn't ever call poll. Without the poll call, pollmap doesn't get updated and the input callback doesn't get called. Change this so that it calls poll every loop iteration, even if there's an idle function.

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/UnixMainLoop.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/UnixMainLoop.cs
@@ -176,16 +176,15 @@ namespace Terminal.Gui {
 		{
 			UpdatePollMap ();
 
-			if (CheckTimers (wait, out var pollTimeout)) {
-				return true;
-			}
+			bool checkTimersResult = CheckTimers (wait, out var pollTimeout);
 
 			var n = poll (pollmap, (uint)pollmap.Length, pollTimeout);
 
 			if (n == KEY_RESIZE) {
 				winChanged = true;
 			}
-			return n >= KEY_RESIZE || CheckTimers (wait, out pollTimeout);
+
+			return checkTimersResult || n >= KEY_RESIZE;
 		}
 
 		bool CheckTimers (bool wait, out int pollTimeout)


tig (approved):
Lovely! Thank you for tracking this down.

@BDisp since you are familiar with the UnixMainLoop code, could you please review this PR as well? This is a sensitive/fragile area of the codebase so we should be extra careful. Once you do, and find no issues, I'll merge.

tig:
@AaronLieberman Could you add a unit test (to `MainLoopTests.cs`) that a) fails without this fix and b) succeeds with it?

tig:
This is a nasty enough bug that I think we should hold v1.9 until @BDisp can review it. If @AaronLieberman can't get a Unit Test done quickly, that's ok. We can add one via a subsequent PR for a future release.

BDisp (approved):
Thanks for this fix @AaronLieberman. Unlike `WindowsDriver` and `NetDriver` where the user input are handled by `ManualResetEventSlim` which is set when there is any keystroke, on `CursesDriver` the `poll` method is never hit if a idle is set over and over. Good work.

BDisp:
You are right. Actually we only test against `FakeDriver`. The only solution I see for doing that is by using the `testenvironments.json`, which I already implemented and make conditional unit tests that run one specific OS. But that is for doing someday if it really will be done 😃

PR #2284, opened 2023-01-09
Fixes #2181 - Adds configuration manager

### LINKED ISSUE #29: User configurable Color Themes
Should include other color themes, currently we have the Norton-esque theme, we could include a TurboVision-esque theme.   The support is already in place, it is a matter of allocating new colors in each driver.

### LINKED ISSUE #2181: Add Visual Style manager
With 

* the potential of a new Button look (#2144 / #2166) that may cause apps to break when enabled
* a proliferation of visual style settings on other Views (e.g. `TreeStyle`)
* the existing `ColorScheme` scheme and requests like #29
* and more to come...

I suggest we invest in a Visual Style manager.

Basic idea:

* Build a JSON schema for the existing visual style settings on the various Views; it's important that this supports the library as it is NOW. 
* Enhance existing View classes to have `static` "default" style properties (similar to what I did in my PR to #2166 that enabled UI Catalog to set the `Button` style globally.
* Build a component that reads a JSON doc and applies the specified styles such that all views that support it get their defaults set when the library loads.
* Build a mechanism for apps to configure per-app state
  * A `visualstyle.json` file that gets built into the app's assembly.
* Build a mechanism for storing user-defined per-app state. 
  * A `~/.tui` folder for global settings. `~/.tui/UICatalogApp.visualstyle.json` would set the visual style of any app with a main class name of `UICatalogApp`. Unless...
  * A `./.tui` folder for local settings. 
    * `./.tui/UICatalogApp.visualstyle.json` would set the visual style of any app with a main class name of `UICatalogApp` launched from anywhere under `./`. 
    * `./.tui/visualstyle.json` would set the default visual style of any app launched from anywhere under `./`.

I'm sure this has been invented before and we can plagiarize. 

Let the debate begin! :-)

tig:
Fixes #2181
Fixes #29 

Adds a Configuration Manager.

# Configuration Management

Terminal.Gui provides configuration and theme management for Terminal.Gui applications via the [`ConfigurationManager`](~/api/Terminal.Gui/Terminal.Gui.Configuration.

1) **Settings**. Settings are applied to the [`Application`](~/api/Terminal.Gui/Terminal.Gui.Application.yml) class. Settings are accessed via the `Settings` property of the [`ConfigurationManager`](~/api/Terminal.Gui/Terminal.Gui.Configuration.ConfigurationManager.yml) class.
2) **Themes**. Themes are a named collection of settings impacting how applications look. The default theme is named "Default". The built-in configuration stored within the Terminal.Gui library defines two additional themes: "Dark", and "Light". Additional themes can be defined in the configuration files.
3) **AppSettings**. AppSettings allow applicaitons to use the  [`ConfigurationManager`](~/api/Terminal.Gui/Terminal.Gui.Configuration.ConfigurationManager.yml) to store and retrieve application-specific settings.

The The [`ConfigurationManager`](~/api/Terminal.Gui/Terminal.Gui.Configuration.ConfigurationManager.yml) will look for configuration files in the `.tui` folder in the user's home directory (e.g. `C:/Users/username/.tui` or `/usr/username/.tui`), the folder where the Terminal.Gui application was launched from (e.g. `./.tui`), or as a resource within the Terminal.Gui application's main assembly.

Settings that will apply to all applications (global settings) reside in files named config.json. Settings that will apply to a specific Terminal.Gui application reside in files named appname.config.json, where appname is the assembly name of the application (e.g. `UICatalog.config.json`).

Settings are applied using the following precedence (higher precedence settings overwrite lower precedence settings):

1. App specific settings found in the users's home directory (`~/.tui/appname.config.json`). -- Highest precedence.

2. App specific settings found in the directory the app was launched from (`./.tui/appname.config.json`).

3. App settings in app resources (`Resources/config.json`).

4. Global settings found in the the user's home directory (`~/.tui/config.json`).

5. Global settings found in the directory the app was launched from (`./.tui/config.json`).

6. Default settings defined in the Terminal.Gui assembly -- Lowest precedence.

The `UI Catalog` application provides an example of how to use the [`ConfigurationManager`](~/api/Terminal.Gui/Terminal.Gui.Configuration.ConfigurationManager.yml) class to load and save configuration files. The `Configuration Editor` scenario provides an editor that allows users to edit the configuration files. UI Catalog also uses a file system watcher to detect changes to the configuration files to tell [`ConfigurationManager`](~/api/Terminal.Gui/Terminal.Gui.Configuration.ConfigurationManager.yml) to reaload them; allowing users to change settings without having to restart the application.

# What Can Be Configured

## Settings

Settings for the [`Application`](~/api/Terminal.Gui/Terminal.Gui.Application.yml) class.
    * [QuitKey](~/api/Terminal.Gui/Terminal.Gui.Application.yml#QuitKey)
    * [AlternateForwardKey](~/api/Terminal.Gui/Terminal.Gui.Application.yml#AlternateForwardKey)
    * [AlternateBackwardKey](~/api/Terminal.Gui/Terminal.Gui.Application.yml#AlternateBackwardKey)
    * [UseSystemConsole](~/api/Terminal.Gui/Terminal.Gui.Application.yml#UseSystemConsole)
    * [IsMouseDisabled](~/api/Terminal.Gui/Terminal.Gui.Application.yml#IsMouseDisabled)
    * [HeightAsBuffer](~/api/Terminal.Gui/Terminal.Gui.Application.yml#HeightAsBuffer)

## Themes

A Theme is a named collection of settings that impact the visual style of Terminal.Gui applications. The default theme is named "Default". The built-in configuration stored within the Terminal.Gui library defines two more themes: "Dark", and "Light". Additional themes can be defined in the configuration files. 

The Json property `Theme` defines the name of the theme that will be used. If the theme is not found, the default theme will be used.

Themes support defining ColorSchemes as well as various default settings for Views. Both the default color schemes and user defined color schemes can be configured. See [ColorSchemes](~/api/Terminal.Gui/Terminal.Gui.Colors.yml) for more information.

# Example Configuration File

```json
{
  "$schema": "https://gui-cs.github.io/Terminal.Gui/schemas/tui-config-schema.json",
  "Application.QuitKey": {
    "Key": "Esc"
  },
  "AppSettings": {
    "UICatalog.StatusBar": false
  },
  "Theme": "UI Catalog Theme",
  "Themes": [
    {
      "UI Catalog Theme": {
        "ColorSchemes": [
          {
            "UI Catalog Scheme": {
              "Normal": {
                "Foreground": "White",
                "Background": "Green"
              },
              "Focus": {
                "Foreground": "Green",
                "Background": "White"
              },
              "HotNormal": {
                "Foreground": "Blue",
                "Background": "Green"
              },
              "HotFocus": {
                "Foreground": "BrightRed",
                "Background": "White"
              },
              "Disabled": {
                "Foreground": "BrightGreen",
                "Background": "Gray"
              }
            }
          },
          {
            "TopLevel": {
              "Normal": {
                "Foreground": "DarkGray",
                "Background": "White"
              ...
              }
            }
          }
        ],
        "Dialog.DefaultEffect3D": false
      }
    }
  ]
}
```

# Configuration File Schema

Settings are defined in JSON format, according to the schema found here: 

-----------------------------

# Current PR Status

Latest:

* To add properties, just annotate any static property on any class with `[SerializableConfigurationProperty]`
* Now supports `AppSettings` so apps can use CM for app-specific settings
* Option throw json exceptions on load (off by default)
* Refactored API and schema significantly (again)
* More theme settings; note esp. `Border`

What's Missing:

- [ ] Move file system watcher into library out of UI Catalog
- [ ] Add support to `ConsoleDriver` such that default glyphs can be configured (e.g. `Driver.LeftBracket` or `Driver.VLine`
- [ ] Expose `ConsoleDriver.DiagnosticFlags` 
- [ ] Find and expose all "arbitrary" constants (e.g. `CollectionNavigator.TypingDelay`
- [ ] Figure out how to enable `Command`s to be configured externally!
- [ ] Figure out how to enable `KeyBinding`s to be configured externally!
- [ ] Schema is out of date
- [ ] @tzind's ideas about Stream based API


tig:
Question (@BDisp & @tznind esp):

Should VisualStyleManager be a sub-set of a broader config/settings manager? Or should I keep the whole thing scoped to just visual settings?

tznind:
> Question (@BDisp & @tznind esp):
> 
> Should VisualStyleManager be a sub-set of a broader config/settings manager? Or should I keep the whole thing scoped to just visual settings?

I think its best to develop one thing (e.g. color management) but architect it so its easy to add new features (e.g. use shadow buttons) and 'feature adjacent' stuff (e.g. custom keybindings). later on.

It would be nice if there was one file that had all the settings in them and for the schema to be able to grow over time (e.g. into keybindings etc).

A single file (with the override mechanic you have discussed) will be easier for a developer to maintain etc.

------------------

Haven't looked in detail at this new PR but it sounds awesome!

There are various UIs and code in TerminalGuiDesigner around creating/setting ColorScheme that might be of interest if you are wanting to have a UICatalog Scenario where you can change colors on demand.  Or it may be easier just to write again as it's not that complicated.

TerminalGuiDesigner color scheme UIs can be seen in this discussion:
https://github.com/gui-cs/Terminal.Gui/discussions/1887

This is the PR where ColorScheme editing was added:
https://github.com/gui-cs/TerminalGuiDesigner/pull/33

tig:
Check out the latest. It's pretty slick.

BDisp:
> Question (@BDisp & @tznind esp):
> 
> Should VisualStyleManager be a sub-set of a broader config/settings manager? Or should I keep the whole thing scoped to just visual settings?

I think that only one file that include all the config/settings with different sections is the better option.

tig:
Here's the current config file in this PR.

[config.json](https://github.com/gui-cs/Terminal.Gui/blob/5b4c9bae70c58d6af4eb5da0309b542aa5c0e61a/Terminal.Gui/Resources/config.json)

Note that it has two primary sections:

- `ApplicationSettings` (which I think I'll rename to `Settings`) 
- `VisualStyles` 

Here's the schema (does not yet have `Settings` defined yet.

[tui-config-schema.json]](https://github.com/gui-cs/Terminal.Gui/blob/5b4c9bae70c58d6af4eb5da0309b542aa5c0e61a/docfx/schemas/visualstyle-schema.json)

BDisp:
> * `ApplicationSettings` (which I think I'll rename to `Settings`)

Everything on a config file is related to settings and I think that only rename to `Application` is enough.

tznind:
Got a crash on startup of UICatalog from `.tui` folder not existing.  Sent a fix https://github.com/tig/Terminal.Gui/pull/7

tig:
Ok y'all. This beast is ready for your review.

Check out

- The UI Catalog scenario "Configuration Editor"
  - Edit one of the files and add a ColorScheme named "FarOut" that is distinctive. Then go back to UI Config and use the ColorS Schemes menu to select it.

BDisp:
I already test it and I like very much the way you organize all the precedence settings. Congrats for your great work and thanks for you provide all this.

tig:
Now includes themes!!!

Czech it out!

tig:
Check out the new Dark & Light themes. Let me know what you think of them and tweaks you might suggest.

I am going to add one more setting (likey default Border style) to prove my schema is flexible enough and then call this thing done. We can add more later...

tig:
#2258 has given me a thought...

Apps want to have their own settings persisted. I have not considered this yet for `ConfigurationManager`.

Let's use UICatalog as an example, now that it has a `SplitPanel` for the Categories and Scenario panes.

If UICatalog wanted to persist the SplitterDistance...

```cs
class UICatalogConfig: Config<UICatalogConfig> {

  public int? SplitterDistance { get; set} = 25;

  public override void GetHardCodedDefaults () {};
  public override void Apply () {};
  public override void CopyUpdatedProperitesFrom (UICatalogConfig changedConfig) 
  {
    if (changedConfig.SplitterDistance.HasValue) {
       SplitterDistance = changedConfig.SplitterDistance
    }
  }
}

void LoadedHandler ()
{
  if (!ConfigurationManager.Config.AppConfig.ContainsKey("UICatalog")) {
  {
      ConfigurationManager.Config.AppConfig("UICatalog") = new UICatalogConfig ();
  }
  ContentPane.SplitterDistance = ConfigurationManager.Config.AppConfig("UICatalog");
}

void ClosingHandler() 
{
  var uiCatalogConfig = ConfigurationManager.Config.GetAppConfig("UICatalog");
  uiCatalogConfig.SplitterDistance = ContentPane.SplitterDistance;
  ConfigurationManager.Config.Save();
}
```

Missing from the current impl is the `Save` method.

Questions raised by this:

1) What config file location is used when `ConfigurationManager.Config.Save()` is called?
    - Perhaps `Save` should let app indicate locations. e.g. (`Save (ConfigLocation.AppLocal)` or `Save (ConfigLocation.AppHome)` or `Save (ConfigLocation.GlobalHome)`)?  `ConfigLocation` could be flag-based enum so that `ConfigLocation.AppLocal | ConfigLocation.AppHome` saves to both `./.tui/UICatalog.config.json` and `~/.tui/UICatalog.config.json`)?
    - If UICatalog wanted to save the `SelectedTheme` the least surprising default location would be `{ "Themes:" }` in `ConfigLocation.AppHome` (so that any instance of UIConfig would use that setting). 
    - Thus, I think the answer to this question is `ConfigurationManager.Config.Save()` should default to `ConfigLocation.AppHome`.

2) Does `"SplitterDistance": 42` go within `{ "Settings: " `, `{ "Themes:" `, `{ "UICatalog:" `, `{ "Settings: { "UICatalog" : `, or ` "Themes" : "ThemeDefinitions": [ "Default" : { "App":`"?
    - The app's `Config` class (`UICatalogConfig` would be persisted to the top of the JSON schema (a sibling of `Settings` and `Themes`). Thus  `SplitterDistance` would be persisted to `{ "UICatalog" : { "SplitterDistance" : 42 }  }` .

BDisp:
1. `UICatalogConfig` class may provide a property pointed to where should be the localization to be used by the save method, depend on the OS.
2. I think` { "Settings: { "UICatalog" : ` is more appropriate.

tznind:
Love how much this is going to empower the end users of Terminal.Gui applications.  Being able to switch between `UseSystemConsole`  true and false alone is huge and the color stuff is icing on the cake (and thats before we get to the rest!).

I've not dived too deeply into the code so this may be off base.  But it would be great if the persistence API was based on `Stream` class so that an API user could use whatever files they wanted or even load/save the config as part of a larger configuration file/system.

For example:

```csharp
// Assume these are opened somewhere
Stream file1;
Stream s3Bucket;
Stream myMainConfigTuiSection;

var manager = new ConfigurationManager();

IState state = manager.CreateStateFromStream(file1)
			.OverrideWith(s3Bucket)
			.OverrideWith(myMainConfigTuiSection);


// Assume this is created by users main code
View rootView;

// This would recursively propogate to children
rootView.SetState(state);

Application.Run(rootView);

var newState = rootView.GetState();
manager.SaveState(newState, s3Bucket);
```

We could still have a default implementation which generates streams for the various files e.g.

```csharp
IConfigStreamCollection files = new DefaultConfigFiles();
IState state = manager.CreateStateFromCollection(files)
```

tig:
@migueldeicaza suggested that the FIleSystemWatcher be builtin to the library, vs. in the app as I did with UICatalog

I resisted this idea initially because any errors in the JSON will throw an exception. Thus, if a global config.json fle is written with an error in it ALL running TUI apps will crash when the file is saved. If FSW is by the app, then on the app that's implemented the watcher will crash.

I suppose I could `catch` all exceptions and just abort the load. But we have no logging mechanism so with this detecting a config file error would be much harder to notice. 

Thoughts?

tig:
Related to the above: The current implementation sets properties on the underlying classes (e.g. Application or FrameView)  vs having those classes call into the configurationmanager API to retrieve settings (and potentially get notified that config has changed). 

I keep going back and forth on this. Would appreciate any ideas or thoughts y'all have.

tig:
> @migueldeicaza suggested that the FIleSystemWatcher be builtin to the library, vs. in the app as I did with UICatalog
> 
> I resisted this idea initially because any errors in the JSON will throw an exception. Thus, if a global config.json fle is written with an error in it ALL running TUI apps will crash when the file is saved. If FSW is by the app, then on the app that's implemented the watcher will crash.
> 
> I suppose I could `catch` all exceptions and just abort the load. But we have no logging mechanism so with this detecting a config file error would be much harder to notice.
> 
> Thoughts?

I just realized another reason against having the library automatically load/apply config changes:

- Only an app knows where it's caching config info at runtime. E.g. UICatalog stores a copy of `IsMouseDisabled` in a global var (miIsMouseDisabled.Checked). For UICatalog to know this setting was changed, it would have to subscribe to some event to know that Application.IsMouseDisabled changed. This is a lot of logic that is currently handled fairly simply.

tznind:
> I suppose I could catch all exceptions and just abort the load. But we have no logging mechanism so with this detecting a config file error would be much harder to notice.

I think this is the correct approach.  If a settings file is broken it should be ignored.  If visibility is an issue then you could always add a `Console.WriteLine("x file is corrupt")` after `Application.Shutdown()` completes.

Also we should ensure that (as far as possible) there's no way to save a broken configuration with the API (i.e. validate before writing to disk).  That way if its broken its because a user edited a file badly.

> @migueldeicaza suggested that the FIleSystemWatcher be builtin to the library, vs. in the app as I did with UICatalog

I think this is nice to have so it works out of the box.  

If there is an interface for the watching/data source class then that would be even better as that would allow for using a different config source backend (e.g. load configs from DB or a GET query etc).

> Related to the above: The current implementation sets properties on the underlying classes (e.g. Application or FrameView) vs having those classes call into the configurationmanager API to retrieve settings (and potentially get notified that config has changed).

> I keep going back and forth on this. Would appreciate any ideas or thoughts y'all have.

Lots of different possible implementations for this! Each is going to have advantages and disadvantages.

If `ConfigurationManager` needs to know the implementation details of all views (i.e. what to save/load) that could grow arms and legs.  

Especially if it has to find children and apply changes to them too.  There are some controls (e.g. `TabView`) where not all possible child views are in the hierarchy at once  (`TabView` hot swaps its content as the user switches views).

If I were implementing it I would probably pass the `Settings` to the root `View` and let it take responsibility for passing it on.  

But this has its own disadvantages e.g. for new Views created during runtime.  A parent would need to hold onto the settings and then give them to newly created instances (e.g. in `Add`).  Some of this complexity can be seen in `ViewTracker` which was an attempt to do this for keybindings only in UICatalog

tig:
> > I suppose I could catch all exceptions and just abort the load. But we have no logging mechanism so with this detecting a config file error would be much harder to notice.
> 
> I think this is the correct approach. If a settings file is broken it should be ignored. If visibility is an issue then you could always add a `Console.WriteLine("x file is corrupt")` after `Application.Shutdown()` completes.

I now agree. I will refactor such that broken settings files don't throw exceptions and at the minimum `Debug.WriteLine` is used effectively. I'll consider a way to have `Application.Shutdown()` do a `ConsoleWriteLine` if needed.

> 
> Also we should ensure that (as far as possible) there's no way to save a broken configuration with the API (i.e. validate before writing to disk). That way if its broken its because a user edited a file badly.

Because I only use System.Text.Json to serialize, I don't think this is a problem. I invested in a defining a schema so that this will only get more robust over time.

> 
> > @migueldeicaza suggested that the FIleSystemWatcher be builtin to the library, vs. in the app as I did with UICatalog
> 
> I think this is nice to have so it works out of the box.
> 
> If there is an interface for the watching/data source class then that would be even better as that would allow for using a different config source backend (e.g. load configs from DB or a GET query etc).

Good idea.

> 
> > Related to the above: The current implementation sets properties on the underlying classes (e.g. Application or FrameView) vs having those classes call into the configurationmanager API to retrieve settings (and potentially get notified that config has changed).
> 
> > I keep going back and forth on this. Would appreciate any ideas or thoughts y'all have.
> 
> Lots of different possible implementations for this! Each is going to have advantages and disadvantages.
> 
> If `ConfigurationManager` needs to know the implementation details of all views (i.e. what to save/load) that could grow arms and legs.
> 
> Especially if it has to find children and apply changes to them too. There are some controls (e.g. `TabView`) where not all possible child views are in the hierarchy at once (`TabView` hot swaps its content as the user switches views).
> 
> If I were implementing it I would probably pass the `Settings` to the root `View` and let it take responsibility for passing it on.
> 
> But this has its own disadvantages e.g. for new Views created during runtime. A parent would need to hold onto the settings and then give them to newly created instances (e.g. in `Add`). Some of this complexity can be seen in `ViewTracker` which was an attempt to do this for keybindings only in UICatalog

I am shooting for a hybrid approach based on the `Config<T>` class. The idea is any class (e.g. `TabView`) can implement `Config<TabView>` within `TabView.cs`. Right now I have no way of discovering these things, but I was planning on using 
a class `[ConfigProperty(name)`] attribute (where `name`) is the name of a static property on the class that exposes the `Config<T>` instance for the class T.

tig:
Another issue is verbosity of the .json. I don't want to make it overly complicated for editing.

Theme's currently look like this:

```json
 "Themes": {
    "ThemeDefinitions": [
      {
        "Default": {
          "ColorSchemes": [
            {
...
            }
          ],
          "DefaultWindowBorderStyle": "Single",
          "DefaultFrameViewBorderStyle": "Single"
        }
      }
    ],
    "SelectedTheme": "Default"
  }
```

What I don't want to have happen is this:

```json
"Themes": {
    "ThemeDefinitions": [
      {
        "Default": {
          "ColorSchemes": [
            {
...
            }
          ],
          "Window" : {
             "DefaultBorderStyle": "Single"
          },
          "FrameView" : {
              "DefaultBorderStyle": "Single"
          }
        }
      }
    ],
    "SelectedTheme": "Default"
  }
```

Or... maybe I do. ???

BDisp:
> Or... maybe I do. ???

In the cases you want add more definitions, is better, like DrawFrameWindow, BorderBrush, etc.

BDisp:
@tig I can't run `NetDriver` either on `Windows` or `WSL` with this PR. It fallback into `WindowsDriver` on `Windows` and into `CursesDriver` on `WSL`.

tig:
I've still got work to do on this, but y'all should check out the latest...

- New, simpler, schema
- To add properties, just annotate any static property on any class with `[SerializableConfigurationProperty]`
- There are two scopes: `SettingsScope` and `ThemeScope`. Theme scope properties make up themes.
- I've added even more settings

Todo:

- [ ] More settings
- [ ] Add a `AppScope` so app devs can add their own settings
- [ ] @tzind's ideas about Stream based API
- [ ] Finish fixing the schema
- [ ] Move file system watcher into libarary
- [ ] even more unit tests

tig:
Major update here, folks,. See the first post for details.

I'm going to merge this soon. Ideally I'll figure out how to make it stream based (per @tznind) and move the file system watcher out of UI catalog first.

tig:
Closing; targetting v2 with #2362 instead.

PR #2294, opened 2023-01-17
Fixes #2293. WindowsDriver forces minimum console height equal to 1, but should allow be 0.

### LINKED ISSUE #2293: WindowsDriver forces minimum console height equal to 1, but should allow be 0.
This is only happens if `HeightAsBuffer` is true. At the time was to avoid an error that I already found a fix that I'll submit a PR.

BDisp:
Fixes #2293 -  Only the 'OutputHandle' can be set a height to 0 and the `ScreenBuffer` must have a minimum height of 1, in the case where we don't want use buffer.

## Pull Request checklist:


tig (approved):
Nice catch & fix!

PR #2301, opened 2023-01-20
Fixes #2296.  `Attribute.Make` works even if a driver isn't loaded

### LINKED ISSUE #2296: Refactor Attribute.Make to work even if a driver has not been loaded
Today:

https://github.com/gui-cs/Terminal.Gui/blob/2e7c2c83775de2cf8153e07cc8050ba07cecdd07/Terminal.Gui/Core/ConsoleDriver.cs#L174-L185

This makes it awkward to have the configuration manager set things up.

I'd like to propose the following:

- Modify `Attribute` such that a value of `-1` for `Attribute.Value` means "Not initialized by driver"
- Modify `Attribute.Make` to set `Value` to `-1` if it is called with no driver 
- Modify `ConsoleDriver.CreateColors` which is called from `ConsoleDriver.Init` to go through `Colors` and if any `Attribute.Value == -`, call `Make` again.

The only question I have about this is the code in `CursesDriver` that sets the `hasColors` parameter to `ConsoleDriver.CreateColors` to `true` in some cases. 

https://github.com/gui-cs/Terminal.Gui/blob/2e7c2c83775de2cf8153e07cc8050ba07cecdd07/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs#L898-L932

@BDisp or @migueldeicaza : Do you know when/why it does this? Is it really still needed?

tig:
Fixes #2296

- Adds a `Invalid` constant to the `Color` enum - this is in prep for Configuration Manager where there is a need to know if a Color has been set or not
- Adds an `HasValidColors` property to `Attribute` that is false if either of the foreground/background colors is `Invalid`
- Changes the `set` semantics of `ColorScheme` such that if an `Attribute` being set is `Invalid`, it does nothing (the alternative was to throw an exception, but this would break existing code that passes around un-initialized `Attributes`; a NOP does no harm).
- Adds an `Initialized` property to `Attribute` that indicates if the `Attribute` was initialized by a drive or not.
- Adds unit tests for the above changes
- Outputs debug spew in `ConsoleDriver` if un-initilaized `Attributes` are passed to `SetAttribute`
- Changes all `ConsoleDriver`s to initialize `currentAttribute` with a  valid attribute 

Note there's a lot of code (esp in Unit Tests) that create new instances of `ColorScheme` and then do things without actually setting any valid values. Previously this all worked because none of these tests are actually inspecting Attributes.

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -160,10 +160,11 @@ namespace Terminal.Gui {
 
 		public override void UpdateScreen () => window.redrawwin ();
 
-		Attribute currentAttribute;
+		Attribute currentAttribute = new Attribute (Color.White, Color.Black);
 
 		public override void SetAttribute (Attribute c)
 		{
+			base.SetAttribute (c);
 			currentAttribute = c;
 			Curses.attrset (currentAttribute);
 		}
@@ -201,6 +202,7 @@ namespace Terminal.Gui {
 
 		public override void SetColors (ConsoleColor foreground, ConsoleColor background)
 		{
+			// BUGBUG: This code is never called ?? See Issue #2300
 			int f = (short)foreground;
 			int b = (short)background;
 			var v = colorPairs [f, b];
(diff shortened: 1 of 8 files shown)


tig:
@BDisp & @tznind - I'd appreciate your eyes on this PR before I merge it. Thanks.

PR #2309, opened 2023-01-29
Fixes #2308. Redraw passed bounds bigger than its client area.

### LINKED ISSUE #2308: Redraw passed `bounds` bigger than its client area
**Describe the bug**
I've encountered this bug in developing #2258 

After resizing a `View` in `LayoutSubviews` to be smaller then the subsequent redraw that is triggered is passing a larger `Rect` into `Redraw` than the bounds.

image

I think this issue is to do with `view.NeedDisplay.Width` being stale and the parent `View` passing the larger width/height of the two:

https://github.com/gui-cs/Terminal.Gui/blob/3d75be4ea60aa670544e39394b44adf6733a39df/Terminal.Gui/Core/View.cs#L1543-L1544

Surely a `View` should never be asked to redraw a bigger area than it currently occupies?

Or is this to do with menus/spill overs etc where a view deliberately paints outside the box? If so then I think quite a few of my views will need to be adjusted to only listen to `Bounds` and not the passed parameter `bounds`.

BDisp:
Fixes #2308 - The `Bounds` client are is now passing correctly to the `Redraw` method. The `NeedDisplay` now have all the old and new area where can be used to clear all the changed area. Normally a container clear all the client area before redraw the sub-views, but there are cases where this is necessary.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -446,14 +446,11 @@ namespace Terminal.Gui {
 		public virtual Rect Frame {
 			get => frame;
 			set {
-				if (SuperView != null) {
-					SuperView.SetNeedsDisplay (frame);
-					SuperView.SetNeedsDisplay (value);
-				}
+				var rect = GetMaxNeedDisplay (frame, value);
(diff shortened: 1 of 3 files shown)


tig (approved):
LGTM!

PR #2316, opened 2023-02-04
An experiment in renaming Terminal.Gui to Terminal.UI

tig:
Let's see how this sits. If we like, then I propose creating a new project for v2 at https://github.com/gui-cs/Terminal.UI instead of having a v2 branch.


tznind:
Would this mean a new nuget package? That would prevent current users from seeing the new version via Dependabot and the NuGet package browser?

If so that would mean a lot of our current users might get left behind and not realise there is a new version.

BDisp:
In this case is preferable to have v1.x and V2 separated.

tig:
We collectively determined this was a bad idea. Closing.

Issue #2319, opened 2023-02-05
Refactor `ConsoleDriver` to minimize duplicated code in all drivers


tig:
As illustrated by 

- #2318 (and many other examples)

...the design of `ConsoleDriver` has resulted in all of the sub-classes to have tons of duplicated code. In v2 we should refactor `ConsoleDriver` to fix this as much as possible. 

## Todos

- [x] Most drawing stuff has been consoldated
  - [x] `AddRune`
  - [x] `UpdateScreen`
  - [x] `Color/Attribute` stuff 
- [ ] #3119 - (IOW, there's more to do in reducing duplicate keyboard related code).
- [ ] #3029 - (Mouse stuff is still massively duplicated)

tig:
Partially fixed by #2320 - much more to do.

PR #2320, opened 2023-02-06
Fixes #2300. Removes unused `ConsoleDriver` APIs 

### LINKED ISSUE #2296: Refactor Attribute.Make to work even if a driver has not been loaded

### LINKED ISSUE #2300: Remove unused methods from `ConsoleDriver`
These methods appear to be unused:

```cs
		/// <summary>
		/// Set Colors from limit sets of colors.
		/// </summary>
		/// <param name="foreground">Foreground.</param>
		/// <param name="background">Background.</param>
		public abstract void SetColors (ConsoleColor foreground, ConsoleColor background);

		// Advanced uses - set colors to any pre-set pairs, you would need to init_color
		// that independently with the R, G, B values.
		/// <summary>
		/// Advanced uses - set colors to any pre-set pairs, you would need to init_color
		/// that independently with the R, G, B values.
		/// </summary>
		/// <param name="foregroundColorId">Foreground color identifier.</param>
		/// <param name="backgroundColorId">Background color identifier.</param>
		public abstract void SetColors (short foregroundColorId, short backgroundColorId);

		/// <summary>
		/// Disables the cooked event processing from the mouse driver.  At startup, it is assumed mouse events are cooked.
		/// </summary>
		public abstract void UncookMouse ();

		/// <summary>
		/// Enables the cooked event processing from the mouse driver
		/// </summary>
		public abstract void CookMouse ();
```

In all existing drivers they are either empty implementations or there are no codepaths that call them.

CursesDriver has code in `SetColors` but there's no code I can find that ever calls `ConsoleDriver.SetColors` or `Driver.SetColors`.

We should remove these methods me thinks, even though it might be ḃreaking change (seems very unlikely).

tig:
Fixes #2300
Parial fix for #2319


UnitTests/Drivers/ConsoleDriverTests.cs
```diff
@@ -440,8 +440,6 @@ public void MakePrintable_Does_Not_Convert_Ansi_Chars_To_Unicode (uint code)
 			Assert.Equal (code, actual.Value);
 		}
 
-		private static object packetLock = new object ();
```
BDisp, line 443: Without this lock sometimes the `TestVKPacket` fails. Do you think it isn't actually necessary?

tig: This code was accidentally deleted during a merge. I'll add it back.

Issue #2331, opened 2023-02-08
ScrollView may not be honoring clip region; CustomButton shows outside


eonyanov:
I created CustomButton class with FrameView, one Label for fill and one Label for text inside.
Then I put some buttons inside the scrollview.

![image](https://user-images.githubusercontent.com/6646489/217461309-deeaec3f-c7dd-4cfa-acf6-92b0b79e088d.png)

Buttons that are outside of the scrollview display a label.

BDisp:
Can you provide the CustomButton code, please. I seems that you are redraw it in a manner that it don't be respecting the `ScrollView` clip area. `FrameView` when is redrawing also set his clip area and may be are over-spilling the `ScrollView` clip area.

eonyanov:
Add how I create window with ScrollView
```
 public class ScrollViewTestWindow : Window
    {
        private List<Button> buttons;
        private const int BUTTONS_ON_PAGE = 7;
        private const int BUTTON_HEIGHT = 3;

        private ScrollView scrollView;
        
        public ScrollViewTestWindow()
        {
            Title = "ScrollViewTestWindow";
            Width = 80;
            Height = 25;
            
            var titleLabel = new Label("DOCUMENTS")
            {
                X = 0,
                Y = 0
            };
            
            scrollView = new ScrollView()
            {
                X = 0,
                Y = 1,
                Width = 27,
                Height = BUTTONS_ON_PAGE * BUTTON_HEIGHT,
                ShowVerticalScrollIndicator = true,
                ShowHorizontalScrollIndicator = false
            };
            scrollView.ClearKeybindings();          
          
            buttons = new List<Button>();
            Button prevButton = null;
            int count = 20;
            for (int j = 0; j < count; j++)
            {
                Pos yPos = prevButton == null ? 0 : Pos.Bottom(prevButton);
                var button = new ASCIICustomButton(j.ToString(), $"section {j}", 0, yPos, 25, BUTTON_HEIGHT);
                button.Id = $"button{j}";
                scrollView.Add(button);
                buttons.Add(button);
                prevButton = button;
            }

            var closeButton = new ASCIICustomButton( "close", "Close", 0, Pos.Bottom(prevButton), 25, BUTTON_HEIGHT);         
            scrollView.Add(closeButton);
            buttons.Add(closeButton);
            
            var pages = buttons.Count / BUTTONS_ON_PAGE;
            if (buttons.Count % BUTTONS_ON_PAGE > 0)
[...]

PR #2332, opened 2023-02-08
Illustrates #2331 (Scrollview not respecting clip) does not reproduce

BDisp:
Unit test for #2331 - I think there is some issue on how the user is drawing the CustomButton. As you can see in the unit test none of the CustomButton over-spill the `ScrollView`.

## Pull Request checklist:

- [x ] I have checked my code and corrected any poor grammar or misspellings


BDisp:
There is still a bug on this issue and this PR fixes it. If you run the `ScrollViewTests` you'll verify it will fail without the fix. So, the cause of the bug was if the sum of the left+width-1 or the top+height-1 are equal to zero, then there is no available space to draw, but it was drawing.

Issue #92, opened 2018-05-18
Remove dependency on ustring


migueldeicaza:
We might need to introduce string overloads

ebekker:
Isn't the more functional approach in F# to pipe the value, such as `"foobar" |> ustring`?

jpierson:
@ebekker, that definitely looks valid as long as `ustring` is not a constructor but whether it's more functional or not I think is subjective. I typically use the pipe forward operator for larger chains of transformations.

migueldeicaza:
Yeah, I think I might need to make ustring just an internal API, and an optional API everywhere :-(

asik:
You run into the same issue using SharpDX. That's why I always define an implicit conversion operator:

`let inline (!>) (x:^a) : ^b = ((^a or ^b) : (static member op_Implicit : ^a -> ^b) x) `

Then you just write `!> "blabla"` and don't need to worry what the target type is.
Example usage here:
https://gist.github.com/asik/ce229023fb0eb153bde6aacc384d8ac4

Agreed it would be better to avoid needing this, but just to say there's a half-decent workaround.

tig:
Changed title of this Issue from "F# the ustring API makes it annoying to use F#". 

In v2.0 we should remove the dependency on ustring (at least externally) and use System.Rune instead. While ustring is neat, the fact that artifacts of it are exposed via the public Terminal.Gui API is problematic (F# support just being one). Another example is just simple usage of the API requires users to import NStack (see the C# Example in the main README).

PR #2334, opened 2023-02-08
Fixes #2333. TextField is selecting badly a word on double click.

### LINKED ISSUE #2333: TextField is selecting badly a word on double click.
textfield-selection-word-issue

BDisp:
Fixes #2333 - We shouldn't ever use indexing with the Text property to compare chars, because they are indexed as byte instead of the representation of the displayed character.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/TextField.cs
+++ b/Terminal.Gui/Views/TextField.cs
@@ -1064,8 +1064,8 @@ namespace Terminal.Gui {
 				EnsureHasFocus ();
 				int x = PositionCursor (ev);
 				int sbw = x;
-				if (x == text.Count || (x > 0 && (char)Text [x - 1] != ' '
-					|| (x > 0 && (char)Text [x] == ' '))) {
+				if (x == text.Count || (x > 0 && (char)text [x - 1] != ' ')
+					|| (x > 0 && (char)text [x] == ' ')) {
 
 					sbw = WordBackward (x);
 				}
--- a/UnitTests/Views/TextFieldTests.cs
+++ b/UnitTests/Views/TextFieldTests.cs
@@ -1186,6 +1186,21 @@ namespace Terminal.Gui.ViewTests {
 			Assert.Equal ("-", newText);
 			Assert.Equal ("-1", oldText);
 			Assert.Equal ("-", tf.Text.ToString ());
+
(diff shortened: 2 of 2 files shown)


tznind:
Nice find, good catch!

tig:
> We shouldn't ever use indexing with the Text property to compare chars, because they are indexed as byte instead of the representation of the displayed character.

I'm confused by the phrasing here. `Text` is a `ustring` and thus is NOT `byte` based, right?

Regardless, I think it would help make the TextField code more clear if we renamed the `TextField` `text` member to `runes`, e.g.:

Old:

```cs
	public class TextField : View {
		List<Rune> text;
```

New:

```cs
	public class TextField : View {
		List<Rune> runes;
```

Also, note I found this by searching for `Text [`:

```cs
		List<Rune> DeleteSelectedText ()
		{
			ustring actualText = Text;
			SetSelectedStartSelectedLength ();
			int selStart = SelectedStart > -1 ? start : point;
			(var _, var len) = TextModel.DisplaySize (text, 0, selStart, false);
			(var _, var len2) = TextModel.DisplaySize (text, selStart, selStart + length, false);
			(var _, var len3) = TextModel.DisplaySize (text, selStart + length, actualText.RuneCount, false);
			var newText = actualText [0, len] +
				actualText [len + len2, len + len2 + len3];
			ClearAllSelection ();
			point = selStart >= newText.RuneCount ? newText.RuneCount : selStart;
			return newText.ToRuneList ();
		}
```

Is this code incorrect too? I don't think so, since it's counting Runes, but figured i'd ask.

FWIW, I searched the entire project for `(char)Text [` and found only the bad code you fixed:

image

tig:
> I think we only need the `NStack` because of the `Rune.ColumnWidth` and I maybe we can use the `System.Text.Rune` and make the own `Rune.ColumnWidth` method on the `Core` folder and change `ustring` to use `string`. The `ColumnWidth` may not be equal for all fonts type, some of them may have a 1 length and other 2. Thus, we can get rid of the `NStack`, but it's a lot of work. Is it worth for the v2?

Nice work @BDisp. 

See https://github.com/gui-cs/Terminal.Gui/issues/92 ;-)

PR #2344, opened 2023-02-13
Fixes 2343 - LineCanvas not respecting X and Y of clip bounds

### LINKED ISSUE #2343: LineCanvas `GenerateImage` and `Draw` assume bounds' location is always `0, 0`
This test illustrates:

```cs
		[Fact, AutoInitShutdown]
		public void TestLineCanvas_PositiveLocation ()
		{
			var x = 1;
			var y = 1;
			var v = GetCanvas (out var canvas, x, y);

			// outer box
			canvas.AddLine (new Point (x, y), 9, Orientation.Horizontal, BorderStyle.Single);
			canvas.AddLine (new Point (x + 9, y + 0), 4, Orientation.Vertical, BorderStyle.Single);
			canvas.AddLine (new Point (x + 9, y + 4), -9, Orientation.Horizontal, BorderStyle.Single);
			canvas.AddLine (new Point (x + 0, y + 4), -4, Orientation.Vertical, BorderStyle.Single);


			canvas.AddLine (new Point (x + 5, y + 0), 4, Orientation.Vertical, BorderStyle.Single);
			canvas.AddLine (new Point (x + 0, y + 2), 9, Orientation.Horizontal, BorderStyle.Single);

			v.Redraw (v.Bounds);

			string looksLike =
@"    

 ┌────┬───┐
 │    │   │
 ├────┼───┤
 │    │   │
 └────┴───┘";
			TestHelpers.AssertDriverContentsAre (looksLike, output);
		}

		private View GetCanvas (out LineCanvas canvas, int x = 0, int y = 0)
		{
			var v = new View {
				Width = 10,
				Height = 5,
				Bounds = new Rect (x, y, 10, 5)
			};

			var canvasCopy = canvas =  new LineCanvas ();
			v.DrawContentComplete += (r)=> canvasCopy.Draw (v, v.Bounds);

			return v;
		}
```

```

Message: 
Assert.Equal() Failure
                   ↓ (pos 9)
Expected: ┌────┬───┐\n │    │   │\n ├────┼───┤\n │    │   │\n └─···
Actual:   ┌────┬───\n │    │\n ├────┼───\n │    │
                   ↑ (pos 9)
```

I'm confused about the right way to fix this (I need more caffeine or something):

https://github.com/gui-cs/Terminal.Gui/blob/2ead01a3c8fea7dc449483230c6fcab2b2bf7f90/Terminal.Gui/Core/Graphs/LineCanvas.cs#L49-L101

Should the result of `GenerateImage` always have the lines rooted at `0, 0` or should it be offset by `bounds.Location`?

I expected this change to line 68 would fix it:

```cs
						.Select (l => l.Intersects (inArea.X + x, inArea.Y + y))
```

But it didn't. I suspect there's a related bug in the `GetRuneForIntersects` path, but I'm not smart enough see what it is.

I discovered the need for this when attempting to refactor `View` (via `Border.DrawContent` to use `LineCanvas` instead of `Driver.DrawFrame`.

tznind:
Fixes #2343 - `GenerateImage` no longer assumes clip area starts at 0,0

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Core/Graphs/LineCanvas.cs
+++ b/Terminal.Gui/Core/Graphs/LineCanvas.cs
@@ -65,7 +65,7 @@ namespace Terminal.Gui.Graphs {
 				for (int x = 0; x < inArea.Width; x++) {
 
 					var intersects = lines
-						.Select (l => l.Intersects (x, y))
+						.Select (l => l.Intersects (inArea.X + x, inArea.Y + y))
 						.Where (i => i != null)
 						.ToArray ();
 
--- a/Terminal.Gui/Core/Graphs/LineCanvas.cs
+++ b/Terminal.Gui/Core/Graphs/LineCanvas.cs
@@ -79,15 +79,20 @@ namespace Terminal.Gui.Graphs {
 		}
 
 		/// <summary>
-		/// Draws all the lines that lie within the <paramref name="bounds"/> onto
-		/// the <paramref name="view"/> client area.  This method should be called from
+		/// Draws all the lines that lie within the <paramref name="sourceRect"/> onto
+		/// the <paramref name="view"/> client area at given <paramref name="drawOffset"/>.
+		///  This method should be called from
 		/// <see cref="View.Redraw(Rect)"/>.
 		/// </summary>
 		/// <param name="view"></param>
-		/// <param name="bounds"></param>
-		public void Draw (View view, Rect bounds)
+		/// <param name="sourceRect">The area of the canvas to draw.</param>
+		/// <param name="drawOffset">The point within the client area of 
+		/// <paramref name="view"/> to draw at.</param>
+		public void Draw (View view, Rect sourceRect, Point? drawOffset = null)
 		{
-			var runes = GenerateImage (bounds);
+			var offset = drawOffset ?? Point.Empty;
+
+			var runes = GenerateImage (sourceRect);
 
 			var runeRows = runes.GetLength (0);
 			var runeCols = runes.GetLength (1);
@@ -97,7 +102,7 @@ namespace Terminal.Gui.Graphs {
 					var rune = runes [y, x];
 
 					if (rune.HasValue) {
-						view.AddRune (bounds.X + x, bounds.Y + y, rune.Value);
+						view.AddRune (offset.X + x, offset.Y + y, rune.Value);
 					}
 				}
 			}
--- a/Terminal.Gui/Core/Graphs/LineCanvas.cs
+++ b/Terminal.Gui/Core/Graphs/LineCanvas.cs
@@ -89,12 +89,15 @@ namespace Terminal.Gui.Graphs {
 		{
 			var runes = GenerateImage (bounds);
 
-			for (int y = bounds.Y; y < bounds.Height; y++) {
-				for (int x = bounds.X; x < bounds.Width; x++) {
+			var runeRows = runes.GetLength (0);
+			var runeCols = runes.GetLength (1);
+
+			for (int y = 0; y < runeRows; y++) {
+				for (int x = 0; x < runeCols; x++) {
 					var rune = runes [y, x];
 
 					if (rune.HasValue) {
-						view.AddRune (x, y, rune.Value);
+						view.AddRune (bounds.X + x, bounds.Y + y, rune.Value);
 					}
 				}
 			}
--- a/Terminal.Gui/Core/Graphs/LineCanvas.cs
+++ b/Terminal.Gui/Core/Graphs/LineCanvas.cs
@@ -11,10 +11,10 @@ namespace Terminal.Gui.Graphs {
 	/// </summary>
 	public class LineCanvas {
 
-		
+
 		private List<StraightLine> lines = new List<StraightLine> ();
 
-		Dictionary<IntersectionRuneType, IntersectionRuneResolver> runeResolvers = new Dictionary<IntersectionRuneType, IntersectionRuneResolver> { 
+		Dictionary<IntersectionRuneType, IntersectionRuneResolver> runeResolvers = new Dictionary<IntersectionRuneType, IntersectionRuneResolver> {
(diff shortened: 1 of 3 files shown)


tig (commented):
How come  you didn't include the unit test I provided in my Issue?

I have use case where the I want to draw outside of `View.Bounds` and I need to be able to pass `Draw` a `bounds` that is larger than `View.Bounds`.

tznind:
Changing this to draft while I have time to think about this more deeply `Bounds` property is only ever 0,0 since it's setter is:

```csharp
set => Frame = new Rect (frame.Location, value.Size);
```

tig:
View.bounds also defines the nominal clip region too. 

Thinking about it more i don't really require this to be about view.bounds. What I need is an api that takes a rect and renders the canvas to it. I actually modified LineCanvas.Draw at one point in my adventures to treat rect as Screen relative if the view param was null.

tznind:
Ok.  `Draw` signature is now this:

```csharp
/// <summary>
/// Draws all the lines that lie within the <paramref name="sourceRect"/> onto
/// the <paramref name="view"/> client area at given <paramref name="drawOffset"/>.
///  This method should be called from
/// <see cref="View.Redraw(Rect)"/>.
/// </summary>
/// <param name="view"></param>
/// <param name="sourceRect">The area of the canvas to draw.</param>
/// <param name="drawOffset">The point within the client area of 
/// <paramref name="view"/> to draw at.</param>
public void Draw (View view, Rect sourceRect, Point? drawOffset = null)
{
 ...
}
```

I've updated your test to `TestLineCanvas_LeaveMargin_Top1_Left1`.  But I want to make sure I understand the requirements so am happy to add other tests if it doesn't represent your requirement.

The new draw method implementation changes the following:

- You can now draw anywhere in the view client area (specified by offset).  The test gives a draw offset of 1.
- The source rect is the part of the canvas you want drawn and now correctly supports x and y of the `sourceRect`

This method uses `View.AddRune` for rendering so you cannot draw outside of the bounds of the `View`.

We could add a `DrawToScreen` method which does allow for this kind of over-spilling.  It's signature would be something like:

 ```csharp
public void DrawToScreen(Rect sourceRect, Point? screenDrawOffset = null)
```

tig:
> Ok. `Draw` signature is now this:
> 
> ```cs
> /// <summary>
> /// Draws all the lines that lie within the <paramref name="sourceRect"/> onto
> /// the <paramref name="view"/> client area at given <paramref name="drawOffset"/>.
> ///  This method should be called from
> /// <see cref="View.Redraw(Rect)"/>.
> /// </summary>
> /// <param name="view"></param>
> /// <param name="sourceRect">The area of the canvas to draw.</param>
> /// <param name="drawOffset">The point within the client area of 
> /// <paramref name="view"/> to draw at.</param>
> public void Draw (View view, Rect sourceRect, Point? drawOffset = null)
> {
>  ...
> }
> ```
> 
> I've updated your test to `TestLineCanvas_LeaveMargin_Top1_Left1`. But I want to make sure I understand the requirements so am happy to add other tests if it doesn't represent your requirement.
> 
> The new draw method implementation changes the following:
> 
> * You can now draw anywhere in the view client area (specified by offset).  The test gives a draw offset of 1.
> * The source rect is the part of the canvas you want drawn and now correctly supports x and y of the `sourceRect`
> 
> This method uses `View.AddRune` for rendering so you cannot draw outside of the bounds of the `View`.
> 
> > What I need is an api that takes a rect and renders the canvas to it.
> 
> We could add a `DrawToScreen` method which does allow for this kind of over-spilling. It's signature would be something like:
> 
> ```cs
> public void DrawToScreen(Rect sourceRect, Point? screenDrawOffset = null)
> ```

Let me ask this question:

- What use-case drives `LineCanvas` having ANY dependency on `View`?  Why can't it always just operate on screen-relative coordinates and have callers do whatever View->Screen mapping required before using it?

The ONLY place you reference `View` in the current design is within `Draw`, right?

I've been thinking a ton about the current `View` class-hierarchy architecture and one of the biggest drivers of the (what I think is unneeded) complexity is a lack of clarity between: Screen coords (Frame), content-area coords (Bound), and the clip rect for Draw (which, sadly, is named 'bounds' everywhere). 

(Oh, and, BTW, referring to `Frame` as "screen" coords is also broken: Someday (v2?) we will support non-full-screen apps. In this case `Frame` needs to be "`Application.Top.Frame` relative", not screen relative).

I think a primitive drawing API like `LiineCanvas` that really has no need to know about `Views` should be independent of that hierarchy. I'm sorry I didn't notice this when I did the original CR.

tznind:
Fair points.  This is definitely worth exploring to get a good API.  LineCanvas is indeed a primitive self contained class.  Originally the 'output' method of the class was only the public method `public Rune? [,] GenerateImage (Rect inArea)`.  This method is still there and can be used (and this PR fixes its starting offset).

I added `Draw`  method as a helper for rendering to a `View` client area.  Mostly because working out how to iterate over a 2D array is annoying especially for new programmers.  Maybe a more user friendly return type would be `Dictionary<Point,Rune>`?

I can delete the `Draw` helper method to remove the dependency on `View` and we can just deal with the `GenerateImage` method directly instead?

tznind:
> Oh, and, BTW, referring to Frame as "screen" coords is also broken:

Do you mean you want less references to "screen" (e.g. ScreenToView, ViewToScreen)... 

Frame is not the same as screen.  They actually have nothing in common.  Views can be nested and the Frame property is just the location within the immediate parent's client area.  So they are 3 separate things (client area, frame and screen).

I think there is still value in using the 'screen' terminology for what you get when you use `Driver.Move(..)` directly.

tznind:
Ok have switched to Dictionary as return type of `GenerateImage` and removed draw.

So should be super easy to use in any context regardless of whether it is client area or screen coordinates.

Since we have not done any releases this should just be a breaking change for the split container PR (which will be easy to fix).

PR #2349, opened 2023-02-13
Fixes #2348. The Basic Colors scenario doesn't output the blue on white properly.

### LINKED ISSUE #2348: The Basic Colors scenario doesn't output the blue on white properly.
basic-colors-issue

BDisp:
Fixes #2348 - After the adding of the `Color.Invalid` the calculation used doesn't work anymore. Converting to an array and changing from `foreach` to `for` statement fixed the issue.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
@@ -1013,8 +1013,6 @@ namespace Terminal.Gui {
 				return Curses.COLOR_YELLOW | Curses.A_BOLD | Curses.COLOR_GRAY;
 			case Color.White:
 				return Curses.COLOR_WHITE | Curses.A_BOLD | Curses.COLOR_GRAY;
-			case Color.Invalid:
-				return Curses.COLOR_BLACK; 
 			}
 			throw new ArgumentException ("Invalid color code");
 		}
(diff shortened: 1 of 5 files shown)


tig:
> Fixes #2348 - After the adding of the `Color.Invalid` the calculation used doesn't work anymore. Converting to an array and changing from `foreach` to `for` statement fixed the issue.

It was my goal when I added `Color.Invalid` was to do it in a way that wasn't a breaking change. You've demonstated that it WAS a breaking change. What the Scenario is doing is likely a rare use-case, but the fact you had to change Scenario code means what I did was a breaking change.

Since my change to add `Color.Invalid` is still in `develop` we should consider a way to fix the underlying code such that your Scenario code continues to function as it did previously.

What do you think?

BDisp:
> It was my goal when I added `Color.Invalid` was to do it in a way that wasn't a breaking change. You've demonstated that it WAS a breaking change. What the Scenario is doing is likely a rare use-case, but the fact you had to change Scenario code means what I did was a breaking change.
> 
That really is a breaking change. 

> Since my change to add `Color.Invalid` is still in `develop` we should consider a way to fix the underlying code such that your Scenario code continues to function as it did previously.
> 
If you also look at the `InvertedColors` scenario it have a weird behavior. The `Color.Invalid` falls to a valid color which it's not what you want.

> What do you think?

I think it should be removed. I will do that and also add a unit test with a code like the scenario to ensure that the `Color` enum won't break. The `HasValidColors` must be compare by the `-1` value but without having new value in the `Color` enum. I'll submit a commit here with the changes.

tig (approved):
Thank you!

PR #2353, opened 2023-02-18
Adds Snake Scenario

tznind:
snake

## Pull Request checklist:


tznind:
Ok weird, there's a test that never ends after adding this Scenario... need to dig into that.

tig:
Amazing. LineCanvas is game changing for this project.

tznind:
Ok fixed Tests.  I think because the `ScenarioTests` were not disposing of `Scenario` they create the game loop never knew to quit and that caused test to keep running sometimes.

UICatalog/Scenarios/Snake.cs
```diff
@@ -0,0 +1,351 @@
...
+using System.Threading.Tasks;
+using Terminal.Gui;
+using Terminal.Gui.Graphs;
+using Attribute = Terminal.Gui.Attribute;
+
+namespace UICatalog.Scenarios {
+	[ScenarioMetadata (Name: "Snake", Description: "The game of apple eating.")]
+	[ScenarioCategory ("Colors")]
+	public class Snake : Scenario {
+		private bool isDisposed;
```
tig, line 14: I didn't notice this in Animation until after I merged but I don't think this works right. In one run i got an exception in the `while(!isDisposed)` loop because I think I hit exit at just the right time. Are you sure this is the proper way to detect the object has been deleted?

tig (changes requested):

UICatalog/Scenarios/Snake.cs
tznind: Hmn yeah... looks like `Scenario` is disposable (`class Scenario : IDisposable`) but nobody actually ever disposes it.  Should be an easy fix.  Also need to update it to the latest line canvas signature.

tznind:
Fixed the stack overflow, stupid typo :facepalm:

tig (approved):

PR #2355, opened 2023-02-19
Fixes #2354. View.Redraw doesn't clear itself and PositionCursor doesn't ensure focus when a prior view was disabled.

### LINKED ISSUE #2354: View.Redraw doesn't clear itself and PositionCursor doesn't ensure focus when a prior view was disabled.
If a view with subviews is added the clear method isn't executed and only the subviews are redrawn.

Allowing the view been clear by the base class and also ignore the `HasFocus` property in the `Redraw` method. If not ignore (default) the normal color is used if the view doesn't have the focus and the focus color is used if it have the focus. If ignore, then only the normal color is used whether is focused or not.

Only by deriving from `View` class and overridden the `Redraw` method but not needed with the solution above.

**Additional context**
For all this work some more improvements are needed, like on `CanFocus`, `Add`, `PositionCursor`, `Redraw`, `Enabled`.

Code to test the issue:

```cs
			Application.Init ();
			var root1 = new View () {
				Width = 20,
				Height = 10,
				ColorScheme = Colors.Base,
				//IgnoreHasFocusPropertyOnRedraw = true,
				Enabled = false
			};
			var label1 = new Label (new string ('c', 100)) {
				Width = Dim.Fill ()
			};
			var tv1 = new TextView () {
				Y = 2,
				Height = 1,
				Text = new string ('t', 100),
				Width = Dim.Fill (),
				Multiline = false
			};
			var labelFocus1 = new Label (new string ('f', 100)) {
				Y = 4,
				Width = Dim.Fill (),
				CanFocus = true
			};

			root1.Add (label1, tv1, labelFocus1);

			var root2 = new View () {
				X = 3,
				Y = 12,
				Width = 20,
				Height = 10,
				ColorScheme = Colors.Base,
				//IgnoreHasFocusPropertyOnRedraw = true
			};
			var label2 = new Label (new string ('c', 100)) {
				Width = Dim.Fill ()
			};
			var tv2 = new TextView () {
				Y = 2,
				Height = 1,
				Text = new string ('t', 100),
				Width = Dim.Fill (),
				Multiline = false
			};
			var labelFocus2 = new Label (new string ('f', 100)) {
				Y = 4,
				Width = Dim.Fill (),
				CanFocus = true
			};

			root2.Add (label2, tv2, labelFocus2);

			var sb = new StatusBar (new StatusItem [] {
				new StatusItem(Key.F9,"~F9~ Toggle Enable", () => {
					root1.Enabled = !root1.Enabled;
					root2.Enabled = !root2.Enabled;
				}),
				//new StatusItem(Key.F10,"~F10~ Toggle IgnoreHasFocusPropertyOnRedraw", () => {
				//	root1.IgnoreHasFocusPropertyOnRedraw = !root1.IgnoreHasFocusPropertyOnRedraw;
				//	root2.IgnoreHasFocusPropertyOnRedraw = !root2.IgnoreHasFocusPropertyOnRedraw;
				//	root1.SetNeedsDisplay();
				//	root2.SetNeedsDisplay();
				//})
			});

			Application.Top.Add (root1, root2, sb);
			Application.Run (Application.Top);

```

Try this code above and the output is this:

view-redraw-issue

After the fix, please remove all the commented code and the output is this:

view-redraw-fix

BDisp:
Fixes #2354 - A simple view now can be used as container with normal and focus color if `IgnoreHasFocusPropertyOnRedraw` is false (default) or only with the normal color if it's true. Now if it's possible to focus a enable view this is ensured, otherwise the cursor will be invisible.

## Pull Request checklist:

imagem

### DIFF
--- a/Terminal.Gui/Core/Toplevel.cs
+++ b/Terminal.Gui/Core/Toplevel.cs
@@ -908,6 +908,12 @@ namespace Terminal.Gui {
 		{
 			if (!IsMdiContainer) {
 				base.PositionCursor ();
+				if (Focused == null) {
+					EnsureFocus ();
+					if (Focused == null) {
+						Driver.SetCursorVisibility (CursorVisibility.Invisible);
+					}
+				}
 				return;
 			}
 
@@ -920,6 +926,9 @@ namespace Terminal.Gui {
 				}
 			}
 			base.PositionCursor ();
+			if (Focused == null) {
+				Driver.SetCursorVisibility (CursorVisibility.Invisible);
+			}
 		}
 
 		/// <summary>
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1510,11 +1510,7 @@ namespace Terminal.Gui {
 			var clipRect = new Rect (Point.Empty, frame.Size);
 
 			if (ColorScheme != null) {
-				if (!IgnoreHasFocusPropertyOnRedraw) {
-					Driver.SetAttribute (HasFocus ? GetFocusColor () : GetNormalColor ());
-				} else {
-					Driver.SetAttribute (GetNormalColor ());
-				}
+				Driver.SetAttribute (HasFocus ? GetFocusColor () : GetNormalColor ());
 			}
 
 			if (!IgnoreBorderPropertyOnRedraw && Border != null) {
@@ -2701,15 +2697,6 @@ namespace Terminal.Gui {
 		/// itself).
 		/// </summary>
 		public virtual bool IgnoreBorderPropertyOnRedraw { get; set; }
-
-		/// <summary>
-		/// Get or sets whether the view will use <see cref="Terminal.Gui.View.HasFocus"/> 
-		/// to set the <see cref="Terminal.Gui.ColorScheme.Focus"/> color if it's focused.
-		/// If <see langword="false"/> (the default), <see cref="View.Redraw(Rect)"/> will call <see cref="GetFocusColor"/>
-		/// or the <see cref="GetNormalColor"/> method.
-		/// If <see langword="true"/> only the <see cref="GetNormalColor"/> will be called.
-		/// </summary>
-		public virtual bool IgnoreHasFocusPropertyOnRedraw { get; set; }
 
 		/// <summary>
 		/// Pretty prints the View
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -951,6 +951,10 @@ namespace Terminal.Gui {
 				view.tabIndex = tabIndexes.IndexOf (view);
 				addingView = false;
 			}
+			if (view.Enabled && !Enabled) {
+				view.oldEnabled = true;
+				view.Enabled = false;
+			}
 			SetNeedsLayout ();
 			SetNeedsDisplay ();
 			OnAdded (view);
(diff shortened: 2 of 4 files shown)


tig:
Do we have a real user that has requested this? 

In my opinion we should be very leery of adding new features (methods/properties) to v1 that are not critically needed.

Everything we add to v1 now becomes more legacy we need to support/port in v2. 

I would much rather get focused on v2 for things like this.

However, if there is a real customer (developer external to this project) who is requesting this, and can't get their app to work without it, then I'll change my mind.

BDisp:
> Do we have a real user that has requested this?
>
We may not need the `IgnoreHasFocusPropertyOnRedraw` which is a new feature, but I think you didn't yet seen that this PR fixes three bugs:
- Fixes the `View.Redraw` that doesn't clear itself before redraw his subviews.
- Fixes the `Enabled` property if is `true` of a view that is being added to the superview that has the `Enabled` as `false` by not also setting the subview to `false`.
- Fixes the `PositionCursor` where the cursor isn't placed on the available view that can focus, is visible and is enabled when a previous view that was previously focused and his `Enabled` property is set to false.

The `GetFocusColor` method is a helper to get the `Focus` or the `Disabled` color.


> In my opinion we should be very leery of adding new features (methods/properties) to v1 that are not critically needed.
> 
I agree. The only thing I can remove here is the `IgnoreHasFocusPropertyOnRedraw` property. The rest I think is useful for both versions.


> Everything we add to v1 now becomes more legacy we need to support/port in v2.
> 
I agree. Do you want I do remove the `IgnoreHasFocusPropertyOnRedraw` property, then?


> I would much rather get focused on v2 for things like this.
> 
I agree with the cases of adding new features, but not for bug fixes before the v2 is published.


> However, if there is a real customer (developer external to this project) who is requesting this, and can't get their app to work without it, then I'll change my mind.

No there is no customer, that was only me that I detected some bugs and when I was fixing I thinking in a view to force the view to always using the normal color whether it's focused or not, if the `IgnoreHasFocusPropertyOnRedraw` property is `true`.

tig:
> > Do we have a real user that has requested this?
> 
> We may not need the `IgnoreHasFocusPropertyOnRedraw` which is a new feature, but I think you didn't yet seen that this PR fixes three bugs:
> 
> * Fixes the `View.Redraw` that doesn't clear itself before redraw his subviews.
> * Fixes the `Enabled` property if is `true` of a view that is being added to the superview that has the `Enabled` as `false` by not also setting the subview to `false`.
> * Fixes the `PositionCursor` where the cursor isn't placed on the available view that can focus, is visible and is enabled when a previous view that was previously focused and his `Enabled` property is set to false.

I know it seems like more work (and I know I'm guilty of not always doing this right myself), but these should really each be documented as individual Issues and submitted as separate PRs. 

For v1 `develop` (I think we can operate slightly more loosey-goosy in the v2 branch), I am wondering if we should start being super-hardcore about this. I'd be curious to know your reaction to me stating that.

> I agree. Do you want I do remove the `IgnoreHasFocusPropertyOnRedraw` property, then?

Yes.

BDisp:
> I know it seems like more work (and I know I'm guilty of not always doing this right myself), but these should really each be documented as individual Issues and submitted as separate PRs.
> 
Yes you are right. Sometimes I'm so lazy, sorry :-)

> For v1 `develop` (I think we can operate slightly more loosey-goosy in the v2 branch), I am wondering if we should start being super-hardcore about this. I'd be curious to know your reaction to me stating that.

Yes I'm starting working more in the v2 and I propose that you move the View2 code to the View, so we can only be focusing in the v2. But I think that all the v1 bugs that are transversal to the v2 must be merge in the v1 and all the v2 PR must be merged with the develop branch before they are submitted, otherwise there will be a bunch of merge conflicts to resolve.


>>    I agree. Do you want I do remove the IgnoreHasFocusPropertyOnRedraw property, then?

>Yes.
I already removed and changed the issue and PR's titles.

tig:
> > I know it seems like more work (and I know I'm guilty of not always doing this right myself), but these should really each be documented as individual Issues and submitted as separate PRs.
> 
> Yes you are right. Sometimes I'm so lazy, sorry :-)

No need to apologize. We're all in this for fun (but I think it's fun to sometimes pretend this all matters ;-).

> > For v1 `develop` (I think we can operate slightly more loosey-goosy in the v2 branch), I am wondering if we should start being super-hardcore about this. I'd be curious to know your reaction to me stating that.
> 
> Yes I'm starting working more in the v2 and I propose that you move the View2 code to the View, so we can only be focusing in the v2. But I think that all the v1 bugs that are transversal to the v2 must be merge in the v1 and all the v2 PR must be merged with the develop branch before they are submitted, otherwise there will be a bunch of merge conflicts to resolve.

I'm going to do this asap. As part of it, my plan is to remove a bunch of existing functionality that becomes redundant. 

In some cases, I plan on focusing on v2 and then back-porting to v1. I think that either way we'll have some merge issues to deal with. Not fun.

PR #2386, opened 2023-03-01
Fixes #1842  - Add support for toggling selected cells e.g. with space

### LINKED ISSUE #1842: Selecting multiple rows from the TableView element with the mouse
1. I wanted to ask a question about selecting multiple rows from the TableView element with the mouse.  What keyboard shortcut should be used? The Control key does not work as expected.

 2. When selecting with the cursor, is it only possible to select consecutive rows?  Or is there a keyboard shortcut that allows you to select arbitrary lines?

tznind:
Adds support for `Command.ToggleChecked` to `TableView`.   This allows pure keyboard multi selection (e.g. like you would with Ctrl+Click).

- Only works when `MultiSelect` is on 
- No default keybinding.  User enables it by registering a keybinding (e.g. `tableView.AddKeyBinding (Key.Space, Command.ToggleChecked);` )

toggle-selections

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/TableView.cs
+++ b/Terminal.Gui/Views/TableView.cs
@@ -987,8 +987,8 @@ namespace Terminal.Gui {
 				return;
 			}
 
-			var toggles = GetMultiSelectedRegionsContaining(selectedColumn, selectedRow)
-							.Where(s=>s.IsToggled).ToArray ();
+			var regions = GetMultiSelectedRegionsContaining(selectedColumn, selectedRow).ToArray();
+			var toggles = regions.Where(s=>s.IsToggled).ToArray ();
 
 			// Toggle it off
 			if (toggles.Any ()) {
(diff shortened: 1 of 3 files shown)


tznind:
Ready for testing now!  This feature is both nice to have and is an actual solution for https://github.com/gui-cs/Terminal.Gui/issues/1842 (the previous PR was mouse only and there was a workaround in the issue for keyboard).  See the issue OP question:

> When selecting with the cursor, is it only possible to select consecutive rows? Or is there a keyboard shortcut that allows you to select arbitrary lines?

The answer now is yes! just pick a key and set the keybinding.  

In the Scenario I have added this to TableEditor with keybinding of Space and same for the Tests but I don't think we want to claim this key by default as it is quite a niche feature and the key is likely to be bound by existing library users given its prominence.

It will be used in the new FileDialog2 as one of the ways of selecting multiple files files (can still use shift select or ctrl+mouse click).

tig (approved):

Issue #2391, opened 2023-03-02
Setup `v1_develop`, `v2_develop` and CI/CD pipelines for v2


tig:
We're not doing any CI/CD stuff on `v2` right now which is Bad.

tig:
Progress:

- `v2` has been renamed to `v2_develop`. To rename `v2`:

```
git branch -m v2 v2_develop
git fetch origin
git branch -u origin/v2_develop v2_develop
git remote set-head origin -a
```

tig:
PRs to `v2_develop` now run build/unit tests CI/CD and pass. See #2395

BDisp:
@tig when I run `git branch -u origin/v2_develop v2_develop` I receive this error:
```
fatal: the requested upstream branch 'origin/v2_develop' does not exist
hint:
hint: If you are planning on basing your work on an upstream
hint: branch that already exists at the remote, you may need to
hint: run "git fetch" to retrieve it.
hint:
hint: If you are planning to push out a new local branch that
hint: will track its remote counterpart, you may want to use
hint: "git push -u" to set the upstream config as you push.
hint: Disable this message with "git config advice.setUpstreamFailure false"
```
What I have to do?

BDisp:
Forget, I did the next instruction and is all right, thanks.

tig:
@dodexahedron tagging you so you have context.

PR #2395, opened 2023-03-03
Fixing #2391 - Updating CI/CD for v2_develop

tig:


tig:
@BDisp Please double check that I didn't do anything stupid in fixing the `DimFactor` tests in https://github.com/gui-cs/Terminal.Gui/pull/2395/commits/6f16af878bef18f8cb81d700fb07f2f4dbca7943  

Also, I don't really understand why `Border.Title` was getting unset (fixed in https://github.com/gui-cs/Terminal.Gui/pull/2395/commits/ebd9eb44477770f24dc114e6bfc6de3283df8a79). I donl't think it really matters because all that code will go away with the new v2 View design.

Issue #2359, opened 2023-02-20
Standardize all events on the `sender EventArgs` pattern


tig:
For v2.

See: [sender EventArgs pattern](https://learn.microsoft.com/en-us/dotnet/csharp/event-pattern#event-delegate-signatures) 

So we can put relevant info in and expand over time if needed without it being a breaking change.

tig:
Closing as i believe we're good here!

PR #2406, opened 2023-03-11
Fixes - #2359 All events now use sender EventArgs

tznind:
Fixes - #2359  Include a terse summary of the change or which issue is fixed.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Core/Border.cs
+++ b/Terminal.Gui/Core/Border.cs
@@ -131,7 +131,7 @@ namespace Terminal.Gui {
 				}
 			}
 
-			void Border_BorderChanged (Border border)
+			void Border_BorderChanged (object sender, EventArgs e)
 			{
 				Rect frame;
 				if (Border.Child != null && (Border.Child.Width is Dim || Border.Child.Height is Dim)) {
@@ -319,7 +319,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Invoked when any property of Border changes (except <see cref="Child"/>).
 		/// </summary>
-		public event Action<Border> BorderChanged;
+		public event EventHandler BorderChanged;
 
 		private BorderStyle borderStyle;
 		private bool drawMarginFrame;
@@ -1113,7 +1113,7 @@ namespace Terminal.Gui {
 		/// </summary>
 		public virtual void OnBorderChanged ()
 		{
-			BorderChanged?.Invoke (this);
+			BorderChanged?.Invoke (this, new EventArgs());
 		}
 	}
 }
--- a/Terminal.Gui/Core/CollectionNavigator.cs
+++ b/Terminal.Gui/Core/CollectionNavigator.cs
@@ -47,7 +47,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// Event arguments for the <see cref="CollectionNavigator.SearchStringChanged"/> event.
 		/// </summary>
-		public class KeystrokeNavigatorEventArgs {
+		public class KeystrokeNavigatorEventArgs : EventArgs{
 			/// <summary>
 			/// he current <see cref="SearchString"/>.
 			/// </summary>
@@ -66,7 +66,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// This event is invoked when <see cref="SearchString"/>  changes. Useful for debugging.
 		/// </summary>
-		public event Action<KeystrokeNavigatorEventArgs> SearchStringChanged;
+		public event EventHandler<KeystrokeNavigatorEventArgs> SearchStringChanged;
 
 		private string _searchString = "";
 		/// <summary>
@@ -87,7 +87,7 @@ namespace Terminal.Gui {
 		/// <param name="e"></param>
 		public virtual void OnSearchStringChanged (KeystrokeNavigatorEventArgs e)
 		{
-			SearchStringChanged?.Invoke (e);
+			SearchStringChanged?.Invoke (this, e);
 		}
 
 		/// <summary>
--- a/Terminal.Gui/Core/EventArgs/TimeoutEventArgs.cs
+++ b/Terminal.Gui/Core/EventArgs/TimeoutEventArgs.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using Terminal.Gui;
+using static Terminal.Gui.MainLoop;
+
+/// <summary>
+/// <see cref="EventArgs"/> for timeout events (e.g. <see cref="MainLoop.TimeoutAdded"/>)
+/// </summary>
+public class TimeoutEventArgs : EventArgs {
+	/// <summary>
+	/// Gets the timeout callback handler
+	/// </summary>
+	public Timeout Timeout { get; }
+
+	/// <summary>
+	/// Gets the <see cref="DateTime.Ticks"/> in UTC time when the 
+	/// <see cref="Timeout"/> will next execute after.
+	/// </summary>
+	public long Ticks { get; }
+
+
+	/// <summary>
+	/// Creates a new instance of the <see cref="TimeoutEventArgs"/> class.
+	/// </summary>
+	/// <param name="timeout"></param>
+	/// <param name="ticks"></param>
+	public TimeoutEventArgs (Timeout timeout, long ticks)
+	{
+		this.Timeout = timeout;
+		this.Ticks = ticks;
+	}
+}
--- a/Terminal.Gui/Core/MainLoop.cs
+++ b/Terminal.Gui/Core/MainLoop.cs
@@ -97,7 +97,7 @@ namespace Terminal.Gui {
 		/// Invoked when a new timeout is added. To be used in the case
 		/// when <see cref="Application.ExitRunLoopAfterFirstIteration"/> is <see langword="true"/>.
 		/// </summary>
-		public event Action<long> TimeoutAdded;
+		public event EventHandler<TimeoutEventArgs> TimeoutAdded;
 
 		/// <summary>
 		///  Creates a new Mainloop. 
@@ -161,7 +161,7 @@ namespace Terminal.Gui {
 			lock (timeoutsLockToken) {
 				var k = (DateTime.UtcNow + time).Ticks;
 				timeouts.Add (NudgeToUniqueKey (k), timeout);
-				TimeoutAdded?.Invoke (k);
+				TimeoutAdded?.Invoke (this, new TimeoutEventArgs(timeout, k));
 			}
 		}
 
--- a/Terminal.Gui/Core/Window.cs
+++ b/Terminal.Gui/Core/Window.cs
@@ -70,7 +70,7 @@ namespace Terminal.Gui {
 			}
 		}
 
-		void Border_BorderChanged (Border border)
+		void Border_BorderChanged (object sender, EventArgs e)
 		{
 			Rect frame;
 			if (contentView != null && (contentView.Width is Dim || contentView.Height is Dim)) {
--- a/Terminal.Gui/Views/FrameView.cs
+++ b/Terminal.Gui/Views/FrameView.cs
@@ -97,7 +97,7 @@ namespace Terminal.Gui {
 			}
 		}
 
-		void Border_BorderChanged (Border border)
+		void Border_BorderChanged (object sender, EventArgs e)
 		{
 			Rect frame;
 			if (contentView != null && (contentView.Width is Dim || contentView.Height is Dim)) {
--- a/Terminal.Gui/Views/PanelView.cs
+++ b/Terminal.Gui/Views/PanelView.cs
@@ -128,7 +128,7 @@ namespace Terminal.Gui {
 			Child.Initialized -= Child_Initialized;
 		}
 
-		private void Border_BorderChanged (Border obj)
+		private void Border_BorderChanged (object sender, EventArgs e)
 		{
 			AdjustContainer ();
 		}
--- a/UICatalog/Scenarios/CollectionNavigatorTester.cs
+++ b/UICatalog/Scenarios/CollectionNavigatorTester.cs
@@ -142,8 +142,8 @@ namespace UICatalog.Scenarios {
 
 			_listView.SetSource (_items);
 
-			_listView.KeystrokeNavigator.SearchStringChanged += (state) => {
-				label.Text = $"ListView: {state.SearchString}";
+			_listView.KeystrokeNavigator.SearchStringChanged += (s,e) => {
+				label.Text = $"ListView: {e.SearchString}";
 			};
 		}
 
@@ -179,8 +179,8 @@ namespace UICatalog.Scenarios {
 			_treeView.ExpandAll ();
 			_treeView.GoToFirst ();
 
-			_treeView.KeystrokeNavigator.SearchStringChanged += (state) => {
-				label.Text = $"TreeView: {state.SearchString}";
+			_treeView.KeystrokeNavigator.SearchStringChanged += (s,e) => {
+				label.Text = $"TreeView: {e.SearchString}";
 			};
 		}
 
--- a/UnitTests/Application/MainLoopTests.cs
+++ b/UnitTests/Application/MainLoopTests.cs
@@ -264,6 +264,35 @@ namespace Terminal.Gui.ApplicationTests {
 			Assert.False (ml.RemoveTimeout (token));
 		}
 
+		// Timeout Handler Tests
+		[Fact]
+		public void AddTimer_EventFired ()
+		{
+			var ml = new MainLoop (new FakeMainLoop ());
+			var ms = 100;
+			
+			var originTicks = DateTime.UtcNow.Ticks;
+
+			var callbackCount = 0;
+			Func<MainLoop, bool> callback = (loop) => {
+				callbackCount++;
+				return true;
+			};
+
+			object sender = null;
+			TimeoutEventArgs args = null;
+			ml.TimeoutAdded += (s, e) => {
+				sender = s;
+				args = e;
+			};
+
+			var token = ml.AddTimeout (TimeSpan.FromMilliseconds (ms), callback);
+
+			Assert.Same (ml,sender);
+			Assert.NotNull (args.Timeout);
+			Assert.True (args.Ticks - originTicks >= 100 * TimeSpan.TicksPerMillisecond);
+
+		}
 		[Fact]
 		public void AddTimer_Run_Called ()
 		{


tznind:
20 down, 48 to go 😓

Terminal.Gui/Core/View.cs
```diff
@@ -1354,26 +1354,28 @@ public class FocusEventArgs : EventArgs {
...
 		{
+			var view = e.View;
 			view.IsAdded = true;
 			view.x ??= view.frame.X;
 			view.y ??= view.frame.Y;
 			view.width ??= view.frame.Width;
 			view.height ??= view.frame.Height;
 
-			view.Added?.Invoke (this);
+			view.Added?.Invoke (this,e);
```
tznind, line 1367: Hmn so in the old version of `virtual void OnAdded (View view)` the `view` passed into this method was not passed on to `Invoke`.  Instead `this` was passed to `Invoke`.  

That means that the `view` reported by the `Added` event is the parent and not the child being added.

This is why the test is failing.  `Added_Removed` asserts the following:

```csharp
v.Added += (s,e) => {
	Assert.True (v.SuperView == e.View);
};
```

I propose we change this test to instead say

```csharp
v.Added += (s,e) => {
	Assert.True (v.SuperView == s);
	Assert.True (v == e.View);
};
```

tznind: Actually I am confused about this test.  This is what it looks like in `v2_develop`

The description for the `Added` event says:

> Event fired when a subview is being added to this view.

But the test is registering the event on the child (`v`) and then calling add on the parent (`t`).

Is this to say then that `Added` event covers both

- When a subview is added to yourself
- When you are added as a subview to someone else?

```csharp
public void Added_Removed ()
{
	var v = new View (new Rect (0, 0, 10, 24));
	var t = new View ();

	v.Added += (View e) => {
		Assert.True (v.SuperView == e);
	};

	v.Removed += (View e) => {
		Assert.True (v.SuperView == null);
	};

	t.Add (v);
	Assert.True (t.Subviews.Count == 1);

	t.Remove (v);
	Assert.True (t.Subviews.Count == 0);
}
```

BDisp: I agree. This event is normally used by the subview to get notified when the superview has add it and so your logic is perfectly right. By the way thanks for great effort.

tznind: > * When a subview is added to yourself
> 
> * When you are added as a subview to someone else?

Ok this is not the case.  I think the description and my expectation is in fact wrong.  Based on the `Add` method `Added` event is fired on a `View` **when it is aded to a parent**.  And **not** "when a subview is being added" [to it].

```
public virtual void Add (View view)
{
	if (view == null)
		return;
        [...]
	OnAdded (view);
```

BDisp: Yes, that why it's invoke `view.Added?.Invoke (this);`

BDisp: I propose this change, because it the user use the arg passed will think it's the  superview and that why the test is failing.

```cs
public virtual void OnAdded (ViewEventArgs e)
{
	var view = e.View;
	view.IsAdded = true;
	view.x ??= view.frame.X;
	view.y ??= view.frame.Y;
	view.width ??= view.frame.Width;
	view.height ??= view.frame.Height;
	e.View = this;
	view.Added?.Invoke (this, e);
}
```

BDisp: The same for the `OnRemoved`;

tznind: Ok I  have added a new `EventArgs`  called `SuperViewChangedEventArgs` so it is clear to user what view is being talked about (since it contains both `Parent` and `Child`).  Test now passes as:

```csharp
[Fact]
public void Added_Removed ()
{
	var v = new View (new Rect (0, 0, 10, 24));
	var t = new View ();

	v.Added += (s,e) => {
		Assert.Same (v.SuperView, e.Parent);
		Assert.Same (t, e.Parent);
		Assert.Same (v, e.Child);
	};

	v.Removed += (s, e) => {
		Assert.Same (t, e.Parent);
		Assert.Same (v, e.Child);
		Assert.True (v.SuperView == null);
	};

	t.Add (v);
	Assert.True (t.Subviews.Count == 1);

	t.Remove (v);
	Assert.True (t.Subviews.Count == 0);
}
```

BDisp: Much better solution, congrats for your brilliant implementation.

tig (approved):
Fantastic stuff.

What about cancelable events?

tznind:
> Fantastic stuff.

Thanks.  I'm keeping this in draft though as still working my way through all the `event Action<x>` events.  There are a lot!

> What about cancelable events?

Hopefully most would already use an event args with Handled (or similar) so should be ok I think?

tznind:
Only 7 more to go 🏁 👁️ but looks like they will all need event args classes.  

I've added any new EventArgs classes I've created in the EventArgs folder but using the normal namespace (`Terminal.Gui`).  A lot of existing ones are nested within parent classes though which I think isn't ideal (makes them harder to find and take up more editor space in method signatures and generally not be friendly).  But maybe can revisit that later in another PR!

image

tznind:
Ok all the actions are now events! Except a few in `UICatalog` but since thats not part of the shipped library I think we are ok.

BDisp:
@tznind should be more adequate moving all `Terminal.Gui` events to the `\Terminal.Gui\Core\EventArgs\` folder as well? Here a list of them:
```
\Terminal.Gui\Configuration\ConfigurationManager.cs(307):public class ConfigurationManagerEventArgs : EventArgs {
\Terminal.Gui\Configuration\ThemeScope.cs(158):public class ThemeManagerEventArgs : EventArgs {
\Terminal.Gui\Core\Application.cs(1509):public class ResizedEventArgs : EventArgs {
\Terminal.Gui\Core\CollectionNavigator.cs(50):public class KeystrokeNavigatorEventArgs : EventArgs{
\Terminal.Gui\Core\Toplevel.cs(1056):public class ToplevelClosingEventArgs : EventArgs {
\Terminal.Gui\Core\Trees\SelectionChangedEventArgs.cs(7):public class SelectionChangedEventArgs<T> : EventArgs where T : class {
\Terminal.Gui\Core\View.cs(1337):public class FocusEventArgs : EventArgs {
\Terminal.Gui\Core\View.cs(1696):public class KeyEventEventArgs : EventArgs {
\Terminal.Gui\Core\View.cs(2285):public class LayoutEventArgs : EventArgs {
\Terminal.Gui\Core\View.cs(2843):public class MouseEventArgs : EventArgs {
\Terminal.Gui\Core\Window.cs(350):public class TitleEventArgs : EventArgs {
\Terminal.Gui\Views\DateField.cs(428):public class DateTimeEventArgs<T> : EventArgs {
\Terminal.Gui\Views\HexView.cs(638):public class HexViewEditEventArgs : EventArgs {
\Terminal.Gui\Views\HexView.cs(664):public class HexViewEventArgs : EventArgs {
\Terminal.Gui\Views\ListView.cs(945):public class ListViewItemEventArgs : EventArgs {
\Terminal.Gui\Views\ListView.cs(970):public class ListViewRowEventArgs : EventArgs {
\Terminal.Gui\Views\Menu.cs(2011):public class MenuOpeningEventArgs : EventArgs {
\Terminal.Gui\Views\Menu.cs(2040):public class MenuClosingEventArgs : EventArgs {
\Terminal.Gui\Views\RadioGroup.cs(429):public class SelectedItemChangedArgs : EventArgs {
\Terminal.Gui\Views\TableView.cs(21):public class CellActivatedEventArgs : EventArgs {
\Terminal.Gui\Views\TableView.cs(2007):public class SelectedCellChangedEventArgs : EventArgs {
\Terminal.Gui\Views\TabView.cs(777):public class TabMouseEventArgs : EventArgs {
\Terminal.Gui\Views\TabView.cs(874):public class TabChangedEventArgs : EventArgs {
\Terminal.Gui\Views\TextField.cs(1315):public class TextChangingEventArgs : EventArgs {
\Terminal.Gui\Views\TextField.cs(1337):public class TextChangedEventArgs : EventArgs {
\Terminal.Gui\Views\TextView.cs(526):public class HistoryTextItem : EventArgs{
\Terminal.Gui\Views\TextView.cs(2715):public class ContentsChangedEventArgs : EventArgs {
\Terminal.Gui\Views\TileView.cs(67):public class TitleEventArgs : EventArgs {
\Terminal.Gui\Views\TileView.cs(1113):public class SplitterEventArgs : EventArgs {
\Terminal.Gui\Windows\Wizard.cs(98):public class TitleEventArgs : EventArgs {
\Terminal.Gui\Windows\Wizard.cs(621):public class WizardButtonEventArgs : EventArgs {
\Terminal.Gui\Windows\Wizard.cs(669):public class StepChangeEventArgs : EventArgs {
```

tig:
> @tznind should be more adequate moving all `Terminal.Gui` events to the `\Terminal.Gui\Core\EventArgs\` folder as well? Here a list of them:
> 
> ```
> \Terminal.Gui\Configuration\ConfigurationManager.cs(307):public class ConfigurationManagerEventArgs : EventArgs {
> \Terminal.Gui\Configuration\ThemeScope.cs(158):public class ThemeManagerEventArgs : EventArgs {
> \Terminal.Gui\Core\Application.cs(1509):public class ResizedEventArgs : EventArgs {
> \Terminal.Gui\Core\CollectionNavigator.cs(50):public class KeystrokeNavigatorEventArgs : EventArgs{
> \Terminal.Gui\Core\Toplevel.cs(1056):public class ToplevelClosingEventArgs : EventArgs {
> \Terminal.Gui\Core\Trees\SelectionChangedEventArgs.cs(7):public class SelectionChangedEventArgs<T> : EventArgs where T : class {
> \Terminal.Gui\Core\View.cs(1337):public class FocusEventArgs : EventArgs {
> \Terminal.Gui\Core\View.cs(1696):public class KeyEventEventArgs : EventArgs {
> \Terminal.Gui\Core\View.cs(2285):public class LayoutEventArgs : EventArgs {
> \Terminal.Gui\Core\View.cs(2843):public class MouseEventArgs : EventArgs {
> \Terminal.Gui\Core\Window.cs(350):public class TitleEventArgs : EventArgs {
> \Terminal.Gui\Views\DateField.cs(428):public class DateTimeEventArgs<T> : EventArgs {
> \Terminal.Gui\Views\HexView.cs(638):public class HexViewEditEventArgs : EventArgs {
> \Terminal.Gui\Views\HexView.cs(664):public class HexViewEventArgs : EventArgs {
> \Terminal.Gui\Views\ListView.cs(945):public class ListViewItemEventArgs : EventArgs {
> \Terminal.Gui\Views\ListView.cs(970):public class ListViewRowEventArgs : EventArgs {
> \Terminal.Gui\Views\Menu.cs(2011):public class MenuOpeningEventArgs : EventArgs {
> \Terminal.Gui\Views\Menu.cs(2040):public class MenuClosingEventArgs : EventArgs {
> \Terminal.Gui\Views\RadioGroup.cs(429):public class SelectedItemChangedArgs : EventArgs {
> \Terminal.Gui\Views\TableView.cs(21):public class CellActivatedEventArgs : EventArgs {
> \Terminal.Gui\Views\TableView.cs(2007):public class SelectedCellChangedEventArgs : EventArgs {
> \Terminal.Gui\Views\TabView.cs(777):public class TabMouseEventArgs : EventArgs {
> \Terminal.Gui\Views\TabView.cs(874):public class TabChangedEventArgs : EventArgs {
> \Terminal.Gui\Views\TextField.cs(1315):public class TextChangingEventArgs : EventArgs {
> \Terminal.Gui\Views\TextField.cs(1337):public class TextChangedEventArgs : EventArgs {
> \Terminal.Gui\Views\TextView.cs(526):public class HistoryTextItem : EventArgs{
> \Terminal.Gui\Views\TextView.cs(2715):public class ContentsChangedEventArgs : EventArgs {
> \Terminal.Gui\Views\TileView.cs(67):public class TitleEventArgs : EventArgs {
> \Terminal.Gui\Views\TileView.cs(1113):public class SplitterEventArgs : EventArgs {
> \Terminal.Gui\Windows\Wizard.cs(98):public class TitleEventArgs : EventArgs {
> \Terminal.Gui\Windows\Wizard.cs(621):public class WizardButtonEventArgs : EventArgs {
> \Terminal.Gui\Windows\Wizard.cs(669):public class StepChangeEventArgs : EventArgs {
> ```

Shouldn't the eventargs `.cs` files be next to the class that uses them? E.g.  WizardButtonEventArgs.cs should be in the same folder as Wizard.cs?

BDisp:
> Shouldn't the eventargs `.cs` files be next to the class that uses them? E.g. WizardButtonEventArgs.cs should be in the same folder as Wizard.cs?

Maybe you should be right. Although some of them are common to others (like `Menu` and `ContextMenu`), but majority of them are only used by the class who created it. Now I've a lot of doubt about wat I asked :-(

tznind:
They could be moved to the same folder.

There are 2 aspects to this
- Which physical folder they are in in our repo
- Whether they are nested classes or not (either in main file or in their own file but as a `partial class` of the main `View`

### Which Folder
Normally in a csharp project the folder structure would match the namespaces.  But there are arguments for not doing this (e.g. https://medium.com/@steve_verm/matching-namespaces-to-folders-is-an-anti-pattern-e8edb3a6b8e5 ) 

We have 'Views', 'Types', 'Windows' etc.  I quite like this separation and from a client perspective it doesn't matter as they all share the same namespace.

I created a separate folder for `EventArgs` during this PR so I could see them easily together and work with them.  But I am not wed to this idea and can happily move them into the main 'View'/'Windows' folders as applies.

### Nested Classes
A separate issue is whether to declare event args as nested classes e.g.

```csharp
namespace Terminal.Gui;

public class Wizard{
    public class WizardButtonEventArgs : EventArgs
    {
    }
}
```

Or whether they should just be normal root level classes

```csharp
namespace Terminal.Gui;

public class WizardButtonEventArgs : EventArgs
{
}
```

The [MS guidance on nested classes](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types) is basically to avoid doing it .

Theres a big list of 'DO NOT' in that article.  And the only 'Do' case is around access to access private members of parent class.

From a users perspective events are either used as anonymous lamdas e.g.

```csharp
mybtn.Click += (s,e)=>{ /*do a thing*/}
```

In this case it doesn't really matter whether the class is nested or not.  But often a click handler is used instead e.g.

```csharp
wiz.Finished += WizFinished;

private void WizFinished(object sender, Wizard.WizardButtonEventArgs args)
{
   /*do something*/
}
```

To my mind this is adding extra text to the signature and can be confusing to new developers who might not be familiar with this nested syntax. 

It also makes the `WizardButtonEventArgs`  harder to find for an inexperienced user as its never going to show up in autocomplete in any namespaces until they type the name of an actual class then hit dot.

It also means that if you want to reuse an event args in another/new class you suddenly are referencing a nested class in another control (or have to do a breaking change to move it into root).

### Summary
I propose:

- Moving all nested event args classes out into their own files (and not nested in a parent class)
- Scrap the 'EventArgs' folder and instead put the classes in the main View/Window folder (depending on who they are most closely associated with)

But I'm happy to hear other points of view or go into more detail on my reasoning.

BDisp:
> I propose:
> 
>     * Moving all nested event args classes out into their own files (and not nested in a parent class)
> 
Thanks for the understandable explanation and after read it I agree by moving them out of the parent class.

>     * Scrap the 'EventArgs' folder and instead put the classes in the main View/Window folder (depending on who they are most closely associated with)
> 
I agree. All `*EventArgs.cs` must resides in the same folder of the class that use them. I only want to ask if it's has some advantages to group various `EventArgs` class in the same file, e.g. all menu events args class on a file name `MenuEventArgs.cs`.

> But I'm happy to hear other points of view or go into more detail on my reasoning.

In general I agree on all your thought, but I would like to "hear" your opinion of my suggestion.

tig:
> * Moving all nested event args classes out into their own files (and not nested in a parent class)
> * Scrap the 'EventArgs' folder and instead put the classes in the main View/Window folder (depending on who they are most closely associated with)

I also conur with @tznind's proposal.

On having multiple eventargs classes in aa group as @BDisp asks for `Menu`, I also agree.

Restating our "event args rules":

* Events are defined using the `sender EventArgs` pattern.
* Named as `MyClassEventArgs`
* Separate classes from `MyClass`, not nested classes of `MyClass`.
* If the event can be fired by multiple classes, name it appropriately (e.g. `TitleEventArgs`)
* Put the implementation (`.cs` file) in a separate file from `MyClass.cs`, in the same directory.
* It's ok to put multiple `EventArgs` class implementations in a common `.cs` file if they are all related to a root base-class (e.g. all `Menu` related EventArgs implementations can go in `./Views/MenuEventArgs.cs`.

Correct? If so, let's update this as appropriate:

https://github.com/gui-cs/Terminal.Gui/wiki/Framework-Design-Guidelines

tznind:
Ok all nested events are now in their own files.  Thanks for that list @BDisp was very helpful.

BDisp:
@tznind it seems there was misunderstand on what I said and what @tig understood as well, in this comment:
> I agree. All *EventArgs.cs must resides in the same folder of the class that use them. I only want to ask if it's has some advantages to group various EventArgs class in the same file, e.g. all menu events args class on a file name MenuEventArgs.cs.

So, what I referred was the same what @tig said:
> It's ok to put multiple EventArgs class implementations in a common .cs file if they are all related to a root base-class (e.g. all Menu related EventArgs implementations can go in ./Views/MenuEventArgs.cs.

In this case is put all the `Menu` `EventArgs` into the `MenuEventArgs.cs` where will contain:
```
MenuOpeningEventArgs
MenuClosingEventArgs
etc...
```

By the way you have the `MenuOpeningEventArgs.cs` on two places, in the `Core` and the `Views` folders.

Another issue I' seeing is the `GrabbingMouse` and the `UnGrabbingMouse` isn't present in the` Application.cs` on the current PR, did you revert something?

I hope you don't interpret me badly for these comments.

BDisp:
`EventArgs` in a `public partial class` that isn't a `EventArgs` class, still keep them as nested `EventArgs` class. Only `private partial class` must be allowed. Right?

BDisp:
@tznind I submitted the PR https://github.com/tznind/gui.cs/pull/163 to help a little on your awesome work. Thanks.

BDisp:
@tig, @tznind should we have to create default `EventArgs` for empties `EventHandler` for the cases where eventually we need to have arguments?

BDisp:
> @tig, @tznind should we have to create default `EventArgs` for empties `EventHandler` for the cases where eventually we need to have arguments later without breaking changes?

Well, I think we sure don't need to create a default `EventArgs` for empties `EventHandler`, like was done on `ConfigurationManagerEventArgs`, because we can invoke another `EventArgs` type on the argument parameter. In the subscriber method the `EventArgs` args are the same that was passed by the invoke method and thus there is no breaking change if we invoke another `EventArgs` type. I tested it and it work great. I'm wrong?

tznind:
That's right, Now that we are using `object, EventArgs` we can pass any subclass of `EventArgs` we might create later.

BDisp:
Now the problem is the lot of warnings due the use of `#nullable enable`. Some of them have quick fix but others not. What the best solution for those that doesn't have a quick fix?

tznind:
> Now the problem is the lot of warnings due the use of `#nullable enable`. Some of them have quick fix but others not. What the best solution for those that doesn't have a quick fix?

Hmn thats odd, must have been a bad autocomplete suggestion or a side effect of refactoring command I used to pull out to seperate file.  Seems they started appearing in 2ed71e284

I didn't mean to add it so I will just get rid of them.

tznind:
Ah actually some of them were added by @tig for example in UICatalog in c94f9165f4dd0dbfe7ecf4ee66ffcd32cc75401c

See https://github.com/gui-cs/Terminal.Gui/commit/c94f9165f4dd0dbfe7ecf4ee66ffcd32cc75401c#diff-81646b59a9d4df0e26e5e72c0dddde6f499db0ae941481edcd40326d5ec800eeR17-R18

Maybe this is a work in progress, I had better not go chop them out just yet.

BDisp:
Ah ok we wait for @tig answer. Thanks.

tig:
I'm quite confused about nullable. It just confuses the hell out of me. 

We need a clear and concise set of guidelines for v2. I'm not smart enough (apparently) to author such a thing.

tznind:
> I'm quite confused about nullable. It just confuses the hell out of me.
> 
> We need a clear and concise set of guidelines for v2. I'm not smart enough (apparently) to author such a thing.

### Before NRT
C# has value types and classes (aka reference types).  Value types (int, bool, float, struct) cannot be null.  Classes can be null.

A while ago the `?` operator was added to allow Value types to be null (`int?`, `bool?` etc).  

Classes can be null, so `MyClass blah;`  means a variable of type `Myclass`.  The variable will start out as null and may be assigned values or null later on etc.

You do not need the `#nullable enable` directive to use these features and they are not related to NRT - Nullable Reference Types.

### NRT - Nullable Reference Types
You can turn on NRT for a file with the `#nullable enable` or for the whole project as a `csproj` setting.

Enabling NRT changes `MyClass blah;` from meaning "A variable of type 'MyClass'" to meaning "a variable of type 'MyClass' **which will NEVER BE NULL!".

Obviously when you turn this on you immediately start getting compiler warnings that basically say:

"You told me this variable will never be null but then you assigned it null!"

So you have to manually add a question mark to the class i.e. `MyClass? blah`.  Adding the question mark restores the previous interpretation before you enabled NRT (i.e. "A variable of type 'MyClass'").

You also have to do this for methods e.g. if you return null from a method it needs to have the question mark on the class name returned too e.g. `public MyClass? MaybeGetInstance()`.

This means you end up with questionmarks all over the place to limited value.  You can see this when methods like `string? ToString()`.  That method should almost never return null but since theres some corner cases in some persons library where they do return null then the signature needs the question mark.  

Worse still is that when you call `ToString`, because theres a question mark on the Type the compiler makes the assumption that "it could return null!" which means any variable you assign it to needs to allso allow null so you end up with:

```csharp
string? name = thing.ToString();
```
or (my prefered option)

```csharp
string name = thing.ToString() ?? throw new Exception("ToString method unexpectedly returned null");
```

The first case has the big down side that anyone who references or takes `string? name` later on needs to also be considered "might be null".  The second option is you have a throw which will probably never ever get met and is just there to satisfy compiler warnings.
For more info read here: https://devblogs.microsoft.com/dotnet/embracing-nullable-reference-types/

### When is NRT a good idea

NRT is a great idea when you are first starting a project (Terminal.Gui.Designer has it for example) since it is not too much work when writing code to make informed decisions about whether a class instance should be allowed to be null.

NRT does make you think a lot more about nullability and if you have the spare effort to put in (think 2-3 weeks of solid work for a library our size) then it can reduce the number of null references client libraries end up with.

tznind:
> Well if it's for to have sure that a class never can be null and ensuring that all the consumers are instantiating a non null class, that really a good advantage of NRT. But if it's for end up putting `?` around all the code and acting as non NRT, then is better remove the `#nullable enable`, right?

Yes that is the trade off.

NRT is great for a new codebase.  But when you start getting into NRT in an existing big codebase then it starts to proliferate throughout the codebase.  Despite the ability to enable it per file it still ends up spilling everywhere.  

So you either spend a **lot** of effort doing it properly or you do a half way job (either suppressing the warnings or putting `?` on everything everywhere).  At least that is my experience of it.

But its a big enough task that we should probably make a branch specifically for implementing NRT if that's the way we want to go down rather than just turning it on ad-hoc as part of other work.

tig:
What do y'all think about adopting `INotifyPropertyChanged` in v2?

tznind:
> What do y'all think about adopting `INotifyPropertyChanged` in v2?

I think this is a worthy goal but maybe do it as a seperate PR?

tig:
This is gonna be fun... Not.

image

tig:
Excellent work @tznind !! This is huge.

Terminal.Gui/Core/MouseEventEventArgs.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+
+namespace Terminal.Gui {
+	/// <summary>
+	/// Specifies the event arguments for <see cref="Terminal.Gui.MouseEvent"/>. This is a higher-level construct
+	/// than the wrapped <see cref="MouseEvent"/> class and is used for the events defined on <see cref="View"/>
+	/// and subclasses of View (e.g. <see cref="View.MouseEnter"/> and <see cref="View.MouseClick"/>).
+	/// </summary>
+	public class MouseEventEventArgs : EventArgs {
```
tznind, line 9: Looks like I misnamed this now as `MouseEventEventArgs` instead of `MouseEventArgs`

tznind: Fix is in https://github.com/gui-cs/Terminal.Gui/pull/2435

BDisp: @tznind I did the change on purpose to `MouseEventEventArgs` because it's related with the `MouseEvent` class, for the same reason why there is a `MouseFlagsEventArgs` related with `MouseFlags` enum.

PR #2425, opened 2023-03-16
Fixes#2404. Tabbing to MenuBar and open/close.

### LINKED ISSUE #2404: Cannot Tab to MenuBar

I have been exploring using a MenuBar as a dropdown menu.  It works like a dropdown style combo box and looks better and gives clearer indication of what the active/new selections will be.

But there is a problem in that the user cannot Tab to the view.  I have tried setting the following but it does not work.

```csharp
CanFocus = true,
TabStop = true
```

menu-as-combo

Tab to work for MenuBar when `TabStop` is set.

I can probably still do a workaround by intercepting Tab on the `View` before it  and adding a callback to manually popping the menu open.

I will probably have to do this anyway so that the user can Tab out of the menu and back to other views.

BDisp:
Fixes #2404 - The Leave event is always invoked no matter the menu was closed or not. The menu is always closed on losing focus.

## Pull Request checklist:

Here the code I use to make it work:
```cs
	Application.Init ();

	var win = new Window ("Example App (Ctrl+Q to quit)");

	var mb = new MenuBar (
		new []{new MenuBarItem( new []{
			new MenuItem("Item1",null,()=>{}),
			new MenuItem("Item2",null,()=>{}),
		})}) {
			CanFocus = true,
			TabStop = true,
			X = 10,
			Y = 2,
			Width = 10,
			Text = "Open Me",
	};

	mb.Enter += (e) => {
		if (!mb.IsMenuOpen) {
			mb.OpenMenu ();
		}
	};

	win.Add (new TextField ("SomeText") { Y = 2, Width = 9 });
	win.Add (mb);
	Application.Run (win);
	Application.Shutdown ();

```

### DIFF
--- a/Terminal.Gui/Views/Menu.cs
+++ b/Terminal.Gui/Views/Menu.cs
@@ -1047,6 +1047,7 @@ namespace Terminal.Gui {
 		private void MenuBar_Added (View obj)
 		{
 			_initialCanFocus = CanFocus;
+			Added -= MenuBar_Added;
 		}
 
 		bool openedByAltKey;
--- a/Terminal.Gui/Views/Menu.cs
+++ b/Terminal.Gui/Views/Menu.cs
@@ -808,7 +808,7 @@ namespace Terminal.Gui {
 				}
 			}
 
-			host?.openMenu.SetNeedsDisplay ();
+			host?.openMenu?.SetNeedsDisplay ();
 			host.SetNeedsDisplay ();
 		}
 
@@ -1025,6 +1025,8 @@ namespace Terminal.Gui {
 			WantMousePositionReports = true;
 			IsMenuOpen = false;
 
+			Added += MenuBar_Added;
+
 			// Things this view knows how to do
 			AddCommand (Command.Left, () => { MoveLeft (); return true; });
 			AddCommand (Command.Right, () => { MoveRight (); return true; });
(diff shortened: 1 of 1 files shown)


Terminal.Gui/Views/Menu.cs
```diff
@@ -1025,6 +1025,8 @@ public MenuBar (MenuBarItem [] menus) : base ()
 			WantMousePositionReports = true;
 			IsMenuOpen = false;
 
+			Added += MenuBar_Added;
```
tig, line 1028: I think since `MenuBar_Added` is just a single line *and* you never `-=` the delegate, it's probably better to just do 

```cs
Added += (sender) => _initialCanFocus = CanFocus;
```

???

tig (changes requested):

Terminal.Gui/Views/Menu.cs
tig: I like this better. But this is even better:

```cs
private void MenuBar_Added (View sender)
{
	_prevCanFocus = CanFocus;
	Added -= MenuBar_Added;
}
```

"initial" implies the value is only relevant during initialization. But as coded, it is really "the previous version of CanFocus". Right?

Also, we should used `sender` vs `obj` to be more clear.

tig (approved):

PR #2435, opened 2023-03-20
Fix naming on MouseEventArgs

tznind:
Small fix to the events PR that fixes a bad naming of MouseEventArgs


BDisp:
@tznind I did the change on purpose to `MouseEventEventArgs` because it's related with the `MouseEvent` class, for the same reason why there is a `MouseFlagsEventArgs` related with `MouseFlags` enum. I thought it was more appropriate. Sorry for not warning and avoided this incident.

tznind:
Ok I will close this, although it seems quite confusing naming.  I do see how there is a `MouseEvent` type seperate (and contained within) `MouseEventEventArgs` class.  Maybe we can tidy this up later as a breaking change in v2 to be more elegant.

BDisp:
> Ok I will close this, although it seems quite confusing naming. I do see how there is a MouseEvent type seperate (and contained within) MouseEventEventArgs class.

Your change was already merged in the #2406, so this can be closed. 

> Maybe we can tidy this up later as a breaking change in v2 to be more elegant.

You are right. Sorry for that and thanks for your attention.

PR #2444, opened 2023-03-25
Fixes #1880 - Added properties to ScrollView and ScrollBarView that describe how much lines/columns will be scrolled vertically/horizontally

### LINKED ISSUE #1880: Support specifying the number of lines scrolled by the mouse
No, it would be a feature to specify the lines that will be skipped when using the mouse scroll

Considering that if we open a huge text file, scrolling using the mouse scroll that currently only skips one line at a time is tiring

I thought of it more or less like this:
```C#
var tv = new TextView()
{
    // ...
    VerticalLinesToScroll = 5, // Anyway, any number, the default is 1, as it is currently
    HorizontalLinesToScroll = 3 // Anyway, any number, the default is 1, as it is currently, but in this case, the number would be to use with the command Shift + Mouse Scroll, which comes by default to scroll horizontally in the TextView
};
```

Extra note: If the values ​​are not specified, the values ​​will be the same as those that come by default in a `TextView`

Extra Note 2: Hope I'm not bothering you with so many issues

popara96:
Fixes #1880 - Added properties to ScrollView and ScrollBarView that describe how much lines/columns will be scrolled vertically/horizontally.

## Pull Request checklist:


Terminal.Gui/Views/ScrollView.cs
```diff
@@ -38,29 +38,32 @@ public ContentView (Rect frame) : base (frame)
...
 		ContentView contentView;
 		ScrollBarView vertical, horizontal;
+		int scrollLines, scrollCols;
 
 		/// <summary>
 		///  Initializes a new instance of the <see cref="Gui.ScrollView"/> class using <see cref="LayoutStyle.Absolute"/> positioning.
 		/// </summary>
 		/// <param name="frame"></param>
-		public ScrollView (Rect frame) : base (frame)
+		/// <param name="scrollLines">Scroll amount for vertical scrolling</param>
```
tig, line 47: `scrollLines` should be `scrollRows`.

Is there a known use-case where `scrollLines` is different than `scrollCols`?  Just want to make sure we're not adding complexity if it's not needed.

Terminal.Gui/Views/ScrollView.cs
```diff
@@ -176,6 +179,28 @@ void Initialize (Rect frame)
 			}
 		}
 
+		/// <summary>
+		/// Sets how many lines will be scrolled when scrolling vertically.
+		/// </summary>
+		public int ScrollLines {
```
tig, line 185: `ScrollRows`

Terminal.Gui/Views/ScrollView.cs
```diff
@@ -176,6 +179,28 @@ void Initialize (Rect frame)
 			}
 		}
 
+		/// <summary>
+		/// Sets how many lines will be scrolled when scrolling vertically.
```
tig, line 183: See revised help I posted above.

tig (changes requested):
This PR is great. It's a nice addition to the API.

See my comments embedded.

The new functionality needs unit tests before we can merge it.

Terminal.Gui/Views/ScrollBarView.cs
```diff
@@ -46,25 +47,27 @@ public class ScrollBarView : View {
 		/// <param name="size">The size that this scrollbar represents. Sets the <see cref="Size"/> property.</param>
 		/// <param name="position">The position within this scrollbar. Sets the <see cref="Position"/> property.</param>
 		/// <param name="isVertical">If set to <c>true</c> this is a vertical scrollbar, otherwise, the scrollbar is horizontal. Sets the <see cref="IsVertical"/> property.</param>
-		public ScrollBarView (Rect rect, int size, int position, bool isVertical) : base (rect)
+		/// <param name="scrollAmount">Defines how many lines will be scrolled at a time, defaulting to 1. Sets the <see cref="ScrollAmount"/> property.</param>
+		public ScrollBarView (Rect rect, int size, int position, bool isVertical, int scrollAmount) : base (rect)
```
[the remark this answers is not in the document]
popara96: Got it. I could revert all my additions to constructors (leave them as they were), and default the value to 1, so no constructors are changed but devs can still alter the value via object initialization, if that's okay with you.

Terminal.Gui/Views/ScrollView.cs
popara96: I don't think so, it's just that in `public ScrollBarView (View host, bool isVertical, bool showBothScrollIndicator = true) : this (0, 0, isVertical)` I saw that we're initializing another scroll indicator if `showBothScrollIndicator`, so I wanted to work on giving developers a way to initialize it independently of the first one. Though you're right, it does add a bit of complexity without offering a significant value, so I will probably reduce these two variables to just one `scrollAmount` or similar.

BDisp: One situation that can be justified to have different values for lines and cols is if the size of them are very different. So, on a smaller size we may scroll less than a bigger one. In this case I can propose a percentage amount. Thus the scrolling is proportional to his size. Well only an idea.

tig:
Hey @popara96, did you see the changes requested?

tig:
ping @popara96

popara96:
Sorry to have been absent for a while.
Also need to add a couple of simple unit test for this to be complete.

tig:
`ScrollView` is going away in v2. 

Adding a `ScrollAmount` capability to `View`, given #3323, should be trivial.

If this is really wanted/needed in v1 someone can re-open it.

PR #2447, opened 2023-03-26
Fixes #2446. Currently is allowed two views on different toplevels having HasFocus as true.

### LINKED ISSUE #2446: Currently is allowed two views on different toplevels having HasFocus as true.
That is happens and it's a bug. The `Window` is processing `SetFocus` while the `IsInitialized` property `is false` in `ContentView`. This happens in the case where there are another toplevel in the subviews, causing the second toplevel be focused instead the first focused view.
Another issue is the `Window` itself isn't never set the `HasFocus` property to true, when send focus upwards.

BDisp:
Fixes #2446 - When `Superview` is null wasn't allowing `SetFocus` to true. Also the `ContentView` was processing the focus on the `OnCanFocusChanged` while `IsInitialized` is `false`.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -2141,9 +2141,8 @@ namespace Terminal.Gui {
 				FocusFirst ();
 				return focused != null;
 			}
-			var n = tabIndexes.Count;
 			var focusedIdx = -1;
-			for (var i = 0; i < n; i++) {
+			for (var i = 0; i < tabIndexes.Count; i++) {
 				var w = tabIndexes [i];
 
 				if (w.HasFocus) {
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1677,7 +1677,9 @@ namespace Terminal.Gui {
 			if (focused?.hasFocus == true && focused == view)
 				return;
 			if (focused?.hasFocus == true && focused?.SuperView == view) {
-				view.hasFocus = true;
+				if (!view.hasFocus) {
+					view.hasFocus = true;
+				}
 				return;
 			}
 			// Make sure that this view is a subview
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1037,9 +1037,6 @@ namespace Terminal.Gui {
 			view.tabIndex = -1;
 			SetNeedsLayout ();
 			SetNeedsDisplay ();
-			if (subviews.Count < 1) {
-				CanFocus = false;
-			}
 			foreach (var v in subviews) {
 				if (v.Frame.IntersectsWith (touched))
 					view.SetNeedsDisplay ();
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -1673,7 +1673,9 @@ namespace Terminal.Gui {
 				return;
 			if (focused?.hasFocus == true && focused == view)
 				return;
-			if (focused?.hasFocus == true && focused?.SuperView == view) {
+			if ((focused?.hasFocus == true && focused?.SuperView == view)
+				|| view == this) {
+
 				if (!view.hasFocus) {
 					view.hasFocus = true;
 				}
@@ -1715,7 +1717,11 @@ namespace Terminal.Gui {
 				return;
 			}
 
-			SuperView?.SetFocus (this);
+			if (SuperView != null) {
+				SuperView.SetFocus (this);
+			} else {
+				SetFocus (this);
+			}
 		}
 
 		/// <summary>
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -983,7 +983,7 @@ namespace Terminal.Gui {
 			}
 			SetNeedsLayout ();
 			SetNeedsDisplay ();
-      
+
 			OnAdded (new SuperViewChangedEventArgs (this, view));
 			if (IsInitialized && !view.IsInitialized) {
 				view.BeginInit ();
@@ -1676,7 +1676,10 @@ namespace Terminal.Gui {
 				return;
 			if (focused?.hasFocus == true && focused == view)
 				return;
-
+			if (focused?.hasFocus == true && focused?.SuperView == view) {
+				view.hasFocus = true;
+				return;
+			}
 			// Make sure that this view is a subview
 			View c;
 			for (c = view.container; c != null; c = c.container)
@@ -1694,7 +1697,11 @@ namespace Terminal.Gui {
 			focused.EnsureFocus ();
 
 			// Send focus upwards
-			SuperView?.SetFocus (this);
+			if (SuperView != null) {
+				SuperView.SetFocus (this);
+			} else {
+				SetFocus (this);
+			}
 		}
 
 		/// <summary>
--- a/Terminal.Gui/Core/Window.cs
+++ b/Terminal.Gui/Core/Window.cs
@@ -100,7 +100,7 @@ namespace Terminal.Gui {
 
 			public override void OnCanFocusChanged ()
 			{
-				if (MostFocused == null && CanFocus && Visible) {
+				if (IsInitialized && MostFocused == null && CanFocus && Visible) {
 					EnsureFocus ();
 				}
 
--- a/UnitTests/Core/ViewTests.cs
+++ b/UnitTests/Core/ViewTests.cs
@@ -3026,7 +3026,11 @@ At 0,0
 					Assert.False (view3.HasFocus);
 				}
 			};
-			view2.Leave += (s, e) => Application.Current.Remove (view3);
+			view2.Leave += (s, e) => {
+				Application.Current.Remove (view3);
+				view3.Dispose ();
+				view3 = null;
+			};
 			top2.Add (view2);
 			top1.Add (view1, top2);
 			Application.Begin (top1);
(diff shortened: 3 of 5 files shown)


tig:
@BDisp Help me think this through for v2 with the new View design.

I'm not sure I articulate it clearly in the current spec, but I do articulate these new definitions. If you read between the lines, there's a HUGE impact on how multiple "toplevels" will work in v2:

~~~~~~~~~~~~

* *TopLevel* - The v1 term used to describe a view that is both Modal and can have a MenuBar and/or StatusBar. I propose in v2 we deprecate the term `TopLevel` and instead use `Modal` to describe the same thing. We can completely get rid of the `TopLevel` class! I do not think `Modal` should be a class, but a property of `View` that can be set to `true` or `false`.
* *Window* - A View that, by default, has a `Border` and a `Title`. 
    * QUESTION: Why can't this just be a property on `View` (e.g. `View.Border = true`)? Why do we need a `Window` class at all in v2?
* *Tile*, *Tiled*, *Tiling* (NOT IMPLEMENTED YET) - Refer to a form of `ComputedLayout` where SubViews of a `View` are visually arranged such that they abut each other and do not overlap. In a Tiled view arrangement, Z-ordering only comes into play when a developer intentionally causes views to be aligned such that they overlap. Borders that are drawn between the SubViews can optionally support resizing the SubViews (negating the need for `TileView`).
* *Overlap*, *Overlapped*, *Overlapping* (NOT IMPLEMENTED YET) - Refers to a form of `ComputedLayout` where SubViews of a View are visually arranged such that their Frames overlap. In Overlap view arrangements there is a Z-axis (Z-order) in addition to the X and Y dimension. The Z-order indicates which Views are shown above other views.

~~~~~~~~~~~~

If we can pull this off, a huge amount of logic that deals with multiple top-levels, `Mdi`, etc... goes away (or gets refactored to be far more simple and spread around).

Would you be willing to spend some time helping me design this?

BDisp:
> @BDisp Help me think this through for v2 with the new View design.
> 
> I'm not sure I articulate it clearly in the current spec, but I do articulate these new definitions. If you read between the lines, there's a HUGE impact on how multiple "toplevels" will work in v2:
> 
> ========
> 
>     * _TopLevel_ - The v1 term used to describe a view that is both Modal and can have a MenuBar and/or StatusBar. I propose in v2 we deprecate the term `TopLevel` and instead use `Modal` to describe the same thing. We can completely get rid of the `TopLevel` class! I do not think `Modal` should be a class, but a property of `View` that can be set to `true` or `false`.
> 
If you want to change the `TopLevel` name, like `Container` or whatever, I agree, but not named as `Modal`. Only `Dialog` can be modal. Imagine you have multiple toplevels in `Application`. They may be whatever (`Toplevel`, `Window` or `Dialog`). Only the `Dialog` can be the `Application.Current`. After the `Dialog` is closed, any of the other toplevels can be the `Application.Current`, dependent on how they was implemented. If all of them are `Modal` you are forcing that only a toplevel can be the `Application.Current`.

>     * _Window_ - A View that, by default, has a `Border` and a `Title`.
>       
>       * QUESTION: Why can't this just be a property on `View` (e.g. `View.Border = true`)? Why do we need a `Window` class at all in v2?
> 
I agree. I think the `Title` should be implemented by the `Border` class. Any `View` can have a border and so, by implementation a `Title`. The `ContentView` is superseded with the `Padding` thickness.

>     * _Tile_, _Tiled_, _Tiling_ (NOT IMPLEMENTED YET) - Refer to a form of `ComputedLayout` where SubViews of a `View` are visually arranged such that they abut each other and do not overlap. In a Tiled view arrangement, Z-ordering only comes into play when a developer intentionally causes views to be aligned such that they overlap. Borders that are drawn between the SubViews can optionally support resizing the SubViews (negating the need for `TileView`).
> 
A `Padding` of zero can allowed a subview to overlap the superview. e.g. If subview X and Y are zero they will be draw on the superview (0,0). Another situation is when the `Pos/Dim` are sets in a way the subviews are overlapped on resize shrink.

>     * _Overlap_, _Overlapped_, _Overlapping_ (NOT IMPLEMENTED YET) - Refers to a form of `ComputedLayout` where SubViews of a View are visually arranged such that their Frames overlap. In Overlap view arrangements there is a Z-axis (Z-order) in addition to the X and Y dimension. The Z-order indicates which Views are shown above other views.
> 
This sound like the arranged views features. Great.
> 
> =========
> 
> If we can pull this off, a huge amount of logic that deals with multiple top-levels, `Mdi`, etc... goes away (or gets refactored to be far more simple and spread around).
> 
@tig I think that will be a regression by the reasons I exposed above. You may not want `Mdi` at all, but if a user want to have multi toplevels and want to switch through them without recreating them, they can't.

> Would you be willing to spend some time helping me design this?

Of course. I'm finishing the `ComboBox` and end with WIP that simplify the container management. That is only to help to adapt to your v2 View implementation.

BDisp:
@tig but remember that this PR fixes a real bug which I never realized before. Only the `Superview` and the `Superview.MostFocused` hierarchy can have the `HasFocus` equal to `true`. Any another `Superview` can´t have the `HasFocus` equal to true.

tig:
> If you want to change the `TopLevel` name, like `Container` or whatever, I agree, but not named as `Modal`. Only `Dialog` can be modal. Imagine you have multiple toplevels in `Application`. They may be whatever (`Toplevel`, `Window` or `Dialog`). Only the `Dialog` can be the `Application.Current`. After the `Dialog` is closed, any of the other toplevels can be the `Application.Current`, dependent on how they was implemented. If all of them are `Modal` you are forcing that only a toplevel can be the `Application.Current`.

I now realize I mis-wrote something in the spec. What I *meant* to write is:

* *Modal* - *Modal* - The term used when describing a `View` that was created using the `Application.Run(view)` or `Application.Run<T>` APIs. When a View is running as a modal, user input is restricted to just that View until `Application.Run` exits. A `Modal` View has its own `RunState`. 
    * In v1, classes derived from `Dialog` were originally thought to only work modally. However, `Wizard` proved that a `Dialog`-based class can also work non-modally. 
    * In v2, we will simplify the `Dialog` class, and let any class be run via `Applicaiton.Run`. The `Modal` property will be set by `Application.Run` so the class can detect it is running modally if it needs to. 

  * *TopLevel* - The v1 term used to describe a view that can have a MenuBar and/or StatusBar. In v2, we will delete the `TopLevel` class and ensure ANY View can have a menu bar and/or status bar (via `Adornments`).
    * NOTE: There will still be an `Application.Top` which is the `View` that is the root of the `Application`'s view hierarchy.

tig:
> I agree. I think the Title should be implemented by the Border class. Any View can have a border and so, by implementation a Title. The ContentView is superseded with the Padding thickness.

I don't think you've read or understood my spec. The `Border` class gets deleted. Gone. Killed. 

Its functionality is replaced by the `Adornments` frame (currently called `BorderFrame` in my PR until I can remove `Border`). `Adornments` will have several subviews including `LineView`s that draw that frame and a `Label` that draws the title text. Developers can override this by changing what subviews are added to `Adornments`.

tig:
> A Padding of zero can allowed a subview to overlap the superview. e.g. If subview X and Y are zero they will be draw on the superview (0,0). Another situation is when the Pos/Dim are sets in a way the subviews are overlapped on resize shrink.

I'm not understanding you here.

tig:
> I think that will be a regression by the reasons I exposed above. You may not want Mdi at all, but if a user want to have multi toplevels and want to switch through them without recreating them, they can't.

I am not sure I see your point.

In v2, there is only one `Application.Top`. It is the `View` that is the first view in the application. 

If a developer wants an "MDI" like interface, they can turn on the menubar/status bar in the `Application.Top` and add `Window` subviews to their hearts content. With the `LayoutStyle.Overlapped` each of those subviews overlap each other, support dragging/sizing, etc... They can add/remove subviews to the `Application.Top`, hide/show, etc...  There is no need for a `TopLevel` class.

tig:
> @tig but remember that this PR fixes a real bug which I never realized before. Only the `Superview`, `Superview.Focused` and the `Superview.MostFocused` hierarchy can have the `HasFocus` equal to `true`. Any another `Superview` can´t have the `HasFocus` equal to true.

I get it. I'm just wondering if, once things are simplified and `Toplevel` goes away, will this bug be moot?

BDisp:
> I'm not understanding you here.
>
I was thinking on `Border` and I didn't remember that will be deleted in v2, sorry.

BDisp:
> I get it. I'm just wondering if, once things are simplified and `Toplevel` goes away, will this bug be moot?

I see what your point. Do you can test on your WIP at this time?

tig:
> > I get it. I'm just wondering if, once things are simplified and `Toplevel` goes away, will this bug be moot?
> 
> I see what your point. Do you can test on your WIP at this time?

Sadly, no. I'm getting closer, but I keep finding unit tests that are either:

- breaking for real because I broke something
- breaking because they were poorly written and my refactor is enforcing proper behavior

So it will take me a few more days. 

In the meantime, please look at the Project, especially the Backlog column. Those items can all be worked on.

https://github.com/orgs/gui-cs/projects/1/views/1

BDisp:
@tig the changes on the `View` class will be needed normalize better his behavior.

BDisp:
The fix on the `FocusNext` method is very important because if a view is added/removed while the focus process is running, the `tabIndexes` collection may have changed and his enumerationsaved on a variable which may is out of date.

tig (approved):

PR #2467, opened 2023-03-30
Fixes #2465 - Simplify/Enforce LayoutStyle.Absolute when X, Y, Width, and/or Height are null or Absolute()

### LINKED ISSUE #2465: Simplify/Enforce `LayoutStyle.Absolute` when `X`, `Y`, `Width`, and/or `Height` are null or `Absolute()`
I keep finding latent bugs and/or extra code in v1 relative to this topic.

The current API docs for `ComputedLayout` read:

image

IOW, these members are to be ignored if `Absolute` is set. However, in reality, nobody keeps this clear in their heads and there's tons of code where Views with `Absolute` get these members set using absolute values (e.g. `view.X = 3`). 

Plus, `Dim` and `Pos` make this easy with their constructors that take `int`. 

I propose the following for v2 to simplify and clarify all of this: `View.LayoutStyle` is no longer implemented as a field (the `View.layoutStyle` private member), but a calculated property.

v1:
```cs
		public LayoutStyle LayoutStyle {
			get {
				return layoutStyle;
			}
			set {
				layoutStyle = value;
				SetNeedsLayout ();
			}
		}
```

v2:
```cs
		/// <summary>
		/// Gets or sets what <see cref="Terminal.Gui.LayoutStyle"/> the view is using. 
		/// </summary>
		/// <remarks>
		/// <para>
		/// Setting this property to <see cref="Terminal.Gui.LayoutStyle.Absolute"/> will cause <see cref="Frame"/> to determine the size and position of the view. <see cref="X"/> and <see cref="Y"/> will be 
		/// set to <see cref="Dim.DimAbsolute"/> using <see cref="Frame"/>.
		/// </para>
		/// <para>
		/// Setting this property to <see cref="Terminal.Gui.LayoutStyle.Computed"/> will cause the view to use the <see cref="LayoutSubviews"/> method to 
		/// size and position of the view. If either of the <see cref="X"/> and <see cref="Y"/> properties are `null` they will be set to <see cref="Pos.PosAbsolute"/> using
		/// the current value of <see cref="Frame"/>. 
		/// If either of the <see cref="Width"/> and <see cref="Height"/> properties are `null` they will be set to <see cref="Dim.DimAbsolute"/> using <see cref="Frame"/>.
		/// </para>
		/// </remarks>
		/// <value>The layout style.</value>
		public LayoutStyle LayoutStyle {
			get {
				if ((X == null || X is Pos.PosAbsolute) &&  (Y == null || Y is Pos.PosAbsolute) && 
					(Width == null || Width is Dim.DimAbsolute) && (Height == null || Height is Dim.DimAbsolute)) {
					return LayoutStyle.Absolute;
				}
				return LayoutStyle.Computed;
			}
			set {
				switch (value) {
				case LayoutStyle.Absolute:
					X = Frame.X;
					Y = Frame.Y;
					Width = Frame.Width;
					Height = Frame.Height;
					break;
					
				case LayoutStyle.Computed:
					if (X == null) X = Frame.X;
					if (Y == null) Y = Frame.Y;
					if (Width == null) Width = Frame.Width;
					if (Height == null) Height = Frame.Height;
					break;
				}

				SetNeedsLayout ();
			}
		}
```

The implications of this are fairly straightforward, but there are sure to be some subtle gotchas in existing code and architecture. 

- I think once this is done, there's no need for X, Y, Width, or Height to ever be `null`. They are either of type Pos/Dim.Absolute or another. Their setters should throw `InvalidArgumentException` if `value` is `null`.

Thoughts? Comments?

tig:
Fixes #2465

### DIFF
--- a/Terminal.Gui/View/Layout/ViewLayout.cs
+++ b/Terminal.Gui/View/Layout/ViewLayout.cs
@@ -293,7 +293,6 @@ public partial class View {
 #if DEBUG
 			if (LayoutStyle == LayoutStyle.Computed && !IsInitialized) {
 				Debug.WriteLine ($"WARNING: Bounds is being accessed before the View has been initialized. This is likely a bug in {this}");
-				Debug.WriteLine ($"The Frame is set before the View has been initialized. So it isn't a bug.Is by design.");
 			}
 #endif // DEBUG
 			//var frameRelativeBounds = Padding?.Thickness.GetInside (Padding.Frame) ?? new Rect (default, Frame.Size);
@@ -1102,8 +1101,6 @@ public partial class View {
 				Width = newFrameSize.Width;
 			}
 		}
-		// BUGBUG: This call may be redundant
-		TextFormatter.Size = GetTextFormatterSizeNeededForTextAndHotKey ();
 		return boundsChanged;
 	}
 
--- a/Terminal.Gui/View/Layout/ViewLayout.cs
+++ b/Terminal.Gui/View/Layout/ViewLayout.cs
@@ -30,22 +30,31 @@ public partial class View {
 	Rect _frame;
 
 	/// <summary>
-	/// Gets or sets the frame for the view. The frame is relative to the <see cref="SuperView"/>'s <see cref="Bounds"/>.
+	/// Gets or sets location and size of the view. The frame is relative to the <see cref="SuperView"/>'s <see cref="Bounds"/>.
 	/// </summary>
-	/// <value>The frame.</value>
+	/// <value>The rectangle describing the location and size of the view, in coordinates relative to the <see cref="SuperView"/>.</value>
 	/// <remarks>
 	/// <para>
-	///    Change the Frame when using the <see cref="Terminal.Gui.LayoutStyle.Absolute"/> layout style to move or resize views. 
+	/// Change the Frame when using the <see cref="LayoutStyle.Absolute"/> layout style to move or resize views.
 	/// </para>
 	/// <para>
-	///    Altering the Frame of a view will trigger the redrawing of the
-	///    view as well as the redrawing of the affected regions of the <see cref="SuperView"/>.
+	/// Altering the Frame will change <see cref="LayoutStyle"/> to <see cref="LayoutStyle.Absolute"/>.
+	/// Additionally, <see cref="X"/>, <see cref="Y"/>, <see cref="Width"/>, and <see cref="Height"/> will be set
+	/// to the values of the Frame (using <see cref="Pos.PosAbsolute"/> and <see cref="Dim.DimAbsolute"/>).
+	/// </para>
+	/// <para>
+	/// Altering the Frame will eventually (when the view is next drawn) cause the <see cref="LayoutSubview(View, Rect)"/>
+	/// and <see cref="OnDrawContent(Rect)"/> methods to be called.
 	/// </para>
 	/// </remarks>
 	public virtual Rect Frame {
 		get => _frame;
 		set {
 			_frame = new Rect (value.X, value.Y, Math.Max (value.Width, 0), Math.Max (value.Height, 0));
+			//X = _frame.X;
+			//Y = _frame.Y;
+			//Width = _frame.Width;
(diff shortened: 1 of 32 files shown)


Terminal.Gui/Core/View.cs
```diff
@@ -446,70 +446,102 @@ void SetTabIndex ()
...
 		/// </para>
 		/// </remarks>
 		public virtual Rect Frame {
 			get => frame;
 			set {
 				frame = new Rect (value.X, value.Y, Math.Max (value.Width, 0), Math.Max (value.Height, 0));
+				X = frame.X;
+				Y = frame.Y;
+				Width = frame.Width;
+				Height = frame.Height;
```
BDisp, line 472: You are forcing the Pos/Dim to be `Absolute` even they aren't. So, in my opinion the `Frame` setter mustn't set the Pos/Dim values.

Terminal.Gui/Core/View.cs
```diff
@@ -446,70 +446,102 @@ void SetTabIndex ()
...
+					Y = Frame.Y;
+					Width = Frame.Width;
+					Height = Frame.Height;
+					break;
+					
+				case LayoutStyle.Computed:
+					if (X == null) X = Frame.X;
+					if (Y == null) Y = Frame.Y;
+					if (Width == null) Width = Frame.Width;
+					if (Height == null) Height = Frame.Height;
```
BDisp, line 518: Can be replaced by:
```cs
X ??= Frame.X;
Y ??= Frame.Y;
Width ??= Frame.Width;
Height ??= Frame.Height;
```
Additionally in each` Pos/Dim (X, Y, Width, Height)` setter you after must update the `_layoutStyle` field by get the `LayoutStyle` value:
`
_layoutStyle = LayoutStyle;
`

BDisp (changes requested):

tig:
@BDisp and @tznind - I've finally come back to this PR and:

- [x] Merged with v2_develop
- [x] Updated API docs to be more complete and accurate
- [x] Fixed all Views that were using `Bounds` or calling `LayoutSubViews` before `IsInitalized`
- [x] Fixed a few other things (esp in ComboBox) that needed fixing

What I *didn't* do:

- [ ] Actually enforce LayoutStyle as was originally proposed here. I'd love your thoughts on that now.

Please review asap. Thanks!

BDisp:
Something is wrong with `All Views Tester`. It starting at (0,0) but navigate down and then up the views don't start at (0,0).

tig:
> Something is wrong with `All Views Tester`. It starting at (0,0) but navigate down and then up the views don't start at (0,0).

Ah, with mouse. With keyboard it's fine. On it...

BDisp:
> > Something is wrong with `All Views Tester`. It starting at (0,0) but navigate down and then up the views don't start at (0,0).
> 
> Ah, with mouse. With keyboard it's fine. On it...

With the keyboard. Navigate downward until `Dialog` and then navigate upward and you will see that the views aren't at (0,0) anymore.

tig:
> > > Something is wrong with `All Views Tester`. It starting at (0,0) but navigate down and then up the views don't start at (0,0).
> > 
> > 
> > Ah, with mouse. With keyboard it's fine. On it...
> 
> With the keyboard. Navigate downward until `Dialog` and then navigate upward and you will see that the views aren't at (0,0) anymore.
> 
> **Edit:** `Dialog` is centered horizontal and vertical and thus it seems the upper left coordinate is changed for all views.

Fixed.

BDisp:
Some of the views aren't show, like the `TextField`.

tig:
> Some of the views aren't show, like the `TextField`.

Fixed. ;-)

tig:
@BDisp I'd like to merge this so it can be picked up in https://github.com/gui-cs/Terminal.Gui/pull/3078 and https://github.com/BDisp/Terminal.Gui/pull/172 so I can more easily debug (this PR removes a ton of debug spew).

Are you good with my changes here?

BDisp:
Done @tig. Thanks.

PR #2468, opened 2023-03-30
Adds AbsoluteLayout unit tests

tig:
We were missing some fundamental unit tests (a lot of this was covered in other unit tests, but in a haphazard way). 

I plan on writing a similar suite for ComputedLayout.


tig:
See related: #2465

PR #2479, opened 2023-04-01
Fixes #2478. Currently it isn't possible to draw over a modal view.

### LINKED ISSUE #2478: Currently it isn't possible to draw over a modal view.
Why this is needed? Because we may need to popup something on a dialog that can spillover the `Dialog` bounds, like a combo drop-down list that doesn't fit his dimension. For this reason it isn't possible to display the `ContextMenu` of a `TextField` on a `Dialog`, which has to refactored to force draw after all the run loop processes are completed.

There is some constrains to achieve this feature:
- Application.Top is always redrawing if  it isn't need, because it isn't cleaning the `LayoutNeeded ` and the `NeedDisplay`.
- There is a lack of an event that could be invoked after all processes all completed.
- `SetSubViewNeedsDisplay` is recursively setting the `ChildNeedsDisplay` and the `container.ChildNeedsDisplay` to `true` even if they are already set as `true` wasting time in the process.

BDisp:
Fixes #2478 - Added `IterationComplete` action which do the trick. The only doubt is if it also work if two or more views want to use it as well.

## Pull Request checklist:


BDisp:
@tig, @tznind the cause of this issue is the override by toplevels that doesn't call the `OnDrawContentComplete` after the contents are drawn. Should be the `Application.Current` who have to call the `OnDrawContentComplete` after the `state.Toplevel.Redraw (state.Toplevel.Bounds);`. In this case the `IterationComplete` action isn't needed. The suggestion change will be:

```cs
			if (!state.Toplevel.NeedDisplay.IsEmpty || state.Toplevel.ChildNeedsDisplay || state.Toplevel.LayoutNeeded
				|| MdiChildNeedsDisplay ()) {
				state.Toplevel.Redraw (state.Toplevel.Bounds);
				state.Toplevel.OnDrawContentComplete (state.Toplevel.Bounds);
				if (DebugDrawBounds) {
					DrawBounds (state.Toplevel);
				}
				state.Toplevel.PositionCursor ();
				Driver.Refresh ();
			} else {
				Driver.UpdateCursor ();
			}

```

BDisp:
Here is how it looks:

combobox-contextmenu

BDisp:
It's only with me this happens or someone else complain the behavior on after run a unit test and all completed, when I copy something to the clipboard and I paste it still having junk from the `Clipboard` unit tests, even I didn't ran any of his tests? Before this behavior only was happening while the unit test are running, but not after they are finished and than I copy some other selection to the `Clipboard`. It seems that some background work is still running even when there aren't any running unit tests.

Terminal.Gui/Core/Application.cs
```diff
@@ -277,6 +277,14 @@ static void OnQuitKeyChanged (KeyChangedEventArgs e)
 		/// </remarks>
 		public static Action Iteration;
 
+		/// <summary>
+		///   This event is raised on each iteration of the <see cref="MainLoop"/> after all processes are completed.
+		/// </summary>
+		/// <remarks>
+		///   See also <see cref="Timeout"/>
+		/// </remarks>
+		public static Action IterationComplete;
```
tig, line 286: Nothing wrong with this as a new primitive event, but would it be better for this PR if there were an event that was draw related?

In v2, i'd like us to revisit how `Redraw` and the related events are structured. I started this in my View2 PR...

- `Redraw` is mis-named. It should just be `Draw` (it also shouldn't take a `Bounds` because implementors can just use `this.Bounds`; this is a cause of a bunch of bugs BTW).
- `Draw` should simply call
  - `OnDraw`
  - `OnDrawFrames`
  - `OnDrawContent`
  - `OnDrawContentComplete`

Each of those invokes an event and each is cancellable (except OnDrawContentComplete).

With this, your tests could subscribe to `DrawContentComplete` instead of `IterationComplete`?

tig (changes requested):

Terminal.Gui/Core/Application.cs
BDisp: That the point @tig I mentioned in this comment https://github.com/gui-cs/Terminal.Gui/pull/2479#issuecomment-1493371057. It's almost have this ready, only finishing some unit tests. The `IterationComplete` will be removed. Only the `DrawContentComplete` is more accurate to draw anything in the current run state toplevel, because some toplevels consume the `Redraw` method without calling the `OnDrawContentComplete` method later,

BDisp:
> In v2, i'd like us to revisit how `Redraw` and the related events are structured. I started this in my View2 PR...
> 
>     * `Redraw` is mis-named. It should just be `Draw` (it also shouldn't take a `Bounds` because implementors can just use `this.Bounds`; this is a cause of a bunch of bugs BTW).
> 
>     * `Draw` should simply call
>       
>       * `OnDraw`
>       * `OnDrawFrames`
>       * `OnDrawContent`
>       * `OnDrawContentComplete`
> 
> 
> Each of those invokes an event and each is cancellable (except OnDrawContentComplete).

@tig I agree with this scenario, but `Draw` mustn't be virtual to avoid derived class not calling the  `OnDraw`, `OnDrawFrames`, `OnDrawContent` and `OnDrawContentComplete`. Maybe `protected`. What do you think?

tig:
> > In v2, i'd like us to revisit how `Redraw` and the related events are structured. I started this in my View2 PR...
> > ```
> > * `Redraw` is mis-named. It should just be `Draw` (it also shouldn't take a `Bounds` because implementors can just use `this.Bounds`; this is a cause of a bunch of bugs BTW).
> > 
> > * `Draw` should simply call
> >   
> >   * `OnDraw`
> >   * `OnDrawFrames`
> >   * `OnDrawContent`
> >   * `OnDrawContentComplete`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Each of those invokes an event and each is cancellable (except OnDrawContentComplete).
> 
> @tig I agree with this scenario, but `Draw` mustn't be virtual to avoid derived class not calling the `OnDraw`, `OnDrawFrames`, `OnDrawContent` and `OnDrawContentComplete`. Maybe `protected`. What do you think?

Agree. I'm going to do this AFTER my big PR. New issue: https://github.com/gui-cs/Terminal.Gui/issues/2482

Terminal.Gui/Core/View.cs
```diff
@@ -3155,5 +3158,45 @@ public View GetTopSuperView (View view = null, View superview = null)
 
 			return top;
 		}
+
+		/// <summary>
+		/// Finds which view that belong to the <paramref name="start"/> at the provided location.
```
tig, line 3163: I believe all coordiantes are  elative to `start`'s superview, right? If that's correct please add this detail to the API docs. If not please add the right verbiage.

Terminal.Gui/Core/View.cs
```diff
@@ -3155,5 +3158,45 @@ public View GetTopSuperView (View view = null, View superview = null)
...
+		/// </summary>
+		/// <param name="start">The view where to look for.</param>
+		/// <param name="x">The column location.</param>
```
tig, line 3166: The column location of WHAT?

Terminal.Gui/Core/View.cs
```diff
@@ -3155,5 +3158,45 @@ public View GetTopSuperView (View view = null, View superview = null)
...
+		/// <param name="y">The row location.</param>
+		/// <param name="resx">The found view column location.</param>
+		/// <param name="resy">The found view row location.</param>
+		/// <returns>The view that belong to the provided location.</returns>
```
tig, line 3170: "The view that was found at the <praramref name"x"/> and <praramref name"x"/> coordinates. `null` if no view was found."

UnitTests/Menus/ContextMenuTests.cs
```diff
@@ -902,5 +902,85 @@ public void Key_Open_And_Close_The_ContextMenu ()
...
+		{
+			var top = Application.Top;
+			var win = new Window ("Window");
+			top.Add (win);
+			Application.Begin (top);
+			((FakeDriver)Application.Driver).SetBufferSize (20, 15);
+
+			Assert.Equal (new Rect (0, 0, 20, 15), win.Frame);
+			TestHelpers.AssertDriverContentsWithFrameAre (@"
+┌ Window ──────────┐
```
tig, line 917: Please don't add "Titles" to unit tests unless strictly necessary. This makes porting tests to new title/border styles harder.

UnitTests/Menus/ContextMenuTests.cs
```diff
@@ -902,5 +902,85 @@ public void Key_Open_And_Close_The_ContextMenu ()
...
+│                  │
+│                  │
+│                  │
+│                  │
+│                  │
+│                  │
+│                  │
+└──────────────────┘", output);
+
+			var dialog = new Dialog ("Dialog") { X = 2, Y = 2, Width = 15, Height = 4 };
```
tig, line 933: Same. Strictly, there's no need for a dialog title for this test to prove things are working. Just remove "Dialog" from the constructor.

tig (changes requested):

BDisp:
Done @tig. If you find some more incorrect stuff, please don't hesitate to tell me. Thanks.

BDisp:
@tig I'm seeing here a critical change which affect the way how the frame really work. Each view frame location is always zero based and can be negative. The unit test in this PR worked before but not now with the new v2 frame. In the unit test the `Window` has a `Width = 20` and a `Height = 20`. The `Dialog` has this frame (`{Dialog(2)({X=2,Y=5,Width=15,Height=10})}`). The "Popup" should have `{X=3,Y=8,Width=9,Height=1}`, but it have in the v2 `{Button(Popup)({X=2,Y=7,Width=9,Height=1})}`. See bellow how I get the values. Is this right and is how it work now and I have to change the test to fit the changes?
```
 01234567890123456789
0┌──────────────────┐0
1│                  │1
2│                  │2
3│                  │3
4│ 012345678901234  │4
5│0┌─────────────┐0 │5
6│1│             │1 │6
7│2│             │2 │7
8│3│             │3 │8
9│4│             │4 │9
0│5│             │5 │0
1│6│             │6 │1
2│7│             │7 │2
3│8│  [ Popup ]  │8 │3
4│9└─────────────┘9 │4
5│ 012345678901234  │5
6│                  │6
7│                  │7
8│                  │8
9└──────────────────┘9
 01234567890123456789
```

tig:
> @tig I'm seeing here a critical change which affect the way how the frame really work. Each view frame location is always zero based and can be negative. The unit test in this PR worked before but not now with the new v2 frame. In the unit test the `Window` has a `Width = 20` and a `Height = 20`. The `Dialog` has this frame (`{Dialog(2)({X=2,Y=5,Width=15,Height=10})}`). The "Popup" should have `{X=3,Y=8,Width=9,Height=1}`, but it have in the v2 `{Button(Popup)({X=2,Y=7,Width=9,Height=1})}`. See bellow how I get the values. Is this right and is how it work now and I have to change the test to fit the changes?

I'm not following. 

See the `DialogTests.cs` I did have to tweak some button alignment tests; but I think the code is correct based on how the tests look now. Happy to be proven wrong!

BDisp:
The distance from 0 to 2 is the same as 0 to -2, only differ in the direction.

tig:
> How you can have negative location if you don't consider the zero base location? A location X with -2 must be before the 0. If you only consider the zero base on `Application.Top` the `FindDeepestView` will get the wrong location for a view. For this reason the location coordinates for the "Popup" button wasn't found in my test. See bellow my clarification.

I still don't understand. Can you please write a simple unit test that illustrates against the current code?

> why did you disabled the frame ruler and the frame padding from the `Diagnostics` scenario? If this is to maintain, working with scroll bars will be different.

It's a short term bug. Just opened a new issue to track it #2496

BDisp:
Well I already seen how you tweaked the `FindDeepestView` by using the `GetBoundsOffset` method. I'm still prefer the correct frame locations that appear on the screen, like above. The tweak should be used when is draw to the screen. But it's just my opinion.

tig:
> Well I already seen how you tweaked the `FindDeepestView` by using the `GetBoundsOffset` method. I'm still prefer the correct frame locations that appear on the screen, like above. The tweak should be used when is draw to the screen. But it's just my opinion.

`Frame` is super-view relative.
`Bounds` is relative to the view's `Frame`, offset by `GetBoundsOffset()`.

BDisp:
> `Frame` is super-view relative.

@tig if `Frame` is super-view relative why it isn't `{Button(Popup)({X=3,Y=8,Width=9,Height=1})}` as I show above in the first frame rule?

> `Bounds` is relative to the view's `Frame`, offset by `GetBoundsOffset()`.

Right, is the `Bounds` who must get the frame `{X=3,Y=8,Width=9,Height=1}` and call the `GetBoundsOffset()` to tweak his position on the screen and in this case, the `FindDeepestView` method would work as before without calling the `GetBoundsOffset()`.

tig:
> > `Frame` is super-view relative.
> 
> @tig if `Frame` is super-view relative why it isn't `{Button(Popup)({X=3,Y=8,Width=9,Height=1})}` as I show above in the first frame rule?

I have no idea, because you didn't post enough info for me to understand. 

I mis-typed above. `myview.Frame.Location` is relative to `myview.SuperView.Bounds.Location`. This was true in v1.

In v1 `myview.SuperView.Bounds.Location` was always `0, 0` and was tied to `myView.Frame.Location`.

In v2, ``myview.SuperView.Bounds.Location` is `0, 0` (for now; see #2489) and is tied to `myView.Frame.Location` + `myView.GetBoundsOffset()`.

BDisp:
> I have no idea, because you didn't post enough info for me to understand.

frame


Run the `Draw_A_Top_Subview_On_A_Dialog` unit test in this PR and you will see:
`Dialog(2)({X=2,Y=5,Width=15,Height=10})}` That right.
`{Button(Popup)({X=2,Y=7,Width=9,Height=1})}` For me is wrong. It should be `{Button(Popup)({X=3,Y=8,Width=9,Height=1})}` as is shown in the image.

BDisp:
@tig I think you don't understood what I mean with all this. I know the test is passing now because you tweaked the `FindDeepestView` to get the button at the coordinate X=8 and Y=3. My point is that the frame in the button has the wrong frame. On v1 the `Frame` is the screen location for all views. Although is working as you tweaked I think the `myView.Frame` should be the correct position on the screen as the image is showing. But I see I'm talking for nothing.

tig:
> @tig I think you don't understood what I mean with all this. I know the test is passing now because you tweaked the `FindDeepestView` to get the button at the coordinate X=8 and Y=3. My point is that the frame in the button has the wrong frame. On v1 the `Frame` is the screen location for all views. Although is working as you tweaked I think the `myView.Frame` should be the correct position on the screen as the image is showing. But I see I'm talking for nothing.

Please don't get upset. I'm happy to be proven wrong here (really!) but I don't think I understand what you are saying.

However, "On v1 the `Frame` is the *screen* location for all views." is not accurate.

The v1 `View.Frame` documentation reads:

image

The correct statement is "In v1 the `Frame` is the *superview* location for all views."

BDisp:
> Please don't get upset. I'm happy to be proven wrong here (really!) but I don't think I understand what you are saying.
> 
Sorry I'm not. But you are surely wrong here. If you look at the frame rule the `btnPopup` is on `X=3, Y=8`, but the `btnPopup.Frame.Location` is `X=2, Y=7`. I think I found the cause of this working in a wrong way. You are using the `Thickness.Horizontal/Vertical` and you don't need it. This was needed for v1 because the border rectangle was inside the padding and now it's outside. I'm doing the changes to prove to you.

> However, "On v1 the `Frame` is the _screen_ location for all views." is not accurate.
> 
> The v1 `View.Frame` documentation reads:
> 
> image
> 
Yes is relative to the view container, but it's related to the `SuperView` `Application.Top`.

> The correct statement is "In v1 the `Frame` is the _superview_ location for all views."

That is the correct expression and should be the same for the v2 as well.

I hope you finally understood me. Thanks and sorry for something.

BDisp:
@tig, sorry I updated my comment. In In short my only doubt is why the new `myView.Frame` isn't a zero based location.

tig:
> @tig, sorry I updated my comment. In short my only doubt is why the new `myView.Frame` isn't a zero based location.

Can you explain what "zero based location" means to you because I'm confused by what you mean.

This is a bug: `If you look at the frame rule the btnPopup is on X=3, Y=8, but the btnPopup.Frame.Location is X=2, Y=7.`

The new `AbsoluteLayoutTests.cs` are *supposed* to prove that what you show above can't happen.

Please note the "BUGBUG:" comments in it!

`AbsoluteLayout_Change_Frame` and others are intended to prove that `X`/`Y` always track `Frame. Location` and `Height/Width` always track `Frame.Size`. 

Likewise `AbsoluteLayout_Change_X_or_Y_Absolute` and others are intended to prove that `Frame. Location` tracks `X`/`Y` etc...

I was planning on adding a `ComputedLayoutTests.cs` that proves similar things for computed layout but I haven't gotten to it.

Perhaps you can help by 

1) Reviewing `AbsoluteLayoutTests` to ensure I've tested all the scenarios correctly.
2) Creating a `ComputedLayoutTests.cs` from a copy of `AbsoluteLayoutTests.cs`  and have it focus on similar primitve tests.

BDisp:
@tig you are absolutely right. I didn't take an account the padding frame where the button is added and in this case the frame is correct `(X=2, Y=7)`. I'm very sorry for my mistake. The zero base I was referring is the location begin from zero. But it really work as it's expected. Really sorry. The follow redesign explain better.

```
 01234567890123456789
0┌──────────────────┐0
1│                  │1
2│                  │2
3│                  │3
4│ 012345678901234  │4
5│0┌─────────────┐0 │5
6│1│0123456789012│1 │6
7│2│1            │2 │7
8│3│2            │3 │8
9│4│3            │4 │9
0│5│4            │5 │0
1│6│5            │6 │1
2│7│6            │7 │2
3│8│7 [ Popup ]  │8 │3
4│9└─────────────┘9 │4
5│ 012345678901234  │5
6│                  │6
7│                  │7
8│                  │8
9└──────────────────┘9
 01234567890123456789
```

tig (approved):

PR #2504, opened 2023-04-05
Fixes #2503. PosCombine and DimCombine is wrongly looking into the from.InternalSubviews.

### LINKED ISSUE #2503: PosCombine and DimCombine is wrongly looking into the from.InternalSubviews.
The `PosCombine` and `DimCombine` only have to get the left and right sides.

BDisp:
Fixes #2503 - This was causing throws errors in layouts where a subview was be referencing to superview of his superview.

## Pull Request checklist:


Terminal.Gui/View/View.cs
```diff
@@ -2657,16 +2650,27 @@ internal static List<View> TopologicalSort (View superView, IEnumerable<View> no
...
+							result.Add (to);
+						}
+						edges.Remove ((from, to));
+					} else if (from != superView?.GetTopSuperView (to, from) && !ReferenceEquals (from, to)) {
 						if (ReferenceEquals (from.SuperView, to)) {
 							throw new InvalidOperationException ($"ComputedLayout for \"{superView}\": \"{to}\" references a SubView (\"{from}\").");
 						} else {
 							throw new InvalidOperationException ($"ComputedLayout for \"{superView}\": \"{from}\" linked with \"{to}\" was not found. Did you forget to add it to {superView}?");
 						}
-					} else {
```
tig, line 2668: Is it not possible to detect a recursive cycle?

Or, is a recursive cycle impossible?

Or, is it not a bad thing to have a recursive cycle?

Terminal.Gui/Views/ScrollView.cs
```diff
@@ -385,6 +385,7 @@ public override void Redraw (Rect bounds)
 
 			// Fill in the bottom left corner
 			// BUGBUG: ScrollBarView should be responsible for this via contentBottomRightCorner
+			// this is only true if this is hosted by the ScrollBarView and it isn't
```
tig, line 388: Got it now. Thanks for clarifying. Please update the comment to read.

// Fill in the bottom left corner. Note we don't rely on ScrollBarView.contentBottomRightCorner here 
// because that only applies when `ScrollBarView` is hosted.

UICatalog/Scenarios/Scrolling.cs
```diff
@@ -155,9 +155,9 @@ public override void Setup ()
 			void Top_Loaded (object sender, EventArgs args)
 			{
 				// BUGBUG: v2 - this broke somehow - fix later
-				//horizontalRuler.Text = rule.Repeat ((int)Math.Ceiling ((double)(horizontalRuler.Bounds.Width) / (double)rule.Length)) [0..(horizontalRuler.Bounds.Width)] +
-				//	"\n" + "|         ".Repeat ((int)Math.Ceiling ((double)(horizontalRuler.Bounds.Width) / (double)rule.Length)) [0..(horizontalRuler.Bounds.Width)];
-				//verticalRuler.Text = vrule.Repeat ((int)Math.Ceiling ((double)(verticalRuler.Bounds.Height * 2) / (double)rule.Length)) [0..(verticalRuler.Bounds.Height * 2)];
+				horizontalRuler.Text = rule.Repeat ((int)Math.Ceiling ((double)(horizontalRuler.Bounds.Width) / (double)rule.Length)) [0..(horizontalRuler.Bounds.Width)] +
```
tig, line 158: Let's use the new `Ruler` class here.

UnitTests/View/Layout/DimTests.cs
```diff
@@ -1266,5 +1307,49 @@ public void DimPercentPlusOne (int startingDistance, bool testHorizontal)
 				Assert.Equal (51, label.Frame.Height);
 			}
 		}
+
+		[Fact]
+		public void Dim_Referencing_SuperView_Does_Not_Throws ()
```
tig, line 1312: "Throws" should be singular here ("Throw").

UnitTests/View/Layout/LayoutTests.cs
```diff
@@ -43,7 +43,9 @@ public void TopologicalSort_Recursive_Ref ()
 			var sub2 = new View ();
 			root.Add (sub2);
 			sub2.Width = Dim.Width (sub2);
-			Assert.Throws<InvalidOperationException> (() => root.LayoutSubviews ());
+
+			var exception = Record.Exception (root.LayoutSubviews);
```
tig, line 47: I've found debugging unit tests that use `Record.Exception` is extra hard. 

Is there a way to refactor tests like this to not use it?

UnitTests/View/Layout/LayoutTests.cs
```diff
@@ -1805,7 +1807,57 @@ public void Dim_CenteredSubView_85_Percent_Width (int width)
 				break;
 			}
 			_ = TestHelpers.AssertDriverContentsWithFrameAre (expected, output);
+		}
+
+		[Fact, AutoInitShutdown]
+		public void PosConbine_DimCombine_View_With_SubViews ()
```
tig, line 1813: `PosCombine_DimCombine_View_With_SubViews` (m not m)

UnitTests/Views/ScrollViewTests.cs
```diff
@@ -602,16 +602,17 @@ public void Clear_Window_Inside_ScrollView ()
...
 		}
 
-
 		[Fact, AutoInitShutdown]
 		public void DrawTextFormatter_Respects_The_Clip_Bounds ()
 		{
-			var view = new View (new Rect (0, 0, 20, 20));
-			view.Add (new Label ("0123456789abcdefghij"));
-			view.Add (new Label (0, 1, "1\n2\n3\n4\n5\n6\n7\n8\n9\n0"));
+			var rule = "0123456789";
```
tig, line 608: Can you use the new `Ruler` class here instead?

tig: Also, please remove the "Test" from the below so that as we evolve the look & feel of "Title" we don't have to update unit tests.

tig (changes requested):
Fantastic stuff @BDisp. I am very very very appreciative of how well you understand this stuff and your willingness to fix this stuff!!!

Terminal.Gui/View/View.cs
```diff
@@ -1395,7 +1395,8 @@ public Point ScreenToView (int x, int y)
 		public Point ScreenToBounds (int x, int y)
 		{
 			if (SuperView == null) {
-				return new Point (x - Frame.X + GetBoundsOffset ().X, y - Frame.Y + GetBoundsOffset ().Y);
+				var boundsOffset = GetBoundsOffset ();
```
tig, line 1398: I assume you did this to ease debugging? I often do the same.

I believe (but am not 100%) sure that there's no performance penalty for calling `GetBoundsOffset` twice in `return new Point (x - Frame.X + GetBoundsOffset ().X, y - Frame.Y + GetBoundsOffset ().Y);` because the compiler will optimize it.

BDisp: Was due the performance to not call twice.

tig: Do you know for sure it's actually a perf hit? 

I'm not asking you to change it, I'm just very curious.

tig: FYI: https://stackoverflow.com/a/2162893/297526

My favorite part of his answer:

> n short, you cannot possibly know whether writing the code to cache the result rather than recomputing it is actually (1) faster, or (2) better performing. 

IMO, in this situation, we don't even know if this codepath is a critical perf vector. Until we do, my feeling is we should optimize for code readability, testability, and debugabillity.  

I don't think either version makes any difference for readabilty or testability, but I do think your version is easier to debug. So yay. :-)

Terminal.Gui/View/View.cs
BDisp: I couldn't get a unfinished recursive cycle. I only get some recursive cycle when it is expected if a subview references a superview or a super-superview, but it ended this recursive cycle as expected. If we caught an unfinished recursive cycle in the future we'll handle it.

UICatalog/Scenarios/Scrolling.cs
BDisp: @tig I don't understand how you want I draw the label inside the scroll view here, instead of fill the label text once. Did you make same confusion or I'm missing something?

UnitTests/View/Layout/LayoutTests.cs
BDisp: @tig I haven't any problem on debugging this. I put a breakpoint on it and press F11 and I can go smoothly into the `LayoutSubviews` method to debugging. For me when I want to prove that something won't throw an exception, I like to use the `Record.Exception`.

UnitTests/Views/ScrollViewTests.cs
BDisp: > Can you use the new `Ruler` class here instead?

In a scroll view isn't very easy to use it, because it will force to redraw on every iteration. Fill the label text only once is much more easy.

UICatalog/Scenarios/Scrolling.cs
tig: I just hate duplicate code. I wrote `Ruler` recently so that it could be used anywhere we want a ruler.

BDisp: I now understand what you mean. I created the `StringExtensions` as public to be available where is needed.

tig (approved):

PR #2510, opened 2023-04-06
Add many styles to SpinnerView

Nutzzz:
I've added styles to the new SpinnerView based on [cli-spinners](https://github.com/sindresorhus/cli-spinners/blob/master/spinners.json), which can be previewed [here](https://jsfiddle.net/sindresorhus/2eLtsbey/embedded/result/) or using the new SpinnerStyles scenario in the UICatalog.  I used an enum and switch/case rather than a class to match the existing coding style, but a class method could easily be adapted from [Spectre.Console](https://github.com/spectreconsole/spectre.console).

Spinners can now be larger than 1-character wide.  Some styles require special characters and emojis that are not supported by Windows conhost by default (these have been tagged with .HasSpecial).  See <https://spectreconsole.net/best-practices>.  For extra compatibility, a few styles return true for .IsAsciiOnly.

If the number of styles seems over-large, I later decided to add the ability to easily make custom spinners, which arguably negates the need for having a great plethora of built-in styles and the complications these introduce.  However, since I put in the work of adding them to a UICatalog entry, I figured I'd show off what I have at this point if there's any interest.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/SpinnerView.cs
+++ b/Terminal.Gui/Views/SpinnerView.cs
@@ -1,44 +1,271 @@
-﻿using System;
+﻿//------------------------------------------------------------------------------
+// SpinnerStyles below are derived from <https://github.com/sindresorhus/cli-spinners/blob/master/spinners.json>
+// MIT License
+// Copyright (c) Sindre Sorhus <sindresorhus@gmail.com> (https://sindresorhus.com)
+// Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+// The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//
+// Windows Terminal supports Unicode and Emoji characters, but by default conhost shells (e.g., PowerShell and cmd.exe) do not. See <https://spectreconsole.net/best-practices>.
+//------------------------------------------------------------------------------
+
+using System;
+using System.Text;
 
 namespace Terminal.Gui {
+	/// <summary>
+	/// Spinner styles used in a <see cref="SpinnerView"/>.
+	/// </summary>
+#pragma warning disable CS1591 // Missing XML comment for publicly visible type or member
+	public enum SpinnerStyle {
+		Custom,
+		Dots,
+		Dots2,
+		Dots3,
+		Dots4,
+		Dots5,
+		Dots6,
+		Dots7,
+		Dots8,
+		Dots9,
+		Dots10,
+		Dots11,
+		Dots12,
+		Dots8Bit,
+		Line,
+		Line2,
+		Pipe,
+		SimpleDots,
+		SimpleDotsScrolling,
+		Star,
+		Star2,
+		Flip,
+		Hamburger,
+		GrowVertical,
+		GrowHorizontal,
+		Balloon,
+		Balloon2,
+		Noise,
+		Bounce,
+		BoxBounce,
+		BoxBounce2,
+		Triangle,
+		Arc,
+		Circle,
+		SquareCorners,
+		CircleQuarters,
+		CircleHalves,
+		Squish,
+		Toggle,
+		Toggle2,
+		Toggle3,
+		Toggle4,
+		Toggle5,
+		Toggle6,
+		Toggle7,
+		Toggle8,
+		Toggle9,
+		Toggle10,
+		Toggle11,
+		Toggle12,
+		Toggle13,
+		Arrow,
+		Arrow2,
+		Arrow3,
+		BouncingBar,
+		BouncingBall,
+		Smiley,
+		Monkey,
+		Hearts,
+		Clock,
+		Earth,
+		Material,
+		Moon,
+		Runner,
+		Pong,
+		Shark,
+		Dqpb,
+		Weather,
+		Christmas,
+		Grenade,
+		Point,
+		Layer,
+		BetaWave,
+		FingerDance,
+		FistBump,
+		SoccerHeader,
+		MindBlown,
+		Speaker,
+		OrangePulse,
+		BluePulse,
+		OrangeBluePulse,
+		TimeTravelClock,
+		Aesthetic,
+		Aesthetic2,
+	}
+#pragma warning restore CS1591 // Missing XML comment for publicly visible type or member
 
 	/// <summary>
-	/// A 1x1 <see cref="View"/> based on <see cref="Label"/> which displays a spinning
-	/// line character.
+	/// A <see cref="View"/> based on <see cref="Label"/> which displays (by default) a
+	/// spinning line character.
 	/// </summary>
 	/// <remarks>
 	/// By default animation only occurs when you call <see cref="View.SetNeedsDisplay()"/>.
 	/// Use <see cref="AutoSpin"/> to make the automate calls to <see cref="View.SetNeedsDisplay()"/>.
 	/// </remarks>
 	public class SpinnerView : Label {
-		private Rune [] _runes = new Rune [] { '|', '/', '\u2500', '\\' };
+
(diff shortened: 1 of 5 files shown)


Terminal.Gui/Views/SpinnerView.cs
```diff
@@ -1,44 +1,271 @@
...
+		/// <summary>
+		/// Gets or sets whether spinner should go back and forth through frames rather than
+		/// going to the end and starting again at the beginning.
+		/// </summary>
+		public bool SpinBounce { get; set; } = false;
+
+		/// <summary>
+		/// Gets the size in characters of the spinner.
+		/// </summary>
+		public int SpinnerWidth { get => GetSpinnerWidth (); }
```
tig, line 157: I may be missing something, but why is this needed? Can't we just use `View.Width`?

Terminal.Gui/Views/SpinnerView.cs
```diff
@@ -1,44 +1,271 @@
...
-			Width = 1; Height = 1;
+			SpinnerStyle = DEFAULT_STYLE;
+			Width = GetSpinnerWidth ();
+			Height = 1;
+		}
+
+		/// <summary>
+		/// Creates a new instance of the <see cref="SpinnerView"/> class.
+		/// </summary>
+		public SpinnerView (SpinnerStyle style)
```
tig, line 182: Please don't introduce parametrized constructors. Devs can use object initialization instead.

Terminal.Gui/Views/SpinnerView.cs
```diff
@@ -1,44 +1,271 @@
...
+				Frames = frames;
+				Width = GetSpinnerWidth ();
+				SpinDelayInMilliseconds = delay;
+			} else {
+				Width = 1;
+			}
+			Height = 1;
+		}
+
+		private int GetSpinnerWidth ()
```
tig, line 207: Again, regarding `Width` - This raises a question for me:

What happens if `SpinnerWidth` is > `View.Width`?

Terminal.Gui/Views/SpinnerView.cs
```diff
@@ -71,5 +298,1492 @@ protected override void Dispose (bool disposing)
...
+			get => _style;
+			set {
+				_style = value;
+				_bounceReverse = false;
+				SpinDelayInMilliseconds = 80;
+				SpinBounce = false;
+				SpinReverse = false;
+				_hasSpecial = false;
+				switch (value) {
+				case SpinnerStyle.Custom: // Placeholder when user has specified delay and frames manually
```
tig, line 315: What would you say to using a simple class-heirarchy for all this instead of a enum + giant switch statement?

E.g.

```cs
public class SpinnerStyle {
   public int Delay { get ; set } = 80;
   public bool Bounce { get ; set } = false;
...
   public string [] Frames { get ; set; };

   /// location is screen relative like all primitive draw functions (see `Ruler`)
   public virtual void Draw (Point location) {
       // Code from `SpinnerView.Redraw()`
   }
}

public class Dots : SpinnerStyle {
   public Dots () {
        Frames = new string []
					{
						"⠋",
						"⠙",
						"⠹",
						"⠸",
						"⠼",
						"⠴",
						"⠦",
						"⠧",
						"⠇",
						"⠏",
					};
     }
}
...
```

tig (changes requested):
This is gonna be a very cool addition to the project! I love it all!!!

Please see my comments and requested changes.

Terminal.Gui/Views/SpinnerView.cs
tznind: That would be a lot of classes in the root namespace.  

If going down this route I think they should either be nested classes or start with the same text e.g. `SpinnerStyleDots`, `SpinnerStyleBounce` etc.  I hate nested classes so probably the later?

Theres probably an alternative solution involving a dictionary, something like:

```csharp
this.dict = new Dictionary<SpinnerStyle,SpinnerDefinition> ()

this.Add(SpinnerStyle .Dots, new SpinnerDefinition(80,false,
"⠋","⠙","⠹","⠸","⠼","⠴","⠦","⠧","⠇","⠏",));
this.Add(SpinnerStyle .Star, new SpinnerDefinition(80,false,
/*etc*/));
this.Add(SpinnerStyle.Aesthetic, new SpinnerDefinition(80,false,
/*etc*/));
```

SpinnerDefinition would be something like 

```csharp
class SpinnerDefinition
{
	public int Delay { get; }
	public bool Bounce { get; }
	public string [] AnimationFrames { get; set; }
	public SpinnerDefinition (int delay, bool bounce, params string [] animationFrames)
	{
		Delay = delay;
		Bounce = bounce;
		AnimationFrames = animationFrames;
	}
}
```

Nutzzz: Re: "Frames," perhaps it could be called "Animation" or "Sequence" or even just "Spinner" instead?

Re: SpinnerWidth, I imagined it would be possible for the SpinnerView frame to be larger than the animation itself. But I suppose such a case could be accomplished with margins or padding.

Re: classes vs. enum/switch [vs. dictionary, etc.], as I mentioned in the comment to the PR, I was just trying to match the methodology I saw used for other styles. I feel like [nested] classes would be mo' better (especially here with so many possibilities), and I'm happy to change it [to whatever you guys decide. I might offer similar PRs for other styles, so it would be good to decide now. ;-) ]

tig: I think the current `enum` + `switch` is the most fragile, hard to test, and hard to extend of the choices.

I don't have @tznind aversion to nested classes (Thomas, would you mind explaining why you have said aversion?).

tznind: Its a soft aversion 😉 and I have some in FileDialog so it's a bit hypocritical too.

Mostly its based on the reasons in the microsoft design guidlines:

https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types

That said though, this is a good exception to the rule I think.  If I understand the pattern here we are talking about avoiding cluttering in the root namespace with 20+ new classes.  So thats worth it to me.

Lets go with the nested classes approach 🚀

Terminal.Gui/Views/SpinnerView.cs
```diff
@@ -1,44 +1,271 @@
...
-	/// A 1x1 <see cref="View"/> based on <see cref="Label"/> which displays a spinning
-	/// line character.
+	/// A <see cref="View"/> based on <see cref="Label"/> which displays (by default) a
+	/// spinning line character.
 	/// </summary>
 	/// <remarks>
 	/// By default animation only occurs when you call <see cref="View.SetNeedsDisplay()"/>.
 	/// Use <see cref="AutoSpin"/> to make the automate calls to <see cref="View.SetNeedsDisplay()"/>.
 	/// </remarks>
 	public class SpinnerView : Label {
```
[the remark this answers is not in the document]
Nutzzz: Now inherits from View.

Terminal.Gui/Views/SpinnerView.cs
```diff
@@ -1,44 +1,271 @@
...
+		private bool _bounceReverse = false;
+		private SpinnerStyle _style;
 		private int _currentIdx = 0;
 		private DateTime _lastRender = DateTime.MinValue;
 		private object _timeout;
 
+		/// <summary>
+		/// Gets or sets the frames used to animate the spinner.
+		/// </summary>
+		public string [] Frames { get; set; }
```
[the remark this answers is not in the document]
Nutzzz: Now called Sequence.  We need to be able to set the sequence for the Custom style.

Terminal.Gui/Views/SpinnerView.cs
Nutzzz: Removed.

Terminal.Gui/Views/SpinnerView/SpinnerStyle.cs
```diff
@@ -0,0 +1,1699 @@
...
+		const bool DEFAULT_SPECIAL = false;
+
+		/// <summary>
+		/// Gets or sets the number of milliseconds to wait between characters
+		/// in the spin.  Defaults to the SpinnerStyle's Interval value.
+		/// </summary>
+		/// <remarks>This is the maximum speed the spinner will rotate at.  You still need to
+		/// call <see cref="View.SetNeedsDisplay()"/> or <see cref="SpinnerView.AutoSpin"/> to
+		/// advance/start animation.</remarks>
+		public abstract int SpinDelayInMilliseconds { get; }
```
tig, line 46: IMO, delays & timers etc... are almost always in ms. I'd prefer this just be `SpinDelay`; tersness is always better. 

Of course the API docs should call that out too.

tig (approved):

Terminal.Gui/Views/SpinnerView/SpinnerStyle.cs
tznind: We could go even clearer and just use a `TimeSpan` type e.g.

```csharp
/// <summary>
/// Gets or sets the time to wait between characters in the spin.  Defaults to 250 milliseconds.
/// </summary>
/// <remarks>This is the maximum speed the spinner will rotate at.  You still need to
/// call <see cref="View.SetNeedsDisplay()"/> or <see cref="SpinnerView.AutoSpin"/> to
/// advance/start animation.</remarks>
public TimeSpan SpinDelay { get; set; } = TimeSpan.FromMilliseconds(250);
```

tig:
@Nutzzz you've got a test failure here. Just in case you didn't see...

tznind:
Support for `SpinnerView.Style` in designer.  For when v2 comes out.

Thought you might enjoy @Nutzzz :

spinners

On a side note, it seems the AutoSpin implementation results in the other spinners redrawing too.  I guess `SpinnerView` needs a slight behavior adjustment when AutoSpin is false.  I will create a ticket.

PR #2514, opened 2023-04-07
Add Style.ShowHeaders to TableView

Nutzzz:
Very simple change to allow table headers to be hidden (necessary for e.g., using TableView as a single-dimension ListView split into multiple columns).  I'm also currently looking at methods of making such a use case easier (with options for horizontal or vertical item population and horizontal or vertical scrolling).

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/TableView/TableView.cs
+++ b/Terminal.Gui/Views/TableView/TableView.cs
@@ -262,12 +262,14 @@ namespace Terminal.Gui {
 					line++;
 				}
 
-				RenderHeaderMidline (line, columnsToRender);
-				line++;
-
-				if (Style.ShowHorizontalHeaderUnderline) {
-					RenderHeaderUnderline (line, bounds.Width, columnsToRender);
+				if (!Style.HideHeaders) {
+					RenderHeaderMidline (line, columnsToRender);
 					line++;
+
+					if (Style.ShowHorizontalHeaderUnderline) {
+						RenderHeaderUnderline (line, bounds.Width, columnsToRender);
+						line++;
+					}
 				}
 			}
 
@@ -316,8 +318,11 @@ namespace Terminal.Gui {
 		/// <returns></returns>
 		private int GetHeaderHeight ()
 		{
-			if (Style.HideHeaders)
+			if (Style.HideHeaders) {
+				if (Style.ShowHorizontalHeaderOverline)
+					return 1;
 				return 0;
+			}
 
 			int heightRequired = 1;
 
@@ -1583,7 +1588,7 @@ namespace Terminal.Gui {
 
 		private bool ShouldRenderHeaders ()
 		{
-			if (TableIsNullOrInvisible () || Style.HideHeaders)
+			if (TableIsNullOrInvisible ())
 				return false;
 
 			return Style.AlwaysShowHeaders || rowOffset == 0;
@@ -1773,7 +1778,8 @@ namespace Terminal.Gui {
 
 			/// <summary>
 			/// Gets or sets a flag indicating whether to suppress rendering headers of a <see cref="TableView"/>.
-			/// Defaults to false.
+			/// Defaults to false.  ShowHorizontalHeaderOverline may still be used, and AlwaysShowHeaders will
+			/// always show it.
 			/// </summary>
 			public bool HideHeaders { get; set; } = false;
 
--- a/Terminal.Gui/Views/TableView/TableView.cs
+++ b/Terminal.Gui/Views/TableView/TableView.cs
@@ -315,13 +315,16 @@ namespace Terminal.Gui {
 		/// <returns></returns>
 		private int GetHeaderHeight ()
 		{
-			int heightRequired = 1;
-
-			if (Style.ShowHorizontalHeaderOverline)
+			int heightRequired = 0;
+			if (!Style.HideHeaders) {
 				heightRequired++;
 
-			if (Style.ShowHorizontalHeaderUnderline)
-				heightRequired++;
+				if (Style.ShowHorizontalHeaderOverline)
+					heightRequired++;
+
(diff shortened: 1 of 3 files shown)


tznind:
Thanks for looking into this use case.  It sounds exciting.

When new functionality is added to a View there are 2 things we need in addition to the code change:

- A scenario (or change to existing scenario) that demonstrates the feature.
- One or more tests that ensure the feature works

Could you please add a setting in `TableEditor` scenario for toggling this feature.  See `miHeaderOverline` and the other fields for how to do this.

Also please add some tests to `TableViewTests` that confirm the rendering is as expected in different situations.  See tests like `TableView_ExpandLastColumn_False_ExactBounds` as an example.

Terminal.Gui/Views/TableView/TableView.cs
```diff
@@ -1772,6 +1775,11 @@ public class TableStyle {
 			/// </summary>
 			public bool AlwaysShowHeaders { get; set; } = false;
 
+			/// <summary>
+			/// Hide header
```
tznind, line 1779: Please elaborate on this comment e.g.:

> Gets or sets a flag indicating whether to suppress rendering headers of a <see cref="TableView"/>.  Defaults to false.

Terminal.Gui/Views/TableView/TableView.cs
```diff
@@ -315,13 +315,16 @@ private int GetHeaderHeightIfAny ()
 		/// <returns></returns>
 		private int GetHeaderHeight ()
 		{
```
tznind, line 317: Could this be simplified by just adding the following at the top of the method instead?  Typically having less nesting is easier to read and maintain so would appreciate if you could make this change.

```csharp
if (Style.HideHeaders) {
    return 0;
}
```

tznind (changes requested):

tznind:
Bugs

- Sometimes the Back arrow will take you from "c:\somedir\" back to "c:\somedir".  These duplicates should not happen
- Tab Autocomplete has a case sensitivity issue where if you have a file system:

```csharp
/folder/item1
/folder/item2
```

And you enter a capital I (i.e. `"/folder/I"`) then it will correctly propose "item1" and tab completion will work BUT if youpress Down arrow the focus will move to table view (i.e. it will not cycle suggestions).

Terminal.Gui/Views/TableView/TableView.cs
```diff
@@ -315,13 +315,16 @@ private int GetHeaderHeightIfAny ()
...
-			int heightRequired = 1;
```
tznind, line 318: Could this be simplified by just adding the following at the top of the method instead?  Typically having less nesting is easier to read and maintain so would appreciate if you could make this change.

```csharp
if (Style.HideHeaders) {
    return 0;
}
```

tig (approved):
I approve, but @tznind has final call.

Nutzzz:
@tznind Sorry, I decided to change this to allow the top border to be drawn even when HideHeaders is enabled.  Can you re-review?  I modified the TableEditor scenario and added Unit Tests.

tznind:
Looks good to me.  Thanks for taking the time to write the tests 🙂.

I'm on mobile atm so can't test but it looks perfect.  Happy for it to
merge.

Terminal.Gui/Views/TableView/TableView.cs
```diff
@@ -315,16 +316,16 @@ private int GetHeaderHeightIfAny ()
 		/// <returns></returns>
 		private int GetHeaderHeight ()
 		{
-			int heightRequired = 0;
-			if (!Style.HideHeaders) {
-				heightRequired++;
+			if (Style.HideHeaders)
+				return 0;
```
tig, line 324: Please use brackets for all nested statements. Even single lines. Thanks.

tig (commented):
@tznind needs to approve this.

Please see my comment.

Terminal.Gui/Views/TableView/TableView.cs
Nutzzz: This was mooted, but thanks for the reminder.

tznind (approved):

Terminal.Gui/Views/TableView/TableView.cs
tig: > This was mooted, but thanks for the reminder.

Not sure what you mean by `mooted`. 

Before I can merge this, I need you to add brackets to all `if` statements, even if just a single line.

tznind:
I am looking at #2543 and the style options.  I think I want to tweak this a little if that is ok.  All the other options begin with `Show` 

image

It makes me think this option should be `ShowHeaders` and default to `true`

I also think you should be able to mix and match both `ShowHorizontalHeaderOverline` and `ShowHorizontalHeaderUnderline` without caring about whether `ShowHeaders` is true.

I will submit a PR to your branch

tznind:
Here is the PR @Nutzzz  https://github.com/Nutzzz/Terminal.Gui/pull/1 

I think this simplifies the API.  Turning off headers does not change the other lines that are drawn.  Critically I think the underline is important as an option because it displays horizontal scroll indicators on that line so should be an option even without headers.

image
_TableEditor scenario with ShowHeaders changed to false_

Nutzzz:
Yeah, thanks. These changes work for me.

As far as the name I didn't like mixing Hide and Show either, but it was to avoid confusion  with AlwaysShowHeaders, which ought to have been called KeepHeadersOnscreen or something.

As for the underline, I didn't allow it just because it looks pretty weird. While the indicator is helpful, it seems like it could be moved somewhere... else? ...when the underline is disabled, since you're using up a whole line for the benefit of one or two characters.  A scrollbar would be better.  But you're right, let the developer figure out what they want to do.

Issue #2483, opened 2023-04-04
Remove need for `TileView` - Use `LineCanvas` to draw borders and auto-join borders


tig:
With the new `Frames` functionality in v2, it will be possible to have subviews with overlapping `BorderFrames` auto-join borders using `LineCanvas`.

This will require the new `DrawFrames`/`OnDrawFrames` methods to utilize a `LineCanvas` managed by the superview.

I've done a POC (in the `TileView Experiments` scenario) that shows this can work.

tig:
Woot! 

A little proof of concept!

All this really took was 

a) Adding changing `public LineCanvas LineCanvas = new LineCanvas ();` to `View`
b) Modifying `View.ReDraw` to render `LineCanvas` and then clear it (with a new `LineCanvas.Clear` method).
c) Modifying `Frame` to use `Parent?.SuperView?.LineCanvas ?? LineCanvas;` instead of `new LineCanvas`
d) Modifying `Frame` to NOT render `LineCanvas` itself

![D16CspQ 1](https://user-images.githubusercontent.com/585482/231028235-7771432a-ab43-4631-afcd-7c76e9a2a737.gif)

tig:
@tznind and @BDisp - I've been spending too much time coding lately and need to take a break so I can attend to other tings.

With #2527 merged a TON of v2 stuff (see the Backlog in https://github.com/orgs/gui-cs/projects/1) can be worked on.

I'll still be checking in to do PR reviews but won't be writing much code for a while.

Cheers!

tznind:
> I've been spending too much time coding lately and need to take a break so I can attend to other tings.

Thanks for all the awesome work recently.  Our standardisation changes are huge! Its going to be amazing.  But definetly work at a sustainable pace.  Codings a long road 🚗 🚛 🏍️

tznind:
A lot of the code in TileView is about enabling user to move the splitter lines, preventing moving the splitters beyond a max/min size for any pair of tiles and having a nice public API to enable user to dynamically split at runtime.

Basically when combined with TabView you can almost have the full 'dock panel' experience (like visual studio).

I love the free floating frames and docking together and I'm happy to update the implementation of draw/layout in TileView if needed.  But I think that the control and its API surface are worth keeping?

Happy to discuss it more.

Issue #2482, opened 2023-04-03
Refactor `Redraw` - Non-virtual with the right set of virtual `OnXXX` methods


tig:
In v2, i'd like us to revisit how `Redraw` and the related events are structured. I started this in my View2 PR...
 
>     * `Redraw` is mis-named. It should just be `Draw` (it also shouldn't take a `Bounds` because implementors can just use `this.Bounds`; this is a cause of a bunch of bugs BTW).
> 
>     * `Draw` should simply call
>       
>       * `OnDraw`
>       * `OnDrawFrames`
>       * `OnDrawContent`
>       * `OnDrawContentComplete`
> 
> 
> Each of those invokes an event and each is cancellable (except OnDrawContentComplete).

@tig I agree with this scenario, but `Draw` mustn't be virtual to avoid derived class not calling the  `OnDraw`, `OnDrawFrames`, `OnDrawContent` and `OnDrawContentComplete`. Maybe `protected`. What do you think?

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/2479#issuecomment-1494124984_

tig:
This was addressed.

PR #2540, opened 2023-04-13
Fixes #2539. Menu should use Frame instead of DrawFrame.

### LINKED ISSUE #2539: Menu should use Frame instead of DrawFrame

BDisp:
Fixes #2539 - Include a terse summary of the change or which issue is fixed.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/Menu.cs
+++ b/Terminal.Gui/Views/Menu.cs
@@ -477,6 +477,9 @@ namespace Terminal.Gui {
 				WantMousePositionReports = host.WantMousePositionReports;
 			}
 
+			BorderFrame.Thickness = new Thickness (1);
+			BorderFrame.BorderStyle = LineStyle.Single;
+
 			if (Application.Current != null) {
 				Application.Current.DrawContentComplete += Current_DrawContentComplete;
 			}
@@ -534,19 +537,14 @@ namespace Terminal.Gui {
 
 		public override void Redraw (Rect bounds)
 		{
-		}
-
-		// Draws the Menu, within the Frame
-		private void Current_DrawContentComplete (object sender, DrawEventArgs e)
-		{
 			if (barItems.Children == null) {
 				return;
 			}
-			var savedClip = Driver.Clip;
+			var savedClip = Application.Driver.Clip;
 			Application.Driver.Clip = Application.Top.Frame;
 
 			Driver.SetAttribute (GetNormalColor ());
-			DrawFrame (Bounds, padding: 0, fill: true);
+			OnDrawFrames (Frame);
 
 			for (int i = Bounds.Y; i < barItems.Children.Length; i++) {
 				if (i < 0)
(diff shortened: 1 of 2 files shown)


tig (approved):
Hero!

At some point, it seems Menu would be better off just being a view where each menu item is just a `View` (a `Label`?). Each 'frame' (e.g. separators) the menu would then be those View's `Border`.

BDisp:
Well I was very lazy and I only used the view frame and the internal menu separator by changing the coordinates, but it could also be in the way you point to.

tig:
Well your lazyness just caused me a bunch of work because I assumed you had done it carefully. 😠 💘

I'm merging #2483 now, but you'll see me merging in this PR into it broke some Menu tests that I had to disable because I ran out of time.. I'd love it if you'd look at them.

`Menu.cs` will be about 30% shorter if it can be made to fully leverage `LineCanvas`.

Thanks.

BDisp:
@tig in the case of the `Menu` there is no need to use `LineCanvas`, with `Frame` it works well. Beside that I must understand better how it works. I'll see the unit tests.

BDisp:
@tig you are mixing `Frame` with `LineCanvas`. Also the `OnRenderLineCanvas` must be call at the beginning of the Draw. What is bad in your design is that you are drawing the `Border` after the subviews content are drawn. To draw normal frame the `Frame` is ore than sufficient. If you want join borders then we must use the `LineCanvas`, but using both in the same view is very stranger.

BDisp:
@tig I swear I already had the menu to work before but something broke it. Is it not possible to draw a simple edge around the view frame? Run `UICatalog` and you'll see the menu is working as expected.

tig:
> @tig you are mixing `Frame` with `LineCanvas`. Also the `OnRenderLineCanvas` must be call at the beginning of the Draw. What is bad in your design is that you are drawing the `Border` after the subviews content are drawn. To draw normal frame the `Frame` is ore than sufficient. If you want join borders then we must use the `LineCanvas`, but using both in the same view is very stranger.

My design is not fully implemented yet. If you read the spec you'll see that.

When this is all complete the `Frame`s will be drawn INDEPENDENTLY and BEFORE the View's content is drawn.

See https://github.com/gui-cs/Terminal.Gui/issues/2482

BDisp:
> Hero!
> 
> At some point, it seems Menu would be better off just being a view where each menu item is just a `View` (a `Label`?). Each 'frame' (e.g. separators) the menu would then be those View's `Border`.

I understand what you mean but it will imply a full refactor on the `Menu` which I don't know it it worth. I'll add the enable/disable border feature in the menus.

BDisp:
For those much more brave than me here is a suggestion for two items without separator (remember that each member is a `View`):


```
┌───────┐
│ One   │
└───────┘
┌───────┐
│ Two   │
└───────┘

```
Since there is no separator the bottom thickness of the `One` item must be 0 and the top thickness of the `Two` item must also be 0. This will result the follow output:

```
┌───────┐
│ One   │
│ Two   │
└───────┘

```

More harder is when a separator is needed. In this case the bottom thickness of the `One` item must be 1 (default) and the top thickness of the `Two` item must be 0. This will result the follow output:

```
┌───────┐
│ One   │
└───────┘
│ Two   │
└───────┘

```

But the job isn't finished yet. It's needed to join the left with a `├` and on the right with a `┤` and the final result will be the follow output:

```
┌───────┐
│ One   │
├───────┤
│ Two   │
└───────┘

```
I don't want to give any lesson to anyone, but I only alert for the complexity, not for the steps above, but to the views that must be added to the superview and beside dealing with all menu/sub-menus opened, we needed to handle more much more views than currently has.

tig:
The whole point of my latest big pr was to make two views that touch blend borders automatically!

Just set UseSuperviewLineCanvas = true

PR #2527, opened 2023-04-11
Fxes #2483 - Modifying 

tig:
Fixes #2148 

## Pull Request checklist:

- [ ] I've named my PR in the form of "Fixes #issue. Terse description."
- [ ] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [ ] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [ ] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [ ] My changes generate no new warnings
- [ ] I have checked my code and corrected any poor grammar or misspellings
- [ ] I conducted basic QA to assure all features are working

PR #2542, opened 2023-04-13
Fixes menu frame and unit test on the #2483.

BDisp:

## Pull Request checklist:


BDisp:
@tig I broken the border around the listviews on the UICatalog, but I think you'll discovering that fast.

Terminal.Gui/View/View.cs
```diff
@@ -1756,18 +1756,15 @@ protected void ClearNeedsDisplay ()
 		/// <returns></returns>
 		public virtual bool OnDrawFrames ()
 		{
-			if (!IsInitialized) {
-				return false;
-			}
+			//if (!IsInitialized) {
```
tig, line 1759: This is incorrect. All drawing is indeterminate until after the EndInit.

Terminal.Gui/View/View.cs
```diff
@@ -1756,18 +1756,15 @@ protected void ClearNeedsDisplay ()
...
 
 			var prevClip = Driver.Clip;
-			Driver.Clip = ViewToScreen (Frame);
 
-			// TODO: Figure out what we should do if we have no superview
-			//if (SuperView != null) {
-			// TODO: Clipping is disabled for now to ensure we see errors
-			Driver.Clip = new Rect (0, 0, Driver.Cols, Driver.Rows);// screenBounds;// SuperView.ClipToBounds ();
-										//}
+			if (SuperView != null) {
```
tig, line 1765: Please remove this. See my comment above.

I plan on re-enabling clipping later; it's important to see where things are being drawn where they're not supposed to be before we re-implement a proper clipping mechanism (the one in v1 was badly architected and broken).

Terminal.Gui/View/View.cs
```diff
@@ -1805,6 +1802,8 @@ public virtual void Redraw (Rect bounds)
 
 			OnDrawFrames ();
 
+			OnRenderLineCanvas ();
```
tig, line 1805: You are not understanding how this is supposed to work (it's not all there yet, so that's understandable).

`View.OnDrawFrames` writes all borders to `View.LineCanvas`. 

Then, at a later time (any time) `View.OnRenderLineCanvas` will be called to cause the `View.LineCanvas` contents to be rendered.  The full architecture for doing this right is not in place yet. 

When I merged #2527 I had hacked things so that EVERYTHING except for those two unit test worked. 

I'd like you to see if you can fix the unit tests another way.

If you can't that's Ok. we can fix them later when more of the new architecture is in place. They cover edge cases anyway.

tig (changes requested):
I think you should just wait to fix these unit tests until after more of the new architecture is implemented. 

Ok?

Terminal.Gui/View/View.cs
BDisp: Without the changes I made on Views I couldn't fix them, sorry. I'm still confused with all these new code.

BDisp: I reverted some code, but the `Menu` border need to render first for the separator drawing well.

BDisp:
@tig On the `ScrollView` color is failing. Maybe not a big deal but I need to look in better. `OnDrawContentComplete` must run after the `OnRenderLineCanvas`. Otherwise all the work on the `OnDrawContentComplete` will be damaged by the `OnRenderLineCanvas`.

BDisp:
@tig the `Driver.Clip` must be restored before call the `OnRenderLineCanvas` and the `OnDrawContentComplete` must be called after the `OnRenderLineCanvas`, unless this was intentional done for testing.

@tig can you change the `.editorconfig` to set the tab = 2 and not force a space before a parentheses on methods as Method()? Thanks.

tig:
> @tig can you change the `.editorconfig` to set the tab = 2 and not force a space before a parentheses on methods as Method()? Thanks.

Both are a Miguel/Xamarin thing. I don't think we should change it.

tig (approved):

Issue #2562, opened 2023-04-17
Scroll control doesn't work properly


ttool22:
**C# code is below**

var scrollView = new ScrollView();
scrollView.X = 0;
scrollView.Y = 0;
scrollView.Width = Dim.Fill();
scrollView.Height = Dim.Fill();
scrollView.ContentSize = new Size(Console.WindowWidth, Console.WindowHeight * 2);
_____

When I scroll down, I got this error(below), I have win 11 ;(

A clear and concCollection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at Terminal.Gui.View.Redraw(Rect bounds)
   at Terminal.Gui.FrameView.Redraw(Rect bounds)
   at Terminal.Gui.View.Redraw(Rect bounds)
   at Terminal.Gui.ScrollView.Redraw(Rect region)
   at Terminal.Gui.View.Redraw(Rect bounds)
   at Terminal.Gui.Window.Redraw(Rect bounds)
   at Terminal.Gui.Application.RunMainLoopIteration(RunState& state, Boolean wait, Boolean& firstIteration)
   at Terminal.Gui.Application.RunLoop(RunState state, Boolean wait)
   at Terminal.Gui.Application.Run(Toplevel view, Func`2 errorHandler)
   at Terminal.Gui.MessageBox.QueryFull(Boolean useErrorColors, Int32 width, Int32 height, ustring title, ustring message, Int32 

*

BDisp:
I couldn't reproduce this. I see you are using the scroll inside a frame. On what `Terminal.Gui` version you are running, v1 or v2?

tznind:
Or if you are using the Nuget package? If so what version?  Are you able to provide a link to your source code or give any more details.  For example where do you launch the View, do you call Application.Init etc.

ttool22:
I ha have 1.72 version with win 11, I have added scroll inside window
**win.Add(scrollView);**
 **Application.Run(win, (e) =>
{
    MessageBox.ErrorQuery("Error", e.ToString(), "Ok");
    return false;
});**

ttool22:
When I run my app on win 11, controls are mixed  then I fix it by expanding the window with the mouse.

BDisp:
I recommend you to download the current v1 from the repo, which is the `develop` branch and check if the issue still exist.

ttool22:
Now I am using the latest nuget (1.10.1), I have this error
[...]

PR #2568, opened 2023-04-21
Add article on threading

tznind:
Often we encounter users who are using Threads but do not know about `Application.MainLoop.Invoke`.  For example:

> Yes every one sec I have to use the same frameView with new data.

https://github.com/gui-cs/Terminal.Gui/issues/2562#issuecomment-1514559020

I have created an article to explain how to update main UI thread from a background task.

Note that this targets `develop` so current users should see it.  We can cherry pick it into `v2_develop`

## Pull Request checklist:


tznind:
I see though that there is a section in `mainloop.md` but I think its worthy to be its own article since it is so common to see this issue.

migueldeicaza:
This has been on the docs for a very long time:

https://github.com/gui-cs/Terminal.Gui/blob/develop/docfx/articles/mainloop.md#threading

docfx/articles/threading.md
```diff
@@ -0,0 +1,51 @@
+# Threading
+It is common for a developer to run one or more background tasks on a separate
+`Thread` (or `Task`). For example to periodically fetch data. If you want to
```
tig, line 3: Please use  the "infinitive form" in API docs instead of the "impersonal you". The infinitive form of a verb is its basic, uninflected form, typically preceded by the word "to" in English. For example, "to eat," "to play," or "to learn." By using the infinitive form, the sentence becomes more concise and focuses on the action itself.

This would be rewritten as "To make changes to a `View` from a background task, issue the change on the main Terminal.Gui Thread thread."

Miguel tends to write using the "impersonal you" form, so a lot of the old docs still use this. But I've worked hard to rid most API docs of it. Thanks.

tig (changes requested):

En3Tho:
I guess it's worth mentioning that Termital.Gui has it's own SynchronizationContext and it really depends on how this task is started.

(Note: SynchronizationContext is just an abstraction over some kind of Scheduler. In case of Terminal.Gui it's a single UI thread message loop)
 
Few examples with default System.Threading.Task:

```
Button clicked by user -> Processed by UI thread -> Has SynchronizationContext -> SomeAsyncMethodCalled with no ConfigureAwait(false) -> Some awaits inside -> Always jump back to UI's thread SynchronizationContext when await is done -> UI updates are valid

In this case some Task is started from UI thread and SynchronizationContext is saved so technically it's not background.
```

```
Button clicked by user -> Processed by UI thread -> Has SynchronizationContext -> SomeAsyncMethodCalled with ConfigureAwait(false) -> Some awaits inside -> Always jump to thread pool because ConfigureAwait(false) makes it ignore SynchronizationContext -> UI updates are invalid

In this case some Task is started from UI thread and SynchronizationContext is not saved so technically it's background.
```

```
Some work started with Task.Run -> Processed by ThreadPool thread -> No SynchronizationContext -> ... -> UI updates are invalid

This case is quite straight: no SynchronizationContext from the start because it's queued directly on ThreadPool
```

Examples above show that the problem is that when using async machinery you can't always be sure whether you are on UI thread or not.

Possible means to fix this:
1. Do every UI update via Appication.MainLoop (like in the doc) if you're not sure code is running on UI thread after an await.
2. (My favorite) Just write a custom Task that will always run on UI thread via Mainloop.Invoke and run every async UI update using this custom Task. This is very similar to 1) but is done automatically

I hope this helps someone understand it better

tig:
@tznind did you see my change request?

tznind:
Yeah sorry,  not had a chance to get back to this yet.  Its on my todo list.

PR #2576, opened 2023-04-27
Fixes #2575 - TableView to use interface instead of System.Data.DataTable

### LINKED ISSUE #2575: Support `IList` as a data provider for `TableView`.
Refactoring `TableView` to use a new interface `ITableDataSource` instead of `System.Data.DataTable`.  

Interface would be something like

```csharp
interface ITableDataSource
{
    int Rows { get; }
    int Columns { get; }
    string[] ColumnNames { get; }
    object this[int row, int col]
    {
        get;
    }
}
```

This will be breaking but we are working on v2 I think we should consider it as its a nice feature and it is quite a common use case (to use arbitrary objects as the table data items).

That is the approach taken in TermKit and looks like it will make `TableView` much more flexible:

> /// Protocol that describes a matrix made up of columns and rows, where the columns can be labeled
> public protocol DataSource {

https://github.com/migueldeicaza/TermKit/blob/main/Sources/TermKit/Views/DataTable.swift

### The Goal

The more I look at it, the more I'm excited about the idea.

Check this out! How cool would it be for an API user to be able to do something like:

```csharp
var s = new EnumerableTableDataSource<Process>(Process.GetProcesses(),
    new()
    {
        { "Name",(p)=>p.ProcessName},
        { "Id",(p)=>p.Id},
        { "HasExited",(p)=>p.HasExited},
    });

myTableView.Table = s;
```

The `EnumerableTableDataSource<T>` class definition would be something like:

```csharp

class EnumerableTableDataSource<T> : ITableDataSource
{
    private T[] data;
    private string[] cols;
    private Dictionary<string, Func<T, object>> lamdas;

    public EnumerableTableDataSource(IEnumerable<T> data, Dictionary<string, Func<T,object>> columnDefinitions)
    {
        this.data = data.ToArray();
        this.cols = columnDefinitions.Keys.ToArray();
        this.lamdas = columnDefinitions;
    }

    public object this[int row, int col]
    {
        get => this.lamdas[ColumnNames[col]](this.data[row]);
    }

    public int Rows => data.Length;

    public int Columns => cols.Length;

    public string[] ColumnNames => cols;
}
```

_Originally posted by @tznind in https://github.com/gui-cs/Terminal.Gui/issues/2573#issuecomment-1522899866_

tznind:
Fixes #2575 - Changes `TableView` from being based on `System.Data.DataTable` to `ITableDataSource`

This PR makes it super easy to create `TableView` that presents any arbitrary collection.  The user is no longer required to mess about converting their data into `System.Data.DataTable` in order to render.

```csharp
[Fact,AutoInitShutdown]
public void TestEnumerableDataSource_BasicTypes()
{
	var tv = new TableView ();
	tv.ColorScheme = Colors.TopLevel;
	tv.Bounds = new Rect (0, 0, 50, 6);

	tv.Table = new EnumerableTableSource<Type> (
		new Type [] { typeof (string), typeof (int), typeof (float) },
		new () {
			{ "Name", (t)=>t.Name},
			{ "Namespace", (t)=>t.Namespace},
			{ "BaseType", (t)=>t.BaseType}
		});

	tv.LayoutSubviews ();

	tv.Redraw (tv.Bounds);

	string expected =
		@"
┌──────┬─────────┬───────────────────────────────┐
│Name  │Namespace│BaseType                       │
├──────┼─────────┼───────────────────────────────┤
│String│System   │System.Object                  │
│Int32 │System   │System.ValueType               │
│Single│System   │System.ValueType               │";

	TestHelpers.AssertDriverContentsAre (expected, output);
}
```
process-table

_ProcessTable scenario_

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/FileDialog.cs
+++ b/Terminal.Gui/Views/FileDialog.cs
@@ -107,7 +107,7 @@ namespace Terminal.Gui {
 		private MenuBar allowedTypeMenuBar;
 		private MenuBarItem allowedTypeMenu;
 		private MenuItem [] allowedTypeMenuItems;
-		private DataColumn filenameColumn;
+		private int filenameColumn;
 
 		/// <summary>
 		/// Event fired when user attempts to confirm a selection (or multi selection).
@@ -226,7 +226,7 @@ namespace Terminal.Gui {
 				if (this.tableView.SelectedRow <= 0) {
 					this.NavigateIf (k, Key.CursorUp, this.tbPath);
 				}
-				if (this.tableView.SelectedRow == this.tableView.Table.Rows.Count-1) {
+				if (this.tableView.SelectedRow == this.tableView.Table.Rows - 1) {
 					this.NavigateIf (k, Key.CursorDown, this.btnToggleSplitterCollapse);
 				}
 
@@ -318,7 +318,7 @@ namespace Terminal.Gui {
 			this.sorter = new FileDialogSorter (this, this.tableView);
 			this.history = new FileDialogHistory (this);
 
-			this.tableView.Table = this.dtFiles;
+			this.tableView.Table = new DataTableSource(this.dtFiles);
 
 			this.tbPath.TextChanged += (s, e) => this.PathChanged ();
 
@@ -514,7 +514,7 @@ namespace Terminal.Gui {
 
 		private void CycleToNextTableEntryBeginningWith (KeyEventEventArgs keyEvent)
 		{
-			if (tableView.Table.Rows.Count == 0) {
+			if (tableView.Table.Rows == 0) {
 				return;
 			}
 
@@ -539,11 +539,10 @@ namespace Terminal.Gui {
 		{
 			tableView.EnsureValidSelection ();
 			var col = tableView.SelectedColumn;
-			var style = tableView.Style.GetColumnStyleIfAny (tableView.Table.Columns [col]);
+			var style = tableView.Style.GetColumnStyleIfAny (col);
 
 
-			var collection = tableView
-				.Table
+			var collection = dtFiles
 				.Rows
 				.Cast<DataRow> ()
 				.Select ((o, idx) => col == 0 ? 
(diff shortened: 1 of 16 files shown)


tznind:
Marking this ready for review as it is good to merge.  I do want to update FileDialog to use the new interface but since it was all written with DataTable in mind it will be a bit of work to back that out and probably makes sense to do that as a seperate PR.

tig (changes requested):
See my PR.

tznind:
> I am not a fan of how the columns change width as one scrolls. It's disconcerting:

This is currently 'by design'.  

- It is both for performance (not measuring all the data in the table up front)
- Defensive, it prevents 1 long value in a column pushing out all the other columns (potentially off the screen so that horizontal scrolling is required).

The current way of dealing with this is to use the `ColumnStyle` to configure min/max widths etc 

I mentioned in your PR about maybe adding something like `SizeColumnsToAllData` or something but that I think should be a seperate PR?

tig:
I get it now. I think we should add your explanation in uicatalog.cs where I put all the setup code.

tznind:
> I get it now. I think we should add your explanation in uicatalog.cs where I put all the setup code.

Cool that makes sense, how does this read?
[c767aeb](https://github.com/gui-cs/Terminal.Gui/pull/2576/commits/c767aeb957ca4cb981578ae5c964bc1aa24414dd)

tig (approved):

tig:
Oops. As soon as I merged this, i realized moving to TableView broke SearchNavigator for the scenario list. 

@tznind how hard it is to make searchnav work with tableview?

tznind:
> Oops. As soon as I merged this, i realized moving to TableView broke CollectionNavigator for the scenario list.
> 
> @tznind how hard it is to make CollectionNavigator work with tableview?

Its doable because I did a 'good enough' implementation for FileDialog (see `UpdateCollectionNavigator` method).  

But to add it as a core feature of `TableView` would require some care as we don't want to compromise performance when working with large tables.

For example if you have 1 million rows then you don't want to be regularly passing them to a new navigator just because the user changed columns.  

And you don't want to duplicate all the data at all really because it can change.

I think we would want an interface `ICollectionNavigator` and then implement `GetNextMatchingItem` to search the  `ITableSource` data directly (starting at the current offset).

Its definitely doable but requires care.

I am refactoring `FileDialog` to use `ITableSource` at the moment (its making the code much cleaner).  I can look into the collection navigation next.

tznind:
I have documented the breaking change in the breaking changes thread: https://github.com/gui-cs/Terminal.Gui/discussions/2448#discussioncomment-5758572

Issue #2581, opened 2023-04-28
Support ICollectionNavigator in TableView


tznind:
> 
> @tznind how hard it is to make CollectionNavigator work with tableview?

Its doable because I did a 'good enough' implementation for FileDialog (see `UpdateCollectionNavigator` method).  

But to add it as a core feature of `TableView` would require some care as we don't want to compromise performance when working with large tables.

For example if you have 1 million rows then you don't want to be regularly passing them to a new navigator just because the user changed columns.  

And you don't want to duplicate all the data at all really because it can change.

I think we would want an interface `ICollectionNavigator` and then implement `GetNextMatchingItem` to search the  `ITableSource` data directly (starting at the current offset).

Its definitely doable but requires care.

I am refactoring `FileDialog` to use `ITableSource` at the moment (its making the code much cleaner).  I can look into the collection navigation next.

_Originally posted by @tznind in https://github.com/gui-cs/Terminal.Gui/issues/2576#issuecomment-1528081506_

tig:
See #2584 for my workaround in UI Catalog

PR #2584, opened 2023-04-28
Builds CollectionNavigator support into UI Catalog for TableView

tig:
A quick fix until #2581 is implemented.

### DIFF
--- a/UICatalog/UICatalog.cs
+++ b/UICatalog/UICatalog.cs
@@ -378,17 +378,23 @@ namespace UICatalog {
 				// Enable user to find & select a scenario by typing text
 				// TableView does not (currently) have built-in CollectionNavigator support (the ability for the 
 				// user to type and the items that match get selected). We implement it in the app instead. 
-				ScenarioList.KeyPress += (s, a) => {
+				ScenarioList.KeyDown += (s, a) => {
 					if (CollectionNavigator.IsCompatibleKey (a.KeyEvent)) {
 						var newItem = _scenarioCollectionNav?.GetNextMatchingItem (ScenarioList.SelectedRow, (char)a.KeyEvent.KeyValue);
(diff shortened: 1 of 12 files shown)


UICatalog/UICatalog.cs
```diff
@@ -371,13 +371,31 @@ public UICatalogTopLevel ()
...
+							a.Handled = true;
+						}
+					}
+				};
 				ScenarioList.CellActivated += ScenarioView_OpenSelectedItem;
 
+				// TableView typically is a grid where nav keys are biased for moving left/right.
+				ScenarioList.AddKeyBinding (Key.Home, Command.TopHome);
+				ScenarioList.AddKeyBinding (Key.End, Command.BottomEnd);
+				ScenarioList.ClearKeyBinding (Key.A | Key.CtrlMask);
```
tznind, line 397: You can just turn `MultiSelect=false` to prevent this, you don't have to remove the Ctrl+A key.  

Also as an asside you can also multi select with Shift+Up/Down and also Ctrl+Click (if OS supports it).

tig: Setting `MultiSelect = false` does not undo the `AddKeyBinding(ctrl-a)`:

```cs
		/// <summary>
		/// True to allow regions to be selected 
		/// </summary>
		/// <value></value>
		public bool MultiSelect { get; set; } = true;
```

I think it would be reasonable for TableView to adapt it's key bindings based on whether `MutlSelect` is on or not; For now I'll just have UI Catalog clear the binding for Ctrl-A.

tig: To be clear: the reason is Ctrl-A in UI Catalog pops the about box. It could use a different shortcut, but I have muscle memory and frequently test the about box to ensure stuff is working.

PR #2586, opened 2023-04-29
Fixes #2581 Refactor CollectionNavigator so it supports TableView

### LINKED ISSUE #2581: Support ICollectionNavigator in TableView

tznind:
Fixes #2581 - Include a terse summary of the change or which issue is fixed.

So far: 
- [x] Split collection from search implementation
- [x] Make search optimistic (don't get all matches up front, instead use `while`).
- [x] Add a TableView implementation that serves collection from current cell down iteratively

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Text/CollectionNavigator.cs
+++ b/Terminal.Gui/Text/CollectionNavigator.cs
@@ -7,8 +7,7 @@ namespace Terminal.Gui {
 
 	/// <inheritdoc/>
 	/// <remarks>This implementation is based on a static <see cref="Collection"/> of objects.</remarks>
-	public class CollectionNavigator : CollectionNavigatorBase
-	{
+	public class CollectionNavigator : CollectionNavigatorBase {
 		/// <summary>
 		/// The collection of objects to search. <see cref="object.ToString()"/> is used to search the collection.
 		/// </summary>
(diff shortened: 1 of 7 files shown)


tznind:
Ok this now does most of what we want from this feature.  There's some bespoke things I did in FileDialog like ignoring the '.' when navigating file type and ignoring the 'icon' (e.g. `\` on directories).  Need a way to fix that behavior back in.

And needs some tests.

tznind:
Am happy with this PR now so have marked it ready.

PR #2595, opened 2023-05-05
Fixes #2597. Enables standard glyphs to be set via ConfigurationManager

### LINKED ISSUE #2597: Enable users to change standard glyphs
JYv2hb1 1

tig:
Fixes #2597 

As a distraction from working on the new view architecture, and because i've lost hope that we'll ever be able to ask terminals if they support particular glyphs, I decided to refactor how the standard glyphs are set and enable users to change them via config.

See https://github.com/microsoft/terminal/issues/1040#issuecomment-1535203264

## Pull Request checklist:


tig:
@BDisp @tznind etc... please take a look and review.

Terminal.Gui/Configuration/RuneJsonConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Text.Json.Serialization;
+
+namespace Terminal.Gui {
+	/// <summary>
+	/// Json converter for the <see cref="Rune"/> class. Supports
+	/// A string as one of:
+	/// - unicode char (e.g. "☑")
+	/// - U+hex format (e.g. "U+2611")
+	/// - \u format (e.g. "\\u2611")
+	/// A number
```
tznind, line 13: Nice! I like that we are flexible about what we allow.

Terminal.Gui/Views/CheckBox.cs
```diff
@@ -83,11 +83,11 @@ public CheckBox (int x, int y, ustring s, bool is_checked) : base (new Rect (x,
...
 		{
-			charNullChecked = new Rune (Driver != null ? Driver.NullChecked : '?');
-			charChecked = new Rune (Driver != null ? Driver.Checked : '√');
-			charUnChecked = new Rune (Driver != null ? Driver.UnChecked : '╴');
+			charNullChecked = Glyphs.NullChecked;
+			charChecked = Glyphs.Checked;
+			charUnChecked = Glyphs.UnChecked;
 			Checked = is_checked;
-			HotKeySpecifier = new Rune ('_');
+			HotKeySpecifier = Glyphs.HotKeySpecifier;
```
tznind, line 90: Is `HotKeySpecifier` really a Glyph?  

I thought it was more like an escape character that indicates to Terminal.Gui where to change color when rendering (and to pick up the next character as the shortcuts?).

I think if the user was to edit this to another symbol quite a lot of stuff might break in Terminal.Gui apps especially if the author just assumes that its `_` .

Terminal.Gui/Views/TextView.cs
```diff
@@ -1995,7 +1995,14 @@ void WrapTextModel ()
 		/// Indicates whatever the text was changed or not.
 		/// <see langword="true"/> if the text was changed <see langword="false"/> otherwise.
 		/// </summary>
-		public bool IsDirty => _historyText.IsDirty (Text);
+		public bool IsDirty {
+			get {
+				return _historyText.IsDirty (Text);
+			}
+			set {
+				_historyText.Clear (Text);
```
tznind, line 2003: This change is unrelated to glyphs? Could you give some insight into what it's for?

Not a problem just curious.

tznind:
Looks good.  Just a few minor points.

This is a great feature, awesome work! :rocket:

Terminal.Gui/Views/CheckBox.cs
BDisp: If all the `_` are replaced with the `Glyphs.HotKeySpecifier` there aren't any issue. That will imply the users have to change their code to $"{Glyphs.HotKeySpecifier}whatvever". But isn't a big deal with the editors replacement tools.

Terminal.Gui/Views/TextView.cs
BDisp: Well I'm also curious with this. Do you want, if the value is set to false then clear the `_historyText`? If so, there is the `ClearHistoryChanges` method for that.

BDisp:
It's a very great work, congrats.

Terminal.Gui/Views/TextView.cs
tig: Fixes a bug in config editor where i want isdirty to be cleared after saving. Will use the method BDisp suggested. Didn't see that.

Terminal.Gui/Views/CheckBox.cs
tig: Ill remove it from Glyphs. 

I included it as a mental model test and forgot to remove it.

tig:
I got carried away and went further.

- Almost all glyphs used in the project are now covered. Some more still to come
- Refactored LineCanvas to ensure when the config changes, existing LineCanvas' instances get updated.
- Much better docs of each glyph
- Revamped LineDrawing scenario
  - Support LineCanvas tracking attibutes
  - Floating tool window that uses ColorPicker and RadioGroup
  - Added "Add Layer" button vs. color change

tznind:
Maybe a static value that is changed in configs but not reset after the test or something?

```csharp
[xUnit.net 00:00:24.06]       Expected:  ╒═╕ \n╔╡1╞╗\n║╘═╛║\n╚═══╝
[xUnit.net 00:00:24.06]       Actual:    ╒═╖ \n╔╡1╞╗\n║╘═╜║\n╚═══╝
```

tig:
Typo. I blame chatgpt who I used for some of this.

tig:
It'll be fun to extend the LineDrawing scenario to do more. 

Issues 

- it should show what's gonna be drawn as the mouse drags 
- zero length lines are not handled correctly 
- switch layers 
- undo
- erase

tznind:
Yeah that would be cool!

tig:
Ok. This puppy is ready for final review.

PR #2599, opened 2023-05-06
Fixes #2601 - Adds ability to filter TreeView branches/leaves

### LINKED ISSUE #2601: Support filtering in TreeView
Filtering a tree view means not just hiding/showing branches but also considering parents/children that match the filter.  This makes most sense as a core library feature on the view rather than a hacky 'add on' that could be applied by the tree builder.

This will be handy for:
https://github.com/PowerShell/GraphicalTools/pull/179

Since v2 release seems quite far off and will likely take a while to be stable I think it makes sense for this to be in v1 and v2.

Ability to add a filter to a TreeView to show only branches that match given criteria

tznind:
Fixes #2601 - Adds ability to filter TreeView branches/leaves (e.g. free text search).

I've created this targeting `develop` as it has quite a small footprint and its a really nice to have core feature I think.

It should be pretty easy to cherry pick to `v2_develop`

Because a tree is hierarchical its important to consider whether a filter matches the parent or child too.  I went for greatest matching so:

- If a branch matches then all sub branches/leaves show even if they don't match the filter
- If a leaf/branch matches all its parents are automatically shown (to expose the matching child)

As with all things in `TableView` it is only the expanded nodes that get processed.  If a node is not expanded its children are unknown (so not considered for matching).  This is because of how the table builds itself as the user expands rather than loading everything up front (which can be impossible for some data structures - e.g. file system).

filter

Draft as now so I can write some tests

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/TreeViewTextFilter.cs
+++ b/Terminal.Gui/Views/TreeViewTextFilter.cs
@@ -1,0 +1,65 @@
+using System;
+
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// <see cref="ITreeViewFilter{T}"/> implementation which searches the
+	/// <see cref="TreeView{T}.AspectGetter"/> of the model for the given
+	/// <see cref="Text"/>.
+	/// </summary>
+	/// <typeparam name="T"></typeparam>
+	public class TreeViewTextFilter<T> : ITreeViewFilter<T> where T : class {
+		readonly TreeView<T> _forTree;
+
+		/// <summary>
+		/// Creates a new instance of the filter for use with <paramref name="forTree"/>.
+		/// Set <see cref="Text"/> to begin filtering.
+		/// </summary>
+		/// <param name="forTree"></param>
+		/// <exception cref="ArgumentNullException"></exception>
+		public TreeViewTextFilter (TreeView<T> forTree)
+		{
+			_forTree = forTree ?? throw new ArgumentNullException (nameof (forTree));
+		}
+
+		/// <summary>
+		/// The case sensitivity of the search match. 
+		/// Defaults to <see cref="StringComparison.OrdinalIgnoreCase"/>.
+		/// </summary>
+		public StringComparison Comparer { get; set; } = StringComparison.OrdinalIgnoreCase;
+		private string text;
+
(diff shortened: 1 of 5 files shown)


Terminal.Gui/Views/TreeViewTextFilter.cs
```diff
@@ -0,0 +1,65 @@
...
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// <see cref="ITreeViewFilter{T}"/> implementation which searches the
+	/// <see cref="TreeView{T}.AspectGetter"/> of the model for the given
+	/// <see cref="Text"/>.
+	/// </summary>
+	/// <typeparam name="T"></typeparam>
+	public class TreeViewTextFilter<T> : ITreeViewFilter<T> where T : class {
+		readonly TreeView<T> forTree;
```
tig, line 12: Prefix private fields with `_` please.

Terminal.Gui/Views/TreeViewTextFilter.cs
```diff
@@ -0,0 +1,65 @@
...
+
+		/// <summary>
+		/// Creates a new instance of the filter for use with <paramref name="forTree"/>.
+		/// Set <see cref="Text"/> to begin filtering.
+		/// </summary>
+		/// <param name="forTree"></param>
+		/// <exception cref="ArgumentNullException"></exception>
+		public TreeViewTextFilter (TreeView<T> forTree)
+		{
+			this.forTree = forTree ?? throw new ArgumentNullException (nameof (forTree));
```
tig, line 22: And get rid of `this.`

tznind: Done in [42e32b1](https://github.com/gui-cs/Terminal.Gui/pull/2599/commits/42e32b194b750c1fc23ec8bcb4ba819676bf740d)

tznind:
Ok after a bit of trial and git horror I have squashed. to [9c11f61](https://github.com/gui-cs/Terminal.Gui/pull/2599/commits/9c11f61d7dbcda9432cfa8a404a0e78c0bbc3eca) 

I will open another PR to cherry pick into `v2_develop`

PR #2604, opened 2023-05-07
Fixes #2601 - Adds ability to filter TreeView branches/leaves (v2)

### LINKED ISSUE #2601: Support filtering in TreeView

tznind:
This is #2599 but updated to work with `v2_develop`

### DIFF
--- a/Terminal.Gui/Views/ITreeViewFilter.cs
+++ b/Terminal.Gui/Views/ITreeViewFilter.cs
@@ -1,0 +1,14 @@
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// Provides filtering for a <see cref="TreeView"/>.
+	/// </summary>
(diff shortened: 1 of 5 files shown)


tig:
@tznind: Merge conflicts...

tznind:
> @tznind: Merge conflicts...

Have resolved the conflicts, should be good now :)

Issue #666, opened 2020-06-10
Refactor ConsoleDrivers to remove duplicate code


tig:
There is a LOT of common code across the three (now four with Fake) Drivers.

I was just thinking about how to fix bugs related to non-printable chars causing rendering issues on some terminals. For example, in Windows Terminal and Fluent if `AddRune` gets called with `\n` it does this:
![image](https://user-images.githubusercontent.com/585482/84306468-bf950280-ab18-11ea-8e37-b4d623d69152.png)

The fix is to modify `AddRune` to look for non-printable chars like this, that may cause the underlying console to take action and replace them with the right glyph. E.g:

![image](https://user-images.githubusercontent.com/585482/84306987-7d1ff580-ab19-11ea-9fe4-8dc9ad200b62.png)

However, it's stupid to do this 4 times. Instead the `base` should implement the least-common-denominator functionality. 

Right?

BDisp:
I think is already fixed with your creating of the `public static Rune MakePrintable (Rune c)` method and the `TextFormatter` class.

tig:
Woo-hoo! Finally!

It ain't perfect, but this radically simplifies and improves the ability to prove quality.

PR #2606, opened 2023-05-07
Adds "Clip Region" clipping to `ConsoleDriver`

tig:
As part of the new `View` architecture, we will need a more sophisticated clipping mechanism than just a single rectangle. 

This PR adds such a thing, retaining the previous `ConsoleDriver.Clip` functionality.

The clip region is exposed via `ClipRegion` which is a `List<Rect>`. `IsInClipRegion()` (private) tests whether a coordinate is within the region or not. 

`ConsoleDriver.IsValidClip` has been renamed to `IsValidLocation` (it was mis-named before because it also validates the coordinates are within the screen bounds).

Also:
- cleans up ConsoleDriver code

Other than ensuring I didn't break anything and adding unit tests, I've not exercised this functionality for realz yet. That will come soon, hence this is still a draft PR.


tig:
Still a draft. 

I've discovered a bunch of things in all the drivers that need to be tweaked to make this work. No way am I going to do it all 4 times. Instead, I'm going to work on #666 (which I started in the code in this PR) as a separate PR first. It's just crazy how much duplicate/similar code there are in all 4 drivers. Much can be moved into `ConsoleDriver` in a way that makes unit testing it all much better.

Issue #2610, opened 2023-05-08
Refactor `ConsoleDriver`s to support ANSI sequences natively (e.g. Virtual Terminal Sequences on Windows)


tig:
- [ ] Build a new driver that does uses the new [Console Virtual Terminal Sequences](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences) API.
- [ ] Refactor `CursesDriver` to not rely on Curses (as much?)

This may serve as the basis of a new Linux/Mac driver that does not so deeply depend on (and be constrainted by) curses.

More Useful Windows Terminal Issues: 
* https://github.com/microsoft/terminal/issues/10462
* https://github.com/microsoft/terminal/issues/10810
* https://github.com/microsoft/terminal/issues/8000
* https://github.com/microsoft/terminal/issues/3016
* https://github.com/microsoft/terminal/issues/218
* https://github.com/microsoft/terminal/issues/4949

dodexahedron:
So I've been doing a deep dive into the win32 console APIs and the frustratingly fractured/incomplete documentation available for virtual terminal sequences...

I think it's a very excellent goal for TG, and helps make things a bit more consistent, cross-platform.

But, a little voice stopped me and gave me a reality check that I just want to relay, for consideration:

Unless significant work that blocks other work has been done on it, perhaps that's something that might make sense for a VNext?

If done right, it's just internal implementation details, so we should be able to handle it in a way that doesn't affect the external API of TG. So, if following SemVer, that could even be acceptable as a point release, if desired.

Just a thought, since it's a pretty significant undertaking and there's value in a carefully thought-out approach to the implementation, so we don't have to iterate on it too much or work ourselves into avoidable pigeonholes. 🤷‍♂️

Decimation:
@dodexahedron [Spectre.Console](https://github.com/spectreconsole/spectre.console) has a robust cross-platform implementation with VTS support. It's similar to `Terminal.Gui` but is more focused on discrete shell controls.

dodexahedron:
Thanks. :)

I know @tig is aware of that library or may have even looked into it at least on some level, before.

I'm not sure what the current thoughts are on how to take advantage of that kind of thing, be it referencing, using code from parts of it, or drawing inspiration from it. I only came on board recently and have been working on other things, but this stuff is probably going to be at least incidentally relevant to what I'm working on now at some point, so it's good to keep this kind of thing in mind.

[...]

Issue #370, opened 2020-04-15
New terminal backend


migueldeicaza:
Dealing with old ncurses, versions compiled without everything is annoying.   Perhaps the time has come to just support the vt100/xterm family of terminals and have a direct driver.

tig:
Issues that will be resolved by this:

- https://github.com/migueldeicaza/gui.cs/issues/881
-

BDisp:
> Issues that will be resolved by this:
> 
>     * [Modifier key are not fully supported with CursesDriver #881](https://github.com/gui-cs/Terminal.Gui/issues/881)
> 
>     * [The native method "COLOR_PAIRS" does not exist #415](https://github.com/gui-cs/Terminal.Gui/issues/415)

These issue are already closed.

Well, I think the `NetDriver` implemented by @migueldeicaza and enhanced by me, in some way solves part of this, since it'support multi-platform.

tig:
Closing as dupe of

- #2610

PR #2612, opened 2023-05-08
Fixes #666. Refactor `ConsoleDriver`s to simplify and remove duplicated code

### LINKED ISSUE #666: Refactor ConsoleDrivers to remove duplicate code

tig:
Fixes #666 - Include a terse summary of the change or which issue is fixed.

For v2, and as part of #2606, I'm tackling this. Todo:

- [x]  `Move()` and `AddStr()` - there's almost no difference in each of the 4 driver's implementation, and the code is complex and hard to understand. 
- [x]  I have an idea that doing unicode wide manipulation in `AddStr` is the wrong place to do it and will investigate how to move that logic into `UpdateScreen`
- [x]  Each driver defines `_ccol` and `_crow` and `_contents` the same and uses them in the same way. These should be moved into `ConsoleDriver`
- [x]  Move non-driver specific color stuff to `./Drawning`
- [x] NetDriver keyboard handling is broken (e..g. multiple ESC required to quit)
- [x] NetDriver under WSL does not work - (fails to draw correctly)
- [x] Some double clicks not moving focus (WindowsDriver)
- [x] CursesDriver doesn't draw when first started (until first mouse move)
- [x] CursesDriver hangs when moving to ArabicPresentaitonFormsB
- [ ]  Add more primitive unit tests, especially for unicode handling

These issues are NOT fixed in this PR and need to be addressed separately:

- [ ] CursesDriver doesn't render CharMap correctly for wide chars (and other Unicode) - Curses is doing something funky with glyphs that report GetColums() of 1 yet are rendered wide. E.g. 0x2064 (invisible times) is reported as 1 column but is rendered as 2. WindowsDriver & NetDriver correctly render this as 1 column, overlapping the next cell.
- [ ] Refactor Color/True Color
- [ ] Add TrueColor support to CursesDriver
- [ ] Add TrueColor support to NetDriver

## Pull Request checklist:


tig:
@BDisp take a look...

Right now, the only driver that's really using my new codepath is FakeDriver. When I read your message about working on the same thing, I was just about to refactor `AddRune` to not be virtual so there was one implementation that looks something like this:

```cs
	public override void AddRune (System.Rune systemRune)
	{
		var rune = new Rune (systemRune).MakePrintable ();
		var runeWidth = rune.GetColumnWidth ();
		var validLocation = IsValidLocation (Col, Row);

		if (validLocation) {
			if (rune.IsCombiningMark () && Col > 0) {
				// Decode the previous rune
				var previousRune = new Rune (Contents [Row, Col - 1, 0]);
				var newCombo = new StringBuilder ();
				ReadOnlySpan<char> remainingInput = previousRune.ToString ().AsSpan ();
				while (!remainingInput.IsEmpty) {
					// Decode
					OperationStatus opStatus = Rune.DecodeFromUtf16 (remainingInput, out Rune result, out int charsConsumed);

					if (opStatus == OperationStatus.DestinationTooSmall || opStatus == OperationStatus.InvalidData) {
						result = Rune.ReplacementChar;
					}

					newCombo.Append (result);
					// Slice and loop again
					remainingInput = remainingInput.Slice (charsConsumed);
				}
				newCombo.Append (rune);

				var combined = newCombo.ToString ();
				var normalized = !combined.IsNormalized () ? combined.Normalize () : combined;
				Contents [Row, Col - 1, 0] = normalized [0];// BUGBUG: This is wrong, we need to handle the case where the rune is more than one char
				Contents [Row, Col - 1, 1] = CurrentAttribute;
				Contents [Row, Col - 1, 2] = 1;
				Col--;
			} else {
				Contents [Row, Col, 0] = rune.Value;
				Contents [Row, Col, 1] = CurrentAttribute;

				if (Col > 0) {
					var left = new Rune (Contents [Row, Col - 1, 0]);
					if (left.GetColumnWidth () > 1) {
						Contents [Row, Col - 1, 0] = Rune.ReplacementChar.Value;
					}
				}

				if (runeWidth > 1) {
					Col++;
					Contents [Row, Col, 0] = Rune.ReplacementChar.Value;
					Contents [Row, Col, 1] = CurrentAttribute;
					Contents [Row, Col, 2] = 1;
				}
			}
		}
		Col++;
	}
```

To enable `CursesDriver` to render to curses at `AddRune` time, vs in `UpdateScreen`, my plan was to add new virtual methods (e.g. `OnContentsChanged(col, row, Rune, Attribute, Dirty)`.

One thing I just learned is having `Contents` just be an array of tuples of ints is we'll never be able to support combining sequences that don't normalize into a single unicode char. See `ContentsTests.AddStr_With_Combining_Characters` for an example that will never work.

To address this, I propose we change the way `Contents` is defined to be a two dimensional array of structs like this:

```cs
struct ContentsCell {
   string runes;
   Attribute attr;
   bool Dirty;
}
``` 

Also, look at how I refactored things, including the `System.Text.Rune` extension methods.

tig:
See https://github.com/gui-cs/Terminal.Gui/issues/2616

BDisp:
You are doing well. I think with your focused on the `ConsoleDriver` and I with focused on the `System.Text.Rune` will result on the right combination.

tig:
@BDisp please do me a favor and if you make any more changes to any of the ConsoleDrivers, please submit them as PRs to this PR. I've done some significant refactoring and merging from changes to v2_develop are challenging. Thanks.

BDisp:
@tig all logic I added on the `WindowsDriver` is broken here. See in the gif bellow that the replacements chars you added here was on `BMP` code points which has double width. I'll try to make the `WindowsDriver` working with `non-BMP`, which I already did in the `v2_develop` branch, but the cursor doesn't fit the `non-BMP` length. When I finished more or less I'll target the v2_develop and then you merge in this, because it's confused to me.

windowsdriver-broken

tig:
Don't bother...yet. 

I'm rearchitecting the whole thing. 

Please give me another week or so.

BDisp:
Ok @tig. Thanks. The `ContentsCell` idea is great. There is much duplicate code on all drivers that can be put on in the `ConsoleDriver`. It only need do call overridden methods to get/set the rune value in the right places (current, previous or next) and get/set the right colors (current, previous or next), where each driver handles differently. But the conditions can be done all on the AddRune (virtual, not abstract). But I also understand it's very complex too :-)

tig:
> Ok @tig. Thanks. The `ContentsCell` idea is great. There is much duplicate code on all drivers that can be put on in the `ConsoleDriver`. It only need do call overridden methods to get/set the rune value in the right places (current, previous or next) and get/set the right colors (current, previous or next), where each driver handles differently. But the conditions can be done all on the AddRune (virtual, not abstract). But I also understand it's very complex too :-)

There are the following areas where the four drivers need to differ, and my thinking about simplifying each:

1) Emitting changes to `Contents`. Exactly as you describe above; the essence is "`AddStr` modifies `Contents` and `UpdateScreen" reflects any changes since its last call to the platform-specific console APIs. See Note below.
2) Mapping Terminal.Gui's 'Color' to the platform specific APIs. I'm working to reduce the code in each driver in this regard to only the algorithms needed to do this mapping. As I'm doing this I'm trying to ensure adding TrueColor and other styling will be easier. 
3) Cursor style and visibility APIs. I still have more thinking to do here, but there is duplicate code and some of it seems needlessly complex. I'm not convinced the Terminal.Gui Cursor API is ideally designed, but other than NetDriver flickering, it does work.
4) Mouse input handling. I haven't dug into this deeply yet, but man is there a lot of code that seems duplicatlve.
5) Keyboard input handling, I think we can radically simplify this. Terminal.Gui has way too many definitions of a "Key" (`Key`, `ConsoleKey`, `ConsoleKeyInfo`, etc...) and there is a ton of code for mapping between all of these where there seems to be a lot of duplicate logic. I may pun this to a future PR as I simply don't understand a lot of it. 
6) Mapping platform-specific thread synchronization and timers. I'm not planning on touching any of this...when it comes to threading "THAR BE DRAGONS!"

** Note on making `AddRune` non-virtual: I'm worried about two things

a) At one point the drivers had an "immediate mode" where calling AddStr caused the platform APIs to be called synchronously. The idea being, having the console immediately reflect the change. At some point this code became disabled, and all code-paths end up having `UpdateScreen` getting called only via a MainLoop iteration (thus AddStr is asynchronous relative to the console actually changing). I do think there are times when an immediate mode are useful (esp in debugging) and I'd like to see if we can add it back in. 

b) Parsing `Contents` for what has changed (even if it is more encapsulated via `Cell`) may result in poor performance if the underlying API is ANSI escape sequence-stream based. For the current WindowsDriver and NetDriver this may be less of an issue because they provide "direct to buffer" APIs. But Curses doesn't actually do this (in reality it is using ANSI escape sequences under the covers). If we implement #2610 (add a Windows Driver that uses Virtual Terminal Sequences), it will too act in this way. The trick here will be making `UpdateScreen` smart enough to consolidate and coalesce runs of styling such that megabytes of "\0x[x;yH\0x[r;g;bm" that are duplicative don't get emitted.

I now think I understand why Miguel added #370 (New Terminal Backend) to his todo list way back when: He realized the logic for emitting the right ESC sequences should be in Terminal.Gui, and once that happened, nCurses just gets in the way.

Another way of saying this, I think, is: We can build a single implementation that takes changes from `Contents` and emits ESC sequences and it will work on BOTH Windows and Linux now that Windows has a real terminal that actually (mostly) conforms to ANSI esc sequence standards.

BDisp:
> There are the following areas where the four drivers need to differ, and my thinking about simplifying each:
> 
>     1. Emitting changes to `Contents`. Exactly as you describe above; the essence is "`AddStr` modifies `Contents` and `UpdateScreen" reflects any changes since its last call to the platform-specific console APIs. See Note below.
> 
See my response there too.

>     2. Mapping Terminal.Gui's 'Color' to the platform specific APIs. I'm working to reduce the code in each driver in this regard to only the algorithms needed to do this mapping. As I'm doing this I'm trying to ensure adding TrueColor and other styling will be easier.
> 
TrueColor would be great.

>     3. Cursor style and visibility APIs. I still have more thinking to do here, but there is duplicate code and some of it seems needlessly complex. I'm not convinced the Terminal.Gui Cursor API is ideally designed, but other than NetDriver flickering, it does work.
> 
That why I hide the cursor on the `NetDriver` while `UpdateScreen` is executing to avoid the cursor flickering to be more visible, but you commented that code. In `Windows` you can´t reduce that more it's now. That why I decided use the `StringBuilder` class to write the minimum of times, because before, when each char was write to console the flickering was much more visible.

>     4. Mouse input handling. I haven't dug into this deeply yet, but man is there a lot of code that seems duplicatlve.
> 
Yes there are but also each driver handles differently mouse events. Some of them don't send a click event, so we need to create a click event when a released button is captured in the same location when it was pressed. Only the `NetDriver` receives the right mouse events because the internal `NetEvents` does all the heavy stuff.

>     5. Keyboard input handling, I think we can radically simplify this. Terminal.Gui has way too many definitions of a "Key" (`Key`, `ConsoleKey`, `ConsoleKeyInfo`, etc...) and there is a ton of code for mapping between all of these where there seems to be a lot of duplicate logic. I may pun this to a future PR as I simply don't understand a lot of it.
> 
Well this can be more and less identical between WindowsDriver and NetDriver in the response but not on the input, where they are truly different. `CursesDriver` input is also different but as it's escape sequence compatible, much of `EscSeqUtils` methods are used.

>     6. Mapping platform-specific thread synchronization and timers. I'm not planning on touching any of this...when it comes to threading "THAR BE DRAGONS!"
> 
Yes this part is very tricky.
> 
> ** Note on making `AddRune` non-virtual: I'm worried about two things
> 
> a) At one point the drivers had an "immediate mode" where calling AddStr caused the platform APIs to be called synchronously. The idea being, having the console immediately reflect the change. At some point this code became disabled, and all code-paths end up having `UpdateScreen` getting called only via a MainLoop iteration (thus AddStr is asynchronous relative to the console actually changing). I do think there are times when an immediate mode are useful (esp in debugging) and I'd like to see if we can add it back in.
> 
That why the drivers has a `sync` private member for that. Has you can imagine this is private because "immediate mode" can be useful for debugging, but not for a release config. If NetDriver is flickering so much, then if `sync` is true it will flickering much more, because it will write to the console when each char is added to the buffer and every time it's clear a view, draw a view. Well it'll a hell, even for debugging, believe me. I've started a new method for using in the `TestHelpers` that read from the `FakeConsole` buffer, instead of the driver `Contents`. For all this reason the buffer must be write to the console asynchronously, after the buffer has all write and the `Refresh` method will run `UpdateScreen` and `UpdateCursor`. Although it's asynchronous the are executed in sequence and not alternated.

> b) Parsing `Contents` for what has changed (even if it is more encapsulated via `Cell`) may result in poor performance if the underlying API is ANSI escape sequence-stream based. For the current WindowsDriver and NetDriver this may be less of an issue because they provide "direct to buffer" APIs. But Curses doesn't actually do this (in reality it is using ANSI escape sequences under the covers). If we implement #2610 (add a Windows Driver that uses Virtual Terminal Sequences), it will too act in this way. The trick here will be making `UpdateScreen` smart enough to consolidate and coalesce runs of styling such that megabytes of "\0x[x;yH\0x[r;g;bm" that are duplicative don't get emitted.
> 
Don't ever think about adding escape sequences to the `Contents`. It only take location, runes and attributes. We can make this only as part of the `ConsoleDriver` which is common to all drivers. For `WindowsDriver` deal with escape sequences is the same way how the NetDriver active this in the `NetWinVTConsole`, but `WindowsDriver` use `WriteConsoleOutput` which only will write the `ScreenBuffer` contents. Unless you use another API where you can control the output and write to the console a StringBuilder object.

> I now think I understand why Miguel added #370 (New Terminal Backend) to his todo list way back when: He realized the logic for emitting the right ESC sequences should be in Terminal.Gui, and once that happened, nCurses just gets in the way.
> 
Yes I believe, so. I do even think, maybe Miguel can confirm or not, that was the reason he began the `NetDriver` for this purpose.

> Another way of saying this, I think, is: We can build a single implementation that takes changes from `Contents` and emits ESC sequences and it will work on BOTH Windows and Linux now that Windows has a real terminal that actually (mostly) conforms to ANSI esc sequence standards.

Well the ideal is only using one driver that can be compatible on Windows and Linux, like `NetDriver`. But this one has pros and cons on both OS. On `Windows` is so slowing to handle escape sequences. I think it's time to `dotnet` guys perform a refactoring System.Console that not depends on Win32 API to create `Virtual Terminal Sequences` or create a new API that is at root more compatible escape sequences. On `Linux` many of the `Console` properties aren't allowed, but this is the minor of the problems because in return we can leverage the escape sequences to get/set this properties.

tig:
> Don't ever think about adding escape sequences to the `Contents`. 

LOL. 100000% agree. What I'm thinking is `Contents` will be like this (pseudo code):

```cs
using System.Text;


namespace Terminal.Gui; 

/// <summary>
/// Represents a single row/column in a Terminal.Gui rendering surface
/// (e.g. <see cref="LineCanvas"/> and <see cref="ConsoleDriver"/>).
/// </summary>
public class Cell {
	/// <summary>
	/// The list of Runes to draw in this cell. If the list is empty, the cell is blank. If the list contains
	/// more than one Rune, the cell is a combining sequence.
	/// (See #2616 - Support combining sequences that don't normalize)
	/// </summary>
	public List<Rune> Runes { get; set; }

	/// <summary>
	/// The attributes to use when drawing the Glyph.
	/// </summary>
	public Attribute? Attribute { get; set; }

	/// <summary>
	/// Gets or sets a value indicating whether this <see cref="T:Terminal.Gui.Cell"/> has
	/// been modified since the last time it was drawn.
	/// </summary>
	public bool IsDirty { get; set; }
}
```

Then, in `ConsoleDriver`:

```cs
	/// <summary>
	/// The contents of the application output. The driver outputs this buffer to the terminal when <see cref="UpdateScreen"/>
	/// is called.
	/// <remarks>
	/// The format of the array is rows, columns. The first index is the row, the second index is the column.
	/// </remarks>
	/// </summary>
	public Cell [,] Contents { get; internal set; }

```

> It only take location, runes and attributes. We can make this only as part of the `ConsoleDriver` which is common to all drivers. For `WindowsDriver` deal with escape sequences is the same way how the NetDriver active this in the `NetWinVTConsole`, but `WindowsDriver` use `WriteConsoleOutput` which only will write the `ScreenBuffer` contents. Unless you use another API where you can control the output and write to the console a StringBuilder object.

Correct. 

Regarding `_sync` - I'm not convinced we can't expose and API that lets developers turn this on and off. The current codebase removed it, but that was done as a temporary measure to make the code more readable for me as I tried to understand it. 

FWIW, in other places where I've commented out code (e.g. in the cursor stuff) what you're seeing is one of my techniques to really understand how something works. I try to add "BUGBUG" or "TODO" comments when I do this to remind me to revisit it later, but sometimes I forget.

tig:
See 1st post for list of todo's for this PR.

UnitTests/Text/UnicodeTests.cs
```diff
@@ -41,10 +41,10 @@ public void AddRune_On_Clip_Left_Or_Right_Replace_Previous_Or_Next_Wide_Rune_Wit
...
 │これは広いルーンラインです。│
 │これは広いルーンラインです。│
-│これは ┌────────────┐ です。│
-│これは │ワイドルーン│ です。│
-│これは │  {CM.Glyphs.LeftBracket} 選ぶ {CM.Glyphs.RightBracket}  │ です。│
-│これは └────────────┘ です。│
+│これは�┌────────────┐�です。│
+│これは�│ワイドルーン│�です。│
+│これは�│  {CM.Glyphs.LeftBracket} 選ぶ {CM.Glyphs.RightBracket}  │�です。│
+│これは�└────────────┘�です。│
```
BDisp, line 47: Originally when a surrogate pair hasn't enough space to display I replaced it with a space. Now you are using the replacement rune. Is this temporary or you are think let it as is now? It's a matter of personal choice but I only I'm asking if it was for a test or a definitive choice.

BDisp (approved):
@tig I only left a question. See bellow, thanks.

UnitTests/Text/UnicodeTests.cs
tig: I am not sure what the right thing to do is. 

In some cases (e.g. when debugging), it's helpful to have the replacement char shown. But in real-world cases, it may look better to have spaces.

tig: For now, I'm reverting to show the replacement char

PR #2613, opened 2023-05-08
Fixes #2617 - Support for Nerd icons for FileDialog

### LINKED ISSUE #2617: Support same icons as `Terminal-Icons`
Here is a quick stab https://github.com/gui-cs/Terminal.Gui/pull/2613

If you can convert this to an issue I can link it to the above PR.

_Originally posted by @tznind in https://github.com/gui-cs/Terminal.Gui/discussions/2611#discussioncomment-5842639_

tznind:
Adds Nerd font support to file dialog.  To opt into this users should set the following flag on their default theme  (and install the required fonts in the terminal).

```json
{
  "$schema": "https://gui-cs.github.io/Terminal.Gui/schemas/tui-config-schema.json",
  "Themes": [
    {
      "Default": {
        "NerdFonts.Enable": true
      }
    }
    ]
}
```
_\~/.tui/MyProgram.config.json_

- [x] Add extensions and filenames to unicode mapping
- [x] Tests to ensure no bad dictionary keys
- [x] Credit the original repo in source
- [x]  Credit the original repo in FileDialog Scenario
- [x] Make nerd font usage be dependent on config setting.

shot-2023-05-08_22-52-46

Fixes #2617 - Adds support for using Nerd console icons for files/folders in FileDialog.

## Pull Request checklist:


tznind:
I've added icons to the tree view too.  In writing this I realized that that tree did not use `System.IO.Abstractions` so I've had to do a bit more wiring (but good wiring).  Also this fixes directories that are expanded showing the full path all the time (wasting horizontal space).

shot-2023-05-09_22-35-13

Nutzzz:
It might be nice if you could pass the UI state of the current dir to GetNerdIcon() (i.e., focused, selected, expanded/collapsed tree) so the open folder, checked file/folder, folder with plus, etc. could be used.

tznind:
I've added context to IconGetter delegate.  I made a class for it since it might grow later on and don't want to just have an ever increasing `Func<x,y,z>`.

Its a subtle effect but nice.  As you open folders they switch to the 'open folder' icon (as do parents).

shot-2023-05-09_22-35-13

tznind:
This is now good for review! :rocket:

tznind:
Updated class diagram for FileServices

image

Terminal.Gui/Configuration/ConfigurationManager.cs
```diff
@@ -235,6 +235,14 @@ public bool Apply ()
...
 			JsonPropertyName ("Glyphs")]
 		public static GlyphDefinitions Glyphs { get; set; } = new GlyphDefinitions ();
+
+		/// <summary>
+		/// If <see langword="true"/>, enables the use of Nerd unicode symbols.
+		/// This requires specific font(s) to be installed on the users machine
+		/// to work correctly.  Defaults to <see langword="false"/>.
+		/// </summary>
+		[SerializableConfigurationProperty (Scope = typeof (SettingsScope), OmitClassName = true)]
+		public static bool UseNerd { get; set; } = false;
```
tig, line 245: I didn't want CM to fill up with setting definitions. I put Glyphs here because i felt neither `Application` or `ConsoleDriver` was ideal.

Maybe making them ThemeScope properties makes more sense?

tig (approved):
If you ahve strong opinions on where the `NerdFont` setting should go, please say so.

For now we can leave it where you put it...

Terminal.Gui/Configuration/ConfigurationManager.cs
tznind: I wasn't sure where it should go either! But yeah theme sounds like a sensible place for it. I will move it there when I get a chance.

tznind:
Ok I think I have done this right but do not have time to test it out right now.  I think now the user should be able to do:

```json
"Default": {
	 "NerdFonts.Enable":true
	 "ColorSchemes": [
	 {
	      ...
```

tznind:
@tig I have the following json which I think should work (see below).  But I get the following error:

> Terminal.Gui ConfigurationManager encountered the following errors while deserializing configuration files:
Error deserializing ~/.tui/UICatalog.config.json: Expected a JSON object, but got "True".

```json
{
  "$schema": "https://gui-cs.github.io/Terminal.Gui/schemas/tui-config-schema.json",
  "Themes": {
     "Default": {
       "NerdFonts.Enable": true
     }
  }
```

Any ideas what I am doing wrong?

The attribute on NerdFonts was:

```csharp
namespace Terminal.Gui {
	internal class NerdFonts {

		[SerializableConfigurationProperty (Scope = typeof (ThemeScope))]
		public static bool Enable { get; set; } = false;
   [...]
```

tig:
It's a bug in CM. I don't understand it yet, but def a bug.

tznind:
> It's a bug in CM. I don't understand it yet, but def a bug.

Ah ok cool I thought I was doing something wrong.

Ok well I will leave it how it is then and mark this as ready again.  Once the bug is fixed the config will work as intended.

And I think this is a good place for the flag (in NerdFonts) and the scope.

tig:
Ugh. Json isn't as bad as XML, but sometimes just as frustrating.

You're missing a set of braces...

It should look like this:

```cs
  "Themes": [
    {
      "Default": {
        "NerdFonts.Enable": true
      }
    },
    {
      "UI Catalog Theme": {
```

tznind:
Gah! Thanks.  Would never have worked that out.

I've updated the comment in this PR to have correct json.

tig:
Remember, you can do this to enable exceptions for debugging these things (not that it's easy to debug):

```cs

  // Set this to true in a .config file to be loaded to cause JSON parsing errors
  // to throw exceptions. 
  "ConfigurationManager.ThrowOnJsonErrors": false,
```

tig (approved):

tig:
I didn't test this sufficiently before merging. 

- `NerdFonts.cs` should be in ./Text
- Changing to false doesn't seem to disable them
- Ideally, all theme properites get updated at runtime if the config changes and users can see the change without having to re-run the app. For FileDialog it's probably not as big a deal because once changing the setting works any re-opening of the dialog will show the changed theme. But if someone was using this in their main app it'd be great if all properties were updated at runtime. 

image
image

image
image

tig:
This should use `CM.Glyphs...`

image

tznind:
> Changing to false doesn't seem to disable them

There are 2 ways of setting nerd icon use in FileDialog

- In json config
- Explicitly calling `UseNerdForIcons` on `FileDialogStyle`

The FileDialogExample scenario uses the second way (based on which radio button the user selects).  This lets the user try out the icons just pressing ui buttons without having to edit config files.

If you use any other file dialog (e.g. open/save in csv scenario)  you can see the config setting applies:

jsonsetting

tznind:
> if someone was using this in their main app it'd be great if all properties were updated at runtime.

That would be cool.  Is there a configuration changed event?

tig:
> > if someone was using this in their main app it'd be great if all properties were updated at runtime.
> 
> That would be cool. Is there a configuration changed event?

```cs
	/// <summary>
	/// Event fired when an updated configuration has been applied to the  
	/// application.
	/// </summary>
	public static event EventHandler<ConfigurationManagerEventArgs>? Applied;
```

PR #2620, opened 2023-05-10
Fixes #92. Remove dependency on ustring.

### LINKED ISSUE #92: Remove dependency on ustring

BDisp:
Fixes #92 - This is a huge PR which remove the dependency of the `NStack` library and thus all `ustring` was renamed as `string`. The `Rune` library used is from the `System.Text` namespace.

## Pull Request checklist:


tig (changes requested):
Comments as I proceed in my review...

1) You've copied a bunch of helper APIs from nstack that are not used in our library (except in unit tests). I don't think this is a good idea. If we don't have a specific, real-world, use-case for an API it should not be added. The philosophy here is similar to premature performance optimization; it's almost always a bad idea. I will be deleting all of those in my PR.
2) There's no need for both `RuneExtensions` and `RuneUtilities`. I think everything should just hang off `RuneExtensions`.
3) Tons of unit test! Nice! However, please use [Theory] etc... for unit tests instead of just duplicating test code. It makes it very hard to read, change, and expand the tests. 
4) You've mixed Rune tests with String tests. Tests for specific classes/functionality should be as tightly focused on that class, and as simple as, as possible. For example, there's no test of `String.ConsoleWidth` (which I'm renaming to `String.GetColumns`) that tests what happens if the string.IsEmptyOrNull. To demonstrate what I  mean I've added `StringTests.cs` with some examples.
5) ALL the variants of `StringExtensions.Make` seem unnecessary now.  I think they should be removed and code should just do the natural thing with `string` and `Rune` (except maybe `Make(List<Rune> runes` which is handy).

e.g. 
```cs
textToDraw = StringExtensions.Make (new Rune [] { nullCheckedChar, (Rune)' ' }) + item.Title;
sumConsoleWidth += StringExtensions.Make (i).GetColumns ();
```

can just be

```cs
textToDraw = nullCheckedChar + ' ' + item.Title;
sumConsoleWidth +=new Rune(i).GetColumns ();
```

etc...

6) Related, `TextFormatter.GetTextWidth()` can be removed, and calls replaced with `Rune.GetColumns()`

Also, I've added some of the changes I made in #2612 to make merging easier for me later.

This is really fantastic work @BDisp! you rock.

BDisp:
> `textToDraw = nullCheckedChar + ' ' + item.Title;`
>
This isn't valid because `nullCheckedChar` is a `Rune` type and should be:
`textToDraw = nullCheckedChar.ToString() + ' ' + item.Title;`

We have to decide what to do.

tig:
> > `textToDraw = nullCheckedChar + ' ' + item.Title;`
> 
> This isn't valid because `nullCheckedChar` is a `Rune` type and should be: `textToDraw = nullCheckedChar.ToString() + ' ' + item.Title;`
> 
> We have to decide what to do.

Right; my error. I think the best syntax is:

```cs
textToDraw = $"{nullCheckedChar} {item.Title}";
```

But this is perfectly fine too:

```cs
textToDraw = nullCheckedChar.ToString() + ' ' + item.Title;
```

My point is there's really no need for `Make` in either of these cases.

BDisp:
One thing come on mind is to change all glyphs to string. This will be more easy to manipulate with text. What do you think?

BDisp:
> One thing come on mind is to change all glyphs to string. This will be more easy to manipulate with text. What do you think?

I only asked because some glyphs may be set with more than one code point, like the Man, Woman and Girl emoji 👨‍👩‍👧, which is presented with `"\U0001F468\u200D\U0001F469\u200D\U0001F467"` string and cannot be displayed with only one code point.

Since the `System.Text.Rune` overrides the `ToString` then this `textToDraw = $"{nullCheckedChar} {item.Title}";` is enough because the `ToString` is run internally with the syntax.

I'm going to change the `Make` method name to the `String` and only maintaining the  `ToString (Rune [] runes)`, `ToString (List<Rune> runes)` and the `ToString (byte [] bytes, Encoding encoding = null)`, because they have some complexity to be managed always when it's needed.

BDisp:
@tig I had to have to hack the `InternalInit` method to be possible to use the `NetDriver` in the `UICatalog. The `ConfigurationManager` always load the `UseSystemConsole` as `false` even I pass it as `true` in the command line.

tig:
> > One thing come on mind is to change all glyphs to string. This will be more easy to manipulate with text. What do you think?
> 
> 
> 
> I only asked because some glyphs may be set with more than one code point, like the Man, Woman and Girl emoji 👨‍👩‍👧, which is presented with `"\U0001F468\u200D\U0001F469\u200D\U0001F467"` string and cannot be displayed with only one code point.
> 
> 
> 
> Since the `System.Text.Rune` overrides the `ToString` then this `textToDraw = $"{nullCheckedChar} {item.Title}";` is enough because the `ToString` is run internally with the syntax.
> 
> 
> 
> I'm going to change the `Make` method name to the `ToString` and only maintaining the  `ToString (Rune [] runes)`, `ToString (List<Rune> runes)` and the `ToString (byte [] bytes, Encoding encoding = null)`, because they have some complexity to be managed always when it's needed.

Good call. Please open an issue and I'll do that!

tig:
> @tig I had to have to hack the `InternalInit` method to be possible to use the `NetDriver` in the `UICatalog. The `ConfigurationManager` always load the `UseSystemConsole` as `false` even I pass it as `true` in the command line.

Thanks. I saw that too and the right fix is to remove the setter in the Terminal.gui.resources.config.json.

tig:
@BDisp is this ready to merge?

BDisp:
> @BDisp is this ready to merge?

Yes @tig

tig:
See https://github.com/BDisp/Terminal.Gui/pull/143

tig (approved):

PR #2658, opened 2023-05-22
Fixes #2657. Rune parameter in ConsoleDriver.IsRuneSupported is never invalid

### LINKED ISSUE #2657: Rune parameter in ConsoleDriver.IsRuneSupported is never invalid
This method was created to avoid non-bmp code points from be displayed on the drivers that overridden them. If a `Rune` isn't valid an exception will be throw before run it and thus no check is needed. `System.Text.Rune` is more deterministic than the previous Rune from the `NStack` Api and never will create a rune from a invalid parameter. It only have to return true or at least also check for values less than 0. So, `return Rune.IsValid (rune.Value)` is enough.

BDisp:
Fixes #2657 - Using `Rune.IsValid` is the best method to check for a valid rune or not.

## Pull Request checklist:


Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -759,18 +759,15 @@ public abstract class ConsoleDriver {
 		/// does not support displaying this rune.</returns>
 		public virtual bool IsRuneSupported (Rune rune)
 		{
-			if (rune.Value > RuneExtensions.MaxUnicodeCodePoint) {
-				return false;
-			}
-			return true;
+			return Rune.IsValid (rune.Value);
```
tig, line 762: We should just nuke the IsRuneSupported API and have callers use Rune.IsValid(rune.Value) instead.

tig (changes requested):

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
BDisp: @tig I agree nuke the `IsRuneSupported` but also there is no need to have this callers in `AddRune` methods because the parameter of type `Rune` is always valid if this method is executed. Where can happen throwing an exception is when an API call `Driver.AddRune(invalid rune)` but never inside of the `AddRune` method. Please say something after you read this. Thanks.

tig:
I will fix this in #2621. We can close this.

tig:
> I will fix this in #2621. We can close this.

Note, to be clear, I put this API in for a slightly different purpose. Calling it from AddRune is not needed, however, there ARE cases where app code needs to know if the current console driver supports rendering a particular glpyh correctly. See the Snake code in #2621.

So I won't be removing this API, but clarifying its purpose.

BDisp:
> > I will fix this in #2621. We can close this.
> 
> Note, to be clear, I put this API in for a slightly different purpose. Calling it from AddRune is not needed, however, there ARE cases where app code needs to know if the current console driver supports rendering a particular glpyh correctly. See the Snake code in #2621.
> 
I think you have write the wrong #number.

> So I won't be removing this API, but clarifying its purpose.

Yes I understand. You just have to return if the driver supports non-BMP in the overridden methods and always true in the virtual method.

Issue #2616, opened 2023-05-09
Support combining sequences that don't normalize


tig:
See this discussion for more: https://github.com/gui-cs/Terminal.Gui/discussions/2939

What's broken today (v2):

* Combining Marks that don't normalize are not supported
  * `/u0301` at column 0 renders `́ ` even though `/u0301` is a zero-width codepoint.
  * `a/u0301` works because `ConsoleDriver` normalizes it `á`
  * `a/u301/u0301` doesn't work. `ConsoleDriver` normalizes it to `á` and ignores the 2nd `/u0301`.
  * TextField/TextView` strip out CMs on paste.
* non-BMP characters 
  * `WindowsDriver`
    * Not supported by Windows Terminal when using the `WriteConsoleOutput` (what `WindowsDriver` currently does). `WindowsDriver` currently outputs non-BMP characters as `Rune.ReplacementChar`
  * `CursesDriver` 
    * Single-width non-BMP chars output as a space
    * Wide non-BMP characters are output as `*`
  * `NetDriver` 
    * Single-width non-BMP chars correctly
    * Wide non-BMP characters chars output correctly

Note, `TextView/TextField` and other input views do not handle any of this stuff correctly (mostly).

Related:

* https://github.com/gui-cs/Terminal.Gui/issues/2610
*

tig:
@BDisp Whats your latest thinking on this issue? It's been a long time since I thought about this and I've lost track of the progress you've made on this topic. Is PR #2616 still relevant? Thanks!

BDisp:
> [@BDisp](https://github.com/BDisp) Whats your latest thinking on this issue? It's been a long time since I thought about this and I've lost track of the progress you've made on this topic. Is PR [#2616](https://github.com/gui-cs/Terminal.Gui/issues/2616) still relevant? Thanks!

Everything that could be done regarding this issue has already been fully resolved with the implementation of graphemes and therefore, this matter can now be closed.

tig:
Right on! Thanks.

PR #2664, opened 2023-05-22
Fixes #2663 - No more nested classes

### LINKED ISSUE #2663: Per [CA1034](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1034) nested types should not be public.  I would recommend moving those public nested types out from the parent class into the namespace as would a normal class would be.
Per [CA1034](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1034) nested types should not be public.  I would recommend moving those public nested types out from the parent class into the namespace as would a normal class would be.

This is the first project I've encountered with public nested types which did cause some difficulty and confusion during implementation.

_Originally posted by @ThomasNieto in https://github.com/gui-cs/Terminal.Gui/discussions/1940#discussioncomment-5967710_

tznind:
Fixes #2663 - Promote all nested classes to full classes

In #2406 many classes were moved from nested to root level.  This PR explores moving all classes (except those in SpinnerStyle.cs) out to their own classes.

In some cases e.g. `Curses.Window` I named the new class `CursesWindow` which is I think a nice way of grouping things together.  See also e.g. `Bar` (previously `BarSeries.Bar`) is now `BarSeriesBar`

## Pull Request checklist:


Terminal.Gui/Views/TableView/TableStyle.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System.Collections.Generic;
+
+namespace Terminal.Gui {
+
+	public partial class TableView {
```
ThomasNieto, line 5: Looks like `TableStyle` is still nested in `TableView`.

tznind: Yup I have set `<WarningsAsErrors>CA1034</WarningsAsErrors>` and found a bunch more which I've changed to full classes in [915ec1b](https://github.com/gui-cs/Terminal.Gui/pull/2664/commits/915ec1bb74cfbf796bc38e8d7251fe527d7f92c9) :)

tig (changes requested):
Please back out the changes to the ConsoleDrivers and related code. I will take care of this in #2612; it's diverged enough that merging from this PR will be a PITA. Thanks.

tznind:
> Please back out the changes to the ConsoleDrivers and related code. I will take care of this in #2612; it's diverged enough that merging from this PR will be a PITA. Thanks.

I've reverted everything in the `ConsoleDrivers` folder to match `v2_develop`.

I have commented out `<WarningsAsErrors>CA1034</WarningsAsErrors>` (in the .csproj file) so you can work on the console classes without seeing errors about nesting.  But I think it is nice to have as it prevents us creating public nested classes in future (will result in build failure) so if we could turn it back on after the console refactor that would be good.

You can still suppress in source:

```csharp
#pragma warning disable CA1034 // Nested types should not be visible

[...]

#pragma warning restore CA1034 // Nested types should not be visible
```
```

File: Terminal.Gui/ApplicationRunState.cs [hunk 2]
```diff
@@ -0,0 +1,79 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Terminal.Gui {
+	/// <summary>
+	/// The execution state for a <see cref="Toplevel"/> view.
+	/// </summary>
+	public class ApplicationRunState : IDisposable {
```
tig, line 8: For terseness-sake, I think this should remain named RunState.

Terminal.Gui/Configuration/Scope.cs
```diff
@@ -3,209 +3,74 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
-using System.Text.Json;
-using System.Text.Json.Serialization;
 using static Terminal.Gui.ConfigurationManager;
 
 #nullable enable
 
 namespace Terminal.Gui {
```
tig, line 10: While you're in  here, can you please change this to use `namespace Terminal.Gui;`?

Terminal.Gui/Configuration/ScopeJsonConverter.cs
```diff
@@ -0,0 +1,140 @@
+﻿using System;
+using System.Linq;
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+#nullable enable
+
+namespace Terminal.Gui {
```
tig, line 9: Same

Terminal.Gui/Configuration/ThemeManager.cs
```diff
@@ -0,0 +1,209 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text.Json.Serialization;
+
+#nullable enable
+
+namespace Terminal.Gui {
```
tig, line 9: Same

Terminal.Gui/Drawing/LineCanvasCell.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Text;
+
+
+namespace Terminal.Gui {
+
+	/// <summary>
+	/// Represents a single row/column within the <see cref="LineCanvas"/>. Includes the glyph and the foreground/background colors.
+	/// </summary>
+	public class LineCanvasCell {
```
tig, line 9: I really hate how not having nested classes makes it harder to be terse.

What do you think of leaving this `Cell`? Note that this class may end up being part of ConsoleDriver (or a variant of this) when #2616 is addressed.

tig (changes requested):

Terminal.Gui/Configuration/Scope.cs
tznind: I've changed all the new files to use file scoped namespace statements 🚀

Terminal.Gui/Drawing/LineCanvasCell.cs
tznind: Sure, I was thinking it could be confused with having a role in `TableView` but it wasn't a deep concern.  I agree `Cell` is fine as it has xmldoc clearly describing what it does.

Terminal.Gui/ApplicationRunState.cs
tznind: Done

tig (approved):

PR #2685, opened 2023-05-29
Fixes #2683 - Adds an ITableSource which wraps a TreeView<T>

### LINKED ISSUE #2683: Hierarchical table view
This is more of a suggestion request, as I don't think this control would be general enough that too many applications could benefit from it. Feel free to close it, if it's not something you want to deal with in this library directly, which is understandable.

My problem arose, when I wanted to visualize local values in our debugger. In something like VS, this is solved by a table view, where names are in the first column and values are in the second. If a value is compound (like an instance), the row becomes expandable, and subelements are shown as table rows indented. Example:
image

A table view is not enough here, as if I understood from #1285, I can't just embed a tree-view in the second column or something. A tree-view feels very noisy, and values are hard to align after the field names. It's also harder to read, because there's no visual hint about separating names and values anymore.

Could table view be extended somehow to support this? If so, what would it involve? I'd imagine it would have to be a pretty hefty change, considering it builds on `DataTable`.

tznind:
Fixes #2683  - Adds an `ITableSource` which wraps a `TreeView<T>`

@LPeter1997 the quick win is done, now the 90% effort polishing happens ;) 
tree-table-go-go

## Pull Request checklist:


LPeter1997:
Oh my, this looks perfect! Can't wait to put my hands on it

tznind:
Ok I've marked this as ready for review now.  I've added both mouse and keyboard support for expansion/collapse.  I've also put the scenario into TableEditor in UICatalog so it can be more easily tested with other TableView features (styles, checkboxes etc).

Terminal.Gui/Views/TableView/TreeTableSource.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Text;
+
+namespace Terminal.Gui;
+
+/// <summary>
+/// An <see cref="ITableSource"/> with expandable rows.
+/// </summary>
+/// <typeparam name="T"></typeparam>
+public class TreeTableSource<T> : IEnumerableTableSource<T>, IDisposable where T : class {
+	private TreeView<T> tree;
```
tig, line 14: Please prefix privates with `_` and don't use `this.`.

UnitTests/Views/TreeTableSourceTests.cs
```diff
@@ -0,0 +1,287 @@
...
+		tv.Table = checkSource = new CheckBoxTableSourceWrapperByIndex (tv, tv.Table);
+		tv.Style.GetOrCreateColumnStyle (2).MinAcceptableWidth = 1;
+
+		tv.Draw ();
+
+		string expected =
+			@"
+    │ │Name          │Description          │
+├─┼──────────────┼─────────────────────┤
+│☐│├+Lost Highway│Exciting night road  │
```
tig, line 155: Is there a way to factor these tests to test what you're trying to test without relying on text output? Specifically, if the glyph for the check changes in the future, can these tests be made such that they don't need to be adapted?

tig (changes requested):
Great stuff. See my minor comments.

Terminal.Gui/used2.txt
```diff
@@ -0,0 +1,92 @@
+nf-dev-git
```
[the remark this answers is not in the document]
tznind: Oops was not meant to be comitted.  This was a left over temp file from when I was adding the nerd fonts to file dialog.  Have removed.

Terminal.Gui/Views/TableView/TreeTableSource.cs
tznind: Fixed in [bc14970](https://github.com/gui-cs/Terminal.Gui/pull/2685/commits/bc14970f1c7580ffd6a48661d635623187aaa1af)

UnitTests/Views/TreeTableSourceTests.cs
tznind: The test is to confirm that checkboxes play nicely with the tree source.

Since they are both `ITableSource` its important they don't break one another when used together.

The most comprehensive way of doing this is to check expanded/checked statuses but also confirm with draw check.  But I agree we don't want it to be style dependent.  

I have updated the test to explicitly set (and reset) the `ConfigurationManager` runes that should be used.:

https://github.com/gui-cs/Terminal.Gui/pull/2685/commits/3eff9d4d9829620c4d824a90450b7c47a3c75eba

tig (approved):

PR #2703, opened 2023-06-07
Testing v2 publish stuff

tig:


tznind:
I noticed that the current v2 alpha packages have been un-listed (no longer appear in package manager - although can still be resolved if referenced in csproj).  Is it possible to leave at least the latest up? I think having a prereleases nuget package stream would really help people upgrade and test new features earlier?

tig:
This is just because I got busy with other things and was unable to figure out how to make it work. 

The crux iirc is getting gitversion to ignore branches and only use tags. I have an issue open against gitversion asking for help.

tznind:
Cool - no rush.  Was just unsure about the de-listing of packages incase it was something more serious (like there was a problem with the v2 packages).

tig:
No longer valid. Closing.

PR #2731, opened 2023-07-01
Reduce resize poll rate

tznind:
Fixes #2730 - Reduce screen buffer polling to 500ms from 10ms in net driver and 100ms in WindowsDriver

## Pull Request checklist:

- [ ] I've named my PR in the form of "Fixes #issue. Terse description."
- [ ] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [ ] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [ ] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [ ] My changes generate no new warnings
- [ ] I have checked my code and corrected any poor grammar or misspellings
- [ ] I conducted basic QA to assure all features are working

PR #2746, opened 2023-07-12
Fixes #2745 - `NetDriver` key handling is broken on Windows Terminal v1.18+

### LINKED ISSUE #2745: `NetDriver` key handling is broken on Windows Terminal v1.18+
To repro simply run UI Catalog and press ESC a few times.

image

I found this while trying to track down what I broke in #2612.

I kept going back in my commit history to find where the odd behavior I was seeing didn't happen. I eventually just tried v2_develop and discovered this was happening. So I tried it on `develop` too.

tig:
Fixes #2745 (see also this WT issue: https://github.com/microsoft/terminal/issues/15693)

This PR comments out the code in NetDriver that attempted to use ESC sequences to determine if the terminal has been resized. This code is naive and doesn't actually work: If `IsTerminalWithOptions` is `true` we send the "query size" ESC sequence every 10ms and WT 1.18+ dutifully responds. Thus all other keyboard input is swamped.

All of this code will be significantly reworked in v2 (as part of #2612).

@BDisp - You put this code in for a reason so I'm a little leery about simply disabling it as I've done here. Do you remember what scenario this may break (e.g. is it some oddball linux/docker scenario or some other terminal beyond Windows Terminal?). 

## Pull Request checklist:

### DIFF
--- a/UnitTests/Views/PanelViewTests.cs
+++ b/UnitTests/Views/PanelViewTests.cs
@@ -40,11 +40,11 @@ namespace Terminal.Gui.ViewTests {
 		{
 			var pv = new PanelView (new Label ("This is a test."));
 			Assert.NotNull (pv.Child);
-			Assert.Equal (1, pv.Subviews [0].Subviews.Count);
+			Assert.Single (pv.Subviews [0].Subviews);
 
 			pv.Child = null;
 			Assert.Null (pv.Child);
-			Assert.Equal (0, pv.Subviews [0].Subviews.Count);
+			Assert.Empty (pv.Subviews [0].Subviews);
 		}
 
 		[Fact]
@@ -52,11 +52,11 @@ namespace Terminal.Gui.ViewTests {
 		{
 			var pv = new PanelView ();
 			Assert.Null (pv.Child);
-			Assert.Equal (0, pv.Subviews [0].Subviews.Count);
+			Assert.Empty (pv.Subviews [0].Subviews);
 
 			pv.Add (new Label ("This is a test."));
 			Assert.NotNull (pv.Child);
-			Assert.Equal (1, pv.Subviews [0].Subviews.Count);
+			Assert.Single (pv.Subviews [0].Subviews);
 		}
 
 		[Fact]
@@ -64,12 +64,12 @@ namespace Terminal.Gui.ViewTests {
 		{
 			var pv = new PanelView (new Label ("This is a test."));
 			Assert.NotNull (pv.Child);
-			Assert.Equal (1, pv.Subviews [0].Subviews.Count);
+			Assert.Single (pv.Subviews [0].Subviews);
 			Assert.IsType<Label> (pv.Child);
 
 			pv.Add (new TextField ("This is a test."));
 			Assert.NotNull (pv.Child);
-			Assert.Equal (1, pv.Subviews [0].Subviews.Count);
+			Assert.Single (pv.Subviews [0].Subviews);
 			Assert.IsNotType<Label> (pv.Child);
 			Assert.IsType<TextField> (pv.Child);
 		}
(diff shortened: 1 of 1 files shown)


tig:
I've also pulled #2731 into this. Thanks @tznind .

tig:
> > @BDisp - You put this code in for a reason so I'm a little leery about simply disabling it as I've done here. Do you remember what scenario this may break (e.g. is it some oddball linux/docker scenario or some other terminal beyond Windows Terminal?).
> 
> Yes, I know why I put that. It's was to use on Linux because it don't recognize `Console.BufferWidth` and the `Console.BufferHeight` and the escape sequence is querying for terminal resize changes. In `Linux` this is natively fast but on a pseudo terminal on Windows, even with `Windows Terminal` it take longer to process them. But I also realized the query only return the windows size and not the buffer size. So, that can be ignored.

My test results:

- In WT v1.18 - Works fine

- In `Terminator` via WSL and this PR works fine:
rXRRHFH 1

- In `xterm` via WSL this PR works fine:
wSxQ079 1

- ConEmu (not WSL)
  - Resizing does not work (but doesn't work with `develop` either)

BDisp:
As I said, when I implemented that was to get the terminal buffer size (to use with the `EnableConsoleScrolling` but it only get the terminal size) and not the the terminal size. `Linux` read the terminfo to get the `LINES` and `COLS` and the mainloop invoke a event when their values changes and it not need the escape sequence for that, glad you removed it. On `ConEmu` I didn't test it yet.

tig:
Last call before I merge this fix. I'm pretty sure it doesn't break anything else, but I'd feel much better if others had tested it.

Terminal.Gui/ConsoleDrivers/NetDriver.cs
```diff
@@ -196,7 +197,14 @@ void GetConsoleKey ()
...
+					if (cki == null) {
+						// This is an error. We got an ESC while we're in an ESC sequence already
+						// but there's no record of the ESC sequence.
+						// Ignore this for now until we can figure out how we got in this state
+						isEscSeq = false;
+					} else {
+						DecodeEscSeq (ref newConsoleKeyInfo, ref key, cki, ref mod);
+					}
 					cki = null;
 					break;
```
BDisp, line 209: Instead this code I think is better do the follow:

```cs
					DecodeEscSeq (ref newConsoleKeyInfo, ref key, cki, ref mod);
					cki = null;
					if (!Console.KeyAvailable) {
						isEscSeq = false;
					}
					break;
```
So, it process the Esc and if there isn't `Console.KeyAvailable` we've certain that isn't a escape sequence but just a simple Esc and set the `isEscSeq` to false. `cki` is always set to null and if it's a escape sequence the prior condition will add the Esc key to `cki`.

BDisp (changes requested):
@tig please see my review, thanks.

Terminal.Gui/ConsoleDrivers/NetDriver.cs
tig: I think you're right this is slightly more robust than my fix. I'll change to use your suggestion. Thanks!

However, I don't think it's correct to say "if there isn't `Console.KeyAvailable` we've certain that isn't a escape sequence". It is PROBABLY true with Windows Terminal, but it is POSSIBLE that another terminal is sending keys slowly enough that when this line of code executes the next key simply hasn't arrived yet.

BDisp: > However, I don't think it's correct to say "if there isn't `Console.KeyAvailable` we've certain that isn't a escape sequence". It is PROBABLY true with Windows Terminal, but it is POSSIBLE that another terminal is sending keys slowly enough that when this line of code executes the next key simply hasn't arrived yet.

Yes you are right, I already faced that issue before, but in that case the code must be as the original. We really can't absolutely rely on the `Console.KeyAvailable` property but I am unaware how to deal with that issue.

BDisp: What I'm meaning here is when there is a Esc key resulting from a escape sequence, by a mouse event or F1 pressed, the `Console.KeyAvailable` must always be available, but sometimes it can return `false`, which will cause the escape sequence being corrupt.

tig: The right solution is to refactor things such that it's more like a protocol parser that buffers input and not rely on KeyAvailable. 

I have some ideas on this. I'll create a separate Issue where I explain further. 

For now, the current implementation is solid enough i think. But as we reduce our dependency on ncurses for Linux and switch WindowsDriver to use ansi sequencers we should make this more robust.

PR #2770, opened 2023-07-25
Fixes #2761. Small TextView is broken scrolling using scrollbar.

### LINKED ISSUE #2761: Small TextView is broken scrolling using scrollbar
**Describe the bug**
A notably small `TextView` cannot be scrolled correctly using a scrollbar.

This bug relates to #2688 which is corrected. I am mentioning it here because it might need a regression test after this one is corrected. 

**To Reproduce**
1. Add a quite short `TextView`, so that you don't have to insert a lot of text to reproduce the error.
2. Set it's `ReadOnly` property to true and its `Text` property to several lines, such that they exceed its height. Note that @BDisp had noted that the text view being a read only has nothing to do with the problem.
3. Attach a vertical scrollbar to it as demonstrated in the UI catalog.
4. Scroll down to the end using the keyboard arrow keys (I do it programmatically in the screenshot) and then try to scroll up again, by clicking on the scrollbar's arrows or using the mouse's middle button.
5. Notice that you cannot scroll back to the top.

**Expected behavior**
To scroll correctly.

**Screenshots**
Below I am demonstrating the inability to scroll using the scrollbar and the ability to scroll using the keyboard arrow keys with no problem. The bigger `TextView` above does not suffer from the same problem.

ScrollbarProblem

 - OS: Windows
 - Version 10

BDisp:
Fixes #2761 - The principal culprit is the `SetRelativeLayout` being called with the wrong `Frame` value. For e.g. a `TextView` inside a `FrameView` but using the `Window`'s `ContentView.Frame` instead of `FrameView.Frame`.

## Pull Request checklist:


BDisp:
@tig I'll add more unit tests to guarantee that this doesn't happens anymore.

BDisp:
I think it's all done @tig.
@chr1st0scli I added the follow because `BottonOffset` isn't be set as it was sets before subscribing the event and thus ensure get his state on driver initialization.
```cs
				Initialized += TextViewScrollBarView_Initialized;
				OtherScrollBarView.Initialized += OtherScrollBarView_Initialized;

```

tig:
@BDisp can you bring this fix forward into v2_develop too? Thanks.

PR #2818, opened 2023-08-17
V1 branching update

### LINKED ISSUE #2815: Change branch/versioning system
Right now we use Git Flow.

1. `main` always represents the most recent release of `v1`
2. `develop` is the `v1` working branch
3. `v2_develop` is the `v2` working branch
4. There is no `main` for `v2`
6. the .csproj files do not contain real version information
7. We don't effectively used "pre-release" semver

# 4 and 5 are problematic for maintaining multiple versions of T.Gui over time (v1 and v2 etc...).

# 6 makes it hard to determine what version of T.Gui is in use when doing local dev.

Here's my thoughts on addressing this. I've spent a bunch of time digging into this, but am still quite confused anď unclear on the right approach. This Issue is my attempt to write it all down to drive more clear thinking. Help appreciated.

tig:
Partially addresses #2815

### DIFF
--- a/Example/Example.csproj
+++ b/Example/Example.csproj
@@ -5,10 +5,10 @@
     <!-- Version numbers are automatically updated by gitversion when a release is released -->
     <!-- In the source tree the version will always be 1.0 for all projects. -->
     <!-- Do not modify these. -->
-    <AssemblyVersion>1.13.6.0</AssemblyVersion>
-    <FileVersion>1.13.6.0</FileVersion>
-    <Version>1.13.6-pre0002</Version>
-    <InformationalVersion>1.13.6-pre.2+Branch.develop.Sha.90208d6577b9d34bf0c4e9057b4227297bc9118d</InformationalVersion>
+    <AssemblyVersion>1.14.0.0</AssemblyVersion>
+    <FileVersion>1.14.0.0</FileVersion>
+    <Version>1.14.0-pre0109</Version>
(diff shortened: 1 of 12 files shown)


Example/Example.csproj
```diff
@@ -5,10 +5,10 @@
...
     <!-- In the source tree the version will always be 1.0 for all projects. -->
     <!-- Do not modify these. -->
-    <AssemblyVersion>1.13.6.0</AssemblyVersion>
-    <FileVersion>1.13.6.0</FileVersion>
-    <Version>1.13.6-pre0002</Version>
-    <InformationalVersion>1.13.6-pre.2+Branch.develop.Sha.90208d6577b9d34bf0c4e9057b4227297bc9118d</InformationalVersion>
+    <AssemblyVersion>1.14.0.0</AssemblyVersion>
+    <FileVersion>1.14.0.0</FileVersion>
+    <Version>1.14.0-v1-develop-branc0115</Version>
+    <InformationalVersion>1.14.0-v1-develop-branching-update.115+Branch.v1-develop-branching-update.Sha.196872f2340046682a9fc98bceba4894e866e1c5</InformationalVersion>
```
BDisp, line 11: Since this information is repeated on various projects it is possible to use some environment variable or file where the projects can read, like something `<AssemblyVersion>$(AssemblyVersion)</AssemblyVersion>`?

BDisp (commented):
Nice @tig. Please see my comment. Thanks.

Example/Example.csproj
tig: Not sure. I am trying to figure out how to keep this simple.

PR #2885, opened 2023-10-04
Fixes #2872 - Driver memory leaks

### LINKED ISSUE #2872: `WindowsDriver` does not dispose `MainLoop` Tasks - memory leak
I believe this is the cause of the slowdown seen in UI Catalog when you repeatedly start and exit multiple Scenarios...

Once a fix is identified, this should be fixed in both v1 and v2.

Related: #2870 

Repo:

1. Debug.Performance Profiler... with "Memory Usage" on.
2. Take a Snapshot
3. Start and exit `Generic` 5 times
4. Take a Snapshot

Note in first snapshot `WindowsDriver` has a single ref count:

image

Note in 2nd snapshot `WindowsDriver` has 11 refs:
image

10 of those are from `WindowsMainLoop`. 

The `Tasks` created to poll for input (`WindowsInputHandler`) and window resizing (`CheckWinChange`) are not being cancelled properly which is causing `WindowsMainLoop` to not get disposed and since it has a ref to `_consoleDriver` the `WindowsDrivder` instance never gets clean up either.

The fix is to refactor how these Tasks/threads are created and cleaned up. This will require either 

- adding an `End` method to `IMainLoopDriver` that is called from `Application.Cleanup` 
- Adding `IDisposable` to `WiindowsMainLoop` and calling `Dispose` from `WindowsDriver.End`

OR 

- Refactoring the creation and cleanup of the Tasks/threads into `WindowsDriver` such that it manages them instead of `WindowsMainLoop`.

I suspect `CursesDriver` and `NetDriver` have same issue, but I did not test that.

tig:
Fixes #2872 - Driver memory leaks

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Application.cs
+++ b/Terminal.Gui/Application.cs
@@ -464,8 +464,8 @@ namespace Terminal.Gui {
 		///   To make a <see cref="Run(Toplevel, Func{Exception, bool})"/> stop execution, call <see cref="Application.RequestStop"/>.
 		///  </para>
 		///  <para>
-		///   Calling <see cref="Run(Toplevel, Func{Exception, bool})"/> is equivalent to calling <see cref="Begin(Toplevel)"/>, followed by <see cref="RunLoop(RunState, bool)"/>,
-		///   and then calling <see cref="End(RunState)"/>.
+		///   Calling <see cref="Run(Toplevel, Func{Exception, bool})"/> is equivalent to calling <see cref="Begin(Toplevel)"/>,
+		///   followed by <see cref="RunLoop(RunState)"/>, and then calling <see cref="End(RunState)"/>.
 		///  </para>
 		///  <para>
 		///   Alternatively, to have a program control the main loop and 
@@ -481,7 +481,7 @@ namespace Terminal.Gui {
 		///   this method will exit.
 		///  </para>
 		/// </remarks>
-		/// <param name="view">The <see cref="Toplevel"/> to run modally.</param>
+		/// <param name="view">The <see cref="Toplevel"/> to run as a modal.</param>
 		/// <param name="errorHandler">RELEASE builds only: Handler for any unhandled exceptions (resumes when returns true, rethrows when null).</param>
 		public static void Run (Toplevel view, Func<Exception, bool> errorHandler = null)
 		{
--- a/Terminal.Gui/Application.cs
+++ b/Terminal.Gui/Application.cs
@@ -50,6 +50,7 @@ namespace Terminal.Gui {
 		/// <summary>
 		/// If <see langword="true"/>, forces the use of the System.Console-based (see <see cref="NetDriver"/>) driver. The default is <see langword="false"/>.
 		/// </summary>
+		[SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
 		public static bool UseSystemConsole { get; set; } = false;
 
 		// For Unit testing - ignores UseSystemConsole
--- a/Terminal.Gui/Application.cs
+++ b/Terminal.Gui/Application.cs
@@ -242,7 +242,7 @@ namespace Terminal.Gui {
 
 			// BUGBUG: OverlappedTop is not cleared here, but it should be?
 
-			MainLoop?.Stop ();
+			MainLoop?.Dispose ();
 			MainLoop = null;
 			Driver?.End ();
 			Driver = null;
--- a/Terminal.Gui/Application.cs
+++ b/Terminal.Gui/Application.cs
@@ -624,8 +624,8 @@ namespace Terminal.Gui {
 		/// Run one iteration of the <see cref="MainLoop"/>.
 		/// </summary>
 		/// <param name="state">The state returned by <see cref="Begin(Toplevel)"/>.</param>
-		/// <param name="wait">If <see langword="true"/> will execute the <see cref="MainLoop"/> waiting for events. If <see langword="true"/>
-		/// will return after a single iteration.</param>
+		/// <param name="wait">If <see langword="true"/> will execute the <see cref="MainLoop"/> waiting for events. If <see langword="false"/>
+		/// the method will return after a single iteration.</param>
 		/// <param name="firstIteration">Set to <see langword="true"/> if this is the first run loop iteration. Upon return,
 		/// it will be set to <see langword="false"/> if at least one iteration happened.</param>
 		public static void RunMainLoopIteration (ref RunState state, bool wait, ref bool firstIteration)
--- a/Terminal.Gui/Application.cs
+++ b/Terminal.Gui/Application.cs
@@ -470,14 +470,14 @@ namespace Terminal.Gui {
 		///  <para>
 		///   Alternatively, to have a program control the main loop and 
 		///   process events manually, call <see cref="Begin(Toplevel)"/> to set things up manually and then
-		///   repeatedly call <see cref="RunLoop(RunState, bool)"/> with the wait parameter set to false. By doing this
-		///   the <see cref="RunLoop(RunState, bool)"/> method will only process any pending events, timers, idle handlers and
+		///   repeatedly call <see cref="RunLoop(RunState)"/> with the wait parameter set to false. By doing this
+		///   the <see cref="RunLoop(RunState)"/> method will only process any pending events, timers, idle handlers and
 		///   then return control immediately.
 		///  </para>
 		///  <para>
-		///   RELEASE builds only: When <paramref name="errorHandler"/> is <see langword="null"/> any exeptions will be rethrown. 
+		///   RELEASE builds only: When <paramref name="errorHandler"/> is <see langword="null"/> any exceptions will be rethrown. 
 		///   Otherwise, if <paramref name="errorHandler"/> will be called. If <paramref name="errorHandler"/> 
-		///   returns <see langword="true"/> the <see cref="RunLoop(RunState, bool)"/> will resume; otherwise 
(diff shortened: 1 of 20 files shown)


tig:
@BDisp, I'm trying to figure out the `wait` param on `CheckTimers`:

```cs
		/// <summary>
		/// Checks if there are any outstanding timers that need to be processed. Called from the driver's <see cref="IMainLoopDriver"/>.
		/// </summary>
		/// <param name="wait">If <see langword="false"/> <paramref name="waitTimeout"/> will be set to 0.</param>
		/// <param name="waitTimeout">Returns the number of milliseconds remaining in the current timer.</param>
		/// <returns></returns>
		public bool CheckTimers (bool wait, out int waitTimeout)
		{

		if (wait == false) {
			throw new InvalidOperationException ("CheckTimers should not be called with wait == false");
    		}
			long now = DateTime.UtcNow.Ticks;

			if (_timeouts.Count > 0) {
				waitTimeout = (int)((_timeouts.Keys [0] - now) / TimeSpan.TicksPerMillisecond);
				if (waitTimeout < 0)
					// This avoids 'poll' waiting infinitely if 'waitTimeout < 0' until some action is detected
					// This can occur after IMainLoopDriver.Wakeup is executed where the pollTimeout is less than 0
					// and no event occurred in elapsed time when the 'poll' is start running again.
					waitTimeout = 0; 
					return true;
			} else {
				waitTimeout = -1;
			}

			if (!wait) {
				waitTimeout = 0;
			}

			int ic;
			lock (_idleHandlers) {
				ic = _idleHandlers.Count;
			}

			return ic > 0;
		}
```

I have looked over all the code in the project, including unit tests, and I can't find any cases where it will ever bet set to `false`.

I bubbles up to:

`Application.RunMainLoopIteration`:

```cs
/// <summary>
/// Run one iteration of the <see cref="MainLoop"/>.
/// </summary>
/// <param name="state">The state returned by <see cref="Begin(Toplevel)"/>.</param>
/// <param name="wait">If <see langword="true"/> will execute the <see cref="MainLoop"/> waiting for events. If <see langword="false"/>
/// the method will return after a single iteration.</param>
/// <param name="firstIteration">Set to <see langword="true"/> if this is the first run loop iteration. Upon return,
/// it will be set to <see langword="false"/> if at least one iteration happened.</param>
public static void RunMainLoopIteration (ref RunState state, bool wait, ref bool firstIteration)
{
```

image

I added the `throw` you see above and ran the unit tests. It never fires.

So, I believe this whole `wait` parameter starting at `Application.RunMainLoopIteration`, then at `IMainLoopDriver.EventsPending`, then `CheckTimers` is un-used and un-needed.

Am I missing something?

BDisp:
You are right. I also never seen it as false, but it's an option. If false the timeout will be 0 and the loop will continue infinitely.

tig:
@bdisp, also nobody ever checks for a waitTimeout == -1. 

In fact, both Windows and NetDriver have a bug w.r.t this because _keyReady.Wait will be called with a timeout of -1!

```cs

	bool IMainLoopDriver.EventsPending ()
	{
		_waitForProbe.Set ();

		if (_mainLoop.CheckTimersAndIdleHandlers (out var waitTimeout)) {
			return true;
		}

		try {
			if (!_tokenSource.IsCancellationRequested) {
				// BUGBUG: CheckTimers can return false with waitTimeOut set to -1 if there 
				// are no idle handlers or timers. The code below doesn't account for this.
				_keyReady.Wait (waitTimeout, _tokenSource.Token);
			}
		} catch (OperationCanceledException) {
			return true;
		} finally {
			_keyReady.Reset ();
		}

		if (!_tokenSource.IsCancellationRequested) {
			return _inputResult.Count > 0 || _mainLoop.CheckTimersAndIdleHandlers (out _);
		}

		_tokenSource.Dispose ();
		_tokenSource = new CancellationTokenSource ();
		return true;
	}
```

I think I can change CheckTimers to always set waitTimeout = 0 if there are no active timers or idle handlers. Right?

tig:
Here's my proposed new function on `MainLoop`:

```cs
		/// <summary>
		/// Called from <see cref="IMainLoopDriver.EventsPending"/> to check if there are any outstanding timers or idle handlers.
		/// </summary>
		/// <param name="waitTimeout">Returns the number of milliseconds remaining in the current timer. Will be -1 if there are no
		/// active timers.</param>
		/// <returns><see langword="true"/> if there is a timer or idle handler active.</returns>
		public bool CheckTimersAndIdleHandlers (out int waitTimeout)
		{
			var now = DateTime.UtcNow.Ticks;

			waitTimeout = 0;

			lock (_timeouts) {
				if (_timeouts.Count > 0) {
					waitTimeout = (int)((_timeouts.Keys [0] - now) / TimeSpan.TicksPerMillisecond);
					if (waitTimeout < 0) {
						// This avoids 'poll' waiting infinitely if 'waitTimeout < 0' until some action is detected
						// This can occur after IMainLoopDriver.Wakeup is executed where the pollTimeout is less than 0
						// and no event occurred in elapsed time when the 'poll' is start running again.
						waitTimeout = 0;
					}
					return true;
				}
			}

			// There are no timers set, check if there are any idle handlers

			lock (_idleHandlers) {
				return _idleHandlers.Count > 0;
			}
		}
```

BDisp:
> I think I can change CheckTimers to always set waitTimeout = 0 if there are no active timers or idle handlers. Right?

I don't recommend that. Return -1 is normally used to wait until some keystroke occurs. If it's necessary to interrupt, then the `keyReady.Set` will interrupt the wait for `WindowsDriver` and `NetDriver`. The `CursesDriver` write to a file descriptor to interrupt the `poll` call.

tig:
Please explain how it makes sense to pass -1 to _keyReady.Wait?

BDisp:
> Please explain how it makes sense to pass -1 to _keyReady.Wait?

That's the way how `ManualResetEvent` works when the call in the `Wait` handles internally for any event. The `poll` function also accept -1 to wait for an event. Otherwise the mainloop will always running without any event.

tig:
> `ManualResetEvent`

I see now:

image

Very confusing and a great example of where just a few comments in the code would be super helpful.

New version:

```cs
/// <summary>
/// Called from <see cref="IMainLoopDriver.EventsPending"/> to check if there are any outstanding timers or idle handlers.
/// </summary>
/// <param name="waitTimeout">Returns the number of milliseconds remaining in the current timer (if any). Will be -1 if
/// there are no active timers.</param>
/// <returns><see langword="true"/> if there is a timer or idle handler active.</returns>
public bool CheckTimersAndIdleHandlers (out int waitTimeout)
{
	var now = DateTime.UtcNow.Ticks;

	waitTimeout = 0;

	lock (_timeouts) {
		if (_timeouts.Count > 0) {
			waitTimeout = (int)((_timeouts.Keys [0] - now) / TimeSpan.TicksPerMillisecond);
			if (waitTimeout < 0) {
				// This avoids 'poll' waiting infinitely if 'waitTimeout < 0' until some action is detected
				// This can occur after IMainLoopDriver.Wakeup is executed where the pollTimeout is less than 0
				// and no event occurred in elapsed time when the 'poll' is start running again.
				waitTimeout = 0;
			}
			return true;
		}
		// ManualResetEventSlim.Wait, which is called by IMainLoopDriver.EventsPending, will wait indefinitely if
		// the timeout is -1.
		waitTimeout = -1;
	}

	// There are no timers set, check if there are any idle handlers

	lock (_idleHandlers) {
		return _idleHandlers.Count > 0;
	}
}
```

tig:
As I was fixing mem leak this am in windows driver I found the tearing on resize no longer happens if we use the input event! 

I think we can remove the CheckWinChanges task, at least for recent windows / wt versions!

tig:
> As I was fixing mem leak this am in windows driver I found the tearing on resize no longer happens if we use the input event!
> 
> I think we can remove the CheckWinChanges task, at least for recent windows / wt versions!

Meh. I couldn't get it to work; the sizes given in the event and during Init are still wrong and it's not until the first time the ChangeWin thread runs that the right size is returned. 

I did document it all better in `#define HACK_CHECK_WINCHANGED`.

This will be moot once WindowsDriver is refactored via #2610.

tig:
Ok, WindowsDriver is now fixed in this PR.

now on to NetDriver:

image

tig:
FWIW, Performance Profiler doesn't let you specify cmd line args to the Startup Project.

I've enabled a ConfigurationManager setting to force it:

image

tig:
@BDisp, 

I fixed the most egregious memory leak in Netdriver here. 

However, the interplay between the threading Tasks in `NetMainLoop` and `NetEvents` is horrifically complex and I'm not sure I got it right. I'd really appreciate a thorough code review by you.

In addition, there still is a very weird memory leak (or allocation of objects that doesn't make sense):

1. Set "Application.UseSystemConsole": true in `./UICatalog/bin/Debug/net7.0/.tui/UICatalog.config`

```json
{
    "Application.UseSystemConsole" : true
} 
```

3. Start Performance Profiler with memory profiling on
4. Take a snapshot right away
5. Double click on "Generic"
6. Hit ESC to close Generic
7. Take a snapshot

You'll see something like this:
image

Click on the +1612 and see this (sort by "Count Diff"):
image

For some reason ~1000 "Task+DelayPromiseWithCancellation" objects are being created. The number is different each run, but not in an deterministic way. I've also noticed that it has something do with mouse vs. keyboard input. But I've not been able to find a pattern. Nor have I figured out how to fix it (obviously).

Maybe you can see what's going on?

PR #2896, opened 2023-10-09
Fixes #2797 - Refactor Color/TrueColor - Make TrueColor default 

### LINKED ISSUE #2797: Refactor Color/True Color - Make TrueColor default
* The TrueColor support in `v2_develop` works only for Windows Terminal (see #2798 and #2799).
* `Attribute`, `Color`, etc... all need to be refactored to make this work generically so the other drivers can work too.
* TrueColor should be the default; fallback to 16 color mode would be automatic if possible. If not, users should be able to force it via `ConfigurationManager`. E.g. `"Application.Force16ColorMode" : true`.
* Persistence of colors to `ConfigurationManager` needs to be sorted out.

tig:
Fixes #2797 - Refactor Color/True Color - Make TrueColor default 
Partial fix for #457 -

## Plan

- [x] `Color` becomes a class that represents a 32-bit color value accessible via `R`, `G`, and `B` properties (there will be an un-used `A` property as well given that most color systems include an alpha channel, even though there's no standard for that in most terminals).
  - [x]  `Color` will have properties for each of the 16 legacy color names (in the current `Color` enum). This enables most existing app code that uses `Color.White` etc... to work.
  - [x]  `Color` will provide a way to convert between legacy-16-color colors and 24-bit colors (`Value` will return `(this.A<< 24) | (this.R<< 16) | (this.G<< 8) | tc.B`).

- [x] `Attribute` will revert to looking more like it did in v1 with just a `Foreground` and `Background` property (the `TrueColorForeground/Background` properties will be removed.
  - [x] Clean up extra methods that are no longer needed/redundant
    - [x] Simplify constructors and remove duplicate code
    - [x] Remove `Attribute.Make` - use `new` instead.
    - [x] Remove `Attribute.Get` - Was only used in one unit test
    - [x] Remove `Driver.GetColors` - not needed anymore
    - [x] Remove `HasValidColors` stuff - With TrueColor and the map of color names to RGB values now in `Color`, it's not needed.
    - [x] Remove `Initalized` stuff - With TrueColor and the map of color names to RGB values now in `Color`, it's not needed.

- [x] TrueColor should be the default; fallback to 16 color mode would be automatic if possible. 
  - [x] If not, users should be able to force it via ConfigurationManager. E.g. "Application.Force16Colors : true.
  
- [x] Persistence of colors to ConfigurationManager needs to be sorted out.
  - [x] Json schema: Supports reading of `#RGB`, `#RRGGBB`, `#RGBA`, `#RRGGBBAA`, `rgb(r,g,b)`, `rgb(r,g,b,a)`, or a `ColorName`.
     - [x] Alpha read  but never used
     - [x] Alpha is never written - this can be revisited someday
  - [x] Still need to figure out how the ColorNames to RGB values map can be configurable. 
    - [x] New configurable property: `Color.Colors` 👍 
```json
  "Colors": {
    "White": "#F2F2F2",
    "BrightYellow": "#F9F1A5",
    "BrightCyan": "#61D6D6",
    "Cyan": "#3A96DD",
    "BrightMagenta": "#B4009E",
    "Magenta": "#881798",
    "BrightGreen": "#16C60C",
    "DarkGray": "#767676",
    "Green": "#13A10E",
    "BrightBlue": "#3B78FF",
    "Yellow": "#804020",
    "Blue": "#0037DA",
    "Black": "#0C0C0C",
    "BrightRed": "#E74856",
    "Gray": "#CCCCCC",
    "Red": "#C50F1F"
  },
```
    - [ ] Somehow read these from the environment at startup?

## Pull Request checklist:


tig:
I think this is ready!

tig:
Working on ColorPicker... <3

pdXm8rk 1

Terminal.Gui/Drawing/Color.cs
```diff
@@ -9,151 +10,388 @@
...
+			var deltaR = (double)color1.R - (double)color2.R;
+			var deltaG = (double)color1.G - (double)color2.G;
+			var deltaB = (double)color1.B - (double)color2.B;
+
+			return Math.Sqrt (deltaR * deltaR + deltaG * deltaG + deltaB * deltaB);
+		}
+
+		/// <summary>
+		/// Gets or sets the <see cref="Color"/> using a legacy 16-color <see cref="ColorNames"/> value.
+		/// </summary>
```
tznind, line 284: Add

> \<see langword="get"/> will return the closest 16 color match to the true color when no exact value is found.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -365,21 +365,19 @@ public void ClearContents ()
...
 	/// Will be forced to <see langword="true"/> if <see cref="ConsoleDriver.SupportsTrueColor"/> is  <see langword="false"/>, indicating
 	/// that the <see cref="ConsoleDriver"/> cannot support TrueColor.
+	/// </para>
 	/// </remarks>
-	public virtual bool Force16Colors {
-		get => _force16Colors || !SupportsTrueColor;
-		set {
-			_force16Colors = (value || !SupportsTrueColor);
-		}
+	internal virtual bool Force16Colors {
```
tznind, line 378: Why is this no longer public? perhaps API users would like to pass `--oldcolors` to their app or some similar use case? The ability to set directly seems nice, unless that conflicts with the configuration manager somehow.  

Generally having more ways to do the same thing is better.

Terminal.Gui/ConsoleDrivers/FakeDriver/FakeDriver.cs
```diff
@@ -81,8 +81,8 @@ public override void Init (Action terminalResized)
 		FakeConsole.Clear ();
 		ResizeScreen ();
 		// Call InitializeColorSchemes before UpdateOffScreen as it references Colors
-		CurrentAttribute = MakeColor (Color.White, Color.Black);
-		InitializeColorSchemes ();
+		CurrentAttribute = new Attribute (Color.White, Color.Black);
+		//InitializeColorSchemes ();
 		ClearContents ();
```
tznind, line 86: Theres a lot of commented out code in FakeDriver, maybe we should just delete it if it is no longer relevant?

Terminal.Gui/Views/TextField.cs
```diff
@@ -40,7 +40,7 @@ public class TextField : View {
 		/// Gets or sets the foreground <see cref="Color"/> to use when 
 		/// rendering <see cref="Caption"/>.
 		/// </summary>
-		public Color CaptionColor { get; set; } = Color.DarkGray;
+		public Color CaptionColor { get; set; } = (Color)Color.DarkGray;
```
tznind, line 43: I love that it is possible to cast to `Color` from the enum but for readability it might be better in our code to use the constructor directly (even though they are functionally the same).

I think the code is more readable when you have:

```csharp
public Color CaptionColor { get; set; } = new Color(ColorNames.DarkGray);
```

tznind (approved):

tznind:
Awesome work! I have left a few comments, nothing major.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
tig: Because `Application.Force16Colors` IS public (and configurable).

One other thing I'm trying to do is get rid of apps ever referencing `Driver.`.

tznind: Cool, hadn't seen that it was an option on `Application` also.  🚀

Terminal.Gui/Views/TextField.cs
tig: I think it's 6 one way, a half-dozen the other. But I love consistency, so I'm going to follow your suggestion and make all our code use `new`.

PR #2914, opened 2023-10-16
Fixes #2913 - remove Disposed views from Responder.instances when Debugging 

### LINKED ISSUE #2913: Pos objects preventing Garbage Collection of views
As the Title suggests, I have a problem with the fact that Views hold on to their pos objects even after calling `IDisposable.Dispose`.
Because of the nature of `Pos` objects, this prevents the view from being collected by GC. This would normally not be much of an issue, but, as i am dealing with Assembly Loading and unloading, it is a very important requirement for an object to be able to be GC'd immediatly after calling Dispose, as else the Assembly containing the object will not be able to be unloaded when it is expected to be.
However, as there probably aren't too many other people that have to deal with this, i have another reason why this should be resolved: Performance. When creating a Tui Application of large size, you sometimes have to discard large objects that you don't need at the moment, but can easily be rebuilt. This logic however doesn't apply if the object you dereferenced from your program is still fixed in ram by some static variable. 

I think the solution to this is actually relatively simple: Alter the `Dispose` implementation in `base.View.cs` so that it dereferences (Or maybe even destructs) the `Pos` and `Dim` objects stored in the `x`, `y`, `width`, `height`, and any other variable that may or may not store a `Pos` or `Dim` object. Simple as that (At least i think so, so please correct me if I'm wrong).

Creating an override of the `View` class and forcing coworkers to use it instead of the base. But the id have to re-Implement the child classes of `View` (the base). 

**Additional context**
I might create a Pull Request to solve this myself, but only once i have confirmation that this actually would work

a-usr:
Fixes #2913  - remove Disposed views from Responder.instances and remove Pos and Dim objects from View when disposing.

## Pull Request checklist:

Please take a look at the code before accepting / rejecting. Just please

Made to assist [this solution](https://github.com/gui-cs/Terminal.Gui/pull/2915)


tig (changes requested):
Unit tests fail tho. Can you fix?

a-usr:
> Unit tests fail tho. Can you fix?

I'll try, but the UnitTest that is failing seems to have no connection to the changes I made. I'll investigate

a-usr:
I haveCreated a new Issue for this: https://github.com/gui-cs/Terminal.Gui/issues/2923

Terminal.Gui/Core/Responder.cs
```diff
@@ -277,6 +277,9 @@ protected virtual void Dispose (bool disposing)
 
 				// TODO: free unmanaged resources (unmanaged objects) and override finalizer
 				// TODO: set large fields to null
+#if DEBUG_IDISPOSABLE
+				Instances.Remove(this);
+#endif
```
BDisp, line 282: This is on the wrong place. Also this doesn't guarantee that all instances were really disposed (`WasDisposed is true`).
I'll send you a PR to your branch which will include the fix for this and for the unit tests failure.

BDisp: Here is the fix https://github.com/a-usr/Terminal.Gui-FORK/pull/1.

BDisp:
> On a fresh clone of the develop branch the Tests `UnitTests/Terminal.Gui.Drivertests/Contents_Copies_From_OS_Clipboard`, `UnitTests/Terminal.Gui.Drivertests/Contents_Pastes_To_OS_Clipboard` and `UnitTests/Terminal.Gui.ViewTests/KeyBindings_Command` fail, so it seems like the issue isnt caused by the changes i made on my branch. But I will still try to help, although i dont have that much time to work on my "own" projects.

None of this tests fail. Perhaps it's something missing on your development configuration.

a-usr:
> > On a fresh clone of the develop branch the Tests `UnitTests/Terminal.Gui.Drivertests/Contents_Copies_From_OS_Clipboard`, `UnitTests/Terminal.Gui.Drivertests/Contents_Pastes_To_OS_Clipboard` and `UnitTests/Terminal.Gui.ViewTests/KeyBindings_Command` fail, so it seems like the issue isnt caused by the changes i made on my branch. But I will still try to help, although i dont have that much time to work on my "own" projects.
> 
> None of this tests fail. Perhaps it's something missing on your development configuration.

The clipboard tests failed because I didn't have the ms store version of Powershell installed.

The KeyBindings_Command Test fails because the actuall string is separated differently than the expected string (`/` / `.`).

I don't really have the necessary In-depth knowledge of the the inner workings of Terminal.Gui, but if I had to make an educated guess I'd say the separator is not hard-coded considering on my machine the test fails, but on yours it doesn't.

Maybe It has to do something with the OS? 

Again, I don't have had a good enough look at the source code (mainly because it is so much) so I can't really say for sure

BDisp:
> The clipboard tests failed because I didn't have the ms store version of Powershell installed.
> 
Install it from [here](https://github.com/PowerShell/PowerShell/releases/download/v7.3.8/PowerShell-7.3.8-win-x64.msi).

> The KeyBindings_Command Test fails because the actuall string is separated differently than the expected string (`/` / `.`).
> 
How you are running the tests? From command line, VS2022,  VSCode, other? In what OS?

> I don't really have the necessary In-depth knowledge of the the inner workings of Terminal.Gui, but if I had to make an educated guess I'd say the separator is not hard-coded considering on my machine the test fails, but on yours it doesn't.
> 
The tests are running with "en-US" encoder, no mater OS localization you have.

> Maybe It has to do something with the OS?
> 
I don't think so. Mine is Portuguese, which separator and others formats are different from "En-US", but the tests run well.

> Again, I don't have had a good enough look at the source code (mainly because it is so much) so I can't really say for sure

But in this case there is no problem with the code and the unit tests must run without fail. PAY ATENTION you can't copy and paste anything on the computer while the unit tests are running because when the clipboard tests are writing to and reading from clipboard the values may differ.

Terminal.Gui/Core/Responder.cs
a-usr: > This is on the wrong place. Also this doesn't guarantee that all instances were really disposed (`WasDisposed is true`). I'll send you a PR to your branch which will include the fix for this and for the unit tests failure.

however it does remove the current instance from the instances static, doesn't it? 
Compared to checking the entire static every time one element is being disposed, this would also have better performance?

Anyways, i have merged your Pull Request.

a-usr:
> > The KeyBindings_Command Test fails because the actuall string is separated differently than the expected string (`/` / `.`).
> 
> How you are running the tests? From command line, VS2022, VSCode, other? In what OS?

From VS2022 in Windows
> > I don't really have the necessary In-depth knowledge of the the inner workings of Terminal.Gui, but if I had to make an educated guess I'd say the separator is not hard-coded considering on my machine the test fails, but on yours it doesn't.
> 
> The tests are running with "en-US" encoder, no mater OS localization you have.

Are you sure? I modified the build workflow to run on my branch, and all tests are passing now. 
However on my machine they do not. I suppose xUnit is supposed to force the localization? 
My machine has German Localization, wich is why the date is seperated with `.` on my end.

Im closing #2923  for now, as this seems to be an issue with xUnit not being able to force the localization on some machines, and because the tests run perfectly fine now with your fix,

a-usr:
Also, @tig, you could try to merge now, unittests complete without errors now thanks to @BDisp's fix.

BDisp:
> Are you sure? I modified the build workflow to run on my branch, and all tests are passing now. However on my machine they do not. I suppose xUnit is supposed to force the localization? My machine has German Localization, wich is why the date is seperated with `.` on my end.
> 
> Im closing #2923 for now, as this seems to be an issue with xUnit not being able to force the localization on some machines, and because the tests run perfectly fine now with your fix,

Why do you not try to run the modified workflow on your machine and if it all pass, commit and push to this PR.

Terminal.Gui/Core/Responder.cs
BDisp: @a-usr, sorry I forgot to do that. I already sent to your branch a fix for this on https://github.com/a-usr/Terminal.Gui-FORK/pull/4.
This really increase significantly the performance. Thanks for pointing this.

a-usr:
> > Are you sure? I modified the build workflow to run on my branch, and all tests are passing now. However on my machine they do not. I suppose xUnit is supposed to force the localization? My machine has German Localization, wich is why the date is seperated with `.` on my end.
> > Im closing #2923 for now, as this seems to be an issue with xUnit not being able to force the localization on some machines, and because the tests run perfectly fine now with your fix,
> 
> Why do you not try to run the modified workflow on your machine and if it all pass, commit and push to this PR.

as I already said, the one test that actually is failing for me (`UnitTests/Terminal.Gui.ViewTests/DateFieldTests/KeyBindings_Command`) is likely failing because xUnit for some reason can't force the localization properly. This is either a problem with xUnit itself, or a problem with the custom windows environment on my machine, set by my employer.
The Github Action however is working and doesn't produce any errors. Thats why I already pushed it to this PR.

a-usr:
> why do you not try to run the modified workflow on your machine and if it all pass, commit and push to this PR.

Did you maybe think that i modified the workflow to make all tests work? Because i did not. I only enabled a push to my branch to trigger the execution of this workflow, nothing else. I know a merge to the develop branch on my repo would've done the exact same thing, but I am afraid of breaking code that is not mine, especially if i dont understand it completely.

BDisp:
> Also, @tig, you could try to merge now, unittests complete without errors now thanks to @BDisp's fix.

@a-usr don't forget to merge my PR https://github.com/a-usr/Terminal.Gui-FORK/pull/4 first. Thanks.

BDisp:
> > Also, @tig, you could try to merge now, unittests complete without errors now thanks to @BDisp's fix.
> 
> @a-usr don't forget to merge my PR [a-usr#4](https://github.com/a-usr/Terminal.Gui-FORK/pull/4) first. Thanks.

I also submit the PR #2929 for the v2.

a-usr:
> @a-usr don't forget to merge my PR [a-usr#4](https://github.com/a-usr/Terminal.Gui-FORK/pull/4) first. Thanks.

I actually didnt get a notification for that haha. I merged it now, thanks for your effort!

BDisp (changes requested):
I already review this but only @tig can activate the CI and merge. We need to wait.

a-usr:
I didnt know that. Tbh this is my first time doing collaborative work over GitHub

tig:
@BDisp I'm worried about taking the removal of the `static` stuff from `Pos/Dim` into v1. 

Strictly speaking that change is:

- not needed to fix this particular issue
- a breaking change

I think there's a significant risk of regressions to v1 if we take https://github.com/a-usr/Terminal.Gui-FORK/pull/4 as part of the is PR. 

Do you really think we should? I'll defer to you, but it makes me nervous.

BDisp:
> Do you really think we should? I'll defer to you, but it makes me nervous.

@tig the fix really work perfectly without any unit tests failures on both versions and there isn't any breaking change. But a static field shouldn't store object instances that already was disposed. Don't you agree?

tig:
> > Do you really think we should? I'll defer to you, but it makes me nervous.
> 
> @tig the fix really work perfectly without any unit tests failures on both versions and there isn't any breaking change. But a static field shouldn't store object instances that already was disposed. Don't you agree?

I see now. I had it in my mind this removed `static` from public members. But I obv. misunderstood. Thanks!

tig (approved):
Nice!

Issue #2502, opened 2023-04-04
Discuss - There should only be one (`Application.Top`)


tig:
This is all debatable, but here's my thinking:

`TopLevel` is currently built around several concepts that are muddled and have led to overly complex code:

* A `Toplevel` is a view that can host a Menu and StatusBar. Other views cannot.
* A `Toplevel` can run modally, and have its own `Runstate` (via `Application.Run()`. It is unclear why ANY VIEW can't be run this way, but it seems to be a limitation of the current implementation.
* A `Toplevel` can be moved by the user (actually only the `Toplevel` subclass, `Window`). It should be possible to enable the moving of any View (e.g. `View.CanMove = true` or similar).

`Application` is the place where all of the above `TopLevel` gunk happens in addition to other "application" stuff. 

The `MdiContainer stuff` is complex, perhaps overly so, and is not actually used by anyone outside of the project. It's also misnamed because Terminal.Gui doesn't actually support "documents" nor does it have a full "MDI" system like Windows (did). It seems to represent features useful in overlapping Views, but it is super confusing on how this works, and the naming doesn't help.  `Application` is full of code like `if (MdiTop != null && top != MdiTop && top != Current && Current?.Running == false && !top.Running)` which just screams poor-OO design and fragility. This all can be refactored to support specific scenarios and thus be simplified.

# My proposal:

- `Application.Top` should be set when an app starts, and should not be changeable. Changing it after `Application.Begin` has been called should result in an exception.

- The `Application` class should do the following (and nothing else; other things should be moved elsewhere as appropriate):
   - Application-wide settings (like default quit key and diagnostics)
   - Initializing and configuring the console driver
   - Providing `Runstate` and managing the `Mainloop`
   - Providing a simple API for starting applications either with a default `Top` view or a developer-provided one
   - Proxying mouse/keyboard events from the driver to the `Top` 
   
- Stuff that will be moved out of (or deleted) from `Application`:
   - All the `Mdi` stuff
   - All the tracking of `Current` and `toplevels`. This stuff will just be part of `View`.
   - Code for moving from view to view in the focus or z-order heirarchies

- All `Views` should support overlapping within a superview and have a menubar and statusbar. 
- All `Views` should support having subviews be overlapping
[...]

PR #2920, opened 2023-10-17
Fixes #2578 - Updates mouse events to be relative to `View.Bounds`

### LINKED ISSUE #2578: How should `MouseEvent` work with `Margin`, `Border`, and `Padding`?
Right now in v2_develop:  The `me` parameter contains `Frame`-relative coordinates. As a result, any View that wants to override this must adjust for the frames thickness. 

E.g.  From an updated ColorPicker:
 
```cs 
                public override bool MouseEvent (MouseEvent me)
		{
			if (!me.Flags.HasFlag (MouseFlags.Button1Clicked) || !CanFocus) {
				return false;
			}

			SetFocus ();
			Cursor = new Point ((me.X - GetFramesThickness ().Left) / _boxWidth, (me.Y - GetFramesThickness ().Top) / _boxHeight);

			return true;
		}
```

I don't think most View developers should need to deal with this; they should get mouse coordinates relative to the `Bounds`.

I think `Application.ProcessMouseEvent` should be calling `view.GetFramesThickness` and using that to cause `me.X` and `me.Y` to be `Bounds` relative.

If a View implementation wants to get mouse events for its `Margin`, `Border`, or `Padding` it can do so by subscribing to the appropriate mouse events.

What do you all think?

tig:
Fixes #2578 - How should MouseEvent work with Margin, Border, and Padding? 

- [x] Update docs
- [x] Add more unit tests to enforce current behavior
- [x] Refactor `View.ScreenToView/ViewToScreen` to be more sensical in the face of `Bounds != Frame`.
- [x] Update logic that fires mouse events
- [ ] Fix View dragging code. Right now, dragging is handled w/in `TopLevel`. This needs to change to be handled within `Border`. See #https://github.com/gui-cs/Terminal.Gui/issues/2502. Will define `View.Movable` and `View.Sizable` that will control whether a view can be dragged and/or resized.
- [x] Update unit tests to not fail 
- [x] Track down and fix all library and Scenario code that depends on old behavior

Related:
- https://github.com/gui-cs/Terminal.Gui/pull/2927
- https://github.com/gui-cs/Terminal.Gui/issues/2994

## Pull Request checklist:


tig:
@BDisp & @tznind - still a draft (and currently broken), but I'd love your eyes on this asap.

Thanks!

Terminal.Gui/Application.cs
```diff
@@ -1183,38 +1190,75 @@ public static void OnMouseEvent (MouseEventEventArgs a)
...
+						return true;
+					}
 				}
+				return false;
+			}
 
-				if (!view.WantMousePositionReports && a.MouseEvent.Flags == MouseFlags.ReportMousePosition)
-					return;
+			// Work inside-out (Padding, Border, Margin)
+			// TODO: Debate whether inside-out or outside-in is the right strategy
```
tznind, line 1218: Inside out would make the most sense I think.  That would match the way the rest of the mouse events work with Views i.e. the 'most focused' View gets first bite then key goes up the nesting chain.

So:

- View handle mouse event
- Padding handle mouse event
- Frame handle mouse event
- Margin handle mouse event

Terminal.Gui/Input/Responder.cs
```diff
@@ -193,7 +195,8 @@ public virtual bool OnMouseEnter (MouseEvent mouseEvent)
 		}
 
 		/// <summary>
-		/// Method invoked when a mouse event is generated for the last time.
+		/// Called when the mouse has moved outside of the view; the view will no longer receive mouse events (until
+		/// the mouse moves within the vie again and <see cref="OnMouseEnter(Gui.MouseEvent)"/> is called).
```
tznind, line 199: spelling 'vie'

tznind:
This looks great.  If I understand correctly this PR establishes more clearly the idea of bounds (like view client area) and 'frame' i.e. border, padding margin.

This is good because it simplifies things and is mostly backwards compatible so shouldn't complicate upgrades from v1 too much.

🎆 🚀

BDisp:
This is really fantastic. Much more clear, more readable. Thanks for effort.

Terminal.Gui/Application.cs
tig: In my design, view ONLY gets mouse events that occur within the View. So even if Margin/Border/Padding handle the mouse event (or not) the view will never see it (because, by definition, the coords are outside of Bounds.

So the order really doesn't matter.

tig:
I think this is done. I could add more unit tests...

Note the big hack for Toplevel dragging. I decided to not tackle this in this pr but to just hack it until we refactor drag/size per #

PR #2922, opened 2023-10-18
Fixes #2921 - MainLoop refactoring

### LINKED ISSUE #2921: `WindowsDriver` and `CursesDriver`'s MainLoops use an `Action` incorrectly
The right fix is to change both Windows and Curses' MainLoop implementations to do this:

image

IOW, get rid of the `ProcessInput` Action alltogether and just have them call `ProcessInput` on the driver directly.

I tested it and it appears to work fine. Am I missing something (I probably am).

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/pull/2911#discussion_r1363668470_

tig:
Fixes #2921 

IMainLoopDriver and ConsoleDriver were originally designed to be decoupled by Miguel to enable pluggabiliy he imagined. We've learned that all driver implementations end up tightly coupling them.

As a result there is a bunch of complexity (and potential for bugs like #2921) that could be removed by simplifying the design.

This PR will attempt to do this. 

- [x] First step is already done: Fix WindowsDriver and CursesDriver to not use an Action for ProcessInput.
- [x] Change all `Action` based events on ConsoleDriver/Application to use the proper `EventHandler` pattern
- [x] Make `MainLoop`, `IMainLoopDriver` etc... internal implementation details, exposing curical methods/events on `Application` instead. Note this effectively means 3rd parties cannot build `ConsoleDriver` implementations; that's ok because nobody's ever done that anyway.  
- [x] Further simplify the console driver implementations by removing `PrepareToRun` and getting rid of  more duplicated code.

## Pull Request checklist:


BDisp (changes requested):
@tig please see my comments.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -861,7 +861,7 @@ public override void PrepareToRun (MainLoop mainLoop, Action<KeyEvent> keyHandle
 		_mouseHandler = mouseHandler;
 
 		_mainLoop = mainLoop.MainLoopDriver as WindowsMainLoop;
-		_mainLoop.ProcessInput = ProcessInput;
+		_mainLoop.ProcessInput += ProcessInput;
```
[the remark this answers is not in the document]
BDisp: Why not just insert a check to inspect `_mainLoop`?
```cs
if (_mainLoop != null) {
    // The handlers are already set
    return;
}
```

Terminal.Gui/ConsoleDrivers/CursesDriver/UnixMainLoop.cs
```diff
@@ -119,7 +119,7 @@ void IMainLoopDriver.Setup (MainLoop mainLoop)
 		/// <remarks>
 		///	The token parameter is the value returned from AddWatch
 		/// </remarks>
-		public void RemoveWatch (object token)
+		internal void RemoveWatch (object token)
```
BDisp, line 122: You are worry to much on change public to internal on a internal class :-)

BDisp (changes requested):
@tig please see my comments. Thanks.

Terminal.Gui/Application.cs
```diff
@@ -600,6 +617,15 @@ public override void Send (SendOrPostCallback d, object state)
...
 		}
 
+		/// <summary>
+		///   Determines whether there are pending events to be processed.
+		/// </summary>
+		/// <remarks>
+		///   Use this method to probe if events are pending. Typically used to flush the input queue while still
+		///   running code on the main thread.
+		/// </remarks>
+		public static bool EventsPending () => MainLoop.EventsPending ();
```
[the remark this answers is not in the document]
tig: This is not really a new method. All I did here was expose IMainLoop.EventsPending which was previously public. I now see it's not used anywhere, so I've removed it.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -32,15 +32,35 @@ public abstract class ConsoleDriver {
...
+	/// Returns an instance of <see cref="MainLoop"/> using the <see cref="IMainLoopDriver"/> for the driver.
+	/// </summary>
+	/// <returns></returns>
+	internal abstract MainLoop CreateMainLoop ();
+
+	/// <summary>
+	/// Initializes the driver
+	/// </summary>
+	/// <param name="terminalResized">Method to invoke when the terminal is resized.</param>
+	internal abstract void Init (Action terminalResized);
```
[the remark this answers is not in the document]
tig: In the history of gui.cs, nobody but the maintainers (us) has created a ConsoleDriver or IMainLoop implementation. 

I am asserting that nobody ever will in the future either. 

If I'm wrong and someday someone says "I have a new ConsoleDriver I'd like to write" then we can help them do that as part of the library. This will be far simpler than exposing a bunch of fragile public APIs that may or may not work anyway.

Note that in a few minutes I'll be pushing an update to this PR that goes even further. `Init` has NO parameters and `PrepareToRun` will be really close to being able to be deleted as well... ;-)

Terminal.Gui/ConsoleDrivers/CursesDriver/UnixMainLoop.cs
tig: See my comment above. I want it to be explicit in v2 that external parties to the library can NOT create ConsoleDrivers or IMainLoop implementations.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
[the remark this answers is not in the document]
BDisp: > In the history of gui.cs, nobody but the maintainers (us) has created a ConsoleDriver or IMainLoop implementation.
> 
Here is one https://github.com/Blazor-Console/HACC.

> I am asserting that nobody ever will in the future either.
> 
We never know :-)

> If I'm wrong and someday someone says "I have a new ConsoleDriver I'd like to write" then we can help them do that as part of the library. This will be far simpler than exposing a bunch of fragile public APIs that may or may not work anyway.
> 
Ok I see it only must done inside the API.

> Note that in a few minutes I'll be pushing an update to this PR that goes even further. `Init` has NO parameters and `PrepareToRun` will be really close to being able to be deleted as well... ;-)

Wow that rocks :-)

Terminal.Gui/ConsoleDrivers/CursesDriver/UnixMainLoop.cs
BDisp: Then why you don't make the `ConsoleDriver` class internal too?

Terminal.Gui/ConsoleDrivers/CursesDriver/UnixMainLoop.cs
```diff
@@ -187,7 +189,8 @@ void IMainLoopDriver.Iteration ()
 		{
 			if (_winChanged) {
 				_winChanged = false;
-				WinChanged?.Invoke ();
+				_cursesDriver.ProcessInput ();
+				//WinChanged?.Invoke ();
 			}
```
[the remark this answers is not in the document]
tig: See new code. Let me know what you think.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -780,15 +780,11 @@ public COLORREF (uint value)
...
 	WindowsConsole.SmallRect _damageRegion;
-	Action<KeyEvent> _keyHandler;
-	Action<KeyEvent> _keyDownHandler;
-	Action<KeyEvent> _keyUpHandler;
-	Action<MouseEvent> _mouseHandler;
 
 	public WindowsConsole WinConsole { get; private set; }
 
 	public override bool SupportsTrueColor => RunningUnitTests || (Environment.OSVersion.Version.Build >= 14931
-		&& (_isWindowsTerminal || _parentProcessName == "devenv"));
```
BDisp, line 791: "devenv" is returned when is running from the VS2022. "wininit" isn't ensured to be correct because you can have Windows Terminal open with other thing and not being running `Terminal.Gui`.

Terminal.Gui/Application.cs
```diff
@@ -1080,49 +1117,62 @@ static void OnUnGrabbedMouse (View view)
...
+		/// </para>
+		/// </remarks>
+		public static event EventHandler<MouseEventEventArgs> MouseEvent;
 
-		static void ProcessMouseEvent (MouseEvent me)
+		/// <summary>
+		/// Called when a mouse event occurs. Fires the <see cref="MouseEvent"/> event.
+		/// </summary>
+		/// <param name="a"></param>
+		public static void OnMouseEvent (MouseEventEventArgs a)
```
BDisp, line 1140: I really don't understand why you are exposing as public. Why a user will call this and also the keys events? You are changing public objects to internal and now you are changing private objects to public. To call this method a user need to provide a `MouseEvent` and will send a key that wasn't returned from the terminal itself. This method is to propagate an event from the terminal to the focused or grabbed view.

tig: See the unit tests and how I updated them to use this. This is a far better API for being able to "simulate a mouse event". It could be argued it should be `internal`.

BDisp: Ak ok is for simulate then is all right, sorry.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
tig: In my case, "wininit" started being returned when running in the debugger in VS2022. I don't know why, but I never see "devinit" anymore.

Terminal.Gui/Application.cs
tig: I just pushed a change that updated the API docs to reflect this.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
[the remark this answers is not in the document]
tig: > > In the history of gui.cs, nobody but the maintainers (us) has created a ConsoleDriver or IMainLoop implementation.
> 
> Here is one https://github.com/Blazor-Console/HACC.
> 
> > I am asserting that nobody ever will in the future either.
> 
> We never know :-)
> 
> > If I'm wrong and someday someone says "I have a new ConsoleDriver I'd like to write" then we can help them do that as part of the library. This will be far simpler than exposing a bunch of fragile public APIs that may or may not work anyway.

Ok, I was wrong about there not being any examples. Thanks for pointing this out.

We have several choices:

1) Make it so arbitrary ConsoleDrivers could be developed outside of the library (revert to v1 patterns)
2) Extend the v2 pattern in this PR to enable JUST the what is needed for HACC to work.
3) Fold the requirements of HACC into the libary, creating a new built-in `WebConsoleDriver`

I'm not a fan of #1, but #2 or #3 seem very doable.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: @tig I already know why I added the "devenv" into the code. If you run UICatalog directly from cmd or conhost, true color is never support because they don't process Ansi escape sequences. But if you run directly from VS2022, no mater Windows Terminal or Console Host is the default terminal, VS2022 VSConsole can always process Ansi escape sequences and thus supporting true color. So, what is needed to differ here is using `_isWindowsTerminal` to allow or not set the alternate screen and using the `_parentProcessName` to allow or not true color.
The "wininit" seems to be unrelated because it's a OS file (C:\Windows\System32\wininit.exe).

Please change to this:
```cs
	public override bool SupportsTrueColor => RunningUnitTests || (Environment.OSVersion.Version.Build >= 14931
		&& (_isWindowsTerminal || _parentProcessName == "wininit" || _parentProcessName == "devenv"));

```

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -806,7 +809,7 @@ public WindowsDriver ()
 		if (!RunningUnitTests) {
 			_parentProcessName = GetParentProcessName ();
 			_isWindowsTerminal = _parentProcessName == "WindowsTerminal";
-			if (!_isWindowsTerminal && _parentProcessName != "devenv") {
+			if (!_isWindowsTerminal && _parentProcessName != "wininit") {
```
BDisp, line 812: Maybe you can change to this:
```cs
			if (!_isWindowsTerminal && _parentProcessName != "wininit" && _parentProcessName != "devenv") {
				Force16Colors = true;
			}

```

BDisp: In line 1758 you can change to:
```cs
		if (!RunningUnitTests && _isWindowsTerminal) {
			// Disable alternative screen buffer.
			Console.Out.Write (EscSeqUtils.CSI_RestoreCursorAndActivateAltBufferWithBackscroll);
		}

```

tig: On the machine that was (for some reason!) returning "wininit" I'm now getting "devenv" again:

image

tig: On my main dev machine, I get "WindowsTerminal":

image
image
image
image

tig: It's pretty clear to me that relying on the process name is not going to work reliably.

tig: Related: https://github.com/microsoft/terminal/issues/13006

BDisp: > On the machine that was (for some reason!) returning "wininit" I'm now getting "devenv" again:
> 
> image image image image

"devenv" is what also I get when debugging from VS2022.


> On my main dev machine, I get "WindowsTerminal":
> 
> image image image image

"WindowsTerminal" is what I get when running from `Windows Terminal` and attaching it to the VS2022 debugger.


image
image
image

I'm not in the insiders program nor using preview versions, by option.

tig: > "WindowsTerminal" is what I get when running from Windows Terminal and attaching it to the VS2022 debugger.

I'm not attaching to a process. I'm hitting F5 in VS2022. 

Can we please agree that the right solution is to check for WS_SESSION and that looking at the parent process name is a hack that will not work long term? 

It is clear that WT_SESSION is also not reliable, but at least it's the method sanctioned by the WT team (and there's an open issue to fix it: https://github.com/microsoft/terminal/issues/13006)

In the meantime, instead of the hack to find the process name, a workaround for running in the VS2022 debugger is to change the launch profile thusly:

image

I still don't understand why on one of my machines WT_SESSION is set correctly by default, but on another it is not.

tig: Also, I've found that `GetProcessName()` infinite loops:

- Start.Run "pwsh"
- `.\UICatalog\bin\Debug\net7.0\UICatalog.exe`
- Hangs
- Attach to debugger from VS 2022 and hit pause

It's stuck in the `while (queryObj != null)` loop.

BDisp: @tig the better way to check what terminal default is set is read Windows registry, see https://support.microsoft.com/en-us/windows/command-prompt-and-windows-powershell-for-windows-11-6453ce98-da91-476f-8651-5c14d5777c20.

In this case we don't need to get the process name and the method can be removed.

BDisp: Ok I saw your post, in this case we need to use the method to get the process name. I haven't any hang on running the method.

tig: > Ok I saw your post, in this case we need to use the method to get the process name. I haven't any hang on running the method.

Let's make sure you and I are talking about solving the same problem. Here's the problem I think we're trying to solve:

"When I run a Terminal.Gui app in the debugger in VS 2022 I want the WindowsDriver to support True Color. However, some versions of VS don't set WT_SESSION which is the correct way for Terminal.Gui to detect WT. We need a work around."

If this is indeed the problem, the correct fix is the code that is currently in this PR:

```cs
public WindowsDriver ()
{
	if (Environment.OSVersion.Platform == PlatformID.Win32NT) {
		WinConsole = new WindowsConsole ();
		// otherwise we're probably running in unit tests
		Clipboard = new WindowsClipboard ();
	} else {
		Clipboard = new FakeDriver.FakeClipboard ();
	}

	if (!RunningUnitTests) {
		_isWindowsTerminal = Environment.GetEnvironmentVariable ("WT_SESSION") != null;
		if (!_isWindowsTerminal) {
			//_parentProcessName = GetParentProcessName ();
			//_isWindowsTerminal = _parentProcessName == "WindowsTerminal";
			//if (!_isWindowsTerminal && _parentProcessName != "devenv") {
			Force16Colors = true;
			//}
		}
	}
}
```

PLUS, advice to change `launchSettings.json` to set the WT_SESSION env var:

```json
{
  "profiles": {
    "UICatalog": {
      "commandName": "Project",
      "environmentVariables": {
        "WT_SESSION": "yes"
      }
    },
...
```

This solves the problem.

Is this the problem you are talking about?

BDisp: > Is this the problem you are talking about?

Not really, your workaround is suffice when you are debugging from VS2022 by pressing F5 and that work well because VS2022 will always support true color.
I think relying on "WT_SESSION" isn't ensured because if you have others terminals open besides "Windows Terminal" the environment variable will always be available and then you don't have certain if is your application or not that is running on "Windows Terminal". In this case the `GetProcessName` method is more accurate. It's my opinions, for now to debug, your solution is fine.

BDisp: > It's pretty clear to me that relying on the process name is not going to work reliably.

For debug mode you are right, but for release mode is more reliable. Put inside an #if definition.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -1760,9 +1751,9 @@ internal override void End ()
 		WinConsole?.Cleanup ();
 		WinConsole = null;
 
-		if (!RunningUnitTests && (_isWindowsTerminal || _parentProcessName == "devenv")) {
+		if (!RunningUnitTests && _isWindowsTerminal) {
 			// Disable alternative screen buffer.
-			Console.Out.Write (EscSeqUtils.CSI_RestoreCursorAndActivateAltBufferWithBackscroll);
+			Console.Out.Write (EscSeqUtils.CSI_RestoreAltBufferWithBackscroll);
```
BDisp, line 1730: The name is correct but the value must be 1049 and `CSI_RestoreCursorAndActivateAltBufferWithBackscroll` should be 1047. The same for `CSI_ActivateAltBufferNoBackscroll`.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
tig: In what scenario for the release version is there a problem that requires this check?

tig: > > Is this the problem you are talking about?
> 
> Not really, your workaround is suffice when you are debugging from VS2022 by pressing F5 and that work well because VS2022 will always support true color. I think relying on "WT_SESSION" isn't ensured because if you have others terminals open besides "Windows Terminal" the environment variable will always be available and then you don't have certain if is your application or not that is running on "Windows Terminal". In this case the `GetProcessName` method is more accurate. It's my opinion, for now to debug, your solution is fine.

I don't understand your concern. If the user runs a Terminal.Gui app on Windows in a terminal besides Windows Terminal, WT_SESSION will not be set. Environment variables can be either per-process or system wide. In this case we're talking about the WT process setting WT_SESSION, so it will only be scoped to that process.

BDisp: > In what scenario for the release version is there a problem that requires this check?

Easy. Run under cmd or conhost directly. If true is enabled you'll see a black and white screen with a bunch of Ansi escape sequences.

BDisp: Sorry @tig, it seems that you already fix it. Now `_isWindowsTerminal` is false by default and that fix it.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: This issue reappear because of this. The status bar stay visible at the bottom after exit.

Captura de ecrã 2023-10-20 190548

tig: Sometimes I just do dumb things. Fixed.

BDisp: It's possible to invert the values on these fields? Because the current names are inverted.

tig: > It's possible to invert the values on these fields? Because the current names are inverted.

I'm not sure what you mean. I just went back and re-read the documentation and I believe these are accurate:

```cs
	/// <summary>
	/// ESC [ ? 1047 h - Activate xterm alternative buffer (no backscroll)
	/// </summary>
	/// <remarks>
	/// From https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_
	/// Use Alternate Screen Buffer, xterm. 
	/// </remarks>
	public static readonly string CSI_ActivateAltBufferNoBackscroll = CSI + "?1047h";

	/// <summary>
	/// ESC [ ? 1047 l - Restore xterm working buffer (with backscroll)
	/// </summary>
	/// <remarks>
	/// From https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_
	/// Use Normal Screen Buffer, xterm.  Clear the screen first if in the Alternate Screen Buffer.
	/// </remarks>
	public static readonly string CSI_RestoreAltBufferWithBackscroll = CSI + "?1047l";

	/// <summary>
	/// ESC [ ? 1049 h - Save cursor position and activate xterm alternative buffer (no backscroll)
	/// </summary>
	/// <remarks>
	/// From https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_
	/// Save cursor as in DECSC, xterm. After saving the cursor, switch to the Alternate Screen Buffer,
	/// clearing it first.
	/// This control combines the effects of the 1047 and 1048 modes.
	/// Use this with terminfo-based applications rather than the 47 mode.
	/// </remarks>
	public static readonly string CSI_SaveCursorAndActivateAltBufferNoBackscroll = CSI + "?1049h";

	/// <summary>
	/// ESC [ ? 1049 l - Restore cursor position and restore xterm working buffer (with backscroll)
	/// </summary>
	/// <remarks>
	/// From https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_
	/// Use Normal Screen Buffer and restore cursor as in DECRC, xterm.
	/// resource.This combines the effects of the 1047 and 1048  modes.
	/// </remarks>
	public static readonly string CSI_RestoreCursorAndActivateAltBufferWithBackscroll = CSI + "?1049l";
```

tig: Ahh, I now see the typo:

`CSI_RestoreCursorAndActivateAltBufferWithBackscroll` should be `CSI_RestoreCursorAndRestoreAltBufferWithBackscroll`

BDisp: I knew you'd get there :-)

PR #2931, opened 2023-10-24
Fixes #2930. Console.CursorLeft-- throw an exception with NetDriver on Linux (v2).

### LINKED ISSUE #2930: Console.CursorLeft-- throw an exception with NetDriver on Linux (v2).
**Describe the bug**
If non-BMP character exist this code will throw an exception `Console.CursorLeft--;` on the `UpdateScreen` method. The reason is that `Linux` doesn't support `Console.CursorTop` and `Console.CursorLeft` and the returned values are invalid with out of range error.

**To Reproduce**
1. Go to 'WSL: UICatalog -usc'
2. Open 'Text Input Controls' scenario
3. Move the mouse
4. See error:
```
System.ArgumentOutOfRangeException: 'The value must be greater than or equal to zero and less than the console's buffer size in that dimension. (Parameter 'left')
Actual value was 4219246.'
```

**Expected behavior**
Do nothing because only the cursor was moved.

**Screenshots**
If not attached to the debugger:
image

If attached to the debugger:
image

BDisp:
Fixes #2930 - Linux doesn't support `Console.CursorTop` and `Console.CursorLeft` and that why the Ansi escape sequences is used for. The problem with the `CSI_SetCursorPosition` was that the parameters are inverted and thus was disabled.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Application.cs
+++ b/Terminal.Gui/Application.cs
@@ -668,9 +668,6 @@ namespace Terminal.Gui {
 				} else if (Current.SuperView == null && Current?.Modal == true) {
 					Refresh ();
 				}
-				if (Driver.EnsureCursorVisibility ()) {
-					state.Toplevel.SetNeedsDisplay ();
-				}
 			}
 
 			firstIteration = false;
--- a/Terminal.Gui/Application.cs
+++ b/Terminal.Gui/Application.cs
@@ -480,7 +480,7 @@ namespace Terminal.Gui {
 #endif
 				resume = false;
 				var runState = Begin (view);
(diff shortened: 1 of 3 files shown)


BDisp:
@tig I think we can get rid of the `SetWindowPosition` method because we don't use anymore the scroll buffer. In addition it always fallback to `Top = 0` and `Left = 0`. It's only used if `RunningUnitTests` is `true` and I think this should only exist in the `FakeDriver`.

tig (approved):
Nice find and fix. Thank you!

PR #2932, opened 2023-10-24
Partially Addresses #2616. Support combining sequences that don't normalize

### LINKED ISSUE #2616: Support combining sequences that don't normalize

BDisp:
Partially addresses (but does not fix) #2616 

## Pull Request checklist:


UnitTests/ConsoleDrivers/ContentsTests.cs
```diff
@@ -17,40 +17,40 @@ public ContentsTests (ITestOutputHelper output)
 		this.output = output;
 	}
 
-	[Theory]
+	[Theory, AutoInitShutdown]
```
tig, line 20: I do not want these tests to depend on Application. Please remove this.

UnitTests/ConsoleDrivers/ContentsTests.cs
```diff
@@ -17,40 +17,40 @@ public ContentsTests (ITestOutputHelper output)
...
 	[InlineData (typeof (FakeDriver))]
 	//[InlineData (typeof (NetDriver))]
 	//[InlineData (typeof (CursesDriver))]
 	//[InlineData (typeof (WindowsDriver))]
 	public void AddStr_With_Combining_Characters (Type driverType)
 	{
-		var driver = (ConsoleDriver)Activator.CreateInstance (driverType);
-		Application.Init (driver);
-		// driver.Init (null);
+		Application.Init ();
```
tig, line 27: This should not be needed for ConsoleDriver tests.

tig (changes requested):

UnitTests/ConsoleDrivers/ContentsTests.cs
BDisp: Yes, it's needed because it was getting failure on another unit test due any racing condition while this unit test is running. Running the other unit test alone it doesn't give any failure but running all the unit tests I get failing on the other unit test. I don't remember the name of the other fail unit test. But try to run without `Application.Init` and `AutoInitShutdown` and check for yourself.

tig: That's a bug in the unit test system then. Let's figure that out instead of coupling Application to ConsoleDriver.

tig:
PR to this PR incoming with suggested fixes to Unit Tests...

UnitTests/ConsoleDrivers/ContentsTests.cs
BDisp: You have here a dependence on `Application`, but I'll try to fix this.
`Application.Init (driver);`

tig: See my PR I just submitted to this PR.

tig (changes requested):
Please merge https://github.com/gui-cs/Terminal.Gui/pull/2934 and see my comment.

I don't think this fix does what you think it does.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -175,11 +175,14 @@ public void AddRune (Rune rune)
 				// Normalize to Form C (Canonical Composition)
 				string normalized = combined.Normalize (NormalizationForm.FormC);
 
-				Contents [Row, Col - 1].Runes = new List<Rune> { (Rune)normalized [0] }; ;
+				List<Rune> runes = new List<Rune> ();
```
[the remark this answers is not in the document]
BDisp: Well @tig, that code does what you think it does. Of course in the `UpdateScreen` the code must reflect this change, but isn't yet ready.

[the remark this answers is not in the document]
tig: When I reviewed this, I thought the PR was marked as ready for review. I may have mis-read that. Sorry for the unneeded comment.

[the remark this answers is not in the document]
BDisp: No you didn't have mis-read that. I wanted to say that it could be for another PR. This was only to make the unit test pass. But I'll try to do at least in `NetDriver`. I see a problem with the `Console.Write(output)` by not normalizing `ǫ́`. The string builder has the the right string on it but the `Console.Write` doesn't render it on the screen. Can you check that please by doing some test, please? Thanks.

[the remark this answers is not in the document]
BDisp: @tig see what I mean. This should outputted `é́` but show `ę`. I really don't know why because the string builder has the right output.

image

image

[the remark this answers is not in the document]
tig: This is because conhost does not know how to render combining marks that can't be normalized. As far as I can tell NO terminal does. 

Hence why i abandoned trying to make what you're working on work.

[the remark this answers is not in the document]
BDisp: Yes pasting in the Windows Terminal only the second paste well and the first is bad :-). At least the code is right and when terminal support it later it's already done.

image

BDisp:
@tig I already finished this.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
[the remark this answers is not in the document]
tig: I may be wrong: https://github.com/microsoft/terminal/pull/12307

Using `Cascadia Code` in WT

```pwsh
echo "[e`u{0301}`u{0301}]"
```

Prints:
image

Note that not all fonts work. E.g. FiraCode NFM Retna:
image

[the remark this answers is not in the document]
BDisp: Using Cascadia Code in WT

image

tig:
> @tig I already finished this.

I don't think your test in TabView is working. Here's what it should look like:

image

Here's what I see:

image

Another test (in Unicode scenario):

```cs
sb.Append ('e');
sb.Append ('\u0301');
sb.Append ('\u0301');
testlabel = new Label ($"Should be an e with two accents: {sb}") { X = 20, Y = Pos.Y (label), Width = Dim.Percent (50), CanFocus = true, HotKeySpecifier = new Rune ('&') };
Win.Add (testlabel);
```

image

UICatalog/Scenarios/TabViewExample.cs
```diff
@@ -68,6 +68,7 @@ public override void Setup ()
 				 new Label ("This tab has a very long name which should be truncated.  See TabView.MaxTabTextWidth")),
 				 false);
 			tabView.AddTab (new Tab ("Les Mise" + Char.ConvertFromUtf32 (Int32.Parse ("0301", NumberStyles.HexNumber)) + "rables", new Label ("This tab name is unicode")), false);
+			tabView.AddTab (new Tab ("Les Mise" + Char.ConvertFromUtf32 (Int32.Parse ("0328", NumberStyles.HexNumber)) + Char.ConvertFromUtf32 (Int32.Parse ("0301", NumberStyles.HexNumber)) + "rables", new Label ("This tab name has two unicode combining masks")), false);
```
tig, line 71: Simpler:

```cs
tabView.AddTab (new Tab ("Les Mise" + '\u0301' + "rables", new Label ("This tab name is unicode")), false);
tabView.AddTab (new Tab ("Les Mise" + '\u0328' + '\u0301' + "rables", new Label ("This tab name has two unicode combining masks")), false);
```

Also, I think it woudl be better to expand the Unicode scenario than to bury these tests in TabView Example, no?

tig: The glyph that's supposed to appear with `e\u0328\u0301` is 
image

tig (changes requested):
I still don't think this is working. See my comments.

Terminal.Gui/Drawing/Cell.cs
```diff
@@ -23,7 +23,7 @@ public class Cell {
 	///// <remarks>
 	///// Only valid in the rare case where <see cref="Rune"/> is a combining sequence that could not be normalized to a single Rune.
 	///// </remarks>
-	//internal Rune CombiningMark { get; set; }
+	internal Rune CombiningMark { get; set; }
```
tig, line 26: I believe you were correct in asserting this should be `CombiningMarks` and be a list:

image

tig (changes requested):

UICatalog/Scenarios/TabViewExample.cs
BDisp: Well I think I already told you that the string builder appear with the expected unicode but not in the screen. You said that terminal doesn't support it. Do you already saw this https://github.com/gui-cs/Terminal.Gui/pull/2932#discussion_r1372303371?
It's very explicitly explained that the output code is right but the `Console.Write` doesn't render it well.

Terminal.Gui/Drawing/Cell.cs
BDisp: In this case no. It's only needed do add to the `CombiningMarks` if the `normalized` string length on the `ConsoleDriver` is grater than 1, otherwise is 0.

BDisp: I do not waste more time with this. For sure it's a limitation of the terminal. If you can put it working then you'll deserve a medal :-)

tig:
Perhaps we shouldn't try normalizing at all? Like this:

image

Re: NetDriver - I suspect Console.Write can't handle combining marks.

Terminal.Gui/Drawing/Cell.cs
tig: However, I think I was partially right: the .NET Console API does not properly pass combining marks on. Thus combining marks will not work on `NetDriver`.

BDisp:
> Perhaps we shouldn't try normalizing at all? Like this:

There is no problem on the AddRune because on the `UpdateScreen` I can get the correct string in the output variable,

> Re: NetDriver - I suspect Console.Write can't handle combining marks.

Yes I believe is that.

Terminal.Gui/Drawing/Cell.cs
tig: WindowsDriver will need signficant refactoring to utilize `Cell.CombingMarks`. 

a) It will never work with them for `force16colors` because `WriteConsoleOutput` only supports a `char` per buffer position.
b) `WriteConsole` may or may not work like `Console.Write`. Regardless, `WindowsConsole.ExtendedCharInfo` needs to be expanded to hold the combining marks. Might as well just nuke that data structure and replace it with `Cell`, no?

BDisp: The `StringBuilder` handles correctly but the `System.Console` doesn't.

tig:
> > Perhaps we shouldn't try normalizing at all? Like this:
> 
> There is no problem on the AddRune because on the `UpdateScreen` I can get the correct string in the output variable,

Do you have an objection to replacing the current code in this PR with the code I suggested above that removes the calls to normalize?

BDisp:
> Do you have an objection to replacing the current code in this PR with the code I suggested above that removes the calls to normalize?

@tig did you already saw what it do? You are replacing the previous p

> Do you have an objection to replacing the current code in this PR with the code I suggested above that removes the calls to normalize?

This is worst.

image

Terminal.Gui/Drawing/Cell.cs
BDisp: > WindowsDriver will need signficant refactoring to utilize `Cell.CombingMarks`.
> 
> a) It will never work with them for `force16colors` because `WriteConsoleOutput` only supports a `char` per buffer position.

It's possible to put surrogate pair into two consecutive char position and the `WriteConsoleOutput` will render well. The problem is the dealignment caused on the right. I didn't yet found a solution for that but I already can write surrogate pairs on `WindowsDriver`, instead of showing the rune replacements.

> b) `WriteConsole` may or may not work like `Console.Write`. Regardless, `WindowsConsole.ExtendedCharInfo` needs to be expanded to hold the combining marks. Might as well just nuke that data structure and replace it with `Cell`, no?

I really didn't look well into this yet because it deal with Ansi escape sequences which is only supported by the Windows Terminal.

BDisp:
Absolute chaos.

image

BDisp:
@tig sorry but I don't have more head to handle with this. I think it's reasonable for now, attended to the difficult degree you added to this.

tig:
I dove in deep this morning and I think I understand how all this should work much better. I composed a message on my home computer but didn't have a chance to finish my thoughts before I had to run. As soon as I get home I'll finish it.

The net is: We *can* make combining marks work on both NetDriver and WindowsDriver, but the current design of `AddRune/AddStr/UpdateScreen` will not work. In addition `TextFormatter` (and others) are doing things in such a way that will not work.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -175,11 +175,22 @@ public void AddRune (Rune rune)
 				// Normalize to Form C (Canonical Composition)
 				string normalized = combined.Normalize (NormalizationForm.FormC);
 
-				Contents [Row, Col - 1].Rune = (Rune)normalized [0]; ;
+				if (Contents [Row, Col - 1].Rune != default && Contents [Row, Col - 1].Rune != (Rune)' ') {
```
tig, line 178: This is not sufficient. Many, many, characters cannot be combined with a combining mark, not just ' '. Try, for example `[`.

WT (and other platforms) have sophisticated algorithms for figuring out whether a combining mark can combine with a base character. I am trying to find more details on these to see what we can do.

You can easily test this in WT and PS with (Caskadia Cove Nerd Font):

image

> In Unicode, not all base characters can be combined with combining characters. Whether a base character can be combined with a combining character depends on the rules defined in the Unicode Standard. The Unicode Consortium, the organization responsible for maintaining the Unicode Standard, specifies these rules.

> Here are some general guidelines to determine if a base character can be combined with a combining character:

> Character Composition Model: Unicode follows a character composition model, which means that some characters can be composed from a base character and one or more combining characters. This allows for a wide range of characters and diacritics to be represented in text.

> Compatibility: Unicode defines compatibility characters and compatibility composition rules to ensure that combining characters work as expected with base characters. If a base character is defined as a "combining character target," it means that it can be combined with one or more combining characters.

> Normalization Forms: Unicode defines normalization forms (NFC, NFD, NFKC, NFKD) to handle character composition and decomposition. NFC (Normalization Form C) is the most commonly used form for combining characters, and it ensures that text is represented in a composed form when possible.

> Character Properties: Unicode assigns specific properties to each character, including whether it can be used as a base character and whether it can combine with combining characters. You can refer to the Unicode Character Database (UCD) to check the properties of individual characters.

> Combining Class: Combining characters are assigned a "combining class" value, which determines their combining behavior. Base characters and combining characters are combined according to their combining class values.

> Compatibility Decomposition: Some base characters can be combined using compatibility decomposition mappings, even if they don't have explicit combining characters. These mappings are defined in the Unicode Standard.

> To determine whether a specific base character can be combined with a combining character, you can consult the Unicode Standard documentation, specifically the Unicode Character Database (UCD) and the Unicode Technical Reports related to normalization and composition. Additionally, Unicode-aware text processing libraries and programming languages often provide functions or methods for handling character composition and decomposition, making it easier to work with combined characters in text processing.

tig (changes requested):

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
BDisp: @tig I only had to do that because of the `CharacterMap` scenario using the space to separate the runes to avoid incorrect output. I don't have enough knowledge to do better.

tig:
Good info:
* https://github.com/microsoft/terminal/issues/16132

BDisp:
@tig try this code on `WindowsDriver` on the `WriteToConsole` method. It seems to support true color but I need help.

```cs
	[DllImport ("kernel32.dll", SetLastError = true)]
	//public static extern bool SetConsoleTextAttribute (IntPtr hConsoleOutput, CharacterAttributes wAttributes);
	public static extern bool SetConsoleTextAttribute (IntPtr hConsoleOutput, int wAttributes);

	public enum CharacterAttributes {
		FOREGROUND_BLUE = 0x0001,
		FOREGROUND_GREEN = 0x0002,
		FOREGROUND_RED = 0x0004,
		FOREGROUND_INTENSITY = 0x0008,
		BACKGROUND_BLUE = 0x0010,
		BACKGROUND_GREEN = 0x0020,
		BACKGROUND_RED = 0x0040,
		BACKGROUND_INTENSITY = 0x0080,
		COMMON_LVB_LEADING_BYTE = 0x0100,
		COMMON_LVB_TRAILING_BYTE = 0x0200,
		COMMON_LVB_GRID_HORIZONTAL = 0x0400,
		COMMON_LVB_GRID_LVERTICAL = 0x0800,
		COMMON_LVB_GRID_RVERTICAL = 0x1000,
		COMMON_LVB_REVERSE_VIDEO = 0x4000,
		COMMON_LVB_UNDERSCORE = 0x8000
	}


		if (force16Colors) {
			int i = 0;
			//CharInfo [] ci = new CharInfo [charInfoBuffer.Length];
			//foreach (ExtendedCharInfo info in charInfoBuffer) {
			//	ci [i++] = new CharInfo () {
			//		Char = new CharUnion () { UnicodeChar = info.Char },
			//		Attributes = (ushort)(((int)info.Attribute.Foreground.ColorName) | ((int)info.Attribute.Background.ColorName << 4))
			//	};
			//}

			//result = WriteConsoleOutput (_screenBuffer, ci, bufferSize, new Coord () { X = window.Left, Y = window.Top }, ref window);

			_stringBuilder.Clear ();

			//var csbi = new CONSOLE_SCREEN_BUFFER_INFOEX ();
			//csbi.cbSize = (uint)Marshal.SizeOf (csbi);
			//if (!GetConsoleScreenBufferInfoEx (_screenBuffer, ref csbi)) {
			//	var err = Marshal.GetLastWin32Error ();
			//	if (err != 0) {
			//		throw new System.ComponentModel.Win32Exception (err);
			//	}
			//}
			if (!SetCursorPosition (new Coord (0, 0))) {
				var err = Marshal.GetLastWin32Error ();
				if (err != 0) {
					throw new System.ComponentModel.Win32Exception (err);
				}
			}

			Attribute? prev = null;
			foreach (var ci in charInfoBuffer) {
				var attr = ci.Attribute;

				if (attr != prev) {
					prev = attr;
					if (!SetConsoleTextAttribute (_screenBuffer,
						(attr.Foreground.R + attr.Foreground.G + attr.Foreground.B & attr.Foreground.A) |
						(attr.Background.R + attr.Background.G + attr.Background.B & attr.Background.A))) {

						var err = Marshal.GetLastWin32Error ();
						if (err != 0) {
							throw new System.ComponentModel.Win32Exception (err);
						}
					}
				}

				if (!ci.Empty) {
					_stringBuilder.Append (ci.Char);
				}
			}

			//if (!SetCursorPosition (csbi.dwCursorPosition)) {
			//	var err = Marshal.GetLastWin32Error ();
			//	if (err != 0) {
			//		throw new System.ComponentModel.Win32Exception (err);
			//	}
			//}

			string s = _stringBuilder.ToString ();

			result = WriteConsole (_screenBuffer, s, (uint)(s.Length), out uint _, null);

```

tig:
> @tig try this code on `WindowsDriver` on the `WriteToConsole` method. It seems to support true color but I need help.

Happy to. May be a few days though.

I'm not seeing how this is relevant to combining sequences though. Can you explain?

BDisp:
No it's not. Forget.

tig:
- Combining Marks have `GetColumns() == 0`
- `TextFormatter.Draw` is currently doing `var runeWidth = Math.Max (rune.GetColumns (), 1);` and thus incrementing the column when a combining mark is encountered.
- `AddRune` always assumes a combining mark is 0 width and should be combined with the preceding Cell (unless Col == 0). 
- AtlasEngine (`Console.Write("a\u0301\u0301")`) DOES render multiple CMs correctly.

XaOqxdo 1

- However, it ALSO moves the cursor to the right for each CM. Thus `Console.Write("a\u0301\u0301\u0328x")` results in:
image

Note, this is the same thing you see with `echo "[a`u{0301}`u{0301}x]"`:

image

If I `SetCursorPostion()` to the row/col to just to the right of the `a`, WT 'forgets' the CM.

In otherwords, internally AtlasEngine is holding each CM as a used column. 

- AtlasEngine also, treats individual CMs (`Console.Write("\u0301")`) as 1 column width glyphs if in the 1st column.

image

Unless, I'm missing something, until AtlasEngine changes, on Windows, we cannot really support NON-NORMALIZED combining marks in a way that users would expect. Any glyph made up of a base char and n or more combining chars will use base.ColumnWidth() + n columns.

I think for us to support combining marks correctly we need to do the following:

- Always try to normalize. This at least lets the most common combining mark use-case (base + 1 CM) work in many cases. 
- Have `ConsoleDriver.AddRune` treat a combining mark as a 1 column wide glyph. E.g. if `rune.IsCombiningMark() == true` then ConsoleWidth == 1.
- Have `ConsoleDriver.AddStr` enumerate the runes in `str` and add them to `ConsoleDriver.Contents` such that `Cell.CombiningMarks` is setup correctly (if `str[0]` is a CM, special case).
- Drivers' `UpdateScreen` 
   - treat any `Cell.Rune.IsCombiningMark() == true` as taking 1 column.
   - use `Cell.Rune.CombiningMarks().Count > 0` to append the marks to `output`.

tig:
Please see my PR to this PR: https://github.com/BDisp/Terminal.Gui/pull/162

I do not think #2616 can be fixed at this time. I think the best we can do is basically ignore CMs that can't be normalized. My PR to your PR does this.

BDisp:
> Please see my PR to this PR: [BDisp#162](https://github.com/BDisp/Terminal.Gui/pull/162)
> 
> I do not think #2616 can be fixed at this time. I think the best we can do is basically ignore CMs that can't be normalized. My PR to your PR does this.

Done @tig. Thanks.

BDisp:
@tig why you write `u301` twice here `\u0301\u0301\u0328`?

tig:
> @tig why you write `u301` twice here `\u0301\u0301\u0328`?

It was just an example of 3 CMs.

BDisp:
> > @tig why you write `u301` twice here `\u0301\u0301\u0328`?
> 
> It was just an example of 3 CMs.

I see. For have sure it could be normalized. You don't know if there is 3 different combining marks that can normalize on terminal? It would be better.

tig:
> > > @tig why you write `u301` twice here `\u0301\u0301\u0328`?
> > 
> > 
> > It was just an example of 3 CMs.
> 
> I see. For have sure it could be normalized. You don't know if there is 3 different combining marks that can normalize on terminal? It would be better.


`\u0301\u0301\u0328` will not normalize! It results in

image

`e\u0301\u0301\u0328` will not normalize. It results in:

image

You can pile on as many `\u0301` as you want:

image

tig (approved):

PR #2940, opened 2023-10-29
Adds `ANSIDriver` - Pure "Console Virtual Terminal Sequences" driver

### LINKED ISSUE #881: Modifier key are not fully supported with CursesDriver

### LINKED ISSUE #2610: Refactor `ConsoleDriver`s to support ANSI sequences natively (e.g. Virtual Terminal Sequences on Windows)

### LINKED ISSUE #2796: CursesDriver doesn't render wide codepoints correctly
Curses is doing something funky with glyphs that report GetColums() of 1 yet are rendered wide. E.g. 0x2064 (invisible times) is reported as 1 column but is rendered as 2. WindowsDriver & NetDriver correctly render this as 1 column, overlapping the next cell.

I've spent many hours trying to figure this out. I'm at a loss.

JT9gIoo 1

I'm pretty sure curses is actively trying to determine a codepoint is wide and ignoring `wcwidth` (which it does use, and I've verified is that latest).

I suspect the way to address this is to stop using `Curses.addstr` etc... and instead emit ANSI sequences directly to the output stream.

tig:
## Fixes

- Fixes #2610
- Fixes #2796 
- Fixes #881 

## Todo

- [x] Add a new driver called `ANSIDriver` that uses ANSI ESC sequences for drawing and cursor (not input).
- [x] Add `Application.ForceDriver` as a `ConfigurationManager` property to replace `UseSystemConsole` and `UICatalog -usc`
- [ ] Remove unneeded legacy WindowsDriver cruft from `ANSIDriver`
- [ ] Fix bugs & issues on Windows
- [ ] Determine if `ANSIDriver` can be a basis for a new Linux driver too (re: #2796, #881, ...)
- [ ] Update all documentaion

## Pull Request checklist:


BDisp:
@tig sometimes `WindowsDriver` on `Windows Terminal` doesn't refresh after resizing.

image

BDisp:
@tig I understand why you are adding `ANSIDriver` which is a very good idea. But I think the reason for that decision isn't the more correct. In my opinion it should be a cross platform to improve the `NetDriver` instead of `WindowsDriver`. You know that `NetDriver` read and process Ansi escape sequences very slower. Also writing is slower as well. But on Linux it's much faster. So, `ANSIDriver` should use the Win32 API as it now working, as well the input keys and mouse to work on Windows and will use Ansi escape sequences to write to the console. On Linux `ANSIDriver` will use the NetEvents and `System.Console`. Thus `ANSIDriver` would be a cross platform that will run much more faster on Windows and Linux. Right now it's only a duplicate of WindowsDriver that is using Ansi escape sequences to write to console to support true color.

BDisp:
@tig I submitted the PR https://github.com/tig/Terminal.Gui/pull/16 to your branch. When you have time please tell me what you think. Thanks.

BDisp:
For the profile settings work properly this rule must be follow:
- To debug WT the WT settings must be "Let Windows decide or Windows Terminal".
- To debug CH the WT settings must be "Host Windows Console".

PR #2941, opened 2023-10-30
Fixes #2928. In the v2 version, using Chinese will cause UI confusion.

### LINKED ISSUE #2928: In the v2 version, using Chinese will cause UI confusion.
**Describe the bug**
In the v2 version, using Chinese will cause UI confusion, the same code runs well in the v1 version.

**Screenshots**
v1: 
OK

v2:
ERR

 - OS: Windows 11
 - Version [2.0.0-pre.191]

**code**

```
static void Main()
{
    Application.Init();
    var top = Application.Top;
    top.Closing += (sender, e) =>
    {
        var result = MessageBox.Query("Confirm", 
            "Are you sure you want to quit ui?",  0,
            "Yes", "No");
        if (result != 0)
        {
            e.Cancel = true;
        }

    };

    var win = new Window()
    {
        Title = "Test",
        X = 0,
        Y = 0,
        Width = Dim.Fill(),
        Height = Dim.Fill(),
    };
    top.Add(win);


    var buttonPanel = new FrameView()
    {
        Title = "Command",
        X = 0,
        Y = 1,
        Width = Dim.Fill(),
        Height = 5,
    };
    win.Add(buttonPanel);

    
    buttonPanel.Add(
        new Button(1, 1, "你"), // v1: A
        new Button(12, 1, "好"), // v1: B
        new Button(22, 1, "呀")// v1: C
    
    );

    Application.Run();
}
```

BDisp:
Fixes #2928- Fixes wide and non-bmp unit points.

## Pull Request checklist:


BDisp:
This test `Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor` sometimes fail and we need to check why.
```
  Failed Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor [43 ms]
  Error Message:
   System.Exception : Unexpected color White,Black was used at row 0 and col 0 (indexes start at 0).  Color value was White,Black (expected colors were -1,0)
  Stack Trace:
     at Terminal.Gui.TestHelpers.AssertDriverColorsAre(String expectedLook, ConsoleDriver driver, Attribute[] expectedColors) in /home/runner/work/Terminal.Gui/Terminal.Gui/UnitTests/TestHelpers.cs:line 336
   at Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor() in /home/runner/work/Terminal.Gui/Terminal.Gui/UnitTests/Views/TreeViewTests.cs:line 868
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
```

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -32,44 +33,131 @@ internal class WindowsConsole {
...
+
 		if (force16Colors) {
-			int i = 0;
-			CharInfo [] ci = new CharInfo [charInfoBuffer.Length];
-			foreach (ExtendedCharInfo info in charInfoBuffer) {
-				ci [i++] = new CharInfo () {
-					Char = new CharUnion () { UnicodeChar = info.Char },
-					Attributes = (ushort)(((int)info.Attribute.Foreground.ColorName) | ((int)info.Attribute.Background.ColorName << 4))
-				};
+			var csbi = new CONSOLE_SCREEN_BUFFER_INFOEX ();
```
tig, line 90: As far as I can tell, this force16 color stuff has nothing to do with the Chinese issue. 

Can it be removed from this PR? Perhaps we can fold this into the work I'm doing here: https://github.com/gui-cs/Terminal.Gui/pull/2940

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -32,44 +33,131 @@ internal class WindowsConsole {
...
+		newInConsoleMode &= ~(uint)ConsoleModes.EnableQuickEditMode;
+		newInConsoleMode &= ~(uint)ConsoleModes.EnableProcessedInput;
+		InputConsoleMode = newInConsoleMode;
+
+		_originalOutputConsoleMode = OutputConsoleMode;
+		var newOutConsoleMode = _originalOutputConsoleMode;
+		if (!ConsoleDriver.RunningUnitTests && (newOutConsoleMode & (uint)(ConsoleModes.EnableVirtualTerminalProcessing | ConsoleModes.DisableNewlineAutoReturn))
+			< (uint)ConsoleModes.DisableNewlineAutoReturn) {
+
+			newOutConsoleMode |= (uint)(ConsoleModes.EnableVirtualTerminalProcessing | ConsoleModes.DisableNewlineAutoReturn);
```
tig, line 74: Please don't enable VTS in WindowsConsole.cs. We are going to enable it in `ANSIConsole` and try to keep `WindowsConsole` as old-school as possilbe (perhaps ONLY supporting 16 colors)...

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -773,34 +929,97 @@ public COLORREF (uint value)
...
 			Clipboard = new FakeDriver.FakeClipboard ();
 		}
+	}
 
-		// TODO: if some other Windows-based terminal supports true color, update this logic to not
-		// force 16color mode (.e.g ConEmu which really doesn't work well at all).
-		_isWindowsTerminal = Environment.GetEnvironmentVariable ("WT_SESSION") != null;
-		if (!_isWindowsTerminal) {
-			Force16Colors = true;
+	ConsoleHost GetConsoleHost ()
```
tig, line 987: This has nothing to do with Chinese rendering, right? Please remove it from this PR.

tig (changes requested):
Please see my comments.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: @tig with this change I aldo resolved the non-bmp displaying on `WindowsDriver` without using `Windows Terminal`.

image

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: I need that for render the non-bmp properly. Otherwise it appear as bellow (see the black rectangle at right). This is only for output and not for input. If we want old-school work better I think we should keep it.

image

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: Was the only way I could find to run on `Windows Terminal` or on `conhost` dependent of the Windows Terminal configuration.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: Only to remind you that I'm not using Ansi escape sequences on the old-school because it isn't really possible without using Windows Terminal. The only way I could find to resolve this was print each row and column with some control on the last column avoid jumping to the next row if a non-bmp is handled. To achieve this was only possible with the `ConsoleModes.EnableVirtualTerminalProcessing | ConsoleModes.DisableNewlineAutoReturn` together.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: Was the most simple way instead of creating a lot of profile settings for now.

BDisp: Done

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: I already removed `ConsoleModes.EnableVirtualTerminalProcessing`.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: @tig I really need this, otherwise I get the error bellow. This is needed for `conhost` restore the terminal defaults on exit (otherwise you will the last color used by the screen buffer). On Windows Terminal this code throw the exception because `csbi.srWindow.Bottom += 2;` cause a invalid value, since it use alternative screen. But I need this on `conhost` to adjust the size, otherwise the size will decrease always when you open and close a scenario. The only way I have to test both in this PR is using the `GetConsoleHost` method. If you later don't need this do whathever you want.

image

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: Run a release version from cmd.exe (a new line is inserted at end):

image

Run a debug version from cmd.exe:

image

I really need to use `ConsoleModes.EnableVirtualTerminalProcessing | ConsoleModes.DisableNewlineAutoReturn` together for this working on release version as well. It seems that the VS2022 is dealing well with the `VSDebugger` even without `ConsoleModes.EnableVirtualTerminalProcessing`, but not with `cmd` or `conhost`. I only need it for dealing with endlines based on the type of the rune.

tig:
@BDisp  you are chasing your tail here. 

Please see my fix to #2610. You'll note that the `Chinese.cs` scenario works fine:

- Conhost
  - NetDriver 

- WT
  - NetDriver
  - ANSIDriver  

Please stop trying to hack WindowsDriver to fix this; it's not worth it IMO. If people want to use conhost, they should use netdriver. 

Note `ANSIDriver` is a LONG way from being finished, but it is the right long term architecture. I plan on making it the basis for an updated curses driver too!

BDisp:
> @BDisp you are chasing your tail here.
> 
I don't think so.

> Please see my fix to #2610. You'll note that the `Chinese.cs` scenario works fine:
> 
> * Conhost
>   
>   * NetDriver
> * WT
>   
>   * NetDriver
>   * ANSIDriver
> 
Yes, but where is the 'WindowsDriver', the faster one like the 'CursesDriver'?

> Please stop trying to hack WindowsDriver to fix this; it's not worth it IMO. If people want to use conhost, they should use netdriver.
> 
Yes and they will freezing on Windows using `NetDriver`. I hate use it on Windows to be honest.

> Note `ANSIDriver` is a LONG way from being finished, but it is the right long term architecture. I plan on making it the basis for an updated curses driver too!

Yes you are right but to use Ansi escape sequences on Windows only with `Windows Terminal` works more or less. You are forgetting many others users that use `cmd` or `conhost` and I'm showing a workaround to this without Ansi escape sequences to support all code points.

BDisp:
When a user want to run `WindowsDriver` on `cmd` or `conhost` he really want to see this:

image

and not this:

image

BDisp:
@tig I think the best option is submit a PR to your branch and close this one. Do you agree?

PR #2949, opened 2023-10-31
Fixes #2943. Default button bracket glyphs don't work on conhost.

### LINKED ISSUE #2943: Wide Glyphs do not work correctly on `conhost`
Run the scenario `Buttons` with host console and you'll see the same. I really don't know why this happens but I'm sure the right string are passed to the console.

image

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/2928#issuecomment-1785413495_
            
This problem is not just about `Button`. `conhost` operates differently from WT when rendering wide glyphs. 

Run CharMap under conhost to see:
image

Note that `NetDriver` doesn't work correctly either.:
image

BDisp:
Fixes #2943 - Force to `[` and `]` on `WindowsDriver` if `Force16Colors` is true.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
@@ -1545,19 +1545,15 @@ internal class WindowsDriver : ConsoleDriver {
 				}
 				_outputBuffer [position].Empty = false;
 				if (Contents [row, col].Rune.IsBmp) {
-					var rune = Contents[row, col].Rune;
-					if (Force16Colors)
-					{
-                        if (rune == Glyphs.LeftBracket)
-                        {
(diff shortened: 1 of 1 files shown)


tig:
This is not a correct fix. It only fixes the Button issue, not the deeper issue (which is conhost behaves entrely differently from WT w.r.t. wide glpyhs). I've updated #2943 to reflect this.

Note that `NetDriver` doesn't work correctly either. 

In addition, simply overriding the CM.Glyphs in `WindowsDriver` is hacky. If we eventually determine that `conhost` can't handle wide-glyphs we need to actually cause a different set of `CM.Glyphs` to be loaded. I'm going to close this PR. Please discuss this in #2943.

Issue #2814, opened 2023-08-16
Enable Close/Quit/Collapse/Expand buttons on `Border`


drweb86:
# Mini-Spec

Three scenarios so far:

- User clicks on a button in the Border (top/right) that causes a Modal to quit.
- User clicks on a button in the Border (top/right) that causes an Overlapped to close/hide.
- User clicks on a button in the Border (top/right) that causes the view to collapse/expand.

We will create 3 new View subclasses, one for each of these:

- `ApplicationQuitter` - Glyph is an X. `MouseClick`, and `KeyBinding (Space)` invoke `Command.QuitToplevel` (to be renamed to `Command.Close`). The command handler simply calls `Application.RequestStop()`. 
- `ViewHider` -   Glyph is an X. `MouseClick`, and `KeyBinding (Space)` invoke `Command.Close`. The command handler simply calls `SuperView.Parent.Visible = false`. (or perhaps we add `Close()` as a verb on View?).
- `ViewExpander` - Glyph is a up/down arrow.  `MouseClick`, and `KeyBinding (Space)` invoke `Command.ToggleExpandCollapse`. Command hander 
  - Disables/Hides/Enables/Shows `SuperView.Parent`'s subviews
  - Sets the Height of the `SuperView.Parent` to 3 or  `Dim.Auto`

Devs will add these views via `view.Border.Add`.

This is what `ViewExpander` will look like/do:

![9rGXIUs 1](https://github.com/gui-cs/Terminal.Gui/assets/585482/ffd8d852-b255-44b0-a0e0-34ecc20d46cb)

I've built a POC of `ViewExpander` in #3376 

# Background

This is a sub-issue of:

- #2994 

Will be fixed in this PR as a proof:

- #3278 

## Background

**Is your feature request related to a problem? Please describe.**
I want user to be able to close button by [X] button near dialog title from top right side. And I want them to be shown in Query dialogs by default. Because when you pass no button its not clear, that you need Escape button to be pressed to close it.

**Describe the solution you'd like**
For dialogs and querries to have [x] button: Dialog Title              [x]

**Describe alternatives you've considered**
Have close button as now.

BDisp:
I remember a long time ago, to add` [ Ok ]` button on `MessageBox` if no buttons are provided on the parameters, but wasn't accepted at the time. I really don't like when there is no button at all, but also I think if it's only to close the dialog it isn't need to explicit add a button in the parameters. Good luck :-)

[...]

Issue #2563, opened 2023-04-17
Adornments: Enable SubViews in `Margin`, `Border`, and `Padding`


tig:
This is a sub-issue of:

- #2994 

## Todo

- [x] What happens when `Adornment.Add` is called? Fixed in #3278 
  * Nothing special happens, but the various layout and draw methods now do the right thing.
  *  The thickness is not automatically adjusted. 
- [x] How should `MouseEvent`s work? Fixed in #3278 
- [ ] How should `Focus` work? Because `Padding` is in a different view hierarchy from the dialog, `Tab` etc... doesn't tab into the buttons. 
    *  `Padding` - Any views added to `Padding` are included in `Padding.Parent`'s focus chain. If `View.Subviews[^1].HasFocus` and the user presses `Tab` `Padding.SubViews[0]` will get focus. Etc...
    * `Border` - `Ctrl-F10` is used to set focus on `Border.Subviews[0]`. Subsequent `Tab` presses cycle through `Border.Subviews`. `Ctrl-F10` again returns focus to `Border.Parent`.
    * `Margin` - No default behavior. If someone adds views to `Margin` they are on their own to figure out a focus model (I can't think of any use case where it makes sense to add a subview to `Margin`.

## Background

  * Examples of SubViews in Adornments:
    * A `TitleBar` renders the View's `Title` and a horizontal line defining the top of the View. Added to `Border`.
      * One or more `LineView`s that render the View's 'Border` 
    * A `Vertical Scrollbar` adds thickness to `Padding.Right` (or `.Left` when right-to-left language support is added). 
    * A `Horizontal Scrollbar` adds thickness to `Padding.Bottom` when enabled.
    * A `MenuBar` adds thickness to `Padding.Top` (NOTE: This is a change from v1 where `subview.Y = 1` is required).
    * A `StatusBar` adds thickness ot `Padding.Bottom` and is rendered at the bottom of `Padding`.
    * In `Dialog` (and derived classes), the buttons can be added to `Padding` (`Y = Pos.AnchorEnd(1)`).

* Should expanding of, e.g. `Padding.Thickness` be automatic when `Padding.Add` is called? If `Dim.Auto` is implemented, it might be possible for this to happen. For now: I suggest devs adding adornments manually expand thickness as appropriate. In my Dialog prototype, I did this:

```cs
		public void AddButton (Button button)
		{
			if (button == null) {
				return;
			}
			
			_buttons.Add (button);
[...]

Issue #2489, opened 2023-04-04
Create new `ScrollBar` based on a new `Scroll` and remove `ScrollBarView/ScrollView`


tig:
This is a sub-issue of:

- #2994 

## Requires

- #3323 

# Design

Scroll bar support will be based on three views: `ScrollBar` which is a composite of two buttons supporting continuous button press and a `Scroll`, which is just the "middle" part of a scroll bar.

These views are described below.

## `Scroll`

A new View that represents the "inside part" of a scroll bar, minus the arrow buttons

It should look like the existing `ScrollBarView` (but better).

<img width="281" alt="image" src="https://github.com/gui-cs/Terminal.Gui/assets/585482/40489eb1-a7c3-49ee-a2d6-6be3c9891235">

and

<img width="6" alt="image" src="https://github.com/gui-cs/Terminal.Gui/assets/585482/2b9cb0b2-2e6e-482d-8ff0-6d352cc08090">

It should have the following fields/methods/events (roughly):

* `public Orientation Orientation {get ;set;}` - Determines the Orientation of the scroll.
* `public int Size { get;set;}` - The size of content the scroll represents.
* `public int Position { get; set;} - The position, relative to `Size`, to set the scrollbar at.
* `public event PositionChanging`
* `public event PositionChanged`
* `public event SizeChanged`

It'd be cool (but not required) for all the glyphs and styling to be configurable; similar to how `Slider.SliderStyle` works.

The names/types of these should match `Slider` as much as is reasonable for consistency. 

It must not have any internal knowledge of `ScrollBar` or the other views `ScrollBar` contains.

## `ScrollBar` 

A new View that is a composition of 3 subviews

- `public Button Decrease { get; set; }` - Has `WantContinousButtonPressed` set to true
- `public Button Increase { get; set; }` - Has `WantContinousButtonPressed` set to true
- `public Scroll Scroll { get; set; }` 

Exposes `public Orientation Orientation { get; set; }`. This is used to layout the 3 views correctly depending on orientation. Obvioulsy it also needs to change the glyph each of the two buttons have to be correct for orientation too.

This will be just a regular View that can be added as a subview to another view. It will not need any of the `Host` complexity that `ScrollBarView` has.

Exposes:

* `public Orientation Orientation {get ;set;}` - Determines the Orientation of the scroll. A passthrough to  `scrollBar.Scroll.Orientation`.
[...]

Issue #2488, opened 2023-04-04
Refactor `MenuBar` and `StatusBar` to be adornments


tig:
This is a sub-issue of:

- #2994 

## Todo

- [ ] Reimplemen these views based on `Bar`. See https://github.com/gui-cs/Terminal.Gui/issues/3073
- [ ] Remove the dependency between these classes and `Toplevel` and enable any view to host a menubar or statusbar (or any View).
- [ ] By default `Window` will have the menubar and statusbar's added to  it's `Border/Adornment` `Frame` but they will be `Visible = false` by default. We'll expose helper properties to access them:

`win.MenuBar.Visible = true` etc...

tig:
Not entirely sure this is needed.

BDisp:
> Not entirely sure this is needed.

I don't think this make senses because any view has adornments and the question must be  should `MenuBar` and `StatusBar` be created in the parent adornments? Now it's possible to add it anywhere the user want it, right? If affirmative then this can be closed.

tig:
I had two reasons for suggesting this:

1) Views can't render outside of their Viewport. That's a new LAW for v2. But the same law does not apply to Adornments because they are system managed. Menu, in particular, has the need to render outside of the Viewport because a dropdown may be larger than the view. 

2) The `Toplevel` code that tries to be "smart" about Menu/Statusbar is clumsy and bug ridden. I was thinking putting them in Padding might enable this to be simplified.

I still think 1) is semi-valid. I don't think 2) is valid anymore.

I'm going to close this for now though as there's not enough justification.

PR #2980, opened 2023-11-11
Fixes #2882. TabView: 'Frame.DrawFrame(Rect, bool)' is obsolete: 'This method is obsolete in v2. Use use LineCanvas or Frame

### LINKED ISSUE #2882: TabView: 'Frame.DrawFrame(Rect, bool)' is obsolete: 'This method is obsolete in v2. Use use LineCanvas or Frame
image

BDisp:
Fixes #2882 - Used `LineCanvas`.

## Pull Request checklist:


Terminal.Gui/Views/TabView.cs
```diff
@@ -187,9 +187,12 @@ public override void OnDrawContent (Rect contentArea)
 				// How much space do we need to leave at the bottom to show the tabs
 				int spaceAtBottom = Math.Max (0, GetTabHeight (false) - 1);
 				int startAtY = Math.Max (0, GetTabHeight (true) - 1);
+				var lc = new LineCanvas () { Lines = new List<StraightLine> (LineCanvas.Lines) };
```
tznind, line 191: This looks hacky.

The issue is that `DrawFrame` modifies `LineCanvas` and we want to prevent that behaviour?  If this is the case, then adding a `bool` parameter to `DrawFrame` allowing use a fresh `LineCanvas` would be better approach.

```csharp
/// <summary>
/// Draw a frame based on the passed bounds to the screen relative.
/// </summary>
/// <param name="bounds">The bounds view relative.</param>
/// <param name="lineStyle">The line style.</param>
/// <param name="attribute">The color to use.</param>
/// <param name="mergeWithLineCanvas">When drawing the frame, allow it to integrate (join) to other frames in other controls.
/// Or false to simply draw the rect exactly with no side effects.</param>
public void DrawFrame (Rect bounds, LineStyle lineStyle, Attribute? attribute = null, bool mergeWithLineCanvas=true)
```

tig:
When I was designing `Frames` I had the idea that `View`'s `Margin`, `Border`, and `Padding` could be replaced by a View. I didn't complete the implementation of this, but did make `View.CreateFrames` `virtual`:

```cs
/// <summary>
/// Creates the view's <see cref="Frame"/> objects. This internal method is overridden by Frame to do nothing
/// to prevent recursion during View construction.
/// </summary>
internal virtual void CreateFrames ()
{
	void ThicknessChangedHandler (object sender, EventArgs e)
	{
		LayoutFrames ();
		SetNeedsLayout ();
		SetNeedsDisplay ();
	}

	if (Margin != null) {
		Margin.ThicknessChanged -= ThicknessChangedHandler;
		Margin.Dispose ();
	}
	Margin = new Frame () { Id = "Margin", Thickness = new Thickness (0) };
	Margin.ThicknessChanged += ThicknessChangedHandler;
	Margin.Parent = this;

	if (Border != null) {
		Border.ThicknessChanged -= ThicknessChangedHandler;
		Border.Dispose ();
	}
	Border = new Frame () { Id = "Border", Thickness = new Thickness (0) };
	Border.ThicknessChanged += ThicknessChangedHandler;
	Border.Parent = this;

	// TODO: Create View.AddAdornment

	if (Padding != null) {
		Padding.ThicknessChanged -= ThicknessChangedHandler;
		Padding.Dispose ();
	}
	Padding = new Frame () { Id = "Padding", Thickness = new Thickness (0) };
	Padding.ThicknessChanged += ThicknessChangedHandler;
	Padding.Parent = this;
}
```

In addition, I left unfinished how the Frames are drawn. I used a hack of setting `Frame.Id` to `"Border"` etc...:

```cs
// TODO: v2 - this will eventually be two controls: "BorderView" and "Label" (for the title)

if (Id == "Border" && canDrawBorder && Thickness.Top > 0 && maxTitleWidth > 0 && !string.IsNullOrEmpty (Parent?.Title)) {
	var prevAttr = Driver.GetAttribute ();
	if (ColorScheme != null) {
		Driver.SetAttribute (HasFocus ? GetHotNormalColor () : GetNormalColor ());
	} else {
		Driver.SetAttribute (Parent.HasFocus ? Parent.GetHotNormalColor () : Parent.GetNormalColor ());
	}
	DrawTitle (new Rect (borderBounds.X, titleY, maxTitleWidth, 1), Parent?.Title);
	Driver.SetAttribute (prevAttr);
}

```

If we were to complete this idea ("Adornments"), we would have 3 built-in derivations of `Frame`: 

- `MarginView`
- `BorderView`
- `PaddingView`

Then, a view like `TabView` could replace any of these. In TabView's case `TabRowView` would derive from `Frame` (and be named `TabFrame`).

I have an open Issue for this already: https://github.com/gui-cs/Terminal.Gui/issues/2563

Note that in it, I call out that the problem of how Mouse/Keyboard events are handled correctly. In the current v2_develop, `Frame` can never get focus. But for all this to work, we have to change that. 

Note that by doing this it we also can remove `TabView.ContentView` which is a hack that was required in v1... and one of the motivations I had for implementing `Frames`. 

This same solution, BTW, will also simplify `Dialog` and `Wizard` (the button/nav bar just becomes a set of controls in a custom `Border`.  

@BDisp, what would you think of attacking solving all of this such that `TabView` is a great example of how to replace `View.Border` with an advanced implementation that provides all of the Tab navigation etc...?

Related (you'd be taking on enabling all of these Issues to be addressed):
- https://github.com/gui-cs/Terminal.Gui/issues/2563
- https://github.com/gui-cs/Terminal.Gui/issues/2488
- https://github.com/gui-cs/Terminal.Gui/issues/2489
- https://github.com/gui-cs/Terminal.Gui/issues/2407
- https://github.com/gui-cs/Terminal.Gui/issues/2814

What do you think???

tig:
Extra Credit:

Figure out the best way of satisfying this requirement:

- A View where `Border` (or `Margin` or `Padding`) has been overridden, can still have a "standard" border with a Title, frame, and (eventually standard things like close buttons (https://github.com/gui-cs/Terminal.Gui/issues/2563).

There are two potential solutions I can think of:

1) Simply force implementers to wrap such a view within a `View`
2) Allow implementers to add ADDITIONAL adornments (Frames) 

I'm not sure of the best approach, but I suspect #1 is simpler and more obvious.

BDisp:
I think that is a great challenge that must be done by phases. My doubt is if it's possible to intercept all the `LineCanvas` from all them, `Margin`, `Border` and `Padding`.
After I made the `TabWindow` unit test I realized that my hack worked because the `TabView` is clearing part of the top frame bellow of the tab and that why the intersection work. Due to this I'm having an idea to create a new method `DrawIncompleteFrame`.
Thus what is used today may be superseded for other solutions later and we need to walk with a step on a time. But it's excellent ideas that you have.

tig:
> I think that is a great challenge that must be done by phases. My doubt is if it's possible to intercept all the `LineCanvas` from all them, `Margin`, `Border` and `Padding`. After I made the `TabWindow` unit test I realized that my hack worked because the `TabView` is clearing part of the top frame bellow of the tab and that why the intersection work. Due to this I'm having an idea to create a new method `DrawIncompleteFrame`. Thus what is used today may be superseded for other solutions later and we need to walk with a step on a time. But it's excellent ideas that you have.

Not entirely sure what you mean, but I'm glad you think this is the right direction. Hopefully it sounds fun too! :-).

Let's continue this convo in #2994

BDisp:
> Not entirely sure what you mean, but I'm glad you think this is the right direction. Hopefully it sounds fun too! :-).
> 
I' talking about this commit https://github.com/gui-cs/Terminal.Gui/pull/2980/commits/60a6ead9cdf844359f100a7e55e53c183da8f366 and have a new unit test that prove what I said before. I'll submit latter.

```cs
		[Fact, SetupFakeDriver]
		public void TabWindow_With_Border ()
		{
			var canvas = new LineCanvas ();

			// Outer frame
			canvas.AddLine (new Point (0, 0), 10, Orientation.Horizontal, LineStyle.Single);
			canvas.AddLine (new Point (9, 0), 6, Orientation.Vertical, LineStyle.Single);
			canvas.AddLine (new Point (9, 5), -10, Orientation.Horizontal, LineStyle.Single);
			canvas.AddLine (new Point (0, 5), -6, Orientation.Vertical, LineStyle.Single);

			// Tab frame
			canvas.AddLine (new Point (1, 2), -2, Orientation.Vertical, LineStyle.Single);
			canvas.AddLine (new Point (1, 1), 4, Orientation.Horizontal, LineStyle.Single);
			canvas.AddLine (new Point (4, 1), 2, Orientation.Vertical, LineStyle.Single);

			// Inner frame
			canvas.AddLine (new Point (1, 2), 8, Orientation.Horizontal, LineStyle.Single);
			canvas.AddLine (new Point (8, 2), 3, Orientation.Vertical, LineStyle.Single);
			canvas.AddLine (new Point (8, 4), -8, Orientation.Horizontal, LineStyle.Single);
			canvas.AddLine (new Point (1, 4), -3, Orientation.Vertical, LineStyle.Single);

			string looksLike = @"
┌────────┐
│┌──┐    │
│├──┴───┐│
││      ││
│└──────┘│
└────────┘";
			TestHelpers.AssertEqual (output, looksLike, $"{Environment.NewLine}{canvas}");
		}

```
> Let's continue this convo in #2994

Ok. But if I fix this without the `LineCanvas` changes I made, then we need to consider this one and for later changes the #2994. You agree?

tig:
> > Not entirely sure what you mean, but I'm glad you think this is the right direction. Hopefully it sounds fun too! :-).
> 
> I' talking about this commit [60a6ead](https://github.com/gui-cs/Terminal.Gui/commit/60a6ead9cdf844359f100a7e55e53c183da8f366) and have a new unit test that prove what I said before. I'll submit latter.
> 
> ```cs
> 		[Fact, SetupFakeDriver]
> 		public void TabWindow_With_Border ()
> 		{
> 			var canvas = new LineCanvas ();
> 
> 			// Outer frame
> 			canvas.AddLine (new Point (0, 0), 10, Orientation.Horizontal, LineStyle.Single);
> 			canvas.AddLine (new Point (9, 0), 6, Orientation.Vertical, LineStyle.Single);
> 			canvas.AddLine (new Point (9, 5), -10, Orientation.Horizontal, LineStyle.Single);
> 			canvas.AddLine (new Point (0, 5), -6, Orientation.Vertical, LineStyle.Single);
> 
> 			// Tab frame
> 			canvas.AddLine (new Point (1, 2), -2, Orientation.Vertical, LineStyle.Single);
> 			canvas.AddLine (new Point (1, 1), 4, Orientation.Horizontal, LineStyle.Single);
> 			canvas.AddLine (new Point (4, 1), 2, Orientation.Vertical, LineStyle.Single);
> 
> 			// Inner frame
> 			canvas.AddLine (new Point (1, 2), 8, Orientation.Horizontal, LineStyle.Single);
> 			canvas.AddLine (new Point (8, 2), 3, Orientation.Vertical, LineStyle.Single);
> 			canvas.AddLine (new Point (8, 4), -8, Orientation.Horizontal, LineStyle.Single);
> 			canvas.AddLine (new Point (1, 4), -3, Orientation.Vertical, LineStyle.Single);
> 
> 			string looksLike = @"
> ┌────────┐
> │┌──┐    │
> │├──┴───┐│
> ││      ││
> │└──────┘│
> └────────┘";
> 			TestHelpers.AssertEqual (output, looksLike, $"{Environment.NewLine}{canvas}");
> 		}
> ```
> 
> > Let's continue this convo in #2994
> 
> Ok. But if I fix this without the `LineCanvas` changes I made, then we need to consider this one and for later changes the #2994. You agree?

I vote we abandon THIS PR (#2882) for now, until more of #2994 is addressed.

BDisp:
@tznind I'm changing the `Tab` class as `View` and leverage the v2 `Border` by handling the border thickness to get the desired output. Will be possible to merge all the borders of the tabs?

image

tig:
> @tznind I'm changing the `Tab` class as `View` and leverage the v2 `Border` by handling the border thickness to get the desired output. Will be possible to merge all the borders of the tabs?
> 
> image

Consider leveraging the (current prototype form, IMO) ability to make `Border.Thickness = new Thickness(1, 2, 1, 1)` cause the "Title" to be drawn, effectilvy as a  tab:

image

tig:
Looks really nice IMO 'Rounded`
image

tig:
What if...

`Border.TitleStyle = TitleStyle.Tab`

made the top Thickness be 3 and made it look like this:

image

BDisp:
Where is `TitleStyle`?

BDisp:
How about this one?

https://github.com/gui-cs/Terminal.Gui/assets/13117724/fde097ff-5994-423b-b2fc-f641d7e1ee44

tznind:
> handling the border thickness

I think `LineCanvas` supports most line styles, you can check with LineDrawing scenario:

shot-2023-11-21_22-54-25

One thing to note is that for lines to be joined they have to *overlap* not just be adjacent to one another.

These tab styles you guys are coming up with look amazing!  I see some look 2 high and others 3 high.  Maybe we can add a Style setting to let user pick between the various options? No need for 'one size fits all'

Only thing I think we should try to preserve is that there should always be a gap (not a line) under the selected tab title so that the tab content 'flows' naturally up into the tab title area.

BDisp:
I don't know if I +'m capable to do that but any help is welcome and feel free to send PR. I'll revert all the `DrawFrame` in this PR.

tig:
> Where is `TitleStyle`?

I made it up just to express my idea.

BDisp:
> I made it up just to express my idea.

Ah ok , thanks.

I need left/top/right/bottom tees for rounded join. Do you know they exists?

tig:
> > I made it up just to express my idea.
> 
> Ah ok , thanks.
> 
> I need left/top/right/bottom tees for rounded join. Do you know they exists?

I don't think so. I think you'll have to use the regular square tees.

tznind:
T shapes don't come in rounded style :)

BDisp:
I pretend to use borders on all tabs, because there is no horizontal space wasting, the space is always the same. The advantage is all them are separated by borders and thus there is no doubt if it's another tab or a simples space on the tab name.
The show top line will still keeping because is one more line a user can leverage for smaller terminal.
Is there any problems with this for you guys?

BDisp:
I know that we can avoid any border side to being draw by setting his thickness to 0. Also I know that a view frame will be the same size no matter it has border or not. But the bounds size change when the thickness is changed. This means that if we have a text in a view by default it will draw at top. If we set the top border thickness to 0 the text will be draw above the current line.
For e.g. a frame with `(X=0,Y=0,Width=6,Height=3)` with a border thickness of 1 result in a bound of `(X=0,Y=0,Width=4,Height=1)`. If we set the top border thickness to 0, the frame will be the same but the bounds will be `(X=0,Y=0,Width=4,Height=2)`. So the height increase and we have to manage to get the desired behavior.
Well for a border manipulation it's a little heavy to control this behaviors, although we can change the text vertical alignment.
My suggestion if it's justified to create bool properties `DrawLeftBorder`, `DrawTopBorder`, `DrawRightBorder` and `DrawBottomBorder` all as true by default. Using these properties will allow that the border thickness still are the same and will maintain the bounds unchanged. If some the properties are false that side will not be drawn.
@tig, @tznind what do you think of this?

BDisp:
Another way to have always the same bounds sizes when changing the border thickness is manipulating the `Margin` or the `Padding`. Which is preferable?

tig:
> I know that we can avoid any border side to being draw by setting his thickness to 0. Also I know that a view frame will be the same size no matter it has border or not. But the bounds size change when the thickness is changed. This means that if we have a text in a view by default it will draw at top. If we set the top border thickness to 0 the text will be draw above the current line. For e.g. a frame with `(X=0,Y=0,Width=6,Height=3)` with a border thickness of 1 result in a bound of `(X=0,Y=0,Width=4,Height=1)`. If we set the top border thickness to 0, the frame will be the same but the bounds will be `(X=0,Y=0,Width=4,Height=2)`. So the height increase and we have to manage to get the desired behavior. Well for a border manipulation it's a little heavy to control this behaviors, although we can change the text vertical alignment. My suggestion if it's justified to create bool properties `DrawLeftBorder`, `DrawTopBorder`, `DrawRightBorder` and `DrawBottomBorder` all as true by default. Using these properties will allow that the border thickness still are the same and will maintain the bounds unchanged. If some the properties are false that side will not be drawn. @tig, @tznind what do you think of this?

I do not like this idea (if I understand you correctly, which I'm not sure I do). 

If the Frames change, `Bounds` will be changed and `LayoutSubviews` will be called. This should cause the view to be cleared.* Then `OnDrawContent` will be called with a new `Bounds` (which will still always have `Bounds.Location == 0,0`).** I don't understand what the problem with this is.

*The current code is buggy/incorrect in this regard. Right now `view.Clear` is called at the top of `View.OnDrawContent` anytime `NeedsDisplay` is set. See the "TODO: " in Application.cs at line 551.

**This seems silly to me. Why are we using a `Rect` for something where the `Location` will ALWAYS be (0,0). I've considered changing `Bounds` to be a `Point` that describes the size of the `Bounds`. The reason I've not pulled the trigger on this, is I have intuition (which may be wrong) that to address https://github.com/gui-cs/Terminal.Gui/issues/2489 we may want to rename `Bounds` to `ContentArea` and enable `Location` to be negative when there's content to be scrolled.

BDisp:
I was referring about maintaining the Border side thickness grater than zero but decide if any of the side border can be drawn or not.
Imagine a view with a border and you don't want to draw the top border. The result will be the follow. I know we can manipulating this, but if it's only to not draw the top border it's more easy to have a flag to not draw it. It's more or less the `int DrawFrameLine` of the v1 and the suggestion is for using a flag for each border side. But if you don't like it, no problem.

Result in this:

```
┌──────┐      │ Test │
│ Test │      └──────┘
└──────┘
```

And we want this:

```
┌──────┐      
│ Test │      │ Test │
└──────┘      └──────┘
```

BDisp:
> If the Frames change, `Bounds` will be changed and `LayoutSubviews` will be called. This should cause the view to be cleared.* Then `OnDrawContent` will be called with a new `Bounds` (which will still always have `Bounds.Location == 0,0`).** I don't understand what the problem with this is.
> 

> *The current code is buggy/incorrect in this regard. Right now `view.Clear` is called at the top of `View.OnDrawContent` anytime `NeedsDisplay` is set. See the "TODO: " in Application.cs at line 551.
> 
> **This seems silly to me. Why are we using a `Rect` for something where the `Location` will ALWAYS be (0,0). I've considered changing `Bounds` to be a `Point` that describes the size of the `Bounds`. The reason I've not pulled the trigger on this, is I have intuition (which may be wrong) that to address #2489 we may want to rename `Bounds` to `ContentArea` and enable `Location` to be negative when there's content to be scrolled.

I think that makes more sense to considering to be a `Size` because is related with `Width` and `Height`. But since it can have negative location, which I think it will ended to have, then it's better to maintain  `Rect Bounds` and rename `contentArea` parameter to bounds, like it was early on the v1.

tig:
I'm meant Size not Point. Oops.

tig:
> Result in this:
> 
> 
> 
```
> 
> ┌──────┐      │ Test │
> 
> │ Test │      └──────┘
> 
> └──────┘
> 
> ```
> 
> 
> 
> And we want this:
> 
> 
> 
> ```
> 
> ┌──────┐      
> 
> │ Test │      │ Test │
> 
> └──────┘      └──────┘
> 
> 
```

No, we don't. 

What you show there is Margin.Thickness = new Thickness(0,1,0,0).

BDisp:
> No, we don't.
> 
> What you show there is Margin.Thickness = new Thickness(0,1,0,0).

That why I asked what should be the right action in this comment https://github.com/gui-cs/Terminal.Gui/pull/2980#issuecomment-1824967297.

tig:
I guess I didn't understand what you were asking. Sorry about that. Are we clear now?

BDisp:
> I guess I didn't understand what you were asking. Sorry about that. Are we clear now?

No problem. Yes we are clear now. For me it's better to using the margin, because it doesn't have any draw effect but managing thickness. Thanks.

BDisp:
Here is the new `TabView` looks like with full borders on all tabs. Was also added `PageDown` and `PageUp` support.

https://github.com/gui-cs/Terminal.Gui/assets/13117724/c69d2ee2-62e8-4a1e-a659-8f7188ae358b

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -437,7 +437,13 @@ Dim VerifyIsInitialized (Dim dim)
 	/// during set operations and in <see cref="LayoutSubviews"/>.If invalid settings are discovered exceptions will be thrown indicating the error.
 	/// This will impose a performance penalty and thus should only be used for debugging. 
 	/// </remarks>
-	public bool ValidatePosDim { get; set; }
+	public bool ValidatePosDim {
+		get => _validatePosDim;
+		set {
+			_validatePosDim = value;
+			ResizeView (AutoSize);
```
tig, line 444: I don't think turning on/off a diagnostic should have side-effects like calling ResizeView. Why do you think this is necessary here?

Terminal.Gui/Drawing/Thickness.cs
```diff
@@ -108,9 +108,9 @@ public Thickness (int left, int top, int right, int bottom)
 		/// Gets whether the specified coordinates lie within the thickness (inside the bounding rectangle but outside of
 		/// the rectangle described by <see cref="GetInside(Rect)"/>.
 		/// </summary>
-		/// <param name="outside">Describes the location and size of the rectangle that contains the thickness.</param>
-		/// <param name="x"></param>
-		/// <param name="y"></param>
+		/// <param name="outside">The outside rect.</param>
```
tig, line 111: What was wrong with my description? I believe it more fully describes the parameter.  Just saying "The outside rect" is less complete.

Terminal.Gui/View/Frame.cs
```diff
@@ -198,7 +198,9 @@ public override void OnDrawContent (Rect contentArea)
 
 			}
 
-			if (Id == "Border" && canDrawBorder && Thickness.Top > 0 && maxTitleWidth > 0 && !string.IsNullOrEmpty (Parent?.Title)) {
+			if (Id == "Border" && canDrawBorder && Thickness.Top > 0 && maxTitleWidth > 0 && !string.IsNullOrEmpty (Parent?.Title)
```
tig, line 201: I find it curious that you are continuing to add to this code, which I clearly stated was a prototype, instead of implementing Border, Padding, Margin correctly (each as a subclass of Frame)? 

All of this code (my old junk and your new tweaking of that junk) will need to be refactored once there are 3 Frame-based classes, and when Border contains a set of subviews that actually do the drawing vs. using LineCanvas directly.

tig (changes requested):

Terminal.Gui/Drawing/Thickness.cs
BDisp: That was due to the merge conflict, sorry.

Terminal.Gui/View/Layout/ViewLayout.cs
BDisp: If you comment `ResizeView (AutoSize);` the `AutoSize_False_SetWidthHeight_With_Dim_Fill_And_Dim_Absolute_With_Initialization` unit test will fail because `label.AutoSize = true;` and set `label.ValidatePosDim = false;` will resize the label for the text size.

Terminal.Gui/View/Frame.cs
BDisp: I only wanted to get a minimum border for the `TabView` here and was the solution I found.

BDisp:
This PR have a cherry pick of the #3019 because I needed some of the changes on the `View` initialization.

BDisp:
The `TestTreeViewColor` failing again. 😠

BDisp:
I won't inherit `TabRowView` from `Frame` nor remove the `View` container on this PR. So I think this is ready for review.

Terminal.Gui/Application.cs
```diff
@@ -1227,7 +1227,7 @@ public static void OnMouseEvent (MouseEventEventArgs a)
 			}
 		}
 
-		bool FrameHandledMouseEvent (Frame frame)
+		bool FrameHandledMouseEvent (Adornment frame)
```
tig: `AdornmentHandledMouseEvent`

BDisp: I think you didn't pull the last changes because I didn't changes to the `Application.cs`.

BDisp:
@tig this is ready for review. Thanks.

tig:
> @tig this is ready for review. Thanks.

Did you see the comments I already made?

This is looking really fantastic.

BDisp:
I have all comments as resolved. Please unresolved those you want I read. Thanks.

Terminal.Gui/Application.cs
BDisp: Ok I think you really want I do this change on this PR, right? Besides I think it should done on a separate PR but I'll changes in this PR.
Done.

BDisp:
I replaced `FrameHandledMouseEvent` to `AdornmentHandledMouseEvent`.

tig (changes requested):
Can you see them now?

Terminal.Gui/Views/TabView.cs
```diff
@@ -1,737 +1,929 @@
...
+
+/// <summary>
+/// Control that hosts multiple sub views, presenting a single one at once.
+/// </summary>
+public class TabView : View {
+	private Tab _selectedTab;
 
 	/// <summary>
-	/// Control that hosts multiple sub views, presenting a single one at once
+	/// The default <see cref="MaxTabTextWidth"/> to set on new <see cref="TabView"/> controls.
```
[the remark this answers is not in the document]
BDisp: Yes, the `DisplayText` is the real text and the `Text` is only set with the `MaxTabTextWidth` for draw.

Terminal.Gui/Views/Tab.cs
```diff
@@ -1,39 +1,42 @@
...
+public class Tab : View {
+	private string _displayText;
 
 	/// <summary>
-	/// The text to display in a <see cref="TabView"/>
+	/// The text to display in a <see cref="TabView"/>.
 	/// </summary>
 	/// <value></value>
-	public string Text { get => text ?? "Unamed"; set => text = value; }
+	public string DisplayText { get => _displayText ?? "Unamed"; set => _displayText = value; }
```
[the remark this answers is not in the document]
BDisp: I still need the `DisplayText` property to set the real text even it's bigger to display. I'll try to use `Title` instead of `Text` and see what is happen.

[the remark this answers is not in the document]
tig: You're not getting my point. 

But if you just want to move on, it's ok as this is something that can be refined later. What you've done with this so far is really great and what I'm asking for is probably overkill.

Feel free to resolve all the rest of my comments and I'll merge.

Separately, when I get a chance I'll do a write up further explaining myself.

Terminal.Gui/Views/Tab.cs
```diff
@@ -1,39 +1,42 @@
...
 	/// <summary>
-	/// Creates a new tab with the given text hosting a view
+	/// Creates a new tab with the given text hosting a view.
 	/// </summary>
-	/// <param name="text"></param>
-	/// <param name="view"></param>
-	public Tab (string text, View view)
+	/// <param name="displayText">The real text.</param>
+	/// <param name="view">The hosted view.</param>
+	public Tab (string displayText, View view)
```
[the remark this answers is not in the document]
BDisp: Well it already existed and I only changed the parameter name. But I already removed.

BDisp:
I'm sure you have good ideas to improve this more but for me I think it's finished and ready for your review.

tig (approved):

PR #2982, opened 2023-11-11
Fixes #2981. LegendAnnotation: 'Frame.DrawFrame(Rect, bool)' is obsolete: 'This method is obsolete in v2. Use use LineCanvas or Frame

### LINKED ISSUE #2981: LegendAnnotation: 'Frame.DrawFrame(Rect, bool)' is obsolete: 'This method is obsolete in v2. Use use LineCanvas or Frame
image

BDisp:
Fixes #2981 -Used `LineCanvas`.

## Pull Request checklist:


Terminal.Gui/Views/GraphView/Annotations.cs
```diff
@@ -122,7 +122,19 @@ public class LegendAnnotation : IAnnotation {
 		/// <see cref="Bounds"/> and so reduces the width/height
 		/// available for text by 2
 		/// </summary>
-		public bool Border { get; set; } = true;
+		public bool Border {
+			get => _border;
+			set {
+				_border = value;
+				if (_border) {
+					_legend.BorderStyle = LineStyle.Single;
```
tig, line 130: It was my intent that simply changing BorderStyle to something other than None would automatically set the thickness to 1. However, this may be suprising behavior (I'm a big fan of following the "principle of least suprise"). Thoughts?

BDisp: I agree if BorderStyle is other than None then at least set the thickness to 1, if it's 0 and also agree if is None set the thickness to 0. If None is wanted without thickness the the user must set it to 0.

Terminal.Gui/Views/GraphView/Annotations.cs
```diff
@@ -155,21 +170,25 @@ public LegendAnnotation (Rect legendBounds)
...
+				graph.Add (_legend);
 			}
 
-			// start the legend at
-			int y = Bounds.Top + (Border ? 1 : 0);
-			int x = Bounds.Left + (Border ? 1 : 0);
+			_legend.Frame = Bounds;
+
+			if (Border) {
+				_legend.OnDrawFrames ();
```
tig, line 181: Hold on a sec. In the other Issue you said you made the annotation legend a View. But you're not actually making it a SubView of GraphView? That seems really hacky. 

I'd rather you just use this.DrawFrame than jump through the hoops of haivng Legend be a View but not actually be used as a View.

tig (changes requested):

Terminal.Gui/Views/GraphView/Annotations.cs
```diff
@@ -155,21 +170,25 @@ public LegendAnnotation (Rect legendBounds)
 		/// <param name="graph"></param>
 		public void Render (GraphView graph)
 		{
-			if (Border) {
-				graph.Border.DrawFrame (Bounds, true);
+			if (!_legend.IsInitialized) {
+				_legend.ColorScheme = new ColorScheme () { Normal = Application.Driver.GetAttribute () };
+				graph.Add (_legend);
 			}
```
BDisp, line 176: @tig here is the adding the _legend as subview of the graph. If you don't want as view and you want as before I have no problem to change it. It only was to prove the border work for the _legend.

Terminal.Gui/Views/GraphView/Annotations.cs
BDisp: See https://github.com/gui-cs/Terminal.Gui/pull/2982#discussion_r1393439023.

tig (changes requested):

tig:
@tznind - Do you recall why `LegendAnnotations` is not a `View` subclass? Why have it hold it's own `Bounds`, `Border`, etc...?

Terminal.Gui/Views/GraphView/Annotations.cs
BDisp: Right. Deleted.

Terminal.Gui/Views/GraphView/Annotations.cs
BDisp: No, already done. Thanks for the addressed that.

BDisp:
> @tznind - Do you recall why `LegendAnnotations` is not a `View` subclass? Why have it hold it's own `Bounds`, `Border`, etc...?

@tznind I hope I didn't broken anything. @tig you are right, `Bounds` isn't needed. Wass enough only set the `Pos/Dim` based on the `legendBounds` parameter of the constructor.

tig:
> > @tznind - Do you recall why `LegendAnnotations` is not a `View` subclass? Why have it hold it's own `Bounds`, `Border`, etc...?
> 
> @tznind I hope I didn't broken anything. @tig you are right, `Bounds` isn't needed. Wass enough only set the `Pos/Dim` based on the `legendBounds` parameter of the constructor.

This is great progress, and looks like it will work.

I'm of two minds:
1) Leave it as is- good enough; addresses the issue of simplifying things by leveraging `Frame` to some extent.
2) IViewAnnotation.Render is a different pattern than IView.OnDraw.... Ideally the Legend object would be a subview of the GraphView and the "workaround" of explictly calling `DrawFrame/RenderLineCanvas` would not be needed because the natrual order of dawing would just do the right thing.

I'd like @tznind's perspective before we put any more time into this.

tznind:
Good point about view.  Annotations were separate because of the following things:

- could be drawn 'under' the graph.  I.e. the series lines of graph overwrite the annotation
- can be positioned in 'graph space' instead of 'screen space'
- annotations do not have to be square or invalidate their area

The second point could be solved with a new kind of Pos e.g  `AtGraphPosition`.

I think that legend being a view makes great sense.  But for example `PathAnnotation` wouldn't make sense as a `View`

BDisp:
@tig sorry but I want to see the graph and not only empty spaces around decreasing the graph size.

BDisp:
`Failed Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor [33 ms]`

tig:
> @tig sorry but I want to see the graph and not only empty spaces around decreasing the graph size.

Sorry, I don't understand.

BDisp:
> > @tig sorry but I want to see the graph and not only empty spaces around decreasing the graph size.
> 
> Sorry, I don't understand.

Because of all the thickness for margin+border+padding the available space was very smaller to show the graph.

tig:
> > > @tig sorry but I want to see the graph and not only empty spaces around decreasing the graph size.
> > 
> > 
> > Sorry, I don't understand.
> 
> Because of all the thickness for margin+border+padding the available space was very smaller to show the graph.

That's why I made the graph use Dim.Percent(70). It makes it much bigger.

tig (approved):

BDisp:
> That's why I made the graph use Dim.Percent(70). It makes it much bigger.

But it wasn't bigger enough. To see better I needed to maximize or increase the windows size.

tig:
> > That's why I made the graph use Dim.Percent(70). It makes it much bigger.
> 
> But it wasn't bigger enough. To see better I needed to maximize or increase the windows size.

You need a bigger monitor ;-).

In all seriousness, sorry about that. I'll fix it.

BDisp:
> You need a bigger monitor ;-).
> 
> In all seriousness, sorry about that. I'll fix it.

Don't worry I already fix and you already merged :-)

tig:
Ah, you actually broke things tho.

Because of this:

image

Should be `Checked = false`.

No biggie though.

tig:
Feel free to clean it up how you see fit.

BDisp:
Done here #3177.

PR #2984, opened 2023-11-12
Fixes #2983. View need a alternative DrawFrame for the v2.

### LINKED ISSUE #2983: View need a alternative DrawFrame for v2.
This is needed for situations where a Border frame isn't possible to use due some complexity scenario.

BDisp:
Fixes #2983 - Use `LineCanvas` to draw the frame and supersede the obsolete method.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/View/ViewDrawing.cs
+++ b/Terminal.Gui/View/ViewDrawing.cs
@@ -503,5 +503,25 @@ namespace Terminal.Gui {
 			DrawContentComplete?.Invoke (this, new DrawEventArgs (contentArea));
 		}
 
+		/// <summary>
+		/// Draw a frame based on the passed bounds to the screen relative.
+		/// </summary>
+		/// <param name="bounds">The bounds view relative.</param>
+		/// <param name="lineStyle">The line style.</param>
+		/// <param name="attribute">The color to use.</param>
+		public void DrawFrame (Rect bounds, LineStyle lineStyle, Attribute? attribute = null)
+		{
+			var vts = ViewToScreen (bounds);
+			LineCanvas.AddLine (new Point (vts.X, vts.Y), vts.Width,
+				Orientation.Horizontal, lineStyle, attribute);
+			LineCanvas.AddLine (new Point (vts.Right - 1, vts.Y), vts.Height,
+				Orientation.Vertical, lineStyle, attribute);
+			LineCanvas.AddLine (new Point (vts.X, vts.Bottom - 1), vts.Width,
+				Orientation.Horizontal, lineStyle, attribute);
+			LineCanvas.AddLine (new Point (vts.X, vts.Y), vts.Height,
+				Orientation.Vertical, lineStyle, attribute);
+
+			OnRenderLineCanvas ();
+		}
 	}
 }
--- a/UnitTests/View/DrawTests.cs
+++ b/UnitTests/View/DrawTests.cs
@@ -334,6 +334,23 @@ t     ", output);
 			Application.Refresh ();
 			TestHelpers.AssertDriverContentsWithFrameAre ("", output);
 		}
+
+		[Fact, AutoInitShutdown]
+		public void DrawFrame_Test ()
+		{
+			var label = new View () { X = Pos.Center (), Y = Pos.Center (), Text = "test", AutoSize = true };
+			var view = new View () { Width = 10, Height = 5 };
+			view.DrawContentComplete += (s, e) => view.DrawFrame (view.Bounds, LineStyle.Single);
+			view.Add (label);
+			Application.Top.Add (view);
+			Application.Begin (Application.Top);
+
+			TestHelpers.AssertDriverContentsWithFrameAre (@"
+┌────────┐
+│        │
+│  test  │
+│        │
+└────────┘", output);
+		}
 	}
 }
-


BDisp:
@tig the obsolete `Frame.DrawFrame(Rect, bool)` isn't using anymore and maybe can be removed?

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -503,5 +503,25 @@ public virtual void OnDrawContentComplete (Rect contentArea)
 			DrawContentComplete?.Invoke (this, new DrawEventArgs (contentArea));
 		}
 
+		/// <summary>
+		/// Draw a frame based on the passed bounds to the screen relative.
+		/// </summary>
+		/// <param name="bounds">The bounds view relative.</param>
```
tig, line 509: I think this param should be named `rect` and the docs should read "The view relative location and size of the frame."

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -503,5 +503,25 @@ public virtual void OnDrawContentComplete (Rect contentArea)
...
+		/// <param name="lineStyle">The line style.</param>
+		/// <param name="attribute">The color to use.</param>
```
tig, line 511: "The colors to be used."

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -503,5 +503,25 @@ public virtual void OnDrawContentComplete (Rect contentArea)
 			DrawContentComplete?.Invoke (this, new DrawEventArgs (contentArea));
 		}
 
+		/// <summary>
+		/// Draw a frame based on the passed bounds to the screen relative.
```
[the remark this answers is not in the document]
tig: Please add a <remark>:

"The frame will be merged (auto-joined) with any other lines drawn by this View using <see cref="LineCanvas"/>."

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -503,5 +503,25 @@ public virtual void OnDrawContentComplete (Rect contentArea)
...
+			LineCanvas.AddLine (new Point (vts.X, vts.Y), vts.Width,
+				Orientation.Horizontal, lineStyle, attribute);
+			LineCanvas.AddLine (new Point (vts.Right - 1, vts.Y), vts.Height,
+				Orientation.Vertical, lineStyle, attribute);
+			LineCanvas.AddLine (new Point (vts.X, vts.Bottom - 1), vts.Width,
+				Orientation.Horizontal, lineStyle, attribute);
+			LineCanvas.AddLine (new Point (vts.X, vts.Y), vts.Height,
+				Orientation.Vertical, lineStyle, attribute);
+
+			OnRenderLineCanvas ();
```
tig, line 524: I don't think this call is needed. If you replace it with "SetDisplayNeeded()" the LIneCanvas will be rendered in the natural order of things. RIght?

tig (changes requested):
See my comments and questions.

Terminal.Gui/View/ViewDrawing.cs
BDisp: It will have a `mergeWithLineCanvas` parameter as true by default. By default will be merged, otherwise will be rendered immediately.

```cs
			if (mergeWithLineCanvas) {
				LineCanvas.Merge (lc);
			} else {
				OnRenderLineCanvas ();
			}
```

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -503,5 +525,284 @@ public virtual void OnDrawContentComplete (Rect contentArea)
...
+		/// be less or equal to the start.
+		/// </summary>
+		/// <param name="startPos">The view relative of the start and side position.</param>
+		/// <param name="endPos">The view relative of the end and side position.</param>
+		/// <param name="rect">The view relative location and size of the frame.</param>
+		/// <param name="lineStyle">The line style.</param>
+		/// <param name="attribute">The colors to be used.</param>
+		/// <param name="mergeWithLineCanvas">When drawing the frame, allow it to integrate (join) to other frames in other controls.
+		/// Or false to simply draw the rect exactly with no side effects.</param>
+		public void DrawIncompleteFrame ((int start, Side side) startPos, (int end, Side side) endPos, Rect rect, LineStyle lineStyle, Attribute? attribute = null, bool mergeWithLineCanvas = true)
```
tig, line 576: @BDisp I do not understand what all this complex logic is for. Can you please describe the problem you are trying to solve with this?

BDisp: It's a helper method to draw incomplete frames based on start and end. I'm tying to implement it into the `TabView`. If you see the unit tests you'll understand the logic.

tznind:
I'm looking at `DrawIncompleteFrame` and thinking there might be a simpler way to achieve this and create a lot more flexibility for other things.  Here is my premise

Add a method which returns just the lines for the rect:

```csharp

public IEnumerable<StraightLine> GetFrame (Rect rect, LineStyle lineStyle, Attribute? attribute = null)
{
	var vts = ViewToScreen (rect);			
	yield return new StraightLine(new Point (vts.X, vts.Y), vts.Width,
		Orientation.Horizontal, lineStyle, attribute);
	yield return new StraightLine (new Point (vts.Right - 1, vts.Y), vts.Height,
		Orientation.Vertical, lineStyle, attribute);
	yield return new StraightLine (new Point (vts.Right - 1, vts.Bottom - 1), -vts.Width,
		Orientation.Horizontal, lineStyle, attribute);
	yield return new StraightLine (new Point (vts.X, vts.Bottom - 1), -vts.Height,
		Orientation.Vertical, lineStyle, attribute);
}
```

Add an extension method that operates on a collection of lines that lets you create an exclusion area:
```csharp
public static class StraightLineExtensions
{
/// <summary>
	/// Splits or removes all lines in the <paramref name="collection"/> such that none cover the given
	/// exclusion area.
	/// </summary>
	/// <param name="collection">Lines to adjust</param>
	/// <param name="start">First point to remove from collection</param>
	/// <param name="length">The number of sequential points to exclude</param>
	/// <param name="orientation">Orientation of the exclusion line</param>
	/// <returns></returns>
	public static IEnumerable<StraightLine> Exclude (this IEnumerable<StraightLine> collection, Point start, int length, Orientation orientation)
	{
		foreach(var l in collection) {

			// if line intersects the exclusion

			// if it is wholly inclusive of the line
				// remove it from returned collection
			// else if it is longer on both ends
				// split l into two lines
			// else
				// shorten right or left
		}

		return collection;
	}
}
```

What do you think? this would let you create rectangles with gaps in any sides needed and many other cool things besides?

I can probably work out the code for this method if its a go?

BDisp:
> What do you think? this would let you create rectangles with gaps in any sides needed and many other cool things besides?
> 
> I can probably work out the code for this method if its a go?

Any help is welcome. The `DrawIncompleteFrame` know where the side start and where the side end. It also allow to create rectangles with gaps in any sides needed by calling several times with different locations by using the same bounds. But any reduce of code is always better and more cleaning. So, the idea of exclusions is very good.
For now I have some `TabView` unit test failing.
Feel free to do what you think is better. Thanks.

tznind:
@BDisp I have created #3004 as an alternative approach to `DrawIncompleteFrame` which can hopefully simplify things.  Let me know what you think.

BDisp:
> @BDisp I have created #3004 as an alternative approach to `DrawIncompleteFrame` which can hopefully simplify things. Let me know what you think.

I like it very much. More elegant and with less code. Thanks.

In this PR I'm using a non 0 location with a value of 3 to avoid be divisible by 2 and so avoiding bugs on the unit tests. I never more will be using the 0 location for unit tests with frame.

BDisp:
Closing this because was supersede per the #3004.

PR #2998, opened 2023-11-16
v2 - Updates nuget dependencies

tig:


tig:
Test failure is due to upgrading xunit from 

`<PackageReference Include="xunit" Version="2.5.3" />`

to

`<PackageReference Include="xunit" Version="2.6.1" />`

Not sure why. @BDisp any ideas?

PR #3004, opened 2023-11-21
Line canvas exclude method

tznind:

This is a WIP.  Current implementation only works for parallel exclusion, perpendicular exclusion and more tests will be added before I mark this ready.

It is an alternative to the proposed `DrawIncompleteFrame`.

Use Case:
Ability to draw rectangles with parts of the boundary line snipped out e.g. for `TabView`.  Also in general, to apply an 'eraser' effect to a `LineCanvas`.

Approach:
Adds extension method `Exclude` to collections of `StraightLine` (this decouples the method from `LineCanvas`).

exclude

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drawing/LineCanvas.cs
+++ b/Terminal.Gui/Drawing/LineCanvas.cs
@@ -77,6 +77,15 @@ namespace Terminal.Gui {
 			ConfigurationManager.Applied += ConfigurationManager_Applied;
 		}
 
+		/// <summary>
+		/// Creates a new instance with the given <paramref name="lines"/>.
+		/// </summary>
+		/// <param name="lines">Initial lines for the canvas.</param>
+		public LineCanvas (IEnumerable<StraightLine> lines) : this()
+		{
+			_lines = lines.ToList();
+		}
+
 		private void ConfigurationManager_Applied (object? sender, ConfigurationManagerEventArgs e)
 		{
 			foreach (var irr in runeResolvers) {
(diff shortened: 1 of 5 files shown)


tig:
Wow. Just wow. 🤩

tznind:
Now works for perpendicular intersection too (cutting at most 1 cell out).

tznind:
One corner case bug currently is that the resulting lines are always of positive length.  

Now a horizontal line x=1 with length 5 is the same as one with x=5 with length -5

But not so for the special case of length = 1 vs length = -1 when joined to another line.  In this case a `UL` or `LL` is created depending on positivity of the length.

So I think I will need to ensure returned lines keep their +/- length.

See here when cutting horizontally just above the bottom line the bottom corners flip to pointing downwards (+1 length is down on y axis)

flip-down

BDisp:
> One corner case bug currently is that the resulting lines are always of positive length.
> 
No it can't, otherwise you can't control the flow.

> Now a horizontal line x=1 with length 5 is the same as one with x=5 with length -5
> 
That the right behavior.

> But not so for the special case of length = 1 vs length = -1 when joined to another line. In this case a `UL` or `LL` is created depending on positivity of the length.
> 
That was the problem I was facing before and I made confusion that corner could have more than one way to achieve them, but thankfully to you I learned that they must have unique values and the only exception is that no matter if vertical or horizontal are used first.

> So I think I will need to ensure returned lines keep their +/- length.
> 
Make sure it always return the signal value.

> See here when cutting horizontally just above the bottom line the bottom corners flip to pointing downwards (+1 length is down on y axis)

Yes I see, but I think you can manipulate that by getting the Y position where you cutting and the corner case you are manipulating. If the desired effect is the opposite return the right ones.

tznind:
Error is tree view colors unit test.  Should be good to go though.

PR #3024, opened 2023-11-30
Make Driver population code more concise

rccarlson:
This is effectively a NOP change, just makes it a little more terse and easier to read


rccarlson:
Based on the behavior on my machine, I believe this is some kind of no repro error and the commit should pass if tested again. Planning on doing another commit to poke it into trying again

BDisp:
Sometimes that unit test fail. We didn't find the cause yet.

rccarlson:
And it passes. Why do I write code? Why do I do this to myself? Do I enjoy the suffering? Do I just hate myself? It's like an abusive relationship that I keep coming back to. Anyway, PR is ready

Issue #2975, opened 2023-11-10
Modernize `Menu` - Base on `Bar` and address litany of issues


tig:
Related to: 
- #2926
- #3073
- #3852 

Todo: 

- [x] #2539
- [x] Update constructors to add `Command`s and `AddKeyBinding`s instead of using `shortcut` and `action`. 
- [x] Update `ProcessKey/OnKeyPressed` handlers
- [x] Do not allow HotKeys to invoke a menu item if it is not visible. Shotcuts yes, hotkeys no. https://github.com/gui-cs/Terminal.Gui/discussions/1940#discussioncomment-7724400
- [x] `Shortcut` should be `Key` based, not `KeyCode` - 
- [x] `Hotkey` should be `Key` based, not `Rune`
- [x] #4008 
- [x] #4011
- [ ] Update `Editor` Scenario to use `MenuBarv2`
- [ ] Update `Notepad` Scenario to use `MenuBarv2`
- [ ] Mark old `Menu` classes as obsolete. 
  - [ ] Rename to `LegacyMenu` etc... 
  - [ ] Rename `Menuv2` etc... to `Menu`
- [ ] Update `MenuBar` Scenario (or nuke)
- [ ] Update `DynamicMenuBar` Scenario (or nuke or replace)
- [ ] Delete old `Menu` classes

Others:

- [ ] #2404
- [ ] #2108

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/discussions/2974#discussioncomment-7536482_

tig:
Progress on this was made in #3073. Specifically:

- I developed `Shortcut` and `Bar` sufficiently that I'm confident they can be used as a basis.
- Formatting/Layout in `Shortcut` and `Bar` support the current look & feel of both `MenuBar` and `Menu`
- I introduced two temporary classes: `Menuv2` and `MenuBarv2` that are a starting point.

tznind:
If possible can you try to retain ability to modify a existing instance after construction.

This is important for designers [Menu Operations](https://github.com/gui-cs/TerminalGuiDesigner/tree/v2/src/Operations/MenuOperations)

tig:
> If possible can you try to retain ability to modify a existing instance after construction.
> 
> This is important for designers [Menu Operations](https://github.com/gui-cs/TerminalGuiDesigner/tree/v2/src/Operations/MenuOperations)

[...]

PR #3057, opened 2023-12-17
Fixes #3055: Remove `Key.Unknown`

### LINKED ISSUE #3055: Remove the `KeyCode.Unknown`
> > > I propose to remove the `KeyCode.Unknown` which aren't used by the drivers now, since they return `KeyCode.Null` if KeyChar is `'\0'`. Thus avoids using the prior to check if a key is valid or not and only compare if it's `KeyCode.Null`.
> > 
> > 
> > I agree.
> 
> Note that `Key` already doesn't have `Unknown`.

I think it should also been removed from the `KeyCode` enum.

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/2927#issuecomment-1858909499_

tig:
Fixes #3055

KeyCode.ToString now prints "Null" on Null.

Had to fix Menu.


tig:
@BDisp, this good for you?

BDisp:
Ok I see your are using `Key.Empty` to check if it's a valid rather than `KeyCode.Null`. Good work. Thanks.

BDisp:
> @BDisp, this good for you?

Yes thanks.

Terminal.Gui/Views/Menu/Menu.cs
```diff
@@ -432,7 +432,7 @@ void AddKeyBindings (MenuBarItem menuBarItem)
 		foreach (var menuItem in menuBarItem.Children.Where (m => m != null)) {
 			KeyBindings.Add ((KeyCode)menuItem.HotKey.Value, Command.ToggleExpandCollapse);
 			KeyBindings.Add ((KeyCode)menuItem.HotKey.Value | KeyCode.AltMask, Command.ToggleExpandCollapse);
-			if (menuItem.Shortcut != KeyCode.Unknown) {
+			if (menuItem.Shortcut != KeyCode.Null) {
```
BDisp, line 435: Here I've doubt if only a Shift, Alt or Ctrl pass.

BDisp: Ok but `Alt` is acceptable, right?

BDisp (approved):
It seems all good enough to me. Thanks.

Terminal.Gui/Views/Menu/Menu.cs
tig: Regarding a key event with ONLY shift modifiers.... We should ensure they never get sent by drivers.

In otherwords, nobody should REALLY have to call `Key.IsValid`. 

Thus, i think this is ok.

tig: Also, https://github.com/gui-cs/Terminal.Gui/issues/2975 will revamp all of this...

PR #3076, opened 2023-12-23
Fixes #3073: Adds `Bar` as a replacement for `StatusBar` & `Menu`

### LINKED ISSUE #191: Menu hints in StatusBar
To give an example how to use the messaging in a may existing StatusBar control: in theory, each menu item has a short message explaining what it will do, which would be displayed each time the menu item is hovered.

### LINKED ISSUE #2106: StatusBar should support right alignment on last StatusItem
I was playing with the TabView scenarios and hated how Notepad looked. Just so much wasted rows/columns ;-).

In the process, I decided to to move the "Len: " indicator to the status bar. But I really want it aligned right!

image

One way to fix this would be to support `StatusBar.ItemAlignment` property on `StatusBar` that works like `Dialog.ButtonAlignment`.

This would be trick, but wouldn't really get me what I want. The default would be `Left`.

Another idea would be to support an `StatusItem.Alignment` (Left, Center, Right..Justify doesn't make sense).

### LINKED ISSUE #3037: `View.GetKeyFromCommand` does not deal with multiple commands correctly
```cs
	[Fact]
	public void GetKeyFromCommand_MultipleCommands ()
	{
		var view = new View ();
		var commands = new List<Command> () {
			Command.Default,
			Command.Accept
		};
		view.AddKeyBinding (Key.A, commands.ToArray ());

		var key = view.GetKeyFromCommand (commands.ToArray ());
		Assert.Equal (Key.A, key);

		// BUGBUG: This fails and shouldn't
		key = view.GetKeyFromCommand (Command.Default);
		Assert.Equal (Key.A, key);

		// BUGBUG: This fails and shouldn't
		key = view.GetKeyFromCommand (Command.Accept);
		Assert.Equal (Key.A, key);
	}
```

I'll be adding this unit test with the tests commented out in #2927, but will not bother fixing this issue there.

### LINKED ISSUE #3073: Create a new `Bar` view to replace `StatusBar` (and eventually `MenuBar`)
Possible dupe of:

- #2975

I want to refactor StatusBar. It's actually a quite simple control, but was written before...

- View.Text
- View.HotKey with killer ToString support
- Frames (for StatusItem spacing/alignment)
- Consistency in eventing (KeyBindings and no Action). 
- KeyBindingScope.Application

It should just be a View that is Width = Fill and Y = AnchorEnd by default. 

The new StatusItem:

```cs
public class StatusItem : View {
    public override Text {
        get => $"KeyBindings.FirstOrDefault(b => b.Scope != KeyBindingScope.Focus).GetKeyFromCommands(Commands.Default));
}
```

... plus some stuff to auotmatically enable  / disable `Border.Thickness = new Thickness (0, 0, 1, 0)` for the separator.

`StatusBar.Add` is an override that sets `subview.X` appropriately (1st gets `X = 0`, subsequent get `X = Pos.View(prev)`; last gets `Width = Dim.Fill()`).

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3070#issuecomment-1868313288_

tig:
## Fixes 

- [ ] Fixes #3073
- [ ] Fixes #2106
- [ ] Fixes #191

## Todos

*Still very much a Proof of Concept/Prototype*

- [x] Implement `Shortcut` view that holds a `Key`, `KeyBindingScope`, `Title`, and help `Text`. 
  - [x] Renders just like a `MenuItem`.
    - [ ] Proper spacing/alignment of help text.
  - [x] Supports arbitrary `View`s in the `CommandView` (first) position. E.g. `CheckBox`
  - [x] Works with `KeyBindingScope.Application` and `KeyBindingScope.HotKey`.
  - [x] Supports mouse clicks to activate the shortcut
- [x] Implement `Bar` view that
  - [x] Looks just like the old `StatusBar` (`StatusBarStyle = true`, `Orientation.Horizontal`)
  - [ ] Looks just like the old `MenuBar` (`StatusBarStyle = false`, `Orientation.Horizontal`)
  - [ ] Looks just like the old `Menu` (`StatusBarStyle = false`, `Orientation.Vertcal`)
- [x] Update `UI Catalog` to use the new `Bar`
- [ ] Update to leverage `Dim.Auto``
- [ ] Update to leverage `Pos.Align`
- [ ] Replace `Toplevel.StatusBar`

DATQ7MI 1

## Pull Request checklist:


tig:
Now that `Dim.Auto` and `Pos.Align` are merged, I am going to get back to working on this...

BDisp:
> Now that `Dim.Auto` and `Pos.Align` are merged, I am going to get back to working on this...

I think you are heading in the right direction. Go with it :-)

dodexahedron:
I'm confused about KeyBindingScope, though.

That should be implicit, by the object graph.

Otherwise, it breaks some pretty basic assumptions most people make in OOP and especially in C#, and _especially_ when events are involved.

If there is to be such a property, it seems to me it should be defined recursively, so that a node in the graph farther from root never has a higher value than anything closer to root. That turns the object model on its head otherwise.

A simpler option would be for hotkeys to live at the root level, and just have a reference to the object that declared them.

tig:
> I'm confused about KeyBindingScope, though.
> 
> That should be implicit, by the object graph.
> 
> Otherwise, it breaks some pretty basic assumptions most people make in OOP and especially in C#, and _especially_ when events are involved.
> 
> If there is to be such a property, it seems to me it should be defined recursively, so that a node in the graph farther from root never has a higher value than anything closer to root. That turns the object model on its head otherwise.
> 
> A simpler option would be for hotkeys to live at the root level, and just have a reference to the object that declared them.

I find this suggestion hard to parse. First, `KeyBindingScope` was not added or changed in this PR.

Second, your wording is very general, and I struggle to see what you really mean. It would be great if you could explain yourself with some code examples (even pseudo-code). 

Note, I know the design isn't perfect and there's parts of it that worry me, but I have very intentionally exercised it with a bunch of real-world use cases in various new Views and Scenarios and it works great.

This particular PR *definitely* stresses it, which is part of my motivation to do this.

dodexahedron:
Oh certainly, I get that it's a targeted attempt at improving things (and it does).

I intentionally spoke generically, because the concept I intended to convey is a general concept, but it was also a question, though not really properly posed as one, to be fair.

In general, any object should be responsible for its own behaviors, or else a repository pattern is the appropriate alternative (where those objects are managed by a "repository" object that is responsible for telling everything what to do). That's a fundamentally different pattern from how most of TG is designed, though, so that's a non-starter anyway.

The easiest and most appropriate comparisons I can make are to other UI frameworks, since they all kinda set the de facto standards and expectations for developers.[^armDisruptorsAndFire]

So, I'll use WPF, WinForms, and JavaScript as usual.

In all of those, the best place to handle user interaction is as close to the element they are interacting with or that the action is relevant to as possible. The loosest is JS, where you can certainly attach handlers to the window or document if you want, but are strongly advised not to do so.

### The question
What is the aim/purpose of KeyBindingScope?\
I'll leave that question bare, so as not to influence the response, and expand to better answer your request once I have that context.

[^armDisruptorsAndFire]: While it's perfectly cool to be somewhat different/disruptive to conventions and such when we have a good case for doing so, being different for no solid reason is perhaps not so advisable.

tig:
Did you read the keybindingscope docs and keyboard conceptual docs?

dodexahedron:
> Did you read the keybindingscope docs and keyboard conceptual docs?

Not yet. I'll do so later tonight once I'm done with a work emergency, if it isn't terribly late.

dodexahedron:
Put another way, it seems like an over-definition, if that makes sense.

But maybe I'm not grokking something staring me in the face.

Which is the value of the back-and-forth, since I'm always a fan of RTFM regardless. :)

tig:
> It was basically just a clumsy use of the Socratic method to lead to what my thought is on it, with the potential for my thoughts to be supplemented by your input, as well.

totally fair. I was responding on my phone as I was getting in bed. So for me it was basically being lazy.

> The short version of my opinion on it is it's unnecessary, at least as implemented. The _concept_ and the _goal_ are great, but it seems a little bit heavy-handed and kinda unintuitive. Almost like a solution in search of a problem, but not really to that degree.

- ``Focused`` -."Bind this key to me so that when I have focus nobody but me sees the key, but the command is invoked on me when it's pressed". The default.'
- `HotKey`` - "Bound this key to me so that regardless of whether I have focus or not, as long as my superview heirchy has focus, I'll get the command invocation if pressed." - By default, automatcially bound to the hotkey and the `Command.Hotkey`` command, which is, effectively the "Default command" for a view.
- `Application` - "Bound this key to me so that I'll always get the command invocation if pressed." - Enables a view to register an application-wide shortcut key, like `StatusBar` could do in v1 in a very hacky way.

All three of these are very common, real, use-cases.

tig:
Still very confused about why this PR is not actually comparing to v2_develop.

image

@dodexahedron: PLEASE HELP!

dodexahedron:
> totally fair. I was responding on my phone as I was getting in bed. So for me it was basically being lazy.

Ha I feel ya there. I am guilty of that plenty of times myself.\
And I always read it the next day and cringe. 😅 

> * `Focused` -."Bind this key to me so that when I have focus nobody but me sees the key, but the command is invoked on me when it's pressed". The default.'
> * `HotKey` - "Bind this key to me so that regardless of whether I have focus or not, as long as my superview heirchy has focus, I'll get the command invocation if pressed." - By default, automatcially bound to the hotkey and the `Command.Hotkey` command, which is, effectively the "Default command" for a view.
> * `Application` - "Bind this key to me so that I'll always get the command invocation if pressed." - Enables a view to register an application-wide shortcut key, like `StatusBar` could do in v1 in a very hacky way.
> 
> All three of these are very common, real, use-cases.
> 
> It leverages the brilliant `KeyBinding` and `Command` infrastructure @tznind and @BDisp invented and radically simplifies View development. It now is very rare that a View will need to override `OnKey...` or subscribe to `KeyPress` events internally. It also enables configurability.

Right. That model is similar to RoutedCommand in WPF.

I think I need to sleep on this a bit more, too. Right now, I'm still getting the sense of over-definition or at least what feels like an unxpected inversion of control.

Here's what I would say, though, for sake of progress:

Even if we end up coming to some consensus on a change to make to it, the work to do so will be the same then as it would be right now. So, in my opinion, this particular thing isn't really a blocker for this PR necessarily - just an item to keep in mind for potential refinement. Heck, maybe even for a future version.

Though I also do feel like maybe the name of the type/property is part of what is nagging at me, maybe? Perhaps something mentioning Behavior instead of or in addition to scope might be more expressive? I dunno. Like I said, I would need to give it more thought. But right now I'd rather put those cycles into other work. :)

tig:
Closing. Replaced by #3531

PR #3089, opened 2023-12-25
Fixes #3071 & #3079. Key cast and static props are not correct 

### LINKED ISSUE #3071: Casting inconsistent between `Key` and `KeyCode`
**Describe the bug**

`Key` and `KeyCode` support equality e.g. 

```csharp
// Passes
Assert.Equal (KeyCode.A, Key.A);
Assert.Equal (KeyCode.F1, Key.F1);
```

However when casting is involved things behave inconsistently.  The most noticeable case is non character values e.g. F1

```csharp
// Fails
Assert.Equal ((uint)KeyCode.F1, (uint)Key.F1);
```

The int cast on `Key` is going into the 'char' explicit operator which returns 0

```
Expected: 1048587
Actual:   0
```

But there is also inconsistent behaviour when it comes to lower case letters

```csharp
// Pass
Assert.Equal (KeyCode.A, Key.A);
Assert.NotEqual (KeyCode.A, Key.A.WithShift);

// Fail
Assert.Equal ((uint)KeyCode.A, (uint)Key.A);
Assert.NotEqual ((uint)KeyCode.A, (uint)Key.A.WithShift);
```

I think the code is going down the `char` route in order to get to `uint` and I note that the docs describes this as 'lossy'.

```
	/// <summary>
	/// Explicitly cast <see cref="Key"/> to a <see langword="char"/>. The conversion is lossy. 
	/// </summary>
	/// <param name="kea"></param>
	public static explicit operator char (Key kea) => (char)kea.AsRune.Value;
```

**Expected behavior**
Casting to uint should either be impossible or should be consistent

### LINKED ISSUE #3079: Potentially unintended behavior of static pre-defined keys in Keys class (v2)
I am wondering if this behavior of the pre-defined static key fields in the Keys class is intentional or not:

While the fields themselves are readonly, they are **NOT** immutable - only the reference itself is - and a common instance reference is returned to all callers.

Say someone passed a key cloned from one of these via, for example, `Key.X` around in an event, since it's clearly expected, as it inherits from EventArgs. If they set any writable value in that key, it has been set globally for any future cases that static is used, which is probably not what a user would expect.

Currently, this example test would fail:

```
	[Fact]
	public void TheseAreNotSafe ( )
	{
		// Make two local keys, and grab Key.A, which is a reference to a singleton.
		Key aKey1 = Key.A;
		Key aKey2 = Key.A;

		// Assert the starting state that is expected
		Assert.False ( aKey1.Handled );
		Assert.False ( aKey2.Handled );
		Assert.False ( Key.A.Handled );

		// Now set Handled true on one of our local keys
		aKey1.Handled = true;

		// Assert the newly-expected case
		// The last two assertions will fail, because we have actually modified a singleton
		Assert.True ( aKey1.Handled );
		Assert.False ( aKey2.Handled );
		Assert.False ( Key.A.Handled );
	}

```

Changing all of the pre-defined fields to be auto get-only properties would ensure no reference equality between them, and still passes the existing tests.

That can be achieved by simply making every one of the `static readonly Key X = new()`  fields to `static Key X => new()`.

Alternatively, making sure no members can be set (including through side effects) is at least a workaround.

I think returning new instances by making them auto-get-only properties is most likely to have the behavior typically expected of that kind of thing.

### LINKED ISSUE #3081: Review compatibility with Microsoft.Extensions.Configuration
Key is a type that is likely to be in configuration files.

Microsoft.Extensions.Configuration is the standard configuration mechanism, these days.

The following code does not work:

```
class SomeConfig
{
  public Key SomeKey {get;set;}
}

[Fact]
public void RoundTripForConfiguration_SameAsOriginal
{
  SomeConfig c = new();
  c.SomeKey = Key.G.WithShift;
  File.WriteAllText( "configFile.json", JsonSerializer.Serialize( c ) );
  SomeConfig? roundTrippedConfig = new ConfigurationBuilder( ).AddJsonFile( "configFile.json" ).Build( ).Get<SomeConfig>( );
  Assert.Equal( c.SomeKey, roundTrippedConfig.SomeKey ); // Fails
}
```

This is because ConfigurationBuilder doesn't respect custom JsonConverters.

One solution is to make a Key ONLY serialize to a simple string value ("Shift+B" in the example code), rather than the whole object, and provide a string cast to Key that parses and returns Key. At least I'm fairly certain that would work, because I've used a similar solution for other projects.

Another is to implement a custom ConfigurationProvider for Microsoft.Extensions.Configuration. However, that is a lot more work and brings in an additional dependency.

The first way is even something that can be achieved without changing the custom serializer that's already there (KeyJsonConverter), since a user can write their own converter to serialize how they want. Such a custom serializer is as simple as this, on the consumer side:
```
public class AlternativeKeyJsonConverter : JsonConverter<Key>
{
    public override Key Read( ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options )
    {
        if ( Key.TryParse( reader.GetString( ), out Key key ) )
        {
            return key;
        }

        return Key.Empty;
    }

    public override void Write( Utf8JsonWriter writer, Key value, JsonSerializerOptions options )
    {
        writer.WriteStringValue( value.ToString( ) );
    }
}
```

But that still doesn't fix the incompatibility with configuration, because it's still going to be ignored, even if the user explicitly puts their own `[JsonConverter( typeof( AlternativeKeyJsonConverter ) )]` attribute on each Key property.

I am not sure if a specialized `[JsonConstructor]` would help, though I suspect it would not.

Now, while one could say "ok, then they can just `JsonSerializer.Deserialize` it!" well... no, they can't. Not easily, anyway. That forces the entire configuration to have to be treated that way, or else any configuration containing `Key`s would have to be kept separate, which is also non-ideal. Plus it would lose the other benefits of the configuration system, such as optional files, automatic configuration reloading, and layered configuration loading.

I think it's fair to say that Microsoft.Extensions.Configuration is not an edge case, so is worth considering, for any type that is likely to be part of a configuration (of any format).

tig:
Fixes #3071
Fixes #3079
Fixes #3081 

- [x] Adds `new Key (char)` constructor
- [x] Updates `(char)Key` cast to use char constructor
- [x] Adds unit tests
- [x] static standard key properties now always get new instance 
- [x] Adds Key constructor that takes a string
- [x] Changes KeyJsonSerializer to be far simpler (no object) and public. Removes it as an attribute on Key

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Input/Key.cs
+++ b/Terminal.Gui/Input/Key.cs
@@ -72,10 +72,19 @@ public class Key : EventArgs, IEquatable<Key> {
 	public Key () : this (KeyCode.Null) { }
 
 	/// <summary>
-	///   Constructs a new <see cref="Key"/> from the provided Key value
+	/// Constructs a new <see cref="Key"/> from the provided Key value
 	/// </summary>
 	/// <param name="k">The key</param>
 	public Key (KeyCode k) => KeyCode = k;
+
+	/// <summary>
+	/// Constructs a new <see cref="Key"/> from a char.
+	/// </summary>
+	/// <param name="ch"></param>
+	public Key (char ch)
+	{
+		KeyCode = (KeyCode)ch;
+	}
 
 	/// <summary>
 	/// Indicates if the current Key event has already been processed and the driver should stop notifying any other event subscriber.
@@ -280,11 +289,11 @@ public class Key : EventArgs, IEquatable<Key> {
 	public static implicit operator Key (KeyCode keyCode) => new (keyCode);
 
 
-	/// <summary>
-	/// Cast <see langword="char"/> to a <see cref="Key"/>. 
-	/// </summary>
-	/// <param name="ch"></param>
-	public static implicit operator Key (char ch) => new ((KeyCode)ch);
+	///// <summary>
+	///// Cast <see langword="char"/> to a <see cref="Key"/>. 
+	///// </summary>
+	///// <param name="ch"></param>
+	//public static implicit operator Key (char ch) => new ((KeyCode)ch);
 
 	/// <inheritdoc/>
 	public override bool Equals (object obj) => obj is Key k && k.KeyCode == KeyCode;
--- a/Terminal.Gui/Input/Key.cs
+++ b/Terminal.Gui/Input/Key.cs
@@ -109,6 +109,21 @@ public class Key : EventArgs, IEquatable<Key> {
 	}
 
 	/// <summary>
+	/// Constructs a new Key from a string describing the key.
+	/// See <see cref="TryParse(string, out Terminal.Gui.Key)"/> for information
+	/// on the format of the string.
+	/// </summary>
+	/// <param name="str">The string describing the key.</param>
+	public Key (string str)
+	{
+		var result = Key.TryParse (str, out Key key);
+		if (!result) {
+			throw new ArgumentException (@$"Invalid key string: {str}", nameof (str));
+		}
+		KeyCode = key.KeyCode;
+	}
+
+	/// <summary>
 	/// Indicates if the current Key event has already been processed and the driver should stop notifying any other event subscriber.
 	/// Its important to set this value to true specially when updating any View's layout from inside the subscriber method.
 	/// </summary>
--- a/Terminal.Gui/Input/Key.cs
+++ b/Terminal.Gui/Input/Key.cs
@@ -65,7 +65,6 @@ namespace Terminal.Gui;
 /// </list>
 /// </para>
 /// </remarks>
-[JsonConverter (typeof (KeyJsonConverter))]
 public class Key : EventArgs, IEquatable<Key> {
 	/// <summary>
 	/// Constructs a new <see cref="Key"/>
@@ -140,8 +139,6 @@ public class Key : EventArgs, IEquatable<Key> {
 	/// <remarks>
 	/// This property is the backing data for the <see cref="Key"/>. It is a <see cref="KeyCode"/> enum value.
 	/// </remarks>
-	[JsonInclude]
-	[JsonConverter (typeof (KeyCodeJsonConverter))]
 	public KeyCode KeyCode { get; init; }
 
 	/// <summary>
--- a/Terminal.Gui/Input/Key.cs
+++ b/Terminal.Gui/Input/Key.cs
@@ -4,6 +4,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Text;
 using System.Text.Json.Serialization;
+using static System.Runtime.CompilerServices.RuntimeHelpers;
 
 namespace Terminal.Gui;
 
@@ -80,10 +81,31 @@ public class Key : EventArgs, IEquatable<Key> {
 	/// <summary>
 	/// Constructs a new <see cref="Key"/> from a char.
 	/// </summary>
+	/// <remarks>
+	/// <para>
+	/// The key codes for the A..Z keys are encoded as values between 65 and 90 (<see cref="KeyCode.A"/> through <see cref="KeyCode.Z"/>).
+	/// While these are the same as the ASCII values for uppercase characters, they represent *keys*, not characters.
+	/// Therefore, this constructor will store 'A'..'Z' as <see cref="KeyCode.A"/>..<see cref="KeyCode.Z"/> with the
+	/// <see cref="KeyCode.ShiftMask"/> set and will
+	/// store `a`..`z` as <see cref="KeyCode.A"/>..<see cref="KeyCode.Z"/>.
+	/// </para>
+	/// </remarks>
 	/// <param name="ch"></param>
 	public Key (char ch)
 	{
-		KeyCode = (KeyCode)ch;
+		switch (ch) {
+		case >= 'A' and <= 'Z':
+			// Upper case A..Z mean "Shift-char" so we need to add Shift
+			KeyCode = (KeyCode)ch | KeyCode.ShiftMask;
+			break;
+		case >= 'a' and <= 'z':
+			// Lower case a..z mean no shift, so we need to store as Key.A...Key.Z
+			KeyCode = (KeyCode)(ch - 32);
+			return;
+		default:
+			KeyCode = (KeyCode)ch;
+			break;
+		}
 	}
 
 	/// <summary>
@@ -103,7 +125,8 @@ public class Key : EventArgs, IEquatable<Key> {
 	/// <remarks>
 	/// This property is the backing data for the <see cref="Key"/>. It is a <see cref="KeyCode"/> enum value.
 	/// </remarks>
-	[JsonInclude] [JsonConverter (typeof (KeyCodeJsonConverter))]
+	[JsonInclude]
+	[JsonConverter (typeof (KeyCodeJsonConverter))]
 	public KeyCode KeyCode { get; init; }
 
 	/// <summary>
@@ -115,8 +138,11 @@ public class Key : EventArgs, IEquatable<Key> {
 	/// The key value as a Rune. This is the actual value of the key pressed, and is independent of the modifiers.
 	/// </summary>
 	/// <remarks>
-	/// If the key pressed is a letter (a-z or A-Z), this will be the upper or lower case letter depending on whether the shift key is pressed.
-	/// If the key is outside of the <see cref="KeyCode.CharMask"/> range, this will be <see langword="default"/>.
+	/// <para>
+	/// If the key is a letter (a-z or A-Z), the Rune be the upper or lower case letter depending on whether
+	/// the shift key was pressed.
+	/// If the key is outside of the <see cref="KeyCode.CharMask"/> range, the returned Rune will be <see langword="default"/>.
+	/// </para>
 	/// </remarks>
 	public Rune AsRune => ToRune (KeyCode);
 
@@ -124,11 +150,14 @@ public class Key : EventArgs, IEquatable<Key> {
 	/// Converts a <see cref="KeyCode"/> to a <see cref="Rune"/>.
 	/// </summary>
 	/// <remarks>
-	/// If the key is a letter (a-z or A-Z), this will be the upper or lower case letter depending on whether the shift key is pressed.
-	/// If the key is outside of the <see cref="KeyCode.CharMask"/> range, this will be <see langword="default"/>.
+	/// <para>
+	/// If the key is a letter (a-z or A-Z), the Rune be the upper or lower case letter depending on whether
+	/// the shift key was pressed.
+	/// If the key is outside of the <see cref="KeyCode.CharMask"/> range, the returned Rune will be <see langword="default"/>.
+	/// </para>
 	/// </remarks>
 	/// <param name="key"></param>
-	/// <returns>The key converted to a rune. <see langword="default"/> if conversion is not possible.</returns>
+	/// <returns>The key converted to a Rune. <see langword="default"/> if conversion is not possible.</returns>
 	public static Rune ToRune (KeyCode key)
 	{
 		if (key is KeyCode.Null or KeyCode.SpecialMask || key.HasFlag (KeyCode.CtrlMask) || key.HasFlag (KeyCode.AltMask)) {
@@ -286,14 +315,16 @@ public class Key : EventArgs, IEquatable<Key> {
 	/// Cast <see cref="KeyCode"/> to a <see cref="Key"/>. 
 	/// </summary>
 	/// <param name="keyCode"></param>
-	public static implicit operator Key (KeyCode keyCode) => new (keyCode);
+	public static implicit operator Key (KeyCode keyCode) => new Key( keyCode);
 
-
-	///// <summary>
-	///// Cast <see langword="char"/> to a <see cref="Key"/>. 
-	///// </summary>
-	///// <param name="ch"></param>
-	//public static implicit operator Key (char ch) => new ((KeyCode)ch);
+	/// <summary>
+	/// Cast <see langword="char"/> to a <see cref="Key"/>. 
+	/// </summary>
+	/// <remarks>
+	/// See <see cref="Key(char)"/> for more information.
+	/// </remarks>
+	/// <param name="ch"></param>
+	public static implicit operator Key (char ch) => new Key (ch);
 
 	/// <inheritdoc/>
 	public override bool Equals (object obj) => obj is Key k && k.KeyCode == KeyCode;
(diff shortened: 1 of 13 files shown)


Terminal.Gui/Input/Key.cs
```diff
@@ -72,11 +73,41 @@ public class Key : EventArgs, IEquatable<Key> {
...
+			break;
+		case >= 'a' and <= 'z':
+			// Lower case a..z mean no shift, so we need to store as Key.A...Key.Z
+			KeyCode = (KeyCode)(ch - 32);
+			return;
+		default:
+			KeyCode = (KeyCode)ch;
+			break;
+		}
+	}
```
BDisp, line 109: @tig you can leverage this code https://github.com/BDisp/Terminal.Gui/blob/9349d542974c6eb59044b3ca4716cc1a25101321/Terminal.Gui/Input/Key.cs#L195-L253 to get the shift or not based on more range of characters to allow test this:

```cs
	[InlineData ('á', KeyCode.A)]
	[InlineData ('Á', (KeyCode)'Á' | KeyCode.ShiftMask)]

```

BDisp: Here is the full code https://github.com/BDisp/Terminal.Gui/blob/9349d542974c6eb59044b3ca4716cc1a25101321/Terminal.Gui/ConsoleDrivers/ConsoleKeyMapping.cs#L196-L294

tig: I don't think that is correct behavior. Is it  really needed for VK packet support or is it something you think is a nice-to-have?

If it's not absolutely needed for VK packet support then I'd prefer to have A..Z be the only keys supported in the test.

BDisp: Ah if it's only needed for the tests ok, but you still need this for `TextView` recognize that they are valid characters to pass on the `GetIsKeyCodeAtoZ` method to insert the text.

dodexahedron: An aside, as I've been looking at all these tests:

Might I suggest using MemberData to generate test cases instead of all the InlineData?

It's xUnit's analogue to NUnit's TestCaseSource and would allow you to have more exhaustive and authoritative tests with less work and less code.

The general pattern is:

```
[Theory]
[MemberData(nameof(TheoryA_Cases))]
public void TheoryA( string s, int i, object o )
{
  // Test code
}

private static IEnumerable<object[]> TheoryA_Cases()
{
  // Code that provides an IEnumerable of 3-element object arrays that satisfy the parameters of the test.
}
```

xUnit's documentation is sorely lacking, but that's the basic pattern for using that.

I do not know if xUnit has a way to generate tests from combinatorial parameter sources, but that effect can be achieved by just doing the combinatorial loops yourself in the case generator method.

dodexahedron: Someone did contribute an example for xUnit, of a way to provide yourself with fairly easily reusable combinatorial case generators:

https://github.com/xunit/samples.xunit/blob/main/TheoryDataExamples/MatrixTheoryData.cs

That is only for 2 parameters, but extending it to 3 or more is fairly straightforward.

BDisp: You need it because this unit test is wrong:

`[InlineData ('Ó', (KeyCode)'Ó')]`

Should be:

`[InlineData ('Ó', (KeyCode)'Ó' | KeyCode.ShiftMask)]`

Because Ó is a capital letter.

BDisp: Change to this and you'll see.

```cs
	public Key (char ch)
	{
		switch (ch) {
		case >= 'A' and <= 'Z':
			// Upper case A..Z mean "Shift-char" so we need to add Shift
			KeyCode = (KeyCode)ch | KeyCode.ShiftMask;
			break;
		case char when (ch - 32) is >= 'A' and <= 'Z':
			// Lower case a..z mean no shift, so we need to store as Key.A...Key.Z
			KeyCode = (KeyCode)(ch - 32);
			return;
		case >= 'À' and <= 'Ý':
			// Upper case À..Ý mean "Shift-char" so we need to add Shift
			KeyCode = (KeyCode)ch | KeyCode.ShiftMask;
			break;
		case char when (ch - 32) is >= 'à' and <= 'ý':
			// Lower case à..ý mean no shift, so we need to store as Key.A...Key.Z
			KeyCode = (KeyCode)(ch - 32);
			return;
		default:
			KeyCode = (KeyCode)ch;
			break;
		}
	}

```

dodexahedron:
What do you think of maybe also adding a constructor taking string that essentially operates as a Parse method would, and/or adding a Parse method (ie a normal/non-try version of TryParse)?

Terminal.Gui/Input/Key.cs
BDisp: @dodexahedron we also have a `ClassData` unit test here https://github.com/BDisp/Terminal.Gui/blob/9349d542974c6eb59044b3ca4716cc1a25101321/UnitTests/ConsoleDrivers/ConsoleKeyMappingTests.cs#L102.

tig:
💯 👆

Terminal.Gui/Input/Key.cs
BDisp: > I don't think that is correct behavior. Is it really needed for VK packet support or is it something you think is a nice-to-have?
> 
> If it's not absolutely needed for VK packet support then I'd prefer to have A..Z be the only keys supported in the test.

Remember that you created a `Key` class that rely if the shift key is enabled or not for a certain character. So it isn't a matter if it's only needed for VK packet or not. Is how the keystroke really work. If you want to pass the char '!' in a `Key` instance you'll need to know if the shift key will be assed or not. In this case the shift is needed because you'll need to press `Shift+D1` to obtain the '!'.

tig: > you still need this for `TextView` recognize that they are valid characters to pass on the `GetIsKeyCodeAtoZ` method to insert the text.

Why?

BDisp: Because I added the shift key to the '!' and others characters that only can be showing by pressing the shift key. But, like the 'A', it's needed to remove the shift key from the `Key` and only send the character '!', otherwise the key is handled as key binding `(Key)'!' | Key.Shift`.

tig: > > I don't think that is correct behavior. Is it really needed for VK packet support or is it something you think is a nice-to-have?
> > If it's not absolutely needed for VK packet support then I'd prefer to have A..Z be the only keys supported in the test.
> 
> Remember that you created a `Key` class that rely if the shift key is enabled or not for a certain character. So it isn't a matter if it's only needed for VK packet or not. Is how the keystroke really work. If you want to pass the char '!' in a `Key` instance you'll need to know if the shift key will be assed or not. In this case the shift is needed because you'll need to press `Shift+D1` to obtain the '!'.

I disagree. The only special case required is for A...Z. All other cases work fine. I don't see any scenario where an app developer needs to know that `!` was the result of KeyCode.D1 | KeyCode.ShiftMask. All they need to know is the user meant to type '!'. 

If you have a specific use-case that requires all of the added complexity you are suggesting please describe it. 

In the current implementation, if I press KeyCode.D7 | KeyCode.ShiftMask:

- ENG Keyboard: 
image

- DEU Keyboard:
image

This is the behavior I think apps want to see.

(Note: v2_devlelop somehow broke how shift/ctrl/alt key presses on Windows worked and are showing "Null". I will fix this asap).

dodexahedron: > @dodexahedron we also have a `ClassData` unit test here https://github.com/BDisp/Terminal.Gui/blob/9349d542974c6eb59044b3ca4716cc1a25101321/UnitTests/ConsoleDrivers/ConsoleKeyMappingTests.cs#L102.

Right

The point is it should be used for the majority of these tests. Spot checks miss things, and this is a crucial class that warrants positive and exhaustive proof. There are also several cases of duplicated test cases, which underlines that humans writing test case inputs one by one is a highly error-prone process.

dodexahedron: > > > I don't think that is correct behavior. Is it really needed for VK packet support or is it something you think is a nice-to-have?
> > > If it's not absolutely needed for VK packet support then I'd prefer to have A..Z be the only keys supported in the test.
> > 
> > 
> > Remember that you created a `Key` class that rely if the shift key is enabled or not for a certain character. So it isn't a matter if it's only needed for VK packet or not. Is how the keystroke really work. If you want to pass the char '!' in a `Key` instance you'll need to know if the shift key will be assed or not. In this case the shift is needed because you'll need to press `Shift+D1` to obtain the '!'.
> 
> I disagree. The only special case required is for A...Z. All other cases work fine. I don't see any scenario where an app developer needs to know that `!` was the result of KeyCode.D1 | KeyCode.ShiftMask. All they need to know is the user meant to type '!'.
> 
> If you have a specific use-case that requires all of the added complexity you are suggesting please describe it.
> 
> In the current implementation, if I press KeyCode.D7 | KeyCode.ShiftMask:
> 
> * ENG Keyboard:
> 
> image
> * DEU Keyboard:
> 
> image
> This is the behavior I think apps want to see.
> 
> (Note: v2_devlelop somehow broke how shift/ctrl/alt key presses on Windows worked and are showing "Null". I will fix this asap).

Key binds.

How do you know if a user pressed ctrl+shift+D1 or simply pasted the string "!" if you don't communicate the keys?

BDisp: Before the creating of the `Key` class the shift key alone never was added on a character because of this confusion and it also should not be added now for A..Z. This was what starting this confusion. The shift key only must be added to `Key` if it's accomplished with another modifiers but never alone. It only have to be noted in the property `IsShift` bool.

BDisp: The only exceptions where a shift key can be added alone to a `Key` is when a non-printable character is used, like F1..F12, cursor keys, delete, backspace, etc.

dodexahedron: Right and that's fine, because the information is still there. As long as you can round-trip it, you're golden.

dodexahedron: > The only exceptions where a shift key can be added alone to a `Key` is when a non-printable character is used, like F1..F12, cursor keys, delete, backspace, etc.

Makes perfect sense.

Theoretical question, though:

Say I have a string on the clipboard that has all characters my keyboard can produce - letters, digits, symbols, and whitespace - and I paste that string while holding ctrl.

Would characters such as the ! be interpreted as "ctrl+shift+D1"? Or would they come across as just D1? Or would they come across as ctrl+D1? 🤔

BDisp: In my PR I was creating as `(KeyCode)'!' | KeyCode.ShiftMask` to leverage the character. `ConsoleKeyInfo` parses it as `KeyChar='!', Key=ConsoleKey.D1, ConsoleModifiers.Shift`.
Pasting from the clipboard only will return the clipboard content and not the ctrl key and will return as `(KeyCode)'!' | KeyCode.ShiftMask`. In `TextField` you use the `Ctrl+V` to paste and in `TextView` you use the `Ctrl+Y`.
But now the problem adding only a shift key on a character other than A..Z, because the shift key is removed from the `Key` before insert the text in a `TextView` and not for the others characters. Because of this issue, `Terminal.Gui` never added the shift key alone on printable character. Now it's only possible to A..Z and thus is impossible normalize all.

dodexahedron:
I'm also finding, at least from the usage in TerminalGuiDesigner, as I'm adapting it to match the new Key type, that an implicit cast operator to and from string that use Parse and ToString() respectively would be a pretty nice thing to have, and would feel really natural in use. It would also make configuration a ton easier, since M.E.C. looks for string conversions for incoming data to the declared type.

Terminal.Gui/Input/Key.cs
tig: Tell me the user scenario that no longer works.

BDisp: > I don't think that is correct behavior. Is it really needed for VK packet support or is it something you think is a nice-to-have?
> 
With the current `Key` code is the correct behavior if the `GetIsKeyCodeAtoZ` is to persist always because the shift key does not is present only from A..Z. It's good to have, although the shift key must be removed before sending to a consumer and the `IsShift` indicates whether the shift key was pressed.

> If it's not absolutely needed for VK packet support then I'd prefer to have A..Z be the only keys supported in the test.

With the current `Key` code that only is supporting the shift key for A..Z, this is needed for the  VK packet support as well.
I already proved that `'Ó'` is `(Key)'Ó' | Key.Shift` but there are also others printable character that has the shift key, like `!"#$%&/()=?»*ª>;:_`. There are two options:
1. Not including only the shift key to any printable character as before which will not provide the `IsShift` property as `true`.
2. Including the shift key to any printable character which will provide the `IsShift` property as `true`. I prefer this one but if the first is preferable the the `GetIsKeyCodeAtoZ` must be removed and treat all as equal as before.

tig:
> I'm also finding, at least from the usage in TerminalGuiDesigner, as I'm adapting it to match the new Key type, that an implicit cast operator to and from string that use ToString() and Parse() respectively would be a pretty nice thing to have, and would feel really natural in use. It would also make configuration a ton easier, since M.E.C. looks for string conversions for incoming data to the declared type.

Done.

UnitTests/Input/KeyTests.cs
```diff
@@ -53,6 +53,59 @@ public void Constructor_Char (char ch, KeyCode expectedKeyCode)
...
+
+	// TryParse
+	[Theory]
+	[InlineData ("a", KeyCode.A)]
+	[InlineData ("Ctrl+A", KeyCode.A | KeyCode.CtrlMask)]
+	[InlineData ("Alt+A", KeyCode.A | KeyCode.AltMask)]
+	[InlineData ("Shift+A", KeyCode.A | KeyCode.ShiftMask)]
+	[InlineData ("A", KeyCode.A | KeyCode.ShiftMask)]
+	[InlineData ("â", (KeyCode)'â')]
+	[InlineData ("Shift+â", (KeyCode)'â' | KeyCode.ShiftMask)]
```
BDisp, line 65: This is wrong. 'â' is the lower case of 'Â'. So it haven't the shift key. Only the 'Â' has the shift key. Because of only consider the A..Z for the test the logic of the shift key is broken.

BDisp (changes requested):

tig:
@BDisp I'm going to merge this now, ahead of me reviewing the VKPacket PR. I can't keep my changes and your suggestions straight in my head as-is. 

I know we disagree on something fundamental, and I'm eager to understand the crux of our disagreement, but I don't have clarity yet.

PR #3091, opened 2023-12-25
Fixes #3080. Potentially unintended behavior of static pre-defined Attribute.Default in Attribute class (v2)

### LINKED ISSUE #3080: Potentially unintended behavior of static pre-defined Attribute.Default in Attribute class (v2)
Like the #3079 changing the `static` `Attribute.Default` from one instance will also modify other instances. This is what is probably causing the #3075 issue.

```cs
	[Fact]
	public void TheseAreNotSafe ()
	{
		// Make two local attributes, and grab Attribute.Default, which is a reference to a singleton.
		Attribute attr1 = Attribute.Default;
		Attribute attr2 = Attribute.Default;

		// Assert the starting state that is expected
		Assert.Equal (ColorName.White, attr1.Foreground.ColorName);
		Assert.Equal (ColorName.White, attr2.Foreground.ColorName);
		Assert.Equal (ColorName.White, Attribute.Default.Foreground.ColorName);

		// Now set Foreground.ColorName to ColorName.Blue on one of our local attributes
		attr1.Foreground.ColorName = ColorName.Blue;

		// Assert the newly-expected case
		// The last two assertions will fail, because we have actually modified a singleton
		Assert.Equal (ColorName.Blue, attr1.Foreground.ColorName);
		Assert.Equal (ColorName.White, attr2.Foreground.ColorName);
		Assert.Equal (ColorName.White, Attribute.Default.Foreground.ColorName);
	}

```

BDisp:
Fixes #3080 - We need to avoid using static parameters on constructors and use the langword `new` to set instance properties.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drawing/Color.cs
+++ b/Terminal.Gui/Drawing/Color.cs
@@ -663,17 +663,22 @@ public readonly struct Attribute : IEquatable<Attribute> {
 	/// </summary>
 	public Attribute ()
 	{
-		var d = Default;
 		PlatformColor = -1;
-		Foreground = d.Foreground;
-		Background = d.Background;
+		var d = Default;
+		Foreground = new (d.Foreground.ColorName);
+		Background = new (d.Background.ColorName);
 	}
 
 	/// <summary>
 	/// Initializes a new instance with platform specific color value.
 	/// </summary>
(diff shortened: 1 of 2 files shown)


tig (approved):
Nice. Thanks.

PR #3096, opened 2023-12-28
Fixes #3095. WindowsDriver should return the mask keys to IsShift, IsAlt and IsCtrl return the right value.

### LINKED ISSUE #3095: WindowsDriver should return the mask keys to IsShift, IsAlt and IsCtrl return the right value.
If only the `ShiftKey`, `AltKey` or `CtrlKey` are used the `Key.IsShift`, `Key.IsAlt` or `Key.IsCtrl` returns false and should return true.

BDisp:
Fixes #3095 - The drivers should use the mask keys before create a `Key` object.

## Pull Request checklist:


tig:
If the user just presses and releases a shift key it should be registered as just that key. 

This change breaks this.

BDisp:
> If the user just presses and releases a shift key it should be registered as just that key.
> 
> This change breaks this.

A shift key pressed or released is always a shift key. If you check the `wVirtualKeyCode` is the same at key down and key up. I don't understand what it is breaking (`16`).

Captura de ecrã 2023-12-28 161548

Captura de ecrã 2023-12-28 161657

BDisp:
If you are meaning that must returns as `16` the `IsShift` never return true. I think you are misunderstood something here.

tig:
Run the Key scenario. Press ctrl and release it.

BDisp:
The issue is because the `Key` class is removing the `KeyCode.ShiftMask` on all situations. The only situation where it must remove the `KeyCode.ShiftMask`, if it's the only modifier key used, is when it's handling a printable character where only the character must be passed. For e.g. `TextView` cannot handle `KeyCode.A | KeyCode.ShiftKey` and thus it only needed the `KeyCode.A`. So the solution is if only the modifiers keys are pressed without another key then they must be handled in the `Key` class as `KeyCode.ShiftMask`, `KeyCode.AltMask`, `KeyCode.CtrlMask` or as `KeyCode.ShiftKey`, `KeyCode.AltKey`, `KeyCode.CtrlKey`.
I'll try to discover some solution for this.

tig:
I think you're wrong but I cannot yet articulate why. 

Please give me a few days to get a chance to dive in. 

I'm traveling internationally next few days.

BDisp:
> I think you're wrong but I cannot yet articulate why.
> 
> Please give me a few days to get a chance to dive in.
> 
> I'm traveling internationally next few days.

No problem, sorry. I only will add my solution and you only accept it after you have time and if you agree with it, of course.

BDisp:
It's working now. Modifiers key must be valid for key down and key up.

https://github.com/gui-cs/Terminal.Gui/assets/13117724/12c3a10e-f4bb-4cbb-a1a3-f1dc723f4809

tig:
Closing this as it's addressed in your other pr.

Issue #3029, opened 2023-12-04
Refactor `MouseEvent` and `MouseEventEventArgs` to simplify


tig:
Similar in spirit, but somewhat less scope, than #2927 

- [x] `New/OnMouseEvent` now uses just `MouseEvent`.
- [x] `New/OnMouseClicked` now uses just `MouseEvent`.
- [x] Better document/rename `OfX`/`OfY`
- [x] Move mouse stuff out of `Responder` and into `View` so `Responder` can be nuked
- [x] Add a `docs/mouse.md`
- [ ] Beef-up/refactor mouse unit tests
- [ ] Build `MouseBindings` per #https://github.com/gui-cs/Terminal.Gui/issues/3778
- [ ] Merge MouseEvent and MouseEventEventArgs into a single class.
- [ ] Engineer something that removes the need for `Responder.IsOverridden` which breaks all rules of OO design.
- [ ] Ensure all library code uses consistent naming/usage
- [ ] Ensure all Scenario code uses best practices related to new design.
- [ ] Ensure all API documentation is awesome

tig:
Partially fixed in #3372.

PR #3104, opened 2023-12-30
Fixes #3097. Now that View is IDisposable, it should expose a Disposing event.

### LINKED ISSUE #3097: Now that View is IDisposable, it should expose a Disposing event
In v2, View is IDisposable.

However, there's no way, at present, for anything dependent on a view to know when that view has been disposed or is being disposed, other than trying something that's illegal at runtime, situationally.

This can lead to memory leaks, runtime exceptions, and other unexpected behavior, in consumers.

Proposal:

Implement a standard Disposing event, to be invoked in the Dispose method, so that anything dependent on a given View can gracefully clean up, when that view is disposed.

One place this is already relevant is in TerminalGuiDesigner: https://github.com/gui-cs/TerminalGuiDesigner/issues/270

Having such an event _does_ expose the possibility of bad consumer code blocking during Dispose, but that's _their_ fault, not Terminal.Gui's. We could attempt to protect users from themselves by implementing some sort of deadline mechanism, but I don't think that's necessary and has more potential downsides than benefits (plus the complexity of implementing it in the first place).

BDisp:
Fixes #3097 - Added `Disposing` event which will help to take the necessary measure when needed.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Input/Responder.cs
+++ b/Terminal.Gui/Input/Responder.cs
@@ -192,10 +192,10 @@ public class Responder : IDisposable {
 	{
 		// Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
 		Dispose (disposing: true);
+		Disposed?.Invoke (this, EventArgs.Empty);
 		GC.SuppressFinalize (this);
 #if DEBUG_IDISPOSABLE
 		WasDisposed = true;
-		Disposed?.Invoke (this, EventArgs.Empty);
 
 		foreach (var instance in Instances.Where (x => x.WasDisposed).ToList ()) {
 			Instances.Remove (instance);
--- a/Terminal.Gui/Input/Responder.cs
+++ b/Terminal.Gui/Input/Responder.cs
@@ -48,7 +48,7 @@ public class Responder : IDisposable {
 #endif
 
 	/// <summary>
-	/// Invokes subscribers that this is disposed.
+	/// Event raised when <see cref="Dispose()"/> is called to signal that this object has been disposed.
 	/// </summary>
 	public event EventHandler Disposing;
 
--- a/Terminal.Gui/Input/Responder.cs
+++ b/Terminal.Gui/Input/Responder.cs
@@ -48,6 +48,11 @@ public class Responder : IDisposable {
 #endif
 
 	/// <summary>
+	/// Invokes subscribers that this is disposed.
+	/// </summary>
+	public event EventHandler Disposing;
+
+	/// <summary>
 	/// Gets or sets a value indicating whether this <see cref="Responder"/> can focus.
 	/// </summary>
 	/// <value><c>true</c> if can focus; otherwise, <c>false</c>.</value>
@@ -187,6 +192,7 @@ public class Responder : IDisposable {
 	{
 		// Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
 		Dispose (disposing: true);
+		Disposing?.Invoke (this, EventArgs.Empty);
 		GC.SuppressFinalize (this);
 #if DEBUG_IDISPOSABLE
 		WasDisposed = true;
--- a/Terminal.Gui/Input/Responder.cs
+++ b/Terminal.Gui/Input/Responder.cs
@@ -48,9 +48,9 @@ public class Responder : IDisposable {
 #endif
 
 	/// <summary>
-	/// Event raised when <see cref="Dispose()"/> is called to signal that this object has been disposed.
+	/// Event raised when <see cref="Dispose()"/> has been called to signal that this object is being disposed.
 	/// </summary>
-	public event EventHandler Disposed;
+	public event EventHandler Disposing;
 
 	/// <summary>
 	/// Gets or sets a value indicating whether this <see cref="Responder"/> can focus.
@@ -191,8 +191,8 @@ public class Responder : IDisposable {
 	public void Dispose ()
 	{
 		// Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
+		Disposing?.Invoke (this, EventArgs.Empty);
 		Dispose (disposing: true);
-		Disposed?.Invoke (this, EventArgs.Empty);
 		GC.SuppressFinalize (this);
 #if DEBUG_IDISPOSABLE
 		WasDisposed = true;
--- a/Terminal.Gui/Input/Responder.cs
+++ b/Terminal.Gui/Input/Responder.cs
@@ -50,7 +50,7 @@ public class Responder : IDisposable {
 	/// <summary>
 	/// Event raised when <see cref="Dispose()"/> is called to signal that this object has been disposed.
 	/// </summary>
-	public event EventHandler Disposing;
+	public event EventHandler Disposed;
 
 	/// <summary>
 	/// Gets or sets a value indicating whether this <see cref="Responder"/> can focus.
@@ -192,10 +192,10 @@ public class Responder : IDisposable {
 	{
 		// Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
 		Dispose (disposing: true);
-		Disposing?.Invoke (this, EventArgs.Empty);
 		GC.SuppressFinalize (this);
 #if DEBUG_IDISPOSABLE
 		WasDisposed = true;
+		Disposed?.Invoke (this, EventArgs.Empty);
 
 		foreach (var instance in Instances.Where (x => x.WasDisposed).ToList ()) {
 			Instances.Remove (instance);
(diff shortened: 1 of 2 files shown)


Terminal.Gui/Input/Responder.cs
```diff
@@ -47,6 +47,11 @@ public Responder ()
 	}
 #endif
 
+	/// <summary>
+	/// Invokes subscribers that this is disposed.
```
tig, line 51: `/// Event raised when <see cref="Dispose"/> is called to signal that this object has been disposed.`

Terminal.Gui/Input/Responder.cs
```diff
@@ -187,6 +192,7 @@ public void Dispose ()
 	{
 		// Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
 		Dispose (disposing: true);
+		Disposing?.Invoke (this, EventArgs.Empty);
```
tig, line 195: I think this line should be moved to after the `wasDisposed = true` line.

UnitTests/Input/ResponderTests.cs
```diff
@@ -112,11 +113,11 @@ public void IsOverridden_True_IfOverridden ()
 		// MouseEvent is defined on Responder IS overriden on ScrollBarView (but not View)
 		Assert.True (Responder.IsOverridden (new ScrollBarView () { Text = "ScrollBarView overrides MouseEvent" }, "MouseEvent"));
 
-		//// OnKeyDown is defined on View
-		//Assert.True (Responder.IsOverridden (new View () { Text = "View overrides OnKeyDown" }, "OnKeyDown"));
+		// OnKeyDown is defined on View
```
tig, line 116: FWIW, this IsOverriden hack must be removed in v2. It should be removed as part of #https://github.com/gui-cs/Terminal.Gui/issues/3029

UnitTests/Input/ResponderTests.cs
```diff
@@ -129,4 +130,124 @@ public void IsOverridden_True_IfOverridden ()
...
 #endif
 	}
+
+	[Fact]
+	public void Responder_Not_Notifying_Dispose ()
+	{
+		List<View> views = new List<View> ();
+		var container1 = new View () { Id = "Container1" };
+
+		for (int i = 0; i < 20; i++) {
```
tig, line 140: Is there really any value in doing these tests 20 times vs just once? 

This seems like un-needed complexity in the unit tests.

If there is a specific edge case that needs to be tested, then a specific unit test that tests that should be provided.

UnitTests/Input/ResponderTests.cs
```diff
@@ -129,4 +130,124 @@ public void IsOverridden_True_IfOverridden ()
 		Assert.Empty (Responder.Instances);
 #endif
 	}
+
```
tig, line 133: Is there value in having a version of these unit tests that tests all view sub-classes?

Terminal.Gui/Input/Responder.cs
BDisp: Ah I only saw this now, in this case `Disposed` is the right name.

UnitTests/Input/ResponderTests.cs
BDisp: Ok I leave as is and then per #3029 will be removed, right?

UnitTests/Input/ResponderTests.cs
BDisp: I think so, because I'm testing 3 cases:
1. Without disposed notification.
2. With disposed notification only on the second container.
3. With disposed notification only on the first container.

UnitTests/Input/ResponderTests.cs
BDisp: I'll change it to only test a single view for all.

Terminal.Gui/Input/Responder.cs
BDisp: Done.

Terminal.Gui/Input/Responder.cs
```diff
@@ -47,6 +47,11 @@ public Responder ()
 	}
 #endif
 
+	/// <summary>
+	/// Invokes subscribers that this is disposed.
+	/// </summary>
+	public event EventHandler Disposing;
```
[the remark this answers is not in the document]
BDisp: Done.

Terminal.Gui/Input/Responder.cs
BDisp: @tig we need to resolve this. I need to invoke the event before the `GC.SuppressFinalize` call. Then I need your finally opinion of the event name, `Disposing` or `Disposed`?

UnitTests/Input/ResponderTests.cs
BDisp: Ak I think your talking about why not only using the base class `Responder`. The reason was I wanted to use `View` properties, like the `Id`, `Subviews`, etc...

Terminal.Gui/Input/Responder.cs
tig: > @tig we need to resolve this. I need to invoke the event before the `GC.SuppressFinalize` call. Then I need your finally opinion of the event name, `Disposing` or `Disposed`?

Explain why please.

BDisp: Although the event is invoked on the `Dispose` method it must be called before calling the `GC.SuppressFinalize` to avoid unit test failure. So it isn't already disposed, although the `View` already disposed others objects. So do you confirm that you want the event named as `Disposed`. If affirmative then it's already changed.

tig: > Although the event is invoked on the `Dispose` method it must be called before calling the `GC.SuppressFinalize` to avoid unit test failure. So it isn't already disposed, although the `View` already disposed others objects. So do you confirm that you want the event named as `Disposed`. If affirmative then it's already changed.

Are you sure the unit test is valid?

BDisp: Of course it's. Now it only handle one view as you asked. But if you think it isn't valid please say why you are think that.

tig: > Of course it's. Now it only handle one view as you asked. But if you think it isn't valid please say why you are think that.

Sorry, but I don't have time to dive deeply into this and understand it myself. I'd like you to explain to me what fails if the event is fired after `GC.SuppresFinalize`. With my limited understanding it should make any difference. 

I can't answer on what the event should be named without understanding this.

BDisp: Because when the event is invoked `container1` was already had disposed the view and the `container2` could not dispose all the `Responder.Instances`. See the error below. But this failure is arbitrary. Before I push the commit all the unit tests passed locally but failed on the server after pushed it.

```
  Failed Terminal.Gui.InputTests.ResponderTests.Disposed_Event_Notify_All_Subscribers_On_The_First_Container [2 ms]
  Error Message:
   Assert.Empty() Failure: Collection was not empty
Collection: [FrameView()(0,0,0,0), Frame(Margin)(0,0,0,0), Frame(Border)(0,0,0,0), Frame(Padding)(1,1,0,0), FrameView(Test)(0,0,0,0), ···]
  Stack Trace:
     at Terminal.Gui.InputTests.ResponderTests.Disposed_Event_Notify_All_Subscribers_On_The_First_Container() in D:\a\Terminal.Gui\Terminal.Gui\UnitTests\Input\ResponderTests.cs:line 238
  Failed Terminal.Gui.InputTests.ResponderTests.Responder_Not_Notifying_Dispose [< 1 ms]
  Error Message:
   Assert.Empty() Failure: Collection was not empty
Collection: [FrameView()(0,0,0,0), Frame(Margin)(0,0,0,0), Frame(Border)(0,0,0,0), Frame(Padding)(1,1,0,0), FrameView(Test)(0,0,0,0), ···]
  Stack Trace:
     at Terminal.Gui.InputTests.ResponderTests.Responder_Not_Notifying_Dispose() in D:\a\Terminal.Gui\Terminal.Gui\UnitTests\Input\ResponderTests.cs:line 168
```

tig: I still don't understand. 

However, I now think that the event should be `Disposing` and should be fired before `protected virtual void Dispose (bool disposing)` is called.

tig: The API doc should read:

`/// Event raised when <see cref="Dispose()"/> has been called to signal that this object is being disposed.`

BDisp: > I still don't understand.

Me either 😃 but I can't debug it because I don't get any fail locally. Some racing condition caused by the event being called after `GC.SuppressFinalize` which depend of the OS to do the cleanup from the garbage collector.

tig (approved):

PR #3130, opened 2024-01-05
Fixes #3127 - View Layout improvements - Redefines how `LayoutStyle` works. Fixes `AutoSize` etc...

### LINKED ISSUE #3127: Simplify `LayoutStyle` - Make it a computed/read-only property ensure correctness
I've confused myself (again) around ComputedLayout. 

I noticed this morning we throw an exception of Width or Height is absolute and ComputedLayout is set. 

Does this actually make sense?

Did I do this? 

Why?

Why can't someone force a view to a fixed size even when using ComputedLayout?

Maybe I just need more coffee?

tig:
Fixes:

- Fixes #3127

Todo: 

- [x] Makes `LayoutStyle` a read-only, computed property.
- [x] Removes `CheckAbosolute` for `Computed`
- [x] Cleans up when/how `Frame` is set.
- [x] Cleans up when/how `Bounds` is set.
- [x] Merge fix for #3135 (`Toplevel` is confused about dimension - Should use `Computed` and `Dim.Fill` )
  - [x] Refactor unit tests to be able to more clearly focus on `Toplevel`
  - [x] Change `Create` to use `Dim.Fill`
  - [x] Fix `SetCurrentOverlappedAsTop` related issues.
  - [x] Fix other code that depends on old bad behavior
  - [x]  Fix related unit tests that assumed old bad behavior
- [x] Fix `DialogTests.Location_When_Not_Application_Top_Not_Default`
- [x] Fix `AutoSizeTextTests`
- [x] Fix `LabelTests.Label_Draw_Vertical_Wide_TextAlignments`
- [x] Fix `RadioGroupTests.DisplayMode_Width_Height_Vertical_Horizontal_Space`
- [x] Fix `ScrollBarViewTests.Hosting_ShowBothScrollIndicator_Invisible`
- [x] Fix `ScrollBarViewTests.InternalTests`
- [x] Fix `ScrollViewTests.AutoHideScrollBars_False_ShowHorizontalScrollIndicator_ShowVerticalScrollIndicator`
- [x] Fix `TextViewTests.UnwrappedCursorPosition_Event`
- [x] Merged #3019 from @BDisp 
- [x] Significantly updated API docs for related APIs - Read them carefully!
- [x] Cleaned up tons of code to match new standards 
- [x] Provide an all views tester unit test that ensures all built in views center properly (see #3156) 
- [x] Merged develop and undid a bunch of formatting that would cause others merge hell.


tig:
@BDisp & @tznind: as I expected, a big challenge in all this will be getting AutoSize working again.... Esp vertical.

Right now almost all unit tests pass.

Those that don't:
- A bunch of AutoSize = true pass. Many of the failing tests have to do with vertical orientation. Shocking, I know.
- A `Dialogs` test that is odd
- `LabelTests`, `RadioGroupTests`, `TextViewTests` - I suspect these are related to Autosize and will pass once that's figured out
- `ScrollBar/ViewTests` - odd. 

However, I have almost everything else working! I only had to change a small number of unit tests; and as I did so I tried to be very careful to ensure the tests tested the new "correct" behavior. 

To be clear, significant behavior has changed as a result of what I've done here:

## `OnResizeNeeded` no longer changes Frame
But calls `SetRelativeLayout`. This reduces the # of places `Frame` is mucked with and actually makes more sense: That is the whole point of `SetRelativeLayout`.

## `Toplevel` now defaults to `Width/Height = Dim.Fill (0)` 

And thus `LayoutStyle.Computed`.  

## `SetRelativeLayout` now works regardless of `LayoutStyle`.

It also, treats each of `X`, `Y`, `Width`, and `Height` as independent w.r.t. Computed:

```cs
var r = new Rect (newX, newY, newW, newH);
if (Frame != r) {
	// Set the frame. Do NOT use `Frame` as it overwrites X, Y, Width, and Height, making
	// the view LayoutStyle.Absolute.
	_frame = r;
	if (X is Pos.PosAbsolute) {
		_x = Frame.X;
	}
	if (Y is Pos.PosAbsolute) {
		_y = Frame.Y;
	}
	if (Width is Dim.DimAbsolute) {
		_width = Frame.Width;
	}
	if (Height is Dim.DimAbsolute) {
		_height = Frame.Height;
	}

	if (IsInitialized) {
		// TODO: Figure out what really is needed here. All unit tests (except AutoSize) pass as-is
		//LayoutFrames ();
		//TextFormatter.Size = GetTextFormatterSizeNeededForTextAndHotKey ();
		SetNeedsLayout ();
		//SetNeedsDisplay ();
	}
	
	// BUGBUG: Why is this AFTER setting Frame? Seems duplicative.
	if (!SetFrameToFitText ()) {
		TextFormatter.Size = GetTextFormatterSizeNeededForTextAndHotKey ();
	}
}
```

## `Frame` 

```cs
public virtual Rect Frame {
	get => _frame;
	set {
		_frame = new Rect (value.X, value.Y, Math.Max (value.Width, 0), Math.Max (value.Height, 0));

		// If Frame gets set, by definition, the View is now LayoutStyle.Absolute, so
		// set all Pos/Dim to Absolute values.
		_x = _frame.X;
		_y = _frame.Y;
		_width = _frame.Width;
		_height = _frame.Height;
		
		// TODO: Figure out if the below can be optimized.
		if (IsInitialized /*|| LayoutStyle == LayoutStyle.Absolute*/) {
			LayoutFrames ();
			TextFormatter.Size = GetTextFormatterSizeNeededForTextAndHotKey ();
			SetNeedsLayout ();
			SetNeedsDisplay ();
		}
	}
}
```

I'm sure more is broken as this is a DEEP change, but I think you'll agree it is far more "correct" and will simplify things moving forward so it's worth it... 😶

tig:
All unit tests pass. 

Now, what's needed is:

- [ ] A code review from @BDisp  and @tznind 
- [ ] A bunch of manual testing to find out what I broke that's not tested by unit tests.

BDisp:
The `SyncrhonizationContextTests` is failing.

image

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -7,55 +7,85 @@
...
-			//Width = _frame.Width;
-			//Height = _frame.Height;
-			if (IsInitialized || LayoutStyle == LayoutStyle.Absolute) {
+
+			// If Frame gets set, by definition, the View is now LayoutStyle.Absolute, so
+			// set all Pos/Dim to Absolute values.
+			_x = _frame.X;
+			_y = _frame.Y;
+			_width = _frame.Width;
+			_height = _frame.Height;
```
BDisp, line 85: This is so bad. `ColorPicker` scenario is one of that prove this code isn't correct. If the intention was move this code from `OnAdded` to here was a very wrong choice, maybe on the `BeginInit` or `EndInit`, but never on the `Frame` property. If some tests was changed to reflect this is very wrong. The `backgroundColorPicker.X` is a `Pos.Combine` and ends up loosing his location. I really don't know why the needed to convert a `Pos/Dim`, not `Absolute`, to an `Absolute` value. The Frame must be set after `SetRelativeLayout` finish the calculation with absolute values, but the `Pos/Dim` must always preserve his value when they aren't `Absolute`.

BDisp (changes requested):
@tig I'm starting with the `Frame` because is one that more offending the `Pos/Dim` feature.

tig:
The `ColorPicker` View is producing exactly the results I would expect it to. It has a bug in how it is trying to implement it's own "AutoSize" given the new design of `Frame` that I've implemented in this PR. Specifically,

```cs
private void SetInitialProperties ()
{
	CanFocus = true;
	AddCommands ();
	AddKeyBindings ();
	LayoutStarted += (o, a) => {
		Bounds = new Rect (Bounds.Location, new Size (_cols * BoxWidth, _rows * BoxHeight));
	};
}
```

Setting `Bounds` has the same effect as setting `Frame`. 

To fix this, `ColorPicker` should change its logic for positioning the background color picker:

```cs
void SetInitialProperties ()
{
	CanFocus = true;
	AddCommands ();
	AddKeyBindings ();
	LayoutStarted += (o, a) => {
		var thickness = GetFramesThickness ();
		Width  = _cols * BoxWidth + thickness.Vertical;
		Height = _rows * BoxHeight + thickness.Horizontal;
	};
}
```

tig:
> The `SyncrhonizationContextTests` is failing.

I can't reproduce this. Weird as it should have nothing to do with my changes.

BDisp:
> I can't reproduce this. Weird as it should have nothing to do with my changes.

I'm using your PR and only happens when I run all the unit tests. But I also can't pull from your remote because I get some git errors and I don't know why.

Terminal.Gui/View/Layout/ViewLayout.cs
BDisp: Ok sorry, I' seeing that you only set the `_frame` and don't call the `Frame` property anymore in the `SetRelativeLayout` method where you only set the `Pos/Dim` if they are `Absolute`. So, setting directly the `Frame` overwrites the `Pos/Dim` and that begin making sense to me. Good work. Probably many views and scenarios needed/need to be changed to reflect this new behavior.

tig:
> > I can't reproduce this. Weird as it should have nothing to do with my changes.
> 
> I'm using your PR and only happens when I run all the unit tests. But I also can't pull from your remote because I get some git errors and I don't know why.

What's the error?

All tests pass for me with:

- VS 2022's Test Explorer - both Windows and Ubuntu
- Resharpers' Unit Test Explorer
- Github action on Push
- `dotnet clean & dotnet test`

BDisp:
> What's the error?
> 
> All tests pass for me with:
> 
> * VS 2022's Test Explorer - both Windows and Ubuntu
> * Resharpers' Unit Test Explorer
> * Github action on Push
> * `dotnet clean & dotnet test`

I deleted your remote from the .git and fetch again and now it isn't fail anymore. I think the culprit was two branches where you submit 2 PR to my branch with the same name but one start with capital letter and the other with lower letter.

```
bDisp-v2_vkeypacketsimulator-fix_3054
BbDisp-v2_vkeypacketsimulator-fix_3054
```

UnitTests/View/Layout/AbsoluteLayoutTests.cs
```diff
@@ -321,4 +321,69 @@ public void AbsoluteLayout_LayoutSubviews ()
 		Assert.Equal (new Rect (10, 10, 10, 10), v2.Frame);
 		super.Dispose ();
 	}
+
+	[Fact]
+	public void AbsoluteLayout_Setting_Bounds_Location_NotEmpty ()
+	{
+		// TODO: Should we enforce Bounds.X/Y == 0? The code currently ignores value.X/Y which is
+		// TODO: correct behavior, but is silent. Perhaps an exception?
```
BDisp, line 329: I think Bounds is very tied to the Margin, Border and Padding thickness. So, the calculation from Frame to Bounds or Bounds to Frame will always return the same values for each situation. I know you have on minds to use negative Bounds location for scrolling purposes by using the Padding to accommodate the scroll bars. But I think that will never work like that without using a container with different location and size, inside the Bounds. But as you always think far ahead, you may already have some idea about this subject.

UnitTests/View/Layout/AbsoluteLayoutTests.cs
```diff
@@ -321,4 +321,69 @@ public void AbsoluteLayout_LayoutSubviews ()
...
+
+		v.Bounds = newBounds;
+		Assert.True (v.LayoutStyle == LayoutStyle.Absolute);
+		Assert.Equal (newBounds,                                          v.Bounds);
+		Assert.Equal (new Rect (1, 2, newBounds.Width, newBounds.Height), v.Frame);
+		Assert.Equal (new Rect (0, 0, newBounds.Width, newBounds.Height), v.Bounds); 
+		Assert.Equal (Pos.At (1),                                         v.X);
+		Assert.Equal (Pos.At (2),                                         v.Y);
+		Assert.Equal (Dim.Sized (30),                                     v.Width);
+		Assert.Equal (Dim.Sized (40),                                     v.Height);
```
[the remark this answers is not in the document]
tig: I want to fix that. But yeah.

BDisp (approved):
Waiting for the `Dim.Auto` to see the `AutoSize` behavior.

Terminal.Gui/View/Layout/ViewLayout.cs
tig: > Ok sorry, I' seeing that you only set the `_frame` and don't call the `Frame` property anymore in the `SetRelativeLayout` method where you only set the `Pos/Dim` if they are `Absolute`. So, setting directly the `Frame` overwrites the `Pos/Dim` and that begin making sense to me. Good work. Probably many views and scenarios needed/need to be changed to reflect this new behavior.

I think I've fixed all those issues. Or at least the majority of them. 

Please keep testing. 

Also, please `./docfx/build.ps1` and click on  http://localhost:8080.

Read and review all related API docs. I worked hard to update them ALL to reflect the new design. While finding issues in the code is important and useful, it's even more important that the documentation is accurate (and/or you don't have any disagreements with what it says).

tig:
> Waiting for the `Dim.Auto` to see the `AutoSize` behavior.

So am I. 😜

Don't hold your breath. It's a lot of work and I don't know when I can get to it again. But I'm confident this PR will make it easier.

tig:
I think this is ready to merge.

@BDisp ok?

@tznind ok?

BDisp (approved):

Issue #3124, opened 2024-01-05
0001 for year in Datefield


petervdpas:
I can't enter 0001 for year in Datefield... is there something I miss?

MaciekWin3:
Absolutely, you're correct. I attempted to refactor this component to address the issue, but I've encountered some aspects where I'm unsure about the intended behavior. Specifically, there's a field called ShortDate that stores a 2-digit year. While this is convenient, it becomes challenging to distinguish whether the date "06/01/2024" is referring to the year 2024 or 0024. Additionally, I believe it would be beneficial if users could manually specify the date format, as currently, only "MM/dd/yy" and "MM/dd/yyyy" are supported.

I am willing to refactor the code since I've already invested some time in familiarizing myself with the component's codebase. However, I need clarification on the intended behavior to proceed with the changes. @tig can you help?

BDisp:
> Absolutely, you're correct. I attempted to refactor this component to address the issue, but I've encountered some aspects where I'm unsure about the intended behavior. Specifically, there's a field called ShortDate that stores a 2-digit year. While this is convenient, it becomes challenging to distinguish whether the date "06/01/2024" is referring to the year 2024 or 0024. Additionally, I believe it would be beneficial if users could manually specify the date format, as currently, only "MM/dd/yy" and "MM/dd/yyyy" are supported.
> 

With a 2-digit year it's really a challenge to distinguish if the year is referring to the year 0024, 0124, 1924 or 2024. So we can't rely on 2-digit.
The `CultureInfo.CurrentCulture.DateTimeFormat` is also supported which provide much more formats.

> I am willing to refactor the code since I've already invested some time in familiarizing myself with the component's codebase. However, I need clarification on the intended behavior to proceed with the changes. @tig can you help?

I'm preparing a PR to fix only some related issues, but a view like a DatePicker where the user can set it by opening a combo box would be great. Would you are interest on this?

MaciekWin3:
Hey @BDisp, yeah, it sounds great. It will be quite a challenge, but I'll see what I can do. I'll give it a try next week! 😄

tig:
I'm of the opinion that we should leverage as much existing dotnet infrastructure for these Views as possible. The existing implementation is way too bespoke. 

Its well known that date handling is the hardest problem in CompSci 😉

So, big fan of a rewrite that has very little code in it because it uses the work of others smarter than us.

MaciekWin3:
[...]

Issue #1720, opened 2022-05-14
Add ColorPicker pop-up example


tig:
Very related to 

- #2800

I already coded this up as part of my code review of `ColorPicker` (#1668). But I didn't add it because I built it on `ProgressBar Styles` and it really should be part of the `Color Picker` Scenario.

This code shows how to wrap `Color Picker` in a `Dialog` such that when you click on a color, the dialog closes and the `ProgressBar` colors change. 

My suggestion is to add a `ProgressBar` to the `Color Picker` scenario and use this code (or an adaptation) to make it sing:

```csharp
		#region ColorPicker
		ColorName ChooseColor (string text, ColorName colorName)
		{

			var colorPicker = new ColorPicker {
				Title = text,
				SelectedColor = colorName
			};

			var dialog = new Dialog {
				Title = text
			};

			dialog.LayoutComplete += (sender, args) => {
				// TODO: Replace with Dim.Auto
				dialog.X = pbList.Frame.X;
				dialog.Y = pbList.Frame.Height;

				// Once #3127 is merged:
				dialog.LayoutStyle = LayoutStyle.Absolute;
				dialog.Bounds = new Rect (0, 0, colorPicker.Frame.Width, colorPicker.Frame.Height);

				Application.Top.LayoutSubviews ();
			};

			dialog.Add (colorPicker);
			colorPicker.ColorChanged += (s, e) => {
				dialog.RequestStop ();
			};
			Application.Run (dialog);

			var retColor = colorPicker.SelectedColor;
			colorPicker.Dispose ();

			return retColor;
		}

		var fgColorPickerBtn = new Button {
			Text = "Foreground HotNormal Color",
			X = Pos.Center (),
			Y = Pos.Bottom (pbList),
		};
		editor.Add (fgColorPickerBtn);
		fgColorPickerBtn.Clicked += (s, e) => {
			var newColor = ChooseColor (fgColorPickerBtn.Text, editor.ViewToEdit.ColorScheme.HotNormal.Foreground.ColorName);
			var cs = new ColorScheme (editor.ViewToEdit.ColorScheme) {
				HotNormal = new Attribute (newColor, editor.ViewToEdit.ColorScheme.HotNormal.Background)
			};
			editor.ViewToEdit.ColorScheme = cs;
		};

		var bgColorPickerBtn = new Button {
			X = Pos.Center (),
			Y = Pos.Bottom (fgColorPickerBtn),
			Text = "Background HotNormal Color"
		};
[...]

PR #3134, opened 2024-01-07
Fixes #2019: Introduce DatePicker

### LINKED ISSUE #2019: Add DatePicker control
Can I represent more dates in DateField then choose one? I need **DatePicker** here in terminal?

MaciekWin3:
## Fixes 

- #2019  - Introduce DatePicker View

As discussed in issue #3124, I took on the task of implementing the DatePicker control.  I aimed to maximize code reuse and maintain simplicity. The outcome is quite satisfying, leveraging multiple existing components. 

datepicker

I intentionally kept the DatePicker independent of the DateField view for simplicity, as the latter requires improvements. Consequently, I made the deliberate decision for the DatePicker to inherit from TextField to maintain clarity and avoid associated complexities. Going forward, there is potential to merge these components, ensuring, for example, that users cannot input letters into the DateField.

## Pull Request checklist:


BDisp:
@MaciekWin3 I really love it, it's veryyyyy great. Thanks. I would like to suggest that you can create an option that days of previous and the follow month grayed. When the user click on a grayed day that date is selected. If it's complicated don't mind and forget.

BDisp:
It's better validating when typing or paste invalid date.

image

BDisp:
Would be better only click be used to select date than double click?

BDisp:
It should use the `CultureInfo.CurrentCulture` in the constructor by default and the invariant culture in the unit tests. Thus the date will be showing in the user culture in the `DatePicker` scenario.

BDisp:
Another suggestion is the current `TextField.Text` date is selected  on the calendar after it's opened.

MaciekWin3:
> It's better validating when typing or paste invalid date.

Yeah, I skipped validation for the text field of DatePicker because, as I mentioned, in the future, It would be great to merge DateField and DatePicker, with DateField handling all validation. However, if necessary, I can add some validation to this component 😉

>I would like to suggest that you can create an option that days of previous and the follow month grayed. When the user click on a grayed day that date is selected. If it's complicated don't mind and forget.

Similarly, I believe it's achievable, but I prefer waiting for the DateField refactor, which I see you've already started in #3124. I can further refactor it, and then we can merge both components, ensuring the date control part is polished. :D

> Would be better only click be used to select date than double click?

Hmm, that's an interesting perspective. My thought process was that after highlighting a day, you'd need to confirm it, as a single click might accidentally close dialog and force user to  reopen the DatePicker. I understand your suggestion, and both options could work. What do you think?

> Another suggestion is the current TextField.Text date is selected on the calendar after it's opened.

Yes, you are right. The user can either provide a default date or it will be set to the current date by default. Do you think it should be left empty by default?

> It should use the CultureInfo.CurrentCulture in the constructor by default and the invariant culture in the unit tests. Thus the date will be showing in the user culture in the DatePicker scenario.

Although not in the constructor, I set the default format to CultureInfo.CurrentCulture and updated the unit tests.

BDisp:
> Yeah, I skipped validation for the text field of DatePicker because, as I mentioned, in the future, It would be great to merge DateField and DatePicker, with DateField handling all validation. However, if necessary, I can add some validation to this component 😉

> Similarly, I believe it's achievable, but I prefer waiting for the DateField refactor, which I see you've already started in #3131. I can further refactor it, and then we can merge both components, ensuring the date control part is polished. :D

You are right. Let's wait until you merge them. So, don't worry about this for now.

> Hmm, that's an interesting perspective. My thought process was that after highlighting a day, you'd need to confirm it, as a single click might accidentally close dialog and force user to reopen the DatePicker. I understand your suggestion, and both options could work. What do you think?

Well when started test it I check that it wasn't closing after click but then I double clicked and it closed. Also the `TableView` maybe only may work with double click, I really don't know.

> Yes, you are right. The user can either provide a default date or it will be set to the current date by default. Do you think it should be left empty by default?

If the Text is is null or empty, then the default is the current date, I think, otherwise it will be the Text date. I'm also remember a user request to type `0001` as year. But in this case is better not adding all the years in the combo box because of the performance, but only adding a dynamic range size as needed.
Another suggestion in the case of the `DatePicker` doesn't not have the current date, is adding a button "Today" to get it.

> Although not in the constructor, I set the default format to CultureInfo.CurrentCulture and updated the unit tests.

Good and when you refactor all in the future they will be joined.
Things that you maybe could do is using the current culture in the `DatePicker` control to use the user culture for the month names and day of the week. The `DateTime` provide them in the user culture. So the trick is using the int value to get the months and week days to display it instead of fixed strings.
Once more thank you very much for your great work.

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+	/// <summary>
+	/// Initializes a new instance of <see cref="DatePicker"/> with the specified date and format.
+	/// </summary>
+	public DatePicker (DateTime date, string format) : base ()
+	{
+		Format = format;
+		Initialize (date);
+	}
+
+	private void Initialize (DateTime date)
```
tig, line 68: a) No need to use `private`

b) Name this guy `SetInitialProperties`. This is to avoid confusion with `Begin/EndInit`.

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+			.Select (x => x.ToString ())
+			.ToList ();
+
+		if (date.Year < YearsRange.Start.Value || date.Year > YearsRange.End.Value) {
+			MessageBox.ErrorQuery ("Error", "Date year is out of range", "Ok");
+			return;
+		}
+
+		var dialog = new Dialog () {
+			Height = 14,
```
tig, line 86: Why these absolute values?

It's a PITA until `Dim.Auto` is implemented, but subscribe to `LayoutComplete` and adjust the Dim of the Dialog based on the size of the Views you've added.

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+			Width = 40,
+			Title = "Date Picker",
+		};
+
+		calendar = new TableView () {
+			X = 0,
+			Y = Pos.Center () + 1,
+			Width = 22,
```
tig, line 94: Please see if you can figure out how to not make this a literal. At the minimum create a const with a  good name.

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+
+		var yearsLabel = new Label ("Year:") {
+			X = Pos.Right (calendar) + 1,
+			Y = 1,
+			Height = 1,
+		};
+		var comboBoxYear = new ComboBox (years) {
+			X = Pos.Right (calendar) + 1,
+			Y = 2,
+			Width = 12,
```
tig, line 114: Why 12 and not 13 or 11? (Really same comment as above).

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+			Height = 4,
+			SelectedItem = years.IndexOf (date.Year.ToString ())
+		};
+		comboBoxYear.SelectedItemChanged += ChangeYearDate;
+
+		var monthsLabel = new Label ("Month:") {
+			X = Pos.Right (calendar) + 1,
+			Y = 6,
```
tig, line 122: Same. I think this one could be `Y = Pos.Bottom (...)`?

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+			Height = 1,
+		};
+
+		var comboBoxMonth = new ComboBox (months) {
+			X = Pos.Right (calendar) + 1,
+			Y = 7,
+			Width = 12,
```
tig, line 129: Ditto.

tig: (not gonna comment any more on the rest of the int literals. You get the point).

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+
+		if (date.Year < YearsRange.Start.Value || date.Year > YearsRange.End.Value) {
+			MessageBox.ErrorQuery ("Error", "Date year is out of range", "Ok");
+			return;
+		}
+
+		var dialog = new Dialog () {
+			Height = 14,
+			Width = 40,
+			Title = "Date Picker",
```
tig, line 88: Please use `Terminal.Gui.Resources.Strings` to support i18n instead of string literals.

Terminal.Gui/Views/DatePicker.cs
```diff
@@ -0,0 +1,298 @@
...
+using System.Linq;
+
+namespace Terminal.Gui.Views;
+
+/// <summary>
+/// The <see cref="DatePicker"/> <see cref="View"/> Date Picker.
+/// </summary>
+public class DatePicker : TextField {
+
+	private DataTable table;
```
tig, line 19: Use `_table` style for class private members. Thanks.

UICatalog/Scenarios/DatePickers.cs
```diff
@@ -0,0 +1,66 @@
...
+using Terminal.Gui;
+using Terminal.Gui.Views;
+
+namespace UICatalog.Scenarios;
+[ScenarioMetadata (Name: "Date Picker", Description: "Demonstrates how to use DatePicker class")]
+[ScenarioCategory ("Controls")]
+[ScenarioCategory ("DateTime")]
+public class DatePickers : Scenario {
+
+	DatePicker datePicker;
```
tig, line 11: `_datePicker` etc.. please.

tig (changes requested):

BDisp:
@MaciekWin3 now that the #3131 is already merged you are free to refactor the `DateField` as you desire. Good work. Thanks.

UICatalog/Scenarios/DatePickers.cs
MaciekWin3: Done

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done, yeah I get the point 😁

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done, for now I just moved it to separate variables as I couldn't figure out how to set up all the components to make it look good. Could you guide me through the process? I'm eager to understand how it works and learn how to make it function correctly.

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: I designed a calendar using Dim.Fill(), as its height is not constant (it changes depending on the month), and when I was playing with LayoutComplete, I couldn't make it work. The dialog box was always stretched and not centered. I am interested in how you would approach this problem. For now, I've just moved it to separate variables. Once I understand it, I'll be happy to improve that.
The absolute variables are just the ones I found to nicely align the calendar :D

MaciekWin3:
> @MaciekWin3 now that the #3131 is already merged you are free to refactor the `DateField` as you desire. Good work. Thanks.

Awesome! Once this PR is merged, I'll take a look at the DateField and see what can be improved :D Thanks for the heads up.

tig:
@MaciekWin3 - Can you please put all localizable strings in `Resources.Strings`?

MaciekWin3:
> @MaciekWin3 - Can you please put all localizable strings in `Resources.Strings`?

Done

tig:
> > @MaciekWin3 - Can you please put all localizable strings in `Resources.Strings`?
> 
> Done

Sorry, I guess I'm not being clear. In this PR you provide string literals for a whole slew of things, all in english (except "Date Picker").

Except  for debug/diagnostic strings, these should ALL be pulled from `Resources.Strings`. 

Examples:

```cs
_table.Columns.Add ("Su");
_table.Columns.Add ("Mo");
_table.Columns.Add ("Tu");
_table.Columns.Add ("We");
_table.Columns.Add ("Th");
_table.Columns.Add ("Fr");
_table.Columns.Add ("Sa");

MessageBox.ErrorQuery ("Error", "Date year is out of range", "Ok");
```

In addition you used `.ToString()` on the `Month` enum, which will not be correct under a different locale.

This may prove helpful: https://stackoverflow.com/questions/5798908/how-to-produce-localized-date-string-with-cultureinfo

BDisp:
You can leverage the DateTime to get the right name for month, day of week, etc... by using the Current.Culture.

MaciekWin3:
> > > @MaciekWin3 - Can you please put all localizable strings in `Resources.Strings`?
> > 
> > 
> > Done
> 
> Sorry, I guess I'm not being clear. In this PR you provide string literals for a whole slew of things, all in english (except "Date Picker").
> 
> Except for debug/diagnostic strings, these should ALL be pulled from `Resources.Strings`.
> 
> Examples:
> 
> ```cs
> _table.Columns.Add ("Su");
> _table.Columns.Add ("Mo");
> _table.Columns.Add ("Tu");
> _table.Columns.Add ("We");
> _table.Columns.Add ("Th");
> _table.Columns.Add ("Fr");
> _table.Columns.Add ("Sa");
> 
> MessageBox.ErrorQuery ("Error", "Date year is out of range", "Ok");
> ```
> 
> In addition you used `.ToString()` on the `Month` enum, which will not be correct under a different locale.
> 
> This may prove helpful: https://stackoverflow.com/questions/5798908/how-to-produce-localized-date-string-with-cultureinfo

Now we are on the same page! 😁 The calendar labels are now localized, and I have also implemented a couple more improvements. Most importantly, I simplified the code, fixed a couple of issues and as suggested, we are relying more on DateTime instead of handmade enums. The layout is no longer relying on hardcoded values, and we handle the year out of range in code instead of showing an error message.

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done, we no longer rely on hardcoded values.

Terminal.Gui/Views/DatePicker.cs
MaciekWin3: Done, now the dialog layout doesn't use hardcoded values.

BDisp:
It's all perfect @MaciekWin3, thanks.

BDisp:
This is a Portuguese culture. Maybe you want to add to the `Resources.Strings` the `Year:`, `Month:`, `Previous` and `Next`. Then you pull the values from there based on the `Thread.CurrentThread.CurrentUICulture`. See the TextField where the `ContextMenu` pul from `Resources.Strings`, like `Strings.ctxSelectAll`.
Another suggestion is the option to do a shorter column with for the days of week where you can on get the first 2 letters with the first one capitalized. You can also capitalize the first letter on the month combo box. Remember it's only a suggestion. Thanks.

image

tig (changes requested):
Really, really, fantastic stuff!

Should the TextField of the Month be read only? Will users really want to TYPE a month vs. use the dropdown to select?

Same question for Year, but I actually think the answer there is no. Except... if I type "24" I sort expect it to be smart and realize I mean 2024 (and not 1924).

Try this: Change Year to "10". Press Next. It seems like it's working and LOOKS like I'm seeing dates for "2010" (10), but I'm not. 

Another example. Change it to "10", click Next, then drop Year down again. It shows 1900.

Addressing all this adds lots of complexity. So maybe it should just be read only too?

The whole "Click the dot to open..." thing screams of poor visibility. Did you experiment with other UI treatments for this?

Not needed now, but ideas for future refinement:

* Why `-` for empty days? Wouldn't ` ` look cleaner?

* A simpler UI that preserves UI real-estate would be to use the same Glpyhs FileDialog uses for forward/back:

image 
image

* These could be just below the calendar.

* Year could be above.




Future enhancement: Support selecting a date range.

tig:
Some more shower thoughts:

I don't like that `DatePicker` inherits from `DateField`. I would prefer `DatePicker` be a stand-alone view that works fine as a modal and non-modal. Like `Wizard`. The code that lets you pop up a `DatePicker` from a `DateField` should be  sample code in the `DatePickers` scenario. 

I'm thinking through the same things with the new `ColorPicker`. See the exmaple code here: https://github.com/gui-cs/Terminal.Gui/issues/1720. 

The `DatePickers` scenario should also show how to use the view non-modally. One idea I had for that would be to make it a "Famous people's birthday" demo (this would demonstrate the eventing).

Also, I think we should drive consistency for modal views that are "pickers"... The only other "picker" we have (where there's actually code that pops it up) is `ContextMenu`, uses `Key.F10.WithShift`. We should standardize on that. However, don't follow `ContextMenu`'s lead of being a hidden control that has a `Show` command. `ContextMenu` is not a good example IMO. All that logic should just be in the `Visible` setter.

If you're not up for all this fanciness, I get it. You just need a modal date picker. But I'd rather not merge in an API that binds `DatePicker` to `DateField` and can't be used  non-modally from the get go.

BDisp:
> We should standardize on that. However, don't follow `ContextMenu`'s lead of being a hidden control that has a `Show` command. `ContextMenu` is not a good example IMO. All that logic should just be in the `Visible` setter.

The way you say the things, I just have to criticize. Why using the `Visible` property for a popup which we maybe only open once and keep always on memory to consuming unnecessary resources, instead of open and dispose as needed.

tig:
> > We should standardize on that. However, don't follow `ContextMenu`'s lead of being a hidden control that has a `Show` command. `ContextMenu` is not a good example IMO. All that logic should just be in the `Visible` setter.
> 
> The way you say the things, I just have to criticize. Why using the `Visible` property for a popup which we maybe only open once and keep always on memory to consuming unnecessary resources, instead of open and dispose as needed.

I say things the way I say them in the hope of generating debate. I don't take you disagreeing or pointing out flaws in my thinking as criticism, because I know my ideas are sometimes half-baked (or at least not perfect). I love it when people disagree with me and say WHY they disagree. Vigorous and open debate is how real innovation and progress are done! 

And you just did that, so 👌.

Your point is a good one. I'm not convinced the memory we're talking about is that big a deal, but you are right that we should consider it. At the minimum, we should make it so that it's easy to defer creation.

BDisp:
> I say things the way I say them in the hope of generating debate. I don't take you disagreeing or pointing out flaws in my thinking as criticism, because I know my ideas are sometimes half-baked (or at least not perfect). I love it when people disagree with me and say WHY they disagree. Vigorous and open debate is how real innovation and progress are done!
> 
> And you just did that, so 👌.
> 

You made me laugh with this one. 😆 

> Your point is a good one. I'm not convinced the memory we're talking about is that big a deal, but you are right that we should consider it. At the minimum, we should make it so that it's easy to defer creation.

The point is it's for use the `Visible` property we need to know if it's already created and toggle the visibility and by creating an instance and destroying we don't need to worry about that. I thinking the visibility will envolve more code and if we need to iterate with the sub-views modal or non-modal will be more confusing if we have to consider the visibility.

tig:
> The point is it's for use the `Visible` property we need to know if it's already created and toggle the visibility and by creating an instance and destroying we don't need to worry about that. I thinking the visibility will envolve more code and if we need to iterate with the sub-views modal or non-modal will be more confusing if we have to consider the visibility.

I see your point. I concur now.

BDisp:
@tig another thing you are wrong is about the the `DatePicker` will inherits from the `DateField`. @MaciekWin3 don't really want to do after this PR is merged. He want to leverage some validation, edition, copy/paste of the `DateField` to integrate on the `DatePicker`, by also removing the short format data that is better not use it. I think if a user can edit/copy/paste a date directly from the `DateField` is very useful too. So, as he said he want refactor the `DatePicker` with some existing features of the `DateField` and then get rid of the `DateField`. Is this a problem?

MaciekWin3:
> The whole "Click the dot to open..." thing screams of poor visibility. Did you experiment with other UI treatments for this?

I attempted a few approaches, including using an emoji like 📅, but unfortunately, it didn't work out. I agree that this issue needs attention, as the black dot is not aesthetically pleasing. The optimal solution would be to use the emoji 📅, but I couldn't figure that out.

> Why - for empty days? Wouldn't   look cleaner?

Regarding the hyphen character '-', it seems to be the default representation for null in the TableView. Additionally, following @BDisp suggestion, it would be beneficial to implement grayed-out days, like here: 
image
We can consider adding this in future refinements.

> @tig another thing you are wrong is about the the `DatePicker` will inherits from the `DateField`. @MaciekWin3 don't really want to do that after this PR is merged. He want to leverage some validation, edition, copy/paste of the `DateField` to integrate on the `DatePicker`, by also removing the short format data that is better not use it. I think if a user can edit/copy/paste a date directly from the `DateField` is very useful too. So, as he said he want refactor the `DatePicker` with some existing features of the `DateField` and then get rid of the `DateField`. Is this a problem?

As BDisp pointed out, it's essential to clarify that, at the moment, this component inherits from TextField, not DateField. One idea is to merge this component together (after some work is done on both DatePicker and DateField) and introduce some kind of flag. Users can then choose whether they want to exclusively use the datepicker with a readonly text field or set the date in both ways (via the datepicker or manually). I'm holding off on making these changes now, as DateField requires some work, and I need to modify the public API (specifically, remove IsShortDate)."

So, guys, let me know if I should proceed with this pull request and identify any remaining steps. I see that there is a lot that can be improved in this component, and several enhancements can be introduced. However, if these changes pose a challenge, I would prefer to shift my focus to something else. Either way, I appreciate your input and involvement. I'm learning a lot from this collaboration!

tig:
> I'm thinking through the same things with the new `ColorPicker`. See the exmaple code here: #1720.
> 
> The `DatePickers` scenario should also show how to use the view non-modally. One idea I had for that would be to make it a "Famous people's birthday" demo (this would demonstrate the eventing).
> 
> Also, I think we should drive consistency for modal views that are "pickers"... The only other "picker" we have (where there's actually code that pops it up) is `ContextMenu`, uses `Key.F10.WithShift`. We should standardize on that. However, don't follow `ContextMenu`'s lead of being a hidden control that has a `Show` command. `ContextMenu` is not a good example IMO. All that logic should just be in the `Visible` setter.
> 
> If you're not up for all this fanciness, I get it. You just need a modal date picker. But I'd rather not merge in an API that binds `DatePicker` to `DateField` and can't be used non-modally from the get go.

I mis-typed `DatePicker` instead of `TextView`. My point below holds:

> I'm thinking through the same things with the new ColorPicker. See the example code here: https://github.com/gui-cs/Terminal.Gui/issues/1720.

> The DatePickers scenario should also show how to use the view non-modally. One idea I had for that would be to make it a "Famous people's birthday" demo (this would demonstrate the eventing).

> If you're not up for all this fanciness, I get it. You just need a modal date picker. But I'd rather not merge in an API that binds DatePicker to DateField and can't be used non-modally from the get go.

MaciekWin3:
Okay, so simply have DatePicker not being part of the modal and not inheriting from TextField or DateField, right?

BDisp:
> Okay, so simply have DatePicker not being part of the modal and not inheriting from TextField or DateField, right?

I think you should have `DatePicker` as not modal, like tig did with the `ColorPicker`. But you will need to inheriting from the `TextField` to display and edit/copy/paste the date without the `DatePicker`. Try to add/remove the code from the `DateField` to the `DatePicker` and delete it after you satisfied. If the user needs to make it as readonly then is only need to set the property and not allowing to popup the calendar.
The calendar glyph is a good idea but it's a surrogate pair (U+1F4C5 📅), which would not render on all the terminal.

BDisp:
With `NetDriver`:

image

BDisp:
@tig I thought that wider code point already was fixed using true color.

MaciekWin3:
So as suggested I worked out this solution: 
dp

`DatePicker` now inherits from the `View` class. I changed it from a dialog to a standalone component, although it still utilizes `DateField`. Since @BDisp resolved the annoying bug in #3124 , I think it is safe to use.

Now, the question arises: Should `DatePicker` and `DateField` be merged together? In my opinion, no. `DatePicker` occupies valuable screen space, and sometimes users may prefer to use a simple `DateField`. However, we can expand `DateField` to have an option to include an icon, as @BDisp demonstrated, and then display DatePicker in a dialog. Additionally, as mentioned earlier, it would be beneficial to eliminate the `IsShortFormat` flag and support a format defined by the user. But I believe these changes are beyond the scope of introducing `DatePicker`. Currently, even though I left the Format property on DatePicker, it doesn't have any effect.

Since users can now type the date in `DateField`, and changes are reflected in the calendar, I removed ComboBoxes. Supporting them was challenging, and they looked unattractive. It also allows us to remove `YearRange` from properties.

If you think that these changes are heading in the right direction, I will add additional unit tests and also a more interesting example to the UICatalog.

tig:
> `DatePicker` now inherits from the `View` class. I changed it from a dialog to a standalone component, although it still utilizes `DateField`. Since @BDisp resolved the annoying bug in #3124 , I think it is safe to use.

Sweet!

> Now, the question arises: Should `DatePicker` and `DateField` be merged together? In my opinion, no. `DatePicker` occupies valuable screen space, and sometimes users may prefer to use a simple `DateField`. However, we can expand `DateField` to have an option to include an icon, as @BDisp demonstrated, and then display DatePicker in a dialog. Additionally, as mentioned earlier, it would be beneficial to eliminate the `IsShortFormat` flag and support a format defined by the user. But I believe these changes are beyond the scope of introducing `DatePicker`. Currently, even though I left the Format property on DatePicker, it doesn't have any effect.

I concur with this.

> Since users can now type the date in `DateField`, and changes are reflected in the calendar, I removed ComboBoxes. Supporting them was challenging, and they looked unattractive. It also allows us to remove `YearRange` from properties.

Nice. 

> If you think that these changes are heading in the right direction, I will add additional unit tests and also a more interesting example to the UICatalog.

I do!

tig:
> If you want to try do this:
> 
> In the Glyphs.cs:
> 
> `public Rune Calendar { get; set; } = new Rune(0x1F4C5);`
> 
> In the DatePicker:
> 
> ```cs
> Move (Bounds.Right - 2, 0);
> Driver.AddRune (Glyphs.Calendar);
> ```

A fun experiment, yes. However, until we figure out how apps can detect whether a glyph is supported by the terminal (a very hard problem with no solution in sight) we should not try this.

MaciekWin3:
I out of ideas how to make the UICatalog example more interesting, so I left it as it is and focused on simplicity. Once we have all the improvements, I'll try to expand it. Speaking of improvements, I created a list for future work:
* Ranges, a very cool idea proposed by @tig , and I think it would be fun to implement.
* Greyed out days instead of the '-' character in the calendar, proposed by @BDisp .
* A more interesting example in UICatalog.
* Refactor of DateField, especially regarding the format. Both DateField and DatePicker would benefit from that. Currently, the DatePickers Format property does nothing. Also, maybe a property/flag that adds an icon so the user can open the DatePicker dialog from the DateField.
* Add more unit tests

That's all for this PR. I am waiting for you feedback @tig @BDisp

tig (approved):
Nice. Simple and clean.

BDisp:
Yes, that cool. There is only a issue on the `DateField`, which I'll submit a PR, where it's validating wrong the date on different culture than us-US. e.g. if I choose day 13 in the pt-PT culture the date isn't updated because it's parsing as MM-dd-yyyy instead of dd-MM-yyyy. I also will remove the old short date that isn't used anymore nowadays. Finally I will to allow to select text by mouse too. Thanks.

tig:
You have a unit test failure. It seems you're not disposing a view.

See `Run_All_Scenarios` unit test.

tig:
Also `AllViews_Enter_Leave_Events`.

BDisp:
Ah yes, I forgot to alert of the failures. You must override `Dispose` on the `DatePicker` to remove all the sub-views. Before when you used the Dialog it was disposing on closing, now you have to handle that.

MaciekWin3:
So, I guess we are not done yet! 😉 I encountered an obstacle with the `AllViews_Enter_Leave_Events` test. If I understand correctly, since the `DatePicker` view is constructed with smaller views, pressing the `Tab` key doesn't exit the view; instead, it stays inside it, just switching focus within the view. Since I don't want to rewrite the entire unit test just to accommodate this one edge case, do you have any ideas or workarounds? For `Run_All_Scenarios` disposing of every subview works fine.

BDisp:
That's the normal view navigation. One work around is adding a keybinding for Ctrl+Tab (forward) and Ctrl+Shift+Tab (backward) like the `TextView` has. You also can add more Ctrl+PgDown (forward) and Ctrl+PgUp (backward).

MaciekWin3:
> In the `AllViews_Enter_Leave_Events` exclude the `DatePicker` from that unit test, like the `TextView` also does.
> 
> ```cs
> 			if (vType is TextView || vType is DatePicker) {
> 				top.NewKeyDownEvent (new (KeyCode.Tab | KeyCode.CtrlMask));
> 			} else {
> 				top.NewKeyDownEvent (new (KeyCode.Tab));
> 			}
> ```
Yes, but in `DatePicker` we need 4 tabs to leave the component not one. After pressing tab one (like in test) we land on calendar day selection.

MaciekWin3:
This works, so I'll go with that: 
``` csharp
if (vType is TextView) {
	top.NewKeyDownEvent (new (KeyCode.Tab | KeyCode.CtrlMask));
} else if (vType is DatePicker) {
	for (int i = 0; i < 4; i++) {
		top.NewKeyDownEvent (new (KeyCode.Tab | KeyCode.CtrlMask));
	}
} else {
	top.NewKeyDownEvent (new (KeyCode.Tab));
}
```

BDisp:
> This works, so I'll go with that:
> 
> ```cs
> if (vType is TextView) {
> 	top.NewKeyDownEvent (new (KeyCode.Tab | KeyCode.CtrlMask));
> } else if (vType is DatePicker) {
> 	for (int i = 0; i < 4; i++) {
> 		top.NewKeyDownEvent (new (KeyCode.Tab | KeyCode.CtrlMask));
> 	}
> } else {
> 	top.NewKeyDownEvent (new (KeyCode.Tab));
> }
> ```

Great, do the same to the `TextView` because I think is the same keybinding.

MaciekWin3:
Now it failed on something unrealated to this PR: 
```
 Failed Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor [32 ms]
  Error Message:
   System.Exception : Unexpected color White,Black was used at row 0 and col 0 (indexes start at 0).  Color value was White,Black (expected colors were -1,0)
  Stack Trace:
     at Terminal.Gui.TestHelpers.AssertDriverColorsAre(String expectedLook, ConsoleDriver driver, Attribute[] expectedColors) in D:\a\Terminal.Gui\Terminal.Gui\UnitTests\TestHelpers.cs:line 337
   at Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor() in D:\a\Terminal.Gui\Terminal.Gui\UnitTests\Views\TreeViewTests.cs:line 868
  Passed Terminal.Gui.ViewsTests.TreeViewTests.ContentWidth_BiggerAfterExpand [50 ms]
```

BDisp:
> Now it failed on something unrealated to this PR:
> 
> ```
>  Failed Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor [32 ms]
>   Error Message:
>    System.Exception : Unexpected color White,Black was used at row 0 and col 0 (indexes start at 0).  Color value was White,Black (expected colors were -1,0)
>   Stack Trace:
>      at Terminal.Gui.TestHelpers.AssertDriverColorsAre(String expectedLook, ConsoleDriver driver, Attribute[] expectedColors) in D:\a\Terminal.Gui\Terminal.Gui\UnitTests\TestHelpers.cs:line 337
>    at Terminal.Gui.ViewsTests.TreeViewTests.TestTreeViewColor() in D:\a\Terminal.Gui\Terminal.Gui\UnitTests\Views\TreeViewTests.cs:line 868
>    at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
>    at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
>   Passed Terminal.Gui.ViewsTests.TreeViewTests.ContentWidth_BiggerAfterExpand [50 ms]
> ```

Ignore it, sometime it fails.

BDisp:
> Ignore it, sometime it fails.

@tig I only was addressing to this PR because as @MaciekWin3 said the fail isn't related with this PR.

Issue #2491, opened 2023-04-04
Get rid of `Toplevel` - Introduce `Runnable` and `Overlapped` instead


tig:
# This PR is for a major rehaul of how Terminal.Gui deals with non-Tiled view arrangements and "runnable" views.

@bdisp, @tznind, @dodexahedron, and @migueldeicaza - PLEASE READ AND TURN ON YOUR BRAINSTORM MODE

We have a chance to get the design of how all this works in v2 right. We should approach this throwing away existing notions of how Terminal.Gui has always worked in these regards. This does not mean we WILL throw all those things away, but that we should not constrain our thinking at this point.

I've written the below in the spirit of generating debate and ideas. I'm highly motivated to start working on this as I've been noodling on it for years. But my thinking is still very far from clear.

Related Issues:

- #2994
- #2537  

## PRs That Will Partially Address

- #3634 
- #3627 

## Background 

In v2, we have the following layout abilities:

* **Arrangement** - Includes `View.Arrangement` and code in `Border.cs` for determining how a view (via `Border`) can be moved and sized. Right now only `ViewArrangement.Fixed` and `ViewArrangement.Movable` have any implementation.
* **Tiled** - By default, there is no 'z-order` to the Subviews of a View. And in mose use-cases, subviews do not overlap with each other (the exception being when it's done intentionally to create some effect). As a result, the default layout for most TUI apps is "tiled". 
* **Modal** - A modal view is one that is run as an "application" via `Application.Run(Toplevel)` where `Modal == true`. `Dialog`, `Messagebox`, and `Wizard` are the prototypical examples. When run this way, there IS a `z-order` but it is highly-constrained: the modal view has a z-order of 1 and everything else is at 0. 
* **Runnable** - Today, Overlapped and Runnable are intrinsically linked. A runnable view is one where `Application.Run(Toplevel)` is called.  Each *Runnable* view where (`Modal == false`) has it's own `RunState` and is, effectively, a self-contained "application". `Application.Run()` non-preemptively dispatches events (screen, keyboard, mouse , Timers, and Idle handlers) to the associated `Toplevel`. It is possible for such a `Toplevel` to create a thread and call `Application.Run(someotherToplevel)` on that thread, enabling pre-emptive user-interface multitasking (`BackgroundWorkerCollection` does this). 
[...]

PR #3151, opened 2024-01-09
Fixes #3072, #3148. `Dialog` and `MessageBox` key binding do not handle Enter correctly

### LINKED ISSUE #3072: IsDefault on Button is bugged in v2 and has priority over other Buttons
**Describe the bug**
If a View has 2 buttons, one of which is IsDefault then pressing Enter will always select that one regardless of which has input focus

**To Reproduce**
1. Tab to the 'non default' button
2. Press Enter

**Expected behavior**
Click handler on the focused button to be called

**Observed behavior**
The default button handler is invoked.

**Screenshots**
bad-defaults

**Code Repro**

```csharp
Application.Init ();

var d = new Window ();
d.Title = "Blah";
var b1 = new Button ("Button 1 (Default)",true);
b1.Clicked += (s, e) => { 
	Application.RequestStop ();
	MessageBox.Query ("Closed", "ClosedWithButton1", "Ok");
};
var b2 = new Button ("Button 2") {
	Y = 1
};
b2.Clicked += (s, e) => { 
	Application.RequestStop ();
	MessageBox.Query ("Closed", "ClosedWithButton2","Ok");
};

d.Add (b1,b2);

Application.Run (d);	
Application.Shutdown ();
```

**Additional context**

### LINKED ISSUE #3148: MessageBox v2 Returns wrong button when using keyboard navigation
When navigating across buttons with the keyboard, and then hitting enter for 'accept', it always returns the default button value.

```CSharp
using Terminal.Gui;

Application.Init();

var mainWindow = new Window() { Title = "Test", X = 0, Y = 0, Width = Dim.Fill(), Height = Dim.Fill() };
var label = new Label() { X = 0, Y = 0, Text = "-" };
var button = new Button() { X = 0, Y = 2, Text = "Quit", IsDefault = true };
mainWindow.Add(label, button);

button.Clicked += (object sender, System.EventArgs e) => {
	label.Text = "";
	var r = MessageBox.Query(50, 7, "Quit?", "Are you sure you want to quit?", 0, "Yes", "No", "Maybe");
	label.Text = $"{r} = MessageBox.Query()";
};

Application.Top.Add(mainWindow);
Application.Run();
```
Would expect it to behave the same way as a WinForm application and return the button selected when the enter key is hit

MessageBox

tig:
## Fixes

- Fixes #3072
- Fixes #3148 

## Todo:

- [x] Fix `Button`'s key bindings
- [x] Move `Enter` key binding from `Toplevel` to `Dialog`
- [x] Add better unit tests

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/Button.cs
+++ b/Terminal.Gui/Views/Button.cs
@@ -122,8 +122,15 @@ public class Button : View {
 
 		// Override default behavior of View
 		// Command.Default sets focus
-		AddCommand (Command.Accept, () => { OnClicked (); return true; });
+		AddCommand (Command.Accept, () => AcceptKey ());
 		KeyBindings.Add (Key.Space, Command.Default, Command.Accept);
+		KeyBindings.Add (Key.Enter, Command.Default, Command.Accept);
+	}
+
+	bool AcceptKey ()
+	{
+		OnClicked ();
+		return true;
 	}
 
 	/// <summary>
@@ -166,14 +173,6 @@ public class Button : View {
 		}
 	}
 
-	bool AcceptKey ()
-	{
-		//if (!HasFocus) {
-		//	SetFocus ();
-		//}
-		OnClicked ();
-		return true;
-	}
 
 	/// <summary>
 	/// Virtual method to invoke the <see cref="Clicked"/> event.
--- a/Terminal.Gui/Views/Button.cs
+++ b/Terminal.Gui/Views/Button.cs
@@ -120,19 +120,12 @@ public class Button : View {
 
 		OnResizeNeeded ();
 
-		// Override default behavior of View
-		// Command.Default sets focus
-		AddCommand (Command.Accept, () => AcceptKey ());
+		// Override default behavior of View - Command.Default sets focus
+		AddCommand (Command.Accept, () => { OnClicked (); return true; });
 		KeyBindings.Add (Key.Space, Command.Default, Command.Accept);
 		KeyBindings.Add (Key.Enter, Command.Default, Command.Accept);
 	}
-
-	bool AcceptKey ()
-	{
-		OnClicked ();
-		return true;
-	}
-
+	
 	/// <summary>
 	/// Gets or sets whether the <see cref="Button"/> is the default action to activate in a dialog.
 	/// </summary>
--- a/Terminal.Gui/Views/Dialog.cs
+++ b/Terminal.Gui/Views/Dialog.cs
@@ -79,6 +79,8 @@ public class Dialog : Window {
 			return false;
 		});
 
+
+		KeyBindings.Add (KeyCode.Esc, Command.QuitToplevel);
 		// This enables the default button to be activated by the Enter key.
 		KeyBindings.Add (KeyCode.Enter, Command.Accept);
 
@@ -240,17 +242,5 @@ public class Dialog : Window {
 			}
 			break;
 		}
-	}
-
-	// BUGBUG: Why is this not handled by a key binding???
-	///<inheritdoc/>
-	public override bool OnProcessKeyDown (Key a)
-	{
-		switch (a.KeyCode) {
-		case KeyCode.Esc:
-			Application.RequestStop (this);
-			return true;
-		}
-		return false;
 	}
 }
--- a/Terminal.Gui/Views/Dialog.cs
+++ b/Terminal.Gui/Views/Dialog.cs
@@ -70,18 +70,7 @@ public class Dialog : Window {
 		Modal = true;
 		ButtonAlignment = DefaultButtonAlignment;
 
-		// This enables the default button to be activated by the Enter key.
-		AddCommand (Command.Accept, () => {
-			// TODO: Perhaps all views should support the concept of being default?
-			if (Subviews.FirstOrDefault (v => v is Button && ((Button)v).IsDefault && ((Button)v).Enabled) is Button defaultBtn) {
-				defaultBtn.InvokeCommand (Command.Accept);
-				return true;
-			}
-			return false;
-		});
-
-		KeyBindings.Add (KeyCode.Esc, Command.QuitToplevel);
-		KeyBindings.Add (KeyCode.Enter, Command.Accept);
+		KeyBindings.Add (Key.Esc, Command.QuitToplevel);
 
 		if (buttons != null) {
 			foreach (var b in buttons) {
--- a/Terminal.Gui/Views/Dialog.cs
+++ b/Terminal.Gui/Views/Dialog.cs
@@ -70,6 +70,7 @@ public class Dialog : Window {
 		Modal = true;
 		ButtonAlignment = DefaultButtonAlignment;
 
+		// This enables the default button to be activated by the Enter key.
 		AddCommand (Command.Accept, () => {
 			// TODO: Perhaps all views should support the concept of being default?
 			if (Subviews.FirstOrDefault (v => v is Button && ((Button)v).IsDefault && ((Button)v).Enabled) is Button defaultBtn) {
@@ -79,9 +80,7 @@ public class Dialog : Window {
 			return false;
 		});
 
-
 		KeyBindings.Add (KeyCode.Esc, Command.QuitToplevel);
-		// This enables the default button to be activated by the Enter key.
 		KeyBindings.Add (KeyCode.Enter, Command.Accept);
 
 		if (buttons != null) {
--- a/Terminal.Gui/Views/Dialog.cs
+++ b/Terminal.Gui/Views/Dialog.cs
@@ -70,6 +70,18 @@ public class Dialog : Window {
 		Modal = true;
 		ButtonAlignment = DefaultButtonAlignment;
 
+		AddCommand (Command.Accept, () => {
+			// TODO: Perhaps all views should support the concept of being default?
+			if (Subviews.FirstOrDefault (v => v is Button && ((Button)v).IsDefault && ((Button)v).Enabled) is Button defaultBtn) {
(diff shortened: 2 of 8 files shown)


BDisp:
`Dialog` is a sub-class of `Window`, which is a sub-class of `Toplevel` and thus if the focused view doesn't handle the key binding the `Toplevel` must be responsible to send the key binding for a sub-view that need it. If none of the sub-views handle it then the `Toplevel` checks if need it and if it doesn't handled return false. I think this is duplicating the same code through sub-class of `Toplevel`. But I may be wrong.

tig:
> `Dialog` is a sub-class of `Window`, which is a sub-class of `Toplevel` and thus if the focused view doesn't handle the key binding the `Toplevel` must be responsible to send the key binding for a sub-view that need it. If none of the sub-views handle it then the `Toplevel` checks if need it and if it doesn't handled return false. I think this is duplicating the same code through sub-class of `Toplevel`. But I may be wrong.

Please look at the new unit tests for MessageBox and Dialog in this PR and tell me why they're wrong.

I think you're just describing how `OnInvokingKeyBindings` now works.

BDisp:
@tig I achieved the same result by only done 2 changes:

On Dialog.cs remove the follow as you already done because it never run.

```cs
	// BUGBUG: Why is this not handled by a key binding???
	///<inheritdoc/>
	public override bool OnProcessKeyDown (Key a)
	{
		switch (a.KeyCode) {
		case KeyCode.Esc:
			Application.RequestStop (this);
			return true;
		}
		return false;
	}

```

In the Toplevel.cs add this in the line 229 before `if (Subviews.FirstOrDefault (v => v is Button && ((Button)v).IsDefault && ((Button)v).Enabled) is Button defaultBtn) {`

```cs
			// First invoke the focused view
			if (Focused?.InvokeCommand (Command.Accept) == true) {
				return true;
			}

```

This allow the key binding run first for the focused view.

tig:
> @tig I achieved the same result by only done 2 changes on the v2_develop branch:
> 
> On Dialog.cs remove the follow as you already done because it never run and the Esc key is already handled on the Toplevel class.
> 
> ```cs
> 	// BUGBUG: Why is this not handled by a key binding???
> 	///<inheritdoc/>
> 	public override bool OnProcessKeyDown (Key a)
> 	{
> 		switch (a.KeyCode) {
> 		case KeyCode.Esc:
> 			Application.RequestStop (this);
> 			return true;
> 		}
> 		return false;
> 	}
> ```
> 
> In the Toplevel.cs add this in the line 229 before `if (Subviews.FirstOrDefault (v => v is Button && ((Button)v).IsDefault && ((Button)v).Enabled) is Button defaultBtn) {`
> 
> ```cs
> 			// First invoke the focused view
> 			if (Focused?.InvokeCommand (Command.Accept) == true) {
> 				return true;
> 			}
> ```
> 
> This allow the key binding run first for the focused view.

Nice!

Do you agree there's no real reason for `Toplevel` to have a keybinding for `Enter` is a bad idea? 

Does your change pass the new unit tests I provided?

tig:
> the Esc key is already handled on the Toplevel class.

This is not accurate. The `Application.QuitKey` key is handled on the Toplevel class. 

I do NOT think the key to dismiss a `DialogBox` or `MessageBox` should be anything other than `Esc`. I DO think the quit key for an APPLICATION should be configurable. The default for `Application.QuitKey` in v2 right now is `Ctrl.Q`. I believe it should be changed to `Esc`, but that's a different topic.

tig:
> This allow the key binding run first for the focused view.

Actually, THIS is what allows Key Bindings (in general) to run first for the focused view:

image

What you propose for `Toplevel` is a workaround for the fact that `Toplevel` is adding a key binding for `Enter` (incorrectly, in my opinion).

BDisp:
> Nice!
> 
> Do you agree there's no real reason for `Toplevel` to have a keybinding for `Enter` is a bad idea?
> 
I think the Toplevel should have a keybinding for `Enter` key because it and the `Window` can handle the default buttons in the same way as Dialog. Remember that WinForm behave the same way.
But I have another solution without the Toplevel change, which is doing the change on the ViewKeyboard.cs on the OnInvokingKeyBindings method:

```cs
	public virtual bool? OnInvokingKeyBindings (Key keyEvent)
	{
		// fire event
		// BUGBUG: KeyEventArgs doesn't include scope, so the event never sees it.
		bool? handled = null;
		if (HasFocus && keyEvent.Scope == KeyBindingScope.Focused) {
			if (KeyBindings.TryGet (keyEvent.KeyCode, KeyBindingScope.Focused, out var binding)) {
				foreach (var kb in binding.Commands) {
					if (InvokeCommand (kb) == true) {
						handled = true;
					}
				}
				if (handled == true) {
					return true;
				}
			}
		}
		if (keyEvent.Scope == KeyBindingScope.Application || keyEvent.Scope == KeyBindingScope.HotKey) {
			InvokingKeyBindings?.Invoke (this, keyEvent);
			if (keyEvent.Handled) {
				return true;
			}
		}

		// * If no key binding was found, `InvokeKeyBindings` returns `null`.
		//   Continue passing the event (return `false` from `OnInvokeKeyBindings`).
		// * If key bindings were found, but none handled the key (all `Command`s returned `false`),
		//   `InvokeKeyBindings` returns `false`. Continue passing the event (return `false` from `OnInvokeKeyBindings`)..
		// * If key bindings were found, and any handled the key (at least one `Command` returned `true`),
		//   `InvokeKeyBindings` returns `true`. Continue passing the event (return `false` from `OnInvokeKeyBindings`).
		handled = InvokeKeyBindings (keyEvent);
		if (handled != null && (bool)handled) {
			// Stop processing if any key binding handled the key.
			// DO NOT stop processing if there are no matching key bindings or none of the key bindings handled the key
			return true;
		}

		// Now, process any key bindings in the subviews that are tagged to KeyBindingScope.HotKey.
		foreach (var view in Subviews.Where (v => v.KeyBindings.TryGet (keyEvent.KeyCode, KeyBindingScope.HotKey, out var _))) {
			// TODO: I think this TryGet is not needed due to the one in the lambda above. Use `Get` instead?
			if (view.KeyBindings.TryGet (keyEvent.KeyCode, KeyBindingScope.HotKey, out var binding)) {
				keyEvent.Scope = KeyBindingScope.HotKey;
				handled = view.OnInvokingKeyBindings (keyEvent);
				if (handled != null && (bool)handled) {
					return true;
				}
			}
		}

		return handled;
	}

```

> Does your change pass the new unit tests I provided?

We really need to add a key binding for the `Enter` key in the Button.cs as you did.
`KeyBindings.Add (Key.Enter, Command.Default, Command.Accept);`

BDisp:
> > the Esc key is already handled on the Toplevel class.
> 
> This is not accurate. The `Application.QuitKey` key is handled on the Toplevel class.
> 
> I do NOT think the key to dismiss a `DialogBox` or `MessageBox` should be anything other than `Esc`. I DO think the quit key for an APPLICATION should be configurable. The default for `Application.QuitKey` in v2 right now is `Ctrl.Q` (UI Catalog overrides this). I believe it should be changed to `Esc`, but that's a different topic.

I agree. In that case you can add a key binding for the `Esc` key in the Dialog.cs.

BDisp:
> > This allow the key binding run first for the focused view.
> 
> Actually, THIS is what allows Key Bindings (in general) to run first for the focused view:
> 
> image
> 
> What you propose for `Toplevel` is a workaround for the fact that `Toplevel` is adding a key binding for `Enter` (incorrectly, in my opinion).

But for that we need to override the method on Button.cs and there is need because the View class already handle that.

tig:
> @tig to prove what I meaning run this test on this PR and you'll see that it fail for the `Toplevel` and `Window`. That why I think the button has to behave on the same way with all containers and not only for the `Dialog`.

I totally dig seeing this example of TheoryData!

You are correct about `Window`.

I've said before that I think `Toplevel` is a bad design and should be removed. https://github.com/gui-cs/Terminal.Gui/issues/2491

My thinking has changed some since I wrote that, but not the parts about `Toplevel` being overly complex and poorly engineered and needing to be nuked from orbit. We should be actively removing things from `Toplevel` and really, really, pause anytime we consider adding something to it.

Therefore, a compromise:

```cs
		public static TheoryData<Toplevel> ButtonContainers =>
			new () {
				{ new Window () },
				{ new Dialog () }
			};
```

Cool?

tig:
> > What you propose for `Toplevel` is a workaround for the fact that `Toplevel` is adding a key binding for `Enter` (incorrectly, in my opinion).
> 
> But for that we need to override the method on Button.cs and there is need because the View class already handle that.

Not sure what you mean by this. What method are you talking about? Is there a unit test missing from this PR that shows that something is broken?

BDisp:
> I totally dig seeing this example of TheoryData!
> 
> You are correct about `Window`.
> 
> I've said before that I think `Toplevel` is a bad design and should be removed. #2491
> 
> My thinking has changed some since I wrote that, but not the parts about `Toplevel` being overly complex and poorly engineered and needing to be nuked from orbit. We should be actively removing things from `Toplevel` and really, really, pause anytime we consider adding something to it.
> 

So, you want to get rid of the `Toplevel` and thus the `Window` will be the most top container?

> Therefore, a compromise:
> 
> ```cs
> 		public static TheoryData<Toplevel> ButtonContainers =>
> 			new () {
> 				{ new Window () },
> 				{ new Dialog () }
> 			};
> ```
> 
> Cool?

In this case it's cool.

tig:
> So, you want to get rid of the `Toplevel` and thus the `Window` will be the most top container?

Let's discuss here: https://github.com/gui-cs/Terminal.Gui/issues/2491

Issue #2994, opened 2023-11-15
Master Issue: Finish Adornments 


tig:
When I was designing `Frames` I had the idea that `View`'s `Margin`, `Border`, and `Padding` could be replaced by a View. I didn't complete the implementation of this, but did make `View.CreateFrames` `virtual`:

See Sub-issues.

tig:
@BDisp which part of this do you feel like tackling first? 

If it were me, I'd do it in this order (each as a separate PR):

1) Implement `class Margin : Frame`,  `class Border : Frame`,  `class Padding: Frame`, getting rid of the `Id == "Border"` hacks.
2) #2995 - Figure out why auto-join is not always working is essential. 
3) #2814 - Use this to flush out how Mouse/Keyboard/Focus work for the current `Border` implementation 
4) Re-factor TabView to use all of this.
5) #2489
6) ...

But I'm happy to let you approach it how you see fit.

BDisp:
My doubt is if you add views or adornments on the `Margin`, `Border` and `Padding` is if the `LineCanvas` should render all of them. There will be situations where you don't need/want to do that. For now I'm only think on them to use colors and on the `Border` frame to draw border. My head is empty for now :-)

tig:
A thought: When we're done with all of this, instead of having 3 adornments as individual members of `View`:

```cs
		// TODO: Make this cancelable
		/// <summary>
		/// Prepares <see cref="View.LineCanvas"/>. If <see cref="SuperViewRendersLineCanvas"/> is true, only the <see cref="LineCanvas"/> of 
		/// this view's subviews will be rendered. If <see cref="SuperViewRendersLineCanvas"/> is false (the default), this 
		/// method will cause the <see cref="LineCanvas"/> be prepared to be rendered.
		/// </summary>
		/// <returns></returns>
		public virtual bool OnDrawFrames ()
		{
			if (!IsInitialized) {
				return false;
			}

			// Each of these renders lines to either this View's LineCanvas 
			// Those lines will be finally rendered in OnRenderLineCanvas
			Margin?.OnDrawContent (Margin.Bounds);
			Border?.OnDrawContent (Border.Bounds);
			Padding?.OnDrawContent (Padding.Bounds);

			return true;
		}
```

We would have a `List<Frame>` member:

```cs
		// TODO: Make this cancelable
		/// <summary>
[...]

PR #3158, opened 2024-01-11
Renames `Frame`->`Adornment`; changes `Frame` to have a `Border` subclass

tig:
See this discussion: https://github.com/gui-cs/Terminal.Gui/issues/2994#issuecomment-1887861731

- [x] Create `Margin` subclass of `Frame`
- [x] Create `Border` subclass of `Frame`
- [x] Create `Padding` subclass of `Frame`
- [x] Improve robustness of Adornment creation
- [x] `Border` line style inherits from `SuperView` (if not set explicitly).
- [x] Update/tweak `Frames` scenario
- [x] `Frame`->`Adornment`
- [x] `Frames` scenario etc... rename 
- [x] More/better `Adornment` unit tests


tig:
@BDisp take a look! Just a WIP, but...

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -454,7 +454,7 @@ public virtual void OnDrawContent (Rect contentArea)
 	{
 		if (NeedsDisplay) {
 			if (SuperView != null) {
-				Clear (BoundsToScreen (Bounds));
+				Clear (BoundsToScreen (contentArea));
```
BDisp, line 457: `contentArea` is the same as `Bounds` but only screen relative, right?

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -191,15 +191,15 @@ void ThicknessChangedHandler (object sender, EventArgs e)
 			Margin.ThicknessChanged -= ThicknessChangedHandler;
 			Margin.Dispose ();
 		}
-		Margin = new Frame () { Id = "Margin", Thickness = new Thickness (0) };
+		Margin = new Frame () { Thickness = new Thickness (0) };
```
BDisp, line 194: In the future if `Margin` has a different behavior than `Frame` then you will to sub-class it.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -209,7 +209,7 @@ void ThicknessChangedHandler (object sender, EventArgs e)
 			Padding.ThicknessChanged -= ThicknessChangedHandler;
 			Padding.Dispose ();
 		}
-		Padding = new Frame () { Id = "Padding", Thickness = new Thickness (0) };
+		Padding = new Frame () { Thickness = new Thickness (0) };
```
BDisp, line 212: In the future if `Padding` has a different behavior than `Frame` then you will to sub-class it.

Terminal.Gui/View/ViewDrawing.cs
tig: Yes. I was going to remove that param from OnDrawContent but now I'm glad i never did. 

You'll see why soon...

Terminal.Gui/View/Layout/ViewLayout.cs
tig: Yessir. 

I already have one difference: Margin uses SuperView color scheme. Border and padding uses View.

tig:
I started in on adding subviews to border. It's complicated so I'm not going to do it in this pr. I'll neaten this one up a bit more and merge it and do another for that.

BDisp:
> I started in on adding subviews to border. It's complicated so I'm not going to do it in this pr. I'll neaten this one up a bit more and merge it and do another for that.

Ok, I also trying to use the `Padding` of the `TabView` to handle the tan rows and it isn't really easy because all the login for handles with views and sub-views are in View class itself. But I'll continue trying.
Why not change the `Frame` class name to `Adornment`. The sub-views that are added to them are still `View` sub-classes. Thus will not make confusing with the `Rect Frame`.

tig:
On the padding thing: it'd be great if you opened an Issue on that. I'd love to dig into it with you with focus (and a simple example, not TabView). 

On name: I think you know I've been wishy-washy on Frame vs Adornment as the name. Your comment is helpful. I'll strongly consider it!

BDisp:
> On the padding thing: it'd be great if you opened an Issue on that. I'd love to dig into it with you with focus (and a simple example, not TabView).
> 

I was trying in this PR #2980 but didn't commit/push nothing addressed with `Padding`. So, you don't want I continue on here?

> On name: I think you know I've been wishy-washy on Frame vs Adornment as the name. Your comment is helpful. I'll strongly consider it!

Ok at least I did you reflect about it 😃

tig:
> > On the padding thing: it'd be great if you opened an Issue on that. I'd love to dig into it with you with focus (and a simple example, not TabView).
> 
> I was trying in this PR #2980 but didn't commit/push nothing addressed with `Padding`. So, you don't want I continue on here?

When working on architecture, I'm a fan of starting with just unit tests and a very simple Scenario vs. a major change like what needs to happen to TabView. TabView is a pretty complex beast, so yes.

BDisp:
> > On name: I think you know I've been wishy-washy on Frame vs Adornment as the name. Your comment is helpful. I'll strongly consider it!
> 
> Ok at least I did you reflect about it 😃

I really appreciated for you have reconsider to change from `Frame` to `Adornment`. Thanks.

BDisp:
@tig how about rename the `Bounds` to `ContentArea` as it's used on some of the `View` methods?

tig:
> @tig how about rename the `Bounds` to `ContentArea` as it's used on some of the `View` methods?

- https://github.com/gui-cs/Terminal.Gui/issues/3169

PR #3165, opened 2024-01-13
Fixes #3157. DateField is validating wrong the date on different culture than us-US.

BDisp:
Fixes #3157 - Include the fix for the `TextField` issue #3161.

## Pull Request checklist:

- [x] I've named my PR in the form of "Fixes #issue. Terse description."
- [x] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [x] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [x] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [x] My changes generate no new warnings
- [x] I have checked my code and corrected any poor grammar or misspellings
- [x] I conducted basic QA to assure all features are working

PR #3167, opened 2024-01-13
Fixes #3166. TimeField should use TextChanging instead of TextChanged event.

### LINKED ISSUE #3166: TimeField should use TextChanging instead of TextChanged event.

BDisp:
Fixes #3166 - Thus avoid any `StackOverflowException` in future.

## Pull Request checklist:


BDisp:
@tig this seems the unit tests are passed but it isn't true, because there silents errors which unfortunately aren't addressed as errors but only appear on the Output window from Tests combo.

image

image

BDisp:
This is the final line of the output window:

`========== Test run aborted: 0 Tests (0 Passed, 0 Failed, 0 Skipped) run in < 1 ms ==========`

BDisp:
What is really strange is the unit tests doesn't fail on WSL  but only on Windows. Do you realize why @tig?

BDisp:
Here the prove that's fail on `Local Windows Environment`:

image

image

image

Here the prove that isn't fail on `WSL`:

image

image

image

The cause is a bug in the `DateField` that was fixed in the #3165.
In WSL doesn't fail because I use the US locale `date` ("01/14/0024") as the github server and in `Windows` I use the PT locale `date` ("14/01/0024") thus it fail.
So for the US users isn't a problem but for others different locales it will always fail.

PR #3181, opened 2024-01-15
Fixes #3171. Remove View and subclass constructors with parameters.

### LINKED ISSUE #3171: Remove `View` and subclass constructors with parameters
I thought we already had an issue for this, but I can't find it. 

We agreed long ago, for various good reasons, to rid ourselves of constructors that take parameters.

With the merging of 

- #3130 

...the biggest reason they existed is now moot: Absolute layout is no longer determined by how the View was constructed.

Anytime you're doing a PR and see the opportunity to delete one of these, please do so. Some will be harder than others. In all cases it is also an opportunity to double-down on unit tests that test initialization.

Also, any methods called from constructors that are currently named `Initialize` should be renamed to `SetInitialProperties` to avoid confusion with `Begin/EndInit`.

## Todo

- [ ] View
- [ ] Button
- [ ] CheckBox
- [ ] ComboBox
- [ ] ColorPicker - Will fix in #2800 
- [x] DateField - Thank you @BDisp!
- [ ] FrameView
- [ ] Label 
- [ ] ListView
- [ ] ScrollBarView
- [ ] ScrollView
- [ ] ContextMenu
- [ ] OpenDialog
- [ ] SaveDialog
- [ ] TextField 
- [ ] TimeField 
- [ ] Toplevel
- [ ] Window

BDisp:
## Fixes

- Fixes #3171

## Proposed Changes/Todos

* [x] View
* [x] Button
* [x] CheckBox
* [x] ComboBox
* [x] FrameView
* [x] Label
* [x] ListView
* [x] ScrollBarView
* [x] ScrollView
* [x] ContextMenu
* [x] Dialog
* [x] OpenDialog
* [x] SaveDialog
* [ ] FileDialog - Parameter can't be removed because there isn't a public `_fileSystem` field and it's only for testing.
* [x] TextField
* [x] TimeField
* [x] Toplevel
* [x] Window
* [x] RadioGroup
* [x] TextView
* [x] MenuBar
* [x] Menu
* [x] TextValidateField

## Changes checklist:

* [x] Rename `Initialize` to `SetInitialProperties`
* [x] Remove `namespace` braces
* [x] Prefix private fields with underscore
* [x] Remove unused `using`
* [x] Remove parenthesis on objects initializers

## Pull Request checklist:


BDisp:
@tig this is ready for review. Thanks.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -560,7 +560,7 @@ internal void OnResizeNeeded ()
 		var relativeBounds = SuperView is { IsInitialized: true } ? SuperView.Bounds :
 			Application.Top != null && Application.Top.IsInitialized ? Application.Top.Bounds :
 										   Application.Driver?.Bounds ??
-										   new Rect (0, 0, int.MaxValue, int.MaxValue);
+										   Rect.Empty;
```
tig, line 563: Please leave this as it was. This will be important for `Dim.Auto`.

Terminal.Gui/Views/Button.cs
```diff
@@ -61,37 +61,6 @@ public class Button : View {
 	/// </param>
 	public Button (string text, bool is_default = false) : base (text) => SetInitialProperties (text, is_default);
```
tig, line 62: Why can't we get rid of this one too?

Terminal.Gui/Views/FrameView.cs
```diff
@@ -1,68 +1,61 @@
...
-		/// Initializes a new instance of the <see cref="Gui.FrameView"/> class using <see cref="LayoutStyle.Computed"/> layout.
-		/// </summary>
-		/// <param name="title">Title.</param>
-		public FrameView (string title)
-		{
-			SetInitialProperties (Rect.Empty, title, null);
-		}
+	/// <param name="title">Title.</param>
+	/// <param name="views">Views.</param>
+	public FrameView (string title = null, View [] views = null)
```
tig, line 15: Why not nuke this one as well?

It's easy enough to do

```cs
var f = new FrameView() { Title = "Title" };
f.Add(views);
```

Terminal.Gui/Views/FrameView.cs
```diff
@@ -1,68 +1,61 @@
...
 
-		/// <summary>
-		/// Initializes a new instance of the <see cref="Gui.FrameView"/> class using <see cref="LayoutStyle.Computed"/> layout.
-		/// </summary>
-		public FrameView () : this (title: string.Empty) {
+	/// <summary>
+	/// Initializes a new instance of the <see cref="Gui.FrameView"/> class using <see cref="LayoutStyle.Computed"/> layout.
+	/// </summary>
+	/// <param name="title">Title.</param>
+	public FrameView (string title)
```
tig, line 24: Same question as above.

Terminal.Gui/Views/Label.cs
```diff
@@ -21,19 +21,13 @@ public class Label : View {
 	/// <inheritdoc/>
 	public Label (string text, bool autosize = true) : base (text) => SetInitialProperties (autosize);
```
tig, line 22: I'm seeing you didn't remove the `text` constructors. Am I missing a reason why they shouldn't be removed as well?

It's easy enough to do:

```cs
var l = new Label() { Text = "text", AutoSize = false }
```

Terminal.Gui/Views/Menu/MenuBar.cs
```diff
@@ -21,7 +20,7 @@ public class MenuBarItem : MenuItem {
 	/// <param name="parent">The parent <see cref="MenuItem"/> of this if exist, otherwise is null.</param>
 	public MenuBarItem (string title, string help, Action action, Func<bool> canExecute = null, MenuItem parent = null) : base (title, help, action, canExecute, parent)
```
tig, line 21: I wouldn't waste time on MenuBar.  ;-)

tig (changes requested):
See comments.

FWIW, I don't think we have to remove them ALL at once, but I DO think that before we release V2 *ALL* View constructors with parameters should be gone unless there's a really, really good reason for it.

BDisp:
> FWIW, I don't think we have to remove them ALL at once, but I DO think that before we release V2 _ALL_ View constructors with parameters should be gone unless there's a really, really good reason for it.

I thought you only want get rid those with frame parameters. Now it's more clear and only the readonly properties that only be set through a parameter passed on the constructor will be used. I'll procced then.

Terminal.Gui/Views/Button.cs
BDisp: It will too.

Terminal.Gui/Views/Label.cs
BDisp: I now realize you want get rid all the parameters and I'll fix it.

Terminal.Gui/Views/Menu/MenuBar.cs
BDisp: We'll see... 😄

Terminal.Gui/View/Layout/ViewLayout.cs
BDisp: Done.

tig:
> > FWIW, I don't think we have to remove them ALL at once, but I DO think that before we release V2 _ALL_ View constructors with parameters should be gone unless there's a really, really good reason for it.
> 
> I thought you only want get rid those with frame parameters. Now it's more clear and only the readonly properties that only be set through a parameter passed on the constructor will be used. I'll procced then.

For "readonly" parameters, `init` or `priviate init` can be used.

BDisp:
@tig I need your confirmation asap for the cases where all properties that can be set after object initialization, should be removed from the constructors parameters.
The `ComboBox` source cant' be set on object initializer, but only through constructor parameter or after the object initialization. Thus the `SelectedItem` can't also be set on object initializer, but only after the object initialization.
So my question is what you prefer of the follow two unit tests. Your answer must also be valid for `ListView`, `RadioGroup` and others properties that can be set after the object initialization. Thanks.

After the object initialization:

```cs
	[Fact, AutoInitShutdown]
	public void SelectedItem_Must_Be_Set_After_Set_The_Source ()
	{
		var cb = new ComboBox ();
		cb.SetSource (new List<string> () { "One", "Two", "Three" });
		cb.SelectedItem = 1;
		cb.BeginInit ();
		cb.EndInit ();
		cb.LayoutSubviews ();
		Assert.Equal ("Two", cb.Text);
		Assert.NotNull (cb.Source);
		Assert.False (cb.AutoSize);
		Assert.Equal (new Rect (0, 0, 0, 2), cb.Frame);
		Assert.Equal (1, cb.SelectedItem);
	}

```

On object initializer using the source parameter:

```cs
	[Fact, AutoInitShutdown]
	public void Constructor_With_Source_Initialize_With_The_Passed_SelectedItem ()
	{
		var cb = new ComboBox (new List<string> () { "One", "Two", "Three" }) {
			SelectedItem = 1
		};
		cb.BeginInit ();
		cb.EndInit ();
		cb.LayoutSubviews ();
		Assert.Equal ("Two", cb.Text);
		Assert.NotNull (cb.Source);
		Assert.False (cb.AutoSize);
		Assert.Equal (new Rect (0, 0, 0, 2), cb.Frame);
		Assert.Equal (1, cb.SelectedItem);
	}

```

tig:
What's wrong with this?

```cs
[Fact]
[AutoInitShutdown]
public void EnsureKeyEventsDoNotCauseExceptions ()
{
	var comboBox = new ComboBox () {
		Width = 0,
		Source = ListView.MakeWrapper(Enumerable.Range (0, 15).Select (x => x.ToString ()).ToArray ()),
		SelectedItem = 1
	};


	Application.Top.Add (comboBox);

	foreach (var key in (KeyCode [])Enum.GetValues (typeof (KeyCode))) {
		Assert.Null (Record.Exception (() => comboBox.NewKeyDownEvent (new (key))));
	}
}
```

(had to make ListView.MakeWapper public to do this).

BDisp:
> What's wrong with this?

> (had to make ListView.MakeWapper public to do this).

Ops I didn't remember that anymore. Thanks @tig.

BDisp:
> (had to make ListView.MakeWapper public to do this).

Not needed @tig. For that the `ListWrapper` class is already public. The static `ListView.MakeWapper` is only used on the `ListView` constructor to allow create the source, but it probably will be removed as it won't be needed. So I need to the follow:

```cs
	[Fact, AutoInitShutdown]
	public void Constructor_With_Source_Initialize_With_The_Passed_SelectedItem ()
	{
		var cb = new ComboBox {
			Source = new ListWrapper (new List<string> () { "One", "Two", "Three" }),
			SelectedItem = 1
		};
		cb.BeginInit ();
		cb.EndInit ();
		cb.LayoutSubviews ();
		Assert.Equal ("Two", cb.Text);
		Assert.NotNull (cb.Source);
		Assert.False (cb.AutoSize);
		Assert.Equal (new Rect (0, 0, 0, 2), cb.Frame);
		Assert.Equal (1, cb.SelectedItem);
	}
```

Terminal.Gui/Views/FrameView.cs
BDisp: I didn't encounter nothing of this. Maybe was already fixed before.

tig:
@BDisp I updated the title of this PR. You ok with it?

BDisp:
@tig @tznind removing all parameters from constructors that can be defined by object initializers requires a change in procedures that were done in the past, such as executing a method that changes values of other properties without actually all properties being defined.
For example, `AutoSize` is one of those cases where it is `false` and all adjustments are made with that setting in the constructor. When the property is set in the object's initializer to `true`, it is considered a change as if the object was already initialized. This way it will be difficult to deal with these changes while the object is initializing.
It is really bad practice to calculate other properties based on the property being defined without the object being fully initialized and I recognize that I committed it too.
The frame will no longer be defined in the constructor because this parameter or any others will no longer actually exist. The frame can be set before initialization if X, Y, Width and Height are Pos/Dim Absolute. In unit tests, these values only have the value of the properties when `IsInitialized` is `false`. After initialization, property values will be examined and calculated.
Can the `ValidatePosDim` property be removed? I agree and would really appreciate your opinion. Thanks.

tig:
I think so. 

The reality of computed layout means that layout properties, such as height, width, auto size etc. are ALL indeterminate until after initialization. 

This is why I pushed to have all view initialization methods called be SetInitialProperties and why I put debug warnings in. 

The glorious work you are doing actually makes the library even more correct. Because constructors with parameters hide problems.  

I'm currently working on a bunch of autosize related investigations in preparation for finishing Dim.Auto. if there's anyway, you can avoid touching any of the layout/autosize related code as you do the constructor work that would be great.  

Or you can just tell me to pause and I'll wait until you're done with this PR before I do more work.

You might want to peek at the draft PR I have going. It's very drafty but you'll see that I cleaned up a bunch of unit tests and I've made some progress on making the current autosize behavior more clear.

tig:
Oh, and please don't feel like you have to fix all views and all constructors in this one PR. There's no reason why we can't take our time and chip away at it as long as we commit to doing it.  .

BDisp:
> Oh, and please don't feel like you have to fix all views and all constructors in this one PR. There's no reason why we can't take our time and chip away at it as long as we commit to doing it. .

Thanks for your comment. Don't worry. I'm committed to resolving this matter once and for all, if you don't mind of course.
I will post here the small changes needed to resolve this issue.
Just clarify the following for me. If `AutoSize` is true and Width and Height is defined as `DimAbsolute` in the object initializer by the user, the size cannot be lower than that defined by the user but only higher, do you agree with me? In practice we need to add a condition to `SetRelativeLayout` that only sets the _x, _y, _width and _height fields if `AutoSize` is false. Or would you prefer these values to always be defined? I'm just asking so I'm not making changes you don't agree with.
I would prefer to continue to just worry about removing parameters from constructors and ensuring that unit tests pass. Then we worry about resolving conflicts later, do you agree?

tig:
And.... just to ensure we are talking about the same thing. Here's a great example:

```cs
[Fact]
[AutoInitShutdown]
public void AutoSize_False_Label_IsEmpty_True_Return_Null_Lines ()
{
	var text = "Label";
	var label = new Label {
		Width = Dim.Fill () - text.Length,
		Height = 1,
		Text = text,
		AutoSize = false,
	};
...

In `v2_develop` changing `AutoSize` does not impact `Width` and `Height`. I'm not sure this is "correct" so I was experimenting with changing it. When I did, this unit test failed.

Changing the unit test to:

```cs
[Fact]
[AutoInitShutdown]
public void AutoSize_False_Label_IsEmpty_True_Return_Null_Lines ()
{
	var text = "Label";
	var label = new Label {
		AutoSize = false,
		Width = Dim.Fill () - text.Length,
		Height = 1,
		Text = text,
	};
```

Makes the unit test pass.

tig:
> > Oh, and please don't feel like you have to fix all views and all constructors in this one PR. There's no reason why we can't take our time and chip away at it as long as we commit to doing it. .
> 
> Thanks for your comment. Don't worry. I'm committed to resolving this matter once and for all, if you don't mind of course. I will post here the small changes needed to resolve this issue. Just clarify the following for me. If `AutoSize` is true and Width and Height is defined as `DimAbsolute` in the object initializer by the user, the size cannot be lower than that defined by the user but only higher, do you agree with me? In practice we need to add a condition to `SetRelativeLayout` that only sets the _x, _y, _width and _height fields if `AutoSize` is false. Or would you prefer these values to always be defined? I'm just asking so I'm not making changes you don't agree with. 

I'm massively debating that with myself literally right now ;-).

When `Dim.Auto` is done this all changes (e.g. `Width = Dim.Auto(), Height = 4` will work exactly as you'd expect: Only the width of the view will be "automatic". 

What I'm finding, in trying to keep the existing `AutoSize` functionality in place while I build `Dim.Auto` is I'm fighting with existing `AutoSize` behavior that is complex, confusing, and non-deterministic in places. The good news `AutoSize` works very well and has really great capabilities, and I really want to makes sure none of those capabilities are lost when we transition to `Dim.Auto`. 

> I would prefer to continue to just worry about removing parameters from constructors and ensuring that unit tests pass. Then we worry about resolving conflicts later, do you agree?

Yes, I agree. If you have to tweak a unit test though, please at least document where / why you changed it with a `// BUGBUG`..

BDisp:
Just to not complicate matters further, I will not remove the `ValidatePosDim` property now. After this PR is finalized and working correctly, we decide that.

tig:
> Just to not complicate matters further, I will not remove the `ValidatePosDim` property now. After this PR is finalized and working correctly, we decide that.

Let me tackle it as part of `Dim.Auto`, ok?

BDisp:
> Let me tackle it as part of `Dim.Auto`, ok?

Sure.

When I mentioned that the size can only increase, it is only in relation to the frame, of course, because the Width and Height always maintain the same value, right?

BDisp:
Yes @tig it passed like a charm.

image


```cs
	[Fact]
	[AutoInitShutdown]
	public void AutoSize_False_Label_IsEmpty_True_Return_Null_Lines ()
	{
		var text = "Label";
		var label = new Label {
			Width = Dim.Fill () - text.Length,
			Height = 1,
			Text = text,
			AutoSize = false
		};
		var win = new Window {
			Width = Dim.Fill (),
			Height = Dim.Fill ()
		};
		win.Add (label);
		Application.Top.Add (win);
		((FakeDriver)Application.Driver).SetBufferSize (10, 4);

		Assert.Equal (5, text.Length);
		Assert.False (label.AutoSize);
		Assert.Equal (new Rect (0, 0, 3, 1), label.Frame);
		Assert.Equal (new Size (3, 1), label.TextFormatter.Size);
		Assert.Equal (new List<string> { "Lab" }, label.TextFormatter.Lines);
		Assert.Equal (new Rect (0, 0, 10, 4), win.Frame);
		Assert.Equal (new Rect (0, 0, 10, 4), Application.Top.Frame);
		var expected = @"
┌────────┐
│Lab     │
│        │
└────────┘
";

		var pos = TestHelpers.AssertDriverContentsWithFrameAre (expected, _output);
		Assert.Equal (new Rect (0, 0, 10, 4), pos);

		text = "0123456789";
		Assert.Equal (10, text.Length);
		label.Width = Dim.Fill () - text.Length;
		Application.Refresh ();

		Assert.False (label.AutoSize);
		Assert.Equal (new Rect (0, 0, 0, 1), label.Frame);
		Assert.Equal (new Size (0, 1), label.TextFormatter.Size);
		Assert.Equal (new List<string> { string.Empty }, label.TextFormatter.Lines);
		expected = @"
┌────────┐
│        │
│        │
└────────┘
";

		pos = TestHelpers.AssertDriverContentsWithFrameAre (expected, _output);
		Assert.Equal (new Rect (0, 0, 10, 4), pos);
	}

```

BDisp:
@tig I'm having an idea that I think you might agree with. If `AutoSize` is true forcing `DimAbsolute` to keep the value is confusing because it is what is normally set when the frame is changed. I propose creating a new `Pos.PosFixed` and `Dim.DimFixed` and then using this to always force the same position and dimension. What do you think?

tig:
> > Let me tackle it as part of `Dim.Auto`, ok?
> 
> Sure.
> 
> When I mentioned that the size can only increase, it is only in relation to the frame, of course, because the Width and Height always maintain the same value, right?

Correct. These statements in `SetRelativeLayout` are the problem:

```cs
newDimension = AutoSize && autosize > newDimension ? autosize : newDimension;
```

IIRC, both you and I tried to get shrinking working but gave up.

WIth `Dim.Auto`, you'll note that growing and shrinking works great. This is one of the reasons I plan on removing `VIew.AutoSize` when `Dim.Auto` is complete (or at least making the `AutoSize` property be a shortcut for setting `Width = Dim.Auto(), Height = Dim.Auto`).

Pk2fvzB 1

tig:
> @tig I'm having an idea that I think you might agree with. If `AutoSize` is true forcing `DimAbsolute` to keep the value is confusing because it is what is normally set when the frame is changed. I propose creating a new `Pos.PosFixed` and `Dim.DimFixed` and then using this to always force the same position and dimension. What do you think?

I don' think this will be needed when `Dim.Auto` is done.

tig:
> > WIth `Dim.Auto`, you'll note that growing and shrinking works great. This is one of the reasons I plan on removing `VIew.AutoSize` when `Dim.Auto` is complete (or at least making the `AutoSize` property be a shortcut for setting `Width = Dim.Auto(), Height = Dim.Auto`).
> 
> It's really fantastic behavior. But you've already implemented it with the option to only grow up to a certain limit, right? Imagine that we don't want the container to grow too?

In the DimAuto demo scenario, the "container" you refer to is 

```cs
var view = new FrameView () {
			Title = "Type in the TextField to make View grow.",
			X = 3,
			Y = 3,
			Width = Dim.Auto (min: 50),
			Height = Dim.Auto (min: 10)
		};
```

To make it not grow, simply:

```cs
var view = new FrameView () {
			Title = "Type in the TextField to make View grow.",
			X = 3,
			Y = 3,
			Width = 50,
			Height = 10
		};
```

BDisp:
Ok, got it, so Dim.Auto is used in both container and subview views. This is very well put together. Great job.

BDisp:
This PR has many unit tests that have been changed to reflect the `AutoSize` changes to the vertical direction of the text with the correct behavior.

tig:
I think what I may do, when you say your're ready is merge this into 

- https://github.com/gui-cs/Terminal.Gui/pull/3202

I'm partying on the same files/code right now. You suffered the pain last time we were crossing streams, so it's only fair I do it this time.

If you still have a ton more on this branch you want to, do perhaps alooking at what I've done in #3202 would be helpful?

BDisp:
I've seen your code and it's really excellent, clean and simple. My only concern in my branch is simply to remove all constructors with parameters and that all unit tests pass with the correct behavior.
I appreciate that this time you are in charge of merging into your branch and the only thing I ask of you is that you do not modify the unit tests without asking me whether or not I am correct, because I was under the impression that many of them were previously incorrectly modified to accommodate code incorrect.

tig:
Also, for any file in this PR that you touch please do the following:

- Use Resharper's `Ctrl-E C` command to reformat.
- Use `Ctrl-K Ctrl-D` to apply .editorconfig 

This will help me merge because I'm doing the same.

tig:
> I'm not using `Resharper` and I only use `Ctrl-K Ctrl-D`.

Ok, when you're ready, I'll first submit a PR to this PR that does all the code-reformatting. You can then merge it. Then I'll try to merge this with mine.

BDisp:
Which format Resharper's `Ctrl-E C` does?

BDisp:
Ok I'll install Resharper's  extension.

BDisp:
@tig this is the normal Resharper's format? By the way the `Ctrl-E C` doesn't work. I had to go to Extensions->ReSharper->Edit->Cleanup Code...

image

tig:
> @tig this is the normal Resharper's format? 

I think it's a bug related to the fact we use 8 spaces for tabs. See: https://github.com/gui-cs/Terminal.Gui/pull/3207#issuecomment-1907162782

> By the way the `Ctrl-E C` doesn't work. I had to go to Extensions->ReSharper->Edit->Cleanup Code...

I think i let Resharper take over that keystroke.

BDisp:
I was unable to remove the constructor parameters
`public ScrollBarView (View host, bool isVertical, bool showBothScrollIndicator = true)`
because I need to distinguish whether it is hosted or not. This can only be controlled through the parameter in the constructor. The difference is that if it is hosted, the `ScrollBarView` is added to the Host's `Superview`. Otherwise it is added to the `Host` itself. However, there is no need to waste a lot of time on this either because we should implement auto scrolling in `Padding`.

BDisp:
> I think i let Resharper take over that keystroke.

Fixed the `Ctrl-E C` ReSharper's shortcut by resetting all Visual Studio settings (Tools->Import and Export Settings->Reset all settings), then in to the ReSharper->Options->Keyboard->Shortcut Scheme->Visual Studio->Apply Scheme.

Terminal.Gui/Views/ScrollBarView.cs
```diff
@@ -171,9 +172,11 @@ public ScrollBarView (View host, bool isVertical, bool showBothScrollIndicator =
 	public ScrollBarView OtherScrollBarView {
 		get => _otherScrollBarView;
 		set {
-			if (value != null && (value.IsVertical && _vertical || !value.IsVertical && !_vertical)) {
-				throw new ArgumentException ($"There is already a {(_vertical ? "vertical" : "horizontal")} ScrollBarView.");
+			if (value != null && ((value.IsVertical && _vertical) || (!value.IsVertical && !_vertical))) {
```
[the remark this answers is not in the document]
dodexahedron: This, by the way, is a null check, a compatible type check, and the actual value checks all in one, and you can also name the matched pattern for use at the same scope as the control flow statement itself, when advantageous (not just inside the control flow statement's associated block). That also helps avoid having separate declarations and assignments of locals.

dodexahedron (changes requested):
And more general commentary about null checking, which applies to a significant portion of this code:

If you find yourself checking the same variable, field, or property against null multiple times in a given context, that's usually an indicator that you can use null propagation or null coalescing operators to simplify the code, or, in cases that are just too complex to easily simplify, it's both cleaner and more readable, yet will generate the same compiled IL, to assign the result of that common null check to a local bool with a name that explains what it represents, and then use that bool in the control flow statements that previously had the repeated checks.

It's also a thread safety issue to perform multiple checks against the same value like that, because you lose atomicity of the operations and have no guarantee that the reference or the object it refers to has remained unchanged between checks.

That's why switch statements or expressions tend to be the best overall option, since they evaluate the condition once and then perform a lookup and a jump to the code for the target case label.

Terminal.Gui/Views/ScrollBarView.cs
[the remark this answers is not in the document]
BDisp: My concern in this pull request is just removing the parameters from the constructors, doing some code cleanup and making sure all unit tests pass. Doing everything in the same PR is complex, focusing on all the details at once.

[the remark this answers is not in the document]
dodexahedron: That sounds perfectly reasonable to me. Carry on. :)

dodexahedron (approved):
Just closing my review as @BDisp made the good point that it's not in scope for this particular PR.

BDisp:
> For "readonly" parameters, `init` or `priviate init` can be used.

This is only possible when the field `readonly` has a public accessor, right?

BDisp:
ReSharper `Ctrl-E. C` on TextViewTests.cs with 7600 lines freezes and VS2022 complains.

image

dodexahedron:
If something is huge, it's better for both time and for your ability to watch what has happened to perform major automated stuff on smaller portions at a time.

The code cleanup command can operate on a selected portion of code. Select something, use the ctrl-e,c, and let it go.

It sounds to me though like you might have some settings that are causing more processing to be happening than there should be. For some things, if something from another plugin or from visual studio that is set in a way that conflicts with what's happening or that would result in some other automatic action being taken, leading to a loop, you could get that kind of behavior.

I've formatted much larger files than that without issue, though, so there probably is a conflict or loop somewhere.

dodexahedron:
You can also try not running all of the rules at once.

The default rule is everything all at once. Other profiles only do a subset of the operations at once.

BDisp:
I clicked on  "Don't report this extension again" and after a while it ends. I understand that huge file may take some more time to finish. After Code Cleanup nothing was modified. Can you pull please and try if it also happens to you? It's already in the last commit. Thanks.

dodexahedron:
Yeah I can do that when I get back home later.

Things like this are part of why I sometimes do things like I did in the Color issue, such as breaking a class up into multiple nestable files, with the class defined as partial and organized into logical groups per file.

The more stuff VS has going on, the harder it gets to deal with large files in general.

dodexahedron:
Ok, so I checked out commit b9349fa075278e80f782187aeec80535362aeb20

The problem is the settings files are not there at all, so you're just getting whatever defaults plus any changes you might have added on top.

If you merge this commit into your branch, you'll have the latest solution and project settings I've pushed:
7dc2ace8126dd2325ae005886cce45c093e436e3

I'd probably run the sort code profile first, and then run the project default profile, when running the cleanup. That way there's a lot less going on at once, for each run.

dodexahedron:
Just note that merging that may end up as a merge conflict on the settings files themselves, at some point.

But if/when that happens, the right thing to do is to treat the version from v2_develop as correct for the dotsettings files.

BDisp:
Yes I'll use the current v2_develop settings because @tig already said he will handles conflicts merge from some PR, included this one. Thanks so much for your explanation.

dodexahedron:
Cool.

However, there's a catch to that, too.

Some of the changes necessary to make everything follow project guidelines are marked as error.

You may want to wait til we have the formatting redone to use those settings, or you'll need to change the config until we finish that so you can build the code.

Terminal.Gui/Views/Menu/ContextMenu.cs
```diff
@@ -188,13 +188,14 @@ public void Show ()
...
-		_menuBar = new MenuBar (new [] { MenuItems }) {
+		_menuBar = new MenuBar {
 			X = position.X,
 			Y = position.Y,
 			Width = 0,
 			Height = 0,
 			UseSubMenusSingleFrame = UseSubMenusSingleFrame,
-			Key = Key
+			Key = Key,
+			Menus = [MenuItems]
```
dodexahedron, line 198: Woo!

Glad to see collection expressions getting used. 😀

UICatalog/Scenarios/ASCIICustomButton.cs
```diff
@@ -16,15 +17,18 @@ public override void Init ()
...
-				_miSmallerWindow = new MenuItem ("Smaller Window", "", ChangeWindowSize) {
-					CheckType = MenuItemCheckStyle.Checked
-				},
-				null,
-				new MenuItem ("Quit", "", () => Application.RequestStop (), null, null, (KeyCode)Application.QuitKey)
-			})
-		});
+		var menu = new MenuBar {
+			Menus = [
+				new MenuBarItem ("Window Size", new [] {
```
dodexahedron, line 22: I know it's not necessarily in scope for this PR, but I just wanted to point out that anywhere there's a `new[]` or a `new SomeCollectionType();` is a potential and quite likely place that a collection expression is allowed and potentially better, too! 😃 

Others were changed here too, so it seemed relevant to point out. 🤷‍♂️

Terminal.Gui/Views/Menu/ContextMenu.cs
BDisp: At least I'm use them on the changes I really have to do but not all yet.

dodexahedron (commented):
No changes needed. Comments were just for visibility. :)

Terminal.Gui/Views/Menu/ContextMenu.cs
dodexahedron: Haha hey I'm just thrilled to see any use of them. They can have a surprisingly large impact because of how they designed that feature.

BDisp:
@tig this is finally already ready for review. The only I didn't removed was the `FileDialog` - Parameter can't be removed because there isn't a public getter for _fileSystem field. I don't know if it's intentional not exposing it or I can create a `FileSystem` property with a `init` setter.

Terminal.Gui/Views/Menu/MenuBar.cs
BDisp: Done 😃

tznind:
> I don't know if it's intentional not exposing it

There is a public constructor:

```csharp
public FileDialog () : this (new FileSystem ()) { }
```

So you could make this one internal:

```csharp
public FileDialog (IFileSystem fileSystem)
```

> or I can create a FileSystem property with a init setter.

Need to be careful here as the file system prop is shared with the dialog style.  `init` could work but just be a bit careful and ensure there doesn't end up a de-sync between dialog and style classes.

BDisp:
> Need to be careful here as the file system prop is shared with the dialog style. `init` could work but just be a bit careful and ensure there doesn't end up a de-sync between dialog and style classes.

My question was whether I could give the user access to obtain data from the `FileSystem`. If there is no problem with this problem then I will use an `init` setter. It will practically work the same, just giving the user visibility of the `FileSystem` data. Thanks.

tig:
@BDisp did you apply resharper code-cleanup to all the files in the solution in this PR?

For example, it looks like `PosDim.cs` has been cleaned up, but you didn't make any functional changes.

I just want to make sure I know what I'm looking at before I dive in and review.

BDisp:
> @BDisp did you apply resharper code-cleanup to all the files in the solution in this PR?
> 
Yes I did for all the files that were touched by the changes.

> For example, it looks like `PosDim.cs` has been cleaned up, but you didn't make any functional changes.
> 
It changed because I did a find/replace on the solution scope from `TextField () {` to `TextField {`. and it changed the line 153 which is now 163.

> I just want to make sure I know what I'm looking at before I dive in and review.

I understand.

BDisp:
> > I don't know if it's intentional not exposing it
> 
> There is a public constructor:
> 
> ```cs
> public FileDialog () : this (new FileSystem ()) { }
> ```
> 
> So you could make this one internal:
> 
> ```cs
> public FileDialog (IFileSystem fileSystem)
> ```
> 
> > or I can create a FileSystem property with a init setter.
> 
> Need to be careful here as the file system prop is shared with the dialog style. `init` could work but just be a bit careful and ensure there doesn't end up a de-sync between dialog and style classes.

As in the comment is explicitly is for testing ( `/// <remarks>This overload is mainly useful for testing.</remarks>`) I'll change it to internal as recommended by the @tznind. Thanks.

BDisp:
@tig I remembered that I only started running resharper code-cleanup after you asked. Now I reflected that it might not have been done for some files that were previously changed and that perhaps it was not changed in subsequent commitments. After all, there were still 46 files to be carried out by resharper code-cleanup and already corrected in the commit https://github.com/gui-cs/Terminal.Gui/pull/3181/commits/b4650fa95e648a773ddf3b4420a6ce997149d209.

BDisp:
Why R# `Ctrl-E, C` convert simples auto-properties to backed fields? Like the following.

From:
`public object Data { get; set; }`

To:
```cs
object _data;

public object Data {
	get => _data;
	set => _data = value;
}
```

Is this really necessary?

BDisp:
> Why R# `Ctrl-E, C` convert simples auto-properties to backed fields? Like the following.
> 
> From: `public object Data { get; set; }`
> 
> To:
> 
> ```cs
> object _data;
> 
> public object Data {
> 	get => _data;
> 	set => _data = value;
> }
> ```
> 
> Is this really necessary?

I reverted this changes and reapply the R# cleanup code again and this doesn't happened again. I don't know why this conversion was apply before. Solved.

tig:
I think I'll have some bandwidth this week to work on this again.

tig:
@BDisp please revert the TextFormatter changes you made in this PR. 

If you can, also please, revert any code reformatting you did that was not in a file impacted by actually removing the parameters from constructors. 

This will greatly assist me in merging this. Thanks.

tig:
This has been merged into #3202.

However, #3202 is not yet ready because merging this PR into it broke a bunch of stuff I had working.

I'm closing.

PR #3187, opened 2024-01-16
Fixes #3186. HotKeys must be explicitly set (Makes `TextFormatter.FindHotKey` `firstUpperCase` default to `false`)

### LINKED ISSUE #3186: Turn off `firstUpperCase = true` in `TextFormatter.FindHotKey`
```cs
/// <param name="firstUpperCase">If <c>true</c> the legacy behavior of identifying the first upper case character as the HotKey will be enabled.
/// Regardless of the value of this parameter, <c>hotKeySpecifier</c> takes precedence.</param>
```
I'd like to remove this "legacy" behavior and make it so that hot keys are only generated when explicitly called for.

My primary reason for wanting this change is I think it's worse to have no hot key than have a hotkey that conflicts with another one because the automatic behavior created the conflict. 

New method signature:

```cs
/// <summary>
/// Finds the HotKey and its location in text. 
/// </summary>
/// <param name="text">The text to look in.</param>
/// <param name="hotKeySpecifier">The HotKey specifier (e.g. '_') to look for.</param>
/// <param name="hotPos">Outputs the Rune index into <c>text</c>.</param>
/// <param name="hotKey">Outputs the hotKey. <see cref="Key.Empty"/> if not found.</param>
/// <param name="firstUpperCase">If <c>true</c> the legacy behavior of identifying the
/// first upper case character as the HotKey will be enabled.
/// Regardless of the value of this parameter, <c>hotKeySpecifier</c> takes precedence.
/// Defaults to <see langword="false"/>.</param>
/// <returns><c>true</c> if a HotKey was found; <c>false</c> otherwise.</returns>
public static bool FindHotKey (string text, Rune hotKeySpecifier, out int hotPos, out Key hotKey, bool firstUpperCase = false)
```

- [ ] All places where we call this will remove `true`.
- [ ] Unit test will be updated to test appropriately.

This change will cause a lot of app code to function differently. Any place where the developer did not specify `_` there will no longer be a hotkey.

tig:
## Fixes

- Fixes #3186 

## Proposed Changes/Todos

- [x] Change `FindHotKey`
- [x] Fix Unit tests
  - [x] FileDialog
  - [x] Button
  - [x] Checkbox
  - [x] Tons more
- [x] Fix glaring UI Catalog issues 

## Pull Request checklist:


BDisp (approved):
That was the best option. Better don't have hot keys than having a lot of duplicates causing conflicts. Now is user responsibility if it still has conflicts.

Issue #2445, opened 2023-03-25
DateField doesn't work with dd-MMM-yy format


popara96:
**Description**
DateField throws an exception when the system uses dd-MMM-yy format with string literals instead of numbers.

**Steps to reproduce the behavior:**
1. Go to 'Control Panel'->'Clock and Region'->'Change date, time, or number formats'->'Additional settings'->'Date' tab->set 'Short date' to 'dd-MMM-yy'
2. Try to test DateTime in UICatalog

**Expected behavior**
We expect to see DateFields. Instead we get an exception: ___System.FormatException: 'The input string 'Mar' was not in a correct format.'___

**Screenshots**
![image](https://user-images.githubusercontent.com/31167073/227720891-01189bdd-6421-41dc-9cb7-ab83966a96ea.png)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: N/A
 - Version: N/A

tig:
I believe this is fixed with

- #2851 

and 

- #3180

Right?

MaciekWin3:
> I believe this is fixed with
> 
> * [DateField: Allow Formatting #2851](https://github.com/gui-cs/Terminal.Gui/issues/2851)
> 
> and
> 
> * [DateField not reflecting user input #3180](https://github.com/gui-cs/Terminal.Gui/issues/3180)
> 
> Right?

Yeah, that's right. What is more now you can set up format using `Culture` property in DateField.

BDisp:
> I believe this is fixed with
> 
> * [DateField: Allow Formatting #2851](https://github.com/gui-cs/Terminal.Gui/issues/2851)
> 

Yes.

> and
> 
> * [DateField not reflecting user input #3180](https://github.com/gui-cs/Terminal.Gui/issues/3180)
> 
No, the user is asking for a MMM format which will use the month abbreviature as Mar. `DateField` only support short format data based on the culture used, "dd/MM/yyyy", "dd-MM-yyyy", "dd.MM.yyyy". The reason of that is `DateField` only validate numbers and not letters.

MaciekWin3:
What I meant app will no longer crush 😁. Correct me if I am wrong, but with recent changes DateField will just use default format which is `"dd/MM/yyyy"` in this case.

BDisp:
> What I meant app will no longer crush 😁. Correct me if I am wrong, but with recent changes DateField will just use default format which is `"dd/MM/yyyy"` in this case.

[...]

PR #3188, opened 2024-01-18
Fixes #3180: DateField to handle all date formats

### LINKED ISSUE #3180: DateField not reflecting user input
**Describe the bug**
While working on DatePicker improvements, I discovered that DateField is not working as intended. Specifically, the date is not updating when entered via typing:

https://github.com/gui-cs/Terminal.Gui/assets/57466073/32b7bf28-7f58-4f19-9466-9780926ef848

**To Reproduce**
1) Navigate to UICatalog and open the DateTime/DatePicker example.
2) Attempt to change the date using the provided interface.
3) After continuing typing beyond the last digit of the year, the application crashes.

**Expected behavior**
Date should be changed to what user typed

MaciekWin3:
## Fixes #3180: DateField not reflecting user input

In colaboration with @BDisp we want to make DateField to handle all date formats and also fix #3180 

- [x] Handle dates with `\u200f` separator

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/DateField.cs
+++ b/Terminal.Gui/Views/DateField.cs
@@ -20,9 +20,29 @@ namespace Terminal.Gui;
 /// </remarks>
 public class DateField : TextField {
 	DateTime _date;
-	int _fieldLen = 10;
-	string _sepChar;
-	string _format;
+
+
+	/// <summary>
+	/// Format of date. The default is CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern.
+	/// </summary>
+	//public string Format { get; set; } = CultureInfo.CurrentCulture.DateTimeFormat.ShortDatePattern;
+	public string Format { get; set; } = " " + CultureInfo.GetCultureInfo ("pl-PL").DateTimeFormat.ShortDatePattern;
+
+	private int FieldLen => Format.Trim ().Length;
+
+	private char Separator {
+		get {
+			if (!string.IsNullOrEmpty (Format)) {
+				var sep = Format
+					.Where (c => !char.IsLetterOrDigit (c))
+					.Where (c => !char.IsWhiteSpace (c))
+					.FirstOrDefault ();
+				return sep;
+			}
+
+			return '/';
+		}
+	}
 
 	/// <summary>
 	///   DateChanged event, raised when the <see cref="Date"/> property has changed.
@@ -46,15 +66,13 @@ public class DateField : TextField {
 	/// <param name="date"></param>
 	public DateField (DateTime date) : base ("")
 	{
-		Width = _fieldLen + 2;
+		Width = FieldLen + 2;
 		SetInitialProperties (date);
 	}
 
 	void SetInitialProperties (DateTime date)
 	{
-		var cultureInfo = CultureInfo.CurrentCulture;
-		_sepChar = cultureInfo.DateTimeFormat.DateSeparator;
-		_format = $" {cultureInfo.DateTimeFormat.ShortDatePattern}";
+		//CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo ("en-GB");
 		Date = date;
 		CursorPosition = 1;
 		TextChanging += DateField_Changing;
@@ -111,8 +129,6 @@ public class DateField : TextField {
 	void DateField_Changing (object sender, TextChangingEventArgs e)
 	{
 		try {
-			var cultureInfo = CultureInfo.CurrentCulture;
-			DateTimeFormatInfo ccFmt = cultureInfo.DateTimeFormat;
 			int spaces = 0;
 			for (int i = 0; i < e.NewText.Length; i++) {
 				if (e.NewText [i] == ' ') {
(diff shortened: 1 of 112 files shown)


BDisp (approved):
That's excellent thanks.

Terminal.Gui/Views/DateField.cs
```diff
@@ -121,13 +119,13 @@ void DateField_Changing (object sender, TextChangingEventArgs e)
...
+			spaces += FormatLength;
 			string trimedText = e.NewText [..spaces];
-			spaces -= _fieldLen;
+			spaces -= FormatLength;
 			trimedText = trimedText.Replace (new string (' ', spaces), " ");
-			var date = Convert.ToDateTime (trimedText, ccFmt).ToString (ccFmt.ShortDatePattern);
+			var date = Convert.ToDateTime (trimedText).ToString (_format.Trim ());
 			if ($" {date}" != e.NewText) {
-				e.NewText = $" {date}";
+				e.NewText = $" {date}".Replace ("\u200f", "");
```
tig, line 128: Please make "\u200f" a const or at least document what the heck it is. Code written for this library should assume people reading it are 6-year-olds. ;-)

Terminal.Gui/Views/DateField.cs
```diff
@@ -283,10 +305,52 @@ static int GetFormatIndex (string [] fm, string t)
...
+	{
+		var sepChar = separator.Trim ();
+		if (sepChar.Length > 1 && sepChar.Contains ('\u200f')) {
+			sepChar = sepChar.Replace ("\u200f", "");
+		}
+
+		return sepChar;
+	}
+
+	private static string StandardizeDateFormat (string format) =>
```
tig, line 318: This feels very hacky and fragile. There has got to be a more straight forward and readable way of accomplishing what you're doing here. 

Almost all of these map to the same thing, but some appear to map a `?` to the same char that proceeds the `?`. 

This code is obtuse. If this is really the simplest way of doing this then please include comments explaining the logic.

tig (changes requested):

Terminal.Gui/Views/DateField.cs
MaciekWin3: Good idea, done

Terminal.Gui/Views/DateField.cs
MaciekWin3: Yeah, I agree that this code is obtuse. However, other, more sophisticated solutions were much more complex, and none of them worked as we needed to handle numerous edge cases. With the current approach, this method gets the job done, and it truly is the simplest way to make it work. However, I admit it's not the most elegant solution.

tig: Fine. At least document the heck out of it with comments please.

MaciekWin3: @tig I have already included a comment describing the StandardizeDateFormat method. Is there anything else you would like documented?

tig (approved):
Nicely done team!

BDisp:
@MaciekWin3 is there any reason for using a method for `FormatLength`? It's enough to only depend of the `_format.Length`. Furthermore it always fallback to a constant value equal to 10 and thus `private readonly int _format.Length = 10;` fit it very well. `FormatLength` is called on 11 places and takes more time to return from `StandardizeDateFormat` then from `Trim` and finally `Length` when it only needed to get the length from the `_format.Length` which is already be set early or getting the constant value of 10.

@tig to prove that the `StandardizeDateFormat` method created by @MaciekWin3 is working fine I created the following method that ensures always the correct format with always using the culture separator. The unit test passed. I think a three old child could read it 😄 

```cs
	private static string StandardizeDateFormat (string format, string separator)
	{
		// Get rid of all strings that aren't d, M, y and separator 
		string fmt = null; ;
		var numOfSep = 0;
		for (int i = 0; i < format.Length; i++) {
			var c = format [i];
			if (c == 'd' || c == 'M' || c == 'y') {
				fmt += c.ToString ();
			}
			if (c == separator [0]) {
				numOfSep++;
				// Ensures that there are only 2 separators and ignores the remaining ones
				if (numOfSep <= 2) {
					fmt += c.ToString ();
				}
			}
		}

		// Ensures that all the required d, M, y and separator exist
		if (!fmt.Contains ('d')) {
			fmt += "dd";
		}
		if (!fmt.Contains ('M')) {
			fmt += "MM";
		}
		if (!fmt.Contains ('y')) {
			fmt += "yyyy";
		}
		if (!fmt.Contains (separator)) {
			throw new ArgumentException ("Separator must be provide!", nameof (separator));
		}

		// Ensures there are 2 separators and adds them as needed
		var firstChar = fmt [0];
		if (fmt.Sum (c => c == separator [0] ? 1 : 0) < 2) {
			for (int i = 0; i < fmt.Length; i++) {
				var c = fmt [i];
				if (c == firstChar) {
					switch (c) {
					case 'd':
					case 'M':
						i += 2;
						firstChar = InsertSeparator (ref fmt, separator, i);
						break;
					case 'y':
						i += 4;
						firstChar = InsertSeparator (ref fmt, separator, i);
						break;
					}
				}
			}
		}

		// Splits the format by separator
		string [] splitedFmt = fmt.Split (separator);
		// Generate the required formated string with 2 digits for the day, 2 digits for the month,
		// 4 digits for the year, with the respective separator and corresponding ordering of the location
		string newFmt = fmt;
		for (int i = 0; i < splitedFmt.Length; i++) {
			if (splitedFmt [i].Contains ("M") && splitedFmt [i].GetRuneCount () < 2) {
				newFmt = newFmt.Replace ("M", "MM");
			}
			if (splitedFmt [i].Contains ("d") && splitedFmt [i].GetRuneCount () < 2) {
				newFmt = newFmt.Replace ("d", "dd");
			}
			if (splitedFmt [i].Contains ("y") && splitedFmt [i].GetRuneCount () < 4) {
				newFmt = newFmt.Replace ("yy", "yyyy");
			}
		}

		return newFmt;

		static char InsertSeparator (ref string fmt, string separator, int i)
		{
			if (i < 10 && fmt [i] != separator [0]) {
				fmt = fmt.Insert (i, separator [0].ToString ());
			}
			if (i < 10) {
				return fmt [i + 1];
			}
			return default;
		}
	}

```

MaciekWin3:
> @MaciekWin3 is there any reason for using a method for `FormatLength`? It's enough to only depend of the `_format.Length`. Furthermore it always fallback to a constant value equal to 10 and thus `private readonly int _format.Length = 10;` fit it very well. `FormatLength` is called on 11 places and takes more time to return from `StandardizeDateFormat` then from `Trim` and finally `Length` when it only needed to get the length from the `_format.Length` which is already be set early or getting the constant value of 10.

I agree with you. Should I make PR for that or you will take care of this? Also I forget that we were talking about adding functionality to DateField to be able to open DatePicker in modal. What are your thoughts on that? 

>@tig to prove that the StandardizeDateFormat method created by @MaciekWin3 is working fine I created the following method that ensures always the correct format with always using the culture separator. The unit test passed. I think a three old child could read it 😄

😁

BDisp:
See here please https://github.com/gui-cs/Terminal.Gui/issues/2445#issuecomment-1901121600.

PR #3202, opened 2024-01-21
Fixes #3192. Improve correctness / clarity of existing `View.AutoSize` functionality/unit tests

### LINKED ISSUE #3192: Improve correctness / clarity of existing View.AutoSize functionality/unit tests
In 

- https://github.com/gui-cs/Terminal.Gui/pull/3062

...I am studying the existing `AutoSize = false` code paths.

I've noted this before, but there's not an issue on it: 

- `View` should not change it's dimensions if `AutoSize = false` if `Text` is changed. 
- The default size of `View` (and any subclass that does not have an explicit reason) should be `0, 0`.

I'm going to submit a PR to fix this (instead of bundling it with #3062).

tig:
## Fixes

- Fixes #3192

## Proposed Changes/Todos

- [x] Rename/Move AutoSize related unit tests to a) make more clear, b) help diagnose issues by having all tests that test `AutoSize = true` in one place.
- [x] Enforce that it makes no sense to set `Width/Height` if `AutoSize = true`. Update Unit tests to match.
- [x] Modernize `AutoSize` related unit tests to remove dependency on `Application` and `Label`
- [x] Create a new set of `AutoSize = true` unit tests that ensure full understanding of existing capabilities
- [x] `View` with `AutoSize = true` and `Label` defaults to dims of `0,0` if `Text` is empty. `Button` defaults to `5, 1` (`⟦► ◄⟧`).
- [x] Change `TextFormatter.Lines_get` to NOT cause a format. Callers should call `Format` explicitly. Property getters should not cause side effects. In this case, it causes all sorts of mayhem when debugging `TextFormatter` because the debugger display does `Lines_get` all the time.
- [x] Make `View.TextFormatter` be `init` only. 

## Pull Request checklist:


tig:
This puppy is ready for review.

However, it impacts so many files that I will wait until 

- #3207 

Is merged before progressing.

UnitTests/Views/TextValidateFieldTests.cs
```diff
@@ -19,14 +17,13 @@ public void Initialized_With_Cursor_On_First_Editable_Character ()
...
 
-		field.NewKeyDownEvent (new Key (KeyCode.D1));
+		field.NewKeyDownEvent (new (KeyCode.D1));
 
 		Assert.Equal ("--(1___)--", field.Provider.DisplayText);
 		Assert.Equal ("--(1   )--", field.Text);
 	}
 
 	[Fact]
-	[AutoInitShutdown]
```
dodexahedron: 🥳

BDisp: That was already removed and only need to be merges with the v2_develop branch.

https://github.com/gui-cs/Terminal.Gui/blob/a3627aed5e4038b7b105cb98401472f3a5ed4fae/UnitTests/Views/TextValidateFieldTests.cs#L26-L27

tig:
Yay: #3181 has been merged into this branch.

Boo: I overwrote some of my changes in the merge and now have to go back and re-fix them.

But progress!

BDisp:
I hope you fix them easily. I'll based this PR on the implementation of the #2489. Right?

tig:
> I hope you fix them easily. I'll based this PR on the implementation of the #2489. Right?

I'd wait until this is merged if I were you. While this branch has partial "code cleanup" it's not done and it's likely that the formatting and syntax style will change causing more merge hell.

tig:
@BDisp, @tznind, @dodexahedron this is now ready to merge!

I tried applying the R# code clean up settings to the entire project but R# screwed some files up. it seems to mess up anything with simple array initalizers. E.g. 

```cs
    [Theory]
    [CombinatorialData]
    public void Constructor_WithByteRGBValues_AllValuesCorrect ([CombinatorialValues (0, 1, 254)] byte r, [CombinatorialValues (0, 1, 253)] byte g, [CombinatorialValues (0, 1, 252)] byte b) {
        var color = new Color (r, g, b);

        ReadOnlySpan<byte> bytes = [b, g, r, 255];
        int expectedRgba = BitConverter.ToInt32 (bytes);
        uint expectedArgb = BitConverter.ToUInt32 (bytes);

        Assert.Multiple (
            () => Assert.Equal (r, color.R),
            () => Assert.Equal (g, color.G),
            () => Assert.Equal (b, color.B),
            () => Assert.Equal (byte.MaxValue, color.A),
            () => Assert.Equal (expectedRgba, color.Rgba),
            () => Assert.Equal (expectedArgb, color.Argb)
        );
    }
```

Becomes

```cs
    [Theory]
    [CombinatorialData]
    public void Constructor_WithByteRGBValues_AllValuesCorrect (
        [CombinatorialValues (0, 1, 254)] byte r,
        [CombinatorialValues (0, 1, 253)] byte g,
        [CombinatorialValues (0, 1, 252)] byte b
    ) {
        var color = new Color (r, g, b);

        ReadOnlySpan<byte> bytes =  [b]
        [g]
        [r]

        255];
        var expectedRgba = BitConverter.ToInt32 (bytes);
        var expectedArgb = BitConverter.ToUInt32 (bytes);

        Assert.Multiple (() => Assert.Equal (r, color.R),
                         () => Assert.Equal (g, color.G),
                         () => Assert.Equal (b, color.B),
                         () => Assert.Equal (byte.MaxValue, color.A),
                         () => Assert.Equal (expectedRgba, color.Rgba),
                         () => Assert.Equal (expectedArgb, color.Argb));
    }
```

We need to diangose this before proceeding.

BDisp:
This is good.

```cs
    [Theory]
    [CombinatorialData]
    public void Constructor_WithByteRGBValues_AllValuesCorrect (
        [CombinatorialValues (0, 1, 254)] byte r,
        [CombinatorialValues (0, 1, 253)] byte g,
        [CombinatorialValues (0, 1, 252)] byte b
    ) {
        var color = new Color (r, g, b);
```

This is ugly.

```cs
        ReadOnlySpan<byte> bytes =  [b]
        [g]
        [r]

        255];
```

This is good.

```cs
        Assert.Multiple (() => Assert.Equal (r, color.R),
                         () => Assert.Equal (g, color.G),
                         () => Assert.Equal (b, color.B),
                         () => Assert.Equal (byte.MaxValue, color.A),
                         () => Assert.Equal (expectedRgba, color.Rgba),
                         () => Assert.Equal (expectedArgb, color.Argb));
```

tig:
I had an old version of reshaper installed. New version doesn't have this bug. reformatting entire project now.

tig:
Done.

Look about all the files and see if there's any formatting or style that strikes you as egregiously bad. For the most part i like it, but I'm sure I missed something.

There are probably new bugs in this branch too. We touched a TON of stuff.  The changes for this PR especially. But I did spend a ton of time on new/improved unit tests so...

BDisp:
Good hard work, congrats. It's possible to do R# code cleanup for the solution at once, without have to do one by one?

tig:
This is weird. On my other machine, after pulling this branch down and ensuring I have the latest version of R# installed, doing a code clean up on a file (e.g. `ViewLayout.cs`) is applying different formatting than what my other machine provided.

I'm clearly missing something in how R# config works. Frustrating!

BDisp:
Probably is some local configuration in one machine that isn't included on the file settings but only in the options itself.

tig:
@BDisp will you please pull this PR down and see if Ctrl-E, Ctrl-C causes files to change after you do so? 

If it does, delete any `*.DotSettings.user` files and try again.

BDisp:
> @BDisp will you please pull this PR down and see if Ctrl-E, Ctrl-C causes files to change after you do so?
> 
It took about 13 minutes and I got 11 files changed after it finished.

image

> If it does, delete any `*.DotSettings.user` files and try again.

After delete all the `*.DotSettings.user` files which was two, `Terminal.Gui.sln.DotSettings.user` and `Terminal.Gui.csproj.DotSettings.user` it took the same time and with the same files changed.

image

BDisp:
But the changes are only a few differences about format and I think they make sense. I'll send a PR to your branch with the changes for you to compare.

BDisp:
> But the changes are only a few differences about format and I think they make sense. I'll send a PR to your branch with the changes for you to compare.

Done @tig on the https://github.com/tig/Terminal.Gui/pull/21.

dodexahedron:
> I had an old version of reshaper installed. New version doesn't have this bug. reformatting entire project now.

Yep. That'll do it. The latest version includes some important functionality for language version 12 that isn't fully understood by earlier versions.

Rules may still need minor tweaking after this is all finished, but it's great this is almost over!

dodexahedron:
> This is weird. On my other machine, after pulling this branch down and ensuring I have the latest version of R# installed, doing a code clean up on a file (e.g. `ViewLayout.cs`) is applying different formatting than what my other machine provided.
> 
> I'm clearly missing something in how R# config works. Frustrating!

Typically that comes down to local configuration that perhaps could have gotten applied if you ever used the context menus to change a behavior or severity or whatever.

The settings and the way they are layered is extremely flexible, but it's _so_ flexible that it's easy to end up with settings on one machine that differ subtly from another.

I wish they'd put some sort of indicator in the settings UI to indicate if a given setting is overridden by another layer or overrides another layer. It'd make it a lot easier to track down little conflicts without having to wipe or replace local settings (or reconciled them manually, one-by-one, but that's insane).

I also wish there were an option to exclude the default layers - like a checkbox by each, the way it does for additional custom layers you add. The way inheritance of the settings works, you have to explicitly override something that is set in a previously-applied layer to get it back to normal, but the XML by default will not contain anything with an explicitly default value, if you do it in the UI. Therefore, a setting at one layer that sets something non-default, but a later layer wants to return back to a value equal to default won't override, if done in the UI. You can put the XML in the file manually, but it'll get removed the next time the UI is used to modify that layer. 😤

dodexahedron:
Also, something to note, guys:

If your VS settings specify respecting the .editorconfig, or (and especially) if your resharper settings do so, you're going to get rules that are a combination of the two, with potentially unintuitive results. That gets even worse if you also have any formatting settings in Visual Studio still enabled, while ReSharper is otherwise responsible for formatting.

I'd suggest that, once this is fully completed and merged, we maintain a higher vigilance for a while regarding formatting changes when making commits, and try to determine their root causes, so we can refine the rulesets to prevent that kind of thing.

Also, as a general note:

The refactorings that sort code I highly suggest you run independently of others, in a separate commit, going forward, as those tend to result in confusing diffs in many merge tools, and can present as bizarre and nonsensical merge conflicts when combined with other actual code changes in the same file.

dodexahedron:
> > @BDisp will you please pull this PR down and see if Ctrl-E, Ctrl-C causes files to change after you do so?
> 
> It took about 13 minutes and I got 11 files changed after it finished.
> 
> image
> 
> > If it does, delete any `*.DotSettings.user` files and try again.
> 
> After delete all the `*.DotSettings.user` files which was two, `Terminal.Gui.sln.DotSettings.user` and `Terminal.Gui.csproj.DotSettings.user` it took the same time and with the same files changed.
> 
> image

You can run it at any level you like. You can run it on the solution, a project, a folder, a file or group of files, a namespace, a class, a function, or even just a specifically highlighted section of code. :)

I do the highlighted section bit a lot, because I know exactly which code will be affected and that nothing else will change anywhere else, keeping it nicely contained. Plus, it's useful if you're trying to avoid automatic re-formatting of something you explicitly wanted to be written with specific formatting, for readability or something.

dodexahedron:
@tig isn't this actually #3212's work?

I kept looking around for this comment chain after I got back to a computer (replies earlier today were from my phone) and wasn't able to find it until I went to my comment history.

.editorconfig
dodexahedron, line 1: Just a general comment on this file, not necessarily related to this commit, as I've not compared it line-by-line to the ReSharper dotsettings:

Any time .editorconfig or the ReSharper dotsettings files are modified, one should ensure that any corresponding changes supported by the other file are also made, to keep things consistent.

ReactiveExample/LoginViewModel.cs
```diff
@@ -6,72 +6,78 @@
...
+// For example, if you have a WPF or XF app with view models written
+// this way, you can easily port your app to Terminal.Gui by implementing
+// the views with Terminal.Gui classes and ReactiveUI bindings.
+//
+// We mark the view model with the [DataContract] attributes and this
+// allows you to save the view model class to the disk, and then to read
+// the view model from the disk, making your app state persistent.
+// See also: https://www.reactiveui.net../docs/handbook/data-persistence/
+//
+[DataContract]
```
dodexahedron, line 22: Non-review comment I wanted to add about this (but it makes it a review comment now that I've started one):

The JsonSerializer now respects DataContracts!

So, anywhere we currently use certain attributes specific to System.Text.Json can instead be these attributes, which enable broader serialization compatibility and consistency beyond just JSON.

dodexahedron (commented):
A general comment that I should probably make an issue for, for future follow-up but which isn't in-scope for this PR:

We need to be more careful with static members, especially when there is any path to those statics via the public API. They can be problematic in a lot of ways, especially if not synchronized or when exceptions happen.

dodexahedron:
Ooops. That review isn't done yet...

Terminal.Gui/Clipboard/Clipboard.cs
```diff
@@ -1,174 +1,178 @@
...
+                    if (clipData == null) {
+                        // throw new InvalidOperationException ($"{Application.Driver.GetType ().Name}.GetClipboardData returned null instead of string.Empty");
+                        clipData = string.Empty;
+                    }
+
+                    _contents = clipData;
+                }
+            }
+            catch (Exception) {
+                _contents = string.Empty;
```
dodexahedron, line 43: Out of scope for this PR, but this makes me 🤨 at a cursory glance.

dodexahedron:
Note:

As I'm going through the files, since github only lets you have one open review on a PR at a time, there will be a bunch of comments that are out of scope for the PR. I'm just opportunistically writing down thoughts as I look over the code, since this one is basically forcing me to look at everything.

Some might be generic commentary, some might warrant future issues/work, or whatever. If it's clearly out of scope for this PR and I didn't explicitly say so, just assume it is. Actually actionable items for this review should be pretty clear, or at least I'll try to make them so.

Terminal.Gui/Clipboard/Clipboard.cs
dodexahedron: Also, I've been thinking for a while now that we should put hooks for logging in the library, especially around exception handlers, and _especially_ those that eat exceptions.

We don't have to actually write the logs - it can be done via the standard logging infrastructure in modern .net, so anyone who wants logging can provide whatever logging implementation they wish.

Terminal.Gui/Clipboard/IClipboard.cs
```diff
@@ -1,40 +1,26 @@
...
-		/// <summary>
-		/// Sets the operation system clipboard.
-		/// </summary>
-		/// <param name="text"></param>
-		/// <exception cref="NotSupportedException">Thrown if it was not possible to set the clipboard contents.</exception>
-		void SetClipboardData (string text);
+    /// <summary>Gets the operation system clipboard if possible.</summary>
+    /// <param name="result">Clipboard contents read</param>
+    /// <returns>true if it was possible to read the OS clipboard.</returns>
+    bool TryGetClipboardData (out string result);
```
dodexahedron, line 20: (Out of scope)

We should put the [NotNullWhen(true)] attributes on any TryX or other relevant methods for static analysis.

Terminal.Gui/Clipboard/IClipboard.cs
```diff
@@ -1,40 +1,26 @@
...
-	/// </summary>
-	public interface IClipboard {
-		/// <summary>
-		/// Returns true if the environmental dependencies are in place to interact with the OS clipboard.
-		/// </summary>
-		bool IsSupported { get; }
+/// <summary>Definition to interact with the OS clipboard.</summary>
+public interface IClipboard {
+    /// <summary>Returns true if the environmental dependencies are in place to interact with the OS clipboard.</summary>
+    bool IsSupported { get; }
```
dodexahedron, line 6: (out of scope)

I realized something pretty obvious recently that I'm sure everyone implicitly already knows but which I just wanted to ensure awareness of...

Remember that an interface is a minimum contract. Implementers can define _more_ than the interface specifies, just not less.

So, an implementer could make a setter for IsSupported.

**It's their own fault**, but, if our code which uses the interface isn't at least written with that in mind, problems could occur.

That also should usually be part of the decision-making process around whether to put something in an interface, a base class, or both. If it's something that has a narrow definition that isn't likely to be different in any implementation, a base class might be a good idea.

OR, since c# 8, interfaces have now had the default implementation feature, which is pretty rad. With some minor restrictions, you can define an implementation for an interface member which will be used if an implementor does not explicitly provide their own. That could be handy for situations like this one, where this member is supposed to basically provide some sort of guarantee of underlying support for the interface's functionality. Consumers wishing to override that still can, by writing the property on their class as they otherwise would have, but it makes the default be at least mostly reliable, if it can be done in a simple way that covers a broad range of typical cases.

Terminal.Gui/Configuration/DictionaryJsonConverter.cs
dodexahedron, line 1: (out of scope comment/question)

I'm not entirely sure why this class exists. The built-in serializer can handle dictionaries already, in standard notation, usually even without many or any attributes.

Terminal.Gui/Configuration/Scope.cs
```diff
@@ -1,76 +1,65 @@
...
 
-#nullable enable
+namespace Terminal.Gui;
 
-namespace Terminal.Gui {
+/// <summary>
+///     Defines a configuration settings scope. Classes that inherit from this abstract class can be used to define scopes
+///     for configuration settings. Each scope is a JSON object that contains a set of configuration settings.
+/// </summary>
+public class Scope<T> : Dictionary<string, ConfigProperty> { //, IScope<Scope<T>> {
```
dodexahedron, line 10: (out of scope)

It is typically strongly recommended NOT to inherit from `Dictionary<TKey,TValue>`

Instead, the recommendation is to implement appropriate interfaces and for the custom type to _own_ a dictionary that is only exposed via the interface implementations.

Terminal.Gui/Configuration/ScopeJsonConverter.cs
```diff
@@ -1,140 +1,189 @@
-﻿using System;
-using System.Linq;
+﻿#nullable enable
+using System.Diagnostics;
```
dodexahedron, line 2: Since this one has appeared in a few places, as well, this should probably be a global using, as well.

Terminal.Gui/Configuration/ScopeJsonConverter.cs
```diff
@@ -1,140 +1,189 @@
...
-		writer.WriteEndObject ();
-	}
+    public override scopeT Read (ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) {
+        if (reader.TokenType != JsonTokenType.StartObject) {
+            throw new JsonException (
+                $"Expected a JSON object (\"{{ \"propName\" : ... }}\"), but got \"{reader.TokenType}\"."
+            );
+        }
+
+        var scope = (scopeT)Activator.CreateInstance (typeof (scopeT))!;
```
dodexahedron, line 22: (out of scope)

This should be generic all the way, as it calls the default constructor anyway. Therefore, the scopeT (which should also be TScope, by convention) should have a type parameter constraint including `new()` and then this line and similar should just call `new TScope()`. Important for AoT, too.

Terminal.Gui/Configuration/SerializableConfigurationProperty.cs
dodexahedron, line 1: (out of scope)

In modern c# (starting from 11), generics can be used in attributes, now.

This attribute type could benefit from that. Relevant to AoT, too.

Terminal.Gui/Configuration/SettingsScope.cs
```diff
@@ -1,121 +1,114 @@
...
 ///    "Theme" : "Default",
 ///    "Themes": {
 ///    },
 ///  },
-/// </code></example>
-/// <remarks>
-/// </remarks>
+/// </code>
+/// </example>
+/// <remarks></remarks>
```
dodexahedron, line 24: Empty remarks should be removed.

Terminal.Gui/Configuration/ThemeManager.cs
```diff
@@ -55,153 +48,127 @@ namespace Terminal.Gui;
...
-		SelectedTheme = "Default";
-	}
-
-	#region IDictionary
+    static ThemeManager () { } // Make sure it's truly lazy
+    private ThemeManager () { } // Prevent instantiation outside
+    private static string _theme = string.Empty;
+
+    /// <summary>Holds the <see cref="ThemeScope"/> definitions.</summary>
+    [JsonInclude]
```
dodexahedron, line 59: (out of scope)

Statics included in serialization can have unexpected consequences.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
dodexahedron, line 1: I'm skipping the files in this folder for purposes of this review, because they're already getting worked on in one of my active work items.

Terminal.Gui/Drawing/Color.Formatting.cs
```diff
@@ -7,418 +7,562 @@
...
+    ///     </para>
+    /// </remarks>
+    [SkipLocalsInit]
+    public string ToString (
+        [StringSyntax (StringSyntaxAttribute.CompositeFormat)] string? formatString,
+        IFormatProvider? formatProvider = null
+    ) {
+        return (formatString, formatProvider) switch {
+                   // Null or empty string and null formatProvider - Revert to 'g' case behavior
+                   (null or { Length: 0 }, null) => ToString (),
```
dodexahedron, line 97: Hmmm...

I am not sure if I like how the current ruleset results in these getting all these extra newlines.

I don't think it's worth bothering with trying to fix that in this PR, since it's not yet a particularly common situation, but next time I run into it I'll see if there is support for this specific situation.

Terminal.Gui/Drawing/ColorParseException.cs
```diff
@@ -3,90 +3,87 @@
...
-	/// </param>
-	/// <param name="reason">The reason that <paramref name="badValue" /> failed to parse.</param>
-	internal ColorParseException (in ReadOnlySpan<char> colorString, string? badValue = null, string? badValueName = null, string? reason = null) : base (DefaultMessage)
-	{
-		ColorString = colorString.ToString ();
-		BadValue = badValue;
-		BadValueName = badValueName;
-		Reason = reason;
-	}
+    internal const string DefaultMessage = "Failed to parse text as Color.";
```
dodexahedron, line 52: (out of scope)

Should I consider not using a const here, and making this localizable?

Terminal.Gui/Drawing/Color.cs
```diff
@@ -10,273 +10,275 @@
...
-	/// <summary>The bright magenta color.</summary>
-	public const ColorName BrightMagenta = ColorName.BrightMagenta;
-
-	/// <summary>The bright yellow color.</summary>
-	public const ColorName BrightYellow = ColorName.BrightYellow;
-
-	/// <summary>The White color.</summary>
-	public const ColorName White = ColorName.White;
-	#endregion
+public readonly partial record struct Color : ISpanParsable<Color>, IUtf8SpanParsable<Color>, ISpanFormattable,
```
dodexahedron, line 20: Looks like the rules didn't chop this one properly.

The intent was for long lists to get chopped to one per line. Should be fairly easily fixable.

Terminal.Gui/Drawing/ColorScheme.cs
```diff
@@ -4,249 +4,210 @@
...
+    /// <param name="attribute">The attribute to initialize the new instance with.</param>
+    public ColorScheme (Attribute attribute) {
+        _normal = attribute;
+        _focus = attribute;
+        _hotNormal = attribute;
+        _disabled = attribute;
+        _hotFocus = attribute;
+    }
+
+    private readonly Attribute _disabled;
```
dodexahedron, line 43: Are the layout rules correct here?

Don't we want private fields to be first or, if they back a property, directly above their associated properties?

Terminal.Gui/Drawing/Glyphs.cs
dodexahedron, line 1: (out of scope question)

Is this potentially more appropriate as a static type?

Terminal.Gui/Drawing/StraightLineExtensions.cs
dodexahedron, line 1: (out of scope)

Can't/shouldn't several of these just be static member methods of StraightLine?

Terminal.Gui/Drawing/Thickness.cs
```diff
@@ -1,298 +1,273 @@
...
+        Right = right;
+        Bottom = bottom;
+    }
+
+    /// <summary>Gets or sets the width of the lower side of the rectangle.</summary>
+    [JsonInclude]
+    public int Bottom;
+
+    /// <summary>Gets or sets the width of the left side of the rectangle.</summary>
+    [JsonInclude]
```
dodexahedron, line 45: (out of scope)

Fields on reference types typically shouldn't be public.

Either this type should be a `struct`/`record struct` (likely) or these should be exposed as properties for the public API surface.

Terminal.Gui/FileServices/AllowedType.cs
```diff
@@ -1,116 +1,82 @@
...
-	}
-	
-}
\ No newline at end of file
+﻿using Terminal.Gui.Resources;
+
+namespace Terminal.Gui;
+
+/// <summary>Interface for <see cref="FileDialog"/> restrictions on which file type(s) the user is allowed to select/enter.</summary>
+public interface IAllowedType {
```
dodexahedron, line 6: Should be in its own file.

dodexahedron (changes requested):
Alrighty.

Submitting this review as-is for now.

After going through several dozen more files without making out of scope comments on them in the interest of time, I haven't yet found any additional in-scope tweaks that should be made that I haven't already commented on and would fix in all affected files if applied.

Most comments in this review are out of scope, but there are a few that aren't. Those are the ones that don't say out of scope at the top of them.

Terminal.Gui/Drawing/ColorScheme.cs
tig: Can you tell what setting is in correct. I thought I had this right.

tig:
> @tig isn't this actually #3212's work?
> 
> I kept looking around for this comment chain after I got back to a computer (replies earlier today were from my phone) and wasn't able to find it until I went to my comment history.

@dodexahedron Because both myself and @BDisp touched so many files, the only way I could find to merge his PR into this was to apply formatting solution wide. I probably could have not done that but i did, and once i did there was no going back.

So, yes, this is #3121's work, but it got done here. 

I am not confident I have the R# and .editiorconfig settings right (enough) to merge. I was hoping you'd look through this PR at just the code cleanup/formatting stuff and help me get it good enough that we can merge this (forgoing #3212).

.editorconfig
tig: Is there a way to configure R# to always make .editorconfig be in sync?

tig:
I got over wanting to have opening braces on line above (in the spirt of "don't be weird").

Reformatting again now and will push again asap.

Terminal.Gui/Drawing/ColorScheme.cs
dodexahedron: That'll be in the file layout configuration. There's also an extension that adds even more capabilities to that functionality.

I was pretty sure we'd end up with at least a couple things not being exactly as we wanted, and this part was likely to be one of them.

That stuff is powerful. But, with that power, comes a fair amount of complexity in getting it right. Then you pretty much never have to touch it again until the language changes enough to matter.

.editorconfig
dodexahedron: On one of the settings screens you can export to and import from .editorconfig, as well as set what its general behavior should be.

tig:
I think I have the issues with two different machines figured out. R# seems to get confused and sometimes saves settings to the machine wide config here: C:\Users\charlie\AppData\Roaming\JetBrains\Shared\vAny

Also, the various code cleanup configs seemed to be confusing each other (or me). I've now deleted the non-built-in ones we created and am moving forward with just those. It seems to have simplified things.

For the first time I was able to do a new `git clone` of the repo, checkout this branch and hit CTRL-E-CTRL-C on a file I knew was complex and "just right" (e.g. UICatalog.cs or ViewLayout.cs) and have it cause no changes.

Running a full cleanup now. On my lappy so it's gonna take a looong time...

dodexahedron:
> I got over wanting to have opening braces on line above (in the spirt of "don't be weird").
> 
> Reformatting again now and will push again asap.

😅

All good. It does seem like one has to fight with things less and there are fewer corner cases for formatting (especially with new language constructs) if it's on its own line. 🤷‍♂️

dodexahedron:
> I think I have the issues with two different machines figured out. R# seems to get confused and sometimes saves settings to the machine wide config here: C:\Users\charlie\AppData\Roaming\JetBrains\Shared\vAny
> 
> Also, the various code cleanup configs seemed to be confusing each other (or me). I've now deleted the non-built-in ones we created and am moving forward with just those. It seems to have simplified things.
> 
> For the first time I was able to do a new `git clone` of the repo, checkout this branch and hit CTRL-E-CTRL-C on a file I knew was complex and "just right" (e.g. UICatalog.cs or ViewLayout.cs) and have it cause no changes.
> 
> Running a full cleanup now. On my lappy so it's gonna take a looong time...

Yep that's usually what does it.  If I'm working on a project with significantly different settings that are too much of a pain for me to fight with the formatting engine, I just copy the team settings over the machine settings from the start, usually. Plus that nukes any local settings I may have tweaked during other work that I had originally _intended_ to be temporary but then just forgot to clear.

tig:
@dodexahedron before I merge, can you please check that I have things like `var` and `null assignment` correct (or at least not stupidly wrong)?

dodexahedron:
Ha sure thing. Won't be at a PC til later, though.

Terminal.Gui/Configuration/AttributeJsonConverter.cs
```diff
@@ -37,7 +37,7 @@ public override Attribute Read (ref Utf8JsonReader reader, Type typeToConvert, J
         {
             if (reader.TokenType == JsonTokenType.EndObject)
             {
-                if ((foreground == null) || (background == null))
+                if (foreground == null || background == null)
```
dodexahedron, line 40: Null check should be ` if (foreground is not {} || background is not {} `

If ReSharper didn't do that, then either the cleanup profile that was run doesn't have that option checked or the style settings for c# don't have the null check pattern properly defined.

Terminal.Gui/Configuration/ConfigurationManager.cs
```diff
@@ -390,7 +390,7 @@ internal static void AddJsonError (string error)
         }
 
         // If value type, just use copy constructor.
-        if (source.GetType ().IsValueType || (source.GetType () == typeof (string)))
+        if (source.GetType ().IsValueType || source.GetType () == typeof (string))
```
dodexahedron, line 393: For the right-hand side of this, which is a type check, it should be the is operator as well, but I'm not sure if ReSharper has a built-in rule for that, specifically. I'll check and report back here.

Terminal.Gui/Configuration/SettingsScope.cs
```diff
@@ -102,7 +102,7 @@ public class SettingsScope : Scope<SettingsScope>
     /// <param name="resourceName"></param>
     public SettingsScope? UpdateFromResource (Assembly assembly, string resourceName)
     {
-        if ((resourceName == null) || string.IsNullOrEmpty (resourceName))
+        if (resourceName == null || string.IsNullOrEmpty (resourceName))
```
dodexahedron, line 105: Redundant and improper null check.

This should probably be the `ArgumentException.ThrowIfNullOrEmpty(resourceName);` helper both for brevity and because those helper methods actually have a run-time advantage over throwing it manually.

And then callers should be prepared to handle the exception.

Also, note that resources are one of the things that are potentially not AoT-friendly. I began looking into that a few days ago but life stuff got in the way and then I just haven't come back to that, yet, favoring working on outstanding work instead.

Terminal.Gui/ConsoleDrivers/ConsoleKeyMapping.cs
```diff
@@ -31,12 +31,12 @@ public static class ConsoleKeyMapping
...
     /// </returns>
     [DllImport ("user32.dll", EntryPoint = "GetKeyboardLayout", CharSet = CharSet.Unicode)]
-    private extern static nint GetKeyboardLayout (nint idThread);
+    private static extern nint GetKeyboardLayout (nint idThread);
 
     //[DllImport ("user32.dll", EntryPoint = "GetKeyboardLayoutNameW", CharSet = CharSet.Unicode)]
     //extern static uint GetKeyboardLayoutName (uint idThread);
-    [DllImport ("user32.dll")] private extern static nint GetForegroundWindow ();
-    [DllImport ("user32.dll")] private extern static nint GetWindowThreadProcessId (nint hWnd, nint ProcessId);
+    [DllImport ("user32.dll")] private static extern nint GetForegroundWindow ();
```
dodexahedron, line 38: I had considered making a rule in ReSharper targeting DllImport and LibraryImport library names to normalize them on extensionless names, as is the recommendation. The runtime does platform-appropriate name mangling on library names. It's less relevant for Windows, but quite relevant for linux, since different distros use different naming conventions for the same libraries, and just making the two follow the same scheme is nice for consistency anyway.

This sort of rule would probably be pretty easy to create and I'm fairly certain ReSharper can be told to apply custom rules when formatting code.

Terminal.Gui/ConsoleDrivers/CursesDriver/ClipboardImpl.cs
```diff
@@ -5,11 +5,10 @@ namespace Terminal.Gui;
 
 /// <summary>A clipboard implementation for Linux. This implementation uses the xclip command to access the clipboard.</summary>
 /// <remarks>If xclip is not installed, this implementation will not work.</remarks>
-class CursesClipboard : ClipboardBase
+internal class CursesClipboard : ClipboardBase
```
dodexahedron, line 8: I don't think the code cleanup engine is allowed to separate things to one type per file, but I'm noting this file has multiple types.

I'll look for that in the settings, too, when I'm looking for the type check operator refactoring setting.

Terminal.Gui/ConsoleDrivers/CursesDriver/ClipboardImpl.cs
```diff
@@ -168,33 +166,32 @@ private bool CheckSupport ()
     }
 
     [DllImport ("/System/Library/Frameworks/AppKit.framework/AppKit")]
```
dodexahedron, line 168: Explicit paths are strongly discouraged.

Is there a hard requirement for these to be explicit? If so, then we should ensure this is loudly documented as a dependency.

In general, though, the name of the library is typically sufficient and it will use the local system's LD_LIBRARY_PATH location(s). If the dependency isn't in one of those paths, the user is typically expected to add the necessary paths to that env var.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -1010,6 +1010,6 @@ private static int GetSuspendSignal ()
         }
     }
 
-    [DllImport ("libc")] private extern static int killpg (int pgrp, int pid);
-    [DllImport ("libc")] private extern static int uname (nint buf);
+    [DllImport ("libc")] private static extern int killpg (int pgrp, int pid);
```
dodexahedron, line 1013: Hm.

These attributes shouldn't be on the same line.

But now I just remembered you said you did this with just default configuration, yeah? That'll result in a few non-standard things like this.

Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
```diff
@@ -76,9 +74,9 @@ static UnmanagedLibrary ()
         IsMacOSPlatform = platform == PlatformID.Unix && GetUname () == "Darwin";
         IsLinux = platform == PlatformID.Unix && !IsMacOSPlatform;
 
-        IsWindows = (platform == PlatformID.Win32NT)
-                    || (platform == PlatformID.Win32S)
-                    || (platform == PlatformID.Win32Windows);
+        IsWindows = platform == PlatformID.Win32NT
```
dodexahedron, line 77: .net 8 made some improvements to this and has some new types for more comprehensive platform identification than before. Might be worth checking out later on.

Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
```diff
@@ -264,20 +261,20 @@ private static string FirstValidLibraryPath (string [] libraryPathAlternatives)
 
     private static class Windows
     {
-        [DllImport ("kernel32.dll")] internal extern static nint GetProcAddress (nint hModule, string procName);
-        [DllImport ("kernel32.dll")] internal extern static nint LoadLibrary (string filename);
+        [DllImport ("kernel32.dll")] internal static extern nint GetProcAddress (nint hModule, string procName);
```
dodexahedron, line 264: More attributes on same line as method signatures and extensions in library names.

dodexahedron: I'm going to stop commenting on that particular problem, since one settings tweak will fix them all at once, so it's not really necessary for me to call them all out.

Terminal.Gui/ConsoleDrivers/CursesDriver/binding.cs
```diff
@@ -93,7 +92,7 @@ public static int Lines
     //
     public static int addch (int ch)
     {
-        if ((ch < 127) || (ch > 0xffff))
+        if (ch < 127 || ch > 0xffff)
```
dodexahedron, line 95: I think there's a ReSharper rule for these but I'm not certain.

In any case, this sort of thing can be `if (ch is < 127 or > 0xffff)`, which the compiler is capable of optimizing better.

Terminal.Gui/ConsoleDrivers/EscSeqUtils/EscSeqUtils.cs
```diff
@@ -461,7 +461,7 @@ public static ConsoleModifiers GetConsoleModifiers (string value)
     /// <returns>The c1Control returned by <see cref="GetC1ControlChar(char)"/>, code, values and terminating.</returns>
     public static (string c1Control, string code, string [] values, string terminating) GetEscapeResult (char [] kChar)
     {
-        if ((kChar == null) || (kChar.Length == 0))
+        if (kChar == null || kChar.Length == 0)
```
dodexahedron, line 464: Another that might be able to be handled by a rule:

`if (kChar is not { Length: > 0})`

or using spans:

`if (kChar is not [_,..])`

Terminal.Gui/ConsoleDrivers/EscSeqUtils/EscSeqUtils.cs
```diff
@@ -807,10 +807,10 @@ public static (string c1Control, string code, string [] values, string terminati
 
         if ((!isButtonClicked
              && !isButtonDoubleClicked
-             && ((buttonState == MouseFlags.Button1Pressed)
-                 || (buttonState == MouseFlags.Button2Pressed)
-                 || (buttonState == MouseFlags.Button3Pressed)
-                 || (buttonState == MouseFlags.Button4Pressed))
+             && (buttonState == MouseFlags.Button1Pressed
```
dodexahedron, line 810: Pretty sure this isn't something ReSharper will automatically do, but these can be greatly simplified and turned into a single comparison.

Terminal.Gui/ConsoleDrivers/NetDriver.cs
```diff
@@ -1157,7 +1155,7 @@ private void ProcessInput (InputResult inputEvent)
 
     #region Size and Position Handling
 
-    volatile private bool _winSizeChanging;
+    private volatile bool _winSizeChanging;
```
dodexahedron, line 1158: This is only thread-safe on the same logical CPU. Other threads or even the same thread, if it gets shunted to another CPU, are not affected by volatile, which just puts a memory barrier around accesses to the variable.

Terminal.Gui/ConsoleDrivers/NetDriver.cs
```diff
@@ -1685,7 +1683,7 @@ bool IMainLoopDriver.EventsPending ()
 
         if (!_eventReadyTokenSource.IsCancellationRequested)
         {
-            return (_resultQueue.Count > 0) || _mainLoop.CheckTimersAndIdleHandlers (out _);
+            return _resultQueue.Count > 0 || _mainLoop.CheckTimersAndIdleHandlers (out _);
```
dodexahedron, line 1686: If possible, we should provide an override without the out param. Discards still result in a wrapper method and passing the parameter around. Minimal impact if only reference types. Potentially tangible impact if value types can be here.

Terminal.Gui/Drawing/ColorScheme.cs
```diff
@@ -52,19 +52,39 @@ public ColorScheme (Attribute attribute)
...
-    public Attribute Focus { get => _focus; init => _focus = value; }
+    public Attribute Focus
+    {
+        get => _focus;
+        init => _focus = value;
+    }
 
     /// <summary>The foreground and background color for for text in a focused view that indicates a <see cref="View.HotKey"/>.</summary>
-    public Attribute HotFocus { get => _hotFocus; init => _hotFocus = value; }
+    public Attribute HotFocus
```
dodexahedron, line 69: ReSharper is capable of turning these into auto-properties, with appropriate settings.

Terminal.Gui/Drawing/LineCanvas.cs
```diff
@@ -413,13 +413,13 @@ private void ConfigurationManager_Applied (object? sender, ConfigurationManagerE
         bool useDouble = intersects.Any (i => i?.Line.Style == LineStyle.Double);
 
         bool useDashed = intersects.Any (
-                                         i => (i?.Line.Style == LineStyle.Dashed)
-                                              || (i?.Line.Style == LineStyle.RoundedDashed)
+                                         i => i?.Line.Style == LineStyle.Dashed
```
dodexahedron, line 416: Comparisons against possible nulls. These should be revisited at some point (probably very non-critical at the moment).

Terminal.Gui/Input/Key.cs
```diff
@@ -259,7 +259,7 @@ public Key (string str)
     /// </remarks>
     public static bool GetIsKeyCodeAtoZ (KeyCode keyCode)
     {
-        if (((keyCode & KeyCode.AltMask) != 0) || ((keyCode & KeyCode.CtrlMask) != 0))
+        if ((keyCode & KeyCode.AltMask) != 0 || (keyCode & KeyCode.CtrlMask) != 0)
```
dodexahedron, line 262: These can be combined into a single comparison against the bitwise or of the flags to save several operations and stack churn.

Terminal.Gui/Input/KeyBinding.cs
```diff
@@ -97,7 +97,7 @@ public void Add (Key key, KeyBindingScope scope, params Command [] commands)
             throw new ArgumentException (@"At least one command must be specified", nameof (commands));
         }
 
-        if ((key == null) || !key.IsValid)
+        if (key == null || !key.IsValid)
```
dodexahedron, line 100: Another case of improper null check but also place to use an object pattern match, avoiding the binary or.

Terminal.Gui/Input/ShortcutHelper.cs
```diff
@@ -162,7 +162,7 @@ private static string GetKeyToString (KeyCode key, out KeyCode knm)
         uint.TryParse (knm.ToString (), out uint c);
```
dodexahedron, line 162: TryParse is cool, but later checks can be simplified by using the return value.

Terminal.Gui/Text/Autocomplete/PopupAutocomplete.cs
```diff
@@ -187,7 +187,7 @@ public override bool ProcessKey (Key a)
             return ReopenSuggestions ();
         }
 
-        if (closed || (Suggestions.Count == 0))
+        if (closed || Suggestions.Count == 0)
```
dodexahedron, line 190: Opportunity for use of multi-condition switch statement. (`switch ( closed, Suggestions )`)

Terminal.Gui/Text/TextFormatter.cs
```diff
@@ -364,7 +392,7 @@ public virtual string Text
             int x, y;
 
             // Horizontal Alignment
-            if ((Alignment == TextAlignment.Right) || (Alignment == TextAlignment.Justified && !IsLeftToRight (Direction)))
+            if (Alignment == TextAlignment.Right || (Alignment == TextAlignment.Justified && !IsLeftToRight (Direction)))
```
dodexahedron, line 395: Object pattern would simplify this.

dodexahedron (changes requested):
I'm about half way done according to the scroll bar on this beast.

Submitting this now because some of the suggested changes will fix the majority of what I've seen and continue to see, which is mostly the null checking pattern not being correct.

dodexahedron:
@tig I submitted a half-finished review, but the small handful of rule changes to address the in-scope issues I have found so far will go a long way toward making it all good.

dodexahedron:
And, just to be clear, any I didn't mention ReSharper for aren't necessarily in scope for this and are just notes.

I'll check this branch out and make rule changes to address and push those so you can pull them in and keep the reformat activity consolidated here.

dodexahedron:
Hmmm...

Well, I made the rules for null check pattern, but the code cleanup isn't actually applying them in all places.

I suspect it just backs off for safety, if the control flow statement is more than JUST the null check by itself, but that's a guess based on what I'm observing.

We'll probably have to address those manually, unless we can find a way to make it do it automatically in more places.

dodexahedron:
There ya go.

Sent you a PR on your branch for a small set of rules

dodexahedron:
If you run that on your machine for a file that got the order if fields and stuff wrong, does it look like it at least resulted in improving that situation?

I'm not very optimistic about it fixing many or maybe even any of the remaining null checks automatically.

I'm actually filing a bug/suggestion regarding that, because they seem like such simple and obvious cases for it to handle and I'm surprised it's not doing it.

dodexahedron:
Oh and note the profile to use is the "All Rules" profile, because the default profiles ignore code that is excluded by current preprocessor evaluations. This profile applies the same logic to them, as well, regardless of how they evaluate.

dodexahedron:
I've got some more tweaks queued up for the settings, but they're mostly for inspections and thus aren't directly related to this work, since they won't result in a change in code cleanup behavior. They're just for providing additional analyzer tool tips and refactoring actions that can be used or ignored when working on something.

As we make things better and better, we should probably consider turning the severity of the more valuable ones higher. Maybe after we declare v2 solidly beta quality or at whatever point we call a feature freeze?

dodexahedron:
Saw the push.

I'll go back over again in a few minutes.

The light at the end of the tunnel is getting closer!

Terminal.Gui/ConsoleDrivers/CursesDriver/ClipboardImpl.cs
dodexahedron: I didn't see anything that would enable this for automatic code cleanup, so it's just something we'll have to invoke wherever the analyzer complains about it. Oh well.

tig:
NOW. ;-)

dodexahedron:
Nice.

It looks like it took care of a lot of the field re-ordering, with one exception I'll have to fiddle with to get it to behave properly.

The layout rules do specify putting backing fields by their properties, and that rule has a high priority, so I'm not entirely sure without poking it some more why it isn't doing that specific bit.

Otherwise, this is looking a lot better so far. :)

dodexahedron:
That one is so minor, though, that I'm kinda inclined to say, unless we can figure it out in short order, we can leave that for later. The diffs caused by that when code with that issue is touched are just single-line moves, and so are easy to eyeball.

tig:
I'm done with this. Merging!! :-)

dodexahedron:
For sake of completeness and consistency, I just sent you another PR for your branch with .editorconfig synced to dotsettings

tig:
We can do that in another PR. 

I'm shutting down for the superbowl. Cheers!

dodexahedron:
Haha good plan all around.

Have fun!

dodexahedron:
And also, thanks for tackling this monster!

It's rare that something this big gets such a big purely code-quality changeset like this. Greatly appreciated. 😊

I'll make a new issue/PR for the editorconfig stuff and take a quick pass over the docs to be sure they match what we actually did, in the end.

PR #3204, opened 2024-01-22
3170 Color improvements

dodexahedron:
## Fixes
 - A bunch. See #3170. Most center around performance.
 - Implementation of several interfaces for Color

## Proposed Changes/Todos

- [x] Significant re-working of Color, especially around formatting and parsing
- [ ] (WIP) Implement additional tests to cover the rest of the new code plus refactoring old tests

## Pull Request checklist:

- [x] I've named my PR in the form of "Fixes #issue. Terse description."
- [x] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [x] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [x] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [x] My changes generate no new warnings
- [x] I have checked my code and corrected any poor grammar or misspellings
- [x] I conducted basic QA to assure all features are working

PR #3207, opened 2024-01-23
Applies Code Cleanup to Entire v2 Solution

tig:
## Fixes

Instead of continuing to peacemeal code cleanup and cause tons of ongoing merge issues, perhaps the right thing to do is a PR that applies CC to the entire solution.

That's what this PR does. 

## Todos

- [ ] Implement [JetBrains' FREE CleanupCode Command Line Tool](https://www.jetbrains.com/help/resharper/CleanupCode.html) in CI/CD as a requirement for merging a PR into  `v2_develop`
- [ ] Update `.editorconfig` based on agreed upon rules (see below).
- [ ] Merge `v2_develop` into this branch
- [ ] Apply Code Cleanup to entire project
- [ ] Double check all tests pass and nothing broke


tig:
@BDisp, @dodexahedron ...

Y'all have "bigish" PRs in progress.

I would like to take this opportunity to do a massive code clean up. I have been doing it peacemeal and it has caused tons of merge pain. 

To get past this and get v2 in the shape I want it in, here's the plan:

1. I do this PR that does a code cleanup in the entire solution.
2. I pull down each of your branches and take care of merging this branch into them... I'll submit a PR to each.
3. I'll do the same with my several outstanding PRs

Ok?

dodexahedron:
Sounds reasonable to me.

dodexahedron:
For mine, it might be easier if you merge my branch into your working branch, instead, taking my copy for all conflicts, and then formatting the result on your branch.

I made sure the Color stuff had no conflicts with v2_develop when I put that PR in, so just a thought. 🤷‍♂️

tig:
> For mine, it might be easier if you merge my branch into your working branch, instead, taking my copy for all conflicts, and then formatting the result on your branch.
> 
> I made sure the Color stuff had no conflicts with v2_develop when I put that PR in, so if you merge before you format there won't be any. Just a thought. 🤷‍♂️
> 
> And then I can just branch again from whatever you merge back to v2_develop.
> 
> The color PR is basically ready to go. It already has near 100% coverage. The only remaining un-covered code is just a reversed dictionary I added when working on it, but which ended up not actually being needed in the end. If I remove that, it'll be at 100%, with 2377 test cases.

Good plan. Thanks.

tig:
I've always disliked @migueldeicaza's love of 8 chars for tabs. I understand there's a history from Mono, but it's very uncommon and uses a ton of horizontal screen real-estate. Recently I found a bug in Reshaper (that @bdisp) also sees that really screws up XML API docs because of it.

As part of this PR, I'd like to change `.editorconfig` to make our standard tab size 4 just like most of the known dotnet World. This would only be for the v2 codebase, not v1.

Anyone object?

- @tznind ?
- @BDisp ?
- @dodexahedron ?

dodexahedron:
Following on from some brief discussion in #3204, this issue is likely best served by first crafting a configuration for ReSharper's formatting and cleanup functionality to match desired output, sharing that file in source control, and then otherwise carrying out the planned actions in the top post of this issue, but via use of ReSharper instead of native VS, so additional nice-to-haves like sorting and other code _layout_ operations can be done as well.

Once we have a good configuration, I suggest we also then have resharper write out a .editorconfig file based on its settings (which is something it can do), so that settings for Visual Studio users both with and without Rider/ReSharper will have as consistent an experience and behaviors as possible..

Note that Rider is also cross-platform, and Linux dotnet devs seem to absolutely love it. I personally have mixed opinions of Rider vs VS+ReSharper, but any Linux or Mac folks who want to work on TG from their respective environments and don't have a Rider license can also use VSCode, which is capable of handling .editorconfig files, too. So, everyone should be covered. :)

dodexahedron:
So, how do you want to do this?

I feel like a Teams/whiteboard/whatever session would be good for this one, so we can knock it out quickly and with minimal extra steps and latency. That something you'd be amenable to in the next few days? I'm pretty flexible, so your schedule can dictate timing, as far as I'm concerned, if you want to give it a go.

dodexahedron:
Oh!

Also!

To allay your fears of folks not being able to use the same formatting rules, I present to you [JetBrains' FREE CleanupCode Command Line Tool](https://www.jetbrains.com/help/resharper/CleanupCode.html).

That tool is free for use by anyone and can apply the rules specified in the dotsettings files, just like ReSharper would have.

A thought I had was that maybe we could add a git action for PRs to run it on code that's ready for review, if github can do that? If so, that'd mean that, no matter how someone commits the code, what ends up in the master repo will follow the settings defined for the project.

dodexahedron:
> I've always disliked @migueldeicaza's love of 8 chars for tabs. 
Hard, emphatic, and jubilant agree. 😅

I'm also a fan of using spaces instead of tabs, for a lot of reasons, not the least of which is consistent layout in all dev environments, including plain old text editors that have no concept of variable width tabs.

And with modern Visual Studio as well as with ReSharper, it ends up being completely transparent to you when working on the code, because it just does it.

Also, somewhat more than tangentially related (secant-ly related?), are you familiar with/aware of Virtual Formatting capabilities in Resharper (and I think also to a much more limited degree, VS2022 as well)?

That feature allows you to view and work on code that is presented to _you_ how _you_ personally prefer, even if that's wildly different from project settings, while actually writing the text to the physical file with formatting settings as dictated for the project.

It's something that isn't really typically saved to the shared team settings, and instead saved at the user level for the project or solution, or at the machine level (which it will use as your personal defaults for any project you open). In the current release of ReSharper, its a lot easier to set up and more reliable than it used to be. Just something you or anyone else might find handy. :)

tig:
> So, how do you want to do this?
> 
> I feel like a Teams/whiteboard/whatever session would be good for this one, so we can knock it out quickly and with minimal extra steps and latency. That something you'd be amenable to in the next few days? I'm pretty flexible, so your schedule can dictate timing, as far as I'm concerned, if you want to give it a go.

I'm a big fan of getting aligned at a principled level for things where there's likely to be lots of opinions and debate. Far better to drive robust debate on a set of written-down [tenets](https://blog.kindel.com/2020/02/10/tenets/) (rules) first.  

## Tenets for [gui-cs](www.github.com/gui-cs) Code Style (Unless you have better ones)

* **Six-Year-Old Reading Level** - Our code style is biased towards code readability and away from terseness. This is *Systems Software* and needs to stand the test of time. Code should be structured and use variable names that make it readable by a 6-year-old, and comments in code are encouraged. 
* **Consistency, Consistency, Consistency** - We adopt and document our standards for code style and then enforce them ruthlessly. For example, we require code reviews to pay attention to code style, not just functionality. 
* **Don't be Weird** - Like all developers we have opinions, but our opinions on code style are tempered by existing standards. We are biased towards code style that used by Microsoft and other leading dotnet developers. For example, we choose 4 spaces for indentation instead of 8.
* **Set and Forget** - We embrace and encourage the use of technology that makes it easy for contributors to apply best-practice code-style, such as ReSharper. As we do so we are mindful that tools can cause hidden issues and merge hell.
* **Documentation is the Spec** - We care deeply about providing delightful developer documentation and are sticklers for grammar and clarity. If the code and the docs conflict, we are biased to believe what we wrote in the API documentation. This drives a virtuous cycle of clear thinking.

##  [gui-cs](www.github.com/gui-cs) Code Style Rules

Our code style rules are codified in `.editorconfig`. The list below provides a high-level summary of what is (should be) found in `.editorconfig`.

* Use vertical whitespace (one is enough) to break code apart.
* 4 spaces per-indent.
* Spaces, not tabs.
* CamelCase for `public` stuff.
* camelCase for local stuff and parameters.
* `private` class members are prefixed with `_`; `this.` is rarely used.
* Numerical literals should be defined with a human readable `const` or equivalent. 
* Non-diagnostic/debug string literals go in `Strings.resx`.
* Backing fields are declared near the `public` properties that use them.
* Do not place opening brace on next line.
* Prefer modern `switch` patterns over complex `if` statements
* Prefer modern pattern matching

(That's enough for me for now. The above was written to ENCOURAGE disagreement and debate in this thread. Fight me, dammit!).

BDisp:
> * Spaces, not tabs.

I agree with all. Only a doubt about using spaces and not tabs. When I have a long statement and press enter to write the continuation of it, normally it's automatically add a tab, but sometimes after comment and uncomment the indentation is broken, even when I press `Ctrl-K, Ctrl-D`. So I normally I select the lines needed and press Tab or Shift-Tab to reduce indentation. If only spaces are required then I think that is a Git config option, right?

dodexahedron:
I meant how do you want to do this in regards to actually carrying out the changes. Especially since you wanted some guidance with wielding ReSharper effectively, I figured we could save a lot of time and effort by doing it live.

dodexahedron:
Also...

I never learn....

I had a big comment written up that had some thoughts and a few proposals (most of which were just putting things in writing that aren't already), and then screwed up by navigating away on the same tab....  \*sigh\*

I'll write it up again another time, when I get over it. 😅

dodexahedron:
To preface that eventual comment, though, I'll link these documents.

These are meant to be informational, not as a list of "violations" or anything. We already follow most of it. These just have relevance to the stuff I had written up before.

 - [C# Coding Conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions)
   - We only really explicitly go against the bracing style, which is standard to be all control flow braces on their own line
 - [C# Naming Conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names)
 - [dotnet project style rules](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md)
   - Heh. Notice item 1 is the brace thing
 - [dotnet naming guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines)
   - [Type naming](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces)
   - [Type member naming](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members)
 - [Nested type guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types)
 - [Event guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/event)

dodexahedron:
OMG yay!

Github itself saved the day somehow.

I went to the PR list again and then literally clicked the link to come back to this PR because I wanted to look at your revised top post again, and my previous comment text was pre-filled here! 🥳

Anyway...

Here it is:

I meant how do you want to handle actually _doing_ it?

Since you wanted some guidance with wielding ReSharper and all, I just figured we could save a bunch of time and probably also a bunch of unnecessary merging and, more importantly, conflict resolution activity by doing it live instead of a bunch of back and forth.

As for the actual style guidelines, those are all actually pretty typical and in-line with recommended baselines. The _only_ one that is atypical for C\# is the K&R-style curly braces (opening on same line). The c\# style guide recommends Allman/BSD-style, which is all control flow braces on their own line. But that one, honestly, rates like a 2 out of 10 on my care-o-meter.

Otherwise, it does pretty much match the recommendations.

https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#style-guidelines

The main gripe I have with those recommendations is that their suggested line wrapping length is absurdly short and feels like something from the 80s (and they break it themselves a lot).

Some of our expansions on that happen to be things that resharper can help enforce, too. :) For example, keeping backing fields by their properties is an option in the ReSharper settings, so getting that into the dotsettings files will make life even easier.

Here are some tweaks I'd suggest to the current project policies, in no particular order. Some are just putting existing rules into writing, some are for bringing it in line with baselines, some address actual problems, and some are to encourage or enforce things that aren't covered yet:

 - Require explicit access modifiers everywhere (ie no implicit `private`).
   - This is different from the current .editorconfig, which prefers omitting `private`
 - `var` usage either not enforced or enforced opposite of how it currently is in .editorconfig.
   - There are real reasons for this, in some places, which can actually result in compiler output not matching developer intent, especially around some of the C\# 12 features and Spans, where the compiler picks "safe" overloads, which are not necessarily the "best"
   - The dotnet code itself only uses `var` when the type is explicitly named on the right (ie only for `new` statements, and never when it's just the return type of the right).
   - It's point 10 in the dotnet runtime project guidelines [here](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md), which also points out that it prevents use of other modern language constructs like target-typed `new`, which is generally a nice space-saver, if nothing else.
 - Strongly suggest (but not require) use of `nameof` in strings that reference code elements by name.
   - Mostly, this is for constants like attributes, to avoid breakage.
   - I suggest not requiring it because it may have an unacceptable performance impact in certain situations.
 - Recommend use of thread-safe and/or observable collections, when possible
 - Recommend use of `record` types when value equality is desired.
 - Require use of `typeof` when possible, instead of `object.GetType()`
 - One type per file
 - Require file-scoped namespaces
 - Use, but do not overuse namespaces.
   - Prefer namespaces and inheritance over nested types. Types aren't namespaces.
   - In most cases, if it's in a folder, it should be in a same-named namespace. The need for organization is no different in either place.
 - [Avoid nested types as much as possible.](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types)
   - This includes things like before when talking about Pos/Dim, nested types were instantiated, but returned as their containing types, especially with them not being public, which would also be counter to guidelines.
 - Require use of generic forms of types, interfaces, delegates, etc, unless there's a good reason not to.
 - "CamelCase" is typically called PascalCase or InitCaps. "camelCase" is camelCase.
   - This is just a suggestion to use the dotnet-standard terminology.
 - A couple more casing rules, which are real compatibility issues with other languages
   - Never create namespaces with types whose names are the same as the namespace or differ only in case.
   - Never create members of types whose names are the same as the type or differ only in case.
 - Suggest/prefer `partial` type implementations in files with nestable names over the use of `#region`s for organization
   - This one is aimed at the guideline to prefer modern language constructs over older ones.
   - Also has some advantages when working with the code, such as when you have tons of analyzers and real-time testing going on whose performance all tend to scale inversely proportional to the length of the active document.

dodexahedron:
A couple more that came to mind just now:

 - Strongly prefer c\# 12 collection expressions anywhere they can be used
   - It enables optimizations at build time that are otherwise unavailable to the compiler.
   - It sometimes results in more efficient intermediate operations, when the compiler can safely assume and use a span.
   - It goes hand in hand with the guideline to use modern constructs.
   - However, it is quite often dependent on the suggestion not to use var everywhere, since it's not possible to infer a type without a cast, which may cause extra work.
 - Don't create types with the same name as common types in the base class library, ESPECIALLY those in the root System namespace. `Attribute`, I'm looking at you.
 - Always explicitly define enum values, and, per enum design guidelines, make the 0 value explicitly a `None` or similar member, as it is the default value of all enums. Real values should only be 0 if they are guaranteed to always be safe to be equivalent to zero, which is rare.

dodexahedron:
Also, re-reading the existing rules while more awake than last night, I see that a couple bullet points are already there, so just ignore those or accept a fist bump in preferential solidarity. 🤪

dodexahedron:
Just linking the branches from #3212 here for completeness.

Branch containing only the addition and design of the dotsettings files, one commit per related setting group:

https://github.com/dodexahedron/Terminal.Gui/tree/v2_resharper_style_settings_3212

Branch that has only the necessary changes applied, from that point, to resolve the inspections marked as errors by the new settings, in one commit:

https://github.com/dodexahedron/Terminal.Gui/tree/v2_resharper_style_settings_fixed_build_3212

dodexahedron:
I started a discussion thread to contain miscellaneous ReSharper stuff, to cut down on future noise about it in related issues.

https://github.com/gui-cs/Terminal.Gui/discussions/3213

Pointed out some extensions for it that I find handy.

tig:
Closing in lieu of #3202

PR #3220, opened 2024-01-27
Fixes #3212 (partial). ReSharper settings files defined.

### LINKED ISSUE #3212: Solution-wide formatting pass
This is just to put a formal work item in for #3207.

Also, per the discussion that has happened so far in that PR, I think a dedicated branch and PR are warranted to take care of a dependency: Creating and adding appropriate dotsettings files for general use. Then, the final committed code in #3207 will be able to be the result of using and applying those rules, so that we don't end up with more changes in short order.

I've got partial definitions already, with more complete settings pending resolution of discussion around it all.

dodexahedron:
This PR is for merging the new ReSharper settings to v2_develop - nothing more.

I've been using these for the past couple of days and they seem to do the basic job well.

If anyone has any further thoughts or feedback based on these or the discussion in #3212, any and all thoughts are welcome!

## Proposed Changes/Todos

- [x] Add Solution and Project dotsettings shared files for ReSharper/Rider to the repo

## Pull Request checklist:

Ha. For this one, in particular, most of the typical template above is either the entire point of this PR or is not relevant since this is a non-code PR


Terminal.Gui/Terminal.Gui.csproj.DotSettings
dodexahedron, line 1: This, like the other project-level files, sets the R# language analysis level to C# 12.

This one also tells R# to care about localization for strings, which can assist with various operations that are relevant to localization.

Terminal.sln.DotSettings
dodexahedron, line 1: This one applies to the whole solution, and is where all the good stuff lives.

The only apparent difference it actually bothers to enforce, at this time (shown as warnings, so you still aren't _required_ to make the related changes) is that it allows for the use of `var` to not be mandatory, if the type is not "evident," which, among other things, keeps it from changing certain very intentional usages of explicit types to var that would otherwise result in worse performance.

There are also some custom code cleanup profiles that perform subsets of the various operations, for when you either don't want to make a big change or a file is large and would be overly heavy to perform all operations on at once.

There are also code _layout_ rules, which are applied in any cleanup profile that includes the "reorganize code" operation. That ruleset governs how it will organize the code in a given file. That stuff sure is nice even just for the alphabetical sorting applied within most subgroups that it operates on, so you can quickly navigate through the code, even when not in VS, like on github or in a plain text editor.

dodexahedron (commented):
I put comments on one of the project level files and the solution-level file, just for a mor detailed explanation of what, exactly, they are and what they do.

BDisp:
There is a situation on the R# that I don't like, when I need to change code. e.g  I want to change True from `Assert.True (somebool)` to False. The IntelliSense write as `Assert.False () (somebool)`. I always have to delete the parenthesis to correct it. VS2022 IntelliSense write it correctly at once as `Assert.False (somebool)`

dodexahedron:
That's from having the same completions enabled in both VS and ReSharper. Only turn one on at a time..

In your local resharper options, the relevant settings are in these sections.

You can fine-tune it as much as you like:

image

tig:
I'm travelling this weekend with little time to werk this stuff. 

My priorities are:

1) Help @BDisp get #3181 merged
2) Get #3202 merged
3) Get this merged (it may help get those two merged; not sure yet)

dodexahedron:
> I'm travelling this weekend with little time to werk this stuff.
> 
> My priorities are:
> 
> 1. Help @BDisp get [Fixes #3171. Remove View and subclass constructors with parameters. #3181](https://github.com/gui-cs/Terminal.Gui/pull/3181) merged
> 2. Get [Fixes #3192. Improve correctness / clarity of existing `View.AutoSize` functionality/unit tests #3202](https://github.com/gui-cs/Terminal.Gui/pull/3202) merged
> 3. Get this merged (it may help get those two merged; not sure yet)

Sounds good.

 1. 👍 
 2. 👍 
 3. I'd say merge them on whatever order makes sense to you, because I don't think it's going to make much difference in the end. @BDisp has been getting acquainted with R#, and I think was trying to make use of this in at least one issue we recently spoke on, so they may be completely or partially in this format anyway. It's going to be merge conflict city almost no matter what, for whichever happens last, but as long as we just keep an eye on recently merged bits when the actual work to merge the re-formatted code is done, I'm confident we'll be fine.

This particular pull request, though, is _just_ for the settings. The actual re-formatting is intended to be done separately. That was to enable work to continue in other issues, using formatting rules that will be in compliance from the start and thus be easier to resolve after the big reformat.

dodexahedron:
I think this one is now safe to close as obsolete, in favor of #3202, which is now merged.

Additional settings can be defined and applied later on.

The bulk of the goal was achieved by that merge.

PR #3241, opened 2024-02-11
Fixes #3240: editorconfig and project documentation sync with reformatting

### LINKED ISSUE #3240: Post-Cleanup Cleanup
After @tig just merged the monster that was the #3202 PR, there are just some minor things to shore up.

The .editorconfig file needs to be synced with the current dotsettings configuration, so users not using ReSharper or Rider will get at least most of the same behaviors when they do a visual studio reformat, as well as giving them similar analyzer behavior, at least for the things that VS supports.

There may also be documents that need to be validated for consistency with the current settings as applied to the code, so I'll check those and make any adjustments, as necessary.

This one is a non-code change. Settings and documentation only.

dodexahedron:
This PR is for #3240

This makes two non-code changes:
 1. Updates the .editorconfig file to match settings defined in ReSharper, to the extent possible in VS
 2. Updates the CONTRIBUTING.md doc to reflect revised styling rules and add a little more detail to it as well

Ready for review and merge


dodexahedron:
Man. No idea why git decided to note all those other commits, but they're linear history branched from current v2_develop, with the only modified files being the two mentioned.

dodexahedron:
Interesting.... Note that nothing was touched to cause those tests to fail, yet the last commit has test runner failures.

I bet that's one of the heisenbugs I've mentioned elsewhere. Tests pass locally and on an Ubuntu 23.10 machine I cloned this to.

dodexahedron:
Hm. Lots of tests are failing, after a re-clone. And it's mostly for mismatched strings.

You guys seeing test failures from the current v2_develop?

tig (approved):

PR #3245, opened 2024-02-12
Fixes #3229. TextFormatter should have a FillRemaining property.

### LINKED ISSUE #3229: TextFormatter should have a FillRemaining property.
> See the difference if `TextFormatter.Draw` `fillRemaining` parameter is true. All the `ContentArea` is filled with whitespaces and the same background color showing a perfect rectangle. I think I'll create a `FillRemaining` property to be configurable by the views and pass it's value on `View.Draw` method parameter.

 _Originally posted by @BDisp in [#2489](https://github.com/gui-cs/Terminal.Gui/issues/2489#issuecomment-1925439649)_

See the differences:

Image

Image

I'll fix this in the #3181.

BDisp:
## Fixes

- Fixes #3229

## Proposed Changes/Todos

- [x] Add FillRemaining property
- [x] Allow pass the property as parameter to the TextFormatter?.Draw
- [x] Add unit test

## Pull Request checklist:


Terminal.Gui/Text/TextFormatter.cs
```diff
@@ -170,6 +170,13 @@ public TextDirection Direction
         }
     }
 
+    /// <summary>
+    ///     Determines if the bounds width will be used or only the text width will be used,
+    ///     If <see langword="true"/> all the bounds area will be filled with whitespaces and the same background color
+    ///     showing a perfect rectangle.
+    /// </summary>
+    public bool FillRemaining { get; set; }
```
tig, line 178: I'm not convinced this property is needed. 

Why not just have code that needs the fillRemaining parameter to be true pass true?

Terminal.Gui/View/Adornment/Adornment.cs
```diff
@@ -141,7 +141,7 @@ public override void OnDrawContent (Rect contentArea)
             }
         }
 
-        TextFormatter?.Draw (screenBounds, normalAttr, normalAttr, Rect.Empty, false);
+        TextFormatter?.Draw (screenBounds, normalAttr, normalAttr, Rect.Empty, TextFormatter.FillRemaining);
```
tig, line 144: Why not just change the `false` to `true`?

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -390,7 +390,7 @@ public virtual void OnDrawContent (Rect contentArea)
                                  HasFocus ? GetFocusColor () : GetNormalColor (),
                                  HasFocus ? ColorScheme.HotFocus : GetHotNormalColor (),
                                  Rect.Empty,
-                                 false
+                                 TextFormatter.FillRemaining
```
tig, line 393: Why not just change the `false` to `true`?

tig (changes requested):
Unless i'm missing something, I think this PR can be simplied per my comments.

Terminal.Gui/View/Adornment/Adornment.cs
BDisp: Because that way we are forcing to used it as true. By te way this parameter is true by default and in this case we don't need to pass it into the parameter. With a property we let the user the choice to use his value to pass into the parameter.

tig:
`TextFormatter.Draw` is an instance method. It is silly to have BOTH a `fillRemaining` parameter AND an instance property that do the same thing.

I suggest:

- Add `FillRemaining` as a property
- Remove the `fillRemaining` parameter from `Draw`
- Have `Draw` reference `FillRemaining` where it currently references `fillRemaining`

Make sense?

BDisp:
> `TextFormatter.Draw` is an instance method. It is silly to have BOTH a `fillRemaining` parameter AND an instance property that do the same thing.
> 
> I suggest:
> 
> * Add `FillRemaining` as a property
> * Remove the `fillRemaining` parameter from `Draw`
> * Have `Draw` reference `FillRemaining` where it currently references `fillRemaining`
> 
> Make sense?

Yes it makes sense, but you didn't answer about the default property value. Do you agree with this?

`public bool FillRemaining { get; set; } = true;`

tig:
> > `TextFormatter.Draw` is an instance method. It is silly to have BOTH a `fillRemaining` parameter AND an instance property that do the same thing.
> > I suggest:
> > 
> > * Add `FillRemaining` as a property
> > * Remove the `fillRemaining` parameter from `Draw`
> > * Have `Draw` reference `FillRemaining` where it currently references `fillRemaining`
> > 
> > Make sense?
> 
> Yes it makes sense, but you didn't answer about the default property value. Do you agree with this?
> 
> `public bool FillRemaining { get; set; } = true;`

I'm not sure. The call in `View.OnDrawContent`, `Thickness.Draw`, etc... specify false. That's because we need to minimize how much default drawing we're doing for perf reasons. If the background was already drawn, there's no need to make it `true`.

So, I guess I'd pick `false` as the default`.

tig (approved):

PR #3248, opened 2024-02-14
Fixes #3242 - Replaces simple null checks

### LINKED ISSUE #3242: Fix remaining null checking code
Since R# wasn't able to rewrite 100% of the comparisons to null using `operator==`, a manual pass needs to be done to finish fixing those up.

This should be a pretty simple work item.

Tracked in https://github.com/dodexahedron/Terminal.Gui/issues/22

dodexahedron:
## Fixes

- `==` and `!=` should not be used to compare against null
- Tests do not compile and cannot run under Release configuration

## Proposed Changes/Todos

- [x] Replace null comparisons from using equality operators to using the `is` operator to avoid potential issues with overloaded equality operators.
- [x] Conditionally compile test code to match with output in release or debug configurations.
  - This actually looks like possible bad behavior to me. See TODO and BUG comments in change diffs.

## Pull Request checklist:

  - There are no formatting changes in this one at all. It's just the mentioned replacements.


Terminal.Gui/Drawing/ColorScheme.cs
```diff
@@ -215,7 +215,7 @@ private class SchemeNameComparerIgnoreCase : IEqualityComparer<string>
     {
         public bool Equals (string x, string y)
         {
-            if (x != null && y != null)
+            if (x is { } && y is { })
```
tig, line 218: Can we get rid of the two warnings in this file while here?

UnitTests/View/ViewTests.cs
```diff
@@ -1166,15 +1166,23 @@ public void New_Initializes ()
         // BUGBUG: IsInitialized must be true to process calculation
         r.BeginInit ();
         r.EndInit ();
+    #if DEBUG
```
tig, line 1169: I think it would be better to fix `View.ToString` to be consistent between release and debug. Both `View.Text` and `View.TItle` set `View.Id` in DEBUG. Instead, `ToString` should be smarter than this:

```cs
  public override string ToString () { return $"{GetType ().Name}({Id}){Frame}"; }
```
Instead:

```cs
    {
        string id = Id;

        if (string.IsNullOrEmpty (id))
        {
            if (!string.IsNullOrEmpty (Title))
            {
                id = Title;
            } else if (!string.IsNullOrEmpty (Text))
            {
                id = Text;
            }
        }
        return $"{GetType ().Name}({id}){Frame}";
    }
```

(Or similar).

tig (changes requested):
Thank you for this. No fun, but worthy!

See my minor requsets.

Terminal.Gui/Drawing/ColorScheme.cs
dodexahedron: Sure thing. I have a few additional minor code quality changes as well, taking care of other analyzer suggestions.

UnitTests/View/ViewTests.cs
dodexahedron: This gives me an idea.

What if there's a static counter int and, unless an explicit ID is set, it simply falls back to type name plus the next value of that integer?

Simple, clean, and at least similar to how VS would do it in the windows forms designer, with the main difference being that int is shared instead of per leaf type.

Increment is done via Interlocked.Increment for thread safety.

dodexahedron: And, if per-type counters are ever wanted on a type, for cosmetic purposes or whatever, it's as simple as hiding the counter field in that derived class.

tig: I'd support that.

Terminal.Gui/Drawing/ColorScheme.cs
dodexahedron: That class was removed. It was unused, has a built-in equivalent, and is private, so was just unneeded code.

UnitTests/View/ViewTests.cs
dodexahedron: Coolio. Work for another issue, of course.

dodexahedron: Opened #3250 for that. Labeled as low priority. Might go ahead and handle it when doing event work that heavily touches the View class.

tig:
Any idea why the dotnet sdk is not installing in the GH action?

dodexahedron:
Actually hang on a bit.

I have another quality improvement for the Colors class that could be pretty handy for upcoming work.

It's just a wrapper for a Dictionary, so implementing a couple of interfaces is simple and makes use of the class more natural.

dodexahedron:
There we go.

Quick and easy.

Just delegates the interface implementations to the wrapped collection and invokes the CollectionChanged event at the appropriate times.

dodexahedron:
> Any idea why the dotnet sdk is not installing in the GH action?

Hm?

dodexahedron:
Ohhhh is that why you've got all those "trying to fix ci issue" commits on that branch you're working on?

tig:
Seems to have fixed itself. Maybe a CDN issue. I just re-ran the jobs on this PR...

dodexahedron:
Here's an example of a build workflow I have for the SnapsInAZfs project, which has a solution with libraries, an application, and unit test projects for each. It execcutes when a tag is pushed with a name beginning with `build-`.

```yaml
name: "Latest 'build' Tag Status"

on:
  push:
    tags: [ "build-*" ]

jobs:
  build-and-test:
    strategy:
      matrix:
        os: [ubuntu-latest]
    runs-on: ${{ matrix.os }}
    steps:
    - uses: actions/checkout@v3
    - name: Setup .NET
      uses: actions/setup-dotnet@v3
      with:
        dotnet-version: 8.0.x
    - name: Restore dependencies
      run: dotnet restore
    - name: Build
      run: dotnet build --no-restore --configuration Release-R2R
    - name: Test
      run: dotnet test --no-restore --configuration Release-R2R --verbosity minimal
```

dodexahedron:
That test shouldn't be failing either.

I'll do a fresh clone and see what happens.

We're hitting the concurrency problems more frequently, now, and that'll get even worse as we continue to improve performance.

dodexahedron:
Huh.

Ok.

Multiple are failing and I'm pretty sure I see why.

The set accessor on certain properties was automatically removed because it thinks it is unneeded.

I'll annotate those to keep that from happening.

dodexahedron:
Just one property.

Restored the setter and marked so ReSharper won't do that again.

dodexahedron:
Aaannnnd there's a commit to make the CI build happy.

dodexahedron:
Yay

tig (approved):
Lovely work. Thank you.

PR #3249, opened 2024-02-15
Fixes #3039. Fix View.HotKey

### LINKED ISSUE #3039: Fix `View.HotKey` - Unshifted is broken & should be based on `Title`, not `Text`
There are several issues with HotKey support in v2_develop right now:

- [x] When I turned off automatically setting HotKey for first uppercase letter in Text I didn't fix all Scenarios.
- [x] Bare hotkeys (no Alt) don't work. If a subview has a hotkey pof `n` and the user presses `n` the focus should change to that view *without* doing an `Accept`. IOW `Command.Default` should be invoked.
  - [ ] Related, `View` should have an event for the default command (`OnDefaultCommand` fires `DefaultCommand`).
- [x] Alt-Hotkey should invoke `Command.Accept` by default. 
  - [ ] Related, `View` should have an event for the accept command (`OnAccept` fires `Accept`). `OnAccept` should not set focus (`Command.Default` does).
  - [ ] `OnMouseClick` should use `Command.Accept` instead of what it does now.
- [ ] There's no Scenario that really shows off/exercises HotKey support. `Buttons` tries, but meh.
- [x] `View.HotKey` uses `View.Text` (actually `View.TextFormater`) to set the hot key automatically. I don't think this is right. Instead, HotKey should be based off of `Title`.

Views like `Label` and `Button` should *also* use `Text` (they should override `OnTitleChanged` and set `Text` to match and viceversa).

Why?

```cs
fv = new FrameView() { Title = "_FrameView" };
```

This should show a FrameView with the title of _Title. If the frame view is in a dialog and the user presses Alt-F, the Frameview should get focus.

tig:
## Fixes

- Fixes #3039 

## Proposed Changes/Todos

- [x] Bare hotkeys (no Alt) don't work. If a subview has a hotkey of `n` and the user presses `n` the focus should change to that view *without* doing an `Accept`. 
  - [x] `Command.Default` should be invoked.
  - [x] `View` should have an event for the default command (`OnDefaultCommand` fires `DefaultCommand`).
- [x] Alt-Hotkey should invoke `Command.Accept` by default. 
  - [x] `View` should have an event for the accept command (`OnAccept` fires `Accept`). `OnAccept` should not set focus (`Command.Default` does).
  - [x] Unit tests for `OnAccept`/`Accept`
  - [x] `OnMouseClick` should use `Command.Accept` instead of what it does now.
- [x] `View.HotKey` uses `View.Text` (actually `View.TextFormater`) to set the hotkey automatically. HotKey should be based on `Title`.
  - [x] `Label`, `Button`, and other views where having `Text` and `Title` be distinct makes no sense: 
    - [x] `Label` Text mirrors Title 
    - [x] `Button` Text mirrors Title 
    - [x] `Checkbox` Text mirrors Title 
    - [x] Enable a more consistent way where `Text` and `Title` stay in sync (`Title` should be the source of truth)
- [x] Enable "Set focus to paired view" - E.g. `_Label` to the left of a `TextView`, user presses `Alt-L` or `L`; focus should move to the `TextView`. This works today, but only because of hack in `Label`.
- [x] Update Scenarios to explicitly provide hotkey specifiers (where appropriate)
- [x] Create a `HotKeys` Scenario that shows off/exercises HotKey support. 
- [x] `MenuBar` interferes with subview hot keys - In UICatalog try Alt-S or Alt-C...
- [x] `ListView.FocusNext` doesn't always work - something is re-ordering `SubViews`; this can be seen in the Reactive Sample where `Alt-U` does not work.

## Pull Request checklist:


Terminal.Gui/Terminal.Gui.csproj
```diff
@@ -39,8 +39,8 @@
   <!-- =================================================================== -->
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="all" />
```
dodexahedron, line 41: We can and probably should drop this reference, since V2 isn't .NET Framework-compatible.

The reference isn't used, anyway.

dodexahedron (changes requested):
Just a thought while you're touching the csproj.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -1606,7 +1606,8 @@ private KeyCode MapKey (WindowsConsole.ConsoleKeyInfoEx keyInfoEx)
                 }
             }
 
-            return (KeyCode)keyInfo.KeyChar;
+            // Return the Key (not KeyChar!)
```
dodexahedron, line 1609: Note (not about this change specifically but the code above it).

Lines 1585 through 1594 are redundant with the block immediately following on line 1595 and can just be removed, unless this is supposed to be doing something else.

tig:
@dodexahedron, @BDisp, @tznind -- Questions on Command / Event design for Views.

 Right now, we define `Command.Default` as "The default command. For View, this focuses the view."

This made sense to me when I added it. After working through this PR, I now believe I was not thinking clearly. Specifically, it's not that a view should have a super-generic "default" command, but what view needs is a command that gets invoked when a user invokes the `HotKey`. 

- `View` - Causes the view to gain focus.
- `Label` - Causes the next view in the tab-order to gain focus.
- `Button` - Causes the view to gain focus and calls `OnClicked` which fires `Clicked`
- `Checkbox` - Causes the view to gain focus calls `OnToggle` which fires `Toggled` (this is the current behavior, but I challenge it; I would argue this behavior is too aggressive and toggling a check should require two user actions: focus, then select, but that's a debate for another day).
- Most/all-other views inherit `View` behavior of just gaining focus.

Based in this, I want to rename `Command.Default` to `Command.HotKey`. Thoughts?

Related, today we define these `Command` enum values:

* `Command.Accept`: Accepts the current state (e.g. selection, button press etc).
* `Command.Select`: Moves selection to an item (e.g. highlighting a different menu item) without necessarily accepting it.
* `Command.ToggleChecked`: Toggle the checked state.

I think `Command.Accept` and `Command.Select` are fine as-is. I think `ToggleChecked` is superfluous and can be deleted; use `Command.Accept` instead. `Checkbox` would be the only built-in view impacted. 

I also think `Button` should be changed such that `OnClicked` and `Clicked` are replaced with `OnAccept`/`Accept`. Why? Because clicking on a button is the same as invoking the hotkey (or the keys bound to `Accept` (space/enter)). 

Related to the above, I propose we combine a bunch of existing `EventArgs` classes into one. 

```cs
public class StateEventArgs<T> : CancelEventArgs
{
    /// <summary>Creates a new instance of the <see cref="StateEventArgs"/> class.</summary>
    /// <param name="oldValue"></param>
    /// <param name="newValue"></param>
    public StateEventArgs (T? oldValue, T? newValue)
    {
        OldValue = oldValue;
        NewValue = newValue;
    }

    /// <summary>The new state</summary>
    public T? NewValue { get; }

    /// <summary>The previous state</summary>
    public T? OldValue { get; }
}
```

Specifically, all events that convey the state of the view has changed. For `CheckBox` `T` would be `bool?`.  For `TextView.TextChanged` `T` would be `string`, etc...

Any objections to this? Am I missing something? 

I need more consistency here for both THIS PR as well as 

- #3076

BDisp:
I think is reasonable the changes. For `CheckBox` I only want to call the attention that navigating through theirs items aren't actually changing the current selected item and only when pressing the space bar is really accept it. You are really doing a great work.

tig:
> For `CheckBox` I only want to call the attention that navigating through theirs items aren't actually changing the current selected item and only when pressing the space bar is really accept it. 

Right. This is why I think using the keyboard to focus on a checkbox (via `HotKey` or `Tab`) should NOT change the check box state (as `HotKey` does now).

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
tig: Nope.  1589 checks for `&&` and 1595 checks for `||`.

tig:
This is ready for review.

tig:
@dodexahedron, @BDisp, and/or @tznind:

Could I get your eyes on this please? I'd like to get it merged.

Terminal.Gui/Input/Key.cs
```diff
@@ -1,5 +1,7 @@
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using static System.Runtime.CompilerServices.RuntimeHelpers;
```
BDisp, line 4: This namespace is really needed?

BDisp (approved):
Great work. Very tedious indeed. I'll change `KeyCode` to `Key` in my PR too.

Terminal.Gui/Input/Key.cs
dodexahedron: I haven't looked at this yet but, following onto @BDisp's question, I'd add this suggestion:

`using static` should be avoided most of the time, because it makes it unclear that method calls dependent on it actually came from the static class referenced in the `using static`. It also makes naming collisions more perilous.

About the only place I've ever really thought it was a good idea is when dealing with expression trees, since everything is a static method in there and the code is half as long with the using than without it.

tig: Fixed.

PR #3299, opened 2024-03-05
Fixes #3298. `BoundsToScreen`

tig:
## Fixes

- Fixes #3298

## Proposed Changes/Todos

- [x] Remove unused `clamped` param
- [x] Remove all references to `void BoundsToScreen...`
- [x] Remove `void BoundsToScreen` so that only `public virtual Rectangle BoundsToScreen (Rectangle bounds)` remains
- [x] Add sorely needed real unit tests

## Pull Request checklist:

- [x] I've named my PR in the form of "Fixes #issue. Terse description."
- [x] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [x] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [x] I ran `dotnet test` before commit
- [x] I have made corresponding changes to the API documentation (using `///` style comments)
- [x] My changes generate no new warnings
- [x] I have checked my code and corrected any poor grammar or misspellings
- [x] I conducted basic QA to assure all features are working

PR #3300, opened 2024-03-06
Fixes #3298 (redo).  `BoundsToScreen` to `FrameToScreen` improvements

### LINKED ISSUE #3298: `BoundsToScreen` is confusing
## Background

We have two APIs:

```cs
    /// <summary>Converts a <see cref="Bounds"/>-relative region to a screen-relative region.</summary>
    public Rectangle BoundsToScreen (Rectangle region)
    {
        BoundsToScreen (region.X, region.Y, out int screenX, out int screenY, false);

        return region with { X = screenX, Y = screenY };
    }

    /// <summary>
    ///     Converts a <see cref="Bounds"/>-relative coordinate to a screen-relative coordinate. The output is optionally
    ///     clamped to the screen dimensions.
    /// </summary>
    /// <param name="x"><see cref="Bounds"/>-relative column.</param>
    /// <param name="y"><see cref="Bounds"/>-relative row.</param>
    /// <param name="rx">Absolute column; screen-relative.</param>
    /// <param name="ry">Absolute row; screen-relative.</param>
    /// <param name="clamped">
    ///     If <see langword="true"/>, <paramref name="rx"/> and <paramref name="ry"/> will be clamped to the
    ///     screen dimensions (will never be negative and will always be less than <see cref="ConsoleDriver.Cols"/> and
    ///     <see cref="ConsoleDriver.Rows"/>, respectively.
    /// </param>
    public virtual void BoundsToScreen (int x, int y, out int rx, out int ry, bool clamped = true)
    {
        // PERF: Use Point.Offset
        // Already dealing with Point here.
        Point boundsOffset = GetBoundsOffset ();
        rx = x + Frame.X + boundsOffset.X;
        ry = y + Frame.Y + boundsOffset.Y;

        View super = SuperView;

        while (super is { })
        {
            if (super is Adornment)
            {
                // TODO: Move this into Adornment somehow to remove coupling.
                super.BoundsToScreen (rx, ry, out rx, out ry);
            }

            boundsOffset = super.GetBoundsOffset ();
            rx += super.Frame.X + boundsOffset.X;
            ry += super.Frame.Y + boundsOffset.Y;

            super = super.SuperView;
        }

        // The following ensures that the cursor is always in the screen boundaries.
        if (clamped)
        {
            ry = Math.Min (ry, Driver.Rows - 1);
            rx = Math.Min (rx, Driver.Cols - 1);
        }
    }
```

Having two is confusing, esp since one is virtual and overridden in Adornment. 

This Issue is to simplfy. I want to do this independently of 

- https://github.com/gui-cs/Terminal.Gui/pull/3295

to ensure I'm not moving multiple things at once.

tig:
Redo of #3299 - which I screwed up.

## Fixes

- Fixes #3298 

## Proposed Changes/Todos

- [x] Remove unused `clamped` param
- [x] Remove all references to `void BoundsToScreen...`
- [x] Remove `void BoundsToScreen` so that only `public virtual Rectangle BoundsToScreen (Rectangle bounds)` remains
- [x] Add sorely needed real unit tests

## Pull Request checklist:


Terminal.Gui/Application.cs
```diff
@@ -807,7 +807,7 @@ public static void RunIteration (ref RunState state, ref bool firstIteration)
                 || state.Toplevel.SubViewNeedsDisplay
                 || state.Toplevel.LayoutNeeded))
         {
-            state.Toplevel.Clear (Driver.Bounds);
+            Driver.ClearContents ();
```
dodexahedron, line 810: Yay for getting rid of that strange inversion. 😃

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -176,26 +176,31 @@ public virtual Rectangle Bounds
...
-
-            int height = Math.Max (
-                                   0,
-                                   Frame.Size.Height - Margin.Thickness.Vertical - Border.Thickness.Vertical - Padding.Thickness.Vertical
-                                  );
+            //if (!IsInitialized)
+            //{
+                int width = Math.Max (
+                                      0,
+                                      Frame.Size.Width
```
dodexahedron: Maybe a good place to use the Offset methods?

One important thing about properties:

A call to set followed immediately by a get should return the original value, in most cases.

When those semantics aren't feasible, the recommendation is typically to use an explicit set method instead of allowing use of the set accessor, or, if still desired, make the set accessor throw an exception instead of messing with input silently.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -499,21 +504,15 @@ public Pos Y
     public event EventHandler Initialized;
 
     /// <summary>Converts a <see cref="Bounds"/>-relative rectangle to a screen-relative rectangle.</summary>
-    public virtual Rectangle BoundsToScreen (Rectangle bounds)
+    public Rectangle BoundsToScreen (Rectangle bounds)
```
dodexahedron, line 487: use `in` for the parameter.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -499,21 +504,15 @@ public Pos Y
...
     {
+        // Translate bounds to Frame (our SuperView's Bounds-relative coordinates)
         Point boundsOffset = GetBoundsOffset ();
-        bounds.Offset(Frame.X + boundsOffset.X, Frame.Y + boundsOffset.Y);
 
-        View super = SuperView;
+        Rectangle screen = FrameToScreen ();
+        screen.Offset (boundsOffset.X + bounds.X, boundsOffset.Y + bounds.Y);
```
dodexahedron, line 493: Use of Point.Add would make this more efficient.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -499,21 +504,15 @@ public Pos Y
...
 
-        while (super is { })
-        {
-            boundsOffset = super.GetBoundsOffset ();
-            bounds.Offset(super.Frame.X + boundsOffset.X, super.Frame.Y + boundsOffset.Y);
-            super = super.SuperView;
-        }
-
-        return bounds;
+        return new Rectangle(screen.Location, bounds.Size);
```
dodexahedron: Use of `with` here enables better compiler optimization.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -616,12 +615,14 @@ public Thickness GetAdornmentsThickness ()
     ///     Helper to get the X and Y offset of the Bounds from the Frame. This is the sum of the Left and Top properties
     ///     of <see cref="Margin"/>, <see cref="Border"/> and <see cref="Padding"/>.
     /// </summary>
-    public Point GetBoundsOffset ()
+    public virtual Point GetBoundsOffset ()
```
dodexahedron, line 593: This is potentially dangerous to be virtual.

But storing the rectangles is a better solution anyway and makes it less of an issue (maybe none at all).

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -80,33 +80,46 @@ public void AddRune (int col, int row, Rune ch)
         Driver.AddRune (ch);
     }
 
-    /// <summary>Clears the <see cref="Bounds"/> with the normal background color.</summary>
+    /// <summary>Clears the <see cref="Frame"/> with the normal background color.</summary>
     /// <remarks>
     ///     <para>This clears the Bounds used by this view.</para>
     /// </remarks>
-    public void Clear ()
+    public void ClearFrame ()
```
dodexahedron: Rename good. 👍 

But I think the XmlDoc should be a bit more explicit about it being ONLY about drawing this View and that it doesn't have anything to do with any subviews (ie it's not clearing any collections).

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -490,11 +490,11 @@ private void HideList ()
             OnOpenSelectedItem ();
         }
 
-        Rectangle rect = _listview.BoundsToScreen (_listview.IsInitialized ? _listview.Bounds : Rectangle.Empty);
         Reset (true);
-        _listview.Clear (rect);
+        _listview.Clear (_listview.IsInitialized ? _listview.Bounds : Rectangle.Empty);
         _listview.TabStop = false;
         SuperView?.SendSubviewToBack (this);
+        Rectangle rect = _listview.BoundsToScreen (_listview.IsInitialized ? _listview.Bounds : Rectangle.Empty);
```
dodexahedron, line 497: Hmmmm.

This is making me wonder if BoundsToScreen is perhaps a better place for that conditional.

Calls to it would seem to always need that check, no? Plus, it would clean these up and ensure consumers don't make the mistake of leaving it out.

UnitTests/View/Adornment/AdornmentTests.cs
```diff
@@ -15,35 +15,80 @@ public void Bounds_Location_Always_Empty_Size_Correct ()
...
+        Assert.Equal (new (0, 0, view.Border.Frame.Width, view.Border.Frame.Height), view.Border.Bounds);
 
-        Assert.Equal (new Rectangle (0, 0, view.Margin.Frame.Width - marginThickness * 2, view.Margin.Frame.Height - marginThickness * 2), view.Margin.Bounds);
+        Assert.Equal (new (0, 0, view.Padding.Frame.Width , view.Padding.Frame.Height), view.Padding.Bounds);
+    }
 
-        Assert.Equal (new Rectangle (0, 0, view.Border.Frame.Width - borderThickness * 2, view.Border.Frame.Height - borderThickness * 2), view.Border.Bounds);
+    // Test that Adornment.Bounds_get override returns Frame.Size minus Thickness
+    [Theory]
+    [InlineData (0, 0, 0, 0, 0)]
```
dodexahedron, line 48: Excellent place for use of the generators in the combinatorial extension or a generator method if that's too annoying to adapt.

Generator method allows shorter test code, too, since you can assert on the rectangles and points, rather than each individual property.

dodexahedron: This particular one is already a pretty well-written test, so the CombinatorialData thing is a nice fit here.

dodexahedron (commented):
A few general comments and a couple of suggestions before it's finished.

Looking good, in general, though.

Terminal.Gui/View/Adornment/Adornment.cs
```diff
@@ -30,10 +30,22 @@ public Adornment ()
...
+    //}
+
+    /// <summary>
+    /// Gets the rectangle that describes the area of the Adornment. The Location is always (0,0).
+    /// The size is the size of the Frame 
+    /// </summary>
     public override Rectangle Bounds
     {
-        get => new Rectangle (Point.Empty, Thickness?.GetInside (new (Point.Empty, Frame.Size)).Size ?? Frame.Size);
+        get => new (Point.Empty, Frame.Size);
```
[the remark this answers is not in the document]
tig: Changed to with.

On the perf stuff, sure... but not today.

Terminal.Gui/View/Layout/ViewLayout.cs
tig: Help my feeble brain work the Math.Max into a use of Offset?

tig: This is not offsettting a Rectangle, but changing it's Size.

Terminal.Gui/View/Layout/ViewLayout.cs
tig: Help me out here. What would that look like?

Terminal.Gui/View/ViewDrawing.cs
tig:
```cs
    /// <summary>Clears the entire view (<see cref="Frame"/>) with the normal background color.</summary>
    /// <remarks>
    ///     <para>This method will clear the area occupied by the <see cref="Margin"/>, <see cref="Border"/>, and <see cref="Padding"/>, as
    ///     well as the <see cref="Bounds"/>. Use <see cref="Clear()"/> to clear just the <see cref="Bounds"/>.</para>
    /// </remarks>
```

Terminal.Gui/Views/ComboBox.cs
tig: This is old code. I think the check can be removed.

UnitTests/View/Adornment/AdornmentTests.cs
tig: I just haven't had the time to get my head around those things yet.

Terminal.Gui/View/Layout/ViewLayout.cs
tig: It's better than the 3 virtuals that existed before this PR. 

I'm not sure what you mean by "storing rectangles".

Terminal.Gui/View/Layout/ViewLayout.cs
dodexahedron: Naive case is taking your boundsOffset and bounds variables and doing something like:

screen.Offset(boundsOffset.Add(bounds));

And then whatever other cleanup, simplification, or optimization you might see from there.

Basically, avoiding explicit math on the individual properties/fields of Point, Size, and Rectangle is a good thing, whenever possible.

Terminal.Gui/View/Layout/ViewLayout.cs
dodexahedron: The value returned by this method is something that only changes if a value it is dependent on changes. Re-calculating it is wasteful, but not the end of the world, since we're on the stack during most of those operations, though it _is_ a hot code path, so impact is potentially worth it.

UnitTests/View/Adornment/AdornmentTests.cs
dodexahedron: For a lot of our existing tests, it would require some refactoring of the tests themselves.

This one, however, does not take specific "expected" values per case, instead using the inputs to create the expected values (which is the ideal way).

To use that stuff, the method itself needs the `[CombinatorialData]` attribute, and then each parameter gets the appropriate attribute to supply it either with explicit values, a range, or a generator explicitly for just that parameter.[^1]

It then creates the full set of possible cases represented by the cross-join of all of them.

I've done a few in some of the stuff I did for Color, if you want some actual examples.

Just remember it's combinatorial by default. So a method with 5 parameters and 2 values for each means 2^5 = 32 cases. That same test method with ranges covering 2, 3, 4, 5, and 6 values for each, respectively, means 2 * 3 * 4 * 5 * 6 = 720 cases. Or ranges from -10 to 10 for each means 21^5 = 4,084,101 cases... You get the idea.... And Visual studio does NOT behave well when there are that many test cases, and it is also pretty dumb about how it caches all of that, too. 

[^1]: NUnit has all this natively and it looks like the creator of this extension did try to at least mostly emulate the syntax and behavior, so that's cool.

PR #3309, opened 2024-03-08
Fixes #3308. Tweaks R# settings to not re-order on file layout

### LINKED ISSUE #3306: Add `ViewDiagnosticFlags.MouseCursor`
When debugging mouse stuff, I often wish I could see where Terminal.Gui THINKS the mouse is. We may not need this long term, but it'd be useful now as we're debugging `Adornments` and the trickyness around coordinates.

The idea would be, if enabled, the foreground/background color of the `Cell` the mouse is over would invert.

I think this goes in `Application.MouseEvent` (and not `ConsoleDriver` or `View`).

### LINKED ISSUE #3308: Code Cleanup with R# - File Layout is annoying
Some of the current settings are annoying and should be changed:

## R# claims to support placing backing fields above their properties in FIle Layout. 

But this does not work. 

https://youtrack.jetbrains.com/issue/RSRP-484909/Support-Place-backing-field-above-property-in-file-layout

I want this, and it's super annoying that the global reformat we did screwed this up. To this end, I want to change the default Ctrl-E/C profile to NOT apply file layout.

It can be applied separrately with a "File Layout" profile.

## Sorting is annoying

I configured properties, methods, etc... to sort based on name. This is annoying when it actually makes having related things near each other. For example in ViewLayout.cs, the logical order of X, Y, Width, Height is exactly that. But because we have sorting on, they get placed as Height, Width, X, Y.

tig:
## Fixes

- Fixes #3308

### DIFF
--- a/Terminal.Gui/View/Layout/ViewLayout.cs
+++ b/Terminal.Gui/View/Layout/ViewLayout.cs
@@ -1,3 +1,4 @@
+using System.ComponentModel;
 using System.Diagnostics;
 
 namespace Terminal.Gui;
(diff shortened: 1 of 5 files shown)


dodexahedron (approved):
Looks good to me. Nothing terribly weird, AFAICT. 🤷‍♂️

tig:
When I get a chance, I'm gonna remove the impacted files from this PR and just change the settings.

PR #3320, opened 2024-03-13
Fixes #3318. v1-MouseClick should be invoked only once on a mouse click.

### LINKED ISSUE #3318: v1-MouseClick should be invoked only once on a mouse click.
This unit test fails because it returns 3 clicks (1 for button pressed, 1 for button released and 1 for button click). As the event suggest this event is for to be used on a click event and not for any mouse event.

```cs
		[Fact]
		public void OnMouseClick_Is_Only_Fired_Once ()
		{
			var mouseClicks = 0;
			var view = new View ();
			view.MouseClick += (_) => mouseClicks++;

			view.OnMouseEvent (new MouseEvent () { Flags = MouseFlags.Button1Pressed });
			view.OnMouseEvent (new MouseEvent () { Flags = MouseFlags.Button1Released });
			view.OnMouseEvent (new MouseEvent () { Flags = MouseFlags.Button1Clicked });

			Assert.Equal (1, mouseClicks);
		}

```

BDisp:
## Fixes

- Fixes #3318

## Proposed Changes/Todos

- [x] Only consider clicked event of any button
- [x] Added unit test for proof

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -2894,9 +2894,14 @@ namespace Terminal.Gui {
 				return false;
 			}
 
-			var args = new MouseEventArgs (mouseEvent);
-			if (OnMouseClick (args))
-				return true;
+			if (mouseEvent.Flags.HasFlag (MouseFlags.Button1Clicked) || mouseEvent.Flags.HasFlag (MouseFlags.Button2Clicked)
+				|| mouseEvent.Flags.HasFlag (MouseFlags.Button3Clicked) || mouseEvent.Flags.HasFlag (MouseFlags.Button4Clicked)) {
+
+				var args = new MouseEventArgs (mouseEvent);
+				if (OnMouseClick (args)) {
+					return true;
+				}
+			}
 			if (MouseEvent (mouseEvent))
 				return true;
 
--- a/Terminal.Gui/Core/View.cs
+++ b/Terminal.Gui/Core/View.cs
@@ -2894,8 +2894,8 @@ namespace Terminal.Gui {
 				return false;
 			}
 
-			if (mouseEvent.Flags.HasFlag (MouseFlags.Button1Clicked) || mouseEvent.Flags.HasFlag (MouseFlags.Button2Clicked)
-				|| mouseEvent.Flags.HasFlag (MouseFlags.Button3Clicked) || mouseEvent.Flags.HasFlag (MouseFlags.Button4Clicked)) {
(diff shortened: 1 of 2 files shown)


Terminal.Gui/Core/View.cs
```diff
@@ -2894,9 +2894,14 @@ public virtual bool OnMouseEvent (MouseEvent mouseEvent)
 				return false;
 			}
 
-			var args = new MouseEventArgs (mouseEvent);
-			if (OnMouseClick (args))
-				return true;
+			if (mouseEvent.Flags.HasFlag (MouseFlags.Button1Clicked) || mouseEvent.Flags.HasFlag (MouseFlags.Button2Clicked)
```
dodexahedron, line 2897: Much more expensive than necessary, even in Release build configuration.

Aside from performance, it is not atomic to make multiple calls on it unless you make a local copy of it first and, therefore, unless the first one returns true (thus skipping evaluation of the rest since it's `||`), it exposes a race condition.
 
Especially when testing more than one flag, the bitwise operations should be preferred, and should be combined to one check against a constant value of the same underlying type as the enum, if possible. The compiler actually doesn't treat any of those values as constants, here (really seems like it should, doesn't it???), and does a lot of work that causes a lot of casting and copying, which is from the mouseEvent object on the heap, for each call to HasFlag.

Declaring a const int or whatever backs that enum as your mask value (which can still be written with the friendly names) and then making a single `(mouseEvent.Flags & thatMask) != 0` check saves a surprising amount of work and eliminates the race condition on this line.

Alternatively, use of a switch statement for cases like this one is also pretty good and, so long as there's not a guard clause on any case explicitly using the mouseEvent.Flags value, avoids the race condition, as well.

UnitTests/Views/ViewTests.cs
```diff
@@ -4581,5 +4581,19 @@ public void View_Instance_Use_Attribute_Normal_On_Draw ()
 011110
 000000", new Attribute [] { Colors.TopLevel.Normal, Colors.TopLevel.Focus });
 		}
+
+		[Fact]
+		public void OnMouseClick_Is_Only_Fired_Once ()
```
dodexahedron, line 4586: `public void OnMouseClick_Is_Only_Raised_Once ()`

dodexahedron (changes requested):
Thread safety and performance comment for the new code and a terminology nitpick for the test method.

tig (approved):

tig (approved):

PR #3325, opened 2024-03-14
Fixes #3324. Width and Height are throwing before IsInitialized is true.

### LINKED ISSUE #3324: Width and Height are throwing before IsInitialized is true.
When setting the Dim `Width` and `Height` before set `AutoSize` to false throws `System.InvalidOperationException: 'Must set AutoSize to false before setting Width.'
`
Running this unit test will fail.

```cs
    [Fact]
    public void Setting_Width_Height_Before_Set_AutoSize_To_False_Do_Not_Throws ()
    {
        var exception = Record.Exception (() => new Label { Width = 10, AutoSize = false });
        Assert.Null(exception);

        exception = Record.Exception (() => new Label { Height = 10, AutoSize = false });
        Assert.Null(exception);
    }
```

BDisp:
## Fixes

- Fixes #3324

## Proposed Changes/Todos

- [x] Check for IsInitialized before throwing.
- [x] Add unit test to proof

## Pull Request checklist:

### DIFF
--- a/UnitTests/Views/LabelTests.cs
+++ b/UnitTests/Views/LabelTests.cs
@@ -567,12 +567,14 @@ e
     }
 
     [Fact]
-    public void Setting_Width_Height_Before_Set_AutoSize_To_False_Do_Not_Throws ()
+    public void Constructor_Setting_Width_Height_Before_Set_AutoSize_To_False ()
     {
-        var exception = Record.Exception (() => new Label { Width = 10, AutoSize = false });
-        Assert.Null(exception);
+        var label = new Label { Width = 10, AutoSize = false };
+        Assert.Equal("Absolute(10)", label.Width.ToString());
+        Assert.False(label.AutoSize);
 
-        exception = Record.Exception (() => new Label { Height = 10, AutoSize = false });
-        Assert.Null(exception);
+        label = new Label { Height = 10, AutoSize = false };
+        Assert.Equal("Absolute(10)", label.Height.ToString());
+        Assert.False(label.AutoSize);
     }
 }
--- a/UnitTests/Views/LabelTests.cs
+++ b/UnitTests/Views/LabelTests.cs
@@ -566,4 +566,13 @@ e
                                                      );
     }
 
+    [Fact]
+    public void Setting_Width_Height_Before_Set_AutoSize_To_False_Do_Not_Throws ()
+    {
+        var exception = Record.Exception (() => new Label { Width = 10, AutoSize = false });
+        Assert.Null(exception);
+
+        exception = Record.Exception (() => new Label { Height = 10, AutoSize = false });
+        Assert.Null(exception);
+    }
 }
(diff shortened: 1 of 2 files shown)


UnitTests/Views/LabelTests.cs
```diff
@@ -566,4 +566,13 @@ public void Without_Top_Border ()
                                                      );
     }
 
+    [Fact]
+    public void Setting_Width_Height_Before_Set_AutoSize_To_False_Do_Not_Throws ()
+    {
+        var exception = Record.Exception (() => new Label { Width = 10, AutoSize = false });
```
dodexahedron, line 572: This test is inherently redundant with a constructor test, as any thrown exception is a test failure.

Doing this is trying to prove a negative, which isn't really valid.

Instead, if there is a known and expected exception, a test for that exception should be written.

Otherwise, the constructor tests do this already.

That's why they removed the Assert.DoesNotThrow method in XUnit in the first place.

dodexahedron (changes requested):
Just a note about an unnecessary/redundant test case (and why).

UnitTests/Views/LabelTests.cs
BDisp: You suggesting only using the constructor and remove the `Record.Exception`? If it's that how I test the assertion?

dodexahedron: Actually suggesting removing this test entirely.

The known condition that would have thrown an exception before the fix should just be added to the constructor test. Un-fixed code would then fail the constructor test when the constructor is called, which is the appropriate place for it. Fixed code will pass.

dodexahedron: In other words, the existing constructor test was just not complete enough, if it wasn't already failing. Ensuring that case is covered makes the constructor test a stronger and more correct test from the start.

dodexahedron: In my opinion, they shouldn't have removed that assertion method, because there _are_ situations in which it is actually useful. This one just doesn't happen to be one of them, because the actual desired proof is that the constructor succeeds. Asserting that it does not throw is redundant, as an unhandled exception is already a failure of the test without any more work on your part.

BDisp: Done. Thanks.

dodexahedron (approved):
That, too, looks like a reasonable way to do it. :)

Side note for future test refinements around the project:

I imagine a lot of test classes could benefit greatly from having tests doing the general stuff this test is doing (calling a constructor plus an initializer) for the constructors of their respective types, too.

Question for this situation, in this type, in particular (but not meant for this PR):

By the time the constructor plus any associated initializer is finished, is it mandatory or otherwise expected that the IsInitialized value has been explicitly set (not specifically true or false - just explicitly assigned one way or the other)?

If so, we can actually have that enforced by the compiler by use of the `required` modifier on the property, and marking code that sets it with the [SetsRequiredMembers] attribute so that we can be sure there are no further corner cases now or in the future where that can happen.

That would also make initializers like the one you discovered that caused this bug to result in compile-time errors, instead of just leaving open potential run-time exceptions.

PR #3334, opened 2024-03-17
Fixes #3333. Application.UnGrabbedMouse event doesn't allow grab another view after exit.

### LINKED ISSUE #3333: Application.UnGrabbedMouse event doesn't allow grab another view after exit.
The issue is the `MouseGrabView` is always set to null at the end.

BDisp:
## Fixes

- Fixes #3333

## Proposed Changes/Todos

- [x] Sets MouseGrabView to null before call the event
- [x] Change unit test to proof
- [x] ScrollView now grab the mouse if any view want to grab it

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/ScrollView.cs
+++ b/Terminal.Gui/Views/ScrollView.cs
@@ -149,6 +149,8 @@ public class ScrollView : View
                 if (supView == _contentView)
                 {
                     Application.GrabMouse (this);
+
+                    break;
                 }
 
                 supView = supView.SuperView;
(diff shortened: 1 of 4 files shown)


Terminal.Gui/Views/ScrollView.cs
```diff
@@ -134,6 +136,26 @@ public ScrollView ()
...
+        if (parent is { })
+        {
+            var supView = parent.SuperView;
+
+            while (supView is { })
+            {
+                if (supView == _contentView)
+                {
+                    Application.GrabMouse (this);
+                }
```
tig, line 152: Are you sure there shouldn't be a `break;` here to exit the loop?

tig (changes requested):
It seems to me while loop should be cut short once a matching view is found?

Terminal.Gui/Views/ScrollView.cs
BDisp: You are absolutely right. When I did that I think to my self, don't forget to break the loop if found it. Damn and I forgot. Thanks.

tig (approved):
Good stuff! Well done.

PR #3339, opened 2024-03-19
Fixes #3338. `Application.Run/End` -> Callers must dispose Toplevel

### LINKED ISSUE #3338: #3330 broke `Dialog` Dispose
I broke something in

- #3330 

Not sure how I missed this in my testing.... and clearly more/better unit tests are needed.

Repo:

- Run UI Catalog
- Hit Ctrl-A to show About box
- Close About box
- Exit UI Catalog

You'll get an assert that the Dialog for the about box was not disposed.

tig:
Caused by #3300

## Fixes

- Fixes #3338 

## Proposed Changes/Todos

- [x] Find and fix all occurances of code not cleaning up the Toplevel used for `Run`

## Pull Request checklist:

### DIFF
--- a/UICatalog/Scenarios/BackgroundWorkerCollection.cs
+++ b/UICatalog/Scenarios/BackgroundWorkerCollection.cs
@@ -13,12 +13,16 @@ namespace UICatalog.Scenarios;
 [ScenarioCategory ("Controls")]
 public class BackgroundWorkerCollection : Scenario
 {
-    public override void Run () { Application.Run<OverlappedMain> (); }
+    public override void Run ()
+    {
+        Application.Run<OverlappedMain> ();
+        Application.Top.Dispose ();
+    }
 
     private class OverlappedMain : Toplevel
     {
         private readonly MenuBar _menu;
-        private readonly WorkerApp _workerApp;
+        private WorkerApp _workerApp;
         private bool _canOpenWorkerApp;
 
         public OverlappedMain ()
@@ -93,8 +97,6 @@ public class BackgroundWorkerCollection : Scenario
             Activate += OverlappedMain_Activate;
             Deactivate += OverlappedMain_Deactivate;
 
-            Closed += OverlappedMain_Closed;
-
             Application.Iteration += (s, a) =>
                                      {
                                          if (_canOpenWorkerApp && !_workerApp.Running && Application.OverlappedTop.Running)
@@ -165,15 +167,15 @@ public class BackgroundWorkerCollection : Scenario
             return new MenuBarItem ("_Window", new List<MenuItem []> { menuItems.ToArray () });
         }
 
-        private void OverlappedMain_Activate (object sender, ToplevelEventArgs top) { _workerApp.WriteLog ($"{top.Toplevel.Data} activate."); }
-
-        private void OverlappedMain_Closed (object sender, ToplevelEventArgs e)
+        private void OverlappedMain_Activate (object sender, ToplevelEventArgs top)
         {
-            _workerApp.Dispose ();
-            Dispose ();
+            _workerApp?.WriteLog ($"{top.Toplevel.Data} activate.");
         }
 
-        private void OverlappedMain_Deactivate (object sender, ToplevelEventArgs top) { _workerApp.WriteLog ($"{top.Toplevel.Data} deactivate."); }
+        private void OverlappedMain_Deactivate (object sender, ToplevelEventArgs top)
+        {
+            _workerApp?.WriteLog ($"{top.Toplevel.Data} deactivate.");
+        }
         private void Quit () { RequestStop (); }
 
         private MenuBarItem View ()
@@ -207,6 +209,15 @@ public class BackgroundWorkerCollection : Scenario
                                     "_View",
                                     new List<MenuItem []> { menuItems.Count == 0 ? new MenuItem [] { } : menuItems.ToArray () }
                                    );
+        }
+
+        /// <inheritdoc />
+        protected override void Dispose (bool disposing)
+        {
+            _workerApp?.Dispose ();
+            _workerApp = null;
+
+            base.Dispose (disposing);
         }
     }
 
@@ -299,7 +310,6 @@ public class BackgroundWorkerCollection : Scenario
 
         public Staging Staging { get; private set; }
         public event Action<StagingUIController> ReportClosed;
-        public void Run () { Application.Run (this); }
 
         private void OnReportClosed (object sender, EventArgs e)
         {
@@ -447,7 +457,8 @@ public class BackgroundWorkerCollection : Scenario
                                                  _stagingsUi.Add (stagingUI);
                                                  _stagingWorkers.Remove (staging);
 
-                                                 stagingUI.Run ();
+                                                 Application.Run (stagingUI);
+                                                 stagingUI.Dispose ();
                                              }
                                          };
 
@@ -465,8 +476,8 @@ public class BackgroundWorkerCollection : Scenario
 
                 _stagingWorkers.Add (staging, worker);
                 worker.RunWorkerAsync ();
-                stagingUI.Dispose ();
             }
+            stagingUI.Dispose ();
         }
 
         public void WriteLog (string msg)
--- a/UnitTests/Dialogs/DialogTests.cs
+++ b/UnitTests/Dialogs/DialogTests.cs
@@ -1331,4 +1331,33 @@ public class DialogTests
 
         return (Begin (dlg), dlg);
     }
+
+    [Fact]
+    [SetupFakeDriver]
+    [TestRespondersDisposed]
+    public void Run_Does_Not_Dispose_Dialog ()
+    {
+        Init ();
+        Dialog dlg = new ();
+
+        dlg.Ready += Dlg_Ready;
+
+        Run (dlg);
+
+#if DEBUG_IDISPOSABLE
+        Assert.False (dlg.WasDisposed);
+#endif
+
+        dlg.Dispose ();
+
+        Shutdown();
+
+        return;
+
+        void Dlg_Ready (object sender, EventArgs e)
+        {
+            RequestStop ();
+        }
+    }
+
 }
--- a/Terminal.Gui/FileServices/DefaultFileOperations.cs
+++ b/Terminal.Gui/FileServices/DefaultFileOperations.cs
@@ -162,6 +162,7 @@ public class DefaultFileOperations : IFileOperations
         dlg.AddButton (btnCancel);
 
         Application.Run (dlg);
+        dlg.Dispose ();
 
         result = tf.Text;
 
--- a/Terminal.Gui/Views/MessageBox.cs
+++ b/Terminal.Gui/Views/MessageBox.cs
@@ -468,6 +468,7 @@ public static class MessageBox
 
         // Run the modal; do not shutdown the mainloop driver when done
         Application.Run (d);
+        d.Dispose ();
 
         return Clicked;
     }
--- a/UICatalog/KeyBindingsDialog.cs
+++ b/UICatalog/KeyBindingsDialog.cs
@@ -82,6 +82,7 @@ internal class KeyBindingsDialog : Dialog
                            Application.RequestStop ();
                        };
         Application.Run (dlg);
+        dlg.Dispose ();
 
         if (key.HasValue)
         {
--- a/UICatalog/Scenarios/Adornments.cs
+++ b/UICatalog/Scenarios/Adornments.cs
@@ -120,6 +120,8 @@ public class Adornments : Scenario
         Application.Top.Closed += (s, e) => View.Diagnostics = _diagnosticFlags;
 
         Application.Run (editor);
+        editor.Dispose ();
+
         Application.Shutdown ();
     }
 
--- a/UICatalog/Scenarios/CharacterMap.cs
+++ b/UICatalog/Scenarios/CharacterMap.cs
@@ -898,6 +898,7 @@ internal class CharMap : ScrollView
                                    (s as Dialog)?.RequestStop ();
                                };
         Application.Run (waitIndicator);
+        waitIndicator.Dispose ();
 
         if (!string.IsNullOrEmpty (decResponse))
         {
@@ -1011,6 +1012,7 @@ internal class CharMap : ScrollView
             dlg.Add (json);
 
             Application.Run (dlg);
+            dlg.Dispose ();
         }
         else
         {
--- a/UICatalog/Scenarios/CsvEditor.cs
+++ b/UICatalog/Scenarios/CsvEditor.cs
@@ -316,6 +316,7 @@ public class CsvEditor : Scenario
         tf.SetFocus ();
 
         Application.Run (d);
+        d.Dispose ();
 
         enteredText = okPressed ? tf.Text : null;
 
@@ -455,6 +456,7 @@ public class CsvEditor : Scenario
         {
             Open (ofd.Path);
         }
+        ofd.Dispose ();
     }
 
     private void Open (string filename)
--- a/UICatalog/Scenarios/Dialogs.cs
+++ b/UICatalog/Scenarios/Dialogs.cs
@@ -192,6 +192,7 @@ public class Dialogs : Scenario
                                                                        buttonPressedLabel
                                                                       );
                                         Application.Run (dlg);
+                                        dlg.Dispose ();
                                     };
 
         Win.Add (showDialogButton);
--- a/UICatalog/Scenarios/DynamicMenuBar.cs
+++ b/UICatalog/Scenarios/DynamicMenuBar.cs
@@ -417,9 +417,9 @@ public class DynamicMenuBar : Scenario
                 EditMenuBarItem (_menuItem);
             }
 
-            var _btnOk = new Button { IsDefault = true, Text = "Ok" };
+            var btnOk = new Button { IsDefault = true, Text = "Ok" };
 
-            _btnOk.Accept += (s, e) =>
+            btnOk.Accept += (s, e) =>
                               {
                                   if (string.IsNullOrEmpty (TextTitle.Text))
                                   {
@@ -431,21 +431,22 @@ public class DynamicMenuBar : Scenario
                                       Application.RequestStop ();
                                   }
                               };
-            var _btnCancel = new Button { Text = "Cancel" };
+            var btnCancel = new Button { Text = "Cancel" };
 
-            _btnCancel.Accept += (s, e) =>
+            btnCancel.Accept += (s, e) =>
                                   {
                                       TextTitle.Text = string.Empty;
                                       Application.RequestStop ();
                                   };
-            var _dialog = new Dialog { Title = "Enter the menu details.", Buttons = [_btnOk, _btnCancel] };
+            var dialog = new Dialog { Title = "Enter the menu details.", Buttons = [btnOk, btnCancel] };
 
             Width = Dim.Fill ();
             Height = Dim.Fill () - 1;
-            _dialog.Add (this);
+            dialog.Add (this);
             TextTitle.SetFocus ();
             TextTitle.CursorPosition = TextTitle.Text.Length;
-            Application.Run (_dialog);
+            Application.Run (dialog);
+            dialog.Dispose ();
 
             if (valid)
             {
--- a/UICatalog/Scenarios/DynamicStatusBar.cs
+++ b/UICatalog/Scenarios/DynamicStatusBar.cs
@@ -253,9 +253,9 @@ public class DynamicStatusBar : Scenario
                 EditStatusItem (_statusItem);
             }
 
-            var _btnOk = new Button { IsDefault = true, Text = "OK" };
+            var btnOk = new Button { IsDefault = true, Text = "OK" };
 
-            _btnOk.Accept += (s, e) =>
+            btnOk.Accept += (s, e) =>
                               {
                                   if (string.IsNullOrEmpty (TextTitle.Text))
                                   {
@@ -275,21 +275,22 @@ public class DynamicStatusBar : Scenario
                                       Application.RequestStop ();
                                   }
                               };
-            var _btnCancel = new Button { Text = "Cancel" };
+            var btnCancel = new Button { Text = "Cancel" };
 
-            _btnCancel.Accept += (s, e) =>
+            btnCancel.Accept += (s, e) =>
                                   {
                                       TextTitle.Text = string.Empty;
                                       Application.RequestStop ();
                                   };
-            var _dialog = new Dialog { Title = "Enter the menu details.", Buttons = [_btnOk, _btnCancel] };
+            var dialog = new Dialog { Title = "Enter the menu details.", Buttons = [btnOk, btnCancel] };
 
             Width = Dim.Fill ();
             Height = Dim.Fill () - 1;
-            _dialog.Add (this);
+            dialog.Add (this);
             TextTitle.SetFocus ();
             TextTitle.CursorPosition = TextTitle.Text.Length;
-            Application.Run (_dialog);
+            Application.Run (dialog);
+            dialog.Dispose ();
 
             return valid
                        ? new DynamicStatusItem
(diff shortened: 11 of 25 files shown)


tig:
@BDisp Highly likely I messed up the BackgroundWorker scenarios here. They are convoluted examples and very hard to understand. I did my best and I think they are fine, but since you wrote them you should probably take a close look.

BDisp:
The `BackgroundWorker` scenario is really broken. Here is failing because the `Current` toplevel is already disposed and `Running` is still true and responding to mouse report position. I'm afraid something is still not well yet.

image

UnitTests/Dialogs/DialogTests.cs
```diff
@@ -1331,4 +1331,33 @@ public void Zero_Buttons_Works ()
 
         return (Begin (dlg), dlg);
     }
+
+    [Fact]
+    [SetupFakeDriver]
+    [TestRespondersDisposed]
+    public void Run_Does_Not_Dispose_Dialog ()
```
BDisp, line 1338: Did you realized that running this unit test alone never ends because it's using `WindowsDriver` and not `FakeDriver`. Also the `Ready` event isn't invoked because it's called after the `if (MainLoop.Running && MainLoop.EventsPending ())` where it never get out.

BDisp (changes requested):
I'm still think we don't need dispose all the views because the `Application` already handle it in the `End` and in the `Shutdown`,

BDisp:
I sent the PR https://github.com/tig/Terminal.Gui/pull/24 to this branch. I hope you like it.

tig:
I addressed the `WorkerLog` not being movable thusly:

ekQwBPw 1

UnitTests/Dialogs/DialogTests.cs
```diff
@@ -1331,4 +1331,37 @@ public void Zero_Buttons_Works ()
...
+#if DEBUG_IDISPOSABLE
+        Assert.True (dlg.WasDisposed);
+        Assert.True (Top.WasDisposed);
+        Assert.Equal (top, Top);
+#endif
+
+        // This is instance and the user is responsible to set to null or not
+        // because in the Application it's all working as expected.
+        // Fortunately, this instance is not null so that we can obtain the value of its properties
+        Assert.NotNull (dlg);
```
tig, line 1356: Looking at this again... I'm confused.

This code literally says: "`dig` has been disposed, but it's ok to access it because the reference is not `null`".

This is not ok (and is actually the point I raised in filing #3338).

tig (commented):
We're going in circles.

Am I wrong in asserting that it's not correct to access an object that has been disposed?

tig:
I've updated this PR to do the following:

- Moves the `Cancelled` property from `FileDialog` to `Dialog` so all dialogs can be checked for `Cancelled`
- Added a check to `Cancelled` get/set for `WasDisposed`
- Updated unit test to prove disposing Top in `Run/End` is incorrect (unit test current fails).

UnitTests/Dialogs/DialogTests.cs
BDisp: That's the user instance and not the Top and the Current instance. If the user instance is null how you can get a property value after the Current is disposed?

BDisp: It's up to user disposing the instance and up to Application dispose the Current runstate.

BDisp:
> Here's a new test that should pass (but doesn't):

You aren't testing this right.
Instead of `Assert.False (dlg.WasDisposed);`
Should be `Assert.True (dlg.WasDisposed);`

And in the `Canceled` property you must to

```cs
            if (!WasDisposed)
            {
                throw new ObjectDisposedException (GetType ().FullName);
            }
```

The PR that I submitted already do that way. You only asked why the `diag` isn't null. Because it's an instance but return the `WasDisposed` from the `runstate.Toplevel`. Can you explain why you think this is wrong?

UnitTests/Dialogs/DialogTests.cs
BDisp: I think you are looking for an issue where there is none.

tig:
> > Here's a new test that should pass (but doesn't):
> 
> You aren't testing this right. Instead of `Assert.False (dlg.WasDisposed);` Should be `Assert.True (dlg.WasDisposed);`
> 
> And in the `Canceled` property you must to
> 
> ```cs
>             if (!WasDisposed)
>             {
>                 throw new ObjectDisposedException (GetType ().FullName);
>             }
> ```
> 
> The PR that I submitted already do that way. You only asked why the `diag` isn't null. Because it's an instance but return the `WasDisposed` from the `runstate.Toplevel`. Can you explain why you think this is wrong?

Are you really telling me this code isn't correct?

```cs
        get
        {
#if DEBUG_IDISPOSABLE
            if (WasDisposed)
            {
                throw new ObjectDisposedException (GetType ().FullName);
            }
#endif
            return _canceled;
        }
```

BDisp:
> Are you really telling me this code isn't correct?
> 
> ```cs
>         get
>         {
> #if DEBUG_IDISPOSABLE
>             if (WasDisposed)
>             {
>                 throw new ObjectDisposedException (GetType ().FullName);
>             }
> #endif
>             return _canceled;
>         }
> ```

Yes because you are get the Canceled property after the runstate.Toplevel was disposed and it was really disposed. So only call it before.

tig:
Let me be more clear:

Are you saying the code I pasted above is the incorrect way to check if a disposed object is being accessed in general?

BDisp:
> Are you really telling me this code isn't correct?
> 
> ```cs
>         get
>         {
> #if DEBUG_IDISPOSABLE
>             if (WasDisposed)
>             {
>                 throw new ObjectDisposedException (GetType ().FullName);
>             }
> #endif
>             return _canceled;
>         }
> ```

Sorry, the code is correct but you are asserting after the dlg was really disposed. Assert it in the Dlg_Ready method.

BDisp:
This pass:

```cs
    [Fact]
    [AutoInitShutdown]
    public void Can_Access_Cancel_Property_After_Run ()
    {
        var top = Top;

        Dialog dlg = new ();

        dlg.Ready += Dlg_Ready;

        Run (dlg);

#if DEBUG_IDISPOSABLE
        Assert.True (dlg.WasDisposed);
        Assert.True (Top.WasDisposed);
        Assert.Equal (top, Top);
#endif

        Shutdown ();

        return;

        void Dlg_Ready (object sender, EventArgs e)
        {
            ((Dialog)sender).Canceled = true;
            Assert.True (dlg.Canceled);
            RequestStop ();
        }
    }

```

tig:
You are missing the point. 

'Run' is a synchronous method. Callers often need to get state of the toplevel they pass to run (eg Canceled,or in the case of FileDialog, the chosen file). The codebase is littered with code that does this. 

If, when 'Run' returns, the toplevel in question has been Disposed, IT IS AN INVALID OBJECT AND ACCESSING ITS PROPERTIES IS AN ERROR. IT IS ONLY LUCK IT WORKS IN MOST CASES TODAY (specifically because GC has not actually run).

BDisp:
> You are missing the point.
> 
> 'Run' is a synchronous method. Callers often need to get state of the toplevel they pass to run (eg Canceled,or in the case of FileDialog, the chosen file). The codebase is littered with code that does this.
> 
> If, when 'Run' returns, the toplevel in question has been Disposed, IT IS AN INVALID OBJECT AND ACCESSING ITS PROPERTIES IS AN ERROR. IT IS ONLY LUCK IT WORKS IN MOST CASES TODAY (specifically because GC has not actually run).

Ok I get the point but how you will guarantee that all derived View classes will dispose them after run?

BDisp:
You are right, must be the callers who have to dispose Toplevel. Must be possible to create some engineering that can control the last closed toplevel if it was disposed when calling Begin for a new toplevel or when the Shutdown was called. I'll also think better about this and try to help you.

tig:
Here's a thought:

Why have Init create and set Application.Top at all?

What if Run() (no parameters) acted just like Run<TopLevel>)?

Then, we remove the toplevel factory from Init. 

Then, we ensure Run<T>/End disposes the T that's created and Run(t) does not dispose the passed t object?

Any code that currently needs to modify Application.Top today will need to be modified to do

Application.Init();
Toplevel t = new();
...
Application.Run(t). 
t.Dispose();
Application.Shutdown();

Which is actually pretty clean.

BDisp:
> Here's a thought:
> 
> Why have Init create and set Application.Top at all?
> 
Unless we want to determine the real available terminal's area, but this can be achieved by the `Driver.Cols` and `Driver.Rows`. Also to it work as expected, e.g. to re-draw it need to being run. Thus my opinion is we don't need to create a toplevel on the `Init` method.

> What if Run() (no parameters) acted just like Run)?
> 
Didn't understanded this. I think you meaning by removing the Run() method and only use the Run(T). I agree and thus forces the user creating a toplevel before run.

> Then, we remove the toplevel factory from Init.
> 
Correct.

> Then, we ensure Run/End disposes the T that's created and Run(t) does not dispose the passed t object?
> 
Run/End only dispose the runstate created but not the passed T object, I agree.

> Any code that currently needs to modify Application.Top today will need to be modified to do
> 
> Application.Init();
> Toplevel t = new();
> ... Application.Run(t).
> t.Dispose();
> Application.Shutdown();
> 
> Which is actually pretty clean.

Yes, they needs to modified to forces the creating of the toplevel object and this is reasonable.

tig:
> > What if Run() (no parameters) acted just like Run)?
> 
> Didn't understanded this. I think you meaning by removing the Run() method and only use the Run(T). I agree and thus forces the user creating a toplevel before run.

What I mean is: there are currently 3 versions of `Application.Run`:

```cs
 public static void Run (Func<Exception, bool> errorHandler = null);
 public static void Run<T> (Func<Exception, bool> errorHandler = null, ConsoleDriver driver = null);
 public static void Run (Toplevel view, Func<Exception, bool> errorHandler = null);
```

In shorthand these are: `Run()`, `Run<T>`, and `Run(t)`.

What I'm proposing is the first one, `Run()`, currently does this:

```cs
public static void Run (Func<Exception, bool> errorHandler = null) { Run (Top, errorHandler); }
```

It would change to this:

```cs
public static void Run (Func<Exception, bool> errorHandler = null) { Run<Toplevel>(), errorHandler); }
```

BDisp:
It must be as the following, otherwise we it `throw new ArgumentException ($"{top.GetType ().Name} must be derived from TopLevel");`.

```cs
 public static void Run (Func<Exception, bool> errorHandler = null) { Run (new (), errorHandler); }
```

BDisp:
I submitted the PR https://github.com/tig/Terminal.Gui/pull/25, but it's still a WIP.

BDisp:
The `Run<T> Example` scenario isn't re-draw the `Dialog` correctly.

image

tig:
@BDisp after merging your PR tests are failing. The failure is in 

image

Any idea what's going on?

tig:
I just hacked in a fix.

BDisp:
> I just hacked in a fix.

Can you replace with this, please?

`_workerApp?.WriteLog ($"{(top.Toplevel is null ? ((Toplevel)sender).Data : top.Toplevel.Data)} activate.");`

I forgot to commit this before. Thanks.

tig:
@bdiisp please run BackgroundWorkercollection on your copy of this branch, show the Worker and then try to quit with ESC. Does it quit?

If you then use File.Quit does it throw?

BDisp:
> @bdiisp please run BackgroundWorkercollection on your copy of this branch, show the Worker and then try to quit with ESC. Does it quit?
> 
> If you then use File.Quit does it throw?

I sent a better solution for this because the intention is hiding it on pressing Esc. See https://github.com/tig/Terminal.Gui/pull/27.

tig:
Here's my thinking on `Scenario`. I'll either create a new issue/PR to address this or do it here, depending...

I over-engineered `Scenario` and tried to be both too helpful to devs and too cute. My intent was to make it super easy to create new scenarios. I thought it would be neat if Scenario developers only had to write a `Setup` override and the "default" logic in `Init` and `Run` would do magic for them. I also had UICatalog call `Application.Shutdown` when the Scenario exited instead of forcing Scenario authors to do this themselves.

This was a bad idea for these reasons:

1. New developers looking at scenarios as examples don't see what ones supposed to do to setup/teardown a Terminal.Gui app. They don't see the proper way: Call `Init`, and ensure it's bracketed by a `Shutdown`, Call `Run` and make sure the `Toplevel` used gets disposed. 
2. It led to un-needed complexity in 4 overrides: `Init`, `Setup`, `Run`, and `RequestStop`. This is confusing to devs and a survey of the existing scenarios shows this confusion.
3. It makes devs think that `Application.Top` can't be the actuall main view for the scenario, because `Init`, by default creates `Win` and adds it to `Top`. 

My proposal:

- Simplify `Scenario` to have only ONE virtrual metods: `Main`. Developers should treat this just like the standard C# console app `int Main (string [] args)` method.
- Add a `Application.Theme` property (a string) that sets the theme for the app. Do NOT reset this to `string.Empty` in `ResetState`. This way, when UI Catalog sets the Theme, it will be carried into any Scenario that's run. 
- Nuke the fancy UI Catalog concept of `ToplevelColorScheme` which was really just a test that is no longer really needed.
- Change all Scenarios to merge their `Init`, `Setup`, and `Run` overrides into `Main`. 

The Generic `Scenario.Main` will look like this:

```cs
    public override void Main ()
    {
        // Init
        Application.Init ();

        Window appWindow = new ()
        {
            Title = $"{Application.QuitKey} to Quit - Scenario: {GetName ()}",
        };

        var button = new Button { X = Pos.Center (), Y = Pos.Center (), Text = "Press me!" };
        button.Accept += (s, e) => MessageBox.ErrorQuery ("Error", "You pressed the button!", "Ok");
        appWindow.Add (button);

        Application.Run (appWindow);
        appWindow.Dispose ();

        Application.Shutdown ();
    }
```

BDisp:
Returning `Toplevel` only on the Run() and Run<T>() methods are fine. Void is more appropriate to the `Run(v)`.

tig:
> Returning `Toplevel` only on the Run() and Run_T_() methods are fine. Void is more appropriate to the `Run(v)`.
> 
> Why writing < and > are hiding?

Cool thing about `Run<T> and Run()` returning the toplevel is we can do this:

```cs
       Application.Top?.Dispose ();
```

tig:
I think this is ready to merge now. @BDisp you agree?

UICatalog/Scenarios/BackgroundWorkerCollection.cs
```diff
@@ -13,12 +14,25 @@ namespace UICatalog.Scenarios;
 [ScenarioCategory ("Controls")]
 public class BackgroundWorkerCollection : Scenario
 {
-    public override void Run () { Application.Run<OverlappedMain> (); }
+    public override void Init ()
+    {
+        Application.Run<OverlappedMain> ().Dispose ();
+        Application.Top?.Dispose ();
```
BDisp, line 20: I don't know why you have `Dispose` twice and why you removed the `Application.Shutdown` which you want to force the users to use.

BDisp: Should be:

```cs
        Application.Run<OverlappedMain> ().Dispose ();
        Application.Shutdown  ();
```

BDisp:
> I think this is ready to merge now. @BDisp you agree?

Yes, I agree. I only have the question above.

BDisp:
There is still a issue on the BackgroundWorkerCollection scenario.

- Open the scenario.
- Click on View menu and open the WorkApp
- Start two news Run Worker
- After they are open go to first that was open
- Press Esc to close it
- A blank window will be show
- Move the mouse over it
- System.NullReferenceException will be throw

Captura de ecrã 2024-03-25 234626

image

UICatalog/Scenarios/BackgroundWorkerCollection.cs
tig: I misunderstood your code. I thought you were disposing Application.Top in addition to the view returned from Run because it may be a different view since SetCurrentOverlappedAsTop changed Top. 

Removing shutdown was an error on my part.

tig: On your "you want to force users to call Shutdown" comment. 

Do you believe it's ok for apps to not call Application.Shutdown?

BDisp: No, I mean mandatory. It's the right thing to do.

BDisp: I think you're right also call Application.Top?.Dispose (); because it may not be the main. That's ensure not throwing the assertion.

BDisp:
There is still a issue when the `OverlappedTop` is the `Current` after a `OverlappedChildren` was closed and there is still a `OverlappedChildren` visible which must be on the top of the toplevel's stack.

*To repo:
- Open the scenario.
- Click on View menu and open the WorkApp
- Start two news Run Worker
- After they are open go to first that was open
- Press Esc to close it
- Press Esc to close the current
- Press Esc to toggle the visibility of the WorkerApp to false
- The WorkerApp is still visible
- Press again Esc and now the WorkerApp is hidden

The WorkerApp must be hidden after pressing the first Esc

UICatalog/Scenarios/BackgroundWorkerCollection.cs
tig: > I think you're right also call `Application.Top?.Dispose ();` because it may not be the main. That's ensure not throwing the assertion.

When we get around to supporting any view in Run, and refactoring overlapped, we really need to make the concept of when Application.Top gets set more deterministic. It's scary that neither of us can articulate exactly why Top (and/or Current) has changed when Run returns here. 

But, for now, I'm not overly concerned because the only place it currently happens is in user code where someone is both using "TopLevelOverlapped" and background tasks/threading ... and this Scenario is pretty much it.

BDisp: You are right. But this has a explanation. This can happen when we start a new RunWorker and close the scenario before it's be showing. I prevented them for not open if the `OverlappedTop` is null. I also fixed an issue where when the scenario is already closed, after start a new RunWorker and the async Post was is called after the `MainLoop` is null.
Maybe now the situation of the `Application.Top` isn't actually the `OverlappedTop`isn't now happen. I will remove the `Application.Top?.Dispose` and doing some tests to check it.

tig:
I found another bug in BWC:

- Ctrl-R to start a new worker
- ENTER to accept
- ListView crashes

The reason is when you create the staging UI on Ctrl-R you create the listview but don't populate it. By default it has focus, so ENTER goes to the listview. But there's no items so ListView crashes. 

The fix, which I've put in already, is to create the listview `Enabled = false` and in the constructor where you set the listview Source, do `_listView.Enabled = true`.

tig:
@BDisp - Ok if I merge this to v2_develop now? further bugs in BWC Scenario can be fixed in another PR?

BDisp:
> @BDisp - Ok if I merge this to v2_develop now? further bugs in BWC Scenario can be fixed in another PR?

Yes.

BDisp:
> I found another bug in BWC:
> 
> * Ctrl-R to start a new worker
> * ENTER to accept
> * ListView crashes
> 
> The reason is when you create the staging UI on Ctrl-R you create the listview but don't populate it. By default it has focus, so ENTER goes to the listview. But there's no items so ListView crashes.
> 
> The fix, which I've put in already, is to create the listview `Enabled = false` and in the constructor where you set the listview Source, do `_listView.Enabled = true`.

This started given the error because you enabled the `ListView` to start the `RunWorker`. I only enabled it after the work is finished to show the items. The ListView had a bug really but why you enabled it before on the window that will start the work?

tig:
> @tig you didn't change as I did in my PR https://github.com/tig/Terminal.Gui/pull/29/files. Sometimes is preferable I merge and resolving the conflicts with your branch than you manually change them.
> 
> image

Yes, I missed removing the Dispose call on line 488.

However, your code is incorrect. You can't dispose an object and then reference. It needs to be like this:

image

BDisp:
> However, your code is incorrect. You can't dispose an object and then reference. It needs to be like this:

It's correct because bellow is another object and not the same. See it again.

tig:
> > However, your code is incorrect. You can't dispose an object and then reference. It needs to be like this:
> 
> It's correct because bellow is another object and not the same. See it again.

Nope. You're missing it. In no universe is this code correct:

image

BDisp:
Edit:
Ah I see now, sorry. I didn't see the image before.

Here my suggestion:

```cs
            Application.Run (stagingUI);

            if (stagingUI.Staging != null && stagingUI.Staging.StartStaging != null)
            {
                stagingUI.Dispose ();

                staging = new Staging (stagingUI.Staging.StartStaging);
                WriteLog ($"Worker is started at {staging.StartStaging}.{staging.StartStaging:fff}");

                if (_stagingWorkers == null)
                {
                    _stagingWorkers = new Dictionary<Staging, BackgroundWorker> ();
                }

                _stagingWorkers.Add (staging, worker);
                worker.RunWorkerAsync ();
            }
            else
            {
                stagingUI.Dispose ();
            }

```

tig:
":

> Edit: Ah I see now, sorry. I didn't see the image before.
> 
> Here my suggestion:
> 
> ```cs
>             Application.Run (stagingUI);
> 
>             if (stagingUI.Staging != null && stagingUI.Staging.StartStaging != null)
>             {
>                 stagingUI.Dispose ();
> 
>                 staging = new Staging (stagingUI.Staging.StartStaging);
>                 WriteLog ($"Worker is started at {staging.StartStaging}.{staging.StartStaging:fff}");
> 
>                 if (_stagingWorkers == null)
>                 {
>                     _stagingWorkers = new Dictionary<Staging, BackgroundWorker> ();
>                 }
> 
>                 _stagingWorkers.Add (staging, worker);
>                 worker.RunWorkerAsync ();
>             }
>             else
>             {
>                 stagingUI.Dispose ();
>             }
> ```

Where does the `staging = new ...` instance get disposed?

UICatalog/Scenarios/BackgroundWorkerCollection.cs
```diff
@@ -90,29 +105,22 @@ public OverlappedMain ()
...
-        private void Menu_MenuOpening (object sender, MenuOpeningEventArgs menu)
+        private void OverlappedMain_Ready (object sender, EventArgs e)
         {
-            if (!_canOpenWorkerApp)
+            if (_workerApp?.Running == false)
             {
-                _canOpenWorkerApp = true;
+                Application.Run (_workerApp);
 
                 return;
```
BDisp, line 119: This return isn't necessary.

tig:
This is STILL incorrect code:

image

BDisp:
> Where does the `staging = new ...` instance get disposed?
Here.

image

BDisp:
> This is STILL incorrect code:
> 
> image

You are right. Put it before `worker.RunWorkerAsync ();`

tig:
> > Where does the `staging = new ...` instance get disposed?
> > Here.
> 
> image

Given this, then this seems correct:

image

Agree?

UICatalog/Scenarios/BackgroundWorkerCollection.cs
```diff
@@ -493,6 +492,7 @@ public void RunWorker ()
             if (stagingUI.Staging != null && stagingUI.Staging.StartStaging != null)
             {
                 staging = new Staging (stagingUI.Staging.StartStaging);
+                stagingUI.Dispose ();
```
BDisp, line 495: Not yet in the right place. 😃

BDisp:
That's now right. The scenario had a great refactoring. It's much more intuitive.

dodexahedron:
The ultimate solution to this (and a bunch of other things, actually), which I strongly encourage for post-V2, is to use the built-in DI container infrastructure in .net and just inject things into it when and as needed.

That also vastly simplifies things like testing, logging, configuration, and background tasks/services, each in appropriate scopes, plus being a mature, trusted, and standard core of modern .net applications, and also makes extending it even easier.

It's actually not that much effort to port existing code to, and could be done fairly easily for v2 (usually you can so it with almost no changes, in fact, and just get all the goodies for free), but I think we have enough on our plates for v2 already and gotta draw the line _somewhere._

Ok PC says 98% so maybe I can log in shortly, now. 😅

PR #3353, opened 2024-03-27
Fixes #3351. TabIndex with the same setter value but with wrong index return without set the correct value.

### LINKED ISSUE #3351: TabIndex with the same setter value but with wrong index return without set the correct value.
### Discussed in https://github.com/gui-cs/Terminal.Gui/discussions/3319

<div type='discussions-op-text'>

<sup>Originally posted by **dasien** March 13, 2024</sup>
Hello,

I was hoping to get some clarity on the purpose of the `View.TabIndexes` list.  What i was trying to accomplish was to join two views together, and have the first control on View2 `TabIndex` be + 1 from the last control in `View1.TabIndexes`. 

What i am finding is that regardless of what i put for the `TabIndex` value for controls in View 1, even if i say that certain controls have no `TabIndex` and i set the `TabStop` to false, the views with no `TabIndex` and `TabStop` set to false are still listed in the `TabIndexes` list which inflates the list with controls.  As a consequence, the value i put in for `TabIndex` for controls which i want to interact with the `Tab` key is overridden by their order in the `TabIndexes` list.

Thus, if i have a view with a `Label` with `TabStop = false` and nothing set for `TabIndex`, and a `TextField` with `TabStop = true` and `TabIndex = 1`, and I repeat that pattern say 3 times incrementing the `TabIndex` by 1 for each `TextField` so the next `TextField.TabIndex = 2` then 3, etc. the values reported for the `TextField` in the parent `View.Tablndexes` list is 1,3,5,7 respectively rather than the 1,2,3,4 i set.

I am confused why something is overwriting my set `TabIndex` values on controls at all and if there is a way for me to turn that off.  I could understand the system taking a best guess if i don't put values in and don't also turn off the fact that a particular `View` should be a tab stop, but it is overwriting the values which are present and making it difficult for me to compute the correct `TabIndex` of a joined `View`.</div>

BDisp:
## Fixes

- Fixes #3351

## Proposed Changes/Todos

- [x] Check if the index also belong to the view.
- [x] Add unit test to proof

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/View/ViewKeyboard.cs
+++ b/Terminal.Gui/View/ViewKeyboard.cs
@@ -267,7 +267,7 @@ public partial class View
                 return;
             }
 
-            if (_tabIndex == value)
+            if (_tabIndex == value && TabIndexes.IndexOf (this) == value)
             {
                 return;
             }
--- a/UnitTests/View/NavigationTests.cs
+++ b/UnitTests/View/NavigationTests.cs
@@ -1303,13 +1303,39 @@ public class NavigationTests
         v1.TabIndex = 2;
         v2.TabIndex = 1;
         v3.TabIndex = 0;
-        Assert.True (r.TabIndexes.IndexOf (v1) == 2);
-        Assert.True (r.TabIndexes.IndexOf (v2) == 1);
-        Assert.True (r.TabIndexes.IndexOf (v3) == 0);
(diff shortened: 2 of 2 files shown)


BDisp:
`XUnit.Assert` is contradicting itself, frankly. Using `Equal/NotEqual` causes failing on another unit tests.

image

tig (approved):

Issue #3344, opened 2024-03-21
A full-screen `Modal` view should not be movable/sizable


tig:
As noted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/3342#issuecomment-2010882239_

>`Toplevel` created by the `Application.Init` shouldn't be moveable, otherwise it will behave badly because it represent all the available space on a terminal.

It should also not be sizable (not currently supported, but will be when #2537 is addressed).

This is not just about `Toplevel`s created by `Init`. In the `Adornments` scenario, I do not use `Application.Top` but run the Scenario using `Application.Run (editor)`. The `editor` Toplevel is meant to be full screen... and it behaves badly as @BDisp  noted.

To fix this, I propose modifying the View move/drag code in Adornment (which will be moved to Border asap) to check for `Modal == true && IsFullScreen` and ignoring `Arrangement` if so.

We'll have to figure out what `IsFullScreen` means. It could be something like `if (X == 0 && Y == 0 && Width == Fill && Height == Fill)`

BDisp:
> We'll have to figure out what `IsFullScreen` means. It could be something like `if (X == 0 && Y == 0 && Width == Fill && Height == Fill)`

Well, that depends if the Parent has or not superview. If it doesn't have then it should be `if (X == 0 && Y == 0 && Width == Driver.Cols && Height == Driver.Rows)`

tig:
Wasn't what we thought it was. I had a bug in `OnResizeNeeded`. PR incoming.

Issue #3312, opened 2024-03-11
Mouse API makes it way too hard to track button pressed


tig:
To illustrate what I think is wrong with the current Terminal.Gui (and v1) API in this regard, consider this class.

```cs
    public class MouseDemo : View
    {
        public MouseDemo ()
        {
            CanFocus = true;
            MouseEvent += (s, e) =>
                          {
                              if (e.MouseEvent.Flags.HasFlag (MouseFlags.Button1Pressed))
                              {
                                  ColorScheme = Colors.ColorSchemes ["Toplevel"];
                              }
                              if (e.MouseEvent.Flags.HasFlag (MouseFlags.Button1Released))
                              {
                                  ColorScheme = Colors.ColorSchemes ["Dialog"];
                              }
                          };
            MouseLeave += (s, e) => { ColorScheme = Colors.ColorSchemes ["Dialog"]; };
        }
    }
```

I want this view to show the "TopLevel" colorscheme while button1 is pressed; when button1 is released, or the mouse leaves the view, the color should go back to "Dialog".

Here's how it behaves now. Note that if I press button1 outside of the view and drag into the view, the color changes to "Toplevel". This is not what I want. The current API provides me no way of knowing that the `Button1Pressed` flag was set while the mouse was outsdie of my view.

![5DhZHvn 1](https://github.com/gui-cs/Terminal.Gui/assets/585482/8fc7095a-8c99-41af-b7c1-ff5bc6ad0a9b)

Now, I can work around this like this, but this is awful.

```cs
    public class MouseDemo : View
    {
        private bool _button1PressedOnEnter = false;
        public MouseDemo ()
        {
            CanFocus = true;
            MouseEvent += (s, e) =>
                          {
                              if (e.MouseEvent.Flags.HasFlag (MouseFlags.Button1Pressed))
                              {
                                  if (!_button1PressedOnEnter)
                                  {
                                      ColorScheme = Colors.ColorSchemes ["Toplevel"];
                                  }
                              }
                              if (e.MouseEvent.Flags.HasFlag (MouseFlags.Button1Released))
                              {
                                  ColorScheme = Colors.ColorSchemes ["Dialog"];
                                  _button1PressedOnEnter = false;
                              }
                          };
            MouseLeave += (s, e) =>
                          {
                              ColorScheme = Colors.ColorSchemes ["Dialog"];
                              _button1PressedOnEnter = false;
                          };
            MouseEnter += (s, e) =>
                          {
[...]

PR #3372, opened 2024-04-01
Fixes #3370. Adds visual highlight for press and hold (for `Button`, `Checkbox`, `RadioGroup`, and `Border`)

### LINKED ISSUE #3370: Make `Button` support repeat via continuous presses
I want to be able to press and hold on a button and have `Accept` be invoked repeatedly as long as the button is pressed.

This prototype demonstrates:

```cs
        label = new Label ()
        {
            X = 0,
            Y = Pos.Bottom (moveUnicodeHotKeyBtn) + 1,
            Title = "_Numeric Up/Down:",
        };
        var downButton = new Button ()
        {
            CanFocus = false,
            AutoSize = false,
            X = Pos.Right(label)+1,
            Y = Pos.Top (label),
            Height = 1,
            Width = 1,
            NoPadding = true,
            NoDecorations = true,
            Title = $"{CM.Glyphs.DownArrow}",
            WantContinuousButtonPressed = true,
        };

        var numericEdit = new TextField ()
        {
            Text = "1966",
            X = Pos.Right (downButton),
            Y = Pos.Top (downButton),
            Width = 5,
            Height = 1,
        };
        var upButton = new Button ()
        {
            CanFocus = false,
            AutoSize = false,
            X = Pos.Right (numericEdit),
            Y = Pos.Top (numericEdit),
            Height = 1,
            Width = 1,
            NoPadding = true,
            NoDecorations = true,
            Title = $"{CM.Glyphs.UpArrow}",
            WantContinuousButtonPressed = true,
        };
        downButton.Accept += (s, e) =>
                             {
                                 numericEdit.Text = $"{int.Parse(numericEdit.Text) - 1}";
                             };
        upButton.Accept += (s, e) =>
                           {
                               numericEdit.Text = $"{int.Parse (numericEdit.Text) + 1}";
                           };

        Win.Add (label, downButton, numericEdit, upButton);
```

v375JYq 1

In addition to supporting use-cases like the one above, modifying `Button` to behave this way if `WantContinuousButtonPressed` will make it such that all the custom button code in `ScrollBar` can be removed and replaced with just using a `Button`.

tig:
## Fixes

- Fixes #3370

## Proposed Changes/Todos

- [x] Figure out how to visually show press-and-hold: `Color.GetHighlightColor` increases brightness
- [x] Fix logic issue in `View.OnMouseEvent` and `Application.OnMouseEvent`
- [x] Update `Buttons` scenario to demonstrate
- [x] Move code into `View` and add `View.InvertColorsOnPress` property
- [x] Update `Checkbox`, `Button`, `DatePicker`, and `ColorPicker` to utilize.
- [x] Update `RadioGroup` to utilize - partial - entire view inverts, not just the radio item under the mouse
- [x] Update API docs
- [x] Add new unit tests
- [x] Support `Border` for view drag
- [x] Draft (commented out) support for a hover effect
- [ ] Ensure works on `NetDriver` and `CursesDriver` - Dependent on #3374 being fixed (#3375).

## Pull Request checklist:


tig:
@BDisp give this PR a try.

Bring it down locally and note how i've implemented a UI effect on Button that shows when it's been pressed/released.

Note that if you do a mouse button press on a button, and then move off of it, it does NOT click, but if you move it back (while still holding the button) down and then release, it does click.

I think this is what you want:

- "prevents many times when accidently press a button and avoiding the clicked after moving the mouse. 

...AND is what I want:

- Makes it so that you can be moving the mouse a little when you try to click

C8keSye 1

BDisp:
Yes that fit what I said. It isn't clicking in the same point that matters but in the same object. That's good.

tig:
> Yes that fit what I said. It isn't clicking in the same point that matters but in the same object. That's good.

I *think* this means removing the `_processButtonClick` related logic from WindowsDriver, and replicating what I've done in this PR with `Button` into `View`?

I just did a quick test with this:

image

And Button works perfectly (I think).

tig:
Also, I'd love your thoughts on the "repeat accelaration" I put in.

See how the Accept Count accelerates after 2 seconds of holding?

NQNW3Iz 1

https://github.com/gui-cs/Terminal.Gui/blob/1cca002393da3b3fbb87c6b53f6dfb051c61f755/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs#L1711-L1748

Ideally this would be configurable, and I don't like the coupling between `ConsoleDriver` having a reference to `Application.WantContinuousButtonPressedView `. So I'd like to figure out how to move the logic out of `WindowsDriver` and into `Application` (or `View` if that makes sense).

Terminal.Gui/View/ViewMouse.cs
```diff
@@ -114,53 +103,147 @@ protected internal virtual bool OnMouseEvent (MouseEvent mouseEvent)
...
+    ///         Called when the mouse is either clicked or double-clicked. Check
+    ///         <see cref="MouseEvent.Flags"/> to see which button was clicked.
+    ///     </para>
     /// </remarks>
     protected bool OnMouseClick (MouseEventEventArgs args)
     {
         if (!Enabled)
         {
+            // QUESTION: Is this right? Should a disabled view eat mouse clicks?
+            args.Handled = true;
```
BDisp, line 217: Yes, Like disabled mustn't respond to keyboard events they mustn't respond to mouse inputs. The situation where it can be used is in the TGD, which I think @tznind will enable it temporarily to it can be moved, clicked or respond to keyboard for setting something or whatever.

BDisp:
Open `All Views Tester` scenario and move the mouse around the window without pressing. A null reference will be throws .

image

BDisp:
Open `Buttons` scenario and move the mouse around the window without pressing. The screen will be all black. Pressing Esc it's possible to exit.

https://github.com/gui-cs/Terminal.Gui/assets/13117724/5fef59a7-491f-4047-91c2-505f28a56fcf

BDisp:
The window is resizing even without pressing any mouse button but only by moving the mouse.

tig:
> Do this issue is happening to you? Click on the `Toplevel` button and after the `Dialog` is open click on the `Yes` button keeping the button pressed. You'll see that the `Dialog` width increasing before it close.
> 
> 
> 
> https://github.com/gui-cs/Terminal.Gui/assets/13117724/3f7f1565-cb3f-432c-8eb7-925371d1522b
> 
> 
> 
> 

Yes. This is unrelated to this. We need to open a separate issue.

BDisp:
> Yes. This is unrelated to this. We need to open a separate issue.

I know that is unrelated to this. Doesn't we just can talk? Damn.

Terminal.Gui/View/ViewMouse.cs
tig: Responding to events is different than preventing other views from getting them.

`Enabled` means: "Gets or sets whether this View can respond to user interaction."

It does NOT mean "gets or sets whether this View eats all mouse and keyboard input, preventing the Superview from getting them".

Note, in Applicaiton.cs:

https://github.com/gui-cs/Terminal.Gui/blob/43e4a836645036b6aa8acc148d9e714d1bf47e09/Terminal.Gui/Application.cs#L1591C1-L1596C1

In #3323 I need to fix this, like this:

```cs
        while (view is {} && !view.OnMouseEvent (me))
        {
            if (view is Adornment ad)
            {
                view = ad.Parent.SuperView;
            }
            else
            {
                view = view.SuperView;
            }
        }
```

But, for now, I'm going to leave it as-is in this PR.

tig:
> The window is resizing even without pressing any mouse button but only by moving the mouse.
> 
> image
> 
> image

I"ve seen this too. I'll dig in to it.

tig:
> > Yes. This is unrelated to this. We need to open a separate issue.
> 
> I know that is unrelated to this. Doesn't we just can talk? Damn.

Yes, we can "just talk". However, we can "just talk" in an organized way.

Every Issue and PR has a specific topic. It confuses things if we mix topics. It would be more effective if you just created a new issue for things like this if you know it's is unrelated. I've asked you to do this before, but you continue to do it.

tig:
> Open `All Views Tester` scenario and move the mouse around the window without pressing. A null reference will be throws .
> 
> image

Good catch. Looking into it.

Terminal.Gui/View/ViewMouse.cs
BDisp: Ok I see what you are meaning. Yes I agree it should bubble up the event for the superview can handle them, but the current disabled view doesn't process it. So in the case a view is disabled it can't process the event for itself but just returning false. That is the right way to do it.

tig:
> Open `Buttons` scenario and move the mouse around the window without pressing. The screen will be all black. Pressing Esc it's possible to exit.
> 
>  WindowsTerminal_uRMkWTbEX6.mp4

I can't reproduce this. 

However, if I press and hold the mouse button while moving, and cross the border, the Window resizes (which it shouldn't). Do you think this is what you're seeing?

tig:
> > The window is resizing even without pressing any mouse button but only by moving the mouse.
> > image
> > image
> 
> I"ve seen this too. I'll dig in to it.

This is caused by a combo of:

- #3344 
- #3312 

This happens in v2_develop as well.

BDisp:
> Every Issue and PR has a specific topic. It confuses things if we mix topics. It would be more effective if you just created a new issue for things like this if you know it's is unrelated. I've asked you to do this before, but you continue to do it.

Ok I already opened an issue for each of them, but sometimes I only want a positive/negative response before continue. And is in these cases where you complicate the things. e.g. if you respond like "Yes I'm also seeing that issue. You may open an issue for it". That's a good answer, but just saying "It's unrelated with this one. It's better open an issue for it". It's a bad response because you didn't confirmed first it the issue really exist or not.

BDisp:
> I can't reproduce this.
> 
> However, if I press and hold the mouse button while moving, and cross the border, the Window resizes (which it shouldn't). Do you think this is what you're seeing?

You have to double click the `Buttons` scenario and move the mouse while the scenario is opening. With `NetDriver` and `CursesDriver` this issue doesn't happens. Maybe is related with how `Windows Terminal` handles  the mouse using the Win32 API.

BDisp:
This commit https://github.com/gui-cs/Terminal.Gui/pull/3372/commits/a8d6f846d74e99ba53ba2e63eb378e679b8c77e9 Fixes #3379.

Confirmed. Thanks @tig.

BDisp:
> > I can't reproduce this.
> > However, if I press and hold the mouse button while moving, and cross the border, the Window resizes (which it shouldn't). Do you think this is what you're seeing?
> 
> You have to double click the `Buttons` scenario and move the mouse while the scenario is opening. With `NetDriver` and `CursesDriver` this issue doesn't happens. Maybe is related with how `Windows Terminal` handles the mouse using the Win32 API.

This doesn't reproduce on the v2_develop anymore. Maybe I swapped from this branch to the v2_develop and run debug without compile it first. But this branch is reproducing this issue.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -1695,7 +1695,13 @@ private MouseFlags ProcessButtonClick (WindowsConsole.MouseEventRecord mouseEven
 
     private async Task ProcessButtonDoubleClickedAsync ()
     {
-        await Task.Delay (300);
+        // Only delay if there's not a view wanting continuous button presses
+        if (Application.WantContinuousButtonPressedView is null)
+        {
+            // QUESTION: Why 300ms?
+            await Task.Delay (300);
+        }
```
BDisp, line 1703: I think isn't awaiting at all and exit if `WantContinuousButtonPressedView` isn't null another button pressed that could chain on a double click or triple click will always sending button pressed.

BDisp (changes requested):
Maybe I found the cause of the continuous mouse button pressed. It worth take a look at it.

tig:
Reposting this from @bdisp to the right Issue:

> When a button is pressed but not released doesn't restore the original color when the cursor leave the button and the mouse is released outside it. So I think the color should be restored when the mouse leaves the button and not waiting for the mouse releasing.
> 
>  WindowsTerminal_ibEkiAlmbr.mp4

tig:
> Reposting this from @BDisp to the right Issue:
> 
> > When a button is pressed but not released doesn't restore the original color when the cursor leave the button and the mouse is released outside it. So I think the color should be restored when the mouse leaves the button and not waiting for the mouse releasing.
> > WindowsTerminal_ibEkiAlmbr.mp4

I had this working. But then somehow broke it. 

Here's the behavior I want for `WantContinousButtonPresses == false:

## Case 1 - Press/Release - acts like Click

- Press within Bounds
  - Color Highlights
- Release within Bounds
  - Color is now Focus

## Case 2 - Press and move out of bounds, then release

- Press within Bounds
  - Color Highlights
- While holding, move mouse out of Bounds
  - Color stays highlight
- Release out of bounds
   - Color goes to Focus
   - Accept does NOT get invoked
  
## Case 3 - Press and move out of bounds, then move back in, then release

- Press within Bounds
  - Color Highlights
- While holding, move mouse out of Bounds
  - Color stays highlight
- Move back in to bounds
   - Color stays highlight 
- Release in bounds 
   - Accept gets invoked

Like I said, I had this working. This is pretty close to how Windows' buttons work (although newer Windows buttons have an intermediate "you're pressing but you moved out of bounds" color.

BDisp:
That also depends of what you are really doing. If you are dragging a view and want always highlighted until releasing it, that ok. But if you are generating a animation of a button press then if the mouse is outside of his bounds then the color is restored to alert that the button isn't selected if the mouse button is released. You also need to decide if you pressing and holding on a button and move the mouse to another button what the effect you want. Restores the previous and highlight the new button?

tig:
@BDisp - huge update here. Please take a look!

BDisp:
This is weird. If the window cannot move, it must be prevented from moving.

https://github.com/gui-cs/Terminal.Gui/assets/13117724/f4d94262-6d8a-463d-b279-090d13704a6e

BDisp:
This code work but you have to override the `OnDrawContent` on the `Button.cs`:

```cs
    private bool _isMouseEnter;

    /// <inheritdoc />
    protected internal override bool OnMouseEnter (MouseEvent mouseEvent)
    {
        if (mouseEvent.Flags == MouseFlags.ReportMousePosition)
        {
            _isMouseEnter = true;
            SetNeedsDisplay ();
        }

        return base.OnMouseEnter (mouseEvent);
    }

    /// <inheritdoc />
    protected internal override bool OnMouseLeave (MouseEvent mouseEvent)
    {
        if (mouseEvent.Flags == MouseFlags.ReportMousePosition)
        {
            _isMouseEnter = false;
            SetNeedsDisplay ();
        }

        return base.OnMouseLeave (mouseEvent);
    }

    /// <inheritdoc />
    public override void OnDrawContent (Rectangle contentArea)
    {
        TextFormatter?.Draw (
                             BoundsToScreen (contentArea),
                             HasFocus || (_isMouseEnter && CanFocus && !HasFocus) ? GetFocusColor () : GetNormalColor (),
                             HasFocus ? ColorScheme.HotFocus : GetHotNormalColor (),
                             Rectangle.Empty
                            );
    }

```

https://github.com/gui-cs/Terminal.Gui/assets/13117724/44700de6-4556-4d4d-98be-469746a6d915

BDisp:
@tig I hope you like the https://github.com/tig/Terminal.Gui/pull/31. So any view that want this feature only need to set `HighlightOnMouseEnter` property to true.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -1695,7 +1696,13 @@ private MouseFlags ProcessButtonClick (WindowsConsole.MouseEventRecord mouseEven
 
     private async Task ProcessButtonDoubleClickedAsync ()
     {
-        await Task.Delay (300);
+        // Only delay if there's not a view wanting continuous button presses
+        if (Application.WantContinuousButtonPressedView is null)
```
BDisp, line 1700: This isn't necessary because this async method is only called if the `_isButtonDoubleClicked or _isOneFingerDoubleClicked` is `true`. So, even a view which `WantContinuousButtonPressed` is `true` may want a `ButtonTripleClicked` event. Another detail, for a click, double click or triple click event occurs, it means that were also button released each time, which will generate the clicked events.
Am I missing something?

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -1695,7 +1696,13 @@ private MouseFlags ProcessButtonClick (WindowsConsole.MouseEventRecord mouseEven
...
+        {
+            // QUESTION: Why 300ms?
+            await Task.Delay (300);
```
BDisp, line 1703: In the PR #3386 I decreased to 200 to not wait so long.

dodexahedron:
This feature is a nice touch. 👍

dodexahedron:
That approach is still easy to have an all-in-one with, as well. You just include all the modules in an all in one library project and publish single file and 💥: there's your all-in-one.

tig:
> where pressing and holding alt will highlight or underline all the hotkeys of elements on screen,

Sadly, getting just `Alt` key presses is a struggle. It barely works on `WindowsDriver` and seems impossible on *nix.

I *am* excited for #1275 though!

dodexahedron:
Figures 😆 

And yes, I love the improvements!

Feels like things are getting shored up nicely. 🙂

PR #3375, opened 2024-04-02
Fixes #3374 - `Curses` and `Net` drivers don't generate click events unless mouse has not moved

### LINKED ISSUE #3374: Not always getting `Button1Clicked` events on `Netdriver` or `CursesDriver`
On both `NetDriver` and `CursesDriver`:

- Run UI Catalog
- Hit Ctrl-A to show about box
- Move mouse over OK button and press button1 - nothing happens
- Don't move mouse and click OK button again - about box exits

Debugging this by adding `Debug.WriteLine` to `ProcessMouseEvent`:

If the mouse has not moved since last mouse event, we get 

```
Button1Pressed
Button1Released
Button1Clicked
```

If the mouse HAS moved we get

```
Button1Pressed
Button1Released
```

This happens on both these drivers. On WIndowsDriver we get:

```
Button1Pressed
Button1Released
Button1Clicked
```

WindowsDriver has a ton of logic in it, that I don't fully understand, apparently to generated clicked events that neither NetDriver or Curses has.

tig:
## Fixes

- Fixes #3374

## Proposed Changes/Todos

- [x] Add debug spew to further diagnose
- [ ] ???

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/NetDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/NetDriver.cs
@@ -1137,7 +1137,7 @@ internal class NetDriver : ConsoleDriver
                 break;
             case EventType.Mouse:
                 MouseEvent me = ToDriverMouse (inputEvent.MouseEvent);
-                //Debug.WriteLine ($"NetDriver: ({me.X},{me.Y}) - {me.Flags}");
+                Debug.WriteLine ($"NetDriver: ({me.X},{me.Y}) - {me.Flags}");
                 OnMouseEvent (new MouseEventEventArgs (me));
 
                 break;
--- a/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
@@ -1393,18 +1393,20 @@ internal class WindowsDriver : ConsoleDriver
                     break;
                 }
 
+                Debug.WriteLine ($"WindowsDriver: ({me.X},{me.Y}) - {me.Flags}");
+
                 OnMouseEvent (new MouseEventEventArgs (me));
 
                 if (_processButtonClick)
                 {
-                    OnMouseEvent (
-                                  new MouseEventEventArgs (
-                                                           new MouseEvent
-                                                           {
-                                                               X = me.X,
-                                                               Y = me.Y,
-                                                               Flags = ProcessButtonClick (inputEvent.MouseEvent)
-                                                           }));
+                    me = new MouseEvent
+                    {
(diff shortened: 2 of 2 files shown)


tig:
@BDisp - I suspect you have more knowledge of this than I. Do you have any suggestions on how to address this?

BDisp:
The clicked event is only generated if the mouse location is the same on the buttonPressed and buttonReleased. This is the normal behavior of the clicked event. If you press the mouse button and move the mouse keeping the button pressed, when you release the button the location isn't the same and thus the clicked event isn't processed.

tig:
> The clicked event is only generated if the mouse location is the same on the buttonPressed and buttonReleased. This is the normal behavior of the clicked event. If you press the mouse button and move the mouse keeping the button pressed, when you release the button the location isn't the same and thus the clicked event isn't processed.
> 
> This allows a user cancel a click event if the button was pressed and then regret do it and move the mouse to another place and release it avoiding a clicked event.

I'm not sure I love the logic you've described. It means that the user has to hold the mouse perfectly still to cause a clicked event. 

However, we can debate that another day.

Please try the steps I outlined in #3374 - You'll see that even though the mouse is perfectly still, pressing the button and releasing it does not cause either driver to receive a clicked event.

BDisp:
> Please try the steps I outlined in #3374 - You'll see that even though the mouse is perfectly still, pressing the button and releasing it does not cause either driver to receive a clicked event.

The bug is in this line. The `point` isn't assigned if it isn't null. In that line `point` must always be equal to `pos`.

https://github.com/gui-cs/Terminal.Gui/blob/43e4a836645036b6aa8acc148d9e714d1bf47e09/Terminal.Gui/ConsoleDrivers/EscSeqUtils/EscSeqUtils.cs#L778


To fix change to:

`point = pos;`

BDisp:
> I'm not sure I love the logic you've described. It means that the user has to hold the mouse perfectly still to cause a clicked event.
> 
> However, we can debate that another day.

I hope you consider this because many app use that behavior and prevents many times when accidently press a button and avoiding the clicked after moving the mouse. If a user want to catch all the clicks he can use the `ButtonPressed` event.

tig:
> > I'm not sure I love the logic you've described. It means that the user has to hold the mouse perfectly still to cause a clicked event.
> > However, we can debate that another day.
> 
> I hope you consider this because many app use that behavior and prevents many times when accidently press a button and avoiding the clicked after moving the mouse. If a user want to catch all the clicks he can use the `ButtonPressed` event.

Lets discuss this here: 

- #3372

tig:
Stale. Closing.

PR #3376, opened 2024-04-02
Fixes #2114. Adds `View.ShadowStyle` - Enabled but not turned on by default

### LINKED ISSUE #2114: ContextMenu: If there's no MenuBar, F9 should open/close the context menu
Totally a minor issue, but it would be a nice enhancement to make things more consistent.

But consider this: If there IS a MenuBar, and a Context Menu is open, what should F9 do?

### LINKED ISSUE #2144: Norton Commander style buttons: Add half-height bottom shadow for Effect3D

tig:
## Fixes

- #2144 
- #3566 

## Todo

- [x] Extend `Margin` to have a "Shadow" Style - When enabled draws the correct half-height block glyphs in the right places
- [x] Extend HighlightStyle to support shadow animation on Pressed 
- [x] Enable `Button` to use `Margin.EnableShadow`
- [x] Figure out how to make `Button` with shadow work reliably and automatically. 
- [x] Figure out how to have the shadow account for whatever was drawn below. Right now, it blindly overwrites and thus doesn't look right in some situations. I think the right design is to have `Margin.Draw` get called by `Application`, not the View's SuperView.
- [x] Test & Enable `EnableShadow` on other Views where it may make sense visually.

## Background

When I created Frames (now Adornments) i had a vision that `Margin` could be used to enable the 3D effect that we had mostly working in v1, but got negated with all the changes to v2. 

I had a shower thought recently on this and decided to quickly code a prototype up to see if my instincts were right.

This PR demonstrates that I was. The concept is leveraging Margin like this:

```cs
 var button = new Button
 {
     X = Pos.Center (),
     Y = Pos.Center (), Text = "Press me!"
 };
 button.Margin.Thickness = new Thickness (0, 0, 1, 1);
 button.Margin.Add (
                    new View ()
                    {
                        X = Pos.AnchorEnd (1),
                        Y = 1,
                        Width = 1,
                        Height = Dim.Fill (),
                        ColorScheme = new ColorScheme (new Attribute (ColorName.DarkGray))
                    },
                    new View ()
                    {
                        X = 1,
                        Y = Pos.AnchorEnd (1),
                        Width = Dim.Fill(),
                        Height = 1,
                        ColorScheme = new ColorScheme (new Attribute (ColorName.DarkGray))
                    }
                    );

 bool pressed = false;
 button.Accept += (s, e) =>
                  {
                      pressed = !pressed;

                      if (pressed)
                      {
                          button.Margin.Thickness = new Thickness (1, 1, 0, 0);
                      }
                      else
                      {
                          button.Margin.Thickness = new Thickness (0, 0, 1, 1);
                      }
                  };
```

XZjxkvE 1

## Pull Request checklist:


tig:
Latest:

https://i.imgur.com/Vqghso7.gif

It's this easy:

```cs
        var button = new Button
        {
            X = Pos.Center (),
            Y = Pos.Center (), Text = "Press me!",
        };
        button.Margin.Shadow = true;
```

tig:
Vqghso7 1


> Latest:
> 
> Vqghso7 1 [ Vqghso7 1 ](https://private-user-images.githubusercontent.com/585482/332575609-2863ea72-a604-410d-924e-e5b3c14cc1f9.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTYzMjQzNzcsIm5iZiI6MTcxNjMyNDA3NywicGF0aCI6Ii81ODU0ODIvMzMyNTc1NjA5LTI4NjNlYTcyLWE2MDQtNDEwZC05MjRlLWU1YjNjMTRjYzFmOS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNTIxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDUyMVQyMDQxMTdaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT04NWVjMDI5ZjE2ODc1YzlhNDRmMTAzYjkxMDFkZmM4YjgxYTZmNDFjYjA2ZDMyZDIwN2ZlMmJmODA0OTk3NGYyJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.of_lQM6GZ3daZOXH2RQrElMeWkFF96QXfoRIxHnz3WY) [ ](https://private-user-images.githubusercontent.com/585482/332575609-2863ea72-a604-410d-924e-e5b3c14cc1f9.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTYzMjQzNzcsIm5iZiI6MTcxNjMyNDA3NywicGF0aCI6Ii81ODU0ODIvMzMyNTc1NjA5LTI4NjNlYTcyLWE2MDQtNDEwZC05MjRlLWU1YjNjMTRjYzFmOS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNTIxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDUyMVQyMDQxMTdaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT04NWVjMDI5ZjE2ODc1YzlhNDRmMTAzYjkxMDFkZmM4YjgxYTZmNDFjYjA2ZDMyZDIwN2ZlMmJmODA0OTk3NGYyJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.of_lQM6GZ3daZOXH2RQrElMeWkFF96QXfoRIxHnz3WY)
> 
> It's this easy:
> 
> ```cs
>         var button = new Button
>         {
>             X = Pos.Center (),
>             Y = Pos.Center (), Text = "Press me!",
>         };
>         button.Margin.Shadow = true;
> ```

Of course, the animation will only work on WindowsDriver, because it's the only one that supports distinct mouse up/down events.

tig:
Latest 

- Enables `Window.DefaultShadow`  and `Button.DefaultShadow` to be configured via Config Manager.
- The UI Catalog theme sets both to true.
- AdornmentsEditor now has a Shadow checkbox

xT7e2hE 1

(Since the Default theme doesn't explicitly set either they stay set even after changing back to the Default theme).

tig:
@dodexahedron Can you please look at this PR and help me out? I mentioned a similar issue the other day.

This PR was originally rebased with #3415.

#3415 was just merged into v2_develop.

I then rebased this PR on to v2_develop

```
v2_2144-3D-effect> git rebase --onto v2_develop
v2_2144-3D-effect> git pull
v2_2144-3D-effect> git push
Everything up-to-date
```

However, if you look at the "Files Changed" in this PR on Github it is comparing to an old v2_develop. For example, it things Aligner.cs is a new file:

image

I am sure I'm doing something wrong with how I'm rebasing. Do you see anything?

tig:
FWIW, here's how "AdornmentEditor" has evolved in this PR. 

Note the new "Expander" on the Border.

y0uGNZO 1

Terminal.Gui/View/Adornment/Border.cs
```diff
@@ -375,7 +375,7 @@ private void Application_UnGrabbingMouse (object sender, GrabMouseEventArgs e)
         }
     }
 
-#endregion Mouse Support
+    #endregion Mouse Support
```
dodexahedron, line 378: Just an optional/discretional request for visibility:

Regions are exceptions to formatting rules, particularly around sorting, and `partial` types were _partially_ (ha) created as a "better" means of doing regions.

It may be worth considering splitting such things out into nested files using names like (for this one) `Border.Mouse.cs` style of naming, where files named X.Y.cs get nested under X.cs in the solution explorer (unless you turn off nesting like a crazy person).

You can even just put multiple type parts for the same type in the same file, if you want, which still achieves the same result as regions, but doesn't hide things from someone looking for stuff, as regions have an unfortunate tendency to do, being purely preprocessor elements.

tig:
@dodexahedron did you see my question to you above?

tig:
This illustrates an issue with my current design here:

OXKUNLd 1

Need to figure out how to have the shadow account for whatever was drawn below. Right now, it blindly overwrites and thus doesn't look right in some situations. I think the right design is to have Margin.Draw get called by Application, not the View's SuperView.

dodexahedron:
That's kinda been a long-standing conundrum, hasn't it?

What order to draw things in, I mean.

Since we don't have an explicit concept of Z index, it's all just order dependent, which is something that can change at runtime as views are created and destroyed.

As far as what color to use, I've got blend code that's aimed at true color but works for lame color too. I may have even put at least some of it in when I did that work on the Color type a while back. 🤔

Terminal.Gui/View/Adornment/Border.cs
```diff
@@ -198,6 +198,9 @@ public LineStyle LineStyle
 
     private bool _showTitle = true;
 
+    /// <summary>
+    ///     Gets or sets whether the title should be shown. The default is <see langword="true"/>.
+    /// </summary>
     public bool ShowTitle
```
dodexahedron, line 204: I just had a thought about this that I want to test out at some point.

Since we do not have constructors and are thus using a property-based dependency model, I wonder how these might behave if set to a different value in an initializer, especially if other properties any property setter depends on may have already been set. 🤔

My thinking is that it's a potential for strange NullReferenceExceptions and potential slow resource leakage due to event subscriptions from objects whose references haven't even been pushed onto the evaluation stack yet.

Purely theoretical but making the note both for myself and in case anyone already has specific insight on that behavior in the CLR.

tig:
```cs
/// <summary>
///     Defines the style of shadow to be drawn on the right and bottom sides of the <see cref="View"/>. 
/// </summary>
public enum ShadowStyle
{
    /// <summary>
    /// No shadow.
    /// </summary>
    None,

    /// <summary>
    ///    A shadow that is drawn using block elements. Ideal for smaller views such as buttons.
    /// </summary>
    Opaque,

    /// <summary>
    ///     A shadow that is drawn using the underlying text with a darker background. Ideal for larger views such as windows.
    /// </summary>
    Transparent
}
```

tWZl0zl 1

I think Opaque looks better for Button as it's less intrusive. And no need for transparency because buttons can't be moved.

I think Transparent looks better for overlapped Toplevels.

Is it ok that they are different?

Should we turn on these effects by default (they all can be configured via CM).

Terminal.Gui/Resources/config.json
```diff
@@ -10,16 +10,16 @@
   // note that not all values here will be recreated (e.g. the Light and Dark themes and any property initialized
   // null).
   //
-  "$schema": "https://gui-cs.github.io/Terminal.Gui/schemas/tui-config-schema.json",
+  // TODO: V2 - Reference via http
+  "$schema": "../../docfx/schemas/tui-config-schema.json",
```
dodexahedron, line 14: Including it as a content file and just using a relative reference is probably a better idea anyway.

Because changes to the schema otherwise apply to versions that don't have code to go with them.

You could always make it a link to a tag or a permalink, but that also means changes that otherwise could be avoided by just including it. 🤷‍♂️ 

Could even _also_ have a commented out URL or a $comment element with the URL to go with it, to have it both ways for reference.

dodexahedron (approved):
OK, it's claiming I have 3 comments in this review, but I just wrote the one... 🤔 

Maybe it's including previous comments?

Guess I'll see after I submit it...

dodexahedron:
> OK, it's claiming I have 3 comments in this review, but I just wrote the one... 🤔
> 
> Maybe it's including previous comments?
> 
> Guess I'll see after I submit it...

Yep - it was including earlier comments it seems.

Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -78,7 +78,7 @@ internal override int Calculate (int location, int superviewContentSize, View us
                 // TODO: This whole body of code is a WIP (for https://github.com/gui-cs/Terminal.Gui/pull/3451).
                 subviewsSize = 0;
 
-                List<View> visibleSubviews = us.Subviews.Where (v => v.Visible).ToList ();
+                List<View> includedSubviews = us.Subviews.ToList();//.Where (v => !v.ExcludeFromLayout).ToList ();
```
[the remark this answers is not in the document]
tig: Will fix later.

tig:
I have these effects turned off by default and plan on merging it that way...

Why? Because we have so many fragile unit tests that can't deal with 

a) Button not being 1 row high.
b) Margin.Thickness != Empty.

Same thing goes for `BorderStyle != Single`.

Please, please, please, lets not write any more unit tests that depend on visual styling and if you're in a unit test as part of a PR that does this re-write it!

BDisp (approved):
Excellent work.

tig:
Off by default, but both the UI Catalog and NEW Hot Dog Stand themes enable for Button and Dialog.

IobTegt 1

UnitTests/Views/ShortcutTests.cs
```diff
@@ -314,4 +314,260 @@ public void Focus_CanFocus_CommandView_Add_Tracks ()
...
+
+        current.Dispose ();
+    }
+
+    [Theory]
+    //  0123456789
+    // " C  0  A "
+    [InlineData (-1, 0, 0)]
+    [InlineData (0, 1, 1)]
+    [InlineData (1, 0, 1)] // BUGBUG: This should be 1,1,1. We need to fix the logic in the Shortcut class.
```
dodexahedron, line 365: As mentioned here (re-posting here in case the commit comment is not easily visible):

https://github.com/tig/Terminal.Gui/commit/5cf556c376ea4f41399fb323a265f05e776e8641#r143586671

When you want to skip a test temporarily, feed a string to the Skip property, in the attribute constructor, as shown below and in the linked commit comment.

It's something attribute constructors can do that normal ones can't, and is basically like using an initializer.

For this one, this code, instead of the line with the comment and altered data to make it pass, will instead skip and print a message, without failing and without having to feed it a fake value that may or may not continue to pass if other things change:

```csharp
[InlineData (1, 1, 1, Skip = "// BUGBUG: We need to fix the logic in the Shortcut class.")]
```

dodexahedron: And it doesn't lie about the test results, either. 😅

dodexahedron (commented):
A tip for test cases you need to skip.

UnitTests/Views/ShortcutTests.cs
dodexahedron: I didn't know that was there til just when I looked at this.

But it occurred to me that NUnit has that capability, and it isn't exactly an uncommon thing to need to do, so surely xUnit must have it?

And sure enough, it does. :)

dodexahedron: Actually, you know what?

I'll make a tag for the todo list specifically for skipped tests that looks for something like `\[\((?:.*Skip *\= *\")(.*)\" *\)\]` (or whatever I have to tweak that to, to make it just match the string in the actual regex).

That'll give them their own category so they don't add noise to the others.

dodexahedron: Booooooo. The ToDo list feature does not support matching in generic code elements, so even though this pattern quite nicely matches the string after a `Skip = "`, it doesn't work to tag it because it isn't a comment, identifier, or string.

That said, I can just make it look for a specific string in them, like "Skipping".

Oh well.

I'll just let strings match then. Same tags available.

PR #3383, opened 2024-04-03
Fixes #3344. `OnResizeNeeded` using Top's Bounds for Top. 

### LINKED ISSUE #3344: A full-screen `Modal` view should not be movable/sizable

tig:
## Fixes

- Fixes #3344

## Proposed Changes/Todos

- [x] Only border drags now, not margin.
- [x] Found, fixed, updated unit tests.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/View/Adornment/Adornment.cs
+++ b/Terminal.Gui/View/Adornment/Adornment.cs
@@ -24,8 +24,6 @@ public class Adornment : View
     /// <param name="parent"></param>
     public Adornment (View parent)
     {
-        Application.GrabbingMouse += Application_GrabbingMouse;
-        Application.UnGrabbingMouse += Application_UnGrabbingMouse;
         CanFocus = true;
         Parent = parent;
     }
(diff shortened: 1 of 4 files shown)


BDisp (approved):
That the right way to manage the moveable of the view by handle it only on the Border. If has a Margin then the new position must take it at  account. Good work.

Issue #3284, opened 2024-03-01
Make `OnMouseXXX` follow the correct pattern


tig:
Make `OnMouseXXX` follow the correct pattern per https://github.com/gui-cs/Terminal.Gui/issues/3209

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3281#issuecomment-1973798898_

dodexahedron:
Hey I'm all for it, but it's no problem if you want to not bother yet and just leave that for when I do that work. I'll be doing the same thing to a bunch of code anyway, so I already was assuming I'd have to do it.

But less work for me is still cool haha. :)

But, if you do go ahead and knock it out yourself, here's the general way I've got current stuff designed on my old branch:

Almost every event uses `EventHandler<TypeUsedForAndNamedAfterThisEvent>` delegates for the events, where that type is a `public record` type, `sealed` if prudent, does not inherit from `EventArgs` (because we have no need for that), and carries a CancellationToken with it.

I have an interface for all such types to implement, but it's likely to change anyway, so don't worry about that. The CancellationToken is definitely a part of it, though, so if you do that from the start that's awesome.

There are also (Verb)ing and (Verb)ed events, for the majority of them, and the wrappers that invoke them use them in a consistent way.

Generally, that means:
 - The incoming value is first checked against current state. If no change, the events are NOT raised and the operation is canceled or returns, as appropriate in context. Otherwise:
 - If a change _is_ going to be made, the (Verb)ing event is raised.
   - Some events may also store rollback data, if feasible and sensible.
   - Subscribers can respect or ignore cooperative cancellation at their option. That's not on us to dictate.
 - After that, the change is made. If successful, the (Verb)ed event is raised.
 - If, at any point in all of this, an exception is thrown that bubbles up to the wrapper, we can either allow the program to crash, attempt rollback, or eat the exception.
   - Typically, for something like this, letting it go unhandled is probably the right thing to do. We can't safely assume anything about the state of the application at this point. Environment.FailFast() may be warranted.
   - A log hook at this point is probably a good idea, no matter what we do.
   - Rollback attempts can result in stack overflows because the same exception is quite possible to happen during the rollback, too. Should probably really only be done for events that are not exposed in the public API surface.
   - Eating the exception is dangerous in too many ways, especially for a library, so I'm pretty strongly against that.

dodexahedron:
Note that not everything has an -ing and -ed event.

[...]

Issue #3209, opened 2024-01-23
Formalize the Cancellable Work Pattern as v2s core eventing model


dodexahedron:
The *Cancellable Work Pattern* is integral to Terminal.Gui, enabling flexible, decoupled workflows across scheme management, rendering, input, commands, and properties. Proposed helper classes streamline its implementation, particularly for property changes like `Orientation`. The `Scheme` system leverages the CWP for dynamic styling, while the `Command` system’s updates (`Command.Activate`, `Handled`, `PropagatedCommands`) enhance clarity and coordination. Ongoing documentation efforts ensure these improvements are well-communicated, supporting robust UI development in `v2_develop`.

This will be addressed in this PR:

- #4092

## Original Post

As mentioned in discussion #3206, event handling could use some TLC.

This issue is meant to be the main issue, though it may spawn others, depending on just how involved this ends up being.

Work items I wanted to address, many of which already have been partially completed in existing work on v2 or may not currently even exist/be problems in the current code, but I'm listing these out for completeness and as reminders for myself to at least confirm them all:
 - Standardize event delegates according to design recommendations
   - Basically, exactly how the Terminal.Gui design specification already says to do, but with an additional degree of scrutiny around the type arguments to `EventHandler<T>`, considering potential needs for future changes/features added to TG.
 - OnX wrappers for raising events are public, but most should be `protected virtual` and none should be `public`.
 - Declaring the actual events inside interfaces, which are then implemented by classes that need to expose those events.
 - Making sure that events are first defined at the most-derived type that is reasonable or necessary for common handling (the interfaces will help with that, too).
 - Ensure that, when relevant, derived classes ask their base class to raise the event, and that there are no virtual/overridden events nor any events hidden with `new` (I don't think either of those last things exist in the code)
 - Ensuring that standard rules and practices for inheritance and related concepts such as overrides are followed as closely as possible.
 - Consider and, if prudent, implement necessary constructs for thread-safety around event raising
    - Including at least being aware that subscribers may provide an `async void` handler that returns before the associated background work is actually complete. Though this doesn't necessarily mean actively attempting to defend against it, since that's 100% the consumer's fault.
 - Ensure/implement more formal/recommended handling of exceptions thrown in subsribers' handlers, as appropriate, via the following methods, as appropriate:
   - Stopping processing of the invocation list when an exception is caught and adding any relevant information to the stack trace before re-throwing the caught exception
[...]

PR #3393, opened 2024-04-09
Fixes #3312. Mouse API makes it way too hard to track button pressed.

### LINKED ISSUE #3312: Mouse API makes it way too hard to track button pressed

BDisp:
## Fixes

- Fixes #3312

## Proposed Changes/Todos

- [x] Add mouse clicked logic for only to raise on the same view that pressed  it
- [x] If a mouse button was pressed and grabbed the mouse, the OnMouseEnter isn't ran. Now it run
- [x] Created the ProcessMouseEvent method allowing reuse the code
- [x] Add unit test
- [x] Add hot color for highlight
- [x] Some formatting

## Pull Request checklist:


Terminal.Gui/Application.cs
```diff
@@ -1432,6 +1435,9 @@ private static void OnUnGrabbedMouse (View view)
 
     // Used by OnMouseEvent to track the last view that was clicked on.
     internal static View? _mouseEnteredView;
+    internal static View? _lastViewButtonPressed;
```
tig, line 1438: We should not be adding more static fields to View. We've learned the hard way that static fields are dangerous and error-prone. We should be moving to REMOVE static fields from View.

I think these pieces of data should be part of `Application` (perhaps encapsulated in a class that `Application` holds a static reference to... e.g. `class MouseState`, or `ConsoleDriver`.

One way to think about this is:

- For instance objects (like `View`) avoid static fields
- `Application` knows what a `View` is and it's ok for `Application` to hold references to `View` objects or `view`-specific objects.
- `ConsoleDriver` should NOT know what `View` is, but since `ConsoleDriver`s are singletons, like `Application`, they are good places to hold persistent data.

Do you agree?

Terminal.Gui/Application.cs
```diff
@@ -143,6 +143,9 @@ internal static void ResetState ()
 
         // Mouse
         _mouseEnteredView = null;
+        _lastViewButtonPressed = null;
+        _canProcessClickedEvent = true;
+        _isMouseDown = false;
```
tig, line 148: We already have `MouseFlags.Button1Pressed`. The ENTIRE point of this Issue (#3312) is that we have confused the meaning of `Pressed`. 

Instead of a design that reflects how users actually use the mouse, we have a mess. 

"Pressed" is a past-tense of the verb "Press". The "-ed" indicates the "Press" happened in the past. Thus "Pressed" LITERALLY means "Is Currently Down".

"Released" is also past-tense. It means "Is No Longer Down".

I believe the terminology we should use (and the designers of Windows, WinForms, Mac Carbon, and most other UI libraries I've worked with) should distinguish, clearly, between 

- The button WAS not pressed, but it just became pressed. Down makes the most sense to me, but the verb Press works too. 
- The button was previously pressed, and continues to be so: Pressed
- The button was previously pressed, but it has just become un-pressed: Released

You are proposing to make things even more confusing. What this code says is:

- The button WAS not pressed, but it just became pressed. Pressed.
- The button was previously pressed, and continues to be so: IsMouseDown
- The button was previously pressed, but it has just become un-pressed: Released

This may seem sublte and insignificant to you, but I don't think it is.

Terminal.Gui/Input/Mouse.cs
```diff
@@ -143,6 +143,11 @@ public class MouseEvent
     /// </remarks>
     public Point ScreenPosition { get; set; }
 
+    /// <summary>
+    ///     Indicates if the current mouse event has first pressed <see langword="true"/>, latest released <see langword="false"/> or none <see langword="null"/>.
+    /// </summary>
+    public bool? IsMouseDown { get; set; }
```
tig, line 149: See my comments above about the words "Pressed" and "Down". 

In addition, I find it confusing that some mouse state is stored in `MouseEvent` and some is stored in `MouseEvent.Flags`. 

I don't have clarity myself on how we decide, but we should decide explicitly.

tig (changes requested):
I have not had a chance to fully review this, nor run it to see how it works. So I can't comment on whether it works.

But i do not think it actually makes tracking button presses easier, and I do not like aspects of the design. I've commented on those in my code review.

Terminal.Gui/Application.cs
BDisp: I agree for not use them in the View class, but this is in the `Application` which is static. I only made them internal for unit tests purposes and not for use them on the `ConsoleDriver`.

Terminal.Gui/Application.cs
BDisp: Well they are only internal static variables to hold and control the flux of the mouse events. Say what names you want for them. They do what you need for the `ViewMouse.cs` handle them through the `IsMouseDown` property.

Terminal.Gui/Input/Mouse.cs
BDisp: I decided to create a separate property to avoid with more confusion with flags. Pressed was always kept press. It's only a semantic matter.

tig: > It's only a semantic matter.

Semantics are actually the MOST important part of getting clarity on an issue. 

I feel VERY strongly about this. 

Please read this:  

https://people.howstuffworks.com/semantics.htm

BDisp: The `MouseDown` and the `MouseUp` are what you proposed for when the mouse button is press at first time and when the mouse is release at end, respectively. I'm only am using the internal fields for that. Do you prefer I create new Flags for them? In my opinion is duplicating flags because if the `IsMouseDown` is true (`MouseDown`) or false (`MouseUp`) the user can get the right button by inspecting the `ButtonPressed` and the `ButtonReleased`. You know I'm not very good with documentation and any help with this is appreciated. So, for me is much more easy you to say what you want I change in code.

BDisp: You are talking about refactor almost everything and how it will to work later. I only used the existent flags and get them work with Down, Up and some more existent bugs .If you want me to read your famous brain with its brilliant ideas, forget it. If you want to go your own way, feel free and you can close this PR. It's one more.

tig: > You are talking about refactor almost everything and how it will to work later.

Yes, I am taking about refactoring at the core. Why? 

Because we know the existing mouse API is has serious problems. For example:

- #3209
- #3284

I am pushing for fixing these things so they are as perfect as they can be. 

I am resisting hacking/patching/workarounds.

BDisp: I know I'm not a perfectionist but you could take what's good about this PR and improve it even more. An API is not created all at once, but rather through constant corrections and improvements. Sometimes it's a mistake to do everything at once.

tig: > I know I'm not a perfectionist but you could take what's good about this PR and improve it even more. An API is not created all at once, but rather through constant corrections and improvements. Sometimes it's a mistake to do everything at once.

I told you above that I didn't have the time to do more than a quick code review.

BDisp: > I told you above that I didn't have the time to do more than a quick code review.

I would say when you have time available. In your quick review you may miss something that you don't immediately understand, as you mentioned that I was using static variables in the `ConsoleDriver` but it is in the `Application`. Only after you properly analyze the code tell me what this PR really does not meet what you initially intended. I'm not referring to the rules of good coding which I know I'm not an expert, but in relation to mouse down, mouse up and mouse move. Thanks.

tig: >  but in relation to mouse down, mouse up and mouse move. 

Do you want me to code the example View I described above and submit it as a PR? Because that is how I would actually dive deep into this.

BDisp: > You misread my comments. I was suggesting that some of these data items that are statics on View in this PR could make sense to be fields on ConsoleDriver.

But I'm not using static variables in the `View` in this PR but in the `Application`. Am I missing something? If so, I apologize. With age, neurons no longer work well ;-)

BDisp: > Do you want me to code the example View I described above and submit it as a PR? Because that is how I would actually dive deep into this.

Of course. That's how I like to work. If you find that something in my PR is really useful but you want to modify something, I would just appreciate it if you sent me PR for my branch and so, as a team, we can resolve what is not right, improving it.

BDisp: It may not be at the correct standards or what you intended, but I think this PR substantially improves mouse processing.

Terminal.Gui/Application.cs
dodexahedron: I also think singletons for this purpose aren't ideal, and are treating Terminal.Gui like an application more than a framework/library. Every static means something a consumer can now only have one instance of in the _entire_ consuming AppDomain, which is problematic for anyone trying to make a multi-user application.

We do need to remember our place in the stack from the consumer perspective, after all. We don't own the app. It owns Terminal.Gui.

My SnapsInAZfs application that uses Terminal.Gui, for example, doesn't even _call_ anything related to terminal.gui unless specifically told to, for example. And using statics for things means that, if users attach to it at the same time (planned future feature), state is shared and that's going to break a lot of things, with the only way to work around it being to run in different processes and have to use IPC mechanisms.

dodexahedron: Well... The planned feature is actually aimed more at letting the same user have multiple UIs up that are using the same process, such as in different panes in a tmux session. Different users are a whole different ballgame.

From my perspective as a consumer of the library (any UI library, for that matter), the same process should be able to host as many instances of Terminal.Gui as I want, so long as they each get handed their own screen buffer. But statics that aren't truly just type properties and are instead simulated globals make that dangerous or impossible.

dodexahedron: And, to be clear, I realize this isn't the only one and there are many others. But they're all resolvable and avoiding adding more will be a big help for that.

dodexahedron (approved):

Terminal.Gui/Application.cs
dodexahedron: Making properties and fields instance instead of static is something for another time and discussion, but I thought I'd mention here since it was already called out.

BDisp: I understand that static components may be safer in cases of threading, but in the case of the mouse or keyboard it makes no sense to create separate instances because they are treated as static for the entire application. Therefore, the greatest care to take into account is to ensure that their values correspond to their current states and are not from previous residual values.

BDisp:
@tig please say if this PR is of interest or not. If it is not necessary I will no longer waste time on it and will close it. Thanks.

tig:
> @tig please say if this PR is of interest or not. If it is not necessary I will no longer waste time on it and will close it. Thanks.

I think we should fix this issue as well as the other deep design issues the current mouse support has. 

I would love it if you tackled that.

However, I feel like you ignored a bunch of feedback I gave in my review.

I'm not sure what to do about that.

Would you like me to do a fresh review of this after you fix the merge conflict?

BDisp:
> Would you like me to do a fresh review of this after you fix the merge conflict?

Yes, I'll fix the merge conflicts, again. Thanks.
But if you also have other better options to improve this, I can close this PR and you can create another one, no problem.

tig:
I think we can use this PR as the base.

Here's a recap of my opinions on this:


> We already have `MouseFlags.Button1Pressed`. The ENTIRE point of this Issue (#3312) is that we have confused the meaning of `Pressed`.
> 
> Instead of a design that reflects how users actually use the mouse, we have a mess.
> 
> "Pressed" is a past-tense of the verb "Press". The "-ed" indicates the "Press" happened in the past. Thus "Pressed" LITERALLY means "Is Currently Down".
> 
> "Released" is also past-tense. It means "Is No Longer Down".
> 
> I believe the terminology we should use (and the designers of Windows, WinForms, Mac Carbon, and most other UI libraries I've worked with) should distinguish, clearly, between
> 
> * The button WAS not pressed, but it just became pressed. Down makes the most sense to me, but the verb Press works too.
> * The button was previously pressed, and continues to be so: Pressed
> * The button was previously pressed, but it has just become un-pressed: Released
> 
> You are proposing to make things even more confusing. What this code says is:
> 
> * The button WAS not pressed, but it just became pressed. Pressed.
> * The button was previously pressed, and continues to be so: IsMouseDown
> * The button was previously pressed, but it has just become un-pressed: Released
> 
> This may seem sublte and insignificant to you, but I don't think it is.

And...

> Let's try this... pretend we are building a new View that wants the following user behavior:
> 
> The view shows the following counters:
> 
> * "clicked"
> * "doubleclicked"
> * "tripleclicked"
> * "down"
> * "pressed"
> * "up"
> 
> Assume the view only cares about Button1.
> 
> * If the user does a typical "mouse click", increment the "clicked" counter by one.
> * If the user repeatedly does a "mouse click" 10 times in a row, the "clicked" counter increases by 10
> * If the user does a "mouse click" twice in a row, the "clicked" counter increases by 2 and the "doubleclicked" counter increases by one
> * If the user does a "mouse click" three times in a row, the "clicked" counter increases by 3, and the "tripleclicked" counter increases by one.
> * If the users presses the mouse down, wiggles it (inside or outside the view) and then releases it, the "down" counter increases by one, the "pressed" counter increases by some number (depends on how long), and the "up" counter increases by one.
> 
> In C#, the setup code for this View would look like this:
> 
> ```cs
> MouseDown += Handle_Down;
> MouseUp += Handle_Down;
> MouseClicked += Handle_Clicked;
> MouseEvent += HandleEvent;
> ```
> 
> What would the logic in `Handle_Down` look like, ideally?
> 
> * It needs to know what mouse button was involved (because I stated a requirement for this view that it only cares about button1).
> 
> Should that be `if (e.HasFlag(Button1Pressed))`? Or `if (e.HasFlag(Button1)`? Or `if (e.Button1 == true)`? Or something else?
> 
> I don't have time right now to ask the same level of questions about `MouseUp`/`MouseDown` etc..., but in my mind that's how we figure out what the right design is.

I think after fixing the merge conflict, you should build a new Scenario that implements the View I describe above. 

We can then use that to debate/refine the design and API.

BDisp:
`MouseUp += Handle_Down;`

I think you meant what's below, right?

`MouseUp += Handle_Up;`

BDisp:
Done @tig.

tig:
Please fix build warnings.

> Done @tig.

What are you referring to with "Done"? 

I don't see a new Scenario as we discussed above, so I'm confused.

Terminal.Gui/Application.cs
dodexahedron: That's not a safe assumption to make, inside or outside the code of Terminal.Gui.

Any concurrent operation from the same or another thread, even within Terminal.Gui itself, can access anything static like that.

If static, it needs careful attention to synchronization, via locking, semaphores, etc., but that is a non-trivial and easily bug-prone task.

tig:
How's your progress on this coming @bdisp? Should it be marked as a draft for now?

BDisp:
> How's your progress on this coming @BDisp? Should it be marked as a draft for now?

I haven't worked on this for a while now. I'll mark it as a draft and take a good look at what you intend, including the scenario. But in my opinion this can only be handled in `Application.OnMouseEvent` but then don't complain as there could probably be even more static variables.

Terminal.Gui/Application.cs
dodexahedron: I don't think you need to close it, really.

Probably plenty of it that can be used, just not that specific piece.

Or we could close, and if you just don't delete the branch, we can still refer to any parts we need, when we get to it.

PR #3402, opened 2024-04-13
Fixes #3366 - Add ability to disable AutoSize

### LINKED ISSUE #3366: Must set AutoSize to false before setting Width
> Must set AutoSize to false before setting Width

I get this exception a lot now when trying to upgrade TGD.  Is it a temporary measure?

It seems like this is an attempt to prevent two conflicting sizing mechanisms (auto and Width).  If it is here to stay then why instead not just make AutoSize a `Dim` e.g.

```csharp
myView.Width = Dim.AutoSize()
// Or
myView.Width = Dim.AutoWidth(myView)
// Or
myView.Width = Dim.AutoSize(myView,width: true)
```

If it were a Dim then it wouldn't be possible to set both, Width would either be auto or other.

Maybe a View has to implement `IAutoSizeable` or something to support this approach.

tznind:
## Fixes

- Fixes #3366

## Proposed Changes/Todos

~Adds a static variable that allows disableing AutoSize.~

__Disables AutoSize instead of throwing Exception when setting Width/Height on a View__

Currently the designer emits code as follows:

```csharp
 this.lblLoading = new Terminal.Gui.Label();
 this.lblLoading.Data = "lblLoading";
 this.lblLoading.Text = "Please wait ...";
 this.lblLoading.Width = 36;
 this.lblLoading.Height = 1;
 this.lblLoading.X = 1;
 this.lblLoading.Y = 1;
 this.lblLoading.TextAlignment = TextAlignment.Left;
 this.Add(this.lblLoading);
```

I looked into adding AutoSize = false on every single view generated and every piece of code emitted.  That kinda worked but was ugly.  Then I realized I had to update all my existing .Designer.cs files code with this `AutoSize=false` too.  

I think this is a better solution, the ability to disable autosize across the board.  

This can be removed once AutoSize itself is removed.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/View/Layout/ViewLayout.cs
+++ b/Terminal.Gui/View/Layout/ViewLayout.cs
@@ -35,13 +35,6 @@ public enum LayoutStyle
 
 public partial class View
 {
-    /// <summary>
-    /// Determines the default usage of <see cref="View.AutoSize"/> in views
-    /// that support it (e.g. <see cref="Button"/>.  Set to false if you frequently
-    /// get Exceptions about changing Width when AutoSize is true.
-    /// </summary>
-    public static bool DefaultAutoSize = true;
-
     #region Frame
 
     private Rectangle _frame;
--- a/Terminal.Gui/View/Layout/ViewLayout.cs
+++ b/Terminal.Gui/View/Layout/ViewLayout.cs
@@ -35,6 +35,13 @@ public enum LayoutStyle
 
 public partial class View
 {
+    /// <summary>
+    /// Determines the default usage of <see cref="View.AutoSize"/> in views
+    /// that support it (e.g. <see cref="Button"/>.  Set to false if you frequently
+    /// get Exceptions about changing Width when AutoSize is true.
+    /// </summary>
+    public static bool DefaultAutoSize = true;
+
     #region Frame
 
     private Rectangle _frame;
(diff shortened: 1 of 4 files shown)


tig:
Love. 

I'll look deeper later today.

tig:
I have a better idea. Now that the exception I put in has weeded out all/most code that is incorrect, we can remove it and instead just change `AutoSize` to `false`:

```cs
    public Dim Width
    {
        get => VerifyIsInitialized (_width, nameof (Width));
        set
        {
            if (Equals (_width, value))
            {
                return;
            }

            _width = value ?? throw new ArgumentNullException (nameof (value), @$"{nameof (Width)} cannot be null");

            if (AutoSize)
            {
                Debug.WriteLine (@$"Must set AutoSize to false before setting {nameof (Width)}.");
                AutoSize = false;
            }

            bool isValidNewAutoSize = AutoSize && IsValidAutoSizeWidth (_width);

            if (IsAdded && AutoSize && !isValidNewAutoSize)
            {
                Debug.WriteLine (@$"Must set AutoSize to false before setting {nameof (Width)}.");
                AutoSize = false;
            }

            OnResizeNeeded ();
        }
    }
```

If I do this, all unit tests pass and I see no functional issues in some light tests...

tznind:
> we can remove it and instead just change AutoSize to false

Works for me!

tznind:
Ah I see how you mean.  Replace the exception with `AutoSize = false;` Yeah that would work nicely.

tznind:
@tig I have changed this PR to do as you suggest (Set AutoSize=false instead of throw Exception).

Please let me know if I have misunderstood.

tig (approved):

PR #3415, opened 2024-04-17
Fixes #1785. Adds `Pos.Align` - Align sets of views horizontally or vertically

### LINKED ISSUE #1785: Add `Pos.Align` to ease aligning subviews: `Left`, `Right`, `Center`, `Justify`
Pos/Dim work wonders for layout. However, aligning a set of controls (e.g. in the center) requires quite a bit of annoying code.

E.g. `Dialog` has this code to center buttons:

image

The new `Wizard` has fairly generic code for center, justify, right, and left alignment of buttons.

I would be easy to extract this code such that it would be easy to take a group of controls and horizontally align them.

The same thing could be done vertically. 

e.g. 

```csharp
void LayoutStartedHandler ()
{
  View.AlignControls (Alignments.Justify, arrayOfControls);
}
```

tig:
## Fixes

- Fixes #1785

## Proposed Changes/Todos

- [x] Prototype & experiment
- [x] Build generic/flexible justification engine that will work for both text and view layout
- [x] Implement prototype `Pos.Justify` that automatically justifies views on the same X or Y axis with the same Justification setting
   - This works. See `Generic` scenario.
   - Note Dialog/Message box is broken.
   - It won't work for real because the automatic behavior is too limiting. 
   - [x] Add GroupID - Default is zero and zero gets auto behavior
- [x] Refactor `Justification` class to be an instance class that can hold all the settings, calling `Justification.Justify(...)` static method
- [x] Add `PosJustification` Scenario
- [x] `Justifier` -> Add events
- [x] `Pos.Justify` -> Cache location
- [x] Update `Dialog` (and `MessageBox`) 
- [x] Remove old `TextAlignment` and `VerticalTextAlignment``enums.
- [x] Rename "Justify" back to "Align" 
- [x] Other Pos/Dim -> Cache location/size?
- [x] Unit Tests
- [x] Change `Alignment` and `AlignmentModes` enums to leverage Maui terminology.
- [x] Enable `EndToStart` alignment in `Aligner`
- [x] Demo `EndToStart` in Pos.Align Scenario

## Pull Request checklist:


tig:
9TY52CS 1

tig:
While this does not yet have sufficient unit tests, it is ready for review.

I'm specifically looking for feedback on the API. 

The typical use-case where someone has a set of horizontal subviews they want justified (e.g. `Dialog`'s buttons) is:

```cs
Button button1 = new () { Title = "Button 1", X = Pos.Justify(Justification.Center), Y = Pos.Bottom() };
Button button2 = new () { Title = "Button 2", X = Pos.Justify(Justification.Center), Y = Pos.Bottom() };
Button button3 = new () { Title = "Button 3", X = Pos.Justify(Justification.Center), Y = Pos.Bottom() };
Add (button1, button2, button3);
```

If there are multiple groups:


```cs
const int CHECKBOXGROUP = 1;
Checkbox cb1 = new () { Title = "Check 1", X = Pos.Justify(Justification.Justified, CHECKBOXGROUP), Y = 0};
Checkbox cb2 = new () { Title = "Check 2", X = Pos.Justify(Justification.Justified, CHECKBOXGROUP), Y = 0};
Checkbox cb3 = new () { Title = "Check 3", X = Pos.Justify(Justification.Justified, CHECKBOXGROUP), Y = 0};
Add (cb1, cb2, cb3);

const int BUTTONGROUP = 2;
Button button1 = new () { Title = "Button 1", X = Pos.Justify(Justification.Center, BUTTONGROUP), Y = Pos.Bottom() };
Button button2 = new () { Title = "Button 2", X = Pos.Justify(Justification.Center, BUTTONGROUP), Y = Pos.Bottom() };
Button button3 = new () { Title = "Button 3", X = Pos.Justify(Justification.Center, BUTTONGROUP), Y = Pos.Bottom() };
Add (button1, button2, button3);
```

I'm open for suggestions for other ways to let devs have multiple sets of views that are independently justified other than the somewhat clunky "GroupID" concept. I've tried multiple things, including trying to make it automatic based on the opposite dimension, but what I have now seems to be the cleanest.

BDisp (approved):
LGTM. Great work.

BDisp:
What you might consider is to simultaneously calculate the position of several `GroupID`'s in order to align them horizontally and vertically.

tig:
> What you might consider is to simultaneously calculate the position of several `GroupID`'s in order to align them horizontally and vertically.

I don't understand. Can you elaborate further? What is the use-case?

BDisp:
> I don't understand. Can you elaborate further? What is the use-case?

Imagine you have 3 rows with 3 views on each row and you want justify horizontally and vertically.

tig:
> > I don't understand. Can you elaborate further? What is the use-case?
> 
> Imagine you have 3 rows with 3 views on each row and you want justify horizontally and vertically.

Could you write the example code for this?

tig:
> > > I don't understand. Can you elaborate further? What is the use-case?
> > 
> > 
> > Imagine you have 3 rows with 3 views on each row and you want justify horizontally and vertically.
> 
> Could you write the example code for this?

Nevermind. It already works:

```cs
    private void Setup3by3Grid (Window appWindow)
    {
        var container = new View ()
        {
            Title = "3 by 3",
            BorderStyle = LineStyle.Single,
            X = Pos.AnchorEnd (),
            Y = Pos.AnchorEnd (),
            Width = Dim.Percent (30),
            Height = Dim.Percent (30),
        };

        for (int i = 0; i < 9 ; i++)

        {
            var v = new View ()
            {
                Title = $"{i}",
                BorderStyle = LineStyle.Dashed,
                Height = 3,
                Width = 5
            };

            v.X = Pos.Justify (Justification.Right,  i / 3);
            v.Y = Pos.Justify (Justification.Justified, (i % 3) + 10);

            container.Add (v);
        }


        appWindow.Add (container);
    }
```
oX4Zf55 1

tig:
Man, it would be cool to support `Justification.Wrap` that worked like word wrap.

dodexahedron:
There are so many standard wrapping modes out there it's ridiculous.

The only ones I tend to _dislike_ are the ones where they both justify & stretch, while also still hyphenating words arbitrarily, rather than on reasonable boundaries like syllables (usually just vowel<->consonant), because then it's like...why bother hyphenating it there, you dumb computer?

BDisp:
@tig I'm not seeing horizontal justification there.

tig:
> @tig I'm not seeing horizontal justification there.

Look at the code. For the demo I recorded, I set horizontal justification to center. It also works if horizontal justification is justify.

tig:
@dodexahedron Please help me understand something. This PR shows 215 files changed from v2_develop.

But I've rebased onto v2_develop and, in fact, only a few files are different from v2_develop.

How do I fix a PR like this so it accurately reflects the diff?

tig:
Dialog/MessageBox now use Pos.Justify:

h59XgLe 1

BDisp:
Wouldn't it make sense to change the name of the `Justification` class to `Alignment` because justification is a state of an alignment?

tig:
> `Justification`

Yeah. I intentionally created it as "Justifier" to avoid ambiguity as I was developing it. 

Now that I've removed all old instances of "Alignment" (see latest push), I'm going to rename it back to "align".

tig:
> > `Justification`
> 
> Yeah. I intentionally created it as "Justifier" to avoid ambiguity as I was developing it.
> 
> Now that I've removed all old instances of "Alignment" (see latest push), I'm going to rename it back to "align".

Done.

tig:
@tznind , @dodexahedron , @BDisp - This is now ready for review. I'm eager to get it merged so I can move on.

Please review asap!

BDisp (approved):
Very good work. Thanks.

BDisp:
@tig this flickering is happening with you with the terminal maximized?

WindowsTerminal_ZaKeYhq8nw

dodexahedron:
> @dodexahedron Please help me understand something. This PR shows 215 files changed from v2_develop.
> 
> But I've rebased onto v2_develop and, in fact, only a few files are different from v2_develop.
> 
> How do I fix a PR like this so it accurately reflects the diff?

This sort of thing is usually the result of merging another branch into yours at some point, but I'll take a look at your branch and see if I can figure out what went wrong.

dodexahedron:
Actually...

Before I bother trying to trace history...

That issue was from before the merge strategy was fixed.

Anything branched from a point before that is quite likely to experience fallout from that. There's also a slight chance that, after merging such a branch, one or more following merges - even of branches after the merge strategy was fixed - may also display unexpected diffs.

If you have any other branches from back then, there are ways to fix it, and every one that I'm aware of is pretty tedious.

I'll typically branch from head of what's current and then cherry pick every single commit, in order, from the old branch, skipping any problematic commits along the way (like merges). It's very similar to what an interactive rebase is, but even more manual and controlled.

Terminal.Gui/Drawing/Aligner.cs
```diff
@@ -1,40 +1,75 @@
...
+using static Terminal.Gui.Pos;
+
 namespace Terminal.Gui;
 
 /// <summary>
-///     Controls how the <see cref="Justifier"/> justifies items within a container. 
+///     Controls how the <see cref="Aligner"/> aligns items within a container.
 /// </summary>
-public enum Justification
+public enum Alignment
```
dodexahedron, line 10: Should be in its own file.

Terminal.Gui/Drawing/Aligner.cs
```diff
@@ -1,40 +1,75 @@
...
     ///     </c>
     /// </example>
     Right,
 
+    /// <summary>
+    ///     The items will be aligned to the bottom.
+    ///     Set <see cref="Aligner.PutSpaceBetweenItems"/> to <see langword="true"/> to ensure at least one line between
+    ///     each item.
+    /// </summary>
+    Bottom = Right,
```
dodexahedron, line 60: Why are dissimilar labels sharing values?

There are several reasons every other existing framework uses separate enumerations for vertical and horizontal alignment.

tig:
> > @dodexahedron Please help me understand something. This PR shows 215 files changed from v2_develop.
> > But I've rebased onto v2_develop and, in fact, only a few files are different from v2_develop.
> > How do I fix a PR like this so it accurately reflects the diff?
> 
> This sort of thing is usually the result of merging another branch into yours at some point, but I'll take a look at your branch and see if I can figure out what went wrong.

It's fixed itself.

tig:
> @tig this flickering is happening with you with the terminal maximized?
> 

Nope.

dodexahedron (changes requested):
Before continuing reviewing, since my main comment is non-trivial:

I'm reeaaaallllly not a fan of combining concerns for horizontal and vertical alignment.

It's "weird" and creates complexity.

Every other UI framework out there, including the main two I compare with - HTML/CSS and WPF - uses separate vertical and horizontal alignment properties, each using its own enumerations for the values.

Combining not only bucks that convention, but leads to awkward constructs like `FirstTopRestBottom.`

If you're meaning to control justification strategy with those, that's also appropriate as a separate property with its own enumeration.

Also, every other UI framework out there makes a distinction between alignments of flow and block elements, as well.

TextAlignment really was a perfectly fine and typical enum to exist, but should have been called `HorizontalTextAlignment` since that's what it was.

Mixing these things together _is_ going to have unwanted and confusing results later on, both internally and for consumers.

Even just the enum itself. That's not round-trippable because of shared labels, unless explicit use of a custom serializer is involved. The behavior of enum is undefined for which label will be reported for a value that matches more than one, so you can set it as one label and then ToString gives one of the the other labels sharing its value.

I can't really support this one as-is.

The _behavior_ looks good to me. It's the API that I'm not a fan of.

dodexahedron:
A follow-up thought to that review:

If you wanted to make Alignment and TextAlignment record types, that would also be perfectly fine as a way of bundling alignment-related things together. But the separation of horizontal, vertical, flow/text, and block/View still should be there.

dodexahedron:
Also, a lot of the logic wouldn't really change at all. This is mostly a structural thing.

dodexahedron:
> > > @dodexahedron Please help me understand something. This PR shows 215 files changed from v2_develop.
> > > But I've rebased onto v2_develop and, in fact, only a few files are different from v2_develop.
> > > How do I fix a PR like this so it accurately reflects the diff?
> > 
> > 
> > This sort of thing is usually the result of merging another branch into yours at some point, but I'll take a look at your branch and see if I can figure out what went wrong.
> 
> It's fixed itself.

Woot

dodexahedron:
> @tig this flickering is happening with you with the terminal maximized?
> 
> WindowsTerminal_ZaKeYhq8nw [ WindowsTerminal_ZaKeYhq8nw ](https://private-user-images.githubusercontent.com/13117724/329707156-1fd22942-3ffa-403d-8d8a-4928458efe96.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTUzODI3OTAsIm5iZiI6MTcxNTM4MjQ5MCwicGF0aCI6Ii8xMzExNzcyNC8zMjk3MDcxNTYtMWZkMjI5NDItM2ZmYS00MDNkLThkOGEtNDkyODQ1OGVmZTk2LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA1MTAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNTEwVDIzMDgxMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTE3NGY1NzE0NjY0YTlmZWY3Njc3NTBkYzEwZTIwMjAwNDdjMGY3NjY4ZGFhOGJhNDEzOTViNTgzOGFjZTQyNDgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.FawcMeXxxqnTuyVO2h0w_Jv1Hp82avpSzgIiu2CWt2I) [ ](https://private-user-images.githubusercontent.com/13117724/329707156-1fd22942-3ffa-403d-8d8a-4928458efe96.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTUzODI3OTAsIm5iZiI6MTcxNTM4MjQ5MCwicGF0aCI6Ii8xMzExNzcyNC8zMjk3MDcxNTYtMWZkMjI5NDItM2ZmYS00MDNkLThkOGEtNDkyODQ1OGVmZTk2LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA1MTAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNTEwVDIzMDgxMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTE3NGY1NzE0NjY0YTlmZWY3Njc3NTBkYzEwZTIwMjAwNDdjMGY3NjY4ZGFhOGJhNDEzOTViNTgzOGFjZTQyNDgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.FawcMeXxxqnTuyVO2h0w_Jv1Hp82avpSzgIiu2CWt2I)

I've seen that in various situations at the console in windows. I double it's TG-related, because I've had it while using even just the console itself or other console apps.

I'm betting it's a video driver thing, but I've not really bothered to look into it. 🤷‍♂️

BDisp:
> I've seen that in various situations at the console in windows. I doubt it's TG-related, because I've had it while using even just the console itself or other console apps.
> 
> I'm betting it's a video driver thing, but I've not really bothered to look into it.

I'm pretty sure it's something that is filling in spaces with the current color attribute and displayed on the screen before the cycle has been completed. When the cycle is completed, the buffer is already filled and is properly displayed on the screen. As there is a timer, the screen is always being redesigned and causes this flickering. But this is a very difficult problem to debug.

BDisp:
Although the horizontal and vertical alignment logic is similar, only the orientation is relevant, I think @dodexahedron is right when he says that each orientation must have its own enum's. For example, it is possible to set `VerticalAlignment` to `Alignment.Left`, but it is strange because `Left` refers to horizontal orientation, although I understand that during execution it takes orientation into account.
So my vote would be to keep them separate and change the `enum` and `Alignment` properties to `HorizontalAlignment` for horizontal orientation.

tig:
> Before continuing reviewing, since my main comment is non-trivial:
> 
> I'm reeaaaallllly not a fan of combining concerns for horizontal and vertical alignment.
> 
> It's "weird" and creates complexity.
> 
> Every other UI framework out there, including the main two I compare with - HTML/CSS and WPF - uses separate vertical and horizontal alignment properties, each using its own enumerations for the values.

This is simply not true. Flutter, which is super modern, for example: https://api.flutter.dev/flutter/painting/Alignment-class.html

SwiftUI does too (although it also has distinct structures too): https://developer.apple.com/documentation/swiftui/alignment

In addition, TG's (fairly unique/innovative) computed layout system is fundamentally based on horizontal == vertical. There's not a single line of code in `PosDim.cs` that treats one dimension differently than another. `X` and `Y` are treated identically. `Height` and `Width` are identical. The only difference are the names.

> Combining not only bucks that convention, but leads to awkward constructs like `FirstTopRestBottom.`

It's not weird at all. Look at the code to `Aligner.Align`. It works exactly the same way for h & v. There is no difference. 

Adding another API so that there are separate APIs for h & v would add complexity. `Justify` is only about 50 lines of code. Adding a separate h & v enum, different signatures of `Align` for each, different `Pos.Align` methods, additional constructors to `PosAlign`, etc. would add hundreds of lines of duplicate boilerplate and API documentation. 

I agree `FirstTopRestBottom` is an awkward name, but it's name has nothing to do with the fact that the API combines h & v into a single end-point. The name is a byproduct of me prefering names that clearly say what they do. An alternative I considered was "TopJustify" and "BottomJustify" (and "LeftJustify"/"RightJustify"). but those are not self-explanatory. If you have better names please suggest them.

> 
> If you're meaning to control justification strategy with those, that's also appropriate as a separate property with its own enumeration.
> 
> Also, every other UI framework out there makes a distinction between alignments of flow and block elements, as well.
> 
> TextAlignment really was a perfectly fine and typical enum to exist, but should have been called `HorizontalTextAlignment` since that's what it was.

Having `TextAlignment` and `VerticalTextAlignment` be exactly the same, but with different names was not perfectly fine. First, it was not "designed", it was done this way because vertical support was added later...bolted onto the side. 

> Mixing these things together _is_ going to have unwanted and confusing results later on, both internally and for consumers.

I disagree. It actually makes things simpler. There's one concept of "alignment" and it works for both horizontal and vertical cases, exactly the same way. The only complexity is humans like to use different words when considering left-right vs. up-down. To that end, I used an `enum` with aliases. 

> Even just the enum itself. That's not round-trippable because of shared labels, unless explicit use of a custom serializer is involved. The behavior of enum is undefined for which label will be reported for a value that matches more than one, so you can set it as one label and then ToString gives one of the the other labels sharing its value.

On this, I'm aligned. I don't love the fact that `Alignment` uses aliases. I think the way to address this is to simply make the items that are aliases have distinct values like this:

```cs
public enum Alignment
{
    Left = 0,
    Top = 1,

    Right = 2,
    Bottom = 3,
```

And change the switch statement in `Align()` to

```cs
        switch (alignment)
        {
            case Alignment.Left:
            case Alignment.Top:
...
```

This will make the enum round-trippable etc...

> 
> I can't really justify supporting this one as-is.
> 
> The _behavior_ looks good to me. It's the API that I'm not a fan of.

That's a bummer. I hope you see my perspective and reconsider.

tig:
> Although the horizontal and vertical alignment logic is similar, only the orientation is relevant, 

It is not just "simliar", it is IDENTICAL. 

> I think @dodexahedron is right when he says that each orientation must have its own enum's. For example, it is possible to set `VerticalAlignment` to `Alignment.Left`, but it is strange because `Left` refers to horizontal orientation, although I understand that during execution it takes orientation into account. So my vote would be to keep them separate and change the `enum` and `Alignment` properties to `HorizontalAlignment` for horizontal orientation. But I believe that this change must be very difficult to make at this point when you have already changed everything to `Alignment`. So I advise, if you really accept the suggestion, of course, it is to reset to the time before these changes and just include the relevant commits.

Per my comment above, I strongly disagree with having two enums. By moving to one, I removed a huge amount of complexity (in terms of lines of code, API docs, and logic) form the project. 

I believe my suggestion for modifying `Alignment` to not have aliases, but distinct values will address the key piece of valid feedback: Using aliases makes it very hard to round trip the enum (e.g. in ConfigurationManager).

BDisp:
> It is not just "simliar", it is IDENTICAL.

You are right, that is the most appropriate term.

> Per my comment above, I strongly disagree with having two enums. By moving to one, I removed a huge amount of complexity (in terms of lines of code, API docs, and logic) form the project.

I agree that you disagree and I completely understand your perspective, especially in duplicating code that could be reusable. But I recognize that alignment doesn't have much to do with orientation but rather with the type of alignment we want horizontally and vertically, regardless of the alignment. In this case, I'm starting to agree with you.

> I believe my suggestion for modifying `Alignment` to not have aliases, but distinct values will address the key piece of valid feedback: Using aliases makes it very hard to round trip the enum (e.g. in ConfigurationManager).

As you can see in the image there are two aliases that are never used, at least with their own names. Therefore, your decision to replace aliases with different values will be an added value to the API. Thanks.

image

BDisp:
I have an idea that you might like. Perhaps the `VerticalTextAlignment` property should be removed and just use the `Alignment` property to set the horizontal and vertical alignment. Therefore, by default it will be `Left | Top` for horizontal and vertical alignment. If another is desired, just use `Left | Centered`, or `Centered | Centered`. Unless you want to differentiate horizontal centered and vertical centered separately. Only the `Orientation` enum will define the direction of the text. What do you think?

tig:
> I have an idea that you might like. Perhaps the `VerticalTextAlignment` property should be removed and just use the `Alignment` property to set the horizontal and vertical alignment. Therefore, by default it will be `Left | Top` for horizontal and vertical alignment. If another is desired, just use `Left | Centered`, or `Centered | Centered`. Unless you want to differentiate horizontal centered and vertical centered separately. Only the `Orientation` enum will define the direction of the text. What do you think?

Perhaps, but not for this PR. This should be considered in the great, looming, `TextFormatter` re-write ;-).

tig:
@tznind, @BDisp, @dodexahedron:

Now that #3480 has been merged, I've made final tweaks to this one and it's ready for you to review (again).

7PZOpll 1

Terminal.Gui/Drawing/Aligner.cs
```diff
@@ -328,7 +328,7 @@ int spacesToGive
         return positions;
     }
 
-    private static void CheckSizeCannotBeNegative (int i, IReadOnlyList<int> sizes)
+    private static void CheckSizeCannotBeNegative (int i, ref readonly int [] sizes)
```
dodexahedron, line 362: These are very different things.

Aside from the other comment, the meaning of these two things are:

 - `IReadOnlyList<int>`: Any type implementing the interface, with a guarantee that, aside from unsafe memory shenanigans, the contents cannot and will not be altered and that, in the callee, the original reference can't be altered because the reference itself was passed by value, because that's how C# works.
 - `ref readonly int []`: Can only be a `System.Array<int>`, passed in by read-only reference. How's this different?
   - The parameter is an actual reference to the source array (if the compiler doesn't make a defensive copy - which it shouldn't here).
   - Normal `ref` would have allowed reassignment of the reference, but `readonly` prohibits that (`in` achieves the same in this case and is also callsite compatible, as you found out).
   - It MUST be (a reference to) an `int[]` and no other type is allowed unless it directly inherits from `System.Array<int>`, which isn't possible in human-generated code in C# - and not only because `Array<T>` isn't a "real" type you can access.
   - Typically, you'd only bother ref passing a ref type for certain interop scenarios or when you really do want the callee to be able to reassign the caller's reference to that item, which is perfectly legal and valid, but typically a very specific micro-optimization for very specific situations when in-lining will probably happen anyway.

The original was honestly probably a better type for the method argument. Interfaces are great for these sorts of things.

Terminal.Gui/Drawing/Aligner.cs
```diff
@@ -152,7 +152,7 @@ public int ContainerSize
 
         for (var i = 0; i < sizes.Length; i++)
         {
-            CheckSizeCannotBeNegative (i, sizes);
+            CheckSizeCannotBeNegative (i, in sizes);
```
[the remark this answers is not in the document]
dodexahedron: Strike that `IEnumerable<T>` comment, since the original callee had `IReadOnlyList<T>`, which is probably better - if readonly semantics are mandatory.

Terminal.Gui/Drawing/Aligner.cs
tig: Thanks.

Issue #1316, opened 2021-05-20
UnitTests: Need unit tests that test the real drivers


tig:
We have zero code coverage for some of the most complex pieces of `Terminal.Gui`: The implementations of `ConsoleDriver`.

![image](https://user-images.githubusercontent.com/585482/119006807-8a5a7c80-b945-11eb-9d71-09ef51c39bba.png)

It should be possible to write unit tests that work on Windows for the `WindowsDriver` and `NetDriver` and some that work on Linux for `CursesDriver` and `NetDriver`. Obviously the `CursesDriver` tests won't run on Windows, and the `LinuxDriver` tests won't run on Windows.  

However, we can change the CI/CD build action to run on BOTH Windows and Linux ensuring that these tests do all run when a new pull request is submitted.

BDisp:
I already tried to do this, but without a real terminal, it isn't possible. I would like to do this if there is a way.

tig:
#2799 helped with this.

more needed though.

tig:
#2932 makes even more progress on this.

Now that `Application` has been decoupled from `ConsoleDriver` (almost completely) and `TestHelpers` now supports passing a driver to `VerifyContents...` all tests in `DriverTests` should be upgraded to use `[Theory]` and be decoupled from `Application`... doing so will massively address this issue:

tig:
Enough progress has been made on this with #4376 that I'm closing it. We can continue to refine.

PR #3416, opened 2024-04-17
Partially Fixes #2432 - `Dim.Auto` automatically sizes views based on `Text` - Replaces `View.AutoSize`

### LINKED ISSUE #2432: Finish `Dim.Auto` - `DimAutoStyle.Content` is half-baked
This feature has been partially implemented. `DimAutoStyle.Text` works great and fully replaced `View.AutoSize`. However, `DimAutoStyle.Computed` is still a WIP. We'll close this issue once it's fully implemented and tested.

## Background

See #2431 for some context.

Multiple times I've wanted to have a view dynamically be sized based on it's subviews. Also we have `AutoSize` which sorta does this for `Text`. 

I'd like to propose the following for v2:

- A new feature for ComputedLayout that adds `Dim.Auto ()`
- Replace `AutoSize` with something that builds on `Dim.Auto ()`

I think this should be possible, but haven't thought THAT much about it.

```cs
view.Height = Dim.Auto ();  // Height will be Margin.Height+Padding.Height+Border.Height+(height required for all Subviews to fit).
```

tig:
(this is a new PR as I goofed up the other one)

## Fixes

- Partially Fixes #2432:
  - Fully implements `DimAutoStyle.Text`
  - Replaces `View.AutoSize` (and removes it)
  - Does NOT fully implement `DimAutoStyle.Subviews` (will be a future PR).

## Adds `Dim.Auto` to Computed Layout to automatically size views based on `Text`

- [x] Initial WIP and Scenario to demonstrate concept
- [x] WIP: Support auto-size based on subviews
- [x] Figure out how to make `CheckDimAuto` deal with PosCombine
- [x] Re-investigate the concept of `Dim.Auto (max: n)` and/or `Dim.Auto (max: dim)` 
  - Without this, `AnchorEnd` doesn't work, which is really sad. 
- [x] Merge the fix to 
  - [x] https://github.com/gui-cs/Terminal.Gui/issues/3192
- [x] Support auto-size based on `Text`
- [x] Implemement `Dim.Auto (min: dim)`
- [x] `DimAutoStyle.Text` works 
  - [x] `View.AutoSize` now just uses `Dim.Auto`.
  - [x] Ensure vertical text works
  - [x] `ObsoleteAttribute` addeed
  - [x] No internal `View` base class code uses `AutoSize` 
- [x] Remove `View.AutoSize` and update all code to use `Dim.AutoSize` instead
- [x] Enough Unit tests
- [x] API docs

## For Future PRs

- [ ] Fully implement/test `DimAutoStyle.Subviews`
- [ ] Implemement `Dim.AutoSize (max: dim)`
- [ ] Use `Dialog` and `MessageBox` as test cases - This will require moving the dialog buttons to `Padding` as adornments.
- [ ] Dive deep into all other PosDim objects and get clarity on behavior relative to `Dim.Auto` - Improve unit tests to validate
  - [x] `Dim.Absolute`
  - [ ] `Dim.Auto`
  - [ ] `Dim.Fill`
  - [ ] `Dim.Function`
  - [ ] `Dim.Width`/`Height`
  - [x] `Pos.Absolute`/`At`
  - [ ] `Pos.AnchorEnd`
  - [ ] `Pos.Function`
  - [ ] `Pos.Anchor`
  - [ ] `Pos.Percent`
  - [ ] `Pos.Center`
  - [ ] `Pos.X`/`Y` etc...

## Pull Request checklist:


tig:
Team, 

This PR is now ready for first review. It is not ready for merge, but you should now start getting your eyes on it and consider basing any new work on it.

I will NOT be completing `Dim.Auto (DimAutoStyle.Subviews)` in this PR. Instead this PR will

- Completely replace `View.AutoSize`
- Partially support (likely with bugs) `DimAutoStyle.Subviews`

@BDisp if you want, please fix the additional vertical text issues you say you know how to fix in TextFormatter is a PR to this PR.

BDisp:
Another situation that I am seeing is that the inversion of the horizontal and vertical direction of the text will have to be processed during formatting and not during redrawing on the screen. See the example in the image where the top and bottom are swapped.

image

BDisp:
This is great. @tig do you know why the title isn't updated when the labels grown?

image

tig:
> This is great. @tig do you know why the title isn't updated when the labels grown?
> 

Nope. Seems to be a bug.

tig:
> > This is great. @tig do you know why the title isn't updated when the labels grown?
> 
> Nope. Seems to be a bug.

it has somthing to do with Text being empty during init:

image

tig:
All, as you review this, please read and comment on the new `DimAuto` Deep Dive:

https://github.com/tig/Terminal.Gui/blob/v2_2432-DimAuto/docfx/docs/dimauto.md

docfx/docs/dimauto.md
```diff
@@ -0,0 +1,79 @@
...
+{
+    Orientation = Orientation.Vertical,
+    Type = SliderType.Multiple,
+};
+view.Add (slider);
+```
+
+Note the developer does not need to specify the size of the `Slider`, it will size itself based on the number of options and the orientation. 
+
+Views like `Slider` do this by setting `Width` and `Height` to `Dim.Auto (DimAutoStyle.Content)` in the constructor and and setting the `ContentSize` property to the size that the view should be in the `LayoutStarted` event handler.
```
BDisp, line 100: Typo and and twice.

BDisp (approved):
LGTM. Excellent.

tig:
OK!

`View.AutoSize` is now completely removed. 

I'm getting close to merging this.

@dodexahedron & @tznind - I would LOVE your eyes on this!

BDisp:
I'm preparing the following and doing the unit tests for all text direction variants. I hope you like this.

WindowsTerminal_bgi81AQuRR

BDisp:
> @BDisp I can't figure out what I merged incorrectly, but the cursor is now showing by default on all views.
> 
> Can you tell where I screwed up?

It seems that you already fix it.

tig:
> > @BDisp I can't figure out what I merged incorrectly, but the cursor is now showing by default on all views.
> > Can you tell where I screwed up?
> 
> It seems that you already fix it.

Yep. I think.

What do you think of this new API doc for `PositionCursor`?

```cs
    /// <summary>
    ///     Positions the cursor in the right position based on the currently focused view in the chain.
    /// </summary>
    /// <remarks>
    ///     <para>
    ///         Views that are focusable and want the cursor visible should override <see cref="PositionCursor"/>,
    ///         use <see cref="Move"/> to place the cursor in a location that makes sense, use
    ///         <see cref="ConsoleDriver.SetCursorVisibility"/>
    ///         to make the cursor visible, and return the position where the cursor was placed.
    ///     </para>
    ///     <para>
    ///         Unix terminals do not have  a way of hiding the cursor, so it can be distracting to have the cursor left at
    ///         the last focused view. Views should make sure that they place the cursor in a visually sensible place.
    ///     </para>
    /// </remarks>
    /// <returns>Viewport-relative cursor position. Return <see langword="null"/> to ensure the cursor is not visible.</returns>
    public virtual Point? PositionCursor ()
    {
        if (IsInitialized && CanFocus && HasFocus && ContentSize.HasValue)
        {
            // Base class will position the cursor at the end of the text.
            Point location = Viewport.Location;
            location.X = TextFormatter.HotKeyPos == -1 ? 0 : TextFormatter.CursorPosition;
            location.Y = 0;
            Move (location.X, location.Y);
        }

        // Returning null will hide the cursor.
        return null;
    }
```

BDisp:
```
    ///     <para>
    ///         Unix terminals do not have  a way of hiding the cursor, so it can be distracting to have the cursor left at
    ///         the last focused view. Views should make sure that they place the cursor in a visually sensible place.
    ///     </para>
```

I don't quite understand this. In Unix it is also possible to hide the cursor through ANSI escape sequence. In WSL with version v2 the cursor only blinks if we move the mouse. A view that wants to show the cursor must also call the `SetCursorVisibility` method to define the cursor type.

BDisp:
@tig this change is enough in `CursesDriver`.

```cs
    public override void UpdateCursor ()
    {
        EnsureCursorVisibility ();

        if (!RunningUnitTests && Col >= 0 && Col < Cols && Row >= 0 && Row < Rows)
        {
            Curses.move (Row, Col);
        }

        Curses.raw ();
        Curses.noecho ();
        Curses.refresh ();
    }

```

BDisp:
The only difference is that the `WindowsDriver` immediately assumes the type of cursor that is configured in the WT and the `CursesDriver` initially uses the type of underline cursor in the `TextField` and after selecting the `TextView` the application assumes the cursor configured in the WT.

WindowsTerminal_RKmy2CuW5G

tig:
> ```
>     ///     <para>
>     ///         Unix terminals do not have  a way of hiding the cursor, so it can be distracting to have the cursor left at
>     ///         the last focused view. Views should make sure that they place the cursor in a visually sensible place.
>     ///     </para>
> ```
> 
> I don't quite understand this. In Unix it is also possible to hide the cursor through ANSI escape sequence. In WSL with version v2 the cursor only blinks if we move the mouse. A view that wants to show the cursor must also call the `SetCursorVisibility` method to define the cursor type.

That comment has been there since Miguel wrote the code. It must have been true at some point.

dodexahedron:
For that `PositionCursor()` method:

Do note that use of `Point?` means construction of a Nullable<Point> wrapper, which means a copy will be made of the Point, plus the cost of the rest of that struct, even though you don't see it, _plus_ the cost of null being construction of a `Nullable<Point>` anyway, just with HasValue == false.

To avoid the cost of a wrapper, implementing the TryX pattern is a good option (whether named with Try or not), and then make the argument either ref or out, but not nullable.

dodexahedron:
And, as with all virtuals, be sure to consider what any internal usages of it may mean if it is overridden by a descendent type, as the most-derived is used no matter if it's called from this type or descendants, with the exception of calls to `base.SomeVirtual` of course.

Code in the base declaring type can't make assumptions based on its implementation of a virtual.

Well... I mean... It _can_...  But any such assumptions need to be loudly documented on the virtual as well as whatever internally uses it that is making those assumptions.

dodexahedron:
This part is also a lot of copying:
```csharp
Point location = Viewport.Location;
location.X = TextFormatter.HotKeyPos == -1 ? 0 : TextFormatter.CursorPosition;
location.Y = 0;
```

Better to use a single mutation like:
```csharp
Point location = Viewport.Location with { X = TextFormatter.HotKeyPos == -1 ? 0 : TextFormatter.CursorPosition, Y = 0 };
```

or, what the compiler will probably tell you to do instead, since all properties are being set:
```csharp
Point location = new (TextFormatter.HotKeyPos == -1 ? 0 : TextFormatter.CursorPosition, 0);
```

dodexahedron:
Also, with Y unconditionally set to 0, is a Point struct really even needed for that method?

It'll get created, then copied again for the call to Move, and then whatever happens in Move as well after that.

And that copy is avoidable, too, even if still using a point, by making it an out or ref parameter and having the caller supply the reference.

dodexahedron:
> @tig this change is enough in `CursesDriver`.
> 
> ```cs
>     public override void UpdateCursor ()
>     {
>         EnsureCursorVisibility ();
> 
>         if (!RunningUnitTests && Col >= 0 && Col < Cols && Row >= 0 && Row < Rows)
>         {
>             Curses.move (Row, Col);
>             Curses.raw ();
>             Curses.noecho ();
>             Curses.refresh ();
>         }
>     }
> ```

I'd strongly suggest not ever having anything unit test related in compiled output in any situation other than unit tests.

For this, it would mean defining a compiler constant in the unit test project (that's useful for other things anyway), and then using a `#if` to skip the check if defined, rather than a runtime check on a boolean that shouldn't even be in consuming code and which could potentially interact with a consumer's unit test projects, too.

BDisp:
Now it's working as expected. The views that wants to show the cursor must explicitly to call the `SetCursorVisibility` method. One solution is creating a property `WantShowCursorPosition` in the `View` class and letting it to manage it.
I think you should remove from the  XML document about the Unix not supporting hiding the cursor.

WindowsTerminal_5i7Hsdis8l

tig:
> For that `PositionCursor()` method:
> 
> Do note that use of `Point?` means construction of a `Nullable<Point>` wrapper, which means a copy will be made of the Point, plus the cost of the rest of that struct, even though you don't see it, _plus_ the cost of null being construction of a `Nullable<Point>` anyway, just with HasValue == false.
> 
> To avoid the cost of a wrapper, implementing the TryX pattern is a good option (whether named with Try or not), and then make the argument either ref or out, but not nullable.

FWIW, this is all temporary - This all needs a serious refactor per https://github.com/gui-cs/Terminal.Gui/issues/3444

tig:
> > @tig this change is enough in `CursesDriver`.
> > ```cs
> >     public override void UpdateCursor ()
> >     {
> >         EnsureCursorVisibility ();
> > 
> >         if (!RunningUnitTests && Col >= 0 && Col < Cols && Row >= 0 && Row < Rows)
> >         {
> >             Curses.move (Row, Col);
> >             Curses.raw ();
> >             Curses.noecho ();
> >             Curses.refresh ();
> >         }
> >     }
> > ```
> 
> I'd strongly suggest not ever having anything unit test related in compiled output in any situation other than unit tests.
> 
> For this, it would mean defining a compiler constant in the unit test project (that's useful for other things anyway), and then using a `#if` to skip the check if defined, rather than a runtime check on a boolean that shouldn't even be in consuming code and which could potentially interact with a consumer's unit test projects, too.
> 
> That also means there should almost definitely be an interface defined for this method instead of or in addition to how it is currently declared.

I agree. However, until these are fully addressed, we don't have a ton of choice here:

- https://github.com/gui-cs/Terminal.Gui/issues/1316
- https://github.com/gui-cs/Terminal.Gui/issues/2319

tig:
> Now it's working as expected. The views that wants to show the cursor must explicitly to call the `SetCursorVisibility` method. One solution is creating a property `WantShowCursorPosition` in the `View` class and letting it to manage it. I think you should remove from the XML document about the Unix not supporting hiding the cursor.
> 
> WindowsTerminal_5i7Hsdis8l [ WindowsTerminal_5i7Hsdis8l ](https://private-user-images.githubusercontent.com/13117724/328516067-c387e0b6-9859-4b32-b412-4bf890f82851.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTUwODgzOTEsIm5iZiI6MTcxNTA4ODA5MSwicGF0aCI6Ii8xMzExNzcyNC8zMjg1MTYwNjctYzM4N2UwYjYtOTg1OS00YjMyLWI0MTItNGJmODkwZjgyODUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA1MDclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNTA3VDEzMjEzMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTgwMzQ4OGI2NzhkNzRlZDI0OTI0NzY1NDIzYzU3MGFhNzRhMjY3MjQwNjdiMWQwOWUzNDkxODllN2QzY2E2NDUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.P5y5-oubuN7kpEWUcQv4hcx7M1mFx9EQIXenCeCAoc0) [ ](https://private-user-images.githubusercontent.com/13117724/328516067-c387e0b6-9859-4b32-b412-4bf890f82851.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTUwODgzOTEsIm5iZiI6MTcxNTA4ODA5MSwicGF0aCI6Ii8xMzExNzcyNC8zMjg1MTYwNjctYzM4N2UwYjYtOTg1OS00YjMyLWI0MTItNGJmODkwZjgyODUxLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA1MDclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNTA3VDEzMjEzMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTgwMzQ4OGI2NzhkNzRlZDI0OTI0NzY1NDIzYzU3MGFhNzRhMjY3MjQwNjdiMWQwOWUzNDkxODllN2QzY2E2NDUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.P5y5-oubuN7kpEWUcQv4hcx7M1mFx9EQIXenCeCAoc0)

I am currently working on finalizing this, in a temporary way, in this PR. 

I am taking a short-term approach that partially addresses this:

- https://github.com/gui-cs/Terminal.Gui/issues/3444

Specifically:

- Default cursor is invisible. For a view to enable the cursor it must do two things:
  - Set `View.CursorStyle` (`public CursorVisibility CursorStyle {get;set} = CursorVisibility.Invisible`) to `Default` (or something else).
  - Return something other than `null` from `PositionCursor`

I've found some more serious issues in how `Focused` and `MostFocused` work along the way. These will need to be addrssed in #3444 too.

@dodexahedron You once suggested a way to annotate unit tests that currently prove bad behavior so they can still run/pass but be identified as needing to be fixed later. Can you remind me of the technique? Thanks.

tig:
@BDisp, regarding the cursor stuff in this PR:

- It all seems to work now
- Flickring is gone in ALL DRIVERS (including netdriver!).

PR #3422, opened 2024-04-20
Fixes #3413. Improve redraw performance by making Clip be a Region instead of Rectangle

### LINKED ISSUE #3413: Improve redraw performance by making `Clip` be a `Region` instead of `Rectangle`
TUI apps need to be performant under highly-constrained conditions, such as over SSH. Currently the Terminal.Gui system for drawing tries to optimize/reduce redraws via `ConsoleDriver.Clip` which is a rectangle. This works ok but a system based on an irregular `Region` could be far more efficient.

Ages ago I experimented with re-implementing `Clip` using a list of rectangles in this PR:

- https://github.com/gui-cs/Terminal.Gui/pull/2606

Note I somehow overwrote the goodness in that PR, but much of the work is in older commits, but the essence was:

```cs
		/// <summary>
		/// Gets or sets the clip rectangle that <see cref="AddRune(Rune)"/> and <see cref="AddStr(ustring)"/> are 
		/// subject to. Setting this property is equivalent to calling <see cref="ClearClipRegion()"/>
		/// and <see cref="AddToClipRegion(Rect)"/>.
		/// </summary>
		/// <value>The rectangle describing the bounds of <see cref="ClipRegion"/>.</value>
		public Rect Clip {
			get {
				if (ClipRegion.Count == 0) {
					return new Rect (0, 0, Cols, Rows);
				}

				int minX = ClipRegion.Min (rect => rect.X);
				int minY = ClipRegion.Min (rect => rect.Y);
				int maxX = ClipRegion.Max (rect => rect.X + rect.Width);
				int maxY = ClipRegion.Max (rect => rect.Y + rect.Height);

				return new Rect (minX, minY, maxX - minX, maxY - minY);
			}
			set {
				ClearClipRegion ();
				AddToClipRegion (value);
			}
		}

		List<Rect> _clipRegion = new List<Rect> ();
```

I abandoned that PR because, at the time, `ConsoleDriver` and all the implementations were too fragile with way too much duplicated code. I also realized my "list of rectangles" idea was too naive and I was re-inventing the wheel: `System.Drawing.Region` already exists. 

At some point, tackling this PR will be worthy. 

If we can't get to it before releasing v2, at the least we should ensure the current `View.SetClip` API can be forward-compatible so fixing this Issue is not a breaking change to v2.

## Other Discussions

> I don't think an array of Rectangles is the right approach. I think we want to build the next-gen clipping support on a `Region` class that is either directly this:
> 
> https://learn.microsoft.com/en-us/dotnet/api/system.drawing.region?view=dotnet-plat-ext-8.0
> 
> Or, is a clone of it.
> 
> So, unless you built your experiment on `Region`, no.

By chance I noticed the `Region` class but I didn't use it to avoid having to add another library. I created a `RectangleExtensions` class to manipulate its Rectangle methods with the array. But since it didn't solve the problem with flickering and I didn't use the `Region` class, I'm going to give up on continuing with my PR.

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/3408#issuecomment-2061357235_

BDisp:
## Fixes

- Fixes #3413

## Proposed Changes/Todos

- [x] Create Region class
- [x] Replace Clip with HashSet{Region}
- [x] Create RegionData

## Pull Request checklist:


Terminal.Gui/Drawing/Region.cs
```diff
@@ -73,8 +77,10 @@ public override bool Equals (object? obj)
     /// </summary>
     /// <param name="regions"></param>
     /// <returns></returns>
-    public static dynamic Union (params Region [] regions)
+    public static dynamic? Union (params Region [] regions)
```
dodexahedron, line 80: Please do not use `dynamic` for anything, anywhere.

It's highly reflection-dependent and has a ***massive*** runtime cost that is almost certainly a lot higher than what this is trying to achieve in the first place.

dodexahedron (changes requested):

Terminal.Gui/Drawing/Region.cs
dodexahedron: params argument arrays also force boxing.

This isn't appropriate for the goal of this.

BDisp: > Please do not use `dynamic` for anything, anywhere.
> 
> It's highly reflection-dependent and has a _**massive**_ runtime cost that is almost certainly a lot higher than what this is trying to achieve in the first place.

The use of `dynamic` was massively considered because I wanted something that could only return a nullable of `Rectangle` and `RectangleF`, reusing code without duplication. I still thought about using generic type (`T`) but that would imply a big change in the API to change to `Region<Rectangle>`, `Region<RectangleF>`. Could the methods I created return null, if applicable, and if I used a `Rectangle?` nullable or `RectangleF`? return, it would also imply a major change to the entire code. Therefore, I chose to use `dynamic` because it can return any nullable of any of the two types I mentioned, `Rectangle`, `RectangleF`, without major code changes and without code duplication. Therefore, in my opinion it will be worth the sacrifice.

BDisp: > params argument arrays also force boxing.
> 
> This isn't appropriate for the goal of this.

You're right and it doesn't even make sense to use `params` because the parameter is already an array, so I did something wrong and will remove it. Thanks.

BDisp:
I think you can already have an idea of how this will work. I think it's worth a review.

dodexahedron:
Should be able to take a good look at it tomorrow or, if I'm not totally exhausted tonight, later tonight.

Excited to see it. 👍

dodexahedron:
Okie dokie, I'm gonna have a look at this finally. 🥳

dodexahedron:
This will take a while, BTW, because I'm benchmarking and tracing things, since the whole point is performance. :)

dodexahedron:
A couple of initial benchmarks are showing slower performance than before. I'm pretty sure I see where that's coming from, but I'll trace just to be sure.

I branched from your branch, and I'll send any benchmark methods and whatnot as a PR for you when I'm done to keep the noise down over here but still keep full revision history.

dodexahedron:
So far, it's mostly what I suspected before. But that RegionData class is also lighting things up a lot, too.

The good news is I don't think it's going to be terribly hard to fix.

I need to finish analyzing things though, to be sure I have the whole picture.

But I also have a question to help inform my analysis:

What is the purpose of the desire to have both Rectangle and RectangleF support?

dodexahedron (changes requested):
There are a lot of issues with this, many of which are pretty severe in terms of both CPU and memory, including a nearly-unavoidable memory leak that will start happening at the first gen2 garbage collection in the application.

Benchmarks and profiling of some very simple uses had slower performance than the original by 3x-40x early in the execution, and increased on what looks like a small-order exponential or power curve to as much as 800x after running for 10 minutes, adding and closing a simple view with a couple of buttons on a 5 second timer to simulate someone changing screens.

I started writing up specifics from my notes, but I'll finish that up tomorrow because I want to sleep on it and look at it with fresh eyes.

Right now, though, this isn't prudent to merge.

BDisp:
If you have improvement changes please submit a PR to this branch. Thanks.

dodexahedron:
> If you have improvement changes please submit a PR to this branch. Thanks.

Will do. :)

I have to push it to my branch of course since I can't write to yours. But you can merge that into yours when that's done, so it's pretty much the same thing.

To be clear, I love the intent of this, I love the effort, and I love the idea of unifying some of the stuff that it unifies, providing a simpler and more consistent API across the library, both internally and for consumers. All are ***great*** goals.

Unfortunately, because of how the C# language, Roslyn, RyuJIT, and hardware work, those goals have to be achieved via some slightly different methods. In particular, since 100% of this is reliant on reflection, it would make the entire Terminal.Gui library unable to be used with AoT or anything else involving trimming, which is becoming a more and more relevant concern with the way things are heading in .net.

And some of that is just because that's how some of these features (in .net itself, I mean) organically grew up, and are now stuck that way at least for the foreseeable future, because backward-compatibility is generally priority 1 for Microsoft.

`dynamic` is, unfortunately, one of those, in the vast majority of situations. It has some pretty high costs even beyond it being reflection-dependent in most cases. One of the nastiest and completely non-obvious effects is that `dynamic` has higher compile-time precedence for method and type resolution than even `object` does, meaning that the presence of `dynamic` in a code path, even if there's a generic on top of it with a fairly strict type filter[^exceptvaluetypes], will win because the compiler can't know ahead of time that an object provided to a `dynamic` symbol isn't actually an EXACT match for how you're about to use it, nor can it generate any of that ahead of time, in most situations, especially in externally visible code or code that may be called as the result of any externally-visible code path (so never, for us). It always has to be resolved at run-time, and IL code has to be generated and emitted _on the fly_. It also can't innately prevent someone from trying to feed it an object of an illegal type, which then means you have to filter it out yourself. But I'll put comments in-line to explain, in context.

Aside from `dynamic`, some of the other _explicit_ uses of reflection here, such as type checking, point out some things that I think will be helpful to use as a sort of learning example, so one of my intentions (since it won't really take much time because the _logic_ is the same) is to first just do a quick fix of those to do things how they "should" do them, just without the reflection, whether that exact "fixed" code gets used or not in the end, because it's a useful concept in a lot of places, even where reflection or even generics aren't anywhere to be seen, that you most likely will end up using in almost every program you ever write. 🙂

Plus, I like making things efficient and I like teaching/explaining/whatever, so hey - a chance to do two things I like doing at once. 😅 

[^exceptvaluetypes]: The main exception is that a generic type parameter constraint of `struct` or certain other special-case types that inherit from ValueType (which would be illegal for us to write, explicitly), or things like `Enum`. In a very select subset of cases and with a full guarantee of a closed model (which being a public type pretty much precludes), the compiler may be able to generate a slightly better than bad code path for those, while still exposing them as `dynamic` at design time, if it doesn't have to export the symbol. But you should generally consider `dynamic` to be the worst of all worlds - even worse than `object` or an unconstrained generic type parameter.

dodexahedron:
> > What is the purpose of the desire to have both Rectangle and RectangleF support, when fractional pixels aren't really a thing in a console buffer?
> 
> Because the original use them and the graph view too.

I may end up addressing that while fixing this, then, if I get the motivation to do so. Though those can also just be converted between each other a lot cheaper than using reflection or the Dynamic Language Runtime.

dodexahedron:
OK, so here's how this looks so far.

I'm batching sets of changes into individual branches, each of which just continues from the end of the previous one.

There's a little bit of housekeeping like running R# rules and such, at the beginning of each branch.

Then, I'm basically doing pairs of commits, where each pair is, in this order:
 1. Just code comments, in-line, saying what we're about to do and why
 2. Actually performing the specific changes mentioned in the first one and removing the comments

So, each pair is basically as if only the second one was ever done, in the end, but the process and explanations are in chronological order, so they're not scattered around the code.

I didn't start off doing it like that, so the first couple of changes don't have the first part with the comments about what's going to happen. But the commit comments say what they did, and they're very small.

I pushed the first branch a few minutes ago, visible [here](https://github.com/BDisp/Terminal.Gui/compare/v2_region-regiondata-feature_3413...dodexahedron:Terminal.Gui:v2_3467_region_performance_pass_1_replacing_reflection_and_dynamic)

There's also a little bit of stuff at the very beginning that will get used later, for benchmarking and whatnot.

The first commit that matters for Region is 2ea0895ad734d611faddec2d4c33de6a5609b34c , which gets rid of IDisposable, which isn't needed for that type. Just follow along each commit from there and you'll see the process.

If you have any questions or anything, just let me know. :)

dodexahedron:
Wrll this sucks. My internet connection just went out, so I won't be pushing any commits I'm making right now until later.

But this one is starting off with a ton of explanation thats going to matter for the rest and for a lot of things in general outside of this, so I'm not sure if I want to leave that as code comments or post it somewhere easier to read (probably that).

It'll take a while to finish writing anyway, so maybe the cable will be back by then. 🙏

dodexahedron:
Aaannnnd sorry for typos. Phone... 😬

dodexahedron:
Alrighty, internet is back and I've written.... a Lot...

There are a ton of concepts relevant to the next bit.

But it's a lot.

So, what I'm going to do is save all of that and skip a lot of the definitions and such.

Please please please, if anything isn't clear or you want more detail, ask. I've got hundreds of lines written already, so there's a pretty good chance I have it already written.

I'll be making some comment-only commits as soon as I take out the deep dives, in a few minutes.

dodexahedron:
Hang on with that, since I'm modifying those same files.

Commit with big comment blocks coming in a sec.

dodexahedron:
The next branch is here:

https://github.com/dodexahedron/Terminal.Gui/tree/v2_3467_region_performance_pass_1_replacing_reflection_and_dynamic_2

It already has 2 commits for introduction to what's going on next and why, which continue from the end of the previous branch.

Next up will go back to the alternating comment and change commits like before.

BDisp:
It would be easier for me if you could send the link comparing your branch with mine, if you please. Thanks.

dodexahedron:
That's not possible in this format, because the comments are chronologically ordered and change with each step.

dodexahedron:
What I'll do, though, is submit a PR to your fork, which will update with commits as I add them, which should be easier for you to navigate.

Just don't merge it til I'm finished. :)

dodexahedron:
I need to go find my power adapter before my computer battery dies, so gimme a few minutes on that. XD

dodexahedron:
I need to rebase this, which is going to take some time, because the original branch it was based on no longer exists (re-base destroys one and makes a new one).

I have an idea that might avoid me having to do that, but if the first attempt doesn't work, I'm just going to rebase on what you have there now.

But I will not be able to do those PRs if rebases happen underneath the work tree branches. It's sorta like if the v2_develop branch in gui-cs were to suddenly be gone. The commits are all there, somewhere, but how do you merge? 😅 

So, if you could keep the current v2_region-regiondata-feature_3413 branch in your fork based on where it is right now, until this is done, it'll help us both.

If you want to keep working in parallel, you can branch from the end of that branch.

And, once I submit and resolve conflicts and all that, you'll be able to merge the final results of each branch easily without having to try to untangle anything, which I'm sure you'll appreciate. :)

BDisp:
I understand but I'm also correcting and improving the code. The most important thing is that you keep what you think should be changed and then submit a PR for my branch, just like you did when comparing the branches but submitting a PR so it's easier for me to compare the differences. If there's something I don't agree with, I'll comment on your PR, ok? Thanks.

dodexahedron:
As long as you don't rebase the existing branch again, that'll be perfect.

That's the only thing that needs to exist. so that it all works as expected.

I'm almost done with rebases, but I'm going to need to look through the commits to make sure the automatic resolution didn't mess anything up.

Just a few more minutes.

dodexahedron:
Yikes.
Visual studio did not enjoy that. And I'm not feeling like untangling it right now. 😅 

Alrighty then..

So, I'm just going to hang on to those commits and extract the text into a discussion or something at a later time, instead, and just for sake of getting it done make the fixes and submit. It's not actually that many code changes. They're just important ones.

dodexahedron:
And I'll add benchmarking in a separate PR here since I'm separating that anyway.

That'll make this even smaller and easier to compare.

dodexahedron:
That 311eaf3 commit is headed the right direction :)

BDisp:
That's all for today, now I'm going to rest. Thanks for the guidance.

dodexahedron:
Hey one quick question if you see it before you go:

What is the RegionData class actually supposed to contain?

It isn't clear to me what it is.

BDisp:
> Hey one quick question if you see it before you go:
> 
> What is the RegionData class actually supposed to contain?
> 
> It isn't clear to me what it is.

It contains the data on the region area, which will get all the `Rune` from the `ConsoleDriver.Contents` and can be used on another region. See the unit test where there is use case on the `ConsoleDriverTests.cs`.

dodexahedron:
So it's essentially the ScreenBuffer, but only the portion that actually might have a rune at each coordinate?

dodexahedron:
Ok, that helps.

I just needed to get a better idea of the intent behind it, because it is very expensive, as well.

But valuable functions are valuable functions, so I'm sure it can be addressed in a way that achieves the desired functionality at a lower cost.

Have a good night!

tig:
> So it's essentially the ScreenBuffer, but only the portion that actually might have a rune at each coordinate?

I'm not a fan of this. Let's keep the idea of adding a screen buffer separate from the clip region. 

I have ideas around extending LineCanvas to bring a deferred-draw general purpose canvas. What you are proposing here will conflict with that.

BDisp:
No @tig. U already thinking of using Line canvas too, as GraphicsPath. If you want contribute I appreciate, but please don't kill this before is born.

BDisp:
Furthermore, there is no conflict. It is only necessary to create a new `Region(LineCanvas lineCanvas)` constructor.
I was also intending to create an `Xor` method to create an array of regions that do not intersect with the other regions and only draw outside the original arrays defined.

tig:
> No @tig. I already thinking of using LineCanvas too, as GraphicsPath. If you want contribute I appreciate, but please don't kill this before is born.

Write a spec first then. I'll be happy to debate it with you. 

What I don't want to do is have you write a bunch of complex code and then ask me to understand it and give feedback. 

Measure twice, cut once. Please.

tig:
> Furthermore, there is no conflict. It is only necessary to create a new `Region(LineCanvas lineCanvas)` constructor. I was also intending to create an `Xor` method to create an array of regions that do not intersect with the other regions and only draw outside the original arrays defined.

How do you know there's no conflict? I have not had a chance to articulate the concept I have in mind.

If you believe the clip region and drawing canvas need to be tightly coupled, then I need you to explain WHY before you go and build it. I have a vision for how they NEED to be decoupled from each other. 

I do not see why you have to build an offscreen rendering canvas as part of building a non-rectangular clip-region solution.

BDisp:
This is a PR still in WIP and by chance yesterday I thought it would be great to use `LineCanvas` integrated into Region. The objective is to create a `LineCanvas` instance and in the constructor calculate the area that will be occupied. But these are still thoughts that still need to be assimilated and only with debate can we resolve and develop them. Now immediately just criticizing, without at least wanting to understand, is not the most correct attitude.

tig:
> This is a PR still in WIP and by chance yesterday I thought it would be great to use `LineCanvas` integrated into `Region`. The objective is to create a `LineCanvas` instance and in the constructor calculate the area that will be occupied. But these are still thoughts that still need to be assimilated and only with debate can we resolve and develop them. Now immediately just criticizing, without at least wanting to understand, is not the most correct attitude.

If you want to take my direction on this as criticism, that is up to you. I'm not doing it to be critical. 

I have thought a ton about both topics (non-rectangular clip region and a deferred drawing canvas) but have deprioritized it because I do not think it's critical path to shipping v2. They are both cool, nice-to-have, ideas. 

Because I've deprioritized them, I've not had the time to write my thoughts down. I don't want to take that time now because I want to be focused on addressing the things that will actually allow us to ship the Beta of V2.

If you want to spend your time on this, please go ahead. But don't expect me to dive deep into it until after V2 ships. I will not accept a PR that includes a deferred drawing canvas until after v2 is shipped.

We have a rich list of work items/issues that are more important than either topic. I'd prefer you pick something from the [v2 Beta Backlog](https://github.com/orgs/gui-cs/projects/1/views/1) list than this. But you can work on whatever you want to work on. 

To be a clear on my thinking as possible:

- #3422 is intended to introduce the concept of a non-rectangular clip region based on the hypothesis that having that will enable TG apps to work even better over constrained connections. We don't actually have a lot of users complaining about this, so it is not a critical issue and by no means is it something that MUST be done to ship v2.
- The idea of a deferred drawing canvas is a separate idea/concept. It can potentially enable some really cool features and perhaps even perf improvements. It would be cool. But it is also not something that MUST be done to ship v2.
- We have a dozen other Issues that MUST be fixed before we can ship v2. I plan on focusing all of my energy on those until v2 is done and I want to encourage all other contributors to do the same.

BDisp:
I am closing this PR because it is not fulfilling what it was supposed to, according to the original concept. I distorted the code thinking more about adapting to `Terminal.Gui` and neglected what the code should cover, what is really essential. So please accept my apologies for the time I made you waste. Thanks.

tig:
> I am closing this PR because it is not fulfilling what it was supposed to, according to the original concept. I distorted the code thinking more about adapting to `Terminal.Gui` and neglected what the code should cover, what is really essential. So please accept my apologies for the time I made you waste. Thanks.

No need to apologize! I know I learned some things with this, and I'm sure you did too. In my view, learning is what's life is all about. I appreciate your passion and effort and eagerness to make this project better!

dodexahedron:
Dude don't sweat it. :)

It's clear there was a lot learned, and that's great!

Sometimes things just don't pan out how we originally hoped, but there's nothing wrong with an honest, well-intentioned attempt to improve or fix something. 👍

PR #3438, opened 2024-04-28
Fixes #3420 - Adding analyzers to solution and demonstrating simple use

### LINKED ISSUE #3420: Add internal analyzers and source generators
Implementing some internal analyzers and source generators for Terminal.Gui that add enhanced functionality to specified code elements in a standard and high-performance way.

Right now, the only functional generator is one for enum types that adds high-performance extension methods as alternatives to IsDefined and conversions to int and uint. More are planned and in progress, but I wanted to go ahead and file this issue now that it's in a useable state.

There is also an included PowerShell module that exposes an `Update-Analyzers` commandlet that rebuilds the analyzers, since visual studio can't dynamically reload analyzers that have already been loaded. It provides optional switches to automatically close running VS processes with Terminal.sln opened, re-launch the solution upon completion, output fewer informational messages, and to run without interactive prompting - all of which default to off. This makes it really convenient to take care of that tedious process whenever changes are made to analyzer/generator code. Running it as `Update-Analyzers -Force -AutoLaunch -AutoClose` will, without further prompting, close the VS processes, remove the obj and bin folders in the Terminal.Gui and Terminal.Gui.Analyzers.Internal project directories, build the analyzers in both Debug and Release configurations, and then re-load Visual Studio with Terminal.sln.

dodexahedron:
Alrighty, y'all.

This PR is the next phase of introducing the analyzers.

I made some refinements to the PowerShell modules to make them handle paths better, so it doesn't do things like change your current working directory without going back after it finishes.

But the major item is that the Analyzer, in its current state, is now part of the solution, along with its companion debug project and a unit test project for it.

I also annotated an existing enum with the generator attribute and changed a usage of that enum to use the generated code in one place, as a demonstration of it in use.

Remember: Before you open the solution for the first time after pulling this, import the Terminal.Gui.PowerShell.psd1 module and then run `Build-Analyzers` (optionally also with other supported switches it has available - I typically run it as `Build-Analyzers -AutoClose -AutoLaunch -Force -NoClean`).

Next phase will most likely be including the BitVector32 stuff I mentioned.

This is ready to go if it looks good to y'all.

## Fixes

- Fixes #3420

## Proposed Changes/Todos

- [x] Additional refinements to PowerShell modules
- [x] Add the new analyzer projects to the solution
- [x] Wire the analyzer up to the Terminal.Gui project and show a simple use

## Pull Request checklist:

   - The generator documents code it writes.

### DIFF
--- a/Analyzers/Directory.build.props
+++ b/Analyzers/Directory.build.props
@@ -11,8 +11,8 @@
     <DefineTrace>True</DefineTrace>
   </PropertyGroup>
   <ItemGroup>
-      <PackageReference Update="JetBrains.Annotations" Version="2023.3.0" />
-      <PackageReference Update="JetBrains.ExternalAnnotations" Version="10.2.147" />
+    <PackageReference Include="JetBrains.Annotations" Version="2023.3.0" />
+    <PackageReference Include="JetBrains.ExternalAnnotations" Version="10.2.147" />
   </ItemGroup>
   <ItemGroup>
       <Using Include="System.Buffers" />
--- a/Analyzers/Directory.build.props
+++ b/Analyzers/Directory.build.props
@@ -7,6 +7,8 @@
     <Deterministic>true</Deterministic>
     <UTF8OutPut>true</UTF8OutPut>
     <DefineConstants>$(DefineConstants);JETBRAINS_ANNOTATIONS;CONTRACTS_FULL;CODE_ANALYSIS</DefineConstants>
+    <NoLogo>True</NoLogo>
+    <DefineTrace>True</DefineTrace>
   </PropertyGroup>
   <ItemGroup>
       <Using Include="System.Buffers" />
--- a/Analyzers/Directory.build.props
+++ b/Analyzers/Directory.build.props
@@ -11,6 +11,10 @@
     <DefineTrace>True</DefineTrace>
   </PropertyGroup>
   <ItemGroup>
+      <PackageReference Update="JetBrains.Annotations" Version="2023.3.0" />
+      <PackageReference Update="JetBrains.ExternalAnnotations" Version="10.2.147" />
+  </ItemGroup>
+  <ItemGroup>
       <Using Include="System.Buffers" />
       <Using Include="System.Collections.Specialized" />
       <Using Include="System.Numerics" />
--- a/Analyzers/Directory.build.props
+++ b/Analyzers/Directory.build.props
@@ -1,0 +1,17 @@
+<Project>
+  <PropertyGroup>
+    <Nullable>enable</Nullable>
+    <AnalysisLevel>latest-recommended</AnalysisLevel>
+    <WarningLevel>7</WarningLevel>
+    <CharacterSet>UTF-8</CharacterSet>
+    <Deterministic>true</Deterministic>
+    <UTF8OutPut>true</UTF8OutPut>
+    <DefineConstants>$(DefineConstants);JETBRAINS_ANNOTATIONS;CONTRACTS_FULL;CODE_ANALYSIS</DefineConstants>
+  </PropertyGroup>
+  <ItemGroup>
+      <Using Include="System.Buffers" />
+      <Using Include="System.Collections.Specialized" />
+      <Using Include="System.Numerics" />
+      <Using Include="System.Runtime.CompilerServices" />
+  </ItemGroup>
+</Project>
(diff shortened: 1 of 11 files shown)


dodexahedron:
I also had to make a minor change to the generator to fully qualify a couple of code elements, because of a perpetual thorn in my side: our `Attribute` class, which confuses the compiler anywhere within the Terminal.Gui namespace, even with an explicit type alias.

One of the many reasons a flat namespace is unfortunate, to put it mildly. 😆

dodexahedron:
Also, just for easy inspection for the curious, it is currently configured to output the generated code files to disk, which you can see by going here in the Solution Explorer in Visual Studio:

image

Don't worry about that little red dot. All that means is that those files are not in source control (and they shouldn't be).

If you apply the `[GenerateEnumExtensionMethods]` attribute to another enum, you'll see another code file for its extension class over there pretty much instantly.

Give it a spin!

dodexahedron:
Here's what a method that uses HasFlag looks like in IL:
```cil
    IL_0001: call         valuetype Terminal.Gui.ViewDiagnosticFlags Terminal.Gui.View::get_Diagnostics()
    IL_0006: box          Terminal.Gui.ViewDiagnosticFlags
    IL_000b: ldc.i4.4
    IL_000c: box          Terminal.Gui.ViewDiagnosticFlags
    IL_0011: call         instance bool [System.Runtime]System.Enum::HasFlag(class [System.Runtime]System.Enum)
    IL_0016: brfalse.s    IL_0026
    IL_0018: ldarg.0      // this
    IL_0019: callvirt     instance class Terminal.Gui.ColorScheme Terminal.Gui.View::get_ColorScheme()
    IL_001e: ldnull
    IL_001f: call         bool Terminal.Gui.ColorScheme::op_Inequality(class Terminal.Gui.ColorScheme, class Terminal.Gui.ColorScheme)
    IL_0024: br.s         IL_0027
    IL_0026: ldc.i4.0
    IL_0027: stloc.0      // V_0

    IL_0028: ldloc.0      // V_0
    IL_0029: brfalse.s    IL_007a
```

Notice the boxing, which is heinous.

Here's what the method below it looks like, with FastHasFlags from the generator:
```cil
   IL_0001: call         valuetype Terminal.Gui.ViewDiagnosticFlags Terminal.Gui.View::get_Diagnostics()
   IL_0006: ldc.i4.4
   IL_0007: call         bool Terminal.Gui.ViewDiagnosticFlagsExtensions::FastHasFlags(valuetype Terminal.Gui.ViewDiagnosticFlags, valuetype Terminal.Gui.ViewDiagnosticFlags)
   IL_000c: brfalse.s    IL_001c
   IL_000e: ldarg.0      // this
   IL_000f: callvirt     instance class Terminal.Gui.ColorScheme Terminal.Gui.View::get_ColorScheme()
   IL_0014: ldnull
   IL_0015: call         bool Terminal.Gui.ColorScheme::op_Inequality(class Terminal.Gui.ColorScheme, class Terminal.Gui.ColorScheme)
   IL_001a: br.s         IL_001d
   IL_001c: ldc.i4.0
   IL_001d: stloc.0      // V_0

   IL_001e: ldloc.0      // V_0
   IL_001f: brfalse.s    IL_0070
```

No boxing at all.

This represents a speedup of between 50x and 300x for that operation depending on hardware.

dodexahedron:
I've begun implementation of the BitVector generator, which is branched from the end of this branch.

The intent is to generate a record struct that mirrors exactly what is in the source Flags enum, but exposes the flags as booleans, optional (but on by default) cast operators, as well as a bunch of other handy methods for common scenarios. There's also an interface common to all of them that will be implemented by each generated type, for easy and boxing-free means of interacting with them in generic (both literally and in the C# sense) uses.

I also ideally want to make accompanying analyzers which can point out places where the generator can be applied, though I'll have to see how much work that'll mean once I get there. Probably not much, but I can't predict that yet. The search code is almost identical, typically, but we'll see...

There's a slight but non-zero chance that work on this may result in some changes to the other generator, since the basic code is intended to be reusable by multiple generators, but that shouldn't really be a big deal, since it's most likely they'd be internal changes that don't affect use, but any that would affect use would be easy to fix anyway, because I'm not going to like...change the entire structure of the generated code or anything. Probably just naming, if anything.

dodexahedron:
Also: The generated code is 100% trim-friendly, because the runtime behavior does not depend on attributes or reflection. Those are just metadata for design time and generator use.

dodexahedron:
I'm going to make a new issue for the new generator, though, for the record.

Seems like a reasonable enough distinction even if they're related. It's already a new branch anyway. 🤷‍♂️

Unless you have an objection @tig. I don't care either way because a rename is a few keystrokes and you know I'm going to write a book if it's in the same issue or a new issue anyway. 😅

dodexahedron:
Rebased on current v2_develop so I can branch from here for #3443

PR #3449, opened 2024-05-05
Fix potential github action smells

ceddy4395:
## Fixes
Hey! 🙂
I want to contribute the following changes to your workflow:

- Avoid running CI related actions when no source code has changed
- Avoid executing  scheduled workflows on forks
- Avoid jobs without timeouts

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

## Pull Request checklist:

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -4,6 +4,8 @@ on:
   push:
   # only publish v2 (main or develop); v2 is published via the Terminal.GuiV2Docs repo
     branches: [main, develop]
+    paths: 
+      - docfx/**
 
 permissions:
   id-token: write 
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -29,6 +29,7 @@ jobs:
     # Consider using larger runners for possible analysis time improvements.
     runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
     timeout-minutes: ${{ (matrix.language == 'swift' && 120) || 360 }}
+    if: github.repository == 'gui-cs/Terminal.Gui'|| github.event_name == 'schedule'
     permissions:
       actions: read
       contents: read
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -25,7 +25,7 @@ on:
 
 jobs:
   CodeQL-Build:
-
+    if: github.repository == 'gui-cs/Terminal.Gui'|| github.event_name == 'schedule'
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-
+    timeout-minutes: 10
     steps:
     - uses: actions/checkout@v4
 


dodexahedron (approved):
LGTM

Thanks!

Plenty more we could do, too, but these hit some really big sources of noise. 👍

dodexahedron:
Also, a thought for future consideration for us before it disappears from my head again:

Especially with regards to Roslyn analyzers/generators, we may want to consider [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), so that they don't have to get built every time, unless one of the analyzer projects is actually modified. Would speed up the build actions.

.github/workflows/api-docs.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
   # only publish v2 (main or develop); v2 is published via the Terminal.GuiV2Docs repo
```
dodexahedron, line 5: The comment here is not quite accurate with what this actually does.

.github/workflows/codeql.yml
```diff
@@ -29,6 +29,7 @@ jobs:
     # Consider using larger runners for possible analysis time improvements.
     runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
```
dodexahedron, line 30: We could clean this up, too, if we use an actual matrix.

And the default comments from the template are not relevant for us, so they could go away.

.github/workflows/codeql.yml
```diff
@@ -29,6 +29,7 @@ jobs:
...
     timeout-minutes: ${{ (matrix.language == 'swift' && 120) || 360 }}
+    if: github.repository == 'gui-cs/Terminal.Gui'|| github.event_name == 'schedule'
```
dodexahedron, line 32: This is the best part of it all. Thanks for getting fed up and doing it! :)

It's so noisy for forks, for various other reasons.

.github/workflows/codeql.yml
dodexahedron: Actually... This is codeQL scanning and we do not have OS-specific dependencies.

We could do it on just one runner for this workflow unless anyone has any thoughts otherwise.

dodexahedron (approved):
Still fine to me, regardless of comments I just added.

Those are notes for me while I'm completely re-working things and aren't related to the changes made.

.github/workflows/codeql.yml
ceddy4395: Out of curiosity, what runner would in that case take your preference? The Ubuntu one?

.github/workflows/codeql.yml
ceddy4395: No worries, thank you for the feedback, very appreciated! :smile:

.github/workflows/codeql.yml
dodexahedron: May as well be that one. Likely to have the lowest total running time cost.

ceddy4395: Ah yes clear!

tig:
Is this ready for me to merge?

tig (approved):

dodexahedron:
Hey @tig :

Closely related to this, and also might influence what gets added to the gh marketplace action I'm working on:

What do you have the "Actions permissions" and "Workflow Permissions" set to for this repository, under settings > actions > general? I wanna be sure this repo is covered from the start, since it was the original catalyst for making me do it now rather than [undefined future time range] in the first place. :)

And do you have any environments configured? (that's not important - just surveying what's in use here).

And any defined rulesets or branch/tag protections?

Also, any difficulties or wish list items vis-a-vis the actions/workflows/deployment/builds/etc?

Aside from splitting the analyzers - that is dependent on these other tasks I've got outstanding.

tig (approved):

PR #3457, opened 2024-05-08
Fixes #3456. Analyzers unit test are failing on Windows.

### LINKED ISSUE #3456: Analyzers unit test are failing on Windows.
Running the Analyzers unit test in the v2_develop fail on two tests.

BDisp:
## Fixes

- Fixes #3456

## Proposed Changes/Todos

- [x] Normalize new lines for Windows and Unix

## Pull Request checklist:

### DIFF
--- a/Analyzers/Terminal.Gui.Analyzers.Internal.Tests/IndentedTextWriterExtensionsTests.cs
+++ b/Analyzers/Terminal.Gui.Analyzers.Internal.Tests/IndentedTextWriterExtensionsTests.cs
@@ -45,20 +45,19 @@ public class IndentedTextWriterExtensionsTests
         }
 
         writer.Flush ();
-        var result = sb.ToString ();
+        var result = sb.ToString ().Replace ("\r\n", "\n");
 
         Assert.That (
                      result,
                      Is.EqualTo (
-                                 $"""
-                                  
-                                          {scopeClosing}
-                                        {scopeClosing}
-                                      {scopeClosing}
-                                    {scopeClosing}
-                                  {scopeClosing}
-
-                                  """));
+                                 @$"
+        {scopeClosing}
+      {scopeClosing}
+    {scopeClosing}
+  {scopeClosing}
+{scopeClosing}
+".Replace ("\r\n", "\n")
+                                ));
     }
 
     [Test]
@@ -89,23 +88,23 @@ public class IndentedTextWriterExtensionsTests
         }
 
         writer.Flush ();
-        var result = sb.ToString ();
+        var result = sb.ToString ().Replace ("\r\n", "\n").Trim ();
 
         Assert.That (
                      result,
                      Is.EqualTo (
-                                 $"""
-                                  Opening UninterestingEnum
-                                  {scopeOpening}
-                                    Opening UninterestingEnum
-                                    {scopeOpening}
-                                      Opening UninterestingEnum
-                                      {scopeOpening}
-                                        Opening UninterestingEnum
-                                        {scopeOpening}
-                                          Opening UninterestingEnum
-                                          {scopeOpening}
-
-                                  """));
+                                 @$"
+Opening UninterestingEnum
+{scopeOpening}
+  Opening UninterestingEnum
+  {scopeOpening}
+    Opening UninterestingEnum
+    {scopeOpening}
+      Opening UninterestingEnum
+      {scopeOpening}
+        Opening UninterestingEnum
+        {scopeOpening}".Replace ("\r\n", "\n")
+                       .Trim ()
+                                ));
     }
 }


dodexahedron (changes requested):
Please don't change those.

They are written that way intentionally.

Your formatting settings are what's breaking that.

Instead of any of this, just wrap the method in these two comments:

```csharp
// @formatter:off`

...the code...

// @formatter:on
```

That will make the formatter stop touching it.

Take out the replace, too. Line ending conversions are automatic from git and the entire point of the test is to guarantee verbatim output _on any platform_.

In other words, all of these changes need to be reverted and just the comments above should be added, so your formatting rules won't break things.

dodexahedron:
Side note:

This happens to be like... one of the few places a test that does everything on strings is actually a good thing.

The alternative is to do it with Roslyn, but that's a LOT more work, a LOT less maintainable, and a LOT slower.

dodexahedron:
If your git configuration has autocrlf set to anything other than true, do this to set global config:

`git config --global core.autocrlf true`

To check what the current setting is, as applied for the Terminal.Gui repo, cd to the repository root amd run `git config core.autocrlf`, which will output the effective value.

If it is true, line endings are already being converted as appropriate for the platform.

We should _never_ be doing that replace on string literals in code.

BDisp:
I switched to verbatim but the unit test failure was not because of any formatting on my part. The problem is that on Windows this fails because of newlines. On the git server there was no error because it uses Ubuntu whose new lines are represented by '\n' and on Windows it is '\r''\n'.

BDisp:
`git config --global core.autocrlf` already output as `true`. This issue has nothing to do with that.

dodexahedron:
I'm running it on windows. It's not newlines.

Roslyn outputs native newlines on the platform.

The code file contains native newlines on the platform.

The only way for that to not work properly anywhere is for the environment to have a setting or process that is changing it.

dodexahedron:
Try this:

git restore the file and don't open it in anything at all.

Then build the analyzers with powershell.

Then call `dotnet test` on the Terminal.Gui.Analyzers.Internal.Tests.csproj specifically.

That will at least take Visual Studio and any potential extensions, including ReSharper, out of the mix.

dodexahedron:
Also, what's the output of the following in your pwsh session?

```powershell
Get-Culture && Get-UICulture
```

Shouldn't be an encoding issue since powershell is all UTF-8, as are those projects, but just looking for clues.

If it's still not working after using the original reset file, let me know what this outputs, as well, when run from the repository root (or just with the path appropriately fixed for where you're calling it from):

```powershell
(Get-Content .\Analyzers\Terminal.Gui.Analyzers.Internal.Tests\Generators\EnumExtensions\EnumExtensionMethodsIncrementalGeneratorTests.cs -AsByteStream)[0..5]
```

BDisp:
This is what I'm getting in the v2_develop branch. As you can see is due to newlines differ from each other.

image

dodexahedron:
Right. Which is your environment doing it (or failing to do it).

I updated my previous comment with a couple of commands. Can you show me the output of those?

dodexahedron:
I cloned it clean to a new folder and do not have that problem, btw.

dodexahedron:
For now, though, just skip that test. Those aren't the right changes to be making, even if there's a need to change something.

BDisp:
LCID             Name             DisplayName
----             ----             -----------
2070             pt-PT            português (Portugal)
2070             pt-PT            português (Portugal)

dodexahedron:
Hmmmmmm. I wonder if that's got something to do with it. I don't really know why it would, though.

Is Visual Studio showing the code files as UTF-8 encoding for both that test file and the place in the generator where it is built?

dodexahedron:
Real quick: Just to skip that test, add this to that test method:

`[Ignore("Broken for me - DONT COMMIT")]`

And, like it says, don't commit it, since we're just troubleshooting for you right now.

dodexahedron:
This is also making me think we should have the git actions perform builds and tests in more than just en_US, to catch anything possibly locale-related. Whether this one is or not, that would still be helpful, beyond just making xUnit or NUnit do it in the test methods.

dodexahedron:
For reference, here is what I see in the lower-right corner of the code window.

image

Both the test file and the code file containing that extension class show CRLF for me.

BDisp:
This was also happening on XUnit. See the TestHelpers class on the UnitTests project. It isn't related to the culture but because Windows treats the newlines as "\r\n" and Unix as "\n".
@tig please can you confirm that this is also happening with you? If not then it's something on my side. Thanks.

dodexahedron:
I _did_, years ago, before Visual Studio got better at handling it, have an extension in VS that auto-switched line endings when opening files. I do not have it and haven't for years.

But do you happen to have anything like that?

ReSharper can also cause it, depending on settings.

dodexahedron:
> This was also happening on XUnit. See the TestHelpers class on the UnitTests project. It isn't related to the culture but because Windows treats the newlines as "\r\n" and Unix as "\n". @tig please can you confirm that this is also happening with you? If not then it's something on my side. Thanks.

It would _only_ run on unix if this were the case.

And you were the only one getting the line ending issue with xUnit.

It's something environment-specific, and we need to figure out what, so we can fix it correctly.

BDisp:
> For reference, here is what I see in the lower-right corner of the code window.
> 
> image
> 
> Both the test file and the code file containing that extension class show CRLF for me.

Mine too. That is only the format that is saved on the file. When we push to the server git convert to LF. The Ignore only skip the unit test and it really skipped.

dodexahedron:
Yes. That's what autocrlf does.

Git converts ALL line endings on pull.

And you just saw you have CRLF in the files.

Something else is wrong.

dodexahedron:
Yeah run that Get-Content command I put earlier. I wonder if it's getting mangled some other way.

dodexahedron:
Also, do you have Teams, Discord, or WhatsApp? I'm more than happy to chat to make this easier for us than going back and forth here haha.

BDisp:
image

dodexahedron:
filepath is the path to one of those files, as mentioned above it.

dodexahedron:
That's a neat powershell you have there. What is that?

BDisp:
Opps

image

dodexahedron:
Ok that looks fine.

I'd really love to know the actual source of the problem, since you _do_ have the correct line endings, as you saw in VS.

Right now we know the locale is a difference, but that would seem like a strange reason for this.

Was it failing for `dotnet test` on the analyzer test project too, or just running in VS?

dodexahedron:
I mean... Here it is on one machine, Windows 11 Enterprise, locale 1033 (US English)

image

dodexahedron:
And here's on another system right after a fresh clone of the repo and building the analyzers:

image

BDisp:
image

BDisp:
In the repo the command `git config core.autocrlf` output `input`. It's normal?

dodexahedron:
Well, it shouldn't matter, since you saw for yourself that they're CRLF anyway

As far as git goes, it won't change files you have already pulled, which means you will have to re-clone to be sure when checking if there's a difference between those modes for those files.

BDisp:
I really have them as LF. When I make same change it toggle to MIXED. My bad sorry.

dodexahedron:
Yay!

I'm glad we got to the root cause!

Now you shouldn't have to worry about raw string literals breaking things for you any more. :)

I'm wondering if that may have had other random side-effects on other things you may have encountered before, too. 🤔 

Probably not, unless a string with a line break was involved, of course.

dodexahedron:
This gave me another idea for something to add to the general project build instructions and that I can also include in the powershell modules:

A script or just a few specific `git config` commands to run to ensure things work as expected regardless of what someone's global or system-level git config is.

dodexahedron:
Oh yeah - Don't forget to `git restore` that test file, to un-ignore the test! :)

I think both this and its parent issue can be closed, but I'll leave that to you in case you ran into anything else directly related.

BDisp:
But I still have the unit test issues on Windows.   @tig does this situation happen with you on Windows?

dodexahedron:
That makes no sense at all.

If you have LF in one and not in the other, your file is using LF in one and not in the other.

That's not implicit in code and SHOULD fail for you, if your environment is broken like that.

That's the point. The goal isn't to make the test pass - it's to make a test that asserts a specific condition that will only pass IF things are working as expected.

And they aren't, for you, so the root cause (which I pretty much guarantee is that your file contains LF when it shouldn't) needs to be ascertained.

After I fixed the casing issue I submitted a few hours ago, this builds on 3 different windows machines for me, as well as now 5 linux VMs of different distros and versions.

It ain't the code as written, I promise you. You should never have to do what was done in that PR.

dodexahedron:
Here's what the code on the analyzer project side looks like, for the tests involving push methods:

```csharp
        if (declaration is { Length: > 0 })
        {
            w.WriteLine (declaration); // If this is writing NOT your native line ending, there are bigger problems
        }

        w.WriteLine (scopeDelimiter);

        w.Indent++;
```

Any code fed to it is raw string literals, and line breaks will be native to your platform.

dodexahedron:
Did you rebuild the analyzers after you fixed your line endings?

dodexahedron:
You can also confirm by checking the line ending bytes in the generated code files, since I have the configuration set to write those out currently. If they're a mix of LF and CRLF, you either still have a file with LF or you just need to rebuild the analyzers.

dodexahedron:
LMK how it goes if you do any more troubleshooting. I'm heading to bed for the night.

Good luck! I know we can fix this. :)

BDisp:
I had to rebuild the Analyzers three times for this problem to be resolved. Now I no longer have unit test errors. Thanks for your patience.
So I'm closing this.

dodexahedron:
No problem at all.

I saw some remarks in microsoft's feedback for VS that they do, in fact, want to move it to .net core, and it is in the cards.

And that would make it easier for them to handle dynamic loading since AppDomains aren't a thing anymore.

It's just a HUGE effort to make the jump, because Visual Studio is an absolutely enormous colllection of software.

All we can do is wait and hope it comes sooner rather than later. 🤷‍♂️

PR #3461, opened 2024-05-08
Fixes #3460.  `xxxtoxxx` consistency

### LINKED ISSUE #3029: Refactor `MouseEvent` and `MouseEventEventArgs` to simplify

### LINKED ISSUE #3460: Mixing signatures for ScreenToX between `int` and `Point`
`ScreenToContent` uses `Point`
`ScreenToViewport` uses `int`,`int`
`ScreenToFrame` uses `int`,`int`

image

Is it possible to please standardize these? or is there a reason they are not consistent. 

It makes debugging screen pointer issues difficult if you are trying to switch between the two methods to see which is more correct.

tig:
## Fixes

- Fixes #3460
- Partially Fixes #3029 (`X/Y` -> `Position`)

## Proposed Changes/Todos

- [x] `ScreenToFrame` -> takes `Point`
- [x] `ScreenToViewport` -> takes `Point`


dodexahedron:
Hey quick easy optiization:

Make them take it as `in`, too, so reference can be passed by default without needing to be explicitly done at the callsite.

dodexahedron:
That, by the way, is just a good habit to get into when passing structs - especially blittable ones.

dodexahedron:
For certain other cases, there's also now `ref readonly` for method parameters, in c# 12, but that is _slightly_ different. MS has a matrix somewhere showing the effects and compatibility at the callsite for ref, ref readonly, in, and out, for the curious. I think it's linked from the what's new for c# 12.

tig:
`in` -> done.

dodexahedron:
Hm. Why builing new points in a lot of those lines rather than just passing the point itself?

Terminal.Gui/Application.cs
```diff
@@ -1556,7 +1556,7 @@ internal static void OnMouseEvent (MouseEvent mouseEvent)
             return;
         }
 
-        var view = View.FindDeepestView (Current, mouseEvent.X, mouseEvent.Y);
+        var view = View.FindDeepestView (Current, new (mouseEvent.X, mouseEvent.Y));
```
dodexahedron, line 1559: Why new?

Better to make the MouseEventArgs itself have a point and pass that by ref.

dodexahedron: Can be done non-breaking by just adding the property and modifying the existing properties to update the underlying Point.

Terminal.Gui/View/Adornment/Border.cs
```diff
@@ -273,7 +273,7 @@ protected internal override bool OnMouseEvent (MouseEvent mouseEvent)
 
             // Only start grabbing if the user clicks in the Thickness area
             // Adornment.Contains takes Parent SuperView=relative coords.
-            if (Contains (mouseEvent.X + Parent.Frame.X + Frame.X, mouseEvent.Y + Parent.Frame.Y + Frame.Y))
+            if (Contains (new (mouseEvent.X + Parent.Frame.X + Frame.X, mouseEvent.Y + Parent.Frame.Y + Frame.Y)))
```
dodexahedron, line 276: Hm. I'm pretty sure offsetting one of the points by the other two would be less implicit copying.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -147,25 +146,22 @@ public virtual Rectangle FrameToScreen ()
...
-    /// <param name="x">Screen-relative column.</param>
-    /// <param name="y">Screen-relative row.</param>
-    public virtual Point ScreenToFrame (int x, int y)
+    /// <param name="location">Screen-relative coordinate.</param>
+    public virtual Point ScreenToFrame (in Point location)
     {
         if (SuperView is null)
         {
-            return new (x - Frame.X, y - Frame.Y);
+            return new (location.X - Frame.X, location.Y - Frame.Y);
```
dodexahedron, line 154: Offset here is also a good idea.

dodexahedron (changes requested):
Mostly same minor comments elsewhere, so I won't repeat it all over the place.

Otherwise, YAY!

dodexahedron:
There's also an Offset overload that takes Point, too, which is better when you already have one.

Terminal.Gui/Application.cs
dodexahedron: Actually, thinking about it a bit more...

If the second parameter to FindDeepestView were `ref`, you can either supply one if you already have one or create it in the FindDeepestView method.

`out` also can do that, but it's a bit non-obvious and not recommended to consume an `out` param, even though it's legal.

And `ref` has the advantage it won't _let_ you create an ephemeral value at the callsite.

Lots of other options, too, but key is just trying to minimize how many get created and thrown away.

tig:
> Hm. Why building new points in a lot of those lines rather than just passing the point itself?

You mean in the tests? 

> At the first place in a method that assigns it to something, it'll still get value copied anyway.
> 
> And with a non-referenceable ephemeral new() in a method call, it will be copied twice, making the `in` not apply.

I was focused on getting the API right; perf next.

image

How do you suggest I address this?

Terminal.Gui/View/Layout/ViewLayout.cs
tig: image

tig:
I bit the bullet and changed `MouseEvent` to use `Point`.

dodexahedron:
Sounds like it's probably pretty great.

I'm only half awake after a wrong number caller woke me up, so I'm useless for review right now but I'll get on that as soon as I'm up and running tomorrow. :)

PR #3463, opened 2024-05-09
Fixes #3462 - Change file name casing to fix builds on linux

### LINKED ISSUE #3462: File name casing can break builds on linux
OMG...

So...

Linux is case-sensitive for file names...

And we just happen to have two Directory.build.props files.

What's the problem?

Build has to have a capital B, or else they might not be included for the build.

An Ubuntu 24.04 VM I was testing on was having ridiculous dependency resolution and targeting issues that made zero sense...

Until I saw that lower-case b and renamed the files. Then it all worked perfectly.

The powershell modules also weren't loading properly on that VM, because just one of the files has `Powershell` in its name, rather than `PowerShell`, as specified in the manifest.

While pscore can be configured to ignore case on any platform, that...is a bad idea on Linux, so that needs to be renamed, too.

Anyway, here comes a no-code PR to rename 3 files to fix broken builds for case-sensitive environments (so likely quite a few linux distros or versions thereof and maybe MacOS as well).

And case-only changes on Windows are dumb, too, since it ignores case and you have to rename twice to do it. 😒

dodexahedron:
## Fixes

- Fixes #3462 

## Proposed Changes/Todos

- [X] Rename files for case-sensitive systems

## Pull Request checklist:

  - Literally _all_ I did. 

Ready to go whenever. 🤦‍♂️


dodexahedron:
# Critical, for project members to do ASAP

Adding an important note from the issue over here, as well, for everyone to please do ASAP, or at least ***BEFORE they pull this to their local copies*** once it's merged (may as well just do it now and be done with it):

`git config core.ignorecase false`

Without that, you may end up with duplicate files because of casing and the way windows works, and they might even make it back into the repo if certain sequences of not-uncommon actions are performed.

If you do it in your global config, you won't lose the setting if you re-clone on that machine, so it may be a good idea as well:

`git config core.ignorecase false --global`

tig (approved):

tig:
> # Critical, for project members to do ASAP
> Adding an important note from the issue over here, as well, for everyone to please do ASAP, or at least _**BEFORE they pull this to their local copies**_ once it's merged (may as well just do it now and be done with it):
> 
> `git config core.ignorecase false`
> 
> Without that, you may end up with duplicate files because of casing and the way windows works, and they might even make it back into the repo if certain sequences of not-uncommon actions are performed.
> 
> If you do it in your global config, you won't lose the setting if you re-clone on that machine, so it may be a good idea as well:
> 
> `git config core.ignorecase false --global`

This was all working fine before, for years. 

Now, after doing this, if I try to checkout another branch, I get 

image

image

Please advise.

dodexahedron:
Because that branch has the wrong file names and different settings.

Case sensitivity mismatches SUCK.

Those are the extra files mentioned in the comment above and quoted in the reply.

Delete them. They are extras, because the case was wrong before.

dodexahedron:
Basically, any way that you take care of it, it will be taken care of for good unless you've got some weird sort of diamond inheritance or something going on that can't be auto-resolved.

But then you still can just delete the wrong-cased files, which are the ones with lower-case build and lower-case shell in Powershell.

Basically, if it says it's untracked, it doesn't care about that file any more. It's just trying to help you in case you actually meant to keep them around.

As for a root cause analysis?

It's my fault, and I think I can trace it back to almost an exact time when I screwed it up, because it was a last-second rename of stuff for consistency, but I missed especially that s vs S in a string vs CLI file listing. My bad.

Again, just delete anything untracked and you're all set. As long as you don't `git rm` a file that IS in the tree, you're not deleting anything the repo cares about. 👍 

And it's honestly pretty annoying that the rename doesn't work like it's supposed to, even though I did explicitly use the git cli to do it in the end, _just in case it could make windows behave_. But no. Windows gonna Windows.

dodexahedron:
And if you ever are in a situation with any file anywhere untracked and you're unsure and it's blocking you, just stash them. Then you can carry on and address it later when someone can answer/help/etc.

PR #3451, opened 2024-05-07
Fixes #2432. Implements `Dim.Auto` for `Content` including subviews.

tig:
`DimAutoStyle.Text` is now complete and merged in #3416.

## Fixes

- Fixes #2432 

## Proposed Changes/Todos

- [ ] Fully implement/test `DimAutoStyle.Content`
- [ ] Implemement `Dim.AutoSize (max: dim)`
- [ ] Use `Dialog` and `MessageBox` as test cases - This will require moving the dialog buttons to `Padding` as adornments.
- [ ] Dive deep into all other PosDim objects and get clarity on behavior relative to `Dim.Auto` - Improve unit tests to validate
  - [x] `Dim.Absolute`
  - [ ] `Dim.Auto`
  - [ ] `Dim.Fill`
  - [ ] `Dim.Function`
  - [ ] `Dim.Width`/`Height`
  - [x] `Pos.Absolute`/`At`
  - [ ] `Pos.AnchorEnd`
  - [ ] `Pos.Function`
  - [ ] `Pos.Anchor`
  - [ ] `Pos.Percent`
  - [ ] `Pos.Center`
  - [ ] `Pos.X`/`Y` etc...

PR #3480, opened 2024-05-13
Fixes #3469, #3473, #3482  - `Dim.Auto` fixes and `Pos`/`Dim` refactor to support TGD

### LINKED ISSUE #3469: `Dim.Auto (DimAutoStyle.Text)` confuses `TextFormatter` when `Width` is absoulte and `Height` is auto
- Change `CheckBox`'s constructor to set `Height = Dim.Auto (DimAutoStyle.Text)`
- Checkbox unit tests fail when they shouldn't

image

This test sets Width to an absolute. 

```cs
        var checkBox = new CheckBox
        {
            X = 1,
            Y = Pos.Center (),
            Text = "Check this out 你",
            TextAlignment = Alignment.Centered,
            Width = 25
        };
```

For some reason, this is confusing TextFormatter.

### LINKED ISSUE #3473: `Dim.Auto` screws up `Pos.AnchorEnd()`
I missed a use case in my `Dim.Auto` testing. To repro, run the `ComputedLayout` scenario and notice the "Left" button is not aligned with the "Center" and "Right" buttons. 

This only happens if `Button.Height/Width` are both `Dim.Auto()`.

image

Todo:

- [ ] Build a unit test for this
- [ ] Figure out how to fix

### LINKED ISSUE #3482: v2 Pos classes need to support interrogation for TGD
**Executive Summary**
It is no longer possible to know important things about a given Pos because of the use of [primary constructors](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors), whose parameters are innaccessible through reflection.

**Detail**
Consider `Pos.PosView` (aka relative to)

***V1***
In v1 fields were side and View
- These were accessed via reflection in TGD

For example
```csharp
    public static bool IsCombine( [NotNullWhen( true )] Pos? p )
    {
        if ( p == null )
        {
            return false;
        }

        return p.GetType( ).Name == "PosCombine";
    }
```
Also
```csharp
  var fTarget = posView.GetType().GetField("Target") ?? throw new Exception("PosView was missing expected field 'Target'");
  View view = (View?)fTarget.GetValue(posView) ?? throw new Exception("PosView had a null 'Target' view");
```

***V2***
In v2 internals are now visible to TGD (yay!) so can do like

```
if(p is Pos.PosCombine pc)
{
...
}
```

However there are some changes to use [primary constructors](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors)  e.g.

```csharp
internal class PosView (View view, Side side) : Pos
```
_v2 code for PosView_

Since Side is no longer a field, we cannot read it in TGD.  If in the debugger we examine the source for the object we get that it has this secret hidden field thats super sketchy.

```csharp
        [CompilerGenerated]
        private Side _003Cside_003EP;

        public readonly View Target;

        public PosView(View view, Side side)
        {
            _003Cside_003EP = side;
            Target = view;
            base._002Ector();
        }
```

**Fix**
Fix will be to expose as `public readonly` the constructor properties e.g. like Side 

This will enable the TGD to read them and support its UIs for designing/editting the position stuff is at.

```diff

    internal class PosView (View view, Side side) : Pos
    {
         public readonly View Target = view;
+        public Side TargetSide => side;
```

tig:
## Fixes

- Fixes #3469 
- Fixes #3473
- Fixes #3482 
- Makes progress on https://github.com/gui-cs/Terminal.Gui/pull/3451

## Proposed Changes/Todos

- [x] Replicate #3473 with unit tests.

The problem was I had removed some code I thought was dead in `LayoutSubViews` that forced a double calc of pos/dim in order to make `DimAuto` work when subviews depended on the view's dims. That code was very much "prototype" and didnt belong there anyway. I now have a much more elegant soution baked into `SetRelativeLayout` where if the Dim is DimAuto, we calcuate the Dim first (with Pos == 0) and then the Pos (using the resulalant dim). 

This all highlighted clumisness in having `ContentSize` be nullable. So I changed it to non-nullable and added a `SetContentsSize()` method. The API is still a bit awkward to use so I may refactor it again at some point.
 
- [x] Refactor `AutoSizeTrue/False` unit tests into new tests that make sense.

- [x] New issue: Dim.Auto `min:` defines min size of the `Frame`. This should be min size of `Content.Size`!

I'm pretty confident #3473 is now fixed here. 

Instead of focusing on #3469 (which is partially fixed), I'm now going to use this PR to also address:

- #3482 in order to simplify @tznind's work on TGD.

- [x] Remove nested class heirarchy in PosDim.
- [x] Make the Pos/Dim types public - with API documentation that warns of pitfals of using them in normal situations... should not be needed!
- [x] Make settings on the Pos/Dim types be publicly available (read only).

- [x] Verify #3469 is fixed.

## Pull Request checklist:


tig:
@tznind 

Please pull this down and try with TGD to see if I got it right for your needs! You should not need to use reflection or anything fancy anymore to round trip Pos or Dim objects.

dodexahedron:
I made several comments in the issue for this PR addressing serious concerns I have that make me unable to support this PR, at least not as a permanent solution that makes it to V2 release.

They're shorter than my usual, but still not one-liners. 😅 

Please do go over them, though, as I really think there's an XY problem going on here and that we should be actually fixing it.

dodexahedron:
Wait a sec...

Did you actually already do that and I'm misreading these diffs?

I'm gonna pull your branch and have a look.

dodexahedron:
I think there might be an issue or two from me that this will end up applying to, as well.

I'll look for those later, though, since that's not terribly important at the moment.

dodexahedron:
OMG THANK YOU FOR DOING THIS.

I do have some specific comments but I think it might actually be easier to wait til you're done with your tweaks, if you're still working on it.

Big one, though, that I'd request in a separate PR _after_ this is merged: one type per file, pleeeeaaaase 🥰

dodexahedron:
OK seriously?

After pulling this down, this is one of my favorite PRs in all of V2. Top 5 for sure.

And I'm eagerly awaiting the rest of your intended alterations, though I would suggest that, even if for the same issue number, this PR be cut off at some point soon and the remaining work split to another, because the scope is massive and it's already quite difficult to review.

That'll make it easier for us to handle reviewing remaining work items, as they're pushed for PR.

tig:
I've verified 

- #3469 is fixed

This puppy is ready for review.

dodexahedron:
Alrighty.

I've looked over some of it so far, and the first comment/suggestion I have are regarding the static factory methods for the Dim and Pos base classes, and some structural things.

But both of those are goin to make the merge diff even harder to track because of how ubiquitous these types are throughout the library.

So I'm going to keep those in my pocket for issues/PRs to follow the initial merge of this.

I've got about 3/4 of the code left to look at for this, and I imagine I should be able to wrap up by/at some point tonight (like before 2024-05-17T00:00:00-07:00 ideally).

I may also send you PRs for your branch if there's anything simple that doesn't broaden the scope. I've got one in mind already, but that will be easier to just do in code, so that's what I'm leaning toward, at the moment, for that kind of thing.

Terminal.Gui/View/Layout/Pos.cs
```diff
@@ -0,0 +1,694 @@
...
+///     methods on the <see cref="Pos"/> class to create <see cref="Pos"/> objects instead.
+///     </para>
+/// </remarks>
+/// <param name="position"></param>
+public class PosAbsolute (int position) : Pos
+{
+    /// <summary>
+    ///    The position of the <see cref="View"/> in the layout.
+    /// </summary>
+    public int Position { get; } = position;
```
dodexahedron, line 397: What does this even mean?

It's not tested and it's not used elsewhere, except for a really odd use pointed out in the next comment.

Terminal.Gui/View/Layout/Pos.cs
```diff
@@ -0,0 +1,694 @@
...
+    /// <inheritdoc />
+    public override bool Equals (object other) { return other is PosAbsolute abs && abs.Position == Position; }
+
+    /// <inheritdoc />
+    public override int GetHashCode () { return Position.GetHashCode (); }
+
+    /// <inheritdoc />
+    public override string ToString () { return $"Absolute({Position})"; }
+
+    internal override int Anchor (int width) { return Position; }
```
dodexahedron, line 408: What? 🤔

Terminal.Gui/View/Layout/Pos.cs
```diff
@@ -0,0 +1,694 @@
...
+/// <summary>
+///     Represents a position anchored to the end (right side or bottom).
+/// </summary>
+/// <remarks>
+///     <para>
+///     This is a low-level API that is typically used internally by the layout system. Use the various static
+///     methods on the <see cref="Pos"/> class to create <see cref="Pos"/> objects instead.
+///     </para>
+/// </remarks>
+public class PosAnchorEnd : Pos
```
dodexahedron, line 420: I don't think it should be in this PR, but combining all the AchnorXXX types into a single `Anchored` would be nice.

But that's one piece of what I alluded to earlier with it being easier to show it in code and use it in the next change set.

Terminal.Gui/View/Layout/Pos.cs
```diff
@@ -0,0 +1,694 @@
...
+///                 </term>
+///                 <description>
+///                     Creates a <see cref="Pos"/> object that tracks the Bottom (Y+Height) coordinate of the
+///                     specified <see cref="View"/>
+///                 </description>
+///             </item>
+///         </list>
+///     </para>
+/// </remarks>
+public class Pos
```
dodexahedron, line 151: Making these records would reduce code for these, and it also makes the primary constructors make more sense.

Annoyingly, a primary constructor on a `class` has very different meaning and behavior from the primary constructor on a `record [class]`

Value equality, deconstruction to treat it like a tuple, and other goodies come along with a `record` for free. Roslyn source generators are behind that language feature.

# BUT!

Don't do that either, yet, because there are other fundamental changes I have suggestions for that would make that work pointless.

Terminal.Gui/View/Layout/Pos.cs
```diff
@@ -0,0 +1,694 @@
...
+
+    /// <summary>Creates a <see cref="Pos"/> object that tracks the Top (Y) position of the specified <see cref="View"/>.</summary>
+    /// <returns>The <see cref="Pos"/> that depends on the other view.</returns>
+    /// <param name="view">The <see cref="View"/>  that will be tracked.</param>
+    public static Pos Y (View view) { return new PosView (view, Side.Top); }
+
+    /// <summary>Creates a <see cref="Pos"/> object that tracks the Left (X) position of the specified <see cref="View"/>.</summary>
+    /// <returns>The <see cref="Pos"/> that depends on the other view.</returns>
+    /// <param name="view">The <see cref="View"/>  that will be tracked.</param>
+    public static Pos Left (View view) { return new PosView (view, Side.Left); }
```
dodexahedron, line 322: This is odd and not really idiomatic C#.

This sort of thing is typically exposed as a static property with only a get accessor defined.

But I also need to finish with reviewing the entirety of this PR to inform which potential alternative seems most prudent.

And that still is probably better for a new issue/branch/PR.

dodexahedron (commented):
A few comments added just to give you something til I get done with more of the review.

dodexahedron:
Aside from the review I'm doing right now, I do want to comment that our use of these types at all, rather than leaving that behind in V2 in favor of more typical, familiar, and largely built-in types is a lot of wasted effort, IMO.

Everything these do can be achieved with pre-packaged types along with a few of the enums we have, and using those things directly within anything that needs them, via use of interfaces.

I realize that's not a very good explanation but I'm dragging and it's time for bed, so I'll circle back to this tomorrow (well...today).

Terminal.Gui/View/Layout/Pos.cs
tig: This is the value for an absolute Pos. 

If a dev does `view.X = 1`, a `PosAbsolute (1)` is created and the `1` is stored in `Position`. This previously was named `n`.

It is extensively tested (just not specifically).

Terminal.Gui/View/Layout/Pos.cs
tig: `Anchor` *Is* confusing, if that's what you're asking about.  I don't have a great answer.

tig:
> Aside from the review I'm doing right now, I do want to comment that our use of these types at all, rather than leaving that behind in V2 in favor of more typical, familiar, and largely built-in types is a lot of wasted effort, IMO.
> 
> Everything these do can be achieved with pre-packaged types along with a few of the enums we have, and using those things directly within anything that needs them, via use of interfaces.
> 
> I realize that's not a very good explanation but I'm dragging and it's time for bed, so I'll circle back to this tomorrow (well...today).

You're gonna need to explain this better.

Terminal.Gui/View/Layout/Pos.cs
dodexahedron: Well, it's all part of the same issue, really.

(Specifically, though, the "What?" is about Anchor returning Position, because Position is...probably not a good name for that.)

Zoomed out a bit, though:

Pos and Dim are sorta like a higher-level incarnation of the _components_ of a Point and Size, respectively, but with behavioral preference along for the ride.

Most of what they do is really/should be the responsibility of Views or IPositionable or _whatever_ themselves. Mixing data and behavior in what is conceptually a primitive is discouraged, partially because it leads to exactly this sort of confusion - especially since the type is likely to be used in places that stretch the meaning of the names of properties even further.

With the _behavior_ defined more generically, the need for actual separate types evaporates, and a simple enum is then sufficient.

For a common example, that's how HTML/CSS do it. The enum is `position` (unfortunate name in my opinion) and it is an enum for the desired positioning style. Then, the relevant properties on the DOM element itself are used and the rules of the selected behavior applied.

https://www.w3schools.com/Css/css_positioning.asp

Terminal.Gui/View/Layout/Pos.cs
dodexahedron: If there's not an explicit test on a public API element, it's not tested.

All it takes is to change that line and some things will break while others won't, and the reason is obscured slightly, especially the way these types are used.

But, like in my other comments, this is all related to the same small set of core issues that I'll write once rather than scattered around and hard to follow. :)

dodexahedron:
Haha yes. No worries. I fully intend to, with words and code. :)

I'm just going to do it once, in one place, rather than bits and pieces all over the place.

Terminal.Gui/View/Layout/Pos.cs
dodexahedron: It's a factory pattern, which is perfectly normal and common. Just not quite like this.

Even just some naming changes would help. But that isn't important right now. This one is a fairly minor concern AFAIC, especially since you don't _have_ to use them.

dodexahedron:
Alrighty, going through my notes to see what is worth sticking in this review.

My goal is for mostly important stuff or just side notes/ideas that aren't requests for change, for the remainder of this PR review, so we can merge and go from there and feel things out.

Stay tuned!

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+///     Represents a dimension that is a fixed size.
+/// </summary>
+/// <remarks>
+///     <para>
+///         This is a low-level API that is typically used internally by the layout system. Use the various static
+///         methods on the <see cref="Dim"/> class to create <see cref="Dim"/> objects instead.
+///     </para>
+/// </remarks>
+/// <param name="size"></param>
+public class DimAbsolute (int size) : Dim
```
dodexahedron, line 355: This is just an explanation that I felt was important to say since primary constructors are nice but subtle.

Always remember the parameters in the primary constructor exist in a static context, within the stack frame of the calling code.

That's important, but not a problem right here. Just mentioning.

dodexahedron: And that's _also_ true with `record` types, but they also _can_ be init-only auto-properties, if a same-named property isn't also explicitly declared in that record.

That's where I've seen this sometimes trip people up.

But primary constructors are nice for code quality, so thumbs up here. 👍

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// <inheritdoc/>
+    public override string ToString () { return $"Absolute({Size})"; }
+
+    internal override int Anchor (int size) { return Size; }
+
+    internal override int Calculate (int location, int superviewContentSize, View us, Dimension dimension)
+    {
+        // DimAbsolute.Anchor (int size) ignores width and returns n
+        return Math.Max (Anchor (0), 0);
```
dodexahedron, line 376: Just thought of this right now...

Because of GetHashCode needing a proper implementation and because the name of this is goofy and non-matching, how about the backing field be named appropriately for each type (since that field should be private, that is perfectly cool). Then there won't be these odd mismatches that look like an accident if you're not familiar with the code.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+///     Specifies how <see cref="DimAuto"/> will compute the dimension. The default is <see cref="DimAutoStyle.Auto"/>.
+/// </param>
+/// <param name="minimumContentDim">The minimum dimension the View's ContentSize will be constrained to.</param>
+/// <param name="maximumContentDim">The maximum dimension the View's ContentSize will be fit to. NOT CURRENTLY SUPPORTED.</param>
+public class DimAuto (DimAutoStyle style, Dim minimumContentDim, Dim maximumContentDim) : Dim
+{
+    /// <inheritdoc/>
+    public override bool Equals (object other)
+    {
+        return other is DimAuto auto && auto.MinimumContentDim == MinimumContentDim && auto.MaximumContentDim == MaximumContentDim && auto.Style == Style;
```
dodexahedron, line 402: Pattern matching is much more readable and has more optimization opportunities available to the compiler.

ReSharper should be able to swap this one automatically for you if you put your cursor on the is or one of the ==.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+    }
+
+    /// <inheritdoc/>
+    public override int GetHashCode () { return HashCode.Combine (base.GetHashCode (), MinimumContentDim, MaximumContentDim, Style); }
```
dodexahedron, line 406: Love the use of Combine. That's disappointingly rare in the wild, even though it's the easiest, recommended, standard, and generally best way to do this.

Otherwise, same comments as above, of course.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// <summary>
+    ///     Gets the maximum dimension the View's ContentSize will be fit to. NOT CURRENTLY SUPPORTED.
+    /// </summary>
+    public Dim MaximumContentDim { get; } = maximumContentDim;
```
dodexahedron, line 411: So, here's something that can get.... interesting...

With properties not marked `required`, one can still bypass a primary constructor, in certain ways. Unless they take care to set these themselves, all assumptions break.

Just marking them `required` is good enough for now to deal with that and will make any improper uses that may exist anywhere compiler errors so they can be fixed.

dodexahedron: Also a good idea to make them explicitly init and back them with a readonly field, which will let the compiler know it can make these `call` instead of `callvirt`

Specifically for the ones that aren't virtual or override.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+                            if (size > subviewsSize)
+                            {
+                                subviewsSize = size;
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
dodexahedron, line 486: Holy cyclomatic complexity, Batman!

This stuff should get refactored to reduce nesting.

ReSharper should be able to help out with that a lot.

Usually easiest to work from the outside in, if trying to do it on the outermost first and letting it turn it into a switch doesn't have a nice result.

But this whole nested control flow can be one big single switch with pattern matching, performed in the reverse order (most specific first because order matters), in half or fewer lines, a single level of nesting, and without having to follow a binary tree for a human to parse it.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+        // We take the larger of text and content.
+        int max = int.Max (textSize, subviewsSize);
+
+        // And, if min: is set, it wins if larger
+        max = int.Max (max, autoMin);
+
+        // Factor in adornments
+        Thickness thickness = us.GetAdornmentsThickness ();
+
+        if (dimension == Dimension.Width)
```
dodexahedron, line 498:
```csharp
max += dimension switch
{
  Dimension.Width => thickness.Horizontal,
  Dimension.Height => thickness.Vertical // or just use _ instead of an explicit value, to match all other values
};
```

Not just syntactic sugar.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+        {
+            max += thickness.Horizontal;
+        }
+        else
+        {
+            max += thickness.Vertical;
+        }
+
+        // If max: is set, clamp the return - BUGBUG: Not tested
+        return int.Min (max, MaximumContentDim?.Anchor (superviewContentSize) ?? max);
```
dodexahedron, line 508: Why not Math.Clamp, then?

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+    }
+
+    internal override bool ReferencesOtherViews ()
+    {
+        // BUGBUG: This is not correct. _contentSize may be null.
+        return false; //_style.HasFlag (DimAutoStyle.Content);
```
dodexahedron, line 514: Oh, this reminds me of something unrelated.

If there is a relevant concept of "truth" for these types (like <= 0 or something being false and everything else true), `operator true` and `operator false` can be defined.

These types are a really borderline case for it, though, so this is VERY optional.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+///     methods on the <see cref="Dim"/> class to create <see cref="Dim"/> objects instead.
+/// </remarks>
+/// <param name="left">The left dimension.</param>
+/// <param name="right">The right dimension.</param>
+public class DimCombine (bool add, Dim left, Dim right) : Dim
+{
+    /// <summary>
+    ///     Gets whether the two dimensions are added or subtracted.
+    /// </summary>
+    public bool Add { get; } = add;
```
dodexahedron, line 536: Should be an enum (default or byte-backed, but unspecified - thus int - is usually best, performance-wise) with 2 members where the 0 value is ok to be named the desired default behavior.

Requires a whole lot less explanation of the behavior when the code is expressive for no extra cost.

If it feels weird to write a boolean as "IsX" or "DoesX" or "HasX" or whatever, a boolean may not be the best choice, even if the enum is simply 2 values.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+    ///     Gets the right dimension.
+    /// </summary>
+    public Dim Right { get; } = right;
+
+    /// <inheritdoc/>
+    public override string ToString () { return $"Combine({Left}{(Add ? '+' : '-')}{Right})"; }
+
+    internal override int Anchor (int size)
+    {
+        int la = Left.Anchor (size);
```
dodexahedron, line 553: These are costly because at least 3 copies happen, by the end of the method, whether they get used or not.

Should be in-lined at minimum to remove 1-2 copies.

I see why you have the Anchor as a method, but it should be renamed if it stays that way.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// <summary>
+    ///     Diagnostics API to determine if this Dim object references other views.
+    /// </summary>
+    /// <returns></returns>
+    internal override bool ReferencesOtherViews ()
+    {
+        if (Left.ReferencesOtherViews ())
+        {
+            return true;
```
dodexahedron, line 591: Are these possible to reach? 🤔 

Not tracing them manually, it looks like only false is possible, right here, anyway, because the recursive default is false, making it false up the stack.

Also potentially a really good place for some ref love, because the compiler can massively optimize this sort of case at runtime especially in later tiers, but that needs careful implementation, so just a note.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+    /// <inheritdoc/>
+    public override bool Equals (object other) { return other is DimPercent f && f.Percent == Percent && f.UsePosition == UsePosition; }
+
+    /// <inheritdoc/>
+    public override int GetHashCode () { return Percent.GetHashCode (); }
+
+    /// <summary>
+    ///     Gets the percentage.
+    /// </summary>
+    public new float Percent { get; } = percent;
```
dodexahedron, line 627: Many decimal values cannot be represented as floats and we don't need the range of a float, plus concepts like PositiveInfinity, NegativeInfinity, positive 0, negative 0, and NaN are all meaningless and bugs waiting to happen.

This is almost definitely better to be an integer, short, or byte, for our use case, with a bounds check when assigned.

Also (here and elsewhere, too), an annotation for static analysis such as `[ValueRange(0,100)]` is helpful both internally and externally, and even some documentation tools can pick those up, as well.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// <summary>
+    /// </summary>
+    /// <returns></returns>
+    public override string ToString () { return $"Percent({Percent},{UsePosition})"; }
```
dodexahedron, line 632: The above suggestion, if taken, doesn't need a change here.

But if the above suggestion is not taken, Percent should be `Percent:P0`.

Either way, if you intend for these kinds of ToString overrides to emit values used in parsing, the IFormattable and IParsable interfaces (or, more ideally, their Utf8 span-based versions) need to be implemented, because we're not the only parser in town and people will almost definitely want to be able to use standard components for that.

And that's also not really optional, because numeric formatting and parsing is culture-dependent.

That should probably be its own specific issue and branch/change set though just for sake of scope.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// <summary>
+    ///     Gets whether the dimension is computed using the View's position or ContentSize.
+    /// </summary>
+    public bool UsePosition { get; } = usePosition;
```
dodexahedron, line 637: Another one that would be much better as an enum, with the property's name adjusted accordingly.

For example, `PositionReference` or `PositionMode` or whatever, and the enum having the possible behaviors.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    internal override int Anchor (int size) { return (int)(size * Percent); }
```
dodexahedron, line 639: Oh. Another note about Anchor that just came to mind related to the name of it.

The `VerbNoun` standard for the naming really helps here.

It is not clear that this is a pure method.

The assumption, based on convention, would be to take `Anchor` as the verb definition of it, and thus that this intuitively should be doing something related to that. But it has no side effects whatsoever (pure).

There are reasons the conventions exist and they're part of why the C# ecosystem is so nice to work in.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+/// <summary>
+///     Represents a function <see cref="Dim"/> object that computes the dimension by executing the provided function.
+/// </summary>
+/// <remarks>
+///     This is a low-level API that is typically used internally by the layout system. Use the various static
+///     methods on the <see cref="Dim"/> class to create <see cref="Dim"/> objects instead.
+/// </remarks>
+/// <param name="dim"></param>
+public class DimFunc (Func<int> dim) : Dim
```
dodexahedron, line 682: I keep meaning to bring up that this needs a more specific implementation, because it's problematic in a _lot_ of potential uses, when unrestricted like this.

Still a cool thing to have, but something we can address later on and that I've not had the opportunity to specifically think about, yet.

It's like...almost there, too. 🤔

dodexahedron:
Beware that, at least for me, github collapsed several of the comments into a section you have to expand to see. Lame.

dodexahedron:
There are other big items, too, but it's too much for this PR, and some apply to other portions of the code, too, so I'll file issues after this is done and merged for those.

Makes it more approachable in bite-sized chunks, anyway.

dodexahedron:
And I keep wanting to suggest not spending _too_ much time perfecting the xmldoc comments right now, in favor of holding that off til this stabilizes, to save work and make it less enticing to stick to something for sunk cost fallacy reasons, which we all are guilty of sometimes - especially when we get "GetItDoneItis." 😅 

So that's me both suggesting that but also not, if you'd rather do it now. 🤷‍♂️ 

We can turn that inspection up to error before release, too, to make sure we don't miss any, which is better than having them but them getting out of date as implementation drifts from what's written in them.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+    ///     Gets a dimension that is anchored to a certain point in the layout.
+    ///     This method is typically used internally by the layout system to determine the size of a View.
+    /// </summary>
+    /// <param name="width">The width of the area where the View is being sized (Superview.ContentSize).</param>
+    /// <returns>
+    ///     An integer representing the calculated dimension. The way this dimension is calculated depends on the specific
+    ///     subclass of Dim that is used. For example, DimAbsolute returns a fixed dimension, DimFactor returns a
+    ///     dimension that is a certain percentage of the super view's size, and so on.
+    /// </returns>
+    internal virtual int Anchor (int size) { return 0; }
```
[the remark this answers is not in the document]
tig: It can't be a property. It takes a parameter. 

I'll consider renaming it to GetAnchor or something.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    #endregion operators
+
+    #region overrides
+
+    /// <summary>
+    ///     Diagnostics API to determine if this Dim object references other views.
+    /// </summary>
+    /// <returns></returns>
+    internal virtual bool ReferencesOtherViews () { return false; }
```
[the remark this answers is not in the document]
tig: I likely will just remove it. It doesn't actually work as i had planned.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// <inheritdoc/>
+    public override bool Equals (object other) { return other is Dim abs && abs == this; }
+
+    /// <inheritdoc/>
+    public override int GetHashCode () { return Anchor (0).GetHashCode (); }
```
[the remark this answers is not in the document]
tig: I just removed both overrides from the base class (now abstract).

Terminal.Gui/View/Layout/Dim.cs
tig: I'm not sure what you're alluding to here:

- "this is goofy and non-matching"). 
- "these odd missmatches"

Can you look at latest and clarify?

Terminal.Gui/View/Layout/Dim.cs
tig: Nope. Pattern matching doesn't work here. I did restructure to make more readable tho.

Terminal.Gui/View/Layout/Pos.cs
tig: Added a test. Thanks.

Terminal.Gui/View/Layout/Dim.cs
tig: Done.

Terminal.Gui/View/Layout/Dim.cs
tig: See new comment:

```cs
// TODO: This whole body of code is a WIP (for https://github.com/gui-cs/Terminal.Gui/pull/3451).
```

Terminal.Gui/View/Layout/Dim.cs
tig: Please elaborate. I don't understand what "needs a more specific implementation" means.

tig:
> And I keep wanting to suggest not spending _too_ much time perfecting the xmldoc comments right now, in favor of holding that off til this stabilizes, to save work and make it less enticing to stick to something for sunk cost fallacy reasons, which we all are guilty of sometimes - especially when we get "GetItDoneItis." 😅

I don't think one can spend enough time on XML docs. Nothing drives clarity of thought more than writing clear, crisp, and complete English sentences. 

> 
> So that's me both suggesting that but also not, if you'd rather do it now. 🤷‍♂️
> 
> We can turn that inspection up to error before release, too, to make sure we don't miss any, which is better than having them but them getting out of date as implementation drifts from what's written in them.

tig:
@dodexahedron I believe I've addressed all your feedback with the exception of those where I've left the comment unresolved with a question I added.

Terminal.Gui/View/Layout/Dim.cs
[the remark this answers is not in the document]
dodexahedron: Also a perfectly viable strategy, 😆

Terminal.Gui/View/Layout/Dim.cs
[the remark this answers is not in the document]
dodexahedron: 👍 

Something is nagging me about parts of it, but I can't put my finger on it at the moment.

Whenever it does, I'll bring it up. 🤷‍♂️

Terminal.Gui/View/Layout/Dim.cs
dodexahedron: Will do in a bit.

Terminal.Gui/View/Layout/Dim.cs
dodexahedron: Hm. I'm pretty sure I see a way, but it's not important right now anyway. As altered, it looks enough anyway.

Terminal.Gui/View/Layout/Pos.cs
dodexahedron: Yeah. Annoying, but unit tests gonna unit test, and properties gonna property.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    #endregion virtual methods
+
+    #region operators
+
+    /// <summary>Adds a <see cref="Dim"/> to a <see cref="Dim"/>, yielding a new <see cref="Dim"/>.</summary>
+    /// <param name="left">The first <see cref="Dim"/> to add.</param>
+    /// <param name="right">The second <see cref="Dim"/> to add.</param>
+    /// <returns>The <see cref="Dim"/> that is the sum of the values of <c>left</c> and <c>right</c>.</returns>
+    public static Dim operator + (Dim left, Dim right)
```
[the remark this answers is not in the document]
dodexahedron: For sure.

And there are several interfaces in .net 8 that most of these types actually already _almost_ implement, so it'll at least be easy to deal with once appropriate ones get declared on the various types.

I just had a thought related to one of the previous treads in the review that this prompted.

Gonna go over there and put that one down.

Terminal.Gui/View/Layout/Dim.cs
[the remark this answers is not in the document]
dodexahedron: Yeah that's what I was going for.

But it doesn't in all the types that have that property if I'm not misremembering. As in it was a method for some but a property for others or something.

I'll have to flag this for myself to check in the full code both to see whatever you've tweaked for this.

If you go with GetAnchor() - which was the first that came to mind for me as well - Just have more than one overload.  Why not just optional parameters as the logical follow-up question?

Any that can call the parameterless version get an efficiency bonus.

Normally, I'd go with optional parameters, but these types are in close to 100% of .cs files, and are involved with so much stuff that every little bit helps with them.

But I need to look at the code in-situ before I can make further comment on that.

Just a rename to VerbNoun like GetAnchor or GetWhatever, as appropriate for each, is certainly good enough.

Terminal.Gui/View/Layout/Dim.cs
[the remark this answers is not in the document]
dodexahedron: Actually... Upon further consideration, taking it out entirely isn't actually a good idea, because every derived type will then be using the only defined GetHashCode and Equals, which are object.Equals and object.GetHashCode, neither of which is ideal. I need to think about this more, but they're simple enough to plug in once it's figured out, so no biggie.

Terminal.Gui/View/Layout/Dim.cs
dodexahedron: Nice.

I'm sitting down right now to take a look at the current state of it all. 👍

Terminal.Gui/View/Layout/Dim.cs
dodexahedron: I think I put a WIP tag in for resharper, didn't I? If not, I will. I'll check.

One thing that not a lot of people are aware of with switch statements and expressions is that you can have multiple condition components to the switch without using a guard clause.

It looks like a tuple, basically:

```csharp
switch (value1, value2)
{
  case ({ BoolInValue1: true, StringInValue1: "1" }, 420 ):
  {
    // This will have matched if the `value1` object matched the pattern
    // And if value2 is a numeric type or implicitly convertible to a numeric type with the value 420.
  } 
}

You can have as many components as you need, and discards ( `_` ) can be used when you don't care about one or more of them in a given case.

Handy when you have control flow constructs that depend on both complex types and local variables, parameters, etc that wouldn't work without a guard clause otherwise.

And it's faster at run-time than using a guard clause, because it's compile-time constant, whereas guard clauses have to be evaluated _after_ the lookup/jump.

(Guard clauses still are still useful, though, if you _don't_ have a compile-time constant to match against.)

File: Terminal.Gui/View/Layout/Dim.cs (hunk 22 above)
Comment (dodexahedron) [reply Terminal.Gui/View/Layout/Dim.cs]: I will. It's way out of scope for this PR though, so I'll stick it in a new issue if it still looks like it matters later.

It's minor.

**Comment by dodexahedron** [comment]
> > And I keep wanting to suggest not spending _too_ much time perfecting the xmldoc comments right now, in favor of holding that off til this stabilizes, to save work and make it less enticing to stick to something for sunk cost fallacy reasons, which we all are guilty of sometimes - especially when we get "GetItDoneItis." 😅
> 
> I don't think one can spend enough time on XML docs. Nothing drives clarity of thought more than writing clear, crisp, and complete English sentences.
> 
> > So that's me both suggesting that but also not, if you'd rather do it now. 🤷‍♂️
> > We can turn that inspection up to error before release, too, to make sure we don't miss any, which is better than having them but them getting out of date as implementation drifts from what's written in them.

Man I swear I had responded to this, but I must have mis-clicked. I wrote it on another PC, though, so I'll look when I'm able to on that.

The gist was "Agreed - XmlDoc is good and not optional in the finished product, but changes are likely to result in lost/wasted effort."

If you don't mind, no biggie. Just trying to save you some work and time. :)

**Comment by dodexahedron** [comment]
Specifically for this PR, I mean. Not as a SOP.

**Comment by dodexahedron** [comment]
> @dodexahedron I believe I've addressed all your feedback with the exception of those where I've left the comment unresolved with a question I added.
> 
> Really good stuff, and I appreciate your diligence.

Now that I've (I think) acked existing comments, I'm diving back in to pull and look at the current code. :)

My intent for this pass is, unless there are any biggies that should block the refactor til being fixed, to comment on small things but otherwise target a green light for this PR so we can work on things that are actual changes outside just the refactor (several of which we already let sneak in, but oh well).

**Comment by dodexahedron** [comment]
After I cram dinner down my neck, that is. 😆

**Comment by dodexahedron** [comment]
Not done yet. I fell asleep at some point, which was much needed (what a week, man).

And it's 3AM, so I'm headed back to bed. Should have stuff for ya tomorrow (today). Good work. 🙂

**Comment by dodexahedron** [comment]
😭 I lost a review I was working on. I was on my laptop, booted into Ubuntu, writing in Kate, just in case github or my browser acted up.[^morelikepurgatory]

Then the machine hard locked up for a hardware reason I now need to look up the light flash codes for in the repair manual and probably get Dell on-site to replace if necessary. 😤

At least there wasn't anything profound in it, so I can recreate it on another machine. I'll probably do that til I fall asleep, so half an hour or so tonight. 

[^morelikepurgatory]: Well, not lost per se because it'll be there when I get it working again, but it's not accessible to me right now without taking it apart, which I'm not about to do at half til midnight. 😑

**Comment by dodexahedron** [comment]
Ok maybe not so much tonight. After going through some of my github inbox it's already just about midnight, so I'll pick it up tomorrow. Was planning on Tuesday having dedicated time for TG anyway. 🙂

**Review by dodexahedron — CHANGES_REQUESTED** [review]
Just a few, this time.

After these tweaks, I'd say it's fine to merge.

Then we can get to other changes & enhancements.

File: Terminal.Gui/View/Layout/PosView.cs [hunk 27]
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <inheritdoc/>
+    public override int GetHashCode () { return Target.GetHashCode (); }
+
+    /// <inheritdoc/>
+    public override string ToString ()
+    {
+        string sideString = Side switch
+                            {
+                                Side.Left => "left",
```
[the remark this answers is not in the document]
dodexahedron: Oh also. Can use the generated methods here too.

Terminal.Gui/View/Layout/LayoutStyle.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// <summary>
+    ///     Indicates one or more of the <see cref="View.X"/>, <see cref="View.Y"/>, <see cref="View.Width"/>, or
+    ///     <see cref="View.Height"/>
+    ///     objects are relative to the <see cref="View.SuperView"/> and are computed at layout time.  The position and size of
+    ///     the
+    ///     view
+    ///     will be computed based on these objects at layout time. <see cref="View.Frame"/> will provide the absolute computed
+    ///     values.
+    /// </summary>
+    Computed
```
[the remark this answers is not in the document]
tig: In honor of Miguel, I don't want to change this.

Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -0,0 +1,194 @@
...
+    public override string ToString () { return $"Auto({Style},{MinimumContentDim},{MaximumContentDim})"; }
+
+    internal override int Calculate (int location, int superviewContentSize, View us, Dimension dimension)
+    {
+        var textSize = 0;
+        var subviewsSize = 0;
+
+        int autoMin = MinimumContentDim?.GetAnchor (superviewContentSize) ?? 0;
+
+        if (Style.HasFlag (DimAutoStyle.Text))
```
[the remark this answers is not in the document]
tig: I did this:

```cs
/// <summary>
///     Specifies how <see cref="Dim.Auto"/> will compute the dimension.
/// </summary>
[GenerateEnumExtensionMethods]
[Flags]
public enum DimAutoStyle
{
```

But `FastHasFlags()` does show up. `FastIsDefined()` does though.

What am I missing?

Terminal.Gui/View/Layout/LayoutStyle.cs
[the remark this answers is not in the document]
dodexahedron: Ha no prob. Extremely unimportant obviously. 🙂

Terminal.Gui/View/Layout/DimAuto.cs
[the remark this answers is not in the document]
tig: > It's false by default. Set it true in the attribute. There are several available options to control it.
> 
> ...Unless I forgot to turn that on?

Need to explicitly turn it on and the attribute has to come after [Flags}

```cs
[Flags]
[GenerateEnumExtensionMethods (FastHasFlags = true)]

```

dodexahedron:
🥳

Once I finish migrating the Roslyn stuff, I'll get to that draft issue I started a couple days ago.

Issue #464, opened 2020-05-20
Merge Responder into View?


tig:
For now it is split, in case we want to introduce formal view controllers.  But the design becomes very ugly.

tig:
This was prematurely closed.

tig:
This was fixed.

PR #3498, opened 2024-05-23
Fixes #2489. Create new ScrollBar based on a new Scroll and remove ScrollBarView/ScrollView

### LINKED ISSUE #2489: Create new `ScrollBar` based on a new `Scroll` and remove `ScrollBarView/ScrollView`

BDisp:
## Fixes

- Fixes #2489

## Proposed Changes/Todos

- [x] Add Scroll class and unit tests
- [x] Add ScrollSlider class and unit tests
- [x] Add ScrolBar class and unit tests
- [x] Add ScrollButton class
- [x] Add @tig View.ScrollBars
- [ ] Remove ScrollBarView
- [ ] Remove ScrollView

## Pull Request checklist:


BDisp:
I would appreciate the `Scroll` class being reviewed before moving forward with the `ScollBar` class. Thanks.

tig:
Ooooh! I'm super excited to review this! Will do asap.

BDisp:
The width/height of the Scroll is not limited to 1 but only to what the user wants.

WindowsTerminal_or6sJpC3c5

BDisp:
@tig how about you creating a new `Content` or `Viewport` box in the `AdornmentsEditor` to set the foreground and background colors. Instead of using `Top`, `Right`, `Bottom` and `Left`, it can use the Axis, `ContentSize` or whatever. What do you think?

Terminal.Gui/Views/Scroll.cs
```diff
@@ -0,0 +1,355 @@
+﻿//
```
tig, line 1: Please remove this. No need for any of these file comments.

Terminal.Gui/Views/Scroll.cs
```diff
@@ -0,0 +1,355 @@
...
+/// <summary>
+///     Represents the "inside part" of a scroll bar, minus the arrows.
+/// </summary>
+public class Scroll : View
+{
+    /// <inheritdoc/>
+    public Scroll ()
+    {
+        WantContinuousButtonPressed = true;
+        ClearOnVisibleFalse = false;
```
tig, line 20: Why is this needed?

I've never understood why this property is there. It seems like a hack to work around something that should be fixed at a deeper level.

tig (changes requested):
Really slick!!

I've got a PR with some changes that you might like. I'll submit it vs. making a lot of comments.

tig:
> @tig how about you creating a new `Content` or `Viewport` box in the `AdornmentsEditor` to set the foreground and background colors. Instead of using `Top`, `Right`, `Bottom` and `Left`, it can use the Axis, `ContentSize` or whatever. What do you think?

You mean being able to set the Normal.Background/Forground color for the View itself? Like the colorpicker in `Content Scrolling` does?

image

BDisp:
@tig did you reorganize by placing the backfields before the properties by hand, or through some automatic system? I normally use R#'s Ctrl+E+C and it's not reformatting that way.

tig:
> @tig did you reorganize by placing the backfields before the properties by hand, or through some automatic system? I normally use R#'s Ctrl+E+C and it's not reformatting that way.

By hand. It's still broken.

tig:
> > You mean being able to set the Normal.Background/Forground color for the View itself? Like the colorpicker in `Content Scrolling` does?
> 
> Yes or for the view passed on the `ViewToEdit` property. That's more or less the idea.

Greet idea. I'll do it as part of #3376 where I've already completely re-done `AdornmentEditor`. I'm going to rename it to `ViewEditor` as well. I also want to integrate `ViewEditor` into `AllViewsTester` because I use that a lot to audit view behaviors and being able to tweak settings live is super useful.

tig (changes requested):

tig:
I just submitted another PR with quesitons / comments inline.

Terminal.Gui/Views/Scroll.cs
```diff
@@ -0,0 +1,387 @@
...
+        Initialized += Scroll_Initialized;
+        DrawContent += Scroll_DrawContent;
+        MouseEvent += Scroll_MouseEvent;
+        _slider.DrawContent += Scroll_DrawContent;
+        _slider.MouseEvent += Slider_MouseEvent;
+    }
+
+    private readonly View _slider;
+
+    private int _lastLocation = -1;
```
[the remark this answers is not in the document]
BDisp: Practically the behavior will be identical. If it is nullable I will have to check if it is null or not, because it works like a switch. The value is reset when the mouse is released. But if you insist I change it to nullable.

Terminal.Gui/Views/Scroll.cs
```diff
@@ -0,0 +1,387 @@
...
+
+    private bool _wasSliderMouse;
```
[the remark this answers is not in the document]
BDisp: > I don't fully understand what this is used for and why it and `Parent_LayoutComplete` are needed.
> 

The `_wasSliderMouse` field is necessary because I need to distinguish whether the `Position` was changed due to iteration outside the slider or not. When the `Position` changes, the `AdjustSlider` method is called and may, due to rounding in the calculation, undo the slider's new location. That's why I need the `_wasSliderMouse` field to prevent the `AdjustSlider` method from being invalidated.

> This feels like you are working around some deficiency in the library we should fix at a deeper level because having a subview that can be moved via the mouse is a pretty common thing.
> 
> I don't have any suggestions on how to fix, but it makes me curious.

As for `Parent_LayoutComplete` it may not be necessary and I used it more because of some resizing and ensuring the recalculation of new values. I can try removing it and check if the `Scroll` automatically adjusts to the new values, based on the new superview dimension.

tig:
IMO, View sub-classes should not have the word "View" in them as a general principle. `

`ScrollBarView` should be `ScrollBar`.

BDisp:
> IMO, View sub-classes should not have the word "View" in them as a general principle. `
> 
> `ScrollBarView` should be `ScrollBar`.

I agree, but `ScrollBarView` is still the old one. When the new one is created it will just be called `ScrollBar`.

BDisp:
I have to find out why if I move the scroll with the mouse one line the position changes from 0 to 9. However, returning the position to 0 with the mouse and then clicking on the Position button requires the position to reach 10 for the scroll move 1 line. It's strange, I think it should be the same value in both cases, 9 or 10, don't you think?

WindowsTerminal_LgwTkcDO8x

Terminal.Gui/Views/Scroll.cs
```diff
@@ -60,7 +60,7 @@ public int Position
         get => _position;
         set
         {
-            int barSize = Orientation == Orientation.Vertical ? ContentSize.Height : ContentSize.Width;
+            int barSize = Orientation == Orientation.Vertical ? GetContentSize ().Height : GetContentSize ().Width;
```
dodexahedron, line 63: Why was this done all over the file? This is a concurrency problem and a performance issue resulting in more than 2x the work

dodexahedron (changes requested):
Important question in-line.

Terminal.Gui/Views/Scroll.cs
dodexahedron: I'll have an alternative to propose to you in just a few minutes, for this.

dodexahedron:
Also, line 11 of Scroll.cs is dangerous and we either need to seal the class or not access that property in that way in that spot.

A derived class that has overridden that member will be set by that line, and the consumer can't tell it not to without completely hiding the member, which has a completely _different_ set of consequences.

If the implementation depends on that behavior, then the class needs to be sealed.

If the implementation will be ok as long as a derived class doesn't hide the member (which is not our problem), then it should be handled by setting the backing field - not from a virtual member accessor.

dodexahedron:
There's also an important implication of using the Size struct the way it is in View, as a `Nullable<Size>`, but that's for addressing elsewhere, as it's not directly related to this specific PR.

@dodexahedron (tagging so it shows up via an easy search in my queue for later performance pass consideration)

Terminal.Gui/Views/Scroll.cs
dodexahedron: Also. What happens when `value == 0`?

The two checks in the if statement right below it have the same outcome (but ever so slightly lower performance than) this:

```csharp
switch (value)
{
    case < 0:
    case > 0 when value + barSize > Size:
        return;
}
```

So, value==0 will fall through, but require that two checks be performed, in the existing form (`value < 0 || (value > 0 && value + barSize > Size)`) because, when value is 0:
 1. `value < 0` is evaluated unconditionally no matter the value.
     * result is false, so right operand of the `||` is evaluated
 2. `value > 0` is then evaluated unconditionally.
     * Evaluates to false.
 3. The logical && operator function is called, but due to assembly/RPN magic, it short circuits and returns false before evaluating the final operand (as expected).
     * Result is false, and statement ends, passing control to line 70
 4. Line 70 unconditionally calls `OnPositionChanging (_position, value);` with value=0.
 5. Event is raised with that value being 0, which is why the unit test has to special case 0 and cancel.

This can all be avoided by just dealing with the 0 case. That is particularly important if the consumer has subscribed to the event and now we've given them a pointless event called PositionChanging that isn't actually a changing position, and then also have to have the workaround in both the unit tests AND the class to special case the 0.

dodexahedron: Remember:

When the event is raised, you do not have a guarantee of the order the members in the MulticastDelegate will be called. Their code could run first. Our code could run first. Their code could be async or parallel and run concurrently with our code. Who knows. We can't depend on that.

Raising the event for an invalid state should be avoided, not handled post-hoc.

dodexahedron: OK.

So...

That distraction has changed what my recommendation for addressing the original comment was going to be, but it's a bit more involved than is really appropriate to give code snippets for here.

The extra commentary may also change as a result of this, so consider this part first:

Unfortunately, the GetContentSize method creates the root problem (the property had the same problem, but not because it was a property).

So, those methods all need to know the size of the scroll bar, clearly. And that method is a perfectly reasonable idea for delegating the responsibility to a container, in various other situations.

But, even in those other situations, it is sub-optimal, for performance, on top of the local issues described above.

Much more optimally, it should store its own size.

"But how would it handle its container changing size?"

Events!

But the explanation is a lot more than the code itself, a significant amount of which actually does already exist. :)

I'm not sure right at the moment what my ultimate recommendation would be for a "proper" fix, here.

Perhaps just what the original recommendation was going to be, which would be changing the beginning of the set accessor for Position to be:

```csharp
            if (value <= 0)
            {
                return;
            }

            Size contentSize = GetContentSize ();
            int barSize = Orientation == Orientation.Vertical ? contentSize.Height : contentSize.Width;

            if (value + barSize > _size)
            {
                return;
            }
```

This does the following:
 * Covers the 0 case by returning immediately.
 * Also returns before bothering to get and assign `barSize`, which is at least 3 copies.
 * The conditional compiles to the same code as before, so that's an optional change.
 * Uses the field instead of the property in the last check.

This does not address the various other things I brought up, however.

But that can be done another time.

dodexahedron:
Not the fault of this PR, but as I said I would do when we got further along, a while back, I'm now giving a closer look to things like I started talking about in my review.

That stuff, when traced through just a little, ultimately reveals opportunities in the position and size code to avoid a non-trivial amount of heft from things like copies (both implicit and explicit) and possible boxing, as well as to at least somewhat mitigate potential concurrency problems. Most of that would be via `ref` returns and such, for internal code especially, but potentially also could be exposed to the consumer, as well.

I think it's best for that to be done separately from here, though, because of scope.

Terminal.Gui/Views/Scroll.cs
BDisp: > This does the following:
> 
> * Covers the 0 case by returning immediately.

I changed the condition to return if value equals position or is <0 or value + _barSize > Size.
Remember that I have to consider 0 as valid if the current position is different.

> * Also returns before bothering to get and assign `barSize`, which is at least 3 copies.

I created a private field `_barSize` and I think its value will be cached. On the other hand, it will only be executed if both previous conditions are not met.

> * The conditional compiles to the same code as before, so that's an optional change.
> * Uses the field instead of the property in the last check.
> 
> This does not address the various other things I brought up, however.
> 
> But that can be done another time.

I think it's more fluid now but I'm sure you would improve it enormously when you deal with this subject.

BDisp:
I added highlight effect in the slider. I didn't used the `HighlightStyle` enum because the hover isn't working.

https://github.com/gui-cs/Terminal.Gui/assets/13117724/17107090-f122-45f1-9bc9-8b5a6c5f088b

PS:
I need a `ColorSchemeChanged` event to intercept the change before `DrawContent`.

Terminal.Gui/Views/Scroll.cs
dodexahedron: Yep that's a better way for now.

I'll have a look at it in a minute to see if anything jumps out at me.

Terminal.Gui/Views/ScrollBarView.cs
```diff
@@ -40,7 +40,7 @@ public ScrollBarView ()
         WantContinuousButtonPressed = true;
         ClearOnVisibleFalse = false;
 
-        Added += (s, e) => CreateBottomRightCorner (e.Parent);
```
dodexahedron, line 43: I missed these before.

Please use named methods instead of anonymous methods for event subscriptions.

Unsubscription cannot be guaranteed with anonymous.

The `(s,e) => ` is what's creating that anonymous method.

UICatalog/KeyBindingsDialog.cs
```diff
@@ -209,7 +209,7 @@ private void RecordView (View view)
             // (and always was wrong). Parents don't get to be told when new views are added
             // to them
 
-            view.Added += (s, e) => RecordView (e.Child);
+            view.Added += (s, e) => RecordView (e.SubView);
```
dodexahedron, line 212: Another anonymous subscription.

Terminal.Gui/Views/Scroll.cs
```diff
@@ -0,0 +1,408 @@
...
+        _slider.MouseLeave += Slider_MouseLeave;
+    }
+
+    private readonly View _slider;
+
+    private int _lastLocation = -1;
+
+    private bool _wasSliderMouse;
+
+    private int _barSize => Orientation == Orientation.Vertical ? GetContentSize ().Height : GetContentSize ().Width;
```
dodexahedron, line 41: This doesn't quite do what you were thinking, because this won't cache.

This actually is still a property, and created an _additional_ copy and layer of indirection.

When a member is declared with `=>`, it is a get-only calculated property.

If you want a get-once-and-cache behavior for a property, this is how you declare it:

```csharp
private T _propertyName { get; } = expression;
```

But, that's not going to have identical behavior, because it will never evaluate `expression` again.

That's why I ultimately had just suggested the easiest thing for you for now would be to basically paste in that code snippet from one of the later comments.

dodexahedron (changes requested):
Took a look at it. Doesn't quite do what you were hoping, but fix is simple. :)

Terminal.Gui/Views/Scroll.cs
dodexahedron: I'm a bit unclear on what you mean about the 0 case. 0 would mean no position change. And the unit test explicitly cancels for that case anyway.

BDisp: No, 0 is position 0. If position is 3 and  then is defined as 0 it must be considered to invoke the events.

dodexahedron: Ah, ok, gotcha.

That wasn't clear to me last night, but I'll take another look at it later with that in mind.

dodexahedron: But do take note of the comment I added about the _position property (it's not a field). That wasn't quite what I was going for and just moved the same "problem" to a different place.

BDisp: I think you are talking about _barSize and it really isn't a field but a getter property. I only did what you asked for which was moving to only one place. It must be a calculated variable and the way I have it before was better because it's only called once on each place.

dodexahedron: Yep. I understand why you did it, and it's _almost_ what I was going for.

However, using a property or other type of method (remember a property accessor is a method) still leaves open the concurrency problem I mentioned, because it's entirely possible for the same method to ask for the property twice and get a different value for each access.

The methods in this class that use that new property do have multiple uses of it.

Storing the value in a local and using the local throughout the method is safer, and ensures that you will always have the same value at all uses in the scope of the method.

While you did in fact reduce some duplication, that wasn't the main goal - fixing the concurrency issue was the important part.

dodexahedron: Another thing is that making a property dependent upon another class directly, that way, creates a non-obvious and tight coupling between the actual class and the other class. And it being private hides what's actually going on from a consumer, which makes their type actually dependent on that property without knowing it.

Use of a local restricts the coupling to the scope of the specific methods, at least.

What I'd do, if you like the private member means of getting it, is make the property a method instead, call it once right before the first time it is used in those two methods, store the returned int in a local variable, and then use the variable only, after that, in the method.

UICatalog/KeyBindingsDialog.cs
BDisp: ScrollBarView.cs will be removed later, so isn't a problem for now.

Terminal.Gui/Views/Scroll.cs
BDisp: Done see https://github.com/gui-cs/Terminal.Gui/pull/3498/commits/5d2120eac69f2bec85947842b87d78daaa5b26c3.

Terminal.Gui/Views/ScrollBarView.cs
BDisp: `ScrollBarView.cs` will be removed later, so isn't a problem for now.

UICatalog/KeyBindingsDialog.cs
BDisp: That file belongs to `UICatalog` and objects are always discarded on exit. On the other hand, it is not part of this PR so it could be used in a new issue to prevent all these anonymous methods from unsubscribing.

BDisp:
> PS: I need a `ColorSchemeChanged` event to intercept the change before `DrawContent`.

Not needed I just override the `GetNormalColor`. Thanks @tig.

tig:
This is looking really nice!!!

Have you noticed that if you click and drag outside of the scroll bar's superview, mouse tracking gets lost? See this vid for what I mean. 

6NkDc55 1

This doesn't feel right to me. 

Slider does this correctly:

yyjZMyf 1

BDisp:
> This is looking really nice!!!
> 
Thanks.

> Have you noticed that if you click and drag outside of the scroll bar's superview, mouse tracking gets lost? See this vid for what I mean.
> 

Yes I did.

> U6BF1ge 1 [ U6BF1ge 1 ](https://private-user-images.githubusercontent.com/585482/342404879-b3f73af7-2604-4643-ab7b-bb445d6ab9e8.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTkyNDQ0MDQsIm5iZiI6MTcxOTI0NDEwNCwicGF0aCI6Ii81ODU0ODIvMzQyNDA0ODc5LWIzZjczYWY3LTI2MDQtNDY0My1hYjdiLWJiNDQ1ZDZhYjllOC5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNjI0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDYyNFQxNTQ4MjRaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0wNjM1YmQxMTM3MDYxOGI0ZWNhODVmZjEwNTcyMTkzY2MxYjI2OTRmZDkxZDAyZTMyYmY0MjExZmExN2VhNzQzJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.lbuXuyqXuwvzmh0hFMqwKGZ88aIKOOBDxdWWXItdnkc) [ ](https://private-user-images.githubusercontent.com/585482/342404879-b3f73af7-2604-4643-ab7b-bb445d6ab9e8.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTkyNDQ0MDQsIm5iZiI6MTcxOTI0NDEwNCwicGF0aCI6Ii81ODU0ODIvMzQyNDA0ODc5LWIzZjczYWY3LTI2MDQtNDY0My1hYjdiLWJiNDQ1ZDZhYjllOC5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNjI0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDYyNFQxNTQ4MjRaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0wNjM1YmQxMTM3MDYxOGI0ZWNhODVmZjEwNTcyMTkzY2MxYjI2OTRmZDkxZDAyZTMyYmY0MjExZmExN2VhNzQzJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.lbuXuyqXuwvzmh0hFMqwKGZ88aIKOOBDxdWWXItdnkc)
> 
> This doesn't feel right to me.
> 

I think this is correct. The scroll does not scroll because the X axis of the outer frame is smaller than the X axis of the scroll frame. That's why you don't see it move to the right.

> Slider does this correctly:
> 

The slider you see moves vertically because you are dealing with the Y axis and not the X axis.

> yyjZMyf 1 [ yyjZMyf 1 ](https://private-user-images.githubusercontent.com/585482/342404371-a596dac4-2e9d-4225-80f6-48ef65ec4055.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTkyNDQ0MDQsIm5iZiI6MTcxOTI0NDEwNCwicGF0aCI6Ii81ODU0ODIvMzQyNDA0MzcxLWE1OTZkYWM0LTJlOWQtNDIyNS04MGY2LTQ4ZWY2NWVjNDA1NS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNjI0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDYyNFQxNTQ4MjRaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1iZWYzYTBlY2MzNzkzZTk3NzZjNGVmYjAxZjg1OWYwOTY2Y2JhNDQ4NzE3ZDQ4M2U5NTJiNDhmYjFhZGM5NTk5JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.xvHJBBANqnrBUkzGJwR3fVByRWo9_MA4O-nTGclK3bE) [ ](https://private-user-images.githubusercontent.com/585482/342404371-a596dac4-2e9d-4225-80f6-48ef65ec4055.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTkyNDQ0MDQsIm5iZiI6MTcxOTI0NDEwNCwicGF0aCI6Ii81ODU0ODIvMzQyNDA0MzcxLWE1OTZkYWM0LTJlOWQtNDIyNS04MGY2LTQ4ZWY2NWVjNDA1NS5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjQwNjI0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI0MDYyNFQxNTQ4MjRaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1iZWYzYTBlY2MzNzkzZTk3NzZjNGVmYjAxZjg1OWYwOTY2Y2JhNDQ4NzE3ZDQ4M2U5NTJiNDhmYjFhZGM5NTk5JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.xvHJBBANqnrBUkzGJwR3fVByRWo9_MA4O-nTGclK3bE)

tig:
> I think this is correct. The scroll does not scroll because the X axis of the outer frame is smaller than the X axis of the scroll frame. That's why you don't see it move to the right.

I'm not explaining myself well. 

If you move the mouse really fast outside of the view that hosts the scroll bar, mouse tracking within the scroll bar stops.

BDisp:
> I'm not explaining myself well.
> 
> If you move the mouse really fast outside of the view that hosts the scroll bar, mouse tracking within the scroll bar stops.

Ah maybe because the `MouseGrabView` become null and loses the mouse tracking? I'll see that later. I cannot test that now. Can you please test the same with the vertical scroll orientation? Thanks.

dodexahedron:
> > I think this is correct. The scroll does not scroll because the X axis of the outer frame is smaller than the X axis of the scroll frame. That's why you don't see it move to the right.
> 
> I'm not explaining myself well.
> 
> If you move the mouse really fast outside of the view that hosts the scroll bar, mouse tracking within the scroll bar stops.

Good catch.

Man, it would be awesome if we had some automated UI testing to add this sort of thing to. Sucks only having 24 hours in a day, huh? 😅

Terminal.Gui/Views/Scroll/Scroll.cs
```diff
@@ -0,0 +1,203 @@
...
+    public Scroll ()
+    {
+        _slider = new (this);
+        Add (_slider);
+
+        WantContinuousButtonPressed = true;
+        CanFocus = false;
+        Orientation = Orientation.Vertical;
+        Width = Dim.Auto (DimAutoStyle.Content, 1);
+        Height = Dim.Auto (DimAutoStyle.Content, 1);
```
BDisp, line 22: @tig if a derived view want to constraint the `Width` and the `Height` to avoid the user set inappropriate values, how that can be done, if they aren't virtual?

BDisp:
> I'm not explaining myself well.
> 
> If you move the mouse really fast outside of the view that hosts the scroll bar, mouse tracking within the scroll bar stops.

@tig this is already fixed in this commit https://github.com/gui-cs/Terminal.Gui/pull/3498/commits/9b896573ece09018ffe7502157209395b046ed7d.

Terminal.Gui/Views/Scroll/Scroll.cs
BDisp: @tig I need your help here, please. I really want to use `Dim.Auto` on this PR but I'm having problems obtained the maximum value. With `DimAutoStyle.Content`, omitted maximum parameter should use the superview `Viewport` size, I think. I'm wrong with this? I confess that I'm really confuse with this. The bellow test is wrong? I also tried with the commented line.

```cs
    [Fact]
    public void DimAutoStyle_Content_On_SubView_Added_To_A_SubView ()
    {
        var supView = new View { Width = 10, Height = 10 };
        var subView = new View { Width = Dim.Auto (DimAutoStyle.Content, 1), Height = Dim.Auto (DimAutoStyle.Content, 1) };
        //var subView = new View { Width = Dim.Auto (DimAutoStyle.Content, 1), Height = Dim.Auto (DimAutoStyle.Content, 1, Fill ()) };
        supView.Add (subView);

        supView.SetRelativeLayout (supView.Frame.Size);
        Assert.Equal (new (10, 10), subView.Frame.Size);
        Assert.Equal (new (0, 0), subView.Frame.Location);
    }

```

BDisp:
> > I'm not explaining myself well.
> > If you move the mouse really fast outside of the view that hosts the scroll bar, mouse tracking within the scroll bar stops.
> 
> @tig this is already fixed in this commit [9b89657](https://github.com/gui-cs/Terminal.Gui/commit/9b896573ece09018ffe7502157209395b046ed7d).

The only issue is when the mouse go outside of the top stopping responding. This is only happens with WT. With conhost works well. I already opened an issue in https://github.com/microsoft/terminal/issues/17744.

Terminal.Gui/Views/Scroll/Scroll.cs
```diff
@@ -35,6 +30,37 @@ public Scroll ()
 
     private int _size;
 
+    /// <inheritdoc/>
+    public override void EndInit ()
```
dodexahedron, line 34: So, just thinking here...

Should we really have EndInit on the base class `public`?

Seems to me that should be `protected` or, if we have coupling internally outside of its hierarchy, `protected internal` (accessible to everything internal, as well as to derived types both internally and externally).

dodexahedron (commented):
This just made me think...

Terminal.Gui/Views/Scroll/Scroll.cs
dodexahedron: Continuing the thought...

With the Being/End pattern, we are also already set up for making them `async`, as well, which makes even more sense if it remains public.

BDisp: I overridden the `EndInit` instead of using the `Initialized` event because this is a View derived class and where the event is being raised. Any changes on `BeginInit/EndInit` isn't the propose of this PR and you'll need to open an issue for that.

dodexahedron: Yeah it wasn't meant for this PR. Just thinking out loud for future possibilities.

Terminal.Gui/Views/Scroll/Scroll.cs
```diff
@@ -10,8 +10,11 @@ namespace Terminal.Gui;
 public class Scroll : View
 {
     /// <inheritdoc/>
-    public Scroll ()
+    public Scroll () : this (null) { }
```
dodexahedron, line 13: You can just make that an optional parameter for the same effect.

`public Scroll (Scrollbar? host = null)`

BDisp: I know but for `All View Tester` scenario every `View` sub-class must have a default constructor (parameterless) to being instantiating using reflection. Thanks for remind me.

BDisp:
@tig when you have time can you please starting review this before I continue further. Thanks.

Terminal.Gui/Views/Scroll/ScrollBar.cs
```diff
@@ -0,0 +1,118 @@
...
+        Orientation = Orientation.Vertical;
+        Width = Dim.Auto (DimAutoStyle.Content, 1);
+        Height = Dim.Auto (DimAutoStyle.Content, 1);
+
+        _scroll.PositionChanging += Scroll_PositionChanging;
+        _scroll.PositionChanged += Scroll_PositionChanged;
+        _scroll.SizeChanged += _scroll_SizeChanged;
+    }
+
+    private readonly Scroll _scroll;
```
tig, line 35: I know this isn't ready for review yet, but I took a look...

Why are these not just subviews? Why don't they just access `SuperView` to access "host"?

BDisp: `Scroll` and `ScrollBar` can be used separately. When only the `Scroll` is used the superview won't be a `ScrollBar` and thus the `Scroll` don't have `ScrollBar` as host. Am I wrong here?

tig: I'm very confused. 

TG has Subviews and SuperViews to support view containment. ScrollBar contains 3 subviews. Why are you not just adding them with Add instead of inventing a new containment system with "_host"?

BDisp: You've absolutely correct, sorry. I removed all the host parameter on the constructors. I renamed the `_host` to a private method `SupView` and maintained the _scroll, _decrease and _increase for convenience because they help on handling these objects. Fixed in https://github.com/gui-cs/Terminal.Gui/pull/3498/commits/ae3fcef17a44ef1797ab54c1e1ba73b8512bd813. Thanks.

BDisp:
@tig do you remember any change that may cause unit test failing on macOS?

Terminal.Gui/Views/Scroll/ScrollButton.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+        else
+        {
+            ColorScheme = new () { Normal = new (SupView.ColorScheme.Normal.Background, SupView.ColorScheme.Normal.Foreground) };
+        }
+
+        return base.GetNormalColor ();
+    }
+
+    public VariationMode VariationMode { get; init; }
```
tig, line 70: `internal`

Terminal.Gui/Views/Scroll/ScrollButton.cs
```diff
@@ -0,0 +1,142 @@
+﻿#nullable enable
+
+namespace Terminal.Gui;
+
+internal enum VariationMode
```
tig, line 5: One type per file please. 

Please add API docs even though this is internal to explain intent.

tig: Consider just using `NavigationDirection` for this, renaming it to `Direction` to make it generic. 

This way we reduce the number of new types?

Or, `AddOrSubtract` might work too?

Terminal.Gui/Views/Scroll/Scroll.cs
tig: Sorry, I just saw this comment. Do you stil need help?

Terminal.Gui/Views/Scroll/Scroll.cs
```diff
@@ -0,0 +1,214 @@
...
+    /// </summary>
+    protected virtual CancelEventArgs<int> OnPositionChanging (int currentPos, int newPos)
+    {
+        CancelEventArgs<int> args = new (ref currentPos, ref newPos);
+        PositionChanging?.Invoke (this, args);
+
+        return args;
+    }
+
+    /// <summary>Virtual method called when <see cref="Size"/> has changed. Raises <see cref="SizeChanged"/>.</summary>
```
tig, line 177: Not "Virtual"

Terminal.Gui/Views/Scroll/Scroll.cs
```diff
@@ -0,0 +1,214 @@
...
+            Width = Orientation == Orientation.Vertical ? Dim.Fill () : Dim.Fill (1);
+            Height = Orientation == Orientation.Vertical ? Dim.Fill (1) : Dim.Fill ();
+        }
+
+        _slider.AdjustSlider ();
+        SetScrollText ();
+    }
+
+    /// <inheritdoc/>
+    internal override void OnLayoutComplete (LayoutEventArgs args)
```
tig, line 195: Do you know precisely what situations require this override?

I like to ask that each time I'm tempted to ovrride OnLayoutStarted/Complete because often it means I'm not doing something else right. 

Sometimes it is needed though.

tig: Oh, and are you sure it `OnLayoutStarted` doesn't make more sense? in AdjustScroll you neve test any super/subviews, so it seems so.

Terminal.Gui/Views/Scroll/ScrollButton.cs
```diff
@@ -0,0 +1,142 @@
...
+            HotNormal = new (_savedColorScheme.Normal.Foreground, _savedColorScheme.Normal.Foreground),
+            HotFocus = new (_savedColorScheme.HotFocus.Foreground, _savedColorScheme.HotFocus.Foreground),
+            Disabled = new (_savedColorScheme.Disabled.Foreground, _savedColorScheme.Disabled.Foreground)
+        };
+
+        return base.OnMouseEnter (mouseEvent);
+    }
+
+    /// <inheritdoc/>
+    protected internal override bool OnMouseEvent (MouseEvent mouseEvent)
```
tig, line 90: Seems to me that this would be more clear if it was a `MouseClicked` handler vs. override of OnMouseEvent?

Terminal.Gui/Views/Scroll/ScrollButton.cs
```diff
@@ -0,0 +1,142 @@
...
+            case VariationMode.Increase:
+                Text = SupView.Orientation == Orientation.Vertical ? Glyphs.DownArrow.ToString () : Glyphs.RightArrow.ToString ();
+
+                break;
+            default:
+                throw new ArgumentOutOfRangeException ();
+        }
+    }
+
+    private ScrollBar SupView => (SuperView as ScrollBar)!;
```
tig, line 141:
```cs
// Helper to avoid having to cast SuperView on each reference
private ScrollBar SuperViewAsScrollBar => (SuperView as ScrollBar);
```

Terminal.Gui/Views/Scroll/ScrollBar.cs
```diff
@@ -0,0 +1,121 @@
+﻿#nullable enable
+
+using System.ComponentModel;
+
+namespace Terminal.Gui;
+
+/// <summary>ScrollBars are views that display a 1-character scrollbar, either horizontal or vertical</summary>
```
tig, line 7:
```cs
/// <summary>A proportional scroll bar that can be oriented either horizontally or vertically.</summary>
/// <remarks>
///     <para>
///         <see cref="Position"/> indicates the current location between zero and <see cref="Size"/>. 
///     </para>
///     <para>If the scrollbar is larger than three cells, arrow indicators are drawn.</para>
/// </remarks>
```

Terminal.Gui/Views/Scroll/Scroll.cs
```diff
@@ -0,0 +1,214 @@
+﻿#nullable enable
+
+using System.ComponentModel;
+
+namespace Terminal.Gui;
+
+/// <summary>
+///     Provides a proportional control for scrolling through content. Used within a <see cref="ScrollBar"/>.
+/// </summary>
+public class Scroll : View
```
tig, line 10:
```cs
/// <summary>
///     Indicates the position and size of scrolllable content. The indicator can be dragged with the mouse.  Can be oriented either vertically or horizontally. Used within a <see cref="ScrollBar"/>.
/// </summary>
<remarks>
    <para>
         By default, this view cannot be focused and does not support keyboard.
    </para>
</remarks>
```

tig (changes requested):
Some embedded comments & tweaks.

I hope to be able to pull this down locally later today and play with it! 

It is looking really good!

Terminal.Gui/Views/Scroll/ScrollButton.cs
BDisp: The `ScrollButton` class and the `VariationMode` enum are already internal and thus the public keyword is limited by the internal keyword of their main objects.

Terminal.Gui/Views/Scroll/ScrollButton.cs
BDisp: I opted to use the `NavigationDirection.Backward` and `NavigationDirection.Forward`. Thanks for the suggestion.

Terminal.Gui/Views/Scroll/Scroll.cs
BDisp: > Sorry, I just saw this comment. Do you stil need help?

No problem. No I don't, I was confused at that time.

> Not "Virtual"

I removed the virtual. If you want it virtual please let me know. Thanks.

Terminal.Gui/Views/Scroll/Scroll.cs
BDisp: > Do you know precisely what situations require this override?
> 

Mainly for console resize, where recalculation, others than layout itself, may cause change on some calculation that will envolve layout changes.

> I like to ask that each time I'm tempted to ovrride OnLayoutStarted/Complete because often it means I'm not doing something else right.
> 
Sometimes using the `OnLayoutStarted` may cause recursive execution that may cause undesired behaviors.

> Sometimes it is needed though.

Yes, indeed.

> Oh, and are you sure it OnLayoutStarted doesn't make more sense? in AdjustScroll you neve test any super/subviews, so it seems so.

When the `Scroll` is used alone I want allow the user the Pos/Dim he want. When the super-view is a `ScrollBar` only makes sense let the `ScrollBar` handle the correct position and dimension. When the `ScrollBar` is used alone, allow the user also set the Pos/Dim he want.

Terminal.Gui/Views/Scroll/ScrollButton.cs
BDisp: I don't think so because it's handling the `Button1Pressed` since the `WantContinuousButtonPressed` is true. I doubt the `MouseClicked` handles with this because it's needed a `Button1Released` to chain a `Button1Clicked`.

Terminal.Gui/Views/Scroll/ScrollButton.cs
BDisp: Agree. Thanks.

Terminal.Gui/Views/Scroll/ScrollBar.cs
BDisp: Done.

Terminal.Gui/Views/Scroll/Scroll.cs
BDisp: Done. Thanks.

Terminal.Gui/Views/Scroll/ScrollButton.cs
tig: I know that. For correctness tho...

tig:
> @tig do you remember any change that may cause unit test failing on macOS? It's also failing in the v2_develop branch.
> 
> ```
> [xUnit.net 00:01:26.21]     UICatalog.Tests.ScenarioTests.All_Scenarios_Quit_And_Init_Shutdown_Properly(scenarioType: typeof(UICatalog.Scenarios.DatePickers)) [FAIL]
> [xUnit.net 00:01:26.21]       'Date Picker' failed to Quit with Esc after 1500ms and 0 iterations. Force quit.
> [xUnit.net 00:01:26.21]       Stack Trace:
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/UnitTests/UICatalog/ScenarioTests.cs(109,0): at UICatalog.Tests.ScenarioTests.<>c__DisplayClass3_0.<All_Scenarios_Quit_And_Init_Shutdown_Properly>g__ForceCloseCallback|1()
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/Terminal.Gui/Application/MainLoop.cs(379,0): at Terminal.Gui.MainLoop.RunTimers()
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/Terminal.Gui/Application/MainLoop.cs(269,0): at Terminal.Gui.MainLoop.RunIteration()
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/Terminal.Gui/Application/Application.Run.cs(587,0): at Terminal.Gui.Application.RunIteration(RunState& state, Boolean& firstIteration)
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/Terminal.Gui/Application/Application.Run.cs(561,0): at Terminal.Gui.Application.RunLoop(RunState state)
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/Terminal.Gui/Application/Application.Run.cs(439,0): at Terminal.Gui.Application.Run(Toplevel view, Func`2 errorHandler)
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/UICatalog/Scenarios/DatePickers.cs(23,0): at UICatalog.Scenarios.DatePickers.Main()
> [xUnit.net 00:01:26.21]         /Users/runner/work/Terminal.Gui/Terminal.Gui/UnitTests/UICatalog/ScenarioTests.cs(48,0): at UICatalog.Tests.ScenarioTests.All_Scenarios_Quit_And_Init_Shutdown_Properly(Type scenarioType)
> [xUnit.net 00:01:26.21]            at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
> [xUnit.net 00:01:26.21]            at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
> [xUnit.net 00:01:26.21]       Output:
> [xUnit.net 00:01:26.21]         Running Scenario 'UICatalog.Scenarios.DatePickers'
> [xUnit.net 00:01:26.21]         Initialized 'Terminal.Gui.CursesDriver'
> [xUnit.net 00:01:26.22]   Finished:    UnitTests
> [xUnit.net 00:01:26.22] Canceling due to test failure..
> ```

Seems ok now?

BDisp:
> Seems ok now?

Yes, did you do something to fix it?

tig:
> > Seems ok now?
> 
> Yes, did you do something to fix it?

I don't think so

BDisp:
I already merged your PR #3728. Thanks for the great and good changes.

BDisp:
@tig I think the scroll bar in the `CharacterMap` is a good pretext for you think better about the mouse click captured when you are moving the slider with the mouse and then release, by selecting the object where the mouse was released. In my opinion this is a bad behavior because it's doing something that we didn't intent to do it.

BDisp:
@tig do you understand why this strange behavior happens? The slider grabbed the mouse and is scrolling down but the decrease button is getting the `OnMouseEnter` and it shouldn't.

WindowsTerminal_PvgN1o3kHh

Terminal.Gui/Views/Scroll/Scroll.cs
```diff
@@ -0,0 +1,272 @@
...
+    /// <inheritdoc/>
+    public override void EndInit ()
+    {
+        base.EndInit ();
+
+        AdjustScroll ();
+    }
+
+    /// <summary>Get or sets if the view-port is kept in all visible area of this <see cref="Scroll"/></summary>
+    public bool KeepContentInAllViewport
```
tig, line 47: I'm trying to understand this.

In v1 there was a similarly named construct that was super confusing to me. 

In the ContentScrolling work I did, I thought I addressed the user scenario that drove this with 

```cs
/// <summary>
///     If set, <see cref="View.Viewport"/><c>.X</c> can be set values greater than <see cref="View.GetContentSize ()"/>
///     <c>.Width</c> enabling scrolling beyond the right
///     of the content area.
/// </summary>
/// <remarks>
///     <para>
///         When not set, <see cref="View.Viewport"/><c>.X</c> is constrained to <see cref="View.GetContentSize ()"/>
///         <c>.Width - 1</c>.
///         This means the last column of the content will remain visible even if there is an attempt to scroll the
///         Viewport past the last column.
///     </para>
///     <para>
///         The practical effect of this is that the last column of the content will always be visible.
///     </para>
/// </remarks>
AllowXGreaterThanContentWidth = 4,

/// <summary>
///     If set, <see cref="View.Viewport"/><c>.Y</c> can be set values greater than <see cref="View.GetContentSize ()"/>
///     <c>.Height</c> enabling scrolling beyond the right
///     of the content area.
/// </summary>
/// <remarks>
///     <para>
///         When not set, <see cref="View.Viewport"/><c>.Y</c> is constrained to <see cref="View.GetContentSize ()"/>
///         <c>.Height - 1</c>.
///         This means the last row of the content will remain visible even if there is an attempt to scroll the Viewport
///         past the last row.
///     </para>
///     <para>
///         The practical effect of this is that the last row of the content will always be visible.
///     </para>
/// </remarks>
AllowYGreaterThanContentHeight = 8,
```

Specifically:

>  The practical effect of this is that the last column of the content will always be visible.

Can you verify we are talking about the same end-user scenario:

"As a user, when using a ScrollBar to scroll through content, the last row/column of the content never scrolls beyond the visible area."

BDisp: > Specifically:
> 
> > The practical effect of this is that the last column of the content will always be visible.
> 
Correct.

> Can you verify we are talking about the same end-user scenario:
> 
I think so.

> "As a user, when using a ScrollBar to scroll through content, the last row/column of the content never scrolls beyond the visible area."

Correct. But imagine if a user want to use a scroll bar without a `View`. But I may not being dealing with this right. What do you propose?

BDisp:
@tig [this](https://github.com/gui-cs/Terminal.Gui/pull/3498/commits/6cf29c6afd10a961d2296fa6da78208aaa60db86) is how I pretend to implement the scroll bars into the `View`.

BDisp:
> @tig [this](https://github.com/gui-cs/Terminal.Gui/pull/3498/commits/6cf29c6afd10a961d2296fa6da78208aaa60db86) is how I pretend to implement the scroll bars into the `View`.

Damn, I committed without run the unit tests and now I see that instantiate the scroll bars on the initialization of the `View` will increase the number of view not yet disposed. Do you know the best way to workaround this?

tig:
> > @tig [this](https://github.com/gui-cs/Terminal.Gui/pull/3498/commits/6cf29c6afd10a961d2296fa6da78208aaa60db86) is how I pretend to implement the scroll bars into the `View`.
> 
> Damn, I committed without run the unit tests and now I see that instantiate the scroll bars on the initialization of the `View` will increase the number of view not yet disposed. Do you know the best way to workaround this?

Haha! I was literally working on same thing this morning. I was up late last night thinking about this ;-).

See how I used `Lazy<T>` in my PR....

BDisp:
I think it's working much better now. When you have time please test it. Thanks.

Terminal.Gui/Views/Scroll/Scroll.cs
tig: > Correct. 

> But imagine if a user want to use a scroll bar without a View. But I may not being dealing with this right. What do you propose?

If someone wants to use a scroll bar as a standalone View to control another View, and they want to ensure some content never gets scrolled out of visibility, they can code that up as needed. There's no need for ScrollBar to have that functionality.

I propose we don't have a `KeepContentInViewport`-like property. `AllowYGreaterThanContentHeight` should work for most mainline scenarios.

tig:
> image
> 
> I would like to maintaining the `ShowScrollIndicator` property because it will indicates that the visibility is false due the `AutoHideScrollBar` is true (visible if size is greater than the content size or invisible otherwise) and thus this check `get => _showScrollIndicator && Visible;` is crucial to to get the real state if the scroll visibility. If you don't agree please let me know.

I think `ShowScrollIndicator` is misnamed. "Indicator" should be "Bar", and "Show" is a verb, not a noun, so it's very confusing.

I do think you're correct that we need something more than just using `Visible` as I did in my prototype.

Here's a suggestion:

Add to `ViewportSettings`:

```cs
public enum ViewportSettings
{
...
    /// <summary>
    ///     If set, the vertical scroll bar (see <see cref="View.HorizontalScrollBar"/>) will be enabled and automatically made visible
    ///     when the dimension of the <see cref="View.Viewport"/> is smaller than the dimension of <see cref="View.GetContentSize()"/>.
    /// </summary>
    EnableHorizontalScrollBar = 64,

    /// <summary>
    ///     If set, the vertical scroll bar (see <see cref="View.VerticalScrollBar"/>) will be enabled and automatically made visible
    ///     when the dimension of the <see cref="View.Viewport"/> is smaller than the dimension of <see cref="View.GetContentSize()"/>.
    /// </summary>
    EnableVerticalScrollBar = 128,

    /// <summary>
    ///     If set, the horizontal and vertical scroll bars (see cref="View.HorizontalScrollBar"/> and <see cref="View.VerticalScrollBar"/>)
    ///     will be enabled and automatically made visible when the dimension of the <see cref="View.Viewport"/> is smaller than the
    ///     dimension of <see cref="View.GetContentSize()"/>.
    /// </summary>
    EnableScrollBars = EnableHorizontalScrollBar | EnableVerticalScrollBar
```

This moves the automatic behavior out of `ScrollBar` and into `View` and encapsulates that logic in `View.ScrollBars.cs` and consolodates content scrolling functionality.

If these flags are NOT set, then the developer can use `view.Horizontal/VerticalScrollbar = true/false` to manage visibility themselves. 

What do you think of this idea?

tig:
Can I just say how much I love the work @BDisp is doing in this PR? Love. Love. Love.

tig:
@BDisp see my comment here:

- https://github.com/gui-cs/Terminal.Gui/issues/464

When you nuke the old ScrollView etc..., can you please take care of `IsOverriden` too?

tig:
@BDisp what's the status of this PR? 

In #3761 I'm fighting with the old hacky`ScrollView` and `ScrollBarView` code that does a lot of unnatural things w.r.t. drawing and layout. Instead of spending time & energy fixing those things (and views that depend on them), i'd much rather have us spend time getting the new classes working well.

Thanks.

BDisp:
> @BDisp what's the status of this PR?
> 
> In #3761 I'm fighting with the old hacky`ScrollView` and `ScrollBarView` code that does a lot of unnatural things w.r.t. drawing and layout. Instead of spending time & energy fixing those things (and views that depend on them), i'd much rather have us spend time getting the new classes working well.
> 
> Thanks.

Sorry, I never worked on it again. I'm busy with some other stuff. But I'll get back to it as soon as I can.

tig:
> > @BDisp what's the status of this PR?
> > In #3761 I'm fighting with the old hacky`ScrollView` and `ScrollBarView` code that does a lot of unnatural things w.r.t. drawing and layout. Instead of spending time & energy fixing those things (and views that depend on them), i'd much rather have us spend time getting the new classes working well.
> > Thanks.
> 
> Sorry, I never worked on it again. I'm busy with some other stuff. But I'll get back to it as soon as I can.

No problem. I'm working on it right now on top of the branch I'm using for #3798 .

BDisp:
> No problem. I'm working on it right now on top of the branch I'm using for #3798 .

Thanks for the help. Then I'll focused on others things, ok?

tig:
Yep. Love the driver work you're doing!

BDisp:
I've a branch where I already removed the old scroll bars and starting the implementation with the views that were use them. But a long time ago I already have done that before and I had some tedious working on this again. But I'll see if I can reanime again to work with it.

tig:
New PR based on this great work: #3811.

Closing this one.

PR #3502, opened 2024-05-27
Fixes #3497. Removes `LayoutStyle`

### LINKED ISSUE #3479: V2 Beta Release
Per [this discussion](https://github.com/gui-cs/Terminal.Gui/discussions/1940#discussioncomment-9421520), this is the master Issue for the **Terminal.Gui v2 Beta Release**.  

We will close this issue once all exit criteria for the v2 Beta have been met, and the nuget packages are published.

### LINKED ISSUE #3497: `LayoutStyle` is no longer needed - remove it
In v1 there was a big distinction between Absoulte and Computed. In v2 this has been significantly simplified and there's really only one "layout mode/style": Computed/Relative. A dev can choose to use `Pos/Dim.Absolute` if they want, but it's just another form of computed.

To this end, we'll remove `LayoutStyle` and `View.LayoutStyle`.

Other than `AllViewsTester` and a bunch of (old, ugly) unit tests it's not used anywhere anymore.

- [ ] Remove `LayoutStyle.cs"
- [ ] Remove `View.Layout`
- [ ] Updated `AllViewsTester`
- [ ] Update Unit tests
- [ ] Simplify API docs including conceptual docs.

tig:
In v1 there was a big distinction between Absoulte and Computed. In v2 this has been significantly simplified and there's really only one "layout mode/style": Computed/Relative. A dev can choose to use `Pos/Dim.Absolute` if they want, but it's just another form of computed.

(Note: Branch is mis-named: Should be `v2_3497-DeleteLayoutStyle`)

## Fixes

- Fixes #3497

## Todo

- [x] Remove `LayoutStyle.cs"
- [x] Remove `View.Layout`
- [x] Updated `AllViewsTester`
- [x] Update Unit tests
- [x] Simplify API docs including conceptual docs.

## Pull Request checklist:


dodexahedron:
Going throigh my queue now. Will be looking at this next.

Terminal.Gui/View/Layout/Dim.cs
```diff
@@ -7,7 +7,7 @@ namespace Terminal.Gui;
 ///     <para>
 ///         A Dim object describes the dimensions of a <see cref="View"/>. Dim is the type of the
 ///         <see cref="View.Width"/> and <see cref="View.Height"/> properties of <see cref="View"/>. Dim objects enable
-///         Computed Layout (see <see cref="LayoutStyle.Computed"/>) to automatically manage the dimensions of a view.
+///         automatic management of the dimensions of a view.
```
dodexahedron, line 10: I kinda think that whole sentence should just go, since it's almost a repetition of the previous sentence.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -148,8 +145,7 @@ public virtual Point ScreenToFrame (in Point location)
     ///         <see cref="LayoutSubview(View, Size)"/> and <see cref="OnDrawContent(Rectangle)"/> methods to be called.
     ///     </para>
     ///     <para>
-    ///         Changing this property will cause <see cref="Frame"/> to be updated. If the new value is not of type
-    ///         <see cref="PosAbsolute"/> the <see cref="LayoutStyle"/> will change to <see cref="LayoutStyle.Computed"/>.
+    ///         Changing this property will cause <see cref="Frame"/> to be updated.
```
dodexahedron, line 148: This is an unrelated note for thoughts on future work:

It would probably be a good idea to implement INotifyPropertyChanged and react to that here, as well as the other stuff it already does.

Currently, we have more general events for the layout pipeline, which is cool for what they do, but not very useful when all I want to do is, for example, react to only one property changing.

I've already got that in my list of event-related stuff to do, but just pointing out here since I had the thought and this is a place that would be raised.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -858,12 +778,6 @@ internal void CollectAll (View from, ref HashSet<View> nNodes, ref HashSet<(View
         foreach (View? v in from.InternalSubviews)
```
dodexahedron, line 778: I know it wasn't part of this change set, but it's something simple that doesn't need a whole issue to track:

The body of this loop needs to guard against nulls.

Also, we should use a different variable name, instead of `from`, as that's a keyword for C# LINQ statements.

Also a big thread safety problem here, but that's a more ubiquitous thing that _will_ need dedicated issues, thought, and work to handle appropriately.

Terminal.Gui/Views/Button.cs
```diff
@@ -33,8 +33,8 @@ public class Button : View
     private readonly Rune _rightDefault;
     private bool _isDefault;
 
-    /// <summary>Initializes a new instance of <see cref="Button"/> using <see cref="LayoutStyle.Computed"/> layout.</summary>
-    /// <remarks>The width of the <see cref="Button"/> is computed based on the text length. The height will always be 1.</remarks>
+    /// <summary>Initializes a new instance of <see cref="Button"/>.</summary>
+    /// <remarks>The width and height of the <see cref="Button"/> is computed based on the text.</remarks>
```
dodexahedron, line 37: Could use the actual property names in a `<see />` element if you feel like it.

Terminal.Gui/Views/CheckBox.cs
```diff
@@ -11,8 +11,7 @@ public class CheckBox : View
     private bool? _checked = false;
```
dodexahedron, line 11: Hmm. Why is this nullable? For tri-state?

Terminal.Gui/Views/Dialog.cs
```diff
@@ -55,13 +55,11 @@ public class Dialog : Window
...
-    ///     Initializes a new instance of the <see cref="Dialog"/> class using <see cref="LayoutStyle.Computed"/>
-    ///     positioning with no <see cref="Button"/>s.
+    ///     Initializes a new instance of the <see cref="Dialog"/> class.
     /// </summary>
     /// <remarks>
-    ///     By default, <see cref="View.X"/> and <see cref="View.Y"/> are set to <c>Pos.Center ()</c> and
-    ///     <see cref="View.Width"/> and <see cref="View.Height"/> are set to <c>Width = Dim.Percent (85)</c>, centering the
-    ///     Dialog vertically and horizontally.
+    ///     By default, <see cref="View.X"/>, <see cref="View.Y"/>, <see cref="View.Width"/>, and <see cref="View.Height"/> are set
+    ///     such that the dialog will be centered and no larger than 90% of the container's width and height.
```
dodexahedron, line 62: How about this wording to make it parse easier:
```csharp
/// <remarks>
/// By default, <see cref="View.X"/>, <see cref="View.Y"/>, <see cref="View.Width"/>, and <see cref="View.Height"/> are set such that the <see cref="Dialog" /> will be centered in its containing <see cref="View" /> and no larger than 90% of the containing <see cref="View" />'s <see cref="View.Width" /> and <see cref="View.Height" />.
/// </remarks>
```
However the reflow will line it up of course. I just put it on one line here so it'll reformat it to fit if pasted.

But I also am wondering if maybe we should be dogfooding the API's terminology here, as well?

Specifically, instead of talking about a containing view or a container, should we be saying `<see cref="SuperView" />`?

Terminal.Gui/Views/Toplevel.cs
```diff
@@ -397,9 +397,7 @@ public virtual void PositionToplevel (Toplevel top)
         }
 
         if ((superView != top || top?.SuperView is { } || (top != Application.Top && top.Modal) || (top?.SuperView is null && top.IsOverlapped))
-
-            // BUGBUG: Prevously PositionToplevel required LayotuStyle.Computed
-            && (top.Frame.X + top.Frame.Width > maxWidth || ny > top.Frame.Y) /*&& top.LayoutStyle == LayoutStyle.Computed*/)
+            && (top.Frame.X + top.Frame.Width > maxWidth || ny > top.Frame.Y))
         {
             if ((top.X is null || top.X is PosAbsolute) && top.Frame.X != nx)
```
dodexahedron, line 402: X can't be null, can it?

I think a nullability pass on this file is probably in order at some point.

dodexahedron (changes requested):
Looks pretty good to me.

Only a couple of relevant comments, and they're pretty small.

The others are just notes and questions for later stuff.

Terminal.Gui/Views/CheckBox.cs
BDisp: Yes.

dodexahedron: Ok cool.

Suggestion, then: Use an enum, like how WPF does it.

The main reason is that a nullable boolean (`Nullable<bool>`) is more cumbersome to work with than an enum, and is also more costly (cost is a fairly minor concern though).

Here is the source code for that enum as it exists in WPF (except I removed the ` : byte` because that's not helpful for us):

```csharp
    /// <summary>
    /// Visibility - Enum which describes 3 possible visibility options.
    /// </summary>
    public enum Visibility
    {
        /// <summary>
        /// Normally visible.
        /// </summary>
        Visible = 0,
 
        /// <summary>
        /// Occupies space in the layout, but is not visible (completely transparent).
        /// </summary>
        Hidden,
 
        /// <summary>
        /// Not visible and does not occupy any space in layout, as if it doesn't exist.
        /// </summary>
        Collapsed
    }
```

Optional of course.

BDisp: But the `_checked` field is for a tri-state `Boolean` and not for the `Visibility`. It only indicate if a `CheckBox` is `true`, `false` or `null` indicating that is not defined yet.

dodexahedron: Oh whoops yes sorry I don't know why my head went to Visibility out of nowhere. 😅 

There's also a nice property in WPF for Checked which is `IsThreeState`. It defaults to false, and the checked boolean is exactly as you have it - a nullable bool - and its default value is false as well.

Again - totally optional, though a nice and expressive property.

BDisp: Yes, an enum could actually be used for this. It's up to @tig :-)

dodexahedron: Well _this_ one is a `bool?` in WPF, so I'm not interested in that part any more. Sorry I wasn't clear.

The only remaining very optional suggestion I had for this part was adding in that `IsThreeState` bool.

dodexahedron: Just for clarity:

IsThreeState isn't simply informational. It changes the behavior of a checkbox for whether it will _allow_ setting null at runtime.

dodexahedron: But, thinking about that, it's probably more appropriate for that to be in a different PR anyway, if we decide to add it. It's not related to this.

Terminal.Gui/View/Layout/ViewLayout.cs
tig: Theoretically yes. But this code has existed in TG since the beginning of time and there's never been a null issue. 

Threading yes.

Suggest a better name for `from` and I'll change it.

Terminal.Gui/Views/Dialog.cs
tig: Actually, for Modal's there is NO containing View. Currently they use `Driver.Screen` to bound them. @BDisp  and I are debating whether Modals can have a SuperView or not elsewhere. 

I have an instinct that there are use-cases where an app would want this, but am struggling to come up with a real example. Therefore, we'll leave it as-is for now and I'll change this API doc to say "Screen".

Terminal.Gui/Views/Toplevel.cs
tig: No. `X` and `Y` cannot be null.

image

Terminal.Gui/View/Layout/ViewLayout.cs
dodexahedron: > Suggest a better name for `from` and I'll change it.

Hmm.  Maybe `source` or something of that nature?

dodexahedron: In the end, it's a method parameter, so it's about a 0.1 on a scale of 0 to 10 for importance. 🤷‍♂️

If it stays `from`, though, it should probably be prefixed with `@` for formality.

Terminal.Gui/Views/Dialog.cs
dodexahedron: One use case is for windows which, themselves, want to have a modal popup, such as in an MDI-style app.

I was actually originally doing something like that in SnapsInAZfs, but went another direction since the means of achieving it in v1 would have required a lot of code around cases I wasn't interested in trying to predict. 😅

dodexahedron: That's probably something we could do in a feature release, though, rather than taking the time to do it in V2 RTM, isn't it?

PR #3507, opened 2024-05-27
Fixes #3503. Improves "View Editing" for Scenarios (e.g. `AdornmentsEditor`)

### LINKED ISSUE #3503: `AdornmentsEditor` does not size correctly
Oops. I broke this in the Pos.Align PR. 

I will fix this in 

- #3376

tig:
## Fixes

- Fixes #3503

## Proposed Changes/Todos

- [x] Fix #3503 - `Dim.Auto` not working correctly, causing `AdornmentsEditor` to be sized incorrectly.
- [x] Updates `Border` to have an explicit `Border.ShowTitle` property. And adjusts line rendering.
- [x] Refactor `AdornmentEditor` and `AdornmentsEditor` to be more flexible.
- [x] Add prototype expand/collapse functionality (`ExpanderButton`) - See #2814
- [x] Add `AdornmentsEditor` to `AllViewsTester`

## For a future PR

(I don't want to do these now as I'd like to get what's here merged because it fixes #3503).

- [ ] Extract the "Position Editor" ("Settings") from `AllViewsTester` into re-usable class like `AdornmentEditor`
- [ ] Move diagnostic flags settings out of `AdornmentsEditor` into a separate "editor"
- [ ] Add `ColorSchemeEditor` 

## Pull Request checklist:


tig:
An illustration:

GlcffUj 1

BDisp (approved):
Fantastic. Thanks.

Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -94,17 +95,17 @@ internal override int Calculate (int location, int superviewContentSize, View us
...
+                                                           // && v.X is not PosCenter
+                                                           && v.Width is not DimFill).ToList ();
                 }
                 else
                 {
-                    subviews = us.Subviews.Where (v => v.Y is not PosAnchorEnd
-                                                       && v.Y is not PosAlign
-                                                       && v.Y is not PosCenter
-                                                       && v.Height is not DimFill).ToList ();
+                    subviews = includedSubviews.Where (v => v.Y is not PosAnchorEnd
```
dodexahedron, line 105: Not for this PR, but definitely needed for performance and resource usage at some point:

We should make things around collections way lighter than they are such as here.

The power of LINQ shines when it is able to defer evaluation til materializing is actually necessary.

Every time we call ToArray or ToList, the loop is evaluated, which means that, unconditionally, a new collection is constructed, allocated, and copied from the source, and that's VERY expensive.

Especially anything that happens in the render pipeline should be as lightweight as possible, as that code is all very hot.

There are various strategies to help with this, and no single strategy is appropriate for all uses.

Some examples:
 - Using `ArraySegment<T>`s when contiguous ranges of a source array are needed.
 - Using `Memory<T>` or `Span<T>` over collections, as and when appropriate for each of those.
   - Using slices, when using either of those, for pretty much the same reason one would use `ArraySegment<T>` without them.
 - Never calling .ToList() or similar until the very last moment, and only when that's because you now both need to enumerate the data and then operate on the subset ***and*** are also reasonably certain that multiple enumeration would happen if you don't do it, ***and*** that that multiple enumeration is actually more expensive than allocating a new collection (the cutoff for that is situational but often pretty early).

Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -94,17 +95,17 @@ internal override int Calculate (int location, int superviewContentSize, View us
...
+                                                           && v.Y is not PosAlign
```
dodexahedron, line 106: Hm. Looks like this might be eligible for a pattern match. If it can't automatically make it for you, don't worry about it. But if you put the cursor on one of the `&&`s, it should be available in the lightbulb menu, if it can see how to do it.

Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -143,38 +144,42 @@ internal override int Calculate (int location, int superviewContentSize, View us
...
-                int maxCenter = 0;
-                for (var i = 0; i < subviews.Count; i++)
-                {
-                    View v = subviews [i];
-                    maxCenter = dimension == Dimension.Width ? v.Frame.Width : v.Frame.Height;
-                }
-
-                subviewsSize += maxCenter;
-                #endregion Center
+                //#region Center
```
dodexahedron, line 147: This `#region` is pretty unnecessary and just makes the code not sort as expected.

Terminal.Gui/View/Layout/ViewLayout.cs
```diff
@@ -315,6 +315,29 @@ public Pos Y
...
+    public bool ExcludeFromLayout
+    {
+        get => _excludeFromLayout;
+        set
+        {
+            if (value == _excludeFromLayout)
+            {
+                return;
+            }
+            _excludeFromLayout = value;
```
dodexahedron, line 336: Good place to raise `PropertyChanged`

dodexahedron: Also, if we either subscribe to the event internally or use an invoker wrapper, the calls to `SetNeedsLayout` in all these ***could*** move into that and would just be replaced by raising the event, in the property handlers.

Or... Getting fancier...

If there were a non-public event (private protected is probably appropriate) called something like "LayoutInvalidated" or whatever makes sense to you, these calls to `SetNeedsLayout()` could be replaced with raising that event, and use `[CallerMemberName]` and the `sender` to communicate who asked and why, as well as enable components to opt in and out of participating if/as needed.

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -236,6 +236,16 @@ public void Draw ()
 
         OnRenderLineCanvas ();
 
+        // TODO: This is a hack to force the border subviews to draw.
+        if (Border?.Subviews is { })
```
dodexahedron, line 240: This made me go look at the TreeView stuff.

I now have more things on my list to bring up later. 😄

Terminal.Gui/View/ViewDrawing.cs
```diff
@@ -365,7 +375,7 @@ public virtual Attribute GetNormalColor ()
     {
         ColorScheme cs = ColorScheme;
 
-        if (ColorScheme is null)
+        if (cs is null)
```
dodexahedron, line 378: Any block like this:
```csharp
if( cs is null )
{
  cs = new(); // Any assignment - not just new()
}
```

should be this:

```csharp
cs ??= new();
```

[Null coalescing assignment](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/null-coalescing-assignment).

It is exactly the same, without the extra cyclomatic complexity and, if the compiler can reason about what happens above, it may be able to optimize away the check in some cases.

dodexahedron (commented):
Mostly just commentary/discussion prompts for additional work we should do.

PR #3509, opened 2024-05-27
Fixes #3501. `ContentSize` -> `GetContentSize()`/`SetContentSize()`

### LINKED ISSUE #3501: ScrollView.ContentSize no setter
**Describe the bug**

TGD allows you to design properties on views.  For ScrollView this includes `ContentSize`.  This functionality relies on both a public `get;` and `set;`

It seems the setter no longer exists for this property.  Instead there is a method.

image

Is it possible to revert this back to being a property?

```
yield return this.CreateProperty(nameof(ScrollView.ContentSize));
```
_TGD code that advertises ContentSize as designable_

tig:
## Fixes

- Fixes #3501

## Proposed Changes/Todos

- [x] Change `ContentSize` back to a `get;set;` property
- [x] Add `public bool ContentSizeTracksViewport` to indicate whether `ContentSize` was explicilty set or not
- [x] Go back to having separate methods.
- [x] Update API docs
- [x] Adjust unit tests.

## Pull Request checklist:


tig:
@tznind - I just had a hottub thought. 

`set_ContentSize` should be `protected` - Setting it is an implementation detail of the View and thus code outside the View should not be able to set it. The `Content Scrollilng` Scenario DOES set it, but that's because I put the code that sets it in the VIew's Padding for funzies...I'll need to change that.

So... I do NOT think this PR is good-to-go as-is. 

I can't think of any reason TGD needs to set `ContentSize` and thus it should be ignored by it. Am I missing something?

Of course, my comment about regarding the legacy-soon-to-be-nuked `ScrollView` still applies.

dodexahedron:
> I'm going to get rid of it being a property and have `public Size GetContentSize ()` and `internal Size SetContentSize ()` (for tests).

Or, you could keep it a property and just make the setter `protected internal`, `internal`, or `private protected`, as appropriate.

PR #3510, opened 2024-05-28
v2 Fixes #2897. ListView ListWrapper - marking does work if you give an IList which in which the count changes

### LINKED ISSUE #2897: ListView ListWrapper - marking does work if you give an IList which in which the count changes
The symptoms are that if I use listView.SetSource with an IList with an initial count of zero, when I later add items to the list, listView.SetMark() does do anthing at all.

To reproduce it, you can simply try giving listView.SetSource an IList with zero count, then add some lines to it.  Then try marking items.

Here's why I think it is happening:

ListView.SetSource takes an IList object.  I need to dynamically add/remove things from the list, so I wrote my own IList object.  However, the BitArray 'marks' is constructed once when the ListWrapper is constructed, and thus has a fixed size.   Since my IList object is of zero count at startup, SetMark just doesn't work at all.

BDisp:
## Fixes

- Fixes #2897

## Proposed Changes/Todos

- [x] Added CollectionChanged event to the IListDataSource interface
- [x] Change ListWrapper to use ObservableCollection instead of the IList
- [x] Change the scenarios to use the changed implementation
- [x] Change the unit tests to use the changed implementation

## Pull Request checklist:


BDisp:
To test, you can copy and paste this code into a scenario.

```cs
    public override void Main ()
    {
        // Init
        Application.Init ();

        Window appWindow = new ()
        {
            Title = $"{Application.QuitKey} to Quit - Scenario: {GetName ()}"
        };

        var count = 0;
        var source = new ObservableCollection<string> ();

        var listView = new ListView
        {
            Width = 20,
            Height = Dim.Fill (),
            AllowsMarking = true,
            AllowsMultipleSelection = true,
            Source = new ListWrapper<string> (source)
        };
        appWindow.Add (listView);

        var btnAdd = new Button
        {
            X = Pos.Center (),
            Y = Pos.Center (),
            Text = "_Add Item"
        };

        btnAdd.Accept += (s, e) =>
                         {
                             source.Add ($"Item{count}");
                             count++;
                             listView.MoveEnd ();
                         };
        appWindow.Add (btnAdd);

        var btnRemove = new Button
        {
            X = Pos.Center (),
            Y = Pos.Center () + 1,
            Text = "_Remove Item"
        };

        btnRemove.Accept += (s, e) =>
                            {
                                if (source.Count > 0 && listView.SelectedItem > -1)
                                {
                                    source.Remove (source [listView.SelectedItem]);
                                }
                            };
        appWindow.Add (btnRemove);

        Application.Run (appWindow);
        appWindow.Dispose ();

        Application.Shutdown ();
    }

```

@tig the scrolling in the `ListView` is broken. The `MoveEnd` method is now using the code `Viewport = Viewport with { Y = _selected };` but the `Y` does not change.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -16,7 +16,7 @@ public class ComboBox : View
     private readonly ComboListView _listview;
     private readonly int _minimumHeight = 2;
     private readonly TextField _search;
-    private readonly IList _searchset = new List<object> ();
+    private readonly ObservableCollection<object> _searchSet = [];
```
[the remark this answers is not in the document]
BDisp: > A type like `IEnumerable<T>` or `IList<T>` is much more friendly.

But none of them also raise a collection change notification.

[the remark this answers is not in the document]
dodexahedron: I know. You have to do it manually.

But that's the consumer's responsibility - not ours.

If _we_ need to react to collection changes, then the collection shouldn't even be exposed to the consumer, because they can circumvent it.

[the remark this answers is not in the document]
dodexahedron: Ultimately, the key is that second sentence - it's their responsibility.

We don't know, ahead of time, when or if they want things to happen.

What if they want to make several changes to the collection before redrawing?

What if they don't want to re-draw at all?

What if they want to re-draw when a property of something _in_ the collection changes?

There are pretty much infinite possibilities, so we shouldn't try to end-run the consumer.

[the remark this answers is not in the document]
dodexahedron: And that makes it easier on us, anyway. :)

BDisp:
This is the current behavior.

WindowsTerminal_YvE1tHMrhR

Terminal.Gui/Views/ListView.cs
```diff
@@ -264,6 +271,11 @@ public IListDataSource Source
             }
             _source = value;
 
+            if (_source is { })
```
dodexahedron, line 274: This has the potential to cause multiple subscriptions.

Terminal.Gui/Views/ListView.cs
```diff
@@ -501,7 +527,12 @@ public virtual bool MoveEnd ()
 
             if (Viewport.Y + _selected > Viewport.Height - 1)
             {
-                Viewport = Viewport with { Y = _selected };
+                Viewport = Viewport with
```
dodexahedron, line 530: Aren't we using a different term than "Viewport" now?

I've lost track.

dodexahedron (changes requested):
Noticed a multiple-subscription issue.

Terminal.Gui/Views/ListView.cs
BDisp: It's only check for null value here. It only subscribe if not null and it's unsubscribe the previous.

Terminal.Gui/Views/ListView.cs
BDisp: Now we are using `Viewport`. Before was `Bounds`.

Terminal.Gui/Views/ListView.cs
dodexahedron: That's a check for not null, not a check for null.

Terminal.Gui/Views/ListView.cs
```diff
@@ -843,35 +879,64 @@ public Task SetSourceAsync (IList source)
...
     /// <inheritdoc/>
-    public int Length { get; }
+    public int Length { get; private set; }
+
+    void CheckAndResizeMarksIfRequired ()
+    {
+        if (_source != null && _count != _source.Count)
+        {
+            _count = _source.Count;
+            BitArray newMarks = new BitArray (_count);
```
dodexahedron, line 930: BitArray is _extremely_ expensive for what it does and involves a ton of extra heap activity with multiple copies at runtime.

It's good for compact storage - not for performance.

You'd get better performance with a literal List of bools or, better yet, since this is ephemeral, a stackalloced span of bools.

dodexahedron (changes requested):
bit performance hit here

Terminal.Gui/Views/ListView.cs
dodexahedron: There are also a lot of concurrency problems with the way this works as-is.

dodexahedron: Typically, ListView type controls are implemented via having the collection be a collection of items implementing a known interface, rather than a raw reference to a user-supplied collection.

That's how it's done in WPF and WinForms, for example.

Otherwise, the concurrency issues become a mess and it's hard to maintain good performance.

dodexahedron: I would say we should handle it like this:

Go ahead and merge this, just to give it the same behavioral starting point as the same change in v1.

But immediately file a new issue to address these problems, for v2.

Terminal.Gui/Views/ListView.cs
BDisp: I said wrongly for null but I meant for not null. You are right. I'm unsubscribing on the Dispose method and I need to do it here before subscribe the new one.

Terminal.Gui/Views/ListView.cs
BDisp: But this PR is for V2. For V1 it will be a breaking change. It will be better you explain with real or pseudo code.

BDisp: > There are also a lot of concurrency problems with the way this works as-is.

I'm only help with a lack of change collection event.

BDisp: If it's for letting as it was before then it's better leave to the user handle the necessary actions when he is manipulating the collection?

Terminal.Gui/Views/ListView.cs
BDisp: I think I fixed this with the 893a6c8c322816e72d7d3093ebc8197d7c15574a commit with unit tests that is working as expected. Only the `ListWrapper` managed outside the `ListView` the user is responsible to dispose it which will unsubscribe the event. Thanks for to alert me with this.

Terminal.Gui/Views/ComboBox.cs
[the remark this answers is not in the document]
BDisp: I am aware of your comments and subscribe. But as I already explained, I just wanted to provide an event for when a source collection is changed so that the user can take the appropriate actions. If the source has to be manipulated beforehand and the user doesn't want to be bombarded with a bunch of events, he can first set the source to null and manipulate the data he wants and then go back to defining the already changed source.
Searching on Google about the best way to have a collection that notifies the client about changes, what they most advise to use is the `ObservableCollection<T>` or create a new one. If one already exists that works well, the user can create a collection that is compatible with IList<T> and only when defining the Source property must it be created as `new ListWrapper<string> (new ObservableCollection<string> (source))`.

Terminal.Gui/Views/ListView.cs
BDisp: It was always using `BitArray` from the beginning and `List<bool>` for some reason was not used. `Span<bool>` cannot be declared, as it gives the error below. I don't think it's worth changing this because `_marks` isn't even part of `_source`. The fact that it is handled in the `ListWrapper` does not mean that the `_source` changes. They are independent of each other. The only value `_marks` needs is `_source.Count`.

`CS8345 Field or self-implemented property cannot be of type 'Span<bool>' unless it is an instance member of a ref struct.`

dodexahedron: That's a red herring. Don't worry about the Span stuff for now. That has specific rules to follow, by virtue of being a ref struct, and pretty thorough understanding of how `Span<T>` and `Memory<T>` work is necessary to do that stuff without losing your sanity. 😅 

The best or at least most straightforward option with the fewest caveats is to wrap the collection, or else there's going to be a ton of work to do for concurrency on top of it.

The last suggestion I made was basically just to make them be the same and then we can fix it from there. That's mostly because my attention is on other things at the moment and I don't have the bandwidth to provide specific consultation on this particular part, right now, or else I'd be more than happy to give more detail or write something up myself. :)

tig (approved):

Issue #3023, opened 2023-11-30
Enable overriding default Key Bindings with `ConfigurationManager`


tig:
Terminal.Gui provides default key bindings, but those defaults are not configurable by the user. 

`ConfigurationManager` should allow users to redefine key bindings for the system, a user, or an application.

## All built-in view subclasses should use ConfigurationManger to specify the default keybindings. 

For example, `TextField` currently has code like this in it's constructor:

```cs
AddKeyBinding (Key.DeleteChar, Command.DeleteCharRight);
AddKeyBinding (Key.D | Key.CtrlMask, Command.DeleteCharRight);

AddKeyBinding (Key.Delete, Command.DeleteCharLeft);
AddKeyBinding (Key.Backspace, Command.DeleteCharLeft);
...
```

This should be replaced with configuration in `.\Terminal.Gui\Resources\config.json` like this:

```json
"TextView.KeyBindings": {
  "DeleteCharRight" : {
    "Key" : "DeleteChar",
    "Modifiers": [
    ]
  },
  "DeleteCharRight" : {
    "Key" : "D",
    "Modifiers": [
      "Ctrl"
    ]
  }
}
...
```

For this to work,  `View` and any subclass that defines default keybindings should have a member like this:

```cs
	public partial class View : Responder, ISupportInitializeNotification {

		[SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
		public static Dictionary<Command, Key> DefaultKeyBindings { get; set; }
``` 

(This requires more thought - because CM requires config properties to be `static` it's not possible to inherit defualt keybindings!)

dodexahedron:
I think one issue that will crop up in use, and in more places than just trying to use Microsoft.Extensions.Configuration (though that's a big one), is the custom Json serializer that is currently forced by default by being placed on the Key class, rather than used if and when actually needed. It forces a non-standard representation of the object, and requires any deserializer to be aware of the attribute, which CM is not, unfortunately (seems a rather unfortunate shortcoming of CM, if you ask me).

So, four questions:

- Why is the KeyJsonConverter there in the first place?
- Why does it serialize the specific way it does?
- Why is it on the class, rather than referenced when needed?
- Why is it internal?

tig:
> I think one issue that will crop up in use, and in more places than just trying to use Microsoft.Extensions.Configuration (though that's a big one), 

[...]

PR #3515, opened 2024-05-29
Fixes #3512. Ctrl+Z not suspending

### LINKED ISSUE #3512: `Ctrl+Z` not suspending
** Describe the bug**
`Ctrl+Z` not working in version `2.0.0-pre.`

**To Reproduce**
```cs
using Terminal.Gui;

static class Program
{
    class MainWindow : Toplevel
    {
        Label label;
        public MainWindow()
        {
            label = new Label
            {
                Text = "Hello, world!"
            };
            Add(label);
        }
    }

    static void Main(string[] args)
    {

        Application.Init();
        try
        {
            Application.Run<MainWindow>().Dispose();
        }
        finally
        {
            Application.Shutdown();
        }
    }
}

```
**Expected behavior**
When `Ctrl+Z` is pressed, the `app` should be suspended and returned to last state, when brought to foreground (`fg`).

 - OS: macOS 14.4.1
 - Terminal: Terminal and iTerm2
 - Version: Terminal.Gui v2.0.0-pre.949

**Additional context**
This works fine in version `1.16.0`. I tried to test against the first `pre` versions of `2.0.0` and:
*** version <=`2.0.0-pre.231` ***
  - `Ctrl+Z` works always first time, but:
  - when returning to the `app`, `Ctrl+Z` and `Ctrl+Q` (`QuitKey`) don't work;
  - keyboard input is echoing
  - after exiting with `Ctrl+C`, mouse movements output *stuff* to the terminal.
 
*** version >= `2.0.0-pre.232` ***
  - `Ctrl+Z` does nothing.  
  - `OnKeyDown` is called.
 ```cs
        public override bool OnKeyDown(Key keyEvent)
        {
            label.Text = $"Keydown {keyEvent}";
            if (keyEvent == Key.Esc)
                Application.RequestStop();
            if (keyEvent == Key.Z.WithCtrl)
            {
                label.Text += " is Z with Ctrl";
                return false;
            }
            if (keyEvent == Key.Q.WithCtrl)
            {
                label.Text += " is Q with Ctrl";
                return false;
            }
            return true;
        }
```
if the code above is added to `MainWindow`, `Ctrl+Q` (`QuitKey`) is bubbled and Application exits but `Ctrl+Z` just updates `label.Text`.

### PR DESCRIPTION
**Pull request by BDisp**
## Fixes

- Fixes #3512

## Proposed Changes/Todos

- [x] Add #define UNIX_KEY_BINDINGS

## Pull Request checklist:

### CODE REVIEW

File: Terminal.Gui/Views/Toplevel.cs [hunk 1]
```diff
@@ -1,3 +1,5 @@
+#define UNIX_KEY_BINDINGS
```
tig, line 1: I do not want all those other "unix" key bindings enabled on non-Unix platforms. They get in the way of normal usage. They get in the way.

Until https://github.com/gui-cs/Terminal.Gui/issues/3023 is addressed (or something similarly flexible), can you please do this in this PR instead of `#define UNIX_KEY_BINDINGS`

```cs
if (Application.Driver is CursesDriver) 
{
   KeyBindings.Add (Key.Z.WithCtrl, Command.Suspend);
}
```

Thanks.

tig (changes requested):

Terminal.Gui/Views/Toplevel.cs
BDisp: I changed to a conditional `PlatformID.Unix` because now `NetDriver` is also using the `Suspend` method on a `Linux` or `Mac`.

```cs
        if (Environment.OSVersion.Platform == PlatformID.Unix)
        {
            KeyBindings.Add (Key.Z.WithCtrl, Command.Suspend);
        }

```

dodexahedron (commented):
Rather than hot loops, you can just start a free-running timer at startup and an event handler when it ticks.

Should be like... half the code, if even that much.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -315,12 +315,31 @@ public void ClearContents ()
...
+                _dirtyLines [row] = true;
+            }
+        }
+    }
+
+    /// <summary>
+    /// Sets <see cref="Contents"/> as dirty for situations where views
+    /// don't need layout and redrawing, but just refresh the screen.
+    /// </summary>
+    public void SetContentsAsDirty ()
```
tig, line 332: 1) What does this have to do with Ctrl-Z? 
2) I think this would be better implemented as 

```cs
public bool ForceFullUpdate {get;set;}
```

Then in `Update`:

```cs
      if (ForceFullUpdate || !Contents [row, col].IsDirty)
...

ForceFullUpdate = false;
```

tig: And can't it be made `private` since it's only called from a derived class?

tig (changes requested):
Good stuff. See my comments.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
BDisp: > 1. What does this have to do with Ctrl-Z?

This is related with the issue #3525 and was intended to use in the `Suspend` method, so it have to do with Ctrl-Z and I leveraged this branch to implement.

> 2. I think this would be better implemented as
> 
> ```cs
> public bool ForceFullUpdate {get;set;}
> ```
> 
> Then in `Update`:
> 
> ```cs
>       if (ForceFullUpdate || !Contents [row, col].IsDirty)
> ...
> 
> ForceFullUpdate = false;
> ```

I don't agree because if it use a property I have to include that code on two derived class, `CursesDriver` and `NetDriver`. Using a method on the `ConsoleDriver` it's only implemented once. If you don't like the method name, then you may suggest another name.

BDisp: > And can't it be made `private` since it's only called from a derived class?

No, because it is inaccessible in derived classes if I change to private.

Terminal.Gui/Views/Toplevel.cs
tig: Thanks!

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
tig: Sounds good. I appreciate the explanation; makes sense.

tig (approved):

PR #3519, opened 2024-05-30
Fixes #3518. v1 NetDriver throws System.InvalidOperationException when a key is pressed.

### LINKED ISSUE #3518: v1 NetDriver throws System.InvalidOperationException when a key is pressed.
Open de UICatalog and press any key. The `InvalidOperationException` will be throw. Instead of using the keyboard and only using the mouse this issue doesn't happens, but when a key is pressed an exception is immediately throwed.

image

BDisp:
## Fixes

- Fixes #3518

## Proposed Changes/Todos

- [x] Put the Console.ReadKey inside a try/catch block
- [x] Also fixes an issue where the cursor isn't showing when the cursor is visible
- [x] Add support for Suspend method

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/NetDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/NetDriver.cs
@@ -177,7 +177,15 @@ namespace Terminal.Gui {
 			ConsoleKeyInfo newConsoleKeyInfo = default;
 
 			while (true) {
-				ConsoleKeyInfo consoleKeyInfo = Console.ReadKey (true);
+				ConsoleKeyInfo consoleKeyInfo;
+
+				try {
+					consoleKeyInfo = Console.ReadKey (true);
+				} catch (InvalidOperationException ex) {
+
+					return;
+				}
+
 				if ((consoleKeyInfo.KeyChar == (char)Key.Esc && !isEscSeq)
 					|| (consoleKeyInfo.KeyChar != (char)Key.Esc && isEscSeq)) {
 					if (cki == null && consoleKeyInfo.KeyChar != (char)Key.Esc && isEscSeq) {
--- a/Terminal.Gui/ConsoleDrivers/NetDriver.cs
+++ b/Terminal.Gui/ConsoleDrivers/NetDriver.cs
@@ -1027,6 +1027,16 @@ namespace Terminal.Gui {
 
 		public override void Suspend ()
 		{
+			if (Environment.OSVersion.Platform != PlatformID.Unix) {
+				return;
+			}
+
+			StopReportingMouseMoves ();
+			Console.ResetColor ();
+			Console.Clear ();
+			Platform.Suspend ();
+			Application.Refresh ();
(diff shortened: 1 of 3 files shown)


Terminal.Gui/ConsoleDrivers/NetDriver.cs
```diff
@@ -177,7 +177,15 @@ void GetConsoleKey ()
 			ConsoleKeyInfo newConsoleKeyInfo = default;
 
 			while (true) {
-				ConsoleKeyInfo consoleKeyInfo = Console.ReadKey (true);
+				ConsoleKeyInfo consoleKeyInfo;
+
+				try {
+					consoleKeyInfo = Console.ReadKey (true);
+				} catch (InvalidOperationException ex) {
```
dodexahedron, line 184: Just eating the exception?

This seems like a bad idea.

Any redirection of Console.In at any point can cause this on a call to `Console.ReadKey()`, and you don't want to eat that exception then, because it isn't recoverable by the library and represents a breaking change.

dodexahedron (changes requested):
Need to handle this differently.

If possible, avoid the problem rather than eating the exception, because there are other valid reasons for that exception to be thrown.

Terminal.Gui/ConsoleDrivers/NetDriver.cs
BDisp: There is no redirection anywhere. Please can you confirm that? Unless is the console API that redirect itself. Ignoring the exception the application continue running normally.

dodexahedron: In user code.

We are a library, not an application.

And eating exceptions is bad behavior, regardless. We don't know the potential cause of the exception. If we did, we would avoid it or react to it, not eat it.

BDisp: This is easy to react to when we understand why this happens and how to deal with it. This exception is complained about by many users, just do a Google search to confirm it. In my opinion, I think it is because `StdIn` is being used by the terminal API used at the time `Console.ReadKey(bool)` is running. In case of the exception, I have no idea what to do other than ignore it and continue. At least, it has worked well that way. If you know what causes this and can interact with the error, it would be an advantage to explain how to resolve it. I know it's not good practice to ignore exceptions but there are many APIs that generate them and the user already knows that this will happen but they don't need to do anything other than return or ignore and let it flow.

BDisp: I chose to remove all `InvalidOperationException` and use `Task.Delay.Wait` with `CancellationTokenSource` to wait until there is some `ConsoleKeyInfo` available or the `CancellationTokenSource` is cancelled. The cause of the `InvalidOperationException` error is that `Console.ReadKey` cannot be running if `Console.KeyAvailable` is `false`.

tig (approved):

PR #3533, opened 2024-06-10
Fixes #3073. Adds `Shortcut`, `Bar`, and replaces `StatusBar`

### LINKED ISSUE #191: Menu hints in StatusBar

### LINKED ISSUE #2106: StatusBar should support right alignment on last StatusItem

### LINKED ISSUE #3037: `View.GetKeyFromCommand` does not deal with multiple commands correctly

### LINKED ISSUE #3073: Create a new `Bar` view to replace `StatusBar` (and eventually `MenuBar`)

tig:
(New PR & branch - replaces #3076 and #3531 which were messed up).

(Branch is mis-named because dyslexia).

## Fixes 

- [x] Fixes #3073
- [x] Fixes #2106
- [x] Fixes #191

## Todos

*Still very much a Proof of Concept/Prototype*

- [x] Implement `Shortcut` view that holds a `Key`, `KeyBindingScope`, `Title`, and help `Text`. 
  - [x] Renders just like a `MenuItem`.
    - [x] Proper spacing/alignment of help text.
  - [x] Supports arbitrary `View`s in the `CommandView` (first) position. E.g. `CheckBox`
  - [x] Works with `KeyBindingScope.Application` and `KeyBindingScope.HotKey`.
  - [x] Supports mouse clicks to activate the shortcut
- [x] Implement `Bar` view that
  - [x] Looks just like the old `StatusBar` (`StatusBarStyle = true`, `Orientation.Horizontal`)
  - [ ] Looks just like the old `MenuBar` (`StatusBarStyle = false`, `Orientation.Horizontal`)
  - [x] Looks just like the old `Menu` (`StatusBarStyle = false`, `Orientation.Vertcal`)
- [x] New Scenario to show off/test `Shortcut`
- [x] New Scenario to test/show off `Bar`
- [x] Update `UI Catalog` to use the new `Bar`
- [x] Update to leverage `Dim.Auto` - ish 
- [x] Update to leverage `Pos.Align`
- [x] `Shortcut` Unit tests
- [x] Replace `Toplevel.StatusBar`
- [ ] `Bar` Unit tests

7QxcJMo 1

## Pull Request checklist:


tig:
Tons of progress on this.

It is now at the point that replacing `StatusBar` will be trivial. I already updated UI Catalog to use `Bar` instead of `StatusBar`.

`Menu` and `MenuBar` will require a lot more work, but directionally I'm very happy.

![Uploading 7QxcJMo[1].gif…]()

tig:
I'll add more unit tests before merging, and may spend more time on ensuring this theoretically can replace `Menu` and `MenuBar`, but this is ready for review.

tig:
Related question:

In v1 and early v2, for a Hotkey (e.g. "I'm a _Label") to get invoked, the view in question had to be a subview of Top.Focused. And only subviews of Top.Focused were considered.

This mean that even though a hotkey was defined, the view heirarchy might prevent it from working. You could see this in All Views Tester where the size and position FrameViews have subviews with hotkeys. You could only use keyboard to navigate to a view in the OTHER FrameView by tabbing to it first. 

In this PR, I've changed this such that as long as the view is visible (and enabled), the hotkey will cause it to gain focus. I think this is far more natural and sensical.

BDisp:
I also think that all shortcuts predefined by `Application` should be unique. If the user configures an identical shortcut, it will be ignored and the default one defined by the application will be executed. I'm tired of facing this situation with `Windows Terminal`, so be patient.
In the remaining shortcuts, priority should be `Focused`, `SuperView`, `Application`.

> In this PR, I've changed this such that as long as the view is visible (and enabled), the hotkey will cause it to gain focus. I think this is far more natural and sensical.

You are referring to whether `CanFocus` is also true, right?

tig:
> You are referring to whether `CanFocus` is also true, right?

Not entirely. E.g. `Label` is `CanFocus = false`, but the whole point of `Label` is to act as a hotkey for the View it is a label for.

Above, you used the term shortcut. I'd appreciate it if you re-wrote what you used using this terminology (or, if the terminology doesn't make sense, suggest how it could be improved):

### **[HotKey](~/api/Terminal.Gui.View.yml#Terminal_Gui_View_HotKey)** 

A **HotKey** is a keypress that selects a visible UI item. For selecting items across `View`s (e.g. a `Button` in a `Dialog`) the keypress must have the `Alt` modifier. For selecting items within a `View` that are not `View`s themselves, the keypress can be key without the `Alt` modifier.  For example, in a `Dialog`, a `Button` with the text of "_Text" can be selected with `Alt-T`. Or, in a `Menu` with "_File _Edit", `Alt-F` will select (show) the "_File" menu. If the "_File" menu has a sub-menu of "_New" `Alt-N` or `N` will ONLY select the "_New" sub-menu if the "_File" menu is already opened.

By default, the `Text` of a `View` is used to determine the `HotKey` by looking for the first occurrence of the [HotKeySpecifier](~/api/Terminal.Gui.View.yml#Terminal_Gui_View_HotKeySpecifier) (which is underscore (`_`) by default). The character following the underscore is the `HotKey`. If the `HotKeySpecifier` is not found in `Text`, the first character of `Text` is used as the `HotKey`. The `Text` of a `View` can be changed at runtime, and the `HotKey` will be updated accordingly. [HotKey](~/api/Terminal.Gui.View.yml#Terminal_Gui_View_HotKey) is `virtual` enabling this behavior to be customized.

### **[Shortcut](~/api/Terminal.Gui.Shortcut.yml) - An opinionated (visually & API) View for displaying a command, helptext, key.
**

A **Shortcut** is a keypress that invokes a [Command](~/api/Terminal.Gui.Command.yml) or `View`-defined action even if the `View` that defines them is not focused or visible (but the `View` must be enabled). Shortcuts can be any keypress; `Key.A`, `Key.A | Key.Ctrl`, `Key.A | Key.Ctrl | Key.Alt`, `Key.Del`, and `Key.F1`, are all valid. 

`Shortcuts` are used to define application-wide actions (e.g. `Quit`), or actions that are not visible (e.g. `Copy`).

[MenuBar](~/api/Terminal.Gui.MenuBar.yml), [ContextMenu](~/api/Terminal.Gui.ContextMenu.yml), and [StatusBar](~/api/Terminal.Gui.StatusBar.yml) support `Shortcut`s.

tig:
@bidisp - Do you know if there's way to "Un focus" a view? There are times where I'd like to say "Make it so NO view in this View heirarchy has focus".

BDisp:
> @bidisp - Do you know if there's way to "Un focus" a view? There are times where I'd like to say "Make it so NO view in this View heirarchy has focus".

If it is internally you can use `SuperView.Focused = null`.

UICatalog/Scenarios/HexEditor.cs
```diff
@@ -16,21 +16,30 @@ public class HexEditor : Scenario
...
     private bool _saved = true;
-    private StatusItem _siPositionChanged;
+    private Shortcut _siPositionChanged;
     private StatusBar _statusBar;
 
-    public override void Setup ()
+    public override void Main ()
     {
-        Win.Title = GetName () + "-" + _fileName ?? "Untitled";
+        Toplevel app = new Toplevel ()
```
BDisp, line 24: Please add `Application.Init ();` into the scenario to avoid exception been throwing.

UICatalog/Scenarios/InteractiveTree.cs
```diff
@@ -10,52 +10,47 @@ public class InteractiveTree : Scenario
 {
     private TreeView _treeView;
 
-    public override void Setup ()
+    public override void Main ()
     {
-        Win.Title = GetName ();
-        Win.Y = 1; // menu
-        Win.Height = Dim.Fill (1); // status bar
+        var appWindow = new Toplevel ()
```
BDisp, line 15: Please add `Application.Init ();` into the scenario to avoid exception been throwing.

UICatalog/Scenarios/LineViewExample.cs
```diff
@@ -9,80 +9,82 @@ namespace UICatalog.Scenarios;
...
 public class LineViewExample : Scenario
 {
-    public override void Setup ()
+    public override void Main ()
     {
-        Win.Title = GetName ();
-        Win.Y = 1; // menu
-        Win.Height = Dim.Fill (1); // status bar
+        // Setup - Create a top-level application window and configure it.
+        Toplevel appWindow = new ();
```
BDisp, line 15: Please add `Application.Init ();` into the scenario to avoid exception been throwing.

BDisp (changes requested):
LGTM. Only 3 items need your attention.

UICatalog/Scenarios/InteractiveTree.cs
```diff
@@ -10,52 +10,47 @@ public class InteractiveTree : Scenario
...
+                                           new (Key.C.WithCtrl, "Add Child", AddChildNode),
+                                           new (Key.T.WithCtrl, "Add Root", AddRootNode),
+                                           new (Key.R.WithCtrl, "Rename Node", RenameNode)
                                        }
                                       );
-        Top.Add (statusBar);
+        appWindow.Add (statusBar);
+
+        Application.Run (appWindow);
+        appWindow.Dispose ();
```
dodexahedron, line 52: Note that calls to Dispose that are not within a `finally` block (either explicitly or via a `using`) will never get called if there's an unhandled exception.

Probably good to encourage proper disposal in the samples, but of course it's not a big deal. 🤷‍♂️

Terminal.Gui/ConsoleDrivers/CursesDriver/constants.cs
dodexahedron, line 1: It sure will be nice if/when Curses can DIAF.

Terminal.Gui/ConsoleDrivers/EscSeqUtils/EscSeqUtils.cs
```diff
@@ -242,10 +243,10 @@ public enum ClearScreenOptions
 
                 break;
             case "SS3":
-                key = GetConsoleKey (terminator [0], values [0], ref mod);
+                key = GetConsoleKey (terminator [0], values [0], ref mod, ref keyChar);
```
dodexahedron, line 246: I don't love this.

Once there start being multiple outs or non-readonly refs, it's usually time to either use an implicit ValueTuple or a formal return type. Considering the scope of these, this is actually potentially a place where a ref struct could be quite beneficial.

Not worth upending that right now though. Just a note.

Terminal.Gui/ConsoleDrivers/EscSeqUtils/EscSeqUtils.cs
```diff
@@ -342,15 +343,26 @@ public static string GetC1ControlChar (in char c)
...
     /// <returns>The <see cref="ConsoleKey"/> and probably the <see cref="ConsoleModifiers"/>.</returns>
-    public static ConsoleKey GetConsoleKey (char terminator, string? value, ref ConsoleModifiers mod)
+    public static ConsoleKey GetConsoleKey (char terminator, string? value, ref ConsoleModifiers mod, ref char keyChar)
     {
         if (terminator == 'Z')
         {
             mod |= ConsoleModifiers.Shift;
         }
 
+        if (terminator == 'l')
```
dodexahedron, line 356: These could be rolled into the switch as guard clauses for consistency.

Terminal.Gui/Drawing/Aligner.cs
```diff
@@ -108,10 +108,11 @@ public int ContainerSize
             spacesToGive = containerSize - totalItemsSize;
         }
 
+        AlignmentModes mode = alignmentMode & ~AlignmentModes.AddSpaceBetweenItems; // copy to avoid modifying the original
```
dodexahedron, line 111: 🙏 

Did you profile this? I have. This makes a surprisingly significant difference (a good one).

Thank you!

dodexahedron (approved):
Very minor nit picks.

But I gotta go, so I'm sending this your way.

BDisp:
@tig some unit tests are failing. I think what causes this are the changes in ScenarioTests in which you call `Application.RemoveTimeout` when other processes may be occurring in other threads. This manipulation will have to be carried out with thread safety, otherwise it will happen that when `Application.QuitKey` is called it will throw an exception

```ex
System.NullReferenceException: 'Object reference not set to an instance of an object.'

scenario was null.
```

I recommend reverting these changes or ensuring that they are made thread-safe.

By the way, have you already merged this branch with the v2_develop branch (updated, of course)? I check that there are commits that are already included in v2_develop but appear in this PR as new commits.

tig:
> @tig some unit tests are failing...  or ensuring that they are made thread-safe.

Yeah, I was not quite done with making those changes. I was throwing spaghetti a the wall to see what would stick and then had to go to bed. I've now thought it through better and think it is now correct.

Please take a look now and let me know if you agree or not.

BDisp:
Unit tests passed on Windows but not on Linux. Are these changes made really necessary?

tig:
> Unit tests passed on Windows but not on Linux. Are these changes made really necessary?

Here's why:

- I had test failures that were really hard to track down because the scenario tests were failing when unrelated. This makes those scenario tests more robust and less likely to cause other tests to fail.
- Sample code should illustrate correctness. Having a test that proves each `Scenario` calls Init/Shutdown correctly enforces this.
- That code was poorly written originally (by me) and was incorrect. It had a high probability of failing in the future in weird ways.

BDisp:
> Here's why:
> 
> * I had test failures that were really hard to track down because the scenario tests were failing when unrelated. This makes those scenario tests more robust and less likely to cause other tests to fail.

Yes, this could actually be happening when `Application.Init` is called and there are processes running in the background that are using `Terminal.Gui`, namely the `SynchronizationContext` which, being forced to close, will throw an exception, as is expected when it is returned from the respective process.

> * Sample code should illustrate correctness. Having a test that proves each `Scenario` calls Init/Shutdown correctly enforces this.

Because of possible background processes, it is not enough to ensure that Init/Shutdown are in sync. It is also necessary to check if there are still processes running in the background and wait for them to complete before being destroyed by Shutdown.

> * That code was poorly written originally (by me) and was incorrect. It had a high probability of failing in the future in weird ways.

It wasn't poorly written by you, on the contrary. What happens is that `Terminal.Gui` is always evolving and it is impossible to predict future impacts. So we have to have something that checks the state of `Mainloop.IdleHandlers` and waits for its completion before discarding all objects in Shutdown.

tig:
I'm trying to debug the linux unit test failure. This new test passes on Windows but never exits when run on Linux.

Stumped.

```cs
    [Fact]
    public void Run_Iteration_Fires ()
    {
        int iteration = 0;

        Application.Init ();

        Application.Iteration += Application_Iteration;
        Application.Run<Toplevel> ().Dispose ();

        Assert.Equal (1, iteration);
        Application.Shutdown ();

        return;

        void Application_Iteration (object sender, IterationEventArgs e)
        {
            iteration++;
            Application.RequestStop ();
        }
    }
```

tig:
To be clear, the `Iteration` event is never fired on Linux with this test.

tig:
I am calling `Init` without a driver. Hence on Windows, WindowsDriver is being used. 

If I change to `Init(new FakeDriver())` this test succeeds on both Windows and Linux.

Oops.

BDisp:
> I'm trying to debug the linux unit test failure. This new test passes on Windows but never exits when run on Linux.

The application doesn't even end in WindowsDriver.

> To be clear, the `Iteration` event is never fired on Linux with this test.

Not even in WindowsDriver.

BDisp:
So perhaps the failure in the unit tests could be something that could be using the `WindowsDriver` or the `CursesDriver` ending up giving an error.

BDisp:
`AutoInitShutdown` can be removed here because the `Application.Init` is called in the NavigationTests.cs.

image

BDisp (approved):

PR #3552, opened 2024-06-19
Fixes #3551. v2: Big performance impact on ListView/ComboBox's that contain a lot of items

### LINKED ISSUE #3551: v2: Big performance impact on ListView/ComboBox's that contain a lot of items
2.0.0-pre.1588 On lists with a lot of items (4k) the code below when selecting an item from the combo, takes 8 seconds to close the dialog
 
**Expected behavior**
2.0.0-pre.1491 the dialog close is instantaneous

Maybe something to do with ObservableCollection firing events for every item on close?

```CSharp
Application.Init();

var list = new List<string>();

for (int i = 0; i < 4000; i++)
	list.Add($"item{i}");

var combo = new ComboBox() { Width = Dim.Fill(), Height = Dim.Fill() };
combo.SetSource(new ObservableCollection<string>(list));
combo.OpenSelectedItem += (object sender, ListViewItemEventArgs e) => Application.RequestStop();

var dialog = new Dialog() { Title = "Main", Width = Dim.Percent(40), Height = Dim.Percent(50) };
dialog.Add(combo);
Application.Run(dialog);

MessageBox.Query("Selected", combo.Text);

Application.Run();
Application.Shutdown();
```

BDisp:
## Fixes

- Fixes #3551

## Pull Request checklist:


Terminal.Gui/Views/ComboBox.cs
```diff
@@ -738,11 +744,15 @@ private void SetSearchSet ()
             return;
         }
 
+        _listview.SuspendCollectionChangedEvent ();
+
         // force deep copy
```
dodexahedron, line 749: Can you drop a `// PERF: Improve this` comment here?

This is fine as a short-term workaround to allow work to continue, but really should be a release blocker in this state.

dodexahedron: Also, if we merge this PR, can you keep the issue open or re-open it and assign to me after the merge, so it's on my radar?

dodexahedron (changes requested):
Just a request for a todo list tag so I don't forget to revisit this later.

Terminal.Gui/Views/ComboBox.cs
dodexahedron: Maybe @mention me in the PERF comment, too, just for visibility?

BDisp: I can without any problems, but notice that I'm just ignoring it in the `ComboBox` because there is no need for the event to always be invoked just when opening and closing the list. In the `ListView`, the normal behavior continues to be used, that is, the event is invoked when there is a change in the data source, except when the user suspends it, of course.

BDisp: @dodexahedron done.

dodexahedron: Thanks.

Yeah, I'll take a deeper look at how it all fits together as part of a set of whole-library performance passes, since a lot of things are likely to involve more than just specific View types.

PR #3555, opened 2024-06-20
Fixes #3516. Changes default `QuitKey` to `Esc`

### LINKED ISSUE #3516: Change default `Appliation.QuitKey` to `Esc` from `Ctrl-Q`
Ctrl-Q is non-standard across all platforms and un-intuitive. 

Esc is much more natural.

The change simply requires changing this line in `\Terminal.Gui\Resources\config.json`:

image

And updating some unit tests.

tig:
## Fixes

- Fixes #3516 

## Proposed Changes/Todos

- [x] Change to `Esc`
- [x] Fix tests
- [x] Fix scenarios (if any)
- [x] Update docs

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Resources/config.json
+++ b/Terminal.Gui/Resources/config.json
@@ -18,7 +18,7 @@
 
   "Application.AlternateBackwardKey": "Ctrl+PageUp",
   "Application.AlternateForwardKey": "Ctrl+PageDown",
-  "Application.QuitKey": "Ctrl+Q",
+  "Application.QuitKey": "Esc",
   "Application.IsMouseDisabled": false,
   "Theme": "Default",
   "Themes": [
--- a/Terminal.Gui/Resources/config.json
+++ b/Terminal.Gui/Resources/config.json
@@ -10,7 +10,8 @@
   // note that not all values here will be recreated (e.g. the Light and Dark themes and any property initialized
   // null).
   //
-  "$schema": "https://gui-cs.github.io/Terminal.Gui/schemas/tui-config-schema.json",
+  // TODO: V2 - Reference via http
+  "$schema": "../../docfx/schemas/tui-config-schema.json",
 
   // Set this to true in a .config file to be loaded to cause JSON parsing errors
   // to throw exceptions. 
@@ -19,7 +20,6 @@
   "Application.AlternateBackwardKey": "Ctrl+PageUp",
   "Application.AlternateForwardKey": "Ctrl+PageDown",
   "Application.QuitKey": "Esc",
-  "Application.IsMouseDisabled": false,
   "Theme": "Default",
   "Themes": [
     {
--- a/Terminal.Gui/View/ViewKeyboard.cs
+++ b/Terminal.Gui/View/ViewKeyboard.cs
@@ -745,8 +745,8 @@ public partial class View
     /// <summary>
     /// Returns true if Key is bound in this view heirarchy. For debugging
     /// </summary>
-    /// <param name="key"></param>
-    /// <param name="boundView"></param>
+    /// <param name="key">The key to test.</param>
+    /// <param name="boundView">Returns the view the key is bound to.</param>
     /// <returns></returns>
     public bool IsHotKeyKeyBound (Key key, out View boundView)
     {
--- a/Terminal.Gui/View/ViewKeyboard.cs
+++ b/Terminal.Gui/View/ViewKeyboard.cs
@@ -633,6 +633,7 @@ public partial class View
     ///     <para>See <see href="../docs/keyboard.md">for an overview of Terminal.Gui keyboard APIs.</see></para>
     /// </remarks>
     /// <param name="keyEvent">Contains the details about the key that produced the event.</param>
+    /// <param name="scope">The scope.</param>
     /// <returns>
     ///     <see langword="false"/> if the key press was not handled. <see langword="true"/> if the keypress was handled
     ///     and no other view should see it.
@@ -745,6 +746,7 @@ public partial class View
     /// Returns true if Key is bound in this view heirarchy. For debugging
     /// </summary>
     /// <param name="key"></param>
+    /// <param name="boundView"></param>
     /// <returns></returns>
     public bool IsHotKeyKeyBound (Key key, out View boundView)
     {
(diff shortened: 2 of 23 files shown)


BDisp:
There is something strange when clicking a button in a Dialog. Try opening the `InteractiveTree` scenario and clicking Add Root in the status bar. The `Dialog` will open and after typing the node name click on the `Ok` button. The `Dialog` closes and opens again despite `Application.RequestStop` having been called. Therefore, the `Dialog` should not open and the same happens with the `Cancel` button.
It is a good idea to create a unit test to prevent this bad behavior from happening.

tig:
Do you think this has something to do with this PR?

Does this reproduce on v2_develop. I'm not a computer to check.

BDisp:
> Do you think this has something to do with this PR?
> 
> Does this reproduce on v2_develop. I'm not a computer to check.

Yes it reproduce on v2_develop branch and so has nothing with this PR.

dodexahedron:
I have a question.

What happens when an escape literal is received on stdin? Quit, by default?

Because I ran into stuff today that has drastically altered my stance, if that's the proposed change to default behavior, which I'm glad to share if that is, in fact, what it is.

BDisp:
> I have a question.
> 
> What happens when an escape literal is received on stdin? Quit, by default?
> 
> Because I ran into stuff today that has drastically altered my stance, if that's the proposed change to default behavior, which I'm glad to share if that is, in fact, what it is.

If only an escape is received it will Quit, otherwise the ANSI escape sequence will be processed if the terminal supports it, otherwise it will be echoed as text.

dodexahedron:
OK. That I think mostly jives with expectations.

I think it would be good to include some tests for it that do some things such as redirecting stdin and writing esc to it, to test/prove at least that particular case.

Most other ways of testing scenarios and environments that I thought up aren't really things that can be done in that unit test project, but that one is simple and good enough for a basic case that should cover at least some of them, I think.

Here were scenarios I thought up last night (food for thought, maybe, but not hard suggestions really):

 * Asymmetric input if multiple processes are watching stdin (loggers, input filters, password managers, certain keyboard drivers and macros, and a whole range of other things)
 * What if they're kibitzing with another user (That's almost the same as the previous point, but now multiple humans are directly involved)
 * What if they feed it from a pipe or other stream for automation?
 * What if there's an automated process involved? That includes people's UI testing.
 * What if they redirect stdin? Maybe that input comes from a formatted file or even a remote session?
 * What if their mouse input is dependent on it?
    * This isn't uncommon, and is part of why sometimes terminals get screwed up and mouse input causes garbage at the prompt.
 * What if they want to actually format text they're storing in a text field?
 * What if they're using UI theming software (such as Oh My Posh, even), which might want to use escape sequences to change cursor shapes and stuff like that?

And if you've already got that covered by your planned implementation, awesome! 🥳

Terminal.Gui/Views/Shortcut.cs
```diff
@@ -631,7 +631,7 @@ private void UpdateKeyBinding ()
     ///     - if the user presses the HotKey specified by CommandView
     ///     - if HasFocus and the user presses Space or Enter (or any other key bound to Command.Accept).
     /// </summary>
-    protected new bool? OnAccept (CommandContext ctx)
+    protected bool? OnAccept (CommandContext ctx)
```
dodexahedron, line 634: Is the nullable necessary here?

Terminal.Gui/View/ViewKeyboard.cs
```diff
@@ -745,6 +746,7 @@ private bool ProcessSubViewKeyBindings (Key keyEvent, KeyBindingScope scope, ref
     /// Returns true if Key is bound in this view heirarchy. For debugging
     /// </summary>
     /// <param name="key"></param>
```
dodexahedron, line 748: Empty param docs

Terminal.Gui/Resources/config.json
dodexahedron, line 1: A nice piece of formality for later for any custom JSON or other data/config files would be to write a schema document for them.

JSON schema documents are pretty simple and are, themselves, JSON.

Probably something to do during beta, when changes to the schema are unlikely.

That also gives suggestion/completion capabilities to any editor supporting schemas (most of the popular ones do).

dodexahedron (commented):
Just a quick scan taken over the changes to the Terminal.Gui code.

Didn't do in-depth analysis yet and didn't bother with the tests or samples either.

Nothing _critical_ popped out at me anyway on scan.

Only things that prompted any questions were a nullable bool return and a hidden method. But both are pretty minor.

Terminal.Gui/Views/Shortcut.cs
tig: I think so. 

Part of the design of Command/KeyBinding is described here.  For this, I believe we need to support propogating "don't know/yes/no" up and down the stack. OnAccept is right in the middle of all this.

https://github.com/gui-cs/Terminal.Gui/blob/fd683d83f5c279129c7f5ba5345c6c9ba92e2d1e/Terminal.Gui/View/ViewKeyboard.cs#L653-L658

Terminal.Gui/Resources/config.json
tig: I already did this. But for v2 the link is broken:

It is in ./docfx/schemas.  I wasn't sure how else to publish it.

Terminal.Gui/Views/Shortcut.cs
dodexahedron: Works for me.

An opportunity for a small future optimization is there, but it's fairly minor.

I also have been starting to think more about accessibility modifiers, as we get closer to completion. This one, being a kinda non-standard handler, because of not being void, might be a candidate for `private protected` instead of just `protected`.

The resulting visibility is the intersection of protected and internal, so it's visible to derived types in the same assembly only.

dodexahedron: That also lets it continue to do what it needs to do, without having to worry about it being overridden outside of the library.

Terminal.Gui/Resources/config.json
dodexahedron: Perfecto.

I dropped a quick suggestion about it.

We can reference it in the project as a Content item and have it output at build time.

In an ItemGroup in the csproj, you'd put:

```xml
<Content Include="relative/path/to/the/schema.json" CopyToOutputDirectory="PreserveNewest" />
```

dodexahedron: You'd also probably want to be sure to put a Pack element on it, for inclusion in the nuget package.

PR #3556, opened 2024-06-21
Fixes #3557: Mostly non-impacting minor changes to the csproj, a fix for nuget.config, and some design-time QoL stuff

### LINKED ISSUE #3557: Nuget.Config for v2 can behave badly for local CLI builds, plus some minor QoL stuff for Terminal.Gui.csproj and the solution
**Describe the bug**
The nuget.config file we have clears the host's config and then explicitly puts nuget back in.

However, the name it uses doesn't match the default for nuget from Visual Studio, and may possibly not match any custom names someone may have configured.

That's a problem because it breaks sourcelink, since it doesn't also link that item up as a packageSourceMapping. And that breaks the build, when it is the case (as it is on every one of my machines, for example).

For the non-problem QoL stuff, I've just made some minor tweaks to the Terminal.Gui.csproj file, as well as defined some solution filters, which can be used to optionally make your experience, in terms of VS performance, load times, build times, etc., nicer, as well as make CI builds faster and leaner, if they are modified to use the filters instead of just the Terminal.sln itself.

That feature was added back in VS2019 and is really nice, plus it gives a way to ensure that people have more consistent configurations for local builds, rather than relying on individual user preference suo files and such, which (correctly) are not in source control.

I put in a draft PR already: #3556 

Not a whole lot left to do for this one. Just cherry picking common changes from a couple of branches I have right now, to make rebases easier and to make those branches' merges easier when they're ready.

Many or perhaps even all of these changes are also compatible with V1, since they don't actually change any code or pull in any other dependencies, but I'll make a separate branch for that, if it's wanted, since the solution filters will be different, at least.

dodexahedron:
## Fixes

 - Fixes #3557 - Bad nuget behavior for local builds and sourcelink

Aside from that, mostly a few tweaks to the csproj files and other build-related stuff that don't really change anything but are a common base for a couple of branches I have open right now.

Well... Actually one does change/fix a minor thing - configuration conditions have the AnyCPU part removed, so all architectures get those properties, even if an explicit platform is given to the compiler.

This also brings in a couple of solution filters, which you can optionally use both as what you open in Visual Studio or what you target a build at, which don't change the .sln file, but just reduce what's actually loaded into the solution, if you use them. Very handy for making Visual Studio run a bit faster, if you're not wanting to load ALL the projects, and provides consistent templates for people to use, instead of relying on local non-tracked user preference files.

Also useful at build time, for release builds, to speed things up by not making MSBuild and nuget have to figure out everything for projects that aren't part of the build.

This also resolved all warnings in projects in the Analyzers folder.

## Proposed Changes/Todos

 - [x] Add NoSamples.slnf - Solution filter that only has the library, analyzers, and tests.
 - [x] Add Analyzers.slnf - Solution filter that only has the analyzers.
 - [x] Add Release.slnf - Solution filter containing only Terminal.Gui.csproj, UnitTests.csproj, and Terminal.Gui.Analyzers.Internal.csproj, for minimal build requirements.
 - [x] Define instances of "Terminal.Gui" in the Terminal.Gui.csproj file as references to $(AssemblyName).
 - [x] Remove the AnyCPU part of the build configuration conditions. Builds targeting other architectures would otherwise not get those properties.
 - [x] Address a couple of warnings that are very safe to ignore.

## Pull Request checklist:

- [x] My changes resolve a ton of warnings
   - On windows, anyway 

Just tossing this up while I finish cherry-picking bits from a couple of branches that form a nice common base.


dodexahedron:
Also, all or at least most of this stuff will be drop-in portable to v1, as well, if you want.

dodexahedron:
OK actually I will put in an issue, because there's one that touches the nuget.config file to fix bad sourcelink behavior for local command line builds.

dodexahedron:
Think I'm done with this one.

Zero warnings in the build output for the analyzer projects, now, as I addressed several and suppressed others that were just noise due to intentional naming and such.

dodexahedron:
Yep. Ready to go.

Yes, there's more to be done to the project files, but that's in different branches and targeting more specific things and will be in different PRs.

Terminal.Gui/Configuration/SettingsScope.cs
```diff
@@ -118,6 +118,9 @@ public class SettingsScope : Scope<SettingsScope>
             return this;
         }
 
+        // BUG: Not trim-compatible
+        // Not a bug, per se, but it's easily fixable by just loading the file.
+        // Defaults can just be field initializers for involved types.
```
tig, line 123: I really like having a built-in config file for defaults. I hope there's a trim compatible way of doing so.

tig (approved):
LGTM!

Terminal.Gui/Configuration/SettingsScope.cs
dodexahedron: You just initialize the variables in code. Same effect, no embedded text.

dodexahedron: To clarify:

An embedded text file is essentially just a more convoluted string constant, except that it's not a string constant or it would be safe. It's just some anonymous data with no references.

The trim analyzer doesn't understand that the method call used to retrieve it is what "references" it.

Regardless, it's still effectively a hard-coded default value for everything, making it functionally equivalent to Properties/fields having initializers or being initialized in a constructor.

Therefore, the only functional difference on our end is that you do have to make sure a change in one is reflected in the other. An msbuild step or analyzer script that runs at build time could be made to automate that, but I can't imagine we'll be changing defaults with any regularity anyway.

tig: Let me restate: I really like having a single file that contains all defaults. I do not like having defaults spread throughout the project.

dodexahedron: I got that and fully agree with the desire anyway, for multiple reasons, including simply for DRY.

That's what code generation or text templating is for, and why pretty much any project of similar complexity usually does have such things, either as an explicit step in the pre-design process when setting up the build environment (a la autoconf and the standard `configure` script it makes), or (more common and idiomatic for .net) at design time, via Roslyn. The goal is typically both a heavy helping of DRY and transparent automation of otherwise tedious or prone-to-be-forgotten tasks.

Pretty much every other option either results in multiple definitions (ew), incompatibilities/limitations, or are worse means of doing something standard.

For any future project-specific generators, I'd just suggest them being in their own repos, so they can have their output published right on github and automatically pulled into the Terminal.Gui project at design time (one of the most common ways of doing it, when it's not a generator meant for general public consumption), so you don't encounter difficulties with Visual Studio, like the project in the current solution.

Another automatic means of doing it that I've both seen occasionally and used myself a small number of times (mostly from other projects or before access to the Roslyn APIs was public) is having generators/templates/scripts that are needed for solution prep wired up as git hooks on post-checkout and also optionally as a pre-build step in the projects needing their output. That results in you having the files immediately upon checkout, and then a guarantee that, when you actually _build_ the project, the current source is used to update those files, if necessary.

The first solution is better, as you still get real-time generation for new code elements, and if set up as a submodule in this repo, it _feels_ like their source code is in this repo, even though it's not really.

_But_, I bet there are already generators for exactly this specific use already available on nuget. Maybe even something built in already. I haven't looked, in recent history. _That's_ the best solution for use with our custom configuration manager, if any are indeed available.

dodexahedron: All that said, I think that can wait til later anyway. Trim-compatibility is nice and a good goal, but it's not really a development blocker to finish other things up - more like just a release-blocker.

dodexahedron:
Re-based on master as of a little while ago

tig (approved):

PR #3562, opened 2024-06-24
Fixes #2598. Trying to run self-contained application on Ubuntu 22 (Linux-x64), error regarding libdl.so

### LINKED ISSUE #2598: Trying to run self-contained application on Ubuntu 22 (Linux-x64), error regarding libdl.so
**Describe the bug**
I'm getting this error when attempting to run an application I built with .NET 7, Terminal.Gui 1.10.1 targeting Ubuntu 22:
```
Unhandled exception. System.Exception: Curses failed to initialize, the exception is: Unable to load shared library 'libdl.so' or one of its dependencies. In order to help diagnose loading problems, consider using a tool like strace. If you're using glibc, consider setting the LD_DEBUG environment variable.
   at Terminal.Gui.CursesDriver.Init(Action)
   at Terminal.Gui.Application.InternalInit(Func`1, ConsoleDriver , IMainLoopDriver , Boolean )
   at Terminal.Gui.Application.Init(ConsoleDriver , IMainLoopDriver )
   at Program.<Main>$(String[])
Aborted
```
After doing some research, I found this:
https://stackoverflow.com/a/75855054

I also know that v2 is being worked on, so I tried searching on Discussions and Issues for anything about libdl.so but didn't find anything so my apologies if this is already a known issue.

**To Reproduce**
1. Create a simple application using .NET 7 and Terminal.Gui 1.10.1
2. Publish as a single file, self-contained application targeting `linux-x64`
3. chmod +x, attempt to run the application on Ubuntu 22

**Expected behavior**
The application should run, displaying the simple application

 - OS: Ubuntu 22.04.2 LTS

BDisp:
## Fixes

- Fixes #2598

## Proposed Changes/Todos

- [x] Added NET8_0

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
@@ -256,7 +256,7 @@ namespace Unix.Terminal {
 		/// to avoid the dependency on libc-dev Linux.
 		/// </summary>
 		static class CoreCLR {
-#if NET7_0 || NET8_0
+#if NET6_0_OR_GREATER
 			// Custom resolver to support true single-file apps
 			// (those which run directly from bundle; in-memory).
 			//	 -1 on Unix means self-referencing binary (libcoreclr.so)
--- a/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
+++ b/Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
@@ -256,7 +256,7 @@ namespace Unix.Terminal {
 		/// to avoid the dependency on libc-dev Linux.
 		/// </summary>
 		static class CoreCLR {
-#if NET7_0
+#if NET7_0 || NET8_0
 			// Custom resolver to support true single-file apps
 			// (those which run directly from bundle; in-memory).
 			//	 -1 on Unix means self-referencing binary (libcoreclr.so)
--- a/Terminal.Gui/Terminal.Gui.csproj
+++ b/Terminal.Gui/Terminal.Gui.csproj
@@ -20,7 +20,7 @@
     <DebugType>portable</DebugType>
   </PropertyGroup>
   <PropertyGroup>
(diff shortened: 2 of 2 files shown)


Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
```diff
@@ -256,7 +256,7 @@ static class Mono {
 		/// to avoid the dependency on libc-dev Linux.
 		/// </summary>
 		static class CoreCLR {
-#if NET7_0
+#if NET7_0 || NET8_0
```
dodexahedron, line 259: I think you want `NET7_0_OR_GREATER` instead.

Although conditional compilation only matters if the consumer builds it from source.

dodexahedron (changes requested):
Suggested to use the symbol made for this scenario

Terminal.Gui/ConsoleDrivers/CursesDriver/UnmanagedLibrary.cs
dodexahedron: Oh hey. Also...

Even though it's v1, this is probably something that should be updated at some point to use LibraryImport, and conditional compilation is nice to be more visible, such as including or excluding things at the project level, or similar.

But also.... Several things that class does are in netstandard2.0 and thus can just use the built-in methods so this doesn't remain a maintenance burden, like the IsWindows and such. 🤷‍♂️ 

And parts that aren't supported in netstandard2.0 but exist in more recent .net can be polyfilled by various source generators out there on nuget, and then platform-specific parts can use more standard facilities from that. The Meziantou.Polyfill is excellent for that and only emits what's actually used, at compile-time.

BDisp: `LibraryImport` is not supported in all project targets, so I think it is not a good alternative.

image

BDisp: I will use `NET6_0_OR_GREATER` to support `net6.0` which is still currently supported (LTS).

tig (approved):

PR #3565, opened 2024-06-25
Fixes #3564. todo list tweaks

### LINKED ISSUE #3564: (Internal) Better regexes for TODO-list items
I made some simple changes to the R# todo-list tags to allow them to be used more places (such as end of line comments and in...

```csharp
 /* all
  * places
  * of */

 /* multi-line comments */
```

... as well as the full words for the tags and not just the abbreviated forms, for all of them, and better handling and elimination of non-interesting characters around them.

These changes also enable use of multiple tags on a single line, which is something I've wanted to do several times but either ended up putting them on separate lines or just not writing them.

The main remaining limitation that I opted not to address with the new regexes is that the "message" portion of the defined tags will contain any of the tags themselves that follow it on the same line, but the current ones already do that; They just currently can only catch one tag on the same line. So, no loss of functionality.

I also made some small tweaks to them that result in quicker abort and reduced backtracking which makes them a couple milliseconds faster even just on the 10-line test input text I validated them against, which means that it may actually result in a noticeable improvement to responsiveness in VS, especially for large files.

PR will be ready in a few minutes, if I don't get too distracted. 😅

dodexahedron:
Here we go.

All done with this.

Nice little enhancement to the todo list experience.

## Fixes

- Fixes #3564

## Proposed Changes/Todos

- [x] Make todo list regexes cover more cases
- [x] Make todo list regexes work on multi-line comment style
- [x] Enable support for multiple tags on one line
- [x] Added a second-priority naming rule for underscore tolerance

## Pull Request checklist:


tig (approved):

tig:
The build error is because of a test error I introduced in the hotkey PR. I'll fix asap.

BDisp:
> The build error is because of a test error I introduced in the hotkey PR. I'll fix asap.

@tig do you think that the hotkey PR, also broken the `Accept` event not being raising on a default button when the Enter key is pressed?

tig:
> > The build error is because of a test error I introduced in the hotkey PR. I'll fix asap.
> 
> @tig do you think that the hotkey PR, also broken the `Accept` event not being raising on a default button when the Enter key is pressed?

Seems fine to me? 

Note that if the TextFields have focus, it doesn't work because TextField has registered ENTER.

nuTWrak 1

BDisp:
> Note that if the TextFields have focus, it doesn't work because TextField has registered ENTER.

But `TextField` should return false if `Multiline` is false and thus the `Enter` key will be caught by the button.

dodexahedron:
Gah. That's not what I wanted. I forgot that's how github does that...

Hang on

dodexahedron:
There we go.

Re-based on v2_develop 61699d6af5b31109fc06a4d33fe8dc668a7c627f

dodexahedron:
I am gonna make them all search strings and comments, so they work for the unit test skips and such, as mentioned elsewhere. So hold off for a sec. Figured I'd do those here rather than this one and then changing them in another branch too. 🤷‍♂️

dodexahedron:
Okie dokie

All done with that.

Ready whenever.

dodexahedron:
It was difficult reading their nasty XML.

So I split those into a separate "layer" and manually organized the XML...

In any case, the layer thing is kinda nice, so you can turn blocks of settings on and off at will, in your environment, without having to mess around with solution local settings, which override everything and just make it difficult to sort out.

For example, if you don't like the settings from a layer, you can turn them off for yourself by just un-checking that layer in the options manager, rather than redundantly overriding them with settings you already set for your local computer settings.

tig (approved):

PR #3568, opened 2024-06-26
Fixes #3522. Adds `IDesignable`: Ability for Views to load design-time/demo data

### LINKED ISSUE #3522: Enable Views to expose "Design Time" default data?
By default, many Views have no good visual representation unless set with "data". E.g. `ListView`, `RadioGroup`, and `Slider`.

This Issue is to come up with a design and implementation that allows a "designer" (TGD) or "tester" (AllViewsTester/Unit Tests) component to say to a View "Hey, go into a good demo mode".

My current (probably lame) thinking is we'd do this by defining an interface that a View that supports such a thing could implement:

```cs
public interface ISupportsDesignMode
{
   // Call this to tell the View to load "demo data" 
   public bool LoadDemoData ();
}
```
 
## Background

Slider<> is doing strange things when setting it's `Text` property.

```csharp


    [Test]
    public void TestTextSliderString()
    {
        var s = new Slider<string>();
        ClassicAssert.IsEmpty(s.Options);

        var s2 = new Slider<string>();
        s2.Text = "I've got a lovely bunch of coconuts";
        ClassicAssert.IsEmpty(s2.Options);
    }
```

`Text` should be completely independent from `Options`.

It seems `Text` setter is putting data into `Options`.  This is problematic for many reasons.  It happens also with the `<int>` generic.  Notice how if you reverse the properties it will have different result.  

- Set Text first and you can overwrite with Options
- Set Options first and Text overwrites completely the Options

```csharp

    [Test]
    public void TestTextSliderInt()
    {
        var o = new SliderOption<int>("hi", new System.Text.Rune('d'), 32);
        var s1 = new Slider<int>
        {
            Text = "fff",
            Options = new List<SliderOption<int>> { o }
        };

        Assert.That(s1.Options.Single(), Is.EqualTo(o));

        var s2 = new Slider<int>
        {
            // When order is reversed
            Options = new List<SliderOption<int>> {o },
            Text = "fff",
        };

        // It fails
        Assert.That(s2.Options.Single(),Is.EqualTo(o));
    }
```

tig:
## Replaced by #3575

I don't know what rebase error I keep making, but I continue to eff my branches up. Arrrrg.


BDisp (approved):
LGTM. But I would like @tznind take a look on this. Thanks.

Terminal.Gui/View/ISupportsDesignMode.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Terminal.Gui;
+
+/// <summary>
+/// Indicates that the view supports design mode.
+/// </summary>
+public interface ISupportsDesignMode
```
tznind, line 6: What about IDesignable? Sounds snappier?

tznind (approved):
So this function updates the state of the View to an 'example' state?

This makes sense and means some code can come out of TGD ViewFactory.

Makes sense for the view tester too.

I like it!

dodexahedron:
Neat. I'll take a look at this one either later tonight or some time tomorrow.

dodexahedron:
Getting to this in a few minutes, once I go through the last few items at the top of my inbox.

Terminal.Gui/Views/Button.cs
```diff
@@ -177,4 +177,12 @@ protected override void UpdateTextFormatterText ()
             }
         }
     }
+
+    /// <inheritdoc />
+    public bool LoadDemoData (object ctx = null)
```
dodexahedron, line 182: Obviously the signature would need to match the proposal for the interface modification.\
Here's a potentially nice way to do this default behavior:

```csharp
    public bool LoadDemoData<TContext> (in TContext buttonText) where TContext : notnull
    {
        Title = buttonText switch
                {
                    string => buttonText,
                    _=> "_Button"
                };
        return true;
    }
```

Note that the parameter name does not have to match for it to be an interface implementation. Nice way to give a little extra guidance to the consumer via expressive code.

Terminal.Gui/View/IDesignable.cs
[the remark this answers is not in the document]
dodexahedron: The `notnull` constraint gives a compile-time warning, not an error, and only if nullable context is enabled for the consuming code.

dodexahedron (changes requested):
Looks pretty good to me.

But I made a couple of suggestions with code for some design-time and run-time improvements.

Terminal.Gui/View/IDesignable.cs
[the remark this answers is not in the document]
dodexahedron: An implementing class can override one, both, or none, when they have default implementations.

That also makes it easier for a derived type to revert to default behavior by calling the interface method explicitly, rather than as an implicit member method call.

And another thought I just had: Might want to make those virtual, too, since override is likely.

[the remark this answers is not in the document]
tig: Love!

tig:
Odd. WIth this new interface defn, intellisense no longer suggests members aren't implemented if they're not. Why is that?

Regarding other possible things to add to this interface: Not gonna do it until actually needed. First rule of API design is to not expose something publicly until there are at least 3 known use cases.

Terminal.Gui/Views/Button.cs
tig: image

BDisp: Try with a cast `string => (string)buttonText,`.

tig:
Ok, how's this?

Terminal.Gui/Views/Button.cs
dodexahedron: oh yeah. Forgot to say that the name of the type parameter is also not set in stone for implementations. Just have to match the same number of them and same constraints, but otherwise can call them whatever you like.

dodexahedron:
I think your original naming was better. Enable is kinda confusing since that's now a member of the View. I wouldn't expect SomeView.Enable() to load demo data.

dodexahedron:
> Regarding other possible things to add to this interface: Not gonna do it until actually needed. First rule of API design is to not expose something publicly until there are at least 3 known use cases.

Yep, agree there.

In any case, pretty much everything needed is already in the View class and derivatives anyway. Maybe we should take a refinement pass in beta or something to go through and annotate things around the library with appropriate/useful attributes like that?

Terminal.Gui/Views/Menu/MenuBar.cs
```diff
@@ -1594,11 +1594,9 @@ private MenuBar GetMouseGrabViewInstance (View view)
 
 
     /// <inheritdoc />
-    public bool LoadDemoData (object ctx = null)
+    public bool LoadDemoData<T> (T context)
```
dodexahedron: These have to include the where clause or they aren't actually implementations of the interface method.

dodexahedron: Only for the generic method, that is.

The reason it didn't warn you about that is because there is a default implementation, so the class is still valid, but now the default interface implementation is a duplicate on the class that can still be called via the interface like: `IDesignable.LoadDemoData<string>("")` for example.

UICatalog/Scenarios/AllViewsTester.cs
```diff
@@ -348,9 +348,9 @@ private View CreateClass (Type type)
             view.ColorScheme = Colors.ColorSchemes ["Base"];
         }
 
-        if (view.GetType().GetInterface ("IDesignable") is { } iface)
+        if (view is IDesignable iface)
```
dodexahedron, line 351: This check is always true. The explicit cast on line 343 makes it that way.

dodexahedron (changes requested):
Almost there.

I think good to go with these additional little tweaks.

UICatalog/Scenarios/AllViewsTester.cs
tig: Not getting what you're putting down here. If the View created on line 343 doesn't implement `IDesignable` the if statement on line 351 is false.

tig:
> I think your original naming was better. Enable is kinda confusing since that's now a member of the View. I wouldn't expect SomeView.Enable() to load demo data.

Now `EnableForDesign`

UICatalog/Scenarios/AllViewsTester.cs
dodexahedron: I'll look again when I'm back at my PC later. Could have been a brain fart. 🤷‍♂️

UnitTests/Views/ShortcutTests.cs
```diff
@@ -314,4 +314,259 @@ public void Focus_CanFocus_CommandView_Add_Tracks ()
         Assert.False (shortcut.CanFocus);
         Assert.True (shortcut.CommandView.CanFocus);
     }
+
+    [Theory]
+    //  0123456789
+    // " C  0  A "
+    [InlineData (-1, 0)]
```
dodexahedron, line 321: How about, instead of all these inlines:

```csharp
    [Theory]
    [CombinatorialData]
    [InlineData (-1, 0)]
    [InlineData (9, 0)]
    [AutoInitShutdown]
    public void MouseClick_Fires_Accept ([CombinatorialRange (0, 8)] int x, [CombinatorialValues (1)] int expectedAccept)

```

Or similar.

InlineData can still be used for one-off cases, as shown, and are explicit cases beyond the "combinatorial" cases, but do get de-duplicated if you accidentally specify any dupes.

Don't bother if you don't want to, though.

There are other things about the tests that are more important, but that's so far down my list right now. 😅

dodexahedron (commented):
I don't really have a lot to add beyond what we talked about in the issue already.

All I put in here was a suggestion/remindr about CombinatorialData for XUnit. Unimportant.

PR #3571, opened 2024-06-26
Fixes #3570 & #3561. Fixes a bunch of `EventArgs` sins

### LINKED ISSUE #3561: Rename `StateChangedEventArgs<T>` to `CancelEventArgs<T>`
I mis-named this. I originally thought it was just for representing the property that represented the primary state of a View, but that's way to restrictive. This class is useful for any event that is tracking the state of a property.

### LINKED ISSUE #3570: `TextField` prevents default `Button` from working in some cases
This test shows that `TextField` is preventing ENTER from invoking Accept on the default button.

```cs
[Fact]
public void Enter_Prevents_Default_Button_Accept ()
{
    var superView = new Window ();
    var tf = new TextField ();
    var button = new Button ()
    {
        IsDefault = true,
    };

    superView.Add (tf, button);

    var buttonAccept = 0;
    button.Accept += ButtonAccept;

    var textFieldAccept = 0;
    tf.Accept += TextFieldAccept;

    tf.SetFocus ();
    Assert.True (tf.HasFocus);

    superView.NewKeyDownEvent (Key.Enter);
    Assert.Equal (1, textFieldAccept);
    Assert.Equal (0, buttonAccept);

    button.SetFocus();
    superView.NewKeyDownEvent (Key.Enter);
    Assert.Equal (1, textFieldAccept);
    Assert.Equal (1, buttonAccept);

    return;

    void TextFieldAccept (object sender, CancelEventArgs e) { textFieldAccept++; }
    void ButtonAccept (object sender, CancelEventArgs e) { buttonAccept++; }
}
```

We def want `TextField` to support firing `Accept` on `Key.Enter`. However, not in every case.

I use this in `CharMap` for example.

But in the video below, the default button should be invoked when one of the TF's has focus.

## Background

> Do you think that the hotkey PR, also broken the `Accept` event not being raising on a default button when the Enter key is pressed?

Seems fine to me? 

Note that if the TextFields have focus, it doesn't work because TextField has registered ENTER.

nuTWrak 1

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3565#issuecomment-2192523217_

tig:
## Fixes

- Fixes #3561 
- Fixes #3570
- Fixes TextView the same way
- `OnAccept/Accept` now use `HandledEventArgs` instead of `CancelEventArgs`

## Proposed Changes/Todos

- [x] Fix

## Pull Request checklist:


BDisp (approved):

tig:
@BDisp - Please review.

Terminal.Gui/Views/TextView.cs
```diff
@@ -6050,13 +6063,15 @@ private void ProcessPaste ()
...
+    private bool? ProcessReturn ()
     {
         ResetColumnTrack ();
 
         if (!AllowsReturn || _isReadOnly)
         {
-            return false;
+            // By Default pressing ENTER should be ignored (OnAccept will return false or null). Only cancel if the
+            // event was fired and set Cancel = true.
+            return OnAccept () == false;
```
BDisp, line 6074: I have serious doubts here. If the text is read-only and `OnAccept()` returns `true`, is the text allowed to be changed? In my opinion it should not be allowed, even in an inherited class.

BDisp (approved):

Terminal.Gui/Views/TextView.cs
tig: Fixed.

Terminal.Gui/View/CancelEventArgs.cs
```diff
@@ -12,14 +12,67 @@ namespace Terminal.Gui;
...
+/// <summary>
+///     <see cref="EventArgs"/> for events that convey changes to a property of type <typeparamref name="T"/>.
+/// </summary>
+/// <typeparam name="T">The type of the value that was part of the change being canceled.</typeparam>
+/// <remarks>
+///     Events that use this class can be cancellable. Where applicable, the <see cref="CancelEventArgs.Cancel"/> property
+///     should be set to
+///     <see langword="true"/> to prevent the state change from occurring.
+/// </remarks>
+public class EventArgs<T> : EventArgs where T : notnull
```
dodexahedron, line 45: XmlDoc needs a look for accuracy.

I'm also not a fan of the name, but I'd much rather have this code all in there and worry about the name later at the moment.

Terminal.Gui/View/CancelEventArgs.cs
```diff
@@ -12,14 +12,67 @@ namespace Terminal.Gui;
...
+{
+    /// <summary>Initializes a new instance of the <see cref="EventArgs{T}"/> class.</summary>
+    /// <param name="currentValue">The current value of the property.</param>
+    /// <typeparam name="T">The type of the value.</typeparam>
+    public EventArgs (ref readonly T currentValue) : base ()
```
dodexahedron, line 50: I hate this warning, because it's so low-importance, usually... I sometimes suppress it, but it does also have a legitimate underlying point that tends to mean a static factory method that calls an internal or private constructor might be considered.

I think suppressing right here is probably fine, yeah?

Terminal.Gui/View/CancelEventArgs.cs
```diff
@@ -12,14 +12,67 @@ namespace Terminal.Gui;
...
+/// <summary>
+///     <see cref="EventArgs"/> for events that convey changes to a property of type <typeparamref name="T"/>.
+/// </summary>
+/// <typeparam name="T">The type of the value that was part of the change being canceled.</typeparam>
+/// <remarks>
+///     Events that use this class can be cancellable. Where applicable, the <see cref="CancelEventArgs.Cancel"/> property
+///     should be set to
+///     <see langword="true"/> to prevent the state change from occurring.
+/// </remarks>
+public class CancelEventArgsStruct<T> : CancelEventArgs where T : notnull
```
dodexahedron, line 68: This still ended up being necessary after doing the ref readonly, if you dropped the notnull, or was that attempted, specifically?

dodexahedron: I was hoping that would make this unnecessary altogether. :(

Terminal.Gui/View/ViewText.cs
```diff
@@ -69,23 +69,22 @@ public virtual string Text
...
     {
-        TextChanged?.Invoke (this, args);
+        TextChanged?.Invoke (this, EventArgs.Empty);
     }
 
     /// <summary>
     ///     Text changed event, raised when the text has changed.
     /// </summary>
-    public event EventHandler<CancelEventArgs<string>> TextChanged;
+    public event EventHandler<EventArgs> TextChanged;
```
dodexahedron, line 87: Did you mean to use our `EventArgs<T>` you created?

If not, then the event can just be `public event EventHandler? TextChanged;`[^1]. That's actually exactly how that same event is defined in windows forms. Anything derived from EventArgs is therefore accepted.

WPF defines a much richer event [delegate](https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.textchangedeventhandler?view=windowsdesktop-8.0) and [eventargs type](https://learn.microsoft.com/en-us/dotnet/api/system.windows.controls.textchangedeventargs?view=windowsdesktop-8.0) for it, which includes some goodies that make undo handling easier. But that's also got stuff not relevant to TG. Just another reference.


[^1]: The EventHandler delegate [already uses EventArgs for e](https://github.com/dotnet/runtime/blob/v8.0.6/src/libraries/System.Private.CoreLib/src/System/EventHandler.cs#L6). The generic is fine, but will only make a difference if T of `EventArgs<T>` is a value type, since reference types share the runtime concrete implementation.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -185,15 +185,19 @@ public IListDataSource Source
...
+//                HideList ();
+////                ShowHideList (string.Empty);
                 SetNeedsDisplay ();
             }
         }
     }
 
-    /// <summary>The currently selected list item</summary>
+    /// <summary>The text of the currently selected list item</summary>
     public new string Text
```
dodexahedron, line 196: Why `new`?

dodexahedron (commented):
Side commentary mostly, but a couple of small things.

Also, some warnings at a couple of call sites for event dispatchers should be addressed by constructing the object on a separate line and passing it explicitly as a ref.

Terminal.Gui/Application/Application.cs
```diff
@@ -308,7 +308,7 @@ internal static void ResetState (bool ignoreDisposed = false)
         SupportedCultures = GetSupportedCultures ();
         _mainThreadId = Thread.CurrentThread.ManagedThreadId;
         _initialized = true;
-        InitializedChanged?.Invoke (null, new (false, _initialized));
```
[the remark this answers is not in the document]
dodexahedron: Otherwise, if an event is always going to be raised without caring about sender, a delegate that does not include sender should be used, or else it's unclear to subscribers that it will just be ignored.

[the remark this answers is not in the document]
dodexahedron: At minimum, using the current Top or something else representing a high level relevant source for the event may also be something to consider, for the call to Invoke.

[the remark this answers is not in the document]
tig: > Not important at the moment, but a general thought:
> 
> Some of these methods should probably have an optional `object? sender = null` parameter, as well, so it can be passed in the event, if specified, since handlers subscribed to it could potentially want to know who originally prompted the event to be raised.

A communications request, @dodexahedron: I really struggle to parse your comments when you use language like "some of these methods". I can't tell what "these" refers to. In this case, the code lines you comment on has two "methods" in them: GetSupportedCultures() and Invoke(). GetSupportedCultures() is completely irrelevant, so I'm pretty sure that's not what you  are referring to. And `Invoke()` is provided by dotnet, so there's no way for me to add an "optional parameter". 

See what I mean?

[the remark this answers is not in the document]
tig: > object? sender = null

`Application.Init/Shutdown` is decoupled from `View`. So `Top` makes no sense. The only object that might make sense is `Application`, but it's a static/singleton so...

Terminal.Gui/View/CancelEventArgs.cs
tig: > I hate this warning, because it's so low-importance, usually... I sometimes suppress it, but it does also have a legitimate underlying point that tends to mean a static factory method that calls an internal or private constructor might be considered.
> 
> I think suppressing right here is probably fine, yeah?

Yeah. Annoying for sure.

Terminal.Gui/View/CancelEventArgs.cs
tig: Dead code. I was experimenting.

Terminal.Gui/View/ViewText.cs
tig: Yep. already fixed.

tig: No. I changed the semantics of the TextChanged event.

Previously it conveyed old/new but was (clearly) not cancelable. Earlier in this PR I (stupidly) made it `EventHandler<CancelEventArgs<string>>` to make it convey old/new as it had before. By using `CancelEventArgs` I made it confusing.

`Text` can be HUGE. I was an error to originally think it was ok for `TextChanged` to convey EITHER the old or new text as there's no simple way to avoid copies of the string. And some subs may not even care about the new text, much less the old. If a sub wants the new text they can simply access `((View)sender).Text`.  And if a sub wants the old text, they should be using `TextChanging`.

Terminal.Gui/Views/ComboBox.cs
tig: `ComboBox` does a lot of weird things like this. It's quite fragile.

One of the reasons I suggested this:

- #2404

I'm not touching it any more than I already have because I'm terrified of breaking it.

Terminal.Gui/View/ViewText.cs
tig: > public event EventHandler? TextChanged;

Done.

tig:
Oops, I did it: 

Instead of effing around with `CancelEventArgs<bool?>` correctness, I took the challenge of refactoring `CheckBox` to NOT use `bool?` for it's 3 states, and instead defined an enum `CheckState` with three members. 

Push coming soon... Massive #s of changes but it's sooo much better.

BDisp:
> Instead of effing around with

Don't forget to do it for the `MenuItem` as well. Thanks.

tig:
> > Instead of effing around with
> 
> Don't forget to do it for the `MenuItem` as well. Thanks.

I (or someone) will do so as part of

- #2975 

Spending any time on the existing Menu code is not worth it.

tig:
> > > Instead of effing around with
> > 
> > 
> > Don't forget to do it for the `MenuItem` as well. Thanks.
> 
> I (or someone) will do so as part of
> 
> * [Modernize `Menu` - Base on `Bar` and address litany of issues #2975](https://github.com/gui-cs/Terminal.Gui/issues/2975)
> 
> Spending any time on the existing Menu code is not worth it.

Quick note on this: One of the beauties (IMO) of `Bar` is a "checkable menu item' is literally just a `Checkbox`.

dodexahedron:
> Oops, I did it:
> 
> Instead of effing around with `CancelEventArgs<bool?>` correctness, I took the challenge of refactoring `CheckBox` to NOT use `bool?` for it's 3 states, and instead defined an enum `CheckState` with three members.
> 
> Push coming soon... Massive #s of changes but it's sooo much better.

Oooo. Exciting.

I'll pause working on the analyzer nuget for a bit to take a look at this.

dodexahedron:
> > > > Instead of effing around with
> > > 
> > > 
> > > Don't forget to do it for the `MenuItem` as well. Thanks.
> > 
> > 
> > I (or someone) will do so as part of
> > 
> > * [Modernize `Menu` - Base on `Bar` and address litany of issues #2975](https://github.com/gui-cs/Terminal.Gui/issues/2975)
> > 
> > Spending any time on the existing Menu code is not worth it.
> 
> Quick note on this: One of the beauties (IMO) of `Bar` is a "checkable menu item' is literally just a `Checkbox`.

Sounds like how composable things are in WPF, too.

I like. 👍

I'll pull the branch and start review momentarily.

dodexahedron:
When I saw the title of this in my inbox, I couldn't click on it fast enough 😆 

Waiting for my machine to reboot RN because I postponed updates that have now reached the deadline and forced install, so will still be a few minutes before I begin review.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -423,12 +422,12 @@ private bool CancelSelected ()
 
             if (SelectedItem > -1 && _listview.Source?.Count > 0)
             {
-                _search.Text = _text = _listview.Source.ToList () [SelectedItem].ToString ();
+                Text = _listview.Source.ToList () [SelectedItem].ToString ();
```
dodexahedron, line 425: Need a null conditional access on the array element:

```csharp
Text = _listview.Source.ToList () [SelectedItem]?.ToString ();
```

I also don't love that we use `SelectedItem` to mean `SelectedIndex`.

Terminal.Gui/Views/RadioGroup.cs
```diff
@@ -97,7 +97,7 @@ public RadioGroup ()
                         {
                             SelectedItem = (int)ctx.KeyBinding?.Context!;
 
-                            return !OnAccept();
+                            return OnAccept () != false;
```
dodexahedron, line 100: ~~return OnAccept () != false;~~

```csharp
return OnAccept () is not false;
```

Combines the missing null check with a non-overridable type check.

If you want to be clearer, since the tri-state because of null is not all that apparent in a vacuum, you could do this instead:

```csharp
return OnAccept () is true or null;
```

Either way works with the same result.

dodexahedron (changes requested):
A few things, mostly around nullability.

It's probably time for us to turn it on project-wide, honestly.

I didn't mention it in in-line comments, but another nullability problem throughout this and existing code is that event dispatchers haven't been checking for null on their EventArgs parameters, before using them.

Not all of them directly use the EventArgs passed to them, but the ones that do need to take that precaution.

Issue #2485, opened 2023-04-04
`Wizard` is broken - Needs revamping


tig:
Wizard needs to be upgraded to the modern v2 architecture. Right now it's basically broken in v2_develop.

PR #3575, opened 2024-06-29
Fixes #3522. Adds `IDesignable`: Ability for Views to load design-time/demo data

### LINKED ISSUE #3522: Enable Views to expose "Design Time" default data?

tig:
## Fixes

- Fixes #3522

## Proposed Changes/Todos

- [x] Define interface
- [x] Fix `Slider`
- [x] Fix `RadioGroup`
- [x] Fix `ListView`
- [x] Fix `MenuBar`
- [x] ...
- [x] Add unit tests
- [x] Update docs

## Pull Request checklist:


tig:
I am doing rebase wrong, clearly.

I had to re-create this PR by manually copying my changes from #3568 because it was screwed up somehow.

I have now done this 3-4 times and I have no clue what I'm doing wrong.

dodexahedron:
I'll take a quick look at the current state of the branch and see if I can piece together what went wrong. No promises without the original commits still existing, though, which is pretty unlikely unless you went out of your way to keep doing everything on new branches with each push.

dodexahedron:
Taking a look at the commit graph, my first suspicion is those merges.

Dollars to donuts I'd bet those are at least a significant part of the difficulty, because they move the common parent to a place where you're probably going to get strange things like merge conflicts that aren't actually any different between the two copies or duplicate commits, or other annoying/strange junk like that.

That gets worse and worse if other merge activity occurs to the branch you branched from (so v2_develop), that you're trying to rebase your work branch on, after you've also merged _from_ v2_develop into your working branch. It makes a diamond dependency, sorta, and it just tries its best to replay the deltas for you. But especially if you're using the default git diff algorithm, that gets......weird.....

Setting the diff/merge algorithm to histogram sometimes can help a little bit, at least in making the diffs make a bit more sense, but the solution is to never merge outward.

Outward meaning from a branch closer to the root to a branch farther from root.

In general, once you branch, the only merges that branch should ever be involved in are merges into an ancestor with a common base that is no older than the original branch point, or as the recipient of merges from descendent branches of itself.

To "merge" outward (so, to apply changes from an ancestor branch to your working branch) you rebase, effectively changing the branch point, which replays the diffs on the new base, giving you the opportunity to get involved if there are commits it can't auto-apply.

When doing a rebase, an interactive rebase is a good opportunity to squash, drop, or otherwise modify commits of your working branch, since you're already re-writing history for your branch.

However, if the branch you're working on has descendents, ***do not force push a rebase over it***, because you will orphan the other branch, which can be a bitch to fix sometimes because there's no longer a common base. Push to a new branch, instead, if there's a descendent branch.

You can actually _always_ push to a new branch, after a rebase, if you want to preserve the old history for any reason. You would just have to update the PR or create a new one if you did that.

But yeah... Anything that pulls in commits from an ancestor before you merge back into that ancestor has the potential (not a guarantee) to create conflicts and other fun. You can often get away with it, especially if you didn't touch any of the same files, but the nature of merges means that, when it does bite you, it's usually going to bite harder than your average merge conflict.

dodexahedron:
A thought I just had that I've never tried but sounds valid in my head is that, if you have done outward merges like that... If you do an interactive rebase (if it'll let you), you could try to drop the commits for the merges from the ancestor branch into your branch. It _should_ work, even if your code is dependent on what was merged in, because that's all already there, already, at the new ref you're rebasing on.

BDisp (approved):

dodexahedron (commented):
Taking a look at it beyond what we were exploring in the associated issue, around nitpicky stuff with generics.

First thing that comes to mind, just scanning the file list: Are those all the places you wanted to have this, for now (ie no other view types)?

Looking at the actual code now.

Terminal.Gui/View/IDesignable.cs
```diff
@@ -0,0 +1,23 @@
...
+public interface IDesignable
+{
+    /// <summary>
+    ///     Causes the View to enable design-time mode. This typically means that the view will load demo data and
+    ///     be configured to allow for design-time manipulation.
+    /// </summary>
+    /// <param name="context">Optional arbitrary, View-specific, context.</param>
+    /// <typeparam name="TContext">A non-null type for <paramref name="context"/>.</typeparam>
+    /// <returns><see langword="true"/> if the view successfully loaded demo data.</returns>
+    public bool EnableForDesign<TContext> (in TContext context) where TContext : notnull => EnableForDesign ();
```
dodexahedron, line 15: You know...

Can't believe this didn't occur to me while we were talking about it before.

`ref readonly` instead of `in` (yay c# 12!) might actually be the simplest answer here.

The difference is `ref readonly` is mandatory ref without a defensive copy at the callsite, which results in the behavior being the same for reference and value types.

It also carries the restrictions of not modifying the reference, as the `in` keyword does.

It requires no code change at the call sites, whether they currently include the `in` or not. In the former case, there's no change at all. In the latter case, the compiler emits a warning to notify you that the semantics are explicitly ref-always, rather than the C# default of value-unless-told-otherwise. Adding `in` or `ref` at the callsite makes the warning go away, as was the intent of the warning by design.

I think we can then get rid of the notnull, though I'll have to double-check that rule.

But I'm more than happy to tackle that and test it out in a separate branch if you'd rather get this one wrapped up for now.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -243,7 +243,7 @@ public virtual bool Expand ()
     public event EventHandler Expanded;
 
     /// <inheritdoc/>
-    protected internal override bool OnMouseEvent  (MouseEvent me)
+    protected internal override bool OnMouseEvent (MouseEvent me)
```
dodexahedron, line 246: I still really dislike the name of that class (`MouseEvent`). It's not an event. It's event arguments/state. :man_shrugging: 

Irrelevant to this PR of course.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -983,4 +983,14 @@ private void SetInitialProperties (ComboBox container, bool hideDropdownListOnCl
             AddCommand (Command.LineUp, () => _container.MoveUpList ());
         }
     }
+
+    /// <inheritdoc />
+    public bool EnableForDesign ()
+    {
+        var source = new ObservableCollection<string> (["Combo Item 1", "Combo Item two", "Combo Item Quattro", "Last Combo Item"]);
+        SetSource (source);
```
dodexahedron, line 991: Also not relevant to this PR, and mentioned in an actually related PR, but just noting:

The SetSource method needs a bit more tweaking, precisely because of situations like this. The collection can be replaced wholesale, by a reference to a new collection, but anyone subscribed to the old collection's events won't ever know about it.

Easy to fix, though there are multiple ways to do so.

Again, not directly relevant for this PR. Just keeping it visible.

Terminal.Gui/Views/Menu/MenuBar.cs
```diff
@@ -1591,4 +1591,177 @@ private MenuBar GetMouseGrabViewInstance (View view)
...
+        if (context is not Func<string, bool> actionFn)
+        {
+            actionFn = (s) => true;
+        }
+
+        Menus =
+        [
+            new MenuBarItem (
+                             "_File",
+                             new MenuItem []
```
dodexahedron, line 1608: Can use a collection initializer expression here, instead of the array constructor, for free compiler optimizations, if you want, just like is already being done for the `Menus` property already. Might as well. :man_shrugging: 

Plus that'll result in less nesting, which I guess is a bonus.

UICatalog/Scenarios/AllViewsTester.cs
```diff
@@ -348,49 +348,17 @@ private View CreateClass (Type type)
             view.ColorScheme = Colors.ColorSchemes ["Base"];
         }
 
-        // If the view supports a Text property, set it so we have something to look at
-        if (view.GetType ().GetProperty ("Text") != null)
+        if (view is IDesignable designable)
```
dodexahedron, line 351: YAY

Thank you.

So much nicer. Even if it's "just" in samples, this gives me warm fuzzies. :slightly_smiling_face:

dodexahedron (approved):
Looks good to me.

A couple of super minor opportunistic suggestions, but good to go as-is if you don't want to bother.

tig:
> Taking a look at it beyond what we were exploring in the associated issue, around nitpicky stuff with generics.

You are confusing two issues.

> 
> First thing that comes to mind, just scanning the file list: Are those all the places you wanted to have this, for now (ie no other view types)?

Any view that could show meaningful state could support this. There's no pressing need to hit them all now. 

I imagine when we get to 

- #2485 

, for example, we'll want to add this to it to help development/testing.

Terminal.Gui/View/IDesignable.cs
tig: I changed it to `ref readonly`. Left `notnul` in.

Terminal.Gui/Views/Menu/MenuBar.cs
tig: Menu is getting re-written. Not going to touch this code until then.

tig:
> Taking a look at the commit graph, my first suspicion is those merges.
> 

Back when I simply did `git merge` and wasn't trying to be a good boy and `rebase`, I never had these issues. This is def something caused by how I'm rebasing. And it's probably just something stupid.

Here is my typical workflow.

When I want to create a new PR to address an issue I do

```ps1
git checkout v2_develop
git pull upstream v2_develop
# push to my fork if needed, so it matches
git push 
# create new branch
git checkout -b v2_xxx-branch
```

Then I work, and eventually do

```ps1
git add .  ; git commit -m "commit msg"
git push
```

And I use github.com to create the remote `v2_xxx-branch`

Then, if, as I'm working, `v2_develop` changes, I want those changes in `v2_xxx-branch`, so I

```ps1
git checkout v2_develop
git pull upstream v2_develop
git push

git checkout v2_xxx-branch

#double check I've pushed my latest and have the latest of my branch
git push
git pull origin v2_xxx-branch
git rebase --onto v2_develop

# pull branch down and re-apply changes
git pull

# fix any merge issues
git mergetool

# commit
git add . ; git commit -m "rebased with v2_develop"
```

This never feels right, so whatever I'm doing wrong is in the stuff around `git rebase --onto v2_develop`.

BDisp:
> Then, if, as I'm working, `v2_develop` changes, I want those changes in `v2_xxx-branch`, so I
> 
> ```powershell
> git checkout v2_develop
> git pull upstream v2_develop
> git push
> 
> git checkout v2_xxx-branch
> 
> #double check I've pushed my latest and have the latest of my branch
> git push
> git pull origin v2_xxx-branch
> git rebase --onto v2_develop
> 
> # pull branch down and re-apply changes
> git pull
> 
> # fix any merge issues
> git mergetool
> 
> # commit
> git add . ; git commit -m "rebased with v2_develop"
> ```

I think the problem is in the `git pull origin v2_xxx-branch` if you have already some local commit not yet pushed to the `origin`. So I recommend that you push all your local commits to the `origin` or don't pull from the `origin` at all. But if you end up with a force push avoid it because it would may delete any changes that you may want to preserve. I'll recommended using `merge` instead of `rebase` because when a PR is merged into the `v2_develop` all do smoothly. I confess that I do all that work using the VS2022 UI and only on situations more complicated I use the command line.

BDisp:
Sorry, ignore the local commits not yet pushed because you already do a `git push` before the `git pull origin v2_xxx-branch`. Maybe the `git pull origin v2_xxx-branch` isn't really needed because the `origin` is already updated.

BDisp:
Maybe the culprit is the `git pull` after the `git rebase --onto v2_develop`. During this you probably need to resolve any conflicts and only after they are fixed you do `git add . ; git commit -m "rebased with v2_develop"` and then push it. If you are pulling without the rebase is finished, probably you corrupting your origin.

dodexahedron:
> > Taking a look at it beyond what we were exploring in the associated issue, around nitpicky stuff with generics.
> 
> You are confusing two issues.

I suspected I might be.

> 
> > First thing that comes to mind, just scanning the file list: Are those all the places you wanted to have this, for now (ie no other view types)?
> 
> Any view that could show meaningful state could support this. There's no pressing need to hit them all now.
> 
> I imagine when we get to
> 
> * [Refactor Wizard to use View v2 #2485](https://github.com/gui-cs/Terminal.Gui/issues/2485)
> 
> , for example, we'll want to add this to it to help development/testing.

Sounds like a good plan to me. 👍

dodexahedron:
> > Taking a look at the commit graph, my first suspicion is those merges.
> 
> Back when I simply did `git merge` and wasn't trying to be a good boy and `rebase`, I never had these issues. This is def something caused by how I'm rebasing. And it's probably just something stupid.
> 
> Here is my typical workflow.
> 
> When I want to create a new PR to address an issue I do
> 
> ```powershell
> git checkout v2_develop
> git pull upstream v2_develop
> # push to my fork if needed, so it matches
> git push 
> # create new branch
> git checkout -b v2_xxx-branch
> ```
> 
> Then I work, and eventually do
> 
> ```powershell
> git add .  ; git commit -m "commit msg"
> git push
> ```
> 
> And I use github.com to create the remote `v2_xxx-branch`
> 
> Then, if, as I'm working, `v2_develop` changes, I want those changes in `v2_xxx-branch`, so I
> 
> ```powershell
> git checkout v2_develop
> git pull upstream v2_develop
> git push
> 
> git checkout v2_xxx-branch
> 
> #double check I've pushed my latest and have the latest of my branch
> git push
> git pull origin v2_xxx-branch
> git rebase --onto v2_develop
> 
> # pull branch down and re-apply changes
> git pull
> 
> # fix any merge issues
> git mergetool
> 
> # commit
> git add . ; git commit -m "rebased with v2_develop"
> ```
> 
> This never feels right, so whatever I'm doing wrong is in the stuff around `git rebase --onto v2_develop`.

Here's what I do when working on a github repo that I have a fork of, that helps avoid unexpected weirdness:

 * First, sync my fork's v2_develop, on github.
 * Fetch locally to get the updated index.
 * Branch from head of v2_develop.
 * Do my work.
 * If at any point I want to rebase on v2_develop again:
    * Sync my fork on github again and fetch to get the index again
    * Do the rebase
 * Continue working and repeat the above step whenever. Usually also once more before submitting a PR, to make it no-conflict auto-mergeable
 * Submit the PR for dodexahedron/Terminal.Gui/branchI'mworkingOn to gui-cs/Terminal.Gui/v2_develop

dodexahedron:
If I drop or squash anything during the rebases above, I sometimes will push a new branch, to preserve the old one, just in case, and if I remember to do so.

dodexahedron:
This step:

```
git pull origin v2_xxx-branch
```

Has the potential to cause a merge that _could_ cause unexpected annoyances, since the default is to merge from your remote into your local, if they differ.

You can change that behavior default to rebase (what I have set as the default behavior) or to fast-forward only, if you want, or you can specify it at pull-time with --rebase=[base] or --ff-only. Both will avoid a merge.

However, a git checkout already does a pull. It's a shortcut for a `git switch` and a `git pull`. `git pull` itself is already just a shortcut for `git fetch` and `git merge`, by default, which is where the auto-merge into local comes from.

If I do it at the command line, I usually use `git switch` rather than `git checkout` just to be careful that nothing I didn't explicitly intend happens. That merge that happens on a pull/checkout replays the remote on top of your working tree, which is probably not what you actually want, especially if you have dropped/squashed/modified any commits that were common between them, before.

dodexahedron:
> > Then, if, as I'm working, `v2_develop` changes, I want those changes in `v2_xxx-branch`, so I
> > ```powershell
> > git checkout v2_develop
> > git pull upstream v2_develop
> > git push
> > 
> > git checkout v2_xxx-branch
> > 
> > #double check I've pushed my latest and have the latest of my branch
> > git push
> > git pull origin v2_xxx-branch
> > git rebase --onto v2_develop
> > 
> > # pull branch down and re-apply changes
> > git pull
> > 
> > # fix any merge issues
> > git mergetool
> > 
> > # commit
> > git add . ; git commit -m "rebased with v2_develop"
> > ```
> 
> I think the problem is in the `git pull origin v2_xxx-branch` if you have already some local commit not yet pushed to the `origin`. So I recommend that you push all your local commits to the `origin` or don't pull from the `origin` at all. But if you end up with a force push avoid it because it would may delete any changes that you may want to preserve. I'll recommended using `merge` instead of `rebase` because when a PR is merged into the `v2_develop` all do smoothly. I confess that I do all that work using the VS2022 UI and only on situations more complicated I use the command line.

That step stood out to me, too, because of the implicit merge that happens if history differs.

Check out the steps I posted a couple comments up from here for what I do that avoids any of these issues while using the recommended rebase/branch outward, merge inward flow.

I most often use GitKraken to interact with it, but it respects your git configuration and also can show you exactly what it did to achieve an operation, which is part of how I originally learned pieces of how it all works, a while back.

PR #3585, opened 2024-07-07
Fixes #3303. Makes `Thickness` a `record struct`

### LINKED ISSUE #3303: `Thickness` should not be a reference type
Thickness shouldn't be a class/reference type.

TL;DR at the bottom. The majority of everything between here and that is technical backing of that assertion, in my characteristically extremely terse style. (Do I need to provide a /s?)

First off, **it's still good stuff** and I support its existence and the fact that it should be a familiar concept for devs coming to Terminal.Gui from WPF.

### The TL;WDYTWTRT[^1] Version:
<details><summary>Expand if you're bored</summary>
<p>
But it has unfortunate and disproportionate performance impact just from being a class and not a struct.

Note that System.Windows.Thickness is a struct, even though drawing in a Windows desktop app involves even hotter access to this kind of data than a Terminal.Gui app typically would, by a pretty wide margin, so that alone should probably be enough to justify it without what follows:

For one thing, nearly everything that class does is a value semantic, including the additions made to it in #3295.

So, it almost definitely should be a `struct` or (more preferably) a `record struct`, unless there's a _really_ good reason I'm unaware of for why it isn't.

As a minor design comment, the whole type essentially wraps a Rectangle-like structure, but with the slight behavioral differences necessary to work its magic as intended. Thus, it could easily be implemented exactly like that - having a private Rectangle for its underlying value, but exposing things as it currently does (and not exposing that Rectangle at all).

#### But the main problem is that it lives on the heap.

Every operation involving it means we can't do that operation entirely on the stack and have run to the heap (which means methods involving it miss out on some optimization possibilities), and any time a new one is created, even ephemerally, it's _another_ heap allocation, at minimum. In addition to all that, it also means this class is responsible for generating a lot of heap garbage for the GC (which also means updating references on containing objects, on top of anything else).

Any time it exists anywhere in code, access to it automatically requires indirection to the heap, extremely likely (always, for practical purposes), meaning you lose a supremely important thing: cache locality. And then, that compounds with the fact that not a single View is going to be sequentially allocated on the heap, either, meaning truly random access.

If an operation can stay entirely on the stack, you significantly increase cache locality and might not even leave L1 cache. And that's major. Even "resorting" to L2 is typically at least 4x slower than L1, and is typically over _two orders of magnitude_ faster than a single trip to main memory, which use of heap objects greatly increases the chance of, along with a whole bunch of other environment-/platform- dependent realities. And use of a value type, where prudent, at least mitigates most of them.

Even a certain amount of unnecessary copy activity on the stack from sub-optimal use of a value type can pretty easily outperform heap access to an otherwise identical reference type, due to this and various other language, compiler, runtime, and hardware realities.

A _real_ bear of it all that also compounds with all that cache stuff is that reference types are not collocated with their enclosing types. An object containing a Thickness field (either explicit or implicit via an auto-property) is already a pointer to a heap object, and then any reference members of that heap object are themselves pointers to _other_ heap objects. Value types are stored in-line with their enclosing type.

If you want to access a Thickness property of a View, you have the dereference of the View, at one point on the heap, and then the dereference of the Thickness, elsewhere on the heap, before you can even interact with it. And the GC can and will move things around, when it feels like it, and when the heap has become fragmented (which this type contributes to).

And mutability of it is inconsistent. With some operations, you might not even be getting that object back, but a newly-constructed (and therefore heap-allocated) object (an immutable behavior). That new object has to have its values copied in, at _some point_.

Even if you passed an equivalent struct by value, that simple fact alone already means one copy is paid for at the CIL level, and likely 20 or more as well at the machine level, _on top of that_ (meaning multiplicative cost), and that's ***just*** the literal copying of the values, assuming it all happens atomically (which isn't even remotely guaranteed). Consider a simple example for conceptual proof: Why is boxing a value type so expensive that analyzers warn you to help you avoid it and why is it one of the things always mentioned in dotnet performance tuning articles? The main reason is all the above, implicitly started from the first HeapAlloc (or equivalent platform calls on non-Windows platforms) and just getting worse from there. 

AND, as a concurrency concern, that object (which, again, is expected to have reference, not value, semantics) will have field values (which are currently public instance fields, which is also a no-no for reference types) not equal to the values of the object referenced by the containing object, but without static analysis of that situation being possible. Classic race conditions are the basic exposed issue there, with of course undefined effects possible.

And more, but I realized this is probably documented and I can just dump a link.... And yep! At least the c# and dotnet reasons are, anyway (the cache stuff is just something to be aware of because it is important - especially in hot code.

So, with the disclaimer that is hasn't been kept fully up-to-date with the language and the runtime, see https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct .

For how this type is designed, how it behaves, and how it is used and expected to be used, even that doc - which is almost old enough to drive in the US - very strongly supports this being a value type. Plus, some minor tweaks to it and a small number of places it is used (such as passing by reference when relevant) make the case even stronger. But it still likely would perform better even without those changes. And that's even if it stays 4 integers, meaning it's still half the size of the suggested max struct size, which is 4 times native pointer size (so 32 bytes on a 64-bit platform or 16 bytes on a 32-bit platform - Thickness is 16 bytes). The size recommendation is actually the main "outdated" part of that document... But the underlying values are the most minor part of the whole thing (aside from the fields being public).

I think I even mentioned on at least one other tangentially-related issue someone else was working on, a month or two ago, how I had performed the conversion to `record struct` with very little other work needed, didn't I? Or maybe it was commentary on one of my own? 😅 I do know I've at least started it before stopping myself more than once in the past 3 or 4 weeks (including the one I'm actively working on right now), mostly due to both high conflict probability and being way out of scope for existing issues (except maybe the general cleanup issues, but this is too important/has too broad an impact to go there).
</p>
</details> 

### TL;ANGTFT[^2] Points, justified in the above ramblings:

 - Thickness should be a `record struct` or a `readonly record struct` if immutability is wanted.
 - It may benefit from its underlying value being a Rectangle, depending on how ref-y you want to get.
   - If not that, storing its values in a Vector4 or a more advanced type like `Vector128<int>` could possibly be beneficial, especially for certain operations (like growing all 4 underlying values).
   - Or all sorts of options, none of which matter as much as Thickness itself being a struct.
 - Even a surprising amount of wasteful stack copies due to sub-optimal passing of structs is STILL faster than even one copy of an equivalent heap object.
 - All the other compiler-generated goodies you get with a record struct.
 - That was the TL;DR? Dude...

[^1]: Too Long; Who Do You Think Wants To Read That
[^2]: Too Long; Ain't Nobody Got Time For That

tig:
## Fixes

- Fixes #3303

## Proposed Changes/Todos

- [x] `class` -> `record struct`
- [x] Fix anything that breaks
- [x] Use `Vector4` under the covers per @dodexahedron's suggestion

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drawing/Thickness.cs
+++ b/Terminal.Gui/Drawing/Thickness.cs
@@ -14,23 +14,18 @@ namespace Terminal.Gui;
 ///         frame,
 ///         with the thickness widths subtracted.
 ///     </para>
-///     <para>Use the helper API (<see cref="Draw(Rectangle, string)"/> to draw the frame with the specified thickness.</para>
+///     <para>
+///         Use the helper API (<see cref="Draw(Rectangle, string)"/> to draw the frame with the specified thickness.
+///     </para>
+///     <para>
+///         Thickness uses <see langword="float"/> intenrally. As a result, there is a potential precision loss for very
+///         large numbers. This is typically not an issue for UI dimensions but could be relevant in other contexts.
+///     </para>
 /// </remarks>
 public record struct Thickness
 {
-    private Vector4 _sides;
-
-    /// <summary>Initializes a new instance of the <see cref="Thickness"/> class.</summary>
-    public Thickness (float left, float top, float right, float bottom)
-    {
-        _sides = new  (left, top, right, bottom);
-    }
-
     /// <summary>Initializes a new instance of the <see cref="Thickness"/> class with all widths set to 0.</summary>
-    public Thickness ()
-    {
-        _sides = new  (0, 0, 0, 0);
-    }
+    public Thickness () { _sides = Vector4.Zero; }
 
     /// <summary>Initializes a new instance of the <see cref="Thickness"/> class with a uniform width to each side.</summary>
     /// <param name="width"></param>
@@ -52,37 +47,7 @@ public record struct Thickness
         Bottom = bottom;
     }
 
-
-    /// <summary>Gets or sets the width of the left side of the rectangle.</summary>
-    [JsonInclude]
-    public int Left
-    {
-        get => (int)_sides.X;
-        set => _sides.X = value;
-    }
-    /// <summary>Gets or sets the width of the upper side of the rectangle.</summary>
-    [JsonInclude]
-    public int Top
-    {
-        get => (int)_sides.Y;
-        set => _sides.Y = value;
-    }
-
-    /// <summary>Gets or sets the width of the right side of the rectangle.</summary>
-    [JsonInclude]
-    public int Right
-    {
-        get => (int)_sides.Z;
-        set => _sides.Z = value;
-    }
-
-    /// <summary>Gets or sets the width of the lower side of the rectangle.</summary>
-    [JsonInclude]
-    public int Bottom
-    {
-        get => (int)_sides.W;
-        set => _sides.W = value;
-    }
+    private Vector4 _sides;
 
     /// <summary>
     ///     Adds the thickness widths of another <see cref="Thickness"/> to the current <see cref="Thickness"/>, returning a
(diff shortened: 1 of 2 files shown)


BDisp (approved):
LGTM

PR #3593, opened 2024-07-08
Diagnosing xunit failures in github action on ubuntu/macos

tig:


tig:
One successful run!

Is this fixed?

tig:
Twice! Yay.

PR #3597, opened 2024-07-09
Latest packages

tig:
Lastest nuget packages.

### DIFF
--- a/Analyzers/Terminal.Gui.Analyzers.Internal.Debugging/Terminal.Gui.Analyzers.Internal.Debugging.csproj
+++ b/Analyzers/Terminal.Gui.Analyzers.Internal.Debugging/Terminal.Gui.Analyzers.Internal.Debugging.csproj
@@ -6,12 +6,12 @@
     <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.9.2" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.9.2" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.9.2" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.9.2" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="4.10.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.10.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.10.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" PrivateAssets="all" />
     <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="8.0.0" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.9.2" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.10.0" PrivateAssets="all" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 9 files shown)


dodexahedron:
This would have probably been a smarter place for me to post the two I added in #3595 about version ranges to avoid exactly this, along with other related noise and busywork.

tig:
Not sure what you're talking about.

dodexahedron:
I'll grab the comments and cross-post here. Gimme a sec.

dodexahedron:
Here ya go.

The horizontal lines are just where the original comment boundaries were. No real significance.

---

@tig a suggestion to make this stuff less noisy:

If we were to specify version ranges for packages, they don't need to be bumped for every single patch release.

Actual builds still get locked to the version they were compiled with by default, by the packages.lock.json file nuget creates at restore time, so that builds are repeatable without unexpected version changes (one good reason to package those build artifacts on github, for debugging).

The format looks like math ranges:

Instead of an explicit version like"8.0.6" or whatever, a string such as `[8.0.0,9)`, for example, means follow the project roll-forward rules, allowing down to version 8.0.0, inclusive, up to 9 (dot anything), exclusive.

Inclusive and exclusive bounds are legal on both sides, and putting no number on one side of the comma means unbounded on that side. So, `[1.2.3,)` means from 1.2.3 inclusive to any higher version.

Standard Nuget rules apply as usual, like only 3 positions being relevant for version comparison, and a version ending in a hyphen plus any text is considered prerelease.

Cuts down on merge noise and noisy and sometimes conflicting version bump changes in PRs. Also helps avoid annoying compiler warnings or errors caused by mismatched dependencies it doesn't otherwise know how to resolve because it was told only specific versions are ok.

---

You can control them to any level you like, too.

(Edit: Correction now that I'm at my PC..)

9 is exactly equivalent to 9.0 and 9.0.0.

When you want to allow up to but not including the next major version, you use an exclusive upper bound like `[8,9)` which would mean anything from 8.0.0 inclusive to any other version with major 8 and any minor and patch.

Basically an `8.*.*` if that format were legal.

---

There's a website that has a couple of simple tools for analyzing dependency compatibility, most of which revolve around parsing version strings.

Here's an example of it in action

https://nugettools.azurewebsites.net/6.10.1/parse-version-range?versionRange=%5B7.0.20%2C9%29

Pretty nifty yet simple little tool if you're ever unsure of the result of a version range.

dodexahedron:
Those range values replace the explicit version values in the Version attribute of PackageReference elements in project files.

dodexahedron:
As an example from Terminal.Gui.csproj:

```xml
<PackageReference Include="ColorHelper" Version="1.8.1" />
```

could instead be

```xml
<PackageReference Include="ColorHelper" Version="[1.8.1,2)" />
```

That would make it unnecessary to explicitly make commits to bump versions for ColorHelper for all versions from 1.8.1, inclusive, to any other version 1.x.x version, only requiring a change when the major version increases.

Why a closed interval like that? Because major versions, by SemVer, imply breaking changes and thus should have human eyes on them.

dodexahedron:
Repeatability of builds is still maintained with that.

When nuget runs the restore target, it writes a `packages.lock.json` file in the obj directory, which makes any future builds that aren't specified with a forced reevaluation to be performed against the same versions. This already is happening without use of ranges and is part of why you sometimes have to do a clean or even a VS restart to get nuget to behave properly after updating a package. The only change is that, at the time of the first restore in a clean workspace, msbuild reconciles them across everything in the build following the roll-forward rules.

Makes the dependabot PRs almost completely go away and avoids those little merge conflicts in our PRs that are of course easy to resolve but still just needless trivia.

BDisp:
That means the dotnet restore will pull the latest version between 1.8.1 and 2?

dodexahedron:
Correct.

Anything from 1.8.1 up to 1.9999999.9999999 and NOT 2.

If you instead wrote `(1.8.1,2)`, it would be the same except NOT 1.8.1 either.

Exclusive on the left is less useful, but just showing what it means.

Check out that link. It breaks it down quite nicely.

dodexahedron:
Well... To be precise... It follows the roll-forward rules.

Out global.json specifies:

```json
    "rollForward":"latestMinor"
```

Which will have that behavior.

dodexahedron:
If anyone is curious about the available modes for that element in global.json, they are here:
https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#rollforward

The one we have is great IMO. The default is more conservative and chooses only the latest patch release in the specified major.minor. `latestMinor` changes that to the latest minor and patch in the specified major.

dodexahedron:
I went ahead and just did it and put in PR #3601 

If y'all like it, cool. If not, cool. 🤷‍♂️

dodexahedron:
Oh, another note:

If you want to allow prerelease versions for a package, just add something like `-prerelease` to the end of the version number. The specific text after the hyphen does not matter (so `-0` would have the same behavior, for example).

NuGet's default behavior is to ignore prerelease otherwise. Same with explicit versions.

PR #3602, opened 2024-07-10
Fixes #3592. DateField unit tests fail on macos.latest.

### LINKED ISSUE #3592: `DateField` unit tests fail on `macos.latest`
See https://github.com/gui-cs/Terminal.Gui/actions/runs/9831020519/job/27137868075#step:7:6759

@BDisp Can you look into this since you have more familiarity with the culture stuff than I?

BDisp:
## Fixes

- Fixes #3592

## Proposed Changes/Todos

- [ ] Add conditional for macos

## Pull Request checklist:


dodexahedron (changes requested):
This works, but a more formal/appropriate means of achieving platform-specific tests is to use per-platform test methods that have the [`[SupportedOSPlatform(string)]`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.supportedosplatformattribute?view=net-8.0) attribute applied.

For example: `[SupportedOSPlatform("macos")]`

There is also an inverse attribute available: [`[UnsupportedOSPlatform(string)]'](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.unsupportedosplatformattribute?view=net-8.0)

Both can appear multiple times if multiple platforms need to be specified.

dodexahedron:
That can be used in other code, too - not just tests, by the way.

BDisp:
Your suggestion is to only run a unit test that only supports or does not support a certain platform. In this case it is a unit test that runs on all platforms (Windows, Linux and Mac) and there are only constraints in 3 situations on the Mac. I also prefer to use the `OSPlatform` enum because it is more generic.

PR #2900, opened 2023-10-12
Fixes #2800 - Revamp `ColorPicker` and Color Scenarios in UICatalog

tig:
Fixes #2800 - Revamp ColorPicker and Color Scenarios in UICatalog

Depends on #2896 

- [ ] `ColorPicker` View let's users pick
   - [x] From the legacy 16-colors
   - [ ] From the full palette of 24-bit colors
     - [ ] Using RGB
     - [ ] Using HSL
- [ ] `ColorPicker` Scenario sufficiently demonstrates all of the above
- [ ] `True Colors` Scenario shows 16-colors correctly
   - [ ] When `Force16Colors` is true

## Pull Request checklist:

- [ ] I've named my PR in the form of "Fixes #issue. Terse description."
- [ ] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [ ] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [ ] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [ ] My changes generate no new warnings
- [ ] I have checked my code and corrected any poor grammar or misspellings
- [ ] I conducted basic QA to assure all features are working

Issue #3558, opened 2024-06-22
Build .... oddities...


dodexahedron:
OK. So...

#### **TL;DR: I think we should flatten the MSBuild config again, removing the Directory.*.* files and putting their contents in the projects they apply to OR explicitly include them where they're wanted.**

<details>
<summary>
<h4>Do you want me to just fix it and give a quick 5 mile overview reason for them or do you want me to write that book first?</h4>
</summary>

Before I write a book on this one, because it's a combination of multiple problems that are not actually our fault (bugs and limitations of various different components in the chain, mostly), are non-intuitive and have unknown impact that could be anything from wasted junk space to actual differences in program behavior that our tests don't cover (more likely closer to but not entirely the former)...

But yeah... How much detail do you want/need on this one?

Because, while it is a bitch of a set of issues to explain, the fix is extremely simple and understandable by everyone without explanation: Moving some build configuration around and flattening it. We're talking like 3 or 4 file deletions and moving their contents, verbatim, into the csproj files, and the problem is essentially licked. Alternatively, I can just rename them so they aren't automatically picked up by anything and we can put explicit `<Include .../>` elements in the csproj files we want to have them apply to.
</details>

#### **Either way, the end result is getting every tool and process to result in as similar output with equivalent options as possible, with no advanced knowledge of the build tools necessary.**

Here's a high-level and not-very-technical nor scientific intro (in the interest of the closest I can muster to "brevity") to what brought me to this conclusion and wanting to do this:

<details><summary>How I found out about and tracked it all down</summary>
In short? Some of it has been paper-cut misbehaviors of Visual Studio I and lots of people have just dealt with for many years. Some of it is known bad behavior and simply documented as such. And the rest became both very apparent and much more relevant and therefore irritating to me, with all the build tooling work I've been doing in various places over the past few weeks.

All I know, in terms of results, is that builds done in seemingly identical ways on the same system, from clean clones, with as little difference as the directory the build was ***INVOKED*** from can result in different build inputs (like the actual monolithic XML file msbuild creates and then executes) and different build artifacts. And that's not the only way to get different builds. And that's even with the `<Deterministic>true</Deterministic>` element in the csproj files, which doesn't really have much to do with much more than anonymization of source paths in debug symbols.

[...]

PR #3604, opened 2024-07-11
Fixes #2800 - Color picker (supporting hsl, hsv and rgb)

### LINKED ISSUE #2800: Revamp `ColorPicker` and Color Scenarios in UICatalog
image

* `Basic Colors`, `Invert Colors`, and `True Colors` should be combined and refactored to treat true color as default.

Related to #457

tznind:
Here is an updated ColorPicker WIP that supports HSL, HSV and RGB.

Requires a lot of tidy up and tests but this is the outline.  I will investigate integrating with #2900 which has lots of stuff on colors and fallback colors (e.g. `Ansi`).

I studied how it works in Paint.Net and saw the following:

- Hue bar never changes, it always shows full rainbow
- Saturation and Lightness are dependent on Hue and their counterpart.  The bar shows what the color would change to if you clicked in that cell.

For RGB the value rendered on each line indicates 'the color that will result if moving the cursor to the position on the bar'.  This is why the colors of G and B change as you adjust R.  

I think it is a really nice behaviour but happy to take suggestions and/or discuss approach.

I've learned a lot about color models!

image
_Paint.Net Color Picker_

image
_ColorPicker2_

## Fixes

- Fixes #2800

## Proposed Changes/Todos

- [x] When bar area is black triangle should turn grey so it is still visible
- [x] Tests
- [x] Scenario
- [x] Events

## Pull Request checklist:


tig:
This looks glorious!

dodexahedron:
I like.

One thing I had on my personal to-do list was to actually just cherry pick the code from that ColorHelper library that is actually used, since it's also MIT license and avoiding a whole library dependency for just a few methods seems like a nice thing to do for users. Also saves a non-trivial chunk of bytes for the package as well as work for the trimmer, which was actually what prompted me to consider adding that work item.

If you wanna go for it, feel free. Otherwise, I'll just keep it as a low-priority item to address during beta/pre-release of TG. All good either way. 

Could do it as a git submodule, but I've never loved that workflow and it seems to steepen the learning curve for getting going with a project, making it harder for people who aren't black belts in git-kwon-do to grok.

So, my intent is to preserve the original namespaces and formatting of the copied code, and everything and tell the analyzers to buzz off for that code, so it doesn't add any warnings to our build and doesn't get reformatted. The default build configurations would be set up to build it that way. As part of the configuration I'm adding in #3558, I was planning to make a sort of "fully loaded" design time configuration, which would basically exclude that code and pull in the actual package of the same tagged version as the copied code, instead, so anyone actually working on enhancing features of TG would have access to the full library dependency and everything it offers, to make it just a matter of then copying over whatever new stuff gets used, before merge time, for the default configurations.

I actually vaguely remember seeing a Roslyn package out there that automates that last part at build time, even. I'll look into that whenever I get to it.

ALTERNATIVELY, we could just configure the build to trim that dependency ***specifically*** and not include the actual transitive build dependency in the package. At least I'm pretty sure there's a combination of incantations in msbuild to make that happen. If not, oh well. The other ways do it, just with slightly more XML. 🤷‍♂️

tznind:
Can you shed any light on `AllViews_Enter_Leave_Events`

Seems its expecting 2 Enter events but only getting 1.  But from what I can see the test is adding View to page and tabbing into it...

```
Terminal.Gui.ViewsTests.AllViewsTests.AllViews_Enter_Leave_Events(viewType: typeof(Terminal.Gui.ColorPicker2)) [FAIL]
[xUnit.net 00:00:04.71]       Assert.Equal() Failure: Values differ
[xUnit.net 00:00:04.71]       Expected: 2
[xUnit.net 00:00:04.71]       Actual:   1
[xUnit.net 00:00:04.71]       Stack Trace:
[xUnit.net 00:00:04.71]         /Users/runner/work/Terminal.Gui/Terminal.Gui/UnitTests/Views/AllViewsTests.cs(127,0): at Terminal.Gui.ViewsTests.AllViewsTests.AllViews_Enter_Leave_Events(Type viewType)
```

tznind:
Ok this should be good to review @tig, need to understand why AllViews_Enter_Leave_Events is failing 

Then I can either scrap the old ColorPicker or we can maintain both (in which case probably rename ColorPicker2 to TrueColorPicker or something).

UnitTests/Views/ColorPicker2Tests.cs
```diff
@@ -216,6 +216,75 @@ public void ColorPicker_RGB_NoText (Color c, string expected)
...
+        top.Dispose ();
+    }
+
+    public static IEnumerable<object []> ColorPickerTestData_WithTextFields()
+    {
+        yield return new object []
+        {
+            new Color(255, 0, 0),
+            @"
+R:█████████████▲255
```
tig, line 230: I've really grown to despise unit tests that test for visuals. They are extremely hard to maintain over the long run.

Can these tests be replaced with something that either doesn't test contents or is easier to change if elements are updated?

tig (changes requested):

UnitTests/Views/ColorPicker2Tests.cs
tznind: We can make the triangle location from a local variable to a public `get` for the test

And we can replace the diagram code with some test helpers e.g. https://github.com/gui-cs/Terminal.Gui/pull/3604/commits/d4697bdca002e4102ca2d6c70172ee8446a1e9aa

I don't think it will make it more maintainable though mostly just make it harder to figure out what is going on when it does break.

Terminal.Gui/Views/ColorPicker2.cs
```diff
@@ -3,13 +3,29 @@
 
 namespace Terminal.Gui;
 
+/// <summary>
```
tig, line 6: All types should be in their own .cs files.

Terminal.Gui/Views/ColorPicker2.cs
```diff
@@ -210,7 +230,7 @@ public class ColorPicker2 : View
     /// </summary>
     public event EventHandler<ColorEventArgs> ColorChanged;
 
-    private bool updating;
+    private bool _updating;
```
tig, line 233: These kinds of variables always feel hacky to me. Is this stricktly needed, or is there another way of ensuring updates are not problematic?

tig (changes requested):
Let's go ahead and just replace `ColorPicker.cs` with this?

Terminal.Gui/Views/ColorPicker2.cs
tznind: There are 7 views who can each independently trigger a change in color plus the root public property.

- when a bar position changes x3
- when a bar text field changes (currently tied to leave not text changed) x3
- when hex text field changes (again with leave event)

In each case we must update the other 6 (or all 7 if public property is source of change I.e. user calls API direct)

I will have a look, maybe there is a way to handle it with a `SetValueWithoutRaisingEvent` method instead.

UnitTests/Views/ColorPicker2Tests.cs
tznind: Tests now take the format:

```csharp
    [Fact]
    [AutoInitShutdown]
    public void ColorPicker_Construct_DefaultValue ()
    {
        var cp = new ColorPicker2 { Width = 20, Height = 4, Value = new (0, 0) };

        cp.Style.ShowTextFields = false;
        cp.ApplyStyleChanges ();

        var top = new Toplevel ();
        top.Add (cp);
        Application.Begin (top);

        // Should be only a single text field (Hex) because ShowTextFields is false
        Assert.Single (cp.Subviews.OfType<TextField> ());

        cp.Draw ();

        // All bars should be at 0 with the triangle at 0 (+2 because of "H:", "S:" etc)
        var h = GetColorBar (cp, ColorPickerPart.Bar1);
        Assert.Equal ("H:", h.Text);
        Assert.Equal (2, h.TrianglePosition);
        Assert.IsType<HueBar> (h);

        var s = GetColorBar (cp, ColorPickerPart.Bar2);
        Assert.Equal ("S:", s.Text);
        Assert.Equal (2, s.TrianglePosition);
        Assert.IsType<SaturationBar> (s);

        var v = GetColorBar (cp, ColorPickerPart.Bar3);
        Assert.Equal ("V:", v.Text);
        Assert.Equal (2, v.TrianglePosition);
        Assert.IsType<ValueBar> (v);

        var hex = GetTextField (cp, ColorPickerPart.Hex);
        Assert.Equal ("#000000", hex.Text);

        top.Dispose ();
    }
```

BDisp: Which one refers to color?

Terminal.Gui/Views/ColorPicker2.cs
tznind: Implemented in [d6eb6c3](https://github.com/gui-cs/Terminal.Gui/pull/3604/commits/d6eb6c387f25d72f5dbb94992a38d22e2cb0ef5a)

Terminal.Gui/Views/ColorPicker2.cs
tznind: Done, each class now in its own file

UnitTests/Views/ColorPicker2Tests.cs
tznind: > Which one refers to color?

Not sure I understand the question.  Can you elaborate?

BDisp: I think I confused GetColorBar with GetColor. I'm sorry.

UICatalog/Scenarios/Shortcuts.cs
```diff
@@ -320,9 +320,7 @@ private void App_Loaded (object sender, EventArgs e)
 
         var bgColor = new ColorPicker ()
         {
-            CanFocus = false,
-            BoxHeight = 1,
-            BoxWidth = 1,
+            CanFocus = false, Width = 17
```
tznind, line 323: `CanFocus` of false is set on all controls so I have left in... don't really understand this Scenario though.

Terminal.Gui/Views/ColorBar.cs
```diff
@@ -47,6 +47,13 @@ public int Value
         }
     }
 
+    /// <inheritdoc />
+    public void SetValueWithoutRaisingEvent (int v)
```
dodexahedron, line 51: Not a fan of this.

If setting something without raising events is necessary, unsubscribe.

What if the user _wants_ those events to be raised for every individual value for whatever reason?

Otherwise, this should be private.

dodexahedron (changes requested):
Looks mostly good to me, except for the one comment around that public SetValueWithoutRaisingEvent method.

Terminal.Gui/Views/ColorBar.cs
tznind: I added this only because @tig didn't like `bool updating` approach in parent view - see https://github.com/gui-cs/Terminal.Gui/pull/3604#discussion_r1677164640

We need a way to sync all the sub bits of ColorPicker without getting circular events being triggered.

A compromise is perhaps that the method can be made internal?

BDisp: Another suggestion is add a bool property like RaiseEventForAllValues.

tznind: The point of the change was to remove variables and before/after operations.

We do want the event raised for clicks and keyboard changes just not when we are sync a change e.g. when user tupes new value in hex box

BDisp: Sorry, I interacted on impulse due to @dodexahedron saying that the user might want to be notified of all changes. Really, if they are unnecessary, avoiding them from generating events is the best option.

tznind: I have hidden the method and the class as both are internal sub components of ColorPicker.

See  [21897b5](https://github.com/gui-cs/Terminal.Gui/pull/3604/commits/21897b5cba3bb3df168f63a1c9511abe5dc08463)

Is this acceptable compromise @dodexahedron?

dodexahedron: > I added this only because @tig didn't like `bool updating` approach in parent view - see [#3604 (comment)](https://github.com/gui-cs/Terminal.Gui/pull/3604#discussion_r1677164640)
> 
> We need a way to sync all the sub bits of ColorPicker without getting circular events being triggered.
> 
> A compromise is perhaps that the method can be made internal?

Yeah even internal events are a possibility. I haven't given it additional dedicated thought, yet, so at the moment it's just brainstorming essentially.

dodexahedron: > The point of the change was to remove variables and before/after operations.
> 
> We do want the event raised for clicks and keyboard changes just not when we are sync a change e.g. when user tupes new value in hex box

Well... Why not?

Other color pickers I (unreliably) recall usually update in real time.

Paint.net definitely does. Observe:

https://github.com/user-attachments/assets/565cf7bc-606a-4f6a-ac27-cb3f458167c7

However, making that behavior work exactly like that, for all possible 1-8 digit hex strings, would probably require some expansion of the color parse switch block to handle the 1, 2, 4, 5, and 7 character cases, I think. ...Though it's been many moons since I wrote that, so I don't remember all of the cases it handles off the top of my head. I'm pretty sure those are the cases not covered.

Alternatively/more simply, we could have the parse code only raise the event for strings that are meaningful, like 1 (interpret as that nibble repeated 6 times), 2 (that byte repeated 3 times), 3 (it already does that - each nibble is doubled, like on the web), 4 (also does this I'm pretty sure - same as 3 but with alpha), 6 (of course already does this), and 8 (again, already does this). So that only really means adding cases for 1 and 2, I guess? I suppose 5 and 7 _could_ be handled, assuming there's just an omitted leading zero for both, making them just use the 6 and 8 character cases... 🤷‍♂️ 

And of course that struct is natively ARGB, but that's only relevant for naming really because it's just a 32-bit binary value. So, following some of my original thoughts farther down the rabbit hole, I think the way I might end up dealing with that is pretty much how .net 8 and the System.Numerics pseudo-recursive interfaces with static virtuals work: A couple of common interfaces with things like the named colors, and then ColorBlahBlah structs for each pixel format/color space, so they're largely interchangeable, the way things like `INumber<TSelf>` are, regardless of the underlying struct[^IfIKnewThenWhatIKnowNow].

That's where the things like int.Zero come from - they're defined on the static virtual interface member, and then each implementing type brings its own implementation of that member, but it is static, so it is available on the _type argument_ of the interface (so, the color types themselves), meaning it's a largely non-breaking change, insofar as Color itself and its visible API surface goes. 

Anyway... There's what occurred to me while thinking about it right now. Plenty of options available other than those, of course, but I think those would be very low-code and highly idiomatic C#12+ ways to go about it, at least, with details up to the implementation and ultimately irrelevant to the consumer (yay for interfaces!).

[^IfIKnewThenWhatIKnowNow]: Static virtuals on self-referencing generic interfaces are honestly something we should do to Dim/Pos to polish them up and make them seamless, as originally intended when they were nested types hidden from the consumer. They're exactly the solution to it all. When that stuff was happening, I hadn't yet had a chance to get terribly intimate with that language feature yet. I've come to love it because it really does deal with that exact scenario so well. Sure do wish I were familiar with it months ago, when we were trying to figure that all out. It is a bit tricky to initially get the hang of writing because it's VERY particular about the syntax, but it's all strongly-typed static analysis, so you pretty much can't write it wrong and manage to compile it. Score another point for Roslyn.

dodexahedron: You know... Come to think of it...

Paint.Net is licensed with an MIT equivalent, just with an explicit trademark exception (so they should have just used MPL-2.0, but oh well).

So, we could even go so far as to lift the code for that entire control and translate to the Terminal.Gui API, and all we need is a license notice, just like any other dependency already requires (which we need to be sure we shore up in the nupkg before release, btw).

tznind: It already behaves as paint.net one. I.e. real-time sync.

The code here is just to avoid infinite sync looping

dodexahedron: A fix should be pretty quick I think but I'm on my way out the door so I'll look later. It's not a huge deal as a way to, at the end of the day, achieve the same result. But I would request that method not be public, please, so nobody uses it before it gets polished up. Or drop an obsolete attribute on it. Or whatev. 🤷‍♂️ Or just @mention me in a todo in the code if you don't wanna mess with it and I'll include it in one of my ADHD ~~nitpick~~ cleanup passes. 😅

tznind: > But I would request that method not be public

I already made it not public in https://github.com/gui-cs/Terminal.Gui/pull/3604/commits/21897b5cba3bb3df168f63a1c9511abe5dc08463

UICatalog/Scenarios/Shortcuts.cs
tig: `Shortcut`'s `CommandView` can be any view. In most cases, it does not make sense for that view to be focusable (e.g. when `Shortcut` is used w/in `StatusBar`. 

I was using this to test all that.

tig: It's important that Views (e.g. ColorPicker) support mouse events even if `!CanFocus || !HasFocus`.

tig (changes requested):
Really great stuff Thomas!

As noted, having ColorPicker default to `Height/Width = Dim.Auto` would be really helpful. I think you'll find it's not a lot of work (but does require some thought). 

Please go ahead and put `ColorPicker.cs` etc... in `./Views/ColorPicker` ... anytime there's more than ~2 files for a view we should do this.  As noted, a few of the types are generic enough to go in `./Drawing/`.

tznind:
@tig I have moved the classes you mentioned to Drawing.

I set Width/Height to Dim.Auto and removed explicit setting in the `LineDrawing` Scenario.

However this results in a very thin UI:

image


Vertical height looks good, each of the children makes its parent 1 higher so we get height of 4.  But the horizontal layout is not right.  I'd like to understand if this is as designed:

Label is 3 characters width (abs) + 4 characters width (abs) but ColorBar is Dim.Fil().

I assumed that the auto size would effectively inherit the Dim.Fill() of it's content?  But it is instead taking a Width of 12

image

How should I approach this problem? Should ColorPicker just specify Dim.Fill() for its width explicitly?

tig:
> @tig I have moved the classes you mentioned to Drawing.
> 
> I set Width/Height to Dim.Auto and removed explicit setting in the `LineDrawing` Scenario.
> 
> However this results in a very thin UI:
> 
> image
> 
> Vertical height looks good, each of the children makes its parent 1 higher so we get height of 4. But the horizontal layout is not right. I'd like to understand if this is as designed:
> 
> Label is 3 characters width (abs) + 4 characters width (abs) but ColorBar is `Dim.Fill(4)`.
> 
> I assumed that the auto size would effectively inherit the Dim.Fill() of it's content? But it is instead taking a Width of 12 (I guess from its bottom row). Does Fill just get ignored by auto?
> 
> image
> 
> How should I approach this problem? Should ColorPicker just specify Dim.Fill() for its width explicitly?

As noted in [dimauto.md](https://gui-cs.github.io/Terminal.GuiV2Docs/docs/dimauto.html) `Dim.Fill` has no impact on `Dim.Auto`. 

There's no way (without crazy non-deterministic gymnastics) for Dim.FIll to know how wide to fill a view that is automatically sized based on the size of the `Dim.Fill`. 

My recommendation would be for you to follow what `Slider` does in setting `ContentSize` (see `SetContentSize()` in `Slider.cs`). You can come up with "MinimumNaturalSize" for the color picker.

Another strategy would be to use `Dim.Auto`'s `minimumContentSize:` param which should be able to achieve a similar result.

tznind:
Ok should be good to go, test failure looks unrelated?

tig:
@tznind PR incoming with a fix to build issues.

Please make all new/significantly changed files `#nullable enable`

UnitTests/Views/ColorPickerTests.cs
```diff
@@ -1,79 +1,652 @@
...
+﻿using Xunit.Abstractions;
+using Color = Terminal.Gui.Color;
+
+namespace UnitTests.Views;
 
 public class ColorPickerTests
 {
     [Fact]
-    public void Constructors ()
+    [AutoInitShutdown]
```
tig, line 9: I've worked really hard to make it so it's VERY rare where `AutoInitShutdown` is needed for a test. 

Please don't use it unless absolutely needed.  When used, if a test fails, it makes it very hard to find the failure because app state is not reset.

Here's how to fix this test as an example:

```cs
    [Fact]
    [SetupFakeDriver]
    public void ColorPicker_Construct_DefaultValue ()
    {
        var cp = new ColorPicker { Width = 20, SelectedColor = new (0, 0) };

        cp.Style.ShowTextFields = false;
        cp.ApplyStyleChanges ();

        // Should be only a single text field (Hex) because ShowTextFields is false
        Assert.Single (cp.Subviews.OfType<TextField> ());

        cp.Draw ();

        // All bars should be at 0 with the triangle at 0 (+2 because of "H:", "S:" etc)
        var h = GetColorBar (cp, ColorPickerPart.Bar1);
        Assert.Equal ("H:", h.Text);
        Assert.Equal (2, h.TrianglePosition);
        Assert.IsType<HueBar> (h);

        var s = GetColorBar (cp, ColorPickerPart.Bar2);
        Assert.Equal ("S:", s.Text);
        Assert.Equal (2, s.TrianglePosition);
        Assert.IsType<SaturationBar> (s);

        var v = GetColorBar (cp, ColorPickerPart.Bar3);
        Assert.Equal ("V:", v.Text);
        Assert.Equal (2, v.TrianglePosition);
        Assert.IsType<ValueBar> (v);

        var hex = GetTextField (cp, ColorPickerPart.Hex);
        Assert.Equal ("#000000", hex.Text);
    }
```

tig (changes requested):
See my PR and my comment about `#nullable enable`. Thanks.

tznind:
Am away for couple days but will update nullable when I get back.  Thanks.

tig (changes requested):
I was playing with this and exited the ColorPicker Scenario and got this. You have a Dispose bug somewhere.

image

image


Also, I didn't notice this before, but while this is lovely, it is very hard to pick a standard color. The old ColorPicker was lame in that it only handled 16 colors, but it DID make it trival for a user to pick a standard color.

I'm not sure if we should do it in this PR, or later, but we really need something that just shows the standard 16 colors and lets the user pick one. I think there should also be one that lets users pick from the standard Html named colors from html, perhaps as a drop-down list?

https://www.w3schools.com/colors/colors_names.asp

I think the right way to do this is:

- Bring back the old `ColorPicker` as `ColorPicker16`
- Update the scenario to have a class chooser (using RadioGroup or Slider) to pick which class to use.
- Later, add `ColorPickerNames`

Terminal.Gui/Drawing/ColorEventArgs.cs
```diff
@@ -0,0 +1,16 @@
+﻿#nullable enable
+
+namespace Terminal.Gui;
+
+/// <summary>Event arguments for the <see cref="Color"/> events.</summary>
+public class ColorEventArgs : EventArgs
```
tig, line 6: For consistency, it would be nice if this followed the pattern I introduced in IOrientation.

For non-canceable use `EventArgs<Color>`.  For cancelable use `CancelEventArgs<Color>`

tig (changes requested):

Terminal.Gui/Drawing/ColorEventArgs.cs
tznind: Done in ea614953dd89b69db01963cd91bc222b325791db

tznind:
> You have a Dispose bug somewhere.

Can confirm this occurs when switching color models.  Seems was missing Dispose call in `DisposeOldViews`.

Fixed and test added in [e371e99](https://github.com/gui-cs/Terminal.Gui/pull/3604/commits/e371e998722c9e62705668491cf2294b22a2c9d1)

tznind:
> Bring back the old ColorPicker as ColorPicker16

Given we have `ColorModel` in style options we might be able to do it that way instead

```csharp
cp.Style.ColorModel = ColorModel.SixteenColors;
cp.Style.ShowName = true;
cp.ApplyStyleChanges ();
```

This would show something like:
image
_Mock up in which the 16 console colors would show on bar and it would be max length 16_

We can have `ShowName` also work in the context of true color 

image

The values allowed in Name would be 
- If 16 color mode the named console colors
- Any other ColorModel it is the w3c names

If a color is selected in true color mode that is not a named color
- Text field shows `-`

We can improve usability by having the Names text field use `AppendAutocomplete`

What do you think?

tznind:
I have started on the named color support so have converted to draft, here is the ColorPickers scenario with names on:

image

I have to do a lot of work for consistency and tests.

tznind:
Ok I will finish up the name stuff for w3c as I think that still has value as an optional style.

Then see if I can restore and rename the old color picker.

BDisp:
Why not using both in the same view. The HSV could be hide to only show the 16 colors, if true color isn't supported. If a user only want a 16 colors he click on the desire color where the name and value can be showed. If a user want a true color the HSV is sowed above or below the 16 colors. It could be fine that way.

tznind:
The style options don't really make sense if doing 16 colors in a dynamic grid.  This is why I proposed 16 being a bar with the 16 colors as 1x1 rectangles.

But I'd we want to allow 2x2 etc and 2 row layout of colors then it's too different to make sense as the same view.

I think if we like the old view ux it's best to restore it seperate

Terminal.Gui/Drawing/W3CColors.cs
```diff
@@ -0,0 +1,171 @@
+﻿namespace Terminal.Gui;
+
+/// <summary>
+/// Helper class that resolves w3c color names to their hex values
+/// Based on https://www.w3schools.com/colors/color_tryit.asp
+/// </summary>
+public class W3CColors : IColorNameResolver
+{
+    Dictionary<string,Color> _knownColors=new Dictionary<string,Color>(StringComparer.InvariantCultureIgnoreCase)
```
tig, line 9: Shouldn't these be in a string resource. Not for loc reasons, but to minimize memory use when this class is instantiated? 

I think the `new Color("#xxxx")` ends up as code, which is less concerning, but it would be more efficient if these were specified as ints and not strings?

tig (changes requested):

tig:
This this:

```cs
        string white = ColorStrings.GetW3CColorName (0xFFFFFF);
        if (ColorStrings.TryParseW3CColorName (white, out Color color))
        {
               Debug.Assert(0xFFFFF, color);
        }
```

```cs

/// <summary>
///     Provides a mapping between <see cref="Color"/> and the W3C standard color name strings.
/// </summary>
public static class ColorStrings
{
    private static readonly ResourceManager _resourceManager = new ResourceManager (typeof (Strings));

    /// <summary>
    ///     Gets the W3C standard string for <paramref name="color"/>.
    /// </summary>
    /// <param name="color">The color.</param>
    /// <returns><see langword="null"/> if there is no standard color name for the specified color.</returns>
    public static string? GetW3CColorName (Color color)
    {
        // Fetch the color name from the resource file
        return _resourceManager.GetString ($"#{color.R:X2}{color.G:X2}{color.B:X2}", CultureInfo.CurrentCulture);
    }

    /// <summary>
    ///     Parses <paramref name="name"/> and returns <paramref name="color"/> if name is a W3C standard named color.
    /// </summary>
    /// <param name="name">The name to parse.</param>
    /// <param name="color">If successful, the color.</param>
    /// <returns><see langword="true"/> if <paramref name="name"/> was parsed successfully.</returns>
    public static bool TryParseW3CColorName (string name, out Color color)
    {
        // Iterate through all resource entries to find the matching color name
        foreach (DictionaryEntry entry in _resourceManager.GetResourceSet (CultureInfo.CurrentCulture, true, true)!)
        {
            if (entry.Value is string colorName && colorName.Equals (name, StringComparison.OrdinalIgnoreCase))
            {
                // Parse the key to extract the color components
                string key = entry.Key.ToString () ?? string.Empty;
                if (key.StartsWith ("#") && key.Length == 7)
                {
                    if (int.TryParse (key.Substring (1, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int r) &&
                        int.TryParse (key.Substring (3, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int g) &&
                        int.TryParse (key.Substring (5, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int b))
                    {
                        color = new Color (r, g, b);
                        return true;
                    }
                }
            }
        }

        color = default;
        return false;
    }
}
```

Then, in `./Resources/Strings.resx":

image

tig:
Submitted a PR with a potential impl.

tznind:
@tig the side effect of moving color names to a shared resources file is that all other strings are considered colors e.g. 'Save as'

I have added a test to show this.

How do you want to handle that? should there be a seperate resource file for the colors or is it possible to label them somehow so to tell the difference?

image

tig:
The bug is in here:
````cs
public static IEnumerable<string> GetW3CColorNames ()
    {
        foreach (DictionaryEntry entry in _resourceManager.GetResourceSet (CultureInfo.CurrentCulture, true, true)!)
        {
            if (entry.Value is string colorName)
            {
                yield return colorName;
            }
        }
    }
```

Should be

```cs
public static IEnumerable<string> GetW3CColorNames ()
    {
        foreach (DictionaryEntry entry in _resourceManager.GetResourceSet (CultureInfo.CurrentCulture, true, true)!)
        {
            if (entry.Value is string colorName && colorName.StartsWith(`#`))
            {
                yield return colorName;
            }
        }
    }
```

(or similar)

tznind:
Ok, we now have old ColorPicker back in again as ColorPicker16 and ColorPickers scenario lets you switch between the two.

I also fixed a bug in the rendering of ColorPicker16 and finished off the W3C color names which should make it easier to select a known color.

tig:
I submitted another PR with some tweaks.

Can you also please:

- Revert AdornmentEditor etc... to use ColorPicker16
- Fix all warnings

Thanks.

tig:
Is it possible to update the color name in the picker automatically when the sliders are set to a color that matches?

dodexahedron:
It may be worth noting that named colors are environment-dependent, and may not be what their names say, depending on terminal, platform, termcaps, and environment variables like `$LS_COLORS`.

None of which are our problem/fault, but perhaps worth noting at least in docs and in XmlDoc comments around this functionality.

tznind:
> It may be worth noting that named colors are environment-dependent, and may not be what their names say

For this color picker (true color) the colors are the w3c color names not the console colors

I added an interface for the color namer so in future if someone wants non w3c colors/names they can.

dodexahedron:
Gotcha. Sounds great. 👍

tznind:
> Is it possible to update the color name in the picker automatically when the sliders are set to a color that matches?

Yes this is already how it works, however notice that each movement of the triangle moves multiple units in the color space.  For example with RGB if bar width is 100 console characters then each cell is 25.5 units of color space.

keyboard-change-color
_You cannot use mouse to 're-select' the 127 value because it maps between cells_

tznind:
@tig I have restored the old ColorPicker16 for all Scenarios except the LineDrawing and ColorPickers scenarios (although the pickers one does both).

I think I got all the warnings just spelling of 'viewport' and a warning about async tests as well as naming rule violation which is on `ResourceManager _resourceManager` but when fixing IDE1006 there resharper wants to rename with the underscore so maybe some conflicting rules there.

Let me know if I missed any warnings or other things to do to this PR.

tig (approved):
Lovely work Thomas. Thank youl

PR #3612, opened 2024-07-13
Fixes #3611. Localization not working on self-contained single-file.

### LINKED ISSUE #3611: Localization not working on self-contained single-file.
The reason is the `resx` files are embedded on the executable file and thus there is no satellite assemblies for each culture in the folder.
The solution is to return a list of cultures that we know are supported. However, if the project contains the `<InvariantGlobalization>true</InvariantGlobalization>` setting only the `invariant culture` is supported. For other cultures to be supported, it is necessary not to use this configuration or use it as `<InvariantGlobalization>false</InvariantGlobalization>`.

BDisp:
## Fixes

- Fixes #3611

## Proposed Changes/Todos

- [x] Generate supported cultures
- [x] Added assert culture count in the unit test. This must be manually
- [x] Added assert in the SelfContained project and call first Application.Init before Application.Run
- [x] Change this setting to false in the SelfContained project `<InvariantGlobalization>false</InvariantGlobalization>`
- [x] Anyone with a different culture than the invariant, will have the current culture used, right click on the TextField to test it

## Pull Request checklist:


SelfContained/Program.cs
```diff
@@ -10,7 +11,23 @@ public static class Program
     [RequiresUnreferencedCode ("Calls Terminal.Gui.Application.Run<T>(Func<Exception, Boolean>, ConsoleDriver)")]
     private static void Main (string [] args)
     {
-        Application.Run<ExampleWindow> ().Dispose ();
+        Application.Init ();
+
+        if (Equals (Thread.CurrentThread.CurrentUICulture, CultureInfo.InvariantCulture))
```
tig, line 16: Does this need to be in the sample? Does EVERY self-contained-single app need these checks? If not, please remove.

tig (changes requested):

SelfContained/Program.cs
tig: The sample code should only include code that is required to show developers how to do something.

Your response confuses me because you say "it's only needed to ensure it not breaks" and "this is isn't needed because it's only for single-file". But this IS a single-file sample.

BDisp: I was referring the localization not breaking anymore.

tig: I still don't understand. 

Can you please update `SelfContained/Program.cs` to have the bare-minimum code required to be a clear and complete sample for a single file self-contained TUI app?

BDisp: @tig is this solution https://github.com/gui-cs/Terminal.Gui/pull/3612/commits/ba4ab0c8c6aab183a704ab2016f6c765a27b9d5c enough?
Also the respective `SelfContained.csproj` is not compatible with the creation of a single-file self-contained application because in `Debug` mode it references the `Terminal.Gui` project and in `Release` mode it references the `Terminal.Gui` nuget package. My main reason for creating this project was to test and verify that `Terminal.Gui` complies with trimmed applications. Normally a user who wants a single-file application knows how to do it.

tig:
This ready for merge @BDisp ?

BDisp:
> This ready for merge @BDisp ?

Yes.

tig (approved):

PR #3613, opened 2024-07-13
Fixes #3350. Scenario is complex, confusing, and causes errors.

### LINKED ISSUE #3350: `Scenario` is complex, confusing, and causes errors
I over-engineered `Scenario` and tried to be both too helpful to devs and too cute. My intent was to make it super easy to create new scenarios. I thought it would be neat if Scenario developers only had to write a `Setup` override and the "default" logic in `Init` and `Run` would do magic for them. I also had UICatalog call `Application.Shutdown` when the Scenario exited instead of forcing Scenario authors to do this themselves.

This was a bad idea for these reasons:

1. New developers looking at scenarios as examples don't see what ones supposed to do to setup/teardown a Terminal.Gui app. They don't see the proper way: Call `Init`, and ensure it's bracketed by a `Shutdown`, Call `Run` and make sure the `Toplevel` used gets disposed. 
2. It led to un-needed complexity in 4 overrides: `Init`, `Setup`, `Run`, and `RequestStop`. This is confusing to devs and a survey of the existing scenarios shows this confusion.
3. It makes devs think that `Application.Top` can't be the actuall main view for the scenario, because `Init`, by default creates `Win` and adds it to `Top`. 

My proposal:

- Simplify `Scenario` to have only ONE virtual metods: `Main`. Developers should treat this just like the standard C# console app `int Main (string [] args)` method.
- Change `Init` to NOT reset `ConfigurationManager`. This way whatever was configured in UI Catalog will naturally just pass onto the Scenario.
- Change all Scenarios to merge their `Init`, `Setup`, and `Run` overrides into `Main`. 

The Generic `Scenario.Main` will look like this:

```cs
    public override void Main ()
    {
        // Init
        Application.Init ();

        Window appWindow = new ()
        {
            Title = $"{Application.QuitKey} to Quit - Scenario: {GetName ()}",
        };

        var button = new Button { X = Pos.Center (), Y = Pos.Center (), Text = "Press me!" };
        button.Accept += (s, e) => MessageBox.ErrorQuery ("Error", "You pressed the button!", "Ok");
        appWindow.Add (button);

        Application.Run (appWindow);
        appWindow.Dispose ();

        Application.Shutdown ();
    }
```

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3339#issuecomment-2018897248_

BDisp:
## Fixes

- Fixes #3350

## Proposed Changes/Todos

- [x] Change all Scenarios to merge their Init, Setup, and Run overrides into Main.
- [x] Remove obsolete Init, Setup and Run virtual methods.
- [x] Remove obsolete Top and Win properties
- [x] Add GetQuitKeyAndName helper method.

## Pull Request checklist:

### DIFF
--- a/UICatalog/Scenario.cs
+++ b/UICatalog/Scenario.cs
@@ -138,24 +138,9 @@ public class Scenario : IDisposable
     {
     }
 
-    /// <summary>
-    ///     The Toplevel for the <see cref="Scenario"/>. This should be set to <see cref="Terminal.Gui.Application.Top"/>.
-    /// </summary>
-
-    //[ObsoleteAttribute ("This property is obsolete and will be removed in v2. Use Main instead.", false)]
-    public Toplevel Top { get; set; }
-
     /// <summary>Gets the Scenario Name + Description with the Description padded based on the longest known Scenario name.</summary>
     /// <returns></returns>
     public override string ToString () { return $"{GetName ().PadRight (_maxScenarioNameLen)}{GetDescription ()}"; }
-
-    /// <summary>
-    ///     The Window for the <see cref="Scenario"/>. This should be set to <see cref="Terminal.Gui.Application.Top"/> in
-    ///     most cases.
-    /// </summary>
-
-    //[ObsoleteAttribute ("This property is obsolete and will be removed in v2. Use Main instead.", false)]
-    public Window Win { get; set; }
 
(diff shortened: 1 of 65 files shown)


tig (approved):
You rock. Thank you!!!!

PR #3625, opened 2024-07-22
Adds `IOrientation` and `OrientationHelper` - opinionated changing/changed event pattern

tig:
## Fixes

image

I'm also trying to figure out the perfect pattern for Change/Changing events.

## Proposed Changes/Todos

- [x] Implement `IOrientation` and helper; use `RadioGroup` as test subject
- [x] Add unit tests
- [x] Convert `Slider` to use `IOrientation`
- [x] Convert others

## Pull Request checklist:


Terminal.Gui/View/Orientation/IOrientation.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    /// <param name="currentOrientation">The current orientation.</param>
+    /// <param name="newOrientation">The new orientation.</param>
+    /// <returns><see langword="true"/> to cancel the change.</returns>
+    public bool OnOrientationChanging (Orientation currentOrientation, Orientation newOrientation) { return false; }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public event EventHandler<CancelEventArgs<Orientation>> OrientationChanged;
```
BDisp, line 34: I think this should be only the following because the old value isn't needed if wasn't cancelled.

`public event EventHandler<EventArgs<Orientation>> OrientationChanged;`

Don't forget to add doc to this.

Terminal.Gui/View/Orientation/IOrientation.cs
```diff
@@ -0,0 +1,43 @@
...
+
+    /// <summary>
+    ///     Called when <see cref="Orientation"/> has been changed.
+    /// </summary>
+    /// <param name="oldOrientation"></param>
+    /// <param name="newOrientation"></param>
+    /// <returns></returns>
+    public void OnOrientationChanged (Orientation oldOrientation, Orientation newOrientation) { return; }
```
BDisp, line 42: The `oldOrientation` is not needed. If the user want it then he can capture it during the `OnOrientationChanging` event.

BDisp (approved):
LGTM. Only an observation.

dodexahedron:
RE: Changed/Changing:

Best to stick to established convention on that one I think.

Enabling minimal divergence between GUI code like WPF, WinUI, etc and Terminal.Gui is a ***major*** value add.

Anything we do in the public API surface that bucks established convention for common functionality is a pain point for consumers, even if they aren't trying to port something, because established and ubiquitous convention is still established and ubiquitous convention.

tig:
> RE: Changed/Changing:
> 
> Best to stick to established convention on that one I think.
> 
> Enabling minimal divergence between GUI code like WPF, WinUI, etc and Terminal.Gui is a _**major**_ value add.
> 
> Anything we do in the public API surface that bucks established convention for common functionality is a pain point for consumers, even if they aren't trying to port something, because established and ubiquitous convention is still established and ubiquitous convention.
> 
> Even though the intention behind trying to do those things in a hopefully-better way is laudable.
> 
> If we were a million download a month library with little to no alternative? Yeah, we could probably throw some weight around like that. But even .net itself is slowly venturing into the realm of some TG functionality, making compatibility that much more important.

Your comment is not specific enough for me to understand what you are asking. Can you please tell me, specifically, what you think I've done that is out of line with other frameworks here?

dodexahedron:
> > RE: Changed/Changing:
> > Best to stick to established convention on that one I think.
> > Enabling minimal divergence between GUI code like WPF, WinUI, etc and Terminal.Gui is a _**major**_ value add.
> > Anything we do in the public API surface that bucks established convention for common functionality is a pain point for consumers, even if they aren't trying to port something, because established and ubiquitous convention is still established and ubiquitous convention.
> > Even though the intention behind trying to do those things in a hopefully-better way is laudable.
> > If we were a million download a month library with little to no alternative? Yeah, we could probably throw some weight around like that. But even .net itself is slowly venturing into the realm of some TG functionality, making compatibility that much more important.
> 
> Your comment is not specific enough for me to understand what you are asking. Can you please tell me, specifically, what you think I've done that is out of line with other frameworks here?

The intent of that comment was discussion about how to go about this, as an answer to the question that was posed, intentionally in the abstract - not a specific comment about previous stuff.

tig:
> how to go about this...

What specifically is "this"? You could be talking about any number of things from the concept of events for "orientation" to how to define events in general.

dodexahedron:
Ah gotcha.

Specifically, it was in reference to the title's mention of event patterns.

But I think, looking at this again after some time, that I probably misinterpreted that part.

My meaning was aimed at INotifyPropertyChanging/INotifyPropertyChanged, as the standard for notification of changing/changed properties.

tig:
> My meaning was aimed at INotifyPropertyChanging/INotifyPropertyChanged, as the standard for notification of changing/changed properties.

`INotifyPropertyChanging` does not do what many people think it does.  It does NOT prevent the change from happening, only allows you to record that it is happening.

https://stackoverflow.com/a/7996089/297526

There are tons of cases where there's a need to prevent the change from happening. This is what my design enables. There is no standard I've seen for this that's widely adopted. 

And regarding `INotifyPropertyChanged`: It may be used extensively, and it has some good characteristics, but it has some awful ones as well.  It is designed primarily for data binding scenarios and (other some awful things like requiring strings for property names) works great for that. I will not adopt it at the core of TG except in cases where the use case is primarly a data binding case. 

In the future, once we get around to building a XAML-based or -like markup UI system, it will be trivial to add `INotifyPropertyChanged` support on top of the pattern I am showing here. In fact, I think it could even be automated.

PR #3627, opened 2024-07-25
Partially Adresses #2491. Refactors how `Focus` works

### LINKED ISSUE #3645: `ShadowStyle.Transparent` is not rendering bottom-left corner correctly
image

Not sure where I goofed this up, but it was working before.

### LINKED ISSUE #3669: Proposal: Focus/Navigation Changes - e.g. Remove `Tabindex` and `TabIndexes` in v2
# Remove Tabindex and TabIndexes in v2

Why?

- I can't find any real uses of it outside of the `DynamicMenubar` scenario and a simple usage in `FileDialog`.
- I've done a code search of terminal.gui/gui-cs projects and can't find any examples of where it's used.
- TGD doesn't support it https://github.com/gui-cs/TerminalGuiDesigner/issues/79
- It has a ton of complexity and is a source of bugs. The current implementation is full of bugs. I've added a bunch of unit tests in #3627 and fixed some of the bugs, but ...
- The same concept has been deprecated from Maui: https://github.com/dotnet/maui/issues/1646

If there are no objections, I am going to remove `View.TabIndex` and `View.TabIndexes` from Terminal.Gui v2 in #3627. 

## Questions

Q: What if I want to set the tab order? 
A: Simply re-order `Subviews`. 

- Side note: `Dialog` currently has a bug where in most typical use cases where  Views are added with `dlg.Add` after `Buttons` has been set`, the tab order is wrong. The buttons should always be the last items in the order, but are first. I plan on fixing this independently of whether `TabIndex` exists or not by overriding `Add` to always ensure the `Buttons` are at the end of the subview List.

Q: Web frameworks have the concept. E.g. React. Are you sure you want to remove it?
A: Yes, the same issues that plagued the original concept in Win32, then Winforms, then Xamarin, that are outlined in the Maui issue above apply to jQuery, React, etc... 

Q: What about `TabStop`?
A: The need indicates how a View *behaves* in keyboard navigation is real. I've already changed `TabStop` to be of type `TabBehavior` and I think that is still needed.

tig:
This PR paves the way for fixing #2491 by addressing deep design and implementation problems with how `Focus` currently works.

This PR used to be way more ambitions... it was going to eventually fully enable any View to work `Overlapped`, removing the need for `Toplevel` and `IsOverlappedContainer`. That will now come later.

## Fixes

- New design and implementation for how `Focus` works.
- Partially Addresses #2491
- Fixes #3669 
- Fixes #3645 

## Proposed Changes/Todos

- [x] Discovered serious issues with how HasFocus, OnEnter/OnLeave, etc... …work in some edge cases. This will require re-visiting the design at a deep level and fixing some long-standing but ignored issues such as how OnEnter/OnLeave don't follow proper cancelation design. Also, there's a need for keeping track of the old focus state of a tree of subviews when that tree loses focus; FocusDireciton is a hack that causes tons of confusion. See https://github.com/tig/Terminal.Gui/blob/8e70e2ae8faafab7cb8305ec6dbbd552c7bf3a43/docfx/docs/navigation.md
- [x] Figure out how `ViewArrangement.Overlapped` will work. Use `TabIndexes` and all the `View.NextView` focus stuff for navigation (instead of the code in `OverlappedMoveNext`). Use the `Subview` ordering (for just the subviews with ViewArrangement.Overlapped` set) to manage the z-order.
- [x] Rewrite focus engine
- [x] Dozens of new, primitive unit tests
- [x] Updated migration doc  
- [x] Massive code cleanup including making most of `View` `#nullable enable`
- [x] `View Experiments` scenario -> `Navigation Tester`
- [x] TabGroup navigation doesn't work completely right with nested groups. These are edge cases that I may tackle later.
- [x] Fix `TabView` Scenario - above item may address
- [ ] Fix `Wizards` - above item may address

## Pull Request checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ dotnet run
 * [API Documentation](https://gui-cs.github.io/Terminal.GuiV2Docs/api/Terminal.Gui.html)
 * [Documentation Home](https://gui-cs.github.io/Terminal.GuiV2Docs)
 
-The above documentation matches the most recent Nuget release from the `v2_develop` branch. Get the [v1 documentation here]( (This is the v2 API documentation. For v1 go here: https://gui-cs.github.io/Terminal.Gui/api/Terminal.Gui.html)
+The above documentation matches the most recent Nuget release from the `v2_develop` branch. Get the [v1 documentation here](This is the v2 API documentation. For v1 go here: https://gui-cs.github.io/Terminal.Gui/api/Terminal.Gui.html)
 
 See the [`Terminal.Gui/` README](https://github.com/gui-cs/Terminal.Gui/tree/master/Terminal.Gui) for an overview of how the library is structured. 
 
--- a/Terminal.Gui/Application/Application.Keyboard.cs
+++ b/Terminal.Gui/Application/Application.Keyboard.cs
@@ -1,4 +1,6 @@
 ﻿#nullable enable
+using static System.Net.Mime.MediaTypeNames;
+
 namespace Terminal.Gui;
 
 public static partial class Application // Keyboard handling
@@ -287,11 +289,11 @@ public static partial class Application // Keyboard handling
 
         AddCommand (
                     Command.NextView,
-                    static () => Current?.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabStop));
+                    static () => Application.Navigation?.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabStop));
 
         AddCommand (
-                    Command.PreviousView,
-                    static () => Current?.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabStop));
+        Command.PreviousView,
+                    static () => Application.Navigation?.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabStop));
 
         AddCommand (
                     Command.NextViewOrTop,
@@ -305,14 +307,6 @@ public static partial class Application // Keyboard handling
                                 return true;
                             };
 
-                            //// Go back down the focus chain and focus the first TabGroup
-                            //View []? views = Current.GetSubviewFocusChain (NavigationDirection.Forward, TabBehavior.TabGroup);
-
-                            //if (views.Length > 0)
-                            //{
-                            //    View []? subViews = views [0].GetSubviewFocusChain (NavigationDirection.Forward, TabBehavior.TabStop);
-                            //    return subViews? [0].SetFocus ();
-                            //}
 
                             return false;
                         }
@@ -335,14 +329,6 @@ public static partial class Application // Keyboard handling
                                 return true;
                             };
 
-                            //// Go back down the focus chain and focus the first TabGroup
-                            //View []? views = Current.GetSubviewFocusChain (NavigationDirection.Backward, TabBehavior.TabGroup);
-
-                            //if (views.Length > 0)
-                            //{
-                            //    View []? subViews = views [0].GetSubviewFocusChain (NavigationDirection.Backward, TabBehavior.TabStop);
-                            //    return subViews? [0].SetFocus ();
-                            //}
 
                             return false;
                         }
--- a/Terminal.Gui/Application/Application.Keyboard.cs
+++ b/Terminal.Gui/Application/Application.Keyboard.cs
@@ -314,21 +314,14 @@ public static partial class Application // Keyboard handling
                     static () =>
                     {
                         // TODO: This OverlapppedTop tomfoolery goes away in addressing #2491
-                        if (ApplicationOverlapped.OverlappedTop is null)
+                        if (ApplicationOverlapped.OverlappedTop is null && Current is { })
                         {
-                            if (Current is { })
-                            {
-                                return Current.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabGroup);
-                            }
-                        }
-                        else
-                        {
-                            ApplicationOverlapped.OverlappedMovePrevious ();
-
-                            return true;
+                            return Current.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabGroup);
                         }
 
-                        return false;
+                        ApplicationOverlapped.OverlappedMovePrevious ();
+
+                        return true;
                     }
                    );
 
--- a/Terminal.Gui/Application/Application.Keyboard.cs
+++ b/Terminal.Gui/Application/Application.Keyboard.cs
@@ -1,6 +1,4 @@
 ﻿#nullable enable
-using static System.Net.Mime.MediaTypeNames;
-
 namespace Terminal.Gui;
 
 public static partial class Application // Keyboard handling
@@ -289,31 +287,25 @@ public static partial class Application // Keyboard handling
 
         AddCommand (
                     Command.NextView,
-                    static () => Application.Navigation?.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabStop));
+                    static () => Navigation?.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabStop));
 
         AddCommand (
-        Command.PreviousView,
-                    static () => Application.Navigation?.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabStop));
+                    Command.PreviousView,
+                    static () => Navigation?.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabStop));
 
         AddCommand (
                     Command.NextViewOrTop,
                     static () =>
                     {
                         // TODO: This OverlapppedTop tomfoolery goes away in addressing #2491
-                        if (ApplicationOverlapped.OverlappedTop is null && Current is { })
+                        if (ApplicationOverlapped.OverlappedTop is { })
                         {
-                            if (Current.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabGroup))
-                            {
-                                return true;
-                            };
+                            ApplicationOverlapped.OverlappedMoveNext ();
 
-
-                            return false;
+                            return true;
                         }
 
-                        ApplicationOverlapped.OverlappedMoveNext ();
-
-                        return true;
+                        return Navigation?.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabGroup);
                     }
                    );
 
@@ -322,20 +314,14 @@ public static partial class Application // Keyboard handling
                     static () =>
                     {
                         // TODO: This OverlapppedTop tomfoolery goes away in addressing #2491
-                        if (ApplicationOverlapped.OverlappedTop is null && Current is { })
+                        if (ApplicationOverlapped.OverlappedTop is { })
                         {
-                            if (Current.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabGroup))
-                            {
-                                return true;
-                            };
+                            ApplicationOverlapped.OverlappedMovePrevious ();
 
-
-                            return false;
+                            return true;
                         }
 
-                        ApplicationOverlapped.OverlappedMovePrevious ();
-
-                        return true;
+                        return Navigation?.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabGroup);
                     }
                    );
 
@@ -367,8 +353,8 @@ public static partial class Application // Keyboard handling
         KeyBindings.Add (NextTabKey, KeyBindingScope.Application, Command.NextView);
         KeyBindings.Add (PrevTabKey, KeyBindingScope.Application, Command.PreviousView);
 
-        KeyBindings.Add (NextTabGroupKey, KeyBindingScope.Application, Command.NextViewOrTop); // Needed on Unix
-        KeyBindings.Add (PrevTabGroupKey, KeyBindingScope.Application, Command.PreviousViewOrTop); // Needed on Unix
+        KeyBindings.Add (NextTabGroupKey, KeyBindingScope.Application, Command.NextViewOrTop);
+        KeyBindings.Add (PrevTabGroupKey, KeyBindingScope.Application, Command.PreviousViewOrTop);
 
         // TODO: Refresh Key should be configurable
         KeyBindings.Add (Key.F5, KeyBindingScope.Application, Command.Refresh);
--- a/Terminal.Gui/Application/Application.Keyboard.cs
+++ b/Terminal.Gui/Application/Application.Keyboard.cs
@@ -298,21 +298,14 @@ public static partial class Application // Keyboard handling
                     static () =>
                     {
                         // TODO: This OverlapppedTop tomfoolery goes away in addressing #2491
-                        if (ApplicationOverlapped.OverlappedTop is null)
+                        if (ApplicationOverlapped.OverlappedTop is null && Current is { })
                         {
-                            if ((View?)Current is { })
-                            {
-                                return Current.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabGroup);
-                            }
-                        }
-                        else
-                        {
-                            ApplicationOverlapped.OverlappedMoveNext ();
-
-                            return true;
+                            return Current.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabGroup);
                         }
 
-                        return false;
+                        ApplicationOverlapped.OverlappedMoveNext ();
+
+                        return true;
                     }
                    );
 
@@ -323,7 +316,7 @@ public static partial class Application // Keyboard handling
                         // TODO: This OverlapppedTop tomfoolery goes away in addressing #2491
                         if (ApplicationOverlapped.OverlappedTop is null)
                         {
-                            if ((View?)Current is { })
+                            if (Current is { })
                             {
                                 return Current.AdvanceFocus (NavigationDirection.Backward, TabBehavior.TabGroup);
                             }
--- a/Terminal.Gui/Application/Application.Keyboard.cs
+++ b/Terminal.Gui/Application/Application.Keyboard.cs
@@ -144,28 +144,37 @@ public static partial class Application // Keyboard handling
 
                 foreach (Command command in appBinding.Commands)
                 {
-                    if (!CommandImplementations!.ContainsKey (command))
-                    {
-                        throw new NotSupportedException (
-                                                         @$"A KeyBinding was set up for the command {command} ({keyEvent}) but that command is not supported by Application."
-                                                        );
-                    }
-
-                    if (CommandImplementations.TryGetValue (command, out Func<CommandContext, bool?>? implementation))
-                    {
-                        var context = new CommandContext (command, keyEvent, appBinding); // Create the context here
-                        toReturn = implementation (context);
-                    }
-
-                    // if ever see a true then that's what we will return
-                    if (toReturn ?? false)
-                    {
-                        toReturn = true;
-                    }
+                    toReturn = InvokeCommand (command, keyEvent, appBinding);
                 }
 
                 return toReturn ?? true;
             }
+        }
+
+        return false;
+    }
+
+    /// <summary>
+    /// INTENRAL method to invoke one of the commands in <see cref="CommandImplementations"/>
+    /// </summary>
+    /// <param name="command"></param>
+    /// <param name="keyEvent"></param>
+    /// <param name="appBinding"></param>
+    /// <returns></returns>
+    /// <exception cref="NotSupportedException"></exception>
+    private static bool? InvokeCommand (Command command, Key keyEvent, KeyBinding appBinding)
+    {
+        if (!CommandImplementations!.ContainsKey (command))
+        {
+            throw new NotSupportedException (
+                                             @$"A KeyBinding was set up for the command {command} ({keyEvent}) but that command is not supported by Application."
+                                            );
+        }
+
+        if (CommandImplementations.TryGetValue (command, out Func<CommandContext, bool?>? implementation))
+        {
+            var context = new CommandContext (command, keyEvent, appBinding); // Create the context here
+            return implementation (context);
         }
 
         return false;
@@ -364,13 +373,6 @@ public static partial class Application // Keyboard handling
         {
             KeyBindings.Add (Key.Z.WithCtrl, KeyBindingScope.Application, Command.Suspend);
         }
-
-#if UNIX_KEY_BINDINGS
-        KeyBindings.Add (Key.L.WithCtrl, Command.Refresh); // Unix
-        KeyBindings.Add (Key.F.WithCtrl, Command.NextView); // Unix
-        KeyBindings.Add (Key.I.WithCtrl, Command.NextView); // Unix
-        KeyBindings.Add (Key.B.WithCtrl, Command.PreviousView); // Unix
-#endif
     }
 
     /// <summary>
--- a/Terminal.Gui/Application/Application.Keyboard.cs
+++ b/Terminal.Gui/Application/Application.Keyboard.cs
@@ -1,45 +1,49 @@
 ﻿#nullable enable
-using System.Text.Json.Serialization;
-
 namespace Terminal.Gui;
 
 public static partial class Application // Keyboard handling
 {
+    private static Key _nextTabGroupKey = Key.F6; // Resources/config.json overrrides
     private static Key _nextTabKey = Key.Tab; // Resources/config.json overrrides
 
-    /// <summary>Alternative key to navigate forwards through views. Ctrl+Tab is the primary key.</summary>
-    [SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
-    public static Key NextTabKey
-    {
-        get => _nextTabKey;
-        set
-        {
-            if (_nextTabKey != value)
-            {
-                ReplaceKey (_nextTabKey, value);
-                _nextTabKey = value;
-            }
-        }
-    }
+    private static Key _prevTabGroupKey = Key.F6.WithShift; // Resources/config.json overrrides
 
     private static Key _prevTabKey = Key.Tab.WithShift; // Resources/config.json overrrides
 
-    /// <summary>Alternative key to navigate backwards through views. Shift+Ctrl+Tab is the primary key.</summary>
-    [SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
-    public static Key PrevTabKey
-    {
-        get => _prevTabKey;
-        set
-        {
-            if (_prevTabKey != value)
-            {
-                ReplaceKey (_prevTabKey, value);
-                _prevTabKey = value;
-            }
-        }
-    }
+    private static Key _quitKey = Key.Esc; // Resources/config.json overrrides
 
-    private static Key _nextTabGroupKey = Key.F6; // Resources/config.json overrrides
+    static Application () { AddApplicationKeyBindings (); }
+
+    /// <summary>Gets the key bindings for this view.</summary>
+    public static KeyBindings KeyBindings { get; internal set; } = new ();
+
+    /// <summary>
+    ///     Event fired when the user presses a key. Fired by <see cref="OnKeyDown"/>.
+    ///     <para>
+    ///         Set <see cref="Key.Handled"/> to <see langword="true"/> to indicate the key was handled and to prevent
+    ///         additional processing.
+    ///     </para>
+    /// </summary>
+    /// <remarks>
+    ///     All drivers support firing the <see cref="KeyDown"/> event. Some drivers (Curses) do not support firing the
+    ///     <see cref="KeyDown"/> and <see cref="KeyUp"/> events.
+    ///     <para>Fired after <see cref="KeyDown"/> and before <see cref="KeyUp"/>.</para>
+    /// </remarks>
+    public static event EventHandler<Key>? KeyDown;
+
+    /// <summary>
+    ///     Event fired when the user releases a key. Fired by <see cref="OnKeyUp"/>.
+    ///     <para>
+    ///         Set <see cref="Key.Handled"/> to <see langword="true"/> to indicate the key was handled and to prevent
+    ///         additional processing.
+    ///     </para>
+    /// </summary>
+    /// <remarks>
+    ///     All drivers support firing the <see cref="KeyDown"/> event. Some drivers (Curses) do not support firing the
+    ///     <see cref="KeyDown"/> and <see cref="KeyUp"/> events.
+    ///     <para>Fired after <see cref="KeyDown"/>.</para>
+    /// </remarks>
+    public static event EventHandler<Key>? KeyUp;
 
     /// <summary>Alternative key to navigate forwards through views. Ctrl+Tab is the primary key.</summary>
     [SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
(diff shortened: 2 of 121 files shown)


tig:
@BDisp I would very much appreciate you reviewing the new `Navigation.md` overview document in this PR and providing feedback before I go further in implementation. You have a ton of experience with the current system for `Focus` and keyboard navigation.

You may also want to pull down the latest code and play with `ViewExperiments`. It's not fully working but is showing some progress. 

Thanks.

Link here:  https://github.com/tig/Terminal.Gui/blob/8e70e2ae8faafab7cb8305ec6dbbd552c7bf3a43/docfx/docs/navigation.md

BDisp:
Change the title please, otherwise it will close the issue when merged.

dodexahedron:
I know this is a WIP but don't forget nullability context! In the top few lines of the classes I can already see places that need those annotations.

tig:
> I know this is a WIP but don't forget nullability context! In the top few lines of the classes I can already see places that need those annotations.

Question: how does it work with partials? I'm resisting adding it to `View.Navigation.cs` in this PR because I don't want to totally upgrade all of View in this PR.

dodexahedron:
> > I know this is a WIP but don't forget nullability context! In the top few lines of the classes I can already see places that need those annotations.
> 
> Question: how does it work with partials? I'm resisting adding it to `View.Navigation.cs` in this PR because I don't want to totally upgrade all of View in this PR.

It's a per-file directive or, if paired with `#nullable restore`, can be limited to specific sections of code.

dodexahedron:
But IMO it makes the most sense to just turn it on for the whole project at this point, and people can just address warnings as they touch files or _opt out_ rather than opt in, if they need to for some reason.

To silence the warnings in the build, so we only see them at design time in the IDE, you can add `-nowarn nullable` to the build command line in CI.

Another option, which is a half-way means of doing it project-wide, would be to set the `<Nullable>` element in the project to annotations or warnings, to set it to one of those partial modes project-wide.

The [nullable migration strategies](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies) document is quite helpful for understanding the best ways to ease into fully embracing nullability context.

tig:
> But IMO it makes the most sense to just turn it on for the whole project at this point, and people can just address warnings as they touch files or _opt out_ rather than opt in, if they need to for some reason.
> 

Agree. But only after this PR is merged, ok?

tig:
Ok, @tig, @tznind, and @dodexahedron - this is ready for review!

Stuff I'm still working on:

- [ ] I want to make `ViewExperiments.cs` be `NavigationTester.cs`
- [ ] `TabGroup` navigation doesn't work completely right with nested groups. These are edge cases that I may tackle later.

tig:
> Ok, @tig, @tznind, and @dodexahedron - this is ready for review!
> 
> Stuff I'm still working on:
> 
> * [ ]  I want to make `ViewExperiments.cs` be `NavigationTester.cs`
> * [ ]  `TabGroup` navigation doesn't work completely right with nested groups. These are edge cases that I may tackle later.

Recommend you read the new API docs and the migration guide and `navigation.md` as part of your review.

dodexahedron:
Will have a look in a little bit, once I get back home. 👍

Terminal.Gui/Application/Application.Keyboard.cs
```diff
@@ -332,39 +332,71 @@ internal static void AddApplicationKeyBindings ()
                     Command.NextView,
                     () =>
                     {
-                        ApplicationNavigation.MoveNextView ();
-
-                        return true;
+                        View? current = Application.Current;
```
dodexahedron, line 335: No real reason to grab this reference. Just in-line it (`ctrl + r, i` by default in ReSharper with the caret on `current`). 

In fact, lines 335 through 340 can reduce all the way down to:

```csharp
return Application.Current?.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabStop)
```

Also, please make any anonymous function static that can be. It makes a pretty big difference in the JIT output.

If you start them off as `static () => {}` before you even write their bodies, it'll also help force you to avoid closure capture of locals, which is a nice perk.

Looks like this one can be static, since it doesn't reference anything that would disqualify it.

That results in the AddCommand call starting on line 331 turning into something like this:

```csharp
AddCommand (
            Command.NextView,
            static () => Application.Current?.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabStop));
```

Same simplification available with the rest, it look like.

Terminal.Gui/Application/Application.Mouse.cs
```diff
@@ -143,6 +143,12 @@ internal static void OnMouseEvent (MouseEvent mouseEvent)
 
         if (view is { })
         {
+#if DEBUG_IDISPOSABLE
+            if (view.WasDisposed)
+            {
+                throw new ObjectDisposedException (view.GetType ().FullName);
```
dodexahedron, line 149: I'm torn on this one...

GetType() is not something we should be using... But this is also a conditional only emitted if the symbol is defined, so.... I guess it's fine?

In any case, here's something for you, to make all your conditional throws much cleaner:

```csharp

    /// <summary>
    ///     Throws an <see cref="ObjectDisposedException"/> if the provided <paramref name="condition"/> is <see langword="true"/>.
    /// </summary>
    /// <typeparam name="T">
    ///     The type of the object that was disposed. Compiler will figure it out if you provide anything to
    ///     <paramref name="possiblyDisposedObject"/>.
    /// </typeparam>
    /// <param name="condition">The condition upon which, if true, the exception will be thrown. If false, method simply returns.</param>
    /// <param name="possiblyDisposedObject">The object that we are complaining about. Only used by the method for determining type.</param>
    /// <param name="name">The name of the object. If not specified, the full name of its type will be used.</param>
    /// <param name="message">A message to include with the exception. If not specified, "Object already disposed" is used.</param>
    /// <exception cref="ObjectDisposedException">If the condition is true.</exception>
    [Conditional ("DEBUG_IDISPOSABLE")]
    internal static void ThrowObjectDisposedExceptionIfTrue<T> (
        [DoesNotReturnIf (true)] this bool condition,
        T? possiblyDisposedObject = null,
        string? name = null,
        string? message = null
    ) where T : class, IDisposable
    {
        if (condition)
        {
            throw new ObjectDisposedException (name ?? typeof (T).FullName, message ?? "Object was already disposed.");
        }
    }
```

That is written as an extension method on bool but of course can also be called as a regular static method. 

If you don't want to stick it in a static class, just remove the `this` from the condition parameter.

That method replaces the entirety of this:

```csharp
#if DEBUG_IDISPOSABLE
            if (view.WasDisposed)
            {
                throw new ObjectDisposedException (view.GetType ().FullName);
            }
#endif
```

with JUST this:

```csharp
view.WasDisposed.ThrowObjectDisposedExceptionIfTrue(view);
```

or

```csharp
ThrowObjectDisposedExceptionIfTrue(view.WasDisposed,view);
```

No conditional required. The attribute took care of that.

Terminal.Gui/Application/Application.Run.cs
```diff
@@ -99,7 +99,7 @@ public static RunState Begin (Toplevel toplevel)
             else if (ApplicationOverlapped.OverlappedTop is { } && toplevel != Top && TopLevels.Contains (Top!))
             {
                 // BUGBUG: Don't call OnLeave/OnEnter directly! Set HasFocus to false and let the system handle it.
-                Top!.OnLeave (toplevel);
+                //Top!.OnLeave (toplevel);
```
dodexahedron, line 102: Can the line be removed?

Terminal.Gui/Application/Application.Run.cs
```diff
@@ -186,7 +186,14 @@ public static RunState Begin (Toplevel toplevel)
 
         toplevel.LayoutSubviews ();
         toplevel.PositionToplevels ();
-        toplevel.FocusFirst (null);
+
+        // TODO: Should this use FindDeepestFocusableView instead?
```
dodexahedron, line 190: Answering that question:

I think we should just define a `FindDeepestView` method that takes a delegate that is responsible for whatever the deepest view it's supposed to find is all about.

The basic function of finding the deepest view that is, has, or does X is the same, so duplicating that is not very DRY and is a point for bugs/regressions to creep in, as multiple disparate implementations of it diverge due to other changes.

That predicate delegate (or delegates, if appropriate) would only be responsible for filling in the blanks in the recursion for the base case and such.

Or, better yet....

If we implement IEnumerable on View, we can skip that work and use LINQ methods to do it that smarter people have already figured out how to write. 🙂

Terminal.Gui/Application/Application.Run.cs
```diff
@@ -186,7 +186,14 @@ public static RunState Begin (Toplevel toplevel)
...
+        // Try to set initial focus to any TabStop
+        if (!toplevel.AdvanceFocus (NavigationDirection.Forward, TabBehavior.TabStop))
```
dodexahedron, line 192: This feels like something that belongs in a different type, as it's making a specific behavior dictated by another class into something that this class is directly making decisions on. 

Alternatively, can't the AdvanceFocus method make this determination, potentially even based on type checks to avoid reliance on the enum? At minimum, the same operation, but performed by AdvanceFocus is a more appropriate place than here.

Terminal.Gui/Application/Application.cs
```diff
@@ -138,6 +138,8 @@ internal static List<CultureInfo> GetSupportedCultures ()
     // starts running and after Shutdown returns.
     internal static void ResetState (bool ignoreDisposed = false)
     {
+        Application.Navigation = new ApplicationNavigation ();
```
dodexahedron, line 141: Seems unsafe or at least unclean to just charge right into this without any conditions at all, no?

Terminal.Gui/Application/ApplicationNavigation.cs
```diff
@@ -55,14 +55,14 @@ internal void SetFocused (View? value)
...
-    public static bool IsInHierarchy (View start, View? view)
+    public static bool IsInHierarchy (View? start, View? view)
     {
         if (view is null)
         {
             return false;
         }
 
-        if (view == start)
+        if (view == start || start is null)
```
dodexahedron, line 65: How about replacing all of these conditions with a switch similar to the following (or, if all cases are just returns, then the equivalent switch expression, which is super-compact):

```csharp
switch(view,start)
  case (null,_):
    return false;
  case (_,null):
  case (_,_) when view == start:
    return true;
//etc
}
```

dodexahedron (changes requested):
A doc bug and a couple of things that'll reduce line count in this PR a fair bit.

I stopped specifically commenting after a few files for now, since a lot of the things I suggested for those also apply to several other files.

I'll re-review more completely when the next iteration is ready! :)

Loving it so far.

Terminal.Gui/Application/Application.Mouse.cs
dodexahedron: Also, the reason for all these disposal checks goes away if we properly handle lifetime and ownership of IDisposables.

I might move directly addressing some of that stuff up on my priority list. These situations represent pretty critical and potentially even dangerous (security-wise) bugs and should not make it into the final version.

dodexahedron:
> > But IMO it makes the most sense to just turn it on for the whole project at this point, and people can just address warnings as they touch files or _opt out_ rather than opt in, if they need to for some reason.
> 
> Agree. But only after this PR is merged, ok?

I've got no problem with that.

I've long been operating with a .csproj.user file that has it turned on globally, anyway, among a few other analysis-related things I have turned up to 11. You should see the warning counts. 😅

Terminal.Gui/Application/Application.Mouse.cs
dodexahedron: Oh another thing. If you like the extension method style but only want to touch  spcific file, you can define the class as:

```csharp
file static class NiftyThrowHelpers
{
  // Extension methods
}
```

The `file` access modifier makes the type only visible within that single file. You can even define classes with identical names multiple times in the same namespace, so long as they are in different files, if you do that. A potential use case for that is this sort of thing or platform-specific code with a unified interface.

It's even more private than private! 🤯

Terminal.Gui/Application/Application.Keyboard.cs
tig: Actually,, they can all be static. Thanks.

Terminal.Gui/Application/Application.Mouse.cs
tig: Love this, but not for this PR. Please open a new issue for this concept.

Terminal.Gui/Application/Application.Run.cs
tig: It def will be removed when Toplevel goes away as part of fixing #2419!

tig: I didn't remove it because my focus changes probably broke `OverlappedTop`. But I don't care because that's going away in the next big PR towards #2419.

Terminal.Gui/Application/Application.cs
tig: I don't think so. ResetState is intended to be a very strong armed method. 

I keep going back and forth, though on the right semantics for objects like `ApplicationNavigation`, held by a property of a 'singleton static' like `Application`. Yes, I know your opinion on `Application` as a static, and I'd like to fix that eventually but not for v2. 

So... Given `Application` is a singleton-static and needs to be setup/reset for each new "app", what is the right pattern for ensuring non-static object refs it holds to be created/destroyed? 

Since `ApplicationNavigation` is actually used as a singleton, perhaps it should be implemented using the correct singleton pattern? Would appreciate your thoughts on this.

Terminal.Gui/Application/ApplicationNavigation.cs
tig: I'm still getting used to parsing switch expressions... at a glance they are un-readable to me. See the 1st tenet below:

## Tenets for [gui-cs](www.github.com/gui-cs) Code Style (Unless you have better ones)

* **Six-Year-Old Reading Level** - Our code style is biased towards code readability and away from terseness. This is *Systems Software* and needs to stand the test of time. Code should be structured and use variable names that make it readable by a 6-year-old, and comments in code are encouraged. 
* **Consistency, Consistency, Consistency** - We adopt and document our standards for code style and then enforce them ruthlessly. For example, we require code reviews to pay attention to code style, not just functionality. 
* **Don't be Weird** - Like all developers we have opinions, but our opinions on code style are tempered by existing standards. We are biased towards code style that used by Microsoft and other leading dotnet developers. For example, we choose 4 spaces for indentation instead of 8.
* **Set and Forget** - We embrace and encourage the use of technology that makes it easy for contributors to apply best-practice code-style, such as ReSharper. As we do so we are mindful that tools can cause hidden issues and merge hell.
* **Documentation is the Spec** - We care deeply about providing delightful developer documentation and are sticklers for grammar and clarity. If the code and the docs conflict, we are biased to believe what we wrote in the API documentation. This drives a virtuous cycle of clear thinking.

from: https://github.com/gui-cs/Terminal.Gui/pull/3207#issuecomment-1908203789

(Which I meant to pull out into `CONTRIBUTING.md` but forgot. Will do).

Terminal.Gui/Application/Application.Run.cs
tig: @dodexahedron: Your writing is insightful, but the use of the pronoun “This” is frequently ambiguous and makes the reader work too hard to decode. 

For example:

“This feels like something that belongs in a different type, as it’s making a specific behavior dictated by another class into something that this class is directly making decisions on.”

To improve clarity, replace “This” with a specific noun. For instance:

“The behavior dictated by `View` should not be directly decided by `Application`.”

(I actually don't know if the above represents what you meant, BTW).

This change will make your references clearer and your argument stronger and will make your readers lives easier.

Terminal.Gui/Application/Application.Keyboard.cs
dodexahedron: Nice

Terminal.Gui/Application/Application.Mouse.cs
dodexahedron: If not included here, I'll likely stick it in the driver one I'm working on, since I can make use of it there, too.

I'll just do it file-scoped.

Terminal.Gui/Application/Application.Run.cs
dodexahedron: Ha understandable. Intermediate changes. 👍

Terminal.Gui/Application/Application.Run.cs
dodexahedron: Oh yeah totally not for this PR. I didn't think that question was scoped to this PR anyway. :)

Terminal.Gui/Application/Application.Run.cs
dodexahedron: Sorry, I know that still slips in quite a bit. I'm working on that per your previous request, though, so it ain't falling on deaf ears (blind eyes?).

Specifically, this is referring to the code block it annotates, which is that if/else construct that calls the same method (AdvanceFocus) on another object (twice) to take action on that same object. The entire operation is therefore something that `View` does, _to_ `View`, and thus belongs as a method in `View` itself. The AdvanceFocus method itself can still be what it is, but the if/else in this block should be either part of the AdvanceFocus method itself or another method in `View` that calls `AdvanceFocus`.

As part of Application, it's just tight coupling and violation of encapsulation.

Terminal.Gui/Application/Application.cs
dodexahedron: Certainly the nuclear nature of it is understandable.

But what I specifically mean is that just reassigning a reference without ensuring cleanup in at least a naive way may not be ideal. I need to go back and look at the ApplicationNavigation class again, though, to see what (if anything) that is relevant for.

dodexahedron: As for the singleton pattern, I would ***strongly*** advise against continuing to make TG _more_ static than it already is.

Terminal.Gui/Application/Application.Run.cs
dodexahedron: But also, _why can't you just read my mind????_ 😅

tig: > Sorry, I know that still slips in quite a bit. I'm working on that per your previous request, though, so it ain't falling on deaf ears (blind eyes?).
> 
> Specifically, this is referring to the code block it annotates, which is that if/else construct that calls the same method (AdvanceFocus) on another object (twice) to take action on that same object. The entire operation is therefore something that `View` does, _to_ `View`, and thus belongs as a method in `View` itself. The AdvanceFocus method itself can still be what it is, but the if/else in this block should be either part of the AdvanceFocus method itself or another method in `View` that calls `AdvanceFocus`.
> 
> As part of Application, it's just tight coupling and violation of encapsulation.

Great point. 

Turns out I was being stoopid^2 and all that really was needed here was a call to `top.SetFocus` anyway.

Terminal.Gui/Application/ApplicationNavigation.cs
dodexahedron: All good either way, as far as I'm concerned.

I can certainly empathize with learning unfamiliar constructs or things one might just not be highly confident with at a given time. Personally, having worked with C# over its whole existence has kept me on my toes because it never stops changing and I learn something new darn near every day about either the language or .net, still, almost 24 years later! 🤯 

OOOO thanks for moving that whenever you get to it. I always thought it was in a sub-optimal and slightly not-so-discoverable place, before. 👍 

#### Philosophical trivialities, ramblings, and opinionated blathering

Welcome to the wasting of the next...hmm...Maybe 1-5 minutes of your life? 😅 

As to the applicability of the referenced tenet _here_, specifically, ehhhh... I don't think it applies, when we're talking about standard, idiomatic C#, especially with constructs that are not remotely new to the language. C# is pretty big, .net is bigger, and programming is ridicu-hard when you step back and look at what we do. One could pretty arbitrarily pick almost any programming concept and call it complex while someone else thinks it's the simplest and best thing since sliced bread (which I, incidentally, don't find to be so awesome as to deserve its own idiomatic expression 😆).

Things like the concept of readability are more subjective than objective, in particular. I, personally, tend to find **well-written** and **appropriately-used** switches more readable, less prone to easily-made and difficult-to-locate mistakes, clearer in how many cases there actually are,[^myGodItsFullOfStars] and easier to parse in general since they are fully declarative, by specification of the language. I especially like them when they reduce otherwise large and widely-separated conditionals or, on the other end of the spectrum, simple but very subtly different conditions in interdependent statements down to a literal truth table. But those emphasized terms are critical components of it, as you can just as easily make a monstrosity that is hella indecipherable to any of us ugly bags of mostly water.

There are times when one of the three (if/else if/else, switch statement, or switch expression) may have technical advantages as well, but if that might matter I'll point that fact out - definitely will if it's important enough.

Which is why _this one_ is just written as a soft suggestion rather than an imperative with technical justification (you know - in my signature ultra-terse, concise, and devoid of dad joke-esque humor way).

As a reminder: Usually, when I make a suggestion like this, it's as much for presenting alternatives with potential benefits (I won't make any that are _completely_ pointless) as it is for enrichment of those reading the review comment, be they the author of the code, other project members, other people who randomly stumble upon it due to a google search that turns something up that makes them have their lightbulb moment, or myself 6 months from now when I can't figure something out due to coming down with a relapse of stoopid. 😅🤷‍♂️ 

[^myGodItsFullOfStars]: `if` often hides soooooo many possibilities unless it truly is a binary choice and the `else` is fully valid no matter which individual part of the boolean condition made it evaluate to false, and the complexity of that grows polynomially per condition and quadratically component of each condition.

tig:
I'm pretty tired of working on this. It was a massive effort... but well worth it IMO.

I'm bummed about what's left:

- [ ] TabGroup navigation doesn't work completely right with nested groups. These are edge cases that I may tackle later.
- [ ] Fix `TabView` Scenario - above item may address
- [ ] Fix `Wizards` - above item may address

I'm bummed because I want to work on something else for a bit.

At the same time, this is such a big PR I really want it merged.

Any objections to me leaving the above broken (I'll create Issues to track) for a bit?

Terminal.Gui/Application/Application.Mouse.cs
BDisp: Did you already realize that when you click on a `Border` of an open `Menu` this also throw tis exception for the `Border` adornment? The reason is that the `Menu` was disposed and the `HandlePressed` method is grabbing the menu's `Border`.

BDisp (changes requested):
Great work. Some suggestions and possible bugs.

Terminal.Gui/Application/Application.Mouse.cs
tig: > Did you already realize that when you click on a `Border` of an open `Menu` this also throw tis exception for the `Border` adornment? The reason is that the `Menu` was disposed and the `HandlePressed` method is grabbing the menu's `Border`.

Suggestion on how to fix?

BDisp: > Suggestion on how to fix?

The ideal is the `View` class had a `Disposed` property that must also available in `Debug` and Release `modes`. Attempting accessing a disposed object, two situations may occur:
1 - The disposed object is already null and `FindDeepestView` may not see it and it's ok.
2 - The disposed object is not null and `FindDeepestView` should check for the `Disposed` property and not return it, continuing search for the view under that position.

BDisp:
That better but causes select the view under the `Border`.

BDisp:
There is another bug with this by only resizing the border height after click and moving another view.

WindowsTerminal_8yd3k3DdVs

tig:
> There is another bug with this by only resizing the border height after click and moving another view.
> 

No bug. You're not actually selecting the Window by clicking on the Border. Click within the padding.

image

BDisp:
@tig I refer when I colapse the slip the height doesn't update until I move the mouse to another view.

tig:
> @tig I refer when I colapse the slip the height doesn't update until I move the mouse to another view.

Oh. Yeah, that happens in v2_develop as well.

BDisp:
> Oh. Yeah, that happens in v2_develop as well.

Do you think it's a bug on the `Dim.Auto (DimAutoStyle.Content)`?

tig:
> > Oh. Yeah, that happens in v2_develop as well.
> 
> Do you think it's a bug on the `Dim.Auto (DimAutoStyle.Content)`?

We're missing SetLayoutNeeded or LayoutSubviews somewhere.

tig:
> > > Oh. Yeah, that happens in v2_develop as well.
> > 
> > 
> > Do you think it's a bug on the `Dim.Auto (DimAutoStyle.Content)`?
> 
> We're missing SetLayoutNeeded or LayoutSubviews somewhere.

IIRC, I think I suspected it was someting to do with SubViewsNeedsLayout not getting set right. The logic of all of that is pretty complex.

Terminal.Gui/Application/Application.Mouse.cs
dodexahedron: > Why not just use the `view` instance instead of using reflection by calling `GetType` on all occurrences?
> 
> `throw new ObjectDisposedException (nameof (view));`

GetType is banned by the AoT compiler.

Use a generic and `typeof(T)` or type checks when you would otherwise be trying to call GetType(). Don't pass Type objects around if you can help it. Even outside of native AoT, doing that breaks one of the biggest value propositions of the managed runtime and static analysis: type-safety.

The compiler can't tell ahead of time what a Type variable is going to be. But a type parameter is a compile-time construct that _does_ result in a closed model in the final JITed binary.

Terminal.Gui/Application/Application.Run.cs
dodexahedron: > Turns out I was being stoopid^2 and all that really was needed here was a call to `top.SetFocus` anyway.

Exactly where I hoped that would end up once you saw it with fresh eyes. 😉

Hmmm... Instead of "stoopid," I kinda like the sound of "exploring alternative avenues in order to nurture nascent ingenuity," don't you? Much more enterprisey. 😆

dodexahedron:
> I'm pretty tired of working on this. It was a massive effort... but well worth it IMO.
> 
> I'm bummed about what's left:
> 
> * [ ]  TabGroup navigation doesn't work completely right with nested groups. These are edge cases that I may tackle later.
> * [x]  Fix `TabView` Scenario - above item may address
> * [ ]  Fix `Wizards` - above item may address
> 
> I'm bummed because I want to work on something else for a bit.
> 
> At the same time, this is such a big PR I really want it merged.
> 
> Any objections to me leaving the above broken (I'll create Issues to track) for a bit?

Feel you, I do.

...Why does that sound so much worse than "I feel you?" I half wanted to put a 😏 on it.

dodexahedron:
Oh forgot to say...

Yeah, I'll take a final pass over it after all your revisions, in case anything significant sticks out, which I'm starting as soon as I grab a source of caffeine.

Huge it definitely is, so it'll likely be a couple hours before I'm finished, simply to just read it all.

Loving the work. Focus was one of my biggest gripes from the before times, and this is a huge improvement.

Terminal.Gui/Views/Menu/MenuBar.cs
```diff
@@ -1339,14 +1339,14 @@ private bool Select (int index)
     #region Mouse Handling
 
     /// <inheritdoc/>
-    public override bool OnLeave (View view)
+    internal void LostFocus (View view)
     {
-        if (((!(view is MenuBar) && !(view is Menu)) || (!(view is MenuBar) && !(view is Menu) && _openMenu is { })) && !_isCleaning && !_reopen)
+        if (((!(view is MenuBar) && !(view is Menu))) && !_isCleaning && !_reopen)
```
dodexahedron, line 1344: You can make the boolean expression here soooooo much easier to grok by using pattern syntax:

```csharp
if(view is not MenuBar and not Menu && !_isCleaning && !_reopen)
```

Or, by deMorgan, its logical equivalent would be:

```csharp
if ((view is MenuBar or Menu || _isCleaning || _reOpen) is false)

// or

if (!(view is MenuBar or Menu || _isCleaning || _reOpen)) //But I think it's easy to miss the bang.
```

dodexahedron (approved):
I'm not finding any show-stoppers after getting down to TabView.cs.

But it's time for me to hit the hay.

Biggest issues I do see are some pretty obvious performance issues like lines 323 through 325 of ComboBox.cs, where a collection is enumerated multiple times (up to 6 in the worst case). To be fair, that isn't _newly_ broken, so I didn't comment on that and other similar things in-line that weren't actually changed in this PR aside from formatting.

But.

Can you at least drop a `// PERF: Fix this` comment or something at the top of OnHasFocusChanged on [Line 308 of ComboBox.cs](https://github.com/gui-cs/Terminal.Gui/pull/3627/files#diff-2f1387bc4492b95a308f6eb92fda05f8bb7250ae154503edda7717cae4be344eR308) so it doesn't get overlooked later? That one in particular is just being needlessly cruel to the runtime.

There was one little syntax clarity tip I dropped in there, though, to demonstrate an alternative.

I'm gonna say, at this point, unless you want me to finish with the rest tomorrow, which is almost entirely test code, that it looks a'ight to me, with both the "@dodexahedron is bailing for the night" and the "@tig's bored of this one" understanding. 😅

Terminal.Gui/Views/Menu/MenuBar.cs
dodexahedron: Just following up on that one.

Any time you've got a bunch of "not" in an expression, especially if it leads to a ton of parentheses or non-obvious operator precedence, consider deMorgan.

It's also yet another thing R# can just do for you effortlessly, reversibly, and non-destructively. 🥳 

Place the caret on one of the && or || operators and then use the "`Replace && with ||`" (or the other way around) option in the hammer/lightbulb menu. That does deMorgan for you. ***DON'T*** pick the option that says "`Negate && (change semantics)`" because that one will do exactly what it says - invert the actual logical result AND deMorgan it. You probably don't want that.

tig:
yYsChMT 3

tig:
Fixed some overlapped stuff while on a flight. Not really in scope for this PR, but making progress against #2491.

Lots to do to make it all work exactly right, but its promising. With far less code than was required to support overlapped in v1.

rrRYBQH 1

README.md
```diff
@@ -35,7 +35,7 @@ dotnet run
 * [API Documentation](https://gui-cs.github.io/Terminal.GuiV2Docs/api/Terminal.Gui.html)
 * [Documentation Home](https://gui-cs.github.io/Terminal.GuiV2Docs)
 
-_The Documentation matches the most recent Nuget release from the `v2_develop` branch. The documentation for v1 is here: ([![Version](https://img.shields.io/nuget/v/Terminal.Gui.svg)](https://www.nuget.org/packages/Terminal.Gui))_
+The above documentation matches the most recent Nuget release from the `v2_develop` branch. Get the [v1 documentation here](This is the v2 API documentation. For v1 go here: https://gui-cs.github.io/Terminal.Gui/api/Terminal.Gui.html)
```
dodexahedron, line 38: The markdown is still broken.

Format for a URL is `[text you want](url)`

`[v1 documentation here](This is the v2 API documentation. For v1 go here: https://gui-cs.github.io/Terminal.Gui/api/Terminal.Gui.html)` is invalid markdown

Issue #3554, opened 2024-06-20
Multiple views with the same `HotKey` should be supported


tig:
Today, the way HotKey support works is the first view that added the key binding gets it's `Command.Hotkey` command invoked and then processing stops. This means if a layout has views with `_Test`, only the first will get focus when the user presses `Alt-T`.

It would be nicer if each press of the HotKey cycled between all the views in the visible hierarchy that have bound to that hotkey.

dodexahedron:
Agreed.

That's a more natural and typical behavior for things like that.

If they are referenced in a central place, it can be done quite easily.

Since they're delegates, you can just "add" or "subtract" to the delegate (that's also why that works on events - it's exactly the same mechanism.

Doing that then means anything that has subscribed to the delegate will automatically receive the callback, and makes things like cancellation more natural to support, as well.

PR #3634, opened 2024-08-02
Preps for and partially Fixes #2491 - Refactors `Toplevel` and `Application` Navigation

### LINKED ISSUE #2491: Get rid of `Toplevel` - Introduce `Runnable` and `Overlapped` instead

tig:
This is another PR paving the way for fixing 

- #2491

It is a fork of the branch in 

- #3627 
 
at the point where I had done a ton of refactoring, organization, code-clean up, and introduction of primitives required for #2491. 

In order to prepare for addressing #2491 I moved a lot of code around here. In addition, I enabled nullability in big places, and re-organized `View` as well as `Application` to make more understandable and maintainable. As a result, this PR is BIG. 

I want to get this merged because finishing #3627 is going to take a ton of work and I've been coding too much and need to take a break. At the same time, I don't want all the work I did to sit latent, especially since it moves a bunch of code around and will cause merge hell if not integrated quickly. 

At the point where this PR's branch starts all unit tests pass and the new code paths are only active in these situations:

- `View.TabStop` is set to something other than `TabBehavior.TabStop`
- `View.Arrangment` is set to `ViewArrangement.Overlapped`

## Fixes

- Refactors and cleans up existing `Application` and `Toplevel` code without removing existing `Toplevel` or `Overlapped` functionality.
- Makes `Application`, `ConsoleDriver`, and `Toplevel` `#nullable enable` 
- Documents the hell out of existing code and adds more, lower-level, unit tests
- Creates `ViewArrangement.Overlapped` and modifies navigation code to support overlapped views that have this flag set.
- Creates `TabBehavior` enum and changes `View.TabStop` to be of this type. 
- Fixes a bunch of issues in `View.Navigation` without breaking existing functionality.
- Modernizes navigation keys with CM: 
  - Next/Prev TabStop: `Tab` and `Tab.WithShift`
  - Next/Prev TabGroup `F6` and `F6.WithShift` (matches Windows, should work on all platforms)
- Partially Fixes #2491

## Proposed Changes/Todos

- [x] Refactor Application into smaller files
- [x] Make Application use #nullable enable
- [x] Make `ConsoleDriver` use #nullable enable
- [x] Review existing Overlapped and Toplevel unit tests. Add more as a way of learning.
- [x] Moved view navigation out of `Toplevel` and into `Application` (via `ApplicationNavigation` static class)
- [x] Moved `OverlappedTop` navigation out of `Toplevel` and into `ApplicationOverlapped` static class
- [x] Make `ViewExperiments` be a good scenario for exercising all of the above and other stuff for #2491. 

## For #3627 and beyond...

- [ ] Discovered serious issues with how HasFocus, OnEnter/OnLeave, etc... …work in some edge cases. This will require re-visiting the design at a deep level and fixing some long-standing but ignored issues such as how OnEnter/OnLeave don't follow proper cancelation design. Also, there's a need for keeping track of the old focus state of a tree of subviews when that tree loses focus; FocusDireciton is a hack that causes tons of confusion. See https://github.com/tig/Terminal.Gui/blob/8e70e2ae8faafab7cb8305ec6dbbd552c7bf3a43/docfx/docs/navigation.md
- [ ] Figure out how `ViewArrangement.Overlapped` will work. Use `TabIndexes` and all the `View.NextView` focus stuff for navigation (instead of the code in `OverlappedMoveNext`). Use the `Subview` ordering (for just the subviews with ViewArrangement.Overlapped` set) to manage the z-order.
- [ ] Attempt to implement `IRunnable` and replace `Toplevel` with that
- [ ] ...

## Pull Request checklist:


dodexahedron:
Having a look in just a few minutes.

BDisp:
There is a strange behavior with the `ContextMenu`. Open the ContextMenus scenario and right click on the first `TextField`. Then click on another `TextField` and click again on the first one and the cursor isn't showing. If you double click the text will be selected but only clicking doesn't set the cursor on the first `TextField`. But the cursor is always showing on the other `TextField` you clicked.

tig:
> There is a strange behavior with the `ContextMenu`. Open the ContextMenus scenario and right click on the first `TextField`. Then click on another `TextField` and click again on the first one and the cursor isn't showing. If you double click the text will be selected but only clicking doesn't set the cursor on the first `TextField`. But the cursor is always showing on the other `TextField` you clicked.

This repros on v2_develop.

tig:
47wHNuN 1

Note a few things here:

1) Any View can now work "Overlapped". Simply set `View. Arrangement |= ViewArrangement.Overlapped`. The implementation of this is a bit of hack right now, but it illustrates the functionality.
2) `Arrow keys`, `Tab`, and `Shift-Tab` only nav within a TabGroup (a superview with `TabStop == TabBehavior.TabGroup`)
3) `F6` and `Shift-F6` nav between TabGroups (even if the views are tiled). With `ViewArrangement.Overlapped` this *partially* works in this PR.

tig:
One other thing:

- `HotKey`s work across the entire app. If a HotKey is visible, pressing it should invoke it. By default (unless a view overrides the behavior) focus will change to that view. 
- Without https://github.com/gui-cs/Terminal.Gui/issues/3554, this is not perfect, but still nice.

tig:
More:

- `FrameView` now has `TabStop = TabBehavior.TabGroup` and a single border.
- `Window` now has `TabStop = TabBehavior.TabGroup` AND `Arrangement = ViewArrangement.Movable | ViewArrangement.Overlapped` and a double border.

Other than having different `DefaultBorderStyle` they will be the same when we're done with all this.

BDisp (approved):

BDisp:
`TextView` is moving the cursor on mouse moving. This is happening when opening the `Editor` scenario by double click or by pressing the `Key.Enter`. The cursor only should moved on a mouse left/right click or by the keyboard. This isn't happening on the `v2_develop` branch where sometimes may happens when the scenario is opened by double click and moving immediately the mouse.

BDisp:
The `View Experiments` scenario should also move through the tab groups by pressing the F6. But it isn't navigating between the `AdornmentsEditor` and the `Test Frame`.

Terminal.Gui/Views/Shortcut.cs
```diff
@@ -626,10 +626,17 @@ private void UpdateKeyBinding ()
             // Disable the command view key bindings
             CommandView.KeyBindings.Remove (Key);
             CommandView.KeyBindings.Remove (CommandView.HotKey);
-            KeyBindings.Remove (Key);
-            KeyBindings.Add (Key, KeyBindingScope | KeyBindingScope.HotKey, Command.Accept);
 
-            //KeyBindings.Add (Key, KeyBindingScope.HotKey, Command.Accept);
+            if (KeyBindingScope.FastHasFlags (KeyBindingScope.Application))
```
BDisp, line 630: The `UpdateKeyBinding` method isn't updating nothing because in practice it's adding only the new `Key` value but doesn't remove the previous `Key` value, ending up with a shortcut with more than one key if their value are updated.

BDisp (changes requested):
I think that a `Shortcut` which now the `StatusBar` is using, must only have only a `Key ` for each subview.

Terminal.Gui/Views/Shortcut.cs
BDisp: This unit tests will fail:

```cs
    [Fact]
    [AutoInitShutdown]
    public void Changing_Shortcut_Key_Removes_Previous_Key ()
    {
        var newActionCount = 0;

        var statusBar = new StatusBar (
                                       new []
                                       {
                                           new Shortcut (Key.N.WithCtrl, "New", () => newActionCount++)
                                       });
        var top = new Toplevel ();
        top.Add (statusBar);
        var rs = Application.Begin (top);

        Assert.Equal (0, newActionCount);
        Assert.True (Application.OnKeyDown (Key.N.WithCtrl));
        Assert.False (Application.OnKeyDown (Key.W.WithCtrl));
        Assert.Equal (1, newActionCount);

        ((Shortcut)statusBar.Subviews [0]).Key = Key.W.WithCtrl;
        Assert.False (Application.OnKeyDown (Key.N.WithCtrl));
        Assert.True (Application.OnKeyDown (Key.W.WithCtrl));
        Assert.Equal (2, newActionCount);

        Application.End (rs);
        top.Dispose ();
    }

```

tig:
> The `View Experiments` scenario should also move through the tab groups by pressing the F6. But it isn't navigating between the `AdornmentsEditor` and the `Test Frame`.

Yep. This is why I stopped and forked so I can take the time to make it right in the other PR. 

This does not effect use cases without nested TabGroups.

tig:
> I think that a `Shortcut` which now the `StatusBar` is using, must only have only a `Key ` for each subview.

Not sure what you mean. Regardless can you please open a separate issue?

BDisp:
> Not sure what you mean. Regardless can you please open a separate issue?

I wrote a unit tests with a status bar above that reproduces this.

tig:
> > Not sure what you mean. Regardless can you please open a separate issue?
> 
> I wrote a unit tests with a status bar above that reproduces this.

Great catch! Thank you!

Here's a better, more focused unit test based on the one you provided. `Current` is now internal which lets us have tests like this that don't require `AutoInitShutdown`/`Begin`. The nice thing about this is that if a test like this fails it won't cause others to fail. Where if any `AutoInitShutdown` test fails, it's likely to corrupt other tests making diagnosis difficult. 

```cs

    [Fact]
    public void Changing_Key_Removes_Previous ()
    {
        var newActionCount = 0;

        Shortcut shortcut = new Shortcut (Key.N.WithCtrl, "New", () => newActionCount++);
        Application.Current = new Toplevel ();
        Application.Current.Add (shortcut);

        Assert.Equal (0, newActionCount);
        Assert.True (Application.OnKeyDown (Key.N.WithCtrl));
        Assert.False (Application.OnKeyDown (Key.W.WithCtrl));
        Assert.Equal (1, newActionCount);

        shortcut.Key = Key.W.WithCtrl;
        Assert.False (Application.OnKeyDown (Key.N.WithCtrl));
        Assert.True (Application.OnKeyDown (Key.W.WithCtrl));
        Assert.Equal (2, newActionCount);

        Application.Current.Dispose ();
    }
```

BDisp (approved):

tig:
> > > Not sure what you mean. Regardless can you please open a separate issue?
> > 
> > 
> > I wrote a unit tests with a status bar above that reproduces this.
> 
> Great catch! Thank you!
> 
> Here's a better, more focused unit test based on the one you provided. `Current` is now internal which lets us have tests like this that don't require `AutoInitShutdown`/`Begin`. The nice thing about this is that if a test like this fails it won't cause others to fail. Where if any `AutoInitShutdown` test fails, it's likely to corrupt other tests making diagnosis difficult.
> 
> ```cs
>     [Fact]
>     public void Changing_Key_Removes_Previous ()
>     {
>         var newActionCount = 0;
> 
>         Shortcut shortcut = new Shortcut (Key.N.WithCtrl, "New", () => newActionCount++);
>         Application.Current = new Toplevel ();
>         Application.Current.Add (shortcut);
> 
>         Assert.Equal (0, newActionCount);
>         Assert.True (Application.OnKeyDown (Key.N.WithCtrl));
>         Assert.False (Application.OnKeyDown (Key.W.WithCtrl));
>         Assert.Equal (1, newActionCount);
> 
>         shortcut.Key = Key.W.WithCtrl;
>         Assert.False (Application.OnKeyDown (Key.N.WithCtrl));
>         Assert.True (Application.OnKeyDown (Key.W.WithCtrl));
>         Assert.Equal (2, newActionCount);
> 
>         Application.Current.Dispose ();
>     }
> ```

An even better version of the test:

```cs
   [Fact]
   public void Key_Changing_Removes_Previous ()
   {
       Shortcut shortcut = new Shortcut ();

       shortcut.Key = Key.A;
       Assert.Contains (Key.A, shortcut.KeyBindings.Bindings.Keys);

       shortcut.Key = Key.B;
       Assert.DoesNotContain (Key.A, shortcut.KeyBindings.Bindings.Keys);
       Assert.Contains (Key.B, shortcut.KeyBindings.Bindings.Keys);
   }
```

BDisp:
@tig after the creation of the `CheckStyle` enum I already saw some bug on getting the right bool value for bool properties, like the following:

```cs
IsTopLevel = _frmMenuDetails.CkbIsTopLevel?.State == CheckState.UnChecked,
HasSubMenu = _frmMenuDetails.CkbSubMenu?.State == CheckState.UnChecked,

```
As you can see if any of the properties are `false` they will be set incorrectly as `true` because the `State` will always be equal to `CheckState.UnChecked`.

tig:
> @tig after the creation of the `CheckStyle` enum I already saw some bug on getting the right bool value for bool properties, like the following:
> 
> 
> 
> ```cs
> 
> IsTopLevel = _frmMenuDetails.CkbIsTopLevel?.State == CheckState.UnChecked,
> 
> HasSubMenu = _frmMenuDetails.CkbSubMenu?.State == CheckState.UnChecked,
> 
> 
> 
> ```
> 
> As you can see if any of the properties are `false` they will be set incorrectly as `true` because the `State` will always be equal to `CheckState.UnChecked`.

I obviously made some mistakes. 

Not sure why you are commenting here instead of opening a new Issue though.

BDisp:
> I obviously made some mistakes.
> 
> Not sure why you are commenting here instead of opening a new Issue though.

Because their fix make more sense to be done in this branch as it has already fixes for shortcuts. But I'm still investigating why shortcuts menus aren't working on Window.

PR #3638, opened 2024-08-04
Fixes #3109. AOT support with .Net 8.

BDisp:
## Fixes

- Fixes #3109

## Proposed Changes/Todos

- [x] Add a native AOT project.
- [x] Adapt Text.Json to work properly.
- [x] Add in AutoInitShutdownAttribute attribute the possibility to have a different configuration in the config.json and with the unit tests. This avoid to force the config.json having the same configuration as the unit test.
- [x] Fix bug with the color formatting which throw an exception for the Red color.

## Pull Request checklist:

- [x] I've named my PR in the form of "Fixes #issue. Terse description."
- [x] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [x] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [x] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [x] My changes generate no new warnings
- [x] I have checked my code and corrected any poor grammar or misspellings
- [x] I conducted basic QA to assure all features are working

Issue #3493, opened 2024-05-21
Finalize v2 Default Visual Style 


tig:
I'd like apps built with V2 to be visually distinct from v1. Right now, there are some subtle differences. I'd like this Issue to drive discussion on what we want the default look & feel to be.

Here's my current vision:

### So-far:

- TrueColor is default which makes apps look more vibrant (#2798 still needs to be done)

### What's needed: 

I really hope we can figure out these issues for V2 beta. They will be "Right" and will drive obvious visual distinction:

- #457
- #2381

We will then need to pick great default colorschemes.

### So-far:

- LineCanvas & Adornments enables more effective screen real estate management.
- Default glyphs for Button/Checkbox etc... (still needs work)

### What's needed:

- Decide on default border styles for `Window`, `FrameView`, `MessageBox`, `Dialog`, etc...
- Decide on final Title layout (1-3 rows, left, center?)
- #2814 
- #2537 

### What's needed

- Button alignment in `MessageBox`, `Dialog`, and `Wizard`
- `Menu` border or no-border

## So-Far

- FileDialog is radically different
- Slider, ColorPicker are new

What are other's thoughts?

dodexahedron:
Good goals.

I'll more specifically sit down to think about the specifics of this issue and chime in with suggestions after some of my pipeline clears.

BDisp:
The reason why I prefer only the button shadow is hidden when pressed is because the original location of the button pressed is the same as the button released.

tig:
> The reason why I prefer only the button shadow is hidden when pressed is because the original location of the button pressed is the same as the button released.

I think we have these choices when there's a shadow on a button:

1.  Move nothing. Don't hide shadow.
2.  Hide the shadow., but don't move the button
3.  Hide the shadow, move the button one cell to right.
4.  Hide the shadow, move the button one cell to right and one cell down.

The current design (as of #3750) is 3 above:

![iViR1ra 1](https://github.com/user-attachments/assets/3e639561-c3be-4167-8edd-9ed8ffeb966b)

If I understand you correctly you are saying you prefer 2.

In 

- #2144 (here https://github.com/gui-cs/Terminal.Gui/issues/2144#issuecomment-2131348759)

it was discussed and suggested that another system did 3 instead of 4, which was how I originally had it.

I felt 3 was best because with 4 the button moved way too much. 3 has nice motion but is sublte.

What do others think?

tig:
[...]

PR #3639, opened 2024-08-04
Fixes #3523. `MessageBox` buttons centered by default.

### LINKED ISSUE #3523: v2: Button alignment in MessageBox is no longer centered
2.0.0-pre.1491 MessageBox buttons are aligned to the right. They were previously centred. 

If this is a v2 style change, then perhaps we should have a Query() method signature with a ButtonAlignment parameter
```CSharp
MessageBox.Query(50, 7, "Quit?", "Are you sure you want to quit?", 0, "Yes", "No");
```

MessageBox

tig:
## Fixes

- Fixes #3523 
- Adds `MessageBox.DefaultButtonAlignment` config property
- Defaults are now:

image

### DIFF
--- a/Terminal.Gui/Resources/config.json
+++ b/Terminal.Gui/Resources/config.json
@@ -29,8 +29,9 @@
         "Dialog.DefaultButtonAlignmentModes": "AddSpaceBetweenItems",
         "FrameView.DefaultBorderStyle": "Single",
         "Window.DefaultBorderStyle": "Single",
-        "Dialog.DefaultBorderStyle": "Single",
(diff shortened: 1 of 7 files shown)


tig:
MB:

image

Dialog:

image

I'd like to leave these like this for now and make the final call for v2 in 

- #3493

BDisp (approved):
LGTM. In the #3638 I did some changes in the `AutoInitShutdown` attribute, which can be created a separated `ConfigurationManager` attribute where the unit tests can be rely on custom config avoiding forcing the change on the `config.json` file to be align with the unit tests configuration.

PR #3649, opened 2024-08-06
Fixes #3521 - `DimAuto` equality broken

### LINKED ISSUE #3521: DimAuto equality
**Describe the bug**

Regarding equality of DimAuto.  The following test fails

```csharp
    [Test]
    public void TestEquality()
    {
        var a = new DimAuto
        {
            MaximumContentDim = null,
            MinimumContentDim = 1,
            Style = DimAutoStyle.Auto
        };
        var b = new DimAuto
        {
            MaximumContentDim = null,
            MinimumContentDim = 1,
            Style = DimAutoStyle.Auto
        };
        ClassicAssert.IsTrue(a.Equals(b));
        ClassicAssert.IsTrue(a.GetHashCode() == b.GetHashCode());
    }
```

Notice also that even the basic case fails:

```csharp
    [Test]
    public void TestEquality()
    {
        var a = Dim.Auto();
        var b = Dim.Auto();
        ClassicAssert.IsTrue(a.Equals(b));
        ClassicAssert.IsTrue(a.GetHashCode() == b.GetHashCode());
    }
```

Percent does not behave this way, it supports Equals

```csharp
    [Test]
    public void TestEquality3()
    {
        var a = Dim.Percent(32);
        var b = Dim.Percent(32);
        ClassicAssert.IsTrue(a.Equals(b));
        ClassicAssert.IsTrue(a.GetHashCode() == b.GetHashCode());
    }
```

tig:
## Fixes

- Fixes #3521 
-  
## Proposed Changes/Todos

- [x] Add unit tests proving bad behavior
- [x] Engineer fix

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/View/Layout/DimAuto.cs
+++ b/Terminal.Gui/View/Layout/DimAuto.cs
@@ -17,25 +17,30 @@ namespace Terminal.Gui;
 /// </remarks>
 public class DimAuto : Dim
 {
-    private readonly Dim? _maximumContentDim;
-
-    private readonly Dim? _minimumContentDim;
-
-    private readonly DimAutoStyle _style;
-
-    /// <inheritdoc/>
+    /// <inheritdoc />
     public override bool Equals (object? other)
     {
+        if (ReferenceEquals (this, other))
+        {
+            return true;
+        }
+
         if (other is not DimAuto auto)
         {
             return false;
         }
 
-        return auto.MinimumContentDim == MinimumContentDim && auto.MaximumContentDim == MaximumContentDim && auto.Style == Style;
+        return EqualityComparer<Dim?>.Default.Equals (MinimumContentDim, auto.MinimumContentDim) &&
+               EqualityComparer<Dim?>.Default.Equals (MaximumContentDim, auto.MaximumContentDim) &&
+               Style == auto.Style;
     }
 
-    /// <inheritdoc/>
-    public override int GetHashCode () { return HashCode.Combine (MinimumContentDim, MaximumContentDim, Style); }
+    /// <inheritdoc />
+    public override int GetHashCode ()
+    {
+        return HashCode.Combine (MinimumContentDim, MaximumContentDim, Style);
+    }
+    private readonly Dim? _maximumContentDim;
 
     /// <summary>
     ///     Gets the maximum dimension the View's ContentSize will be fit to. NOT CURRENTLY SUPPORTED.
@@ -48,6 +53,8 @@ public class DimAuto : Dim
         init => _maximumContentDim = value;
     }
 
+    private readonly Dim? _minimumContentDim;
+
     /// <summary>
     ///     Gets the minimum dimension the View's ContentSize will be constrained to.
     /// </summary>
@@ -58,6 +65,8 @@ public class DimAuto : Dim
         get => _minimumContentDim;
         init => _minimumContentDim = value;
     }
+
+    private readonly DimAutoStyle _style;
 
     /// <summary>
     ///     Gets the style of the DimAuto.
--- a/Terminal.Gui/View/Layout/DimAuto.cs
+++ b/Terminal.Gui/View/Layout/DimAuto.cs
@@ -15,29 +15,11 @@ namespace Terminal.Gui;
 ///         methods on the <see cref="Dim"/> class to create <see cref="Dim"/> objects instead.
 ///     </para>
 /// </remarks>
+/// <param name="MaximumContentDim">The maximum dimension the View's ContentSize will be fit to. NOT CURRENTLY SUPPORTED.</param>
+/// <param name="MinimumContentDim">The minimum dimension the View's ContentSize will be constrained to.</param>
+/// <param name="Style">The <see cref="DimAutoStyle"/> of the <see cref="DimAuto"/>.</param>
 public record DimAuto (Dim? MaximumContentDim, Dim? MinimumContentDim, DimAutoStyle Style) : Dim
 {
-    /// <summary>
-    ///     Gets the maximum dimension the View's ContentSize will be fit to. NOT CURRENTLY SUPPORTED.
-    /// </summary>
-
-    // ReSharper disable once ConvertToAutoProperty
-    public required Dim? MaximumContentDim { get; init; } = MaximumContentDim;
-
-    /// <summary>
-    ///     Gets the minimum dimension the View's ContentSize will be constrained to.
-    /// </summary>
-
-    // ReSharper disable once ConvertToAutoProperty
-    public required Dim? MinimumContentDim { get; init; } = MinimumContentDim;
-
-    /// <summary>
-    ///     Gets the style of the DimAuto.
-    /// </summary>
-
-    // ReSharper disable once ConvertToAutoProperty
-    public required DimAutoStyle Style { get; init; } = Style;
-
     /// <inheritdoc/>
     public override string ToString () { return $"Auto({Style},{MinimumContentDim},{MaximumContentDim})"; }
 
--- a/Terminal.Gui/View/Layout/DimAuto.cs
+++ b/Terminal.Gui/View/Layout/DimAuto.cs
@@ -15,31 +15,8 @@ namespace Terminal.Gui;
 ///         methods on the <see cref="Dim"/> class to create <see cref="Dim"/> objects instead.
 ///     </para>
 /// </remarks>
-public class DimAuto : Dim
+public record DimAuto : Dim
 {
-    /// <inheritdoc />
-    public override bool Equals (object? other)
-    {
-        if (ReferenceEquals (this, other))
-        {
-            return true;
-        }
-
-        if (other is not DimAuto auto)
-        {
-            return false;
-        }
-
-        return EqualityComparer<Dim?>.Default.Equals (MinimumContentDim, auto.MinimumContentDim) &&
-               EqualityComparer<Dim?>.Default.Equals (MaximumContentDim, auto.MaximumContentDim) &&
-               Style == auto.Style;
-    }
-
-    /// <inheritdoc />
-    public override int GetHashCode ()
-    {
-        return HashCode.Combine (MinimumContentDim, MaximumContentDim, Style);
-    }
     private readonly Dim? _maximumContentDim;
 
     /// <summary>
--- a/Terminal.Gui/View/Layout/DimAuto.cs
+++ b/Terminal.Gui/View/Layout/DimAuto.cs
@@ -59,6 +59,9 @@ public record DimAuto : Dim
     /// <inheritdoc/>
     public override string ToString () { return $"Auto({Style},{MinimumContentDim},{MaximumContentDim})"; }
 
+    /// <inheritdoc />
+    internal override int GetAnchor (int size) => 0;
+
     internal override int Calculate (int location, int superviewContentSize, View us, Dimension dimension)
     {
         var textSize = 0;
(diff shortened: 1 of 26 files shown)


Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -17,25 +17,30 @@ namespace Terminal.Gui;
...
+            return true;
+        }
+
         if (other is not DimAuto auto)
         {
             return false;
         }
 
-        return auto.MinimumContentDim == MinimumContentDim && auto.MaximumContentDim == MaximumContentDim && auto.Style == Style;
+        return EqualityComparer<Dim?>.Default.Equals (MinimumContentDim, auto.MinimumContentDim) &&
```
dodexahedron, line 33: This method of getting equality is non-ideal and, when I turn these into record structs, it'll be even less ideal.

You can actually achieve the whole equality thing by just making them all `record` instead of `class` for now. That comes with implicitly defined memberwise equality for all fields of the type, or you can override it if you _need_ to... But the implicit equality already works like the one you wrote, while also bringing along all the other goodies records bring along for free, including decompose, non-destructive mutation (the `with` operator), and built-in support of the `IEqualityOperators<TSelf,TSelf,bool>` interface.

I have like 3 open branches in which I've turned the Dim types at least into `record` and one or MAYBE two I turned them into `record struct`, but they were before the work you did to make Dim better than it was got started, so they're all pretty much useless now.

But the change to just plain `record` is really straightforward.

And if you switch it to using a primary constructor at the same time (which is optional), it will also be immutable by default, which it already is since the fields are `readonly`.

Basically, like.... any code in this file that has to do with equality just gets to go away, all by writing `record` instead of class.

`Dim` also has to be a record for that, because records are not allowed to inherit from classes and vice versa...which is weird if you ask me, because they're just classes with extra goodies written by Roslyn, and even look like normal classes to other languages anyway. 🤷‍♂️

Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -17,25 +17,30 @@ namespace Terminal.Gui;
...
-
-    private readonly Dim? _minimumContentDim;
-
-    private readonly DimAutoStyle _style;
-
-    /// <inheritdoc/>
+    /// <inheritdoc />
     public override bool Equals (object? other)
     {
+        if (ReferenceEquals (this, other))
```
dodexahedron, line 23: This method in its entirety will go away if the suggestion in the long comment in this review is taken.

Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -17,25 +17,30 @@ namespace Terminal.Gui;
...
-        return auto.MinimumContentDim == MinimumContentDim && auto.MaximumContentDim == MaximumContentDim && auto.Style == Style;
+        return EqualityComparer<Dim?>.Default.Equals (MinimumContentDim, auto.MinimumContentDim) &&
+               EqualityComparer<Dim?>.Default.Equals (MaximumContentDim, auto.MaximumContentDim) &&
+               Style == auto.Style;
     }
 
-    /// <inheritdoc/>
-    public override int GetHashCode () { return HashCode.Combine (MinimumContentDim, MaximumContentDim, Style); }
+    /// <inheritdoc />
+    public override int GetHashCode ()
```
dodexahedron, line 39: This method in its entirety will go away if the suggestion in the long comment in this review is taken.

Terminal.Gui/View/Layout/DimAuto.cs
dodexahedron: Oh also... Being a record means you do not need to have tests for equality, since it's handled in the framework and not in our explicit code any more.

If we, at some point, were to override equality to not mean value equality, then we'd need our own tests again. But these types are values and use value equality, so that shouldn't be an issue.

dodexahedron (changes requested):
Suggestion for a simpler, cleaner, and more robust implementation in the long comment in this review.

dodexahedron:
Will take a look shortly, when I'm at my PC again.

dodexahedron:
> ```cs
>     [Fact]
>     public void DimFunc_Equal ()
>     {
>         Func<int> f1 = () => 0;
>         Func<int> f2 = () => 0;
> 
>         Dim dim1 = Func (f1);
>         Dim dim2 = Func (f2);
>         Assert.Equal (dim1, dim2);
> 
>         f2 = () => 1;
>         dim2 = Func (f2);
>         Assert.NotEqual (dim1, dim2);
>     }
> ```
> 
> After switching to `record` this test is failing.
> 
> I think it was bogus before. `f1` and `f2` are different functions and thus `dim1` and `dim2` are not the same. Right?

💯 

> 
> ```cs
>     [Fact]
>     public void DimFunc_Equal ()
>     {
>         Func<int> f1 = () => 0;
>         Func<int> f2 = () => 0;
> 
>         Dim dim1 = Func (f1);
>         Dim dim2 = Func (f1);
>         Assert.Equal (dim1, dim2);
> 
>         dim2 = Func (f2);
>         Assert.NotEqual (dim1, dim2);
> 
>         f2 = () => 1;
>         dim2 = Func (f2);
>         Assert.NotEqual (dim1, dim2);
>     }
> ```
> 
> I think this test is correct. Correct?

Looks better, yes.

Just for reference, here is how equality of delegates works:

>Two delegates of the same type with the same targets, methods, and invocation lists are considered equal.
> 
>If the two delegates are not of the same type, they are not considered equal.
> 
>The methods and targets are compared for equality as follows:
> 
> * If the two methods being compared are both static and are the same method on the same class, the methods are considered equal and the targets are also considered equal.
> 
> * If the two methods being compared are instance methods and are the same method on the same object, the methods are considered equal and the targets are also considered equal.
> 
> * Otherwise, the methods are not considered to be equal and the targets are also not considered to be equal.
> 
>Two invocation lists are considered identical if they have the same order and the corresponding elements from the two lists represent the same method and target.


Aside from that, I think I need to pull the branch and take a look at a couple of other things about the code.

One thing is that the use of `new` all over the place in them looks odd to me, in the context of the diffs in github. But that doesn't give me the whole picture of course, so I want to see it in-situ.

dodexahedron:
Also, the whole delegate equality thing got me thinking...

I want to be sure your intent is aligned with what this code will do, or at least allow.

Remember that all delegates you can write in dotnet are MulticastDelegate, which is a mutable reference type which contains an invocation _list_.

Take this simple program for example, which showcases both delegate fun and record behavior (immutable value by default): https://dotnetfiddle.net/wEmXkg

dodexahedron (commented):
Added a couple of minor comments about explicit properties in records not being needed when they're already defined in the primary constructor.

Terminal.Gui/View/Layout/DimFill.cs
```diff
@@ -8,19 +8,13 @@ namespace Terminal.Gui;
...
-    public override bool Equals (object? other) { return other is DimFill fill && fill.Margin == Margin; }
-
-    /// <inheritdoc/>
-    public override int GetHashCode () { return Margin.GetHashCode (); }
-
     /// <summary>
     ///     Gets the margin to not fill.
     /// </summary>
-    public int Margin { get; } = margin;
+    public int Margin { get; } = Margin;
```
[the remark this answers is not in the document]
dodexahedron: For the _record_ (woo puns!), this is a perfectly valid and complete record definition:

```csharp
// Super basic but has two init-only properties as shown in the constructor.
public record SomeRecord(string Name, SomeType Value);

// Same record type, but declares the interface it naturally supports for free:
public record SomeRecord(string Name, SomeType Value) : IEqualityOperators<SomeRecord<T>, SomeRecord<T>, bool>;
```

And here it is as generic, partial (so you could perhaps put the non-structural code, if any, in a separate file named to nest under it), and with a type parameter constraint that record types implicitly support, so long as you simply declare it, as noted above:

```csharp
public partial record SomeRecord<T> (string Name, T Value) : IEqualityOperators<SomeRecord<T>, SomeRecord<T>, bool> where T : IEqualityOperators<T, T, bool>;
```

You get all the goodies of a record with all three of these. You get like 50 lines of code for the price of one with the third one, and only slightly less with the other two.

Terminal.Gui/View/Layout/DimPercent.cs
```diff
@@ -8,23 +8,17 @@ namespace Terminal.Gui;
...
-    public override bool Equals (object? other) { return other is DimPercent f && f.Percent == Percent && f.Mode == Mode; }
-
-    /// <inheritdoc/>
-    public override int GetHashCode () { return Percent.GetHashCode (); }
-
     /// <summary>
     ///     Gets the percentage.
     /// </summary>
-    public new int Percent { get; } = percent;
+    public new int Percent { get; } = Percent;
```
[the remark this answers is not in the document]
tig: It's needed in this case because the base `Dim` has a `static` method named `Percent`.

Terminal.Gui/View/Layout/DimFill.cs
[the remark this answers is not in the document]
dodexahedron: Also note, since you CANNOT override the equality operators nor the object.Equals method (and therefore _shouldn't_ override the typed `IEquatable<T>.Equals` method either unless you simply pass it through to the generated Equals), it is almost always unnecessary to test equality for record types.

They come with deconstruction built-in, too, like Tuples, so they can also reduce the number of explicit locals that are copied, in consumers of them, too, if you use those, since they provide `out` references to the properties of the record.

dodexahedron:
Also, it's been an absolute hell of a couple of weeks...

Things are clearing up, though, so I'm gonna have more time and desire to work on TG and/or TG-connected stuff over the next few days, finally. 🥳

tig:
> Also, it's been an absolute hell of a couple of weeks...
> 
> Things are clearing up, though, so I'm gonna have more time and desire to work on TG and/or TG-connected stuff over the next few days, finally. 🥳

Sorry to hear. Life can be hard sometimes.

Thanks for looking at this. Learned a bunch.

dodexahedron:
> > Also, it's been an absolute hell of a couple of weeks...
> > Things are clearing up, though, so I'm gonna have more time and desire to work on TG and/or TG-connected stuff over the next few days, finally. 🥳
> 
> Sorry to hear. Life can be hard sometimes.
> 
> Thanks for looking at this. Learned a bunch.

Excellent! That was the hope. :)

It's an under-utilized feature, but a really handy one (records, I mean). They can save you lots of effort and give some consistency to implementation that you no longer have to do yourself. I love 'em when they can be used.

Since all record types have value semantics out of the box (unless you manually do bad things to break it), record classes, in particular, have the neat property of enabling treatment of them as value types, but they're still reference types, which _can_ be useful in certain scenarios.

These can and should at some point be turned into `readonly record struct` types, but we can do that in another pass. The use of them still looks basically the same at the point of consumption, but they're actually structs at that point.

`readonly` is necessary for those to make them immutable. `record struct` without `readonly` is mutable by default (not sure why they went that way, but whatev).

The reason I wouldn't say just do it now is because it would be smart to add `in` and `ref [readonly]` for them, in various method parameter uses, for best efficiency. While that _could_ just be blanket applied, a more precise touch is best (though not mandatory).

dodexahedron:
Oh and more record use info:

If you want a property declared in the primary constructor to have a default value, like a field initializer, the standard optional parameter syntax is how you do it. It gets you both initialization with no need for callers to supply all parameters, as well as explicit construction without the initialization happening, all in the same line.

dodexahedron:
And you can always (and it's a good idea to) add the `IEqualityOperators<TheRecordType,TheRecordType,bool>` interface to them, because they all already support it. Another thing I'm not sure why they didn't just make default behavior in the generated code.

ReSharper will suggest it for you, so you can add it with a click, rather than writing it yourself.

PR #3653, opened 2024-08-08
Fixes #3652. Setting Menus causes unexpected Exception.

### LINKED ISSUE #3644: Some uses of the `CheckStyle` enum are buggy
Via @bdisp:      @tig after the creation of the `CheckStyle` enum I already saw some bug on getting the right bool value for bool properties, like the following:

```cs
IsTopLevel = _frmMenuDetails.CkbIsTopLevel?.State == CheckState.UnChecked,
HasSubMenu = _frmMenuDetails.CkbSubMenu?.State == CheckState.UnChecked,

```
As you can see if any of the properties are `false` they will be set incorrectly as `true` because the `State` will always be equal to `CheckState.UnChecked`.

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/3634#issuecomment-2267959850_

### LINKED ISSUE #3652: Setting `Menus` causes unexpected Exception
**Describe the bug**

MenuBar.Menus magic property behaves inconsistently when set multiple times.

The setter should replace existing values with the new values. 

Regarding the new Exception that is throw, it is problematic for designer where user might easily create multiple menu items with the same name or shortcut key.  

For this use case we need a setting which suppress the Exception.

Test
```csharp
    [Fact]
    public void Test_SetMenus()
    {
        var mb = new MenuBar ();

        var i1 = new MenuBarItem ("heey", "fff", () => { }, () => { return true; })
        {
            HotKey = new Rune('I')
        };

        mb.Menus = new MenuBarItem [] { i1 };
        mb.Menus = new MenuBarItem [] { i1 };
    }
```
Throws
```csharp
  Message: 
System.InvalidOperationException : A key binding for i exists (KeyBinding { Commands = Terminal.Gui.Command[], Scope = Focused, BoundView = , Context = 0 }).

  Stack Trace: 
KeyBindings.Add(Key key, KeyBinding binding, View boundViewForAppScope) line 32
MenuBar.set_Menus(MenuBarItem[] value) line 178
MenuTests.Test_SetMenus() line 29
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

BDisp:
## Fixes

- Fixes #3652, Fixes #3644

## Proposed Changes/Todos

- [x] Rename Shortcut to ShortcutKey and change type to Key
- [x] Change HotKey type to Key
- [x] Improving add and remove menu items dynamically
- [x] Allow run an action in IsTopLevel menu bar item through HotKey

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Input/Key.cs
+++ b/Terminal.Gui/Input/Key.cs
@@ -1,5 +1,6 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.Text.Json.Serialization;
 
 namespace Terminal.Gui;
 
@@ -975,6 +976,7 @@ public class Key : EventArgs, IEquatable<Key>
     private static Rune _separator = new ('+');
 
     /// <summary>Sets or gets the shortcut delimiter separator. The default is "+".</summary>
+    [SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
     public static Rune Separator
     {
         get => _separator;
--- a/Terminal.Gui/Input/Key.cs
+++ b/Terminal.Gui/Input/Key.cs
@@ -975,7 +975,7 @@ public class Key : EventArgs, IEquatable<Key>
 
     private static Rune _separator = new ('+');
 
-    /// <summary>Sets or gets the shortcut delimiter separator. The default is "+".</summary>
+    /// <summary>Gets or sets the separator character used when parsing and printing Keys. E.g. Ctrl+A. The default is '+'.</summary>
     [SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
     public static Rune Separator
     {
--- a/Terminal.Gui/Input/Key.cs
+++ b/Terminal.Gui/Input/Key.cs
@@ -450,7 +450,7 @@ public class Key : EventArgs, IEquatable<Key>
 
     /// <summary>Pretty prints the KeyEvent</summary>
     /// <returns></returns>
-    public override string ToString () { return ToString (KeyCode, (Rune)'+'); }
+    public override string ToString () { return ToString (KeyCode, ShortcutDelimiter); }
 
     private static string GetKeyString (KeyCode key)
     {
@@ -483,7 +483,7 @@ public class Key : EventArgs, IEquatable<Key>
     ///     The formatted string. If the key is a printable character, it will be returned as a string. Otherwise, the key
     ///     name will be returned.
     /// </returns>
-    public static string ToString (KeyCode key) { return ToString (key, (Rune)'+'); }
+    public static string ToString (KeyCode key) { return ToString (key, ShortcutDelimiter); }
 
     /// <summary>Formats a <see cref="KeyCode"/> as a string.</summary>
     /// <param name="key">The key to format.</param>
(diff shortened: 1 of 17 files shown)


Terminal.Gui/Input/Key.cs
```diff
@@ -971,4 +971,19 @@ out parsedInt
     public static Key F24 => new (KeyCode.F24);
 
     #endregion
+
+    private static Rune _shortcutDelimiter = new ('+');
+
+    /// <summary>Sets or gets the shortcut delimiter separator. The default is "+".</summary>
+    public static Rune ShortcutDelimiter
```
tig, line 978: We should rename this to "Separator". With this PR, this is now used for all `Key` usages and should be more generic.

We should also make this a CM property. It needs to be configurable.

On Windows the standard is +.

On Mac the standard is -.

tig (changes requested):
Really great work!

Terminal.Gui/Input/Key.cs
BDisp: > We should rename this to "Separator". With this PR, this is now used for all `Key` usages and should be more generic.
> 
Done.

> We should also make this a CM property. It needs to be configurable.
> 
Done.

> On Windows the standard is +.
> 
Done.

> On Mac the standard is -.

I think this is irrelevant and I have no idea how to configure this on the `config.json` to achieve this.

Terminal.Gui/Input/Key.cs
```diff
@@ -972,17 +973,18 @@ out parsedInt
 
     #endregion
 
-    private static Rune _shortcutDelimiter = new ('+');
+    private static Rune _separator = new ('+');
 
     /// <summary>Sets or gets the shortcut delimiter separator. The default is "+".</summary>
```
tig, line 978: "Gets or sets the separator character used when parsing and printing Keys. E.g. <c>Ctrl+A</c>. The default is '+'".

Terminal.Gui/Input/Key.cs
```diff
@@ -450,7 +451,7 @@ public override bool Equals (object obj)
 
     /// <summary>Pretty prints the KeyEvent</summary>
```
tig, line 452: KeyEvent->Key

Terminal.Gui/Input/Key.cs
```diff
@@ -972,17 +973,18 @@ out parsedInt
...
     {
-        get => _shortcutDelimiter;
+        get => _separator;
         set
         {
-            if (_shortcutDelimiter != value)
+            if (_separator != value)
             {
-                _shortcutDelimiter = value == default (Rune) ? new ('+') : value;
+                _separator = value == default (Rune) ? new ('+') : value;
```
tig, line 987: I don't think this test for default is needed. _separator is set to "+" above.

tig (changes requested):
Some minor tweaks please.

Terminal.Gui/Input/Key.cs
BDisp: I added a unit test that proves it's needed to test for `Rune` default, otherwise `Ctrl+A` will only print `Ctrl`.

Issue #3675, opened 2024-08-20
Need more cases of `ObjectDisposedException` 


tig:
I just spent several hours fighting bugs in `Menubar` and `ContextMenu` only to discover that they are passing around views (and themsevles) that have been disposed.

We have a lot of this going on and we need to squash it.

This issue is to track adding a bunch of calls like this in key places:

```cs
#if DEBUG_IDISPOSABLE
            if (WasDisposed)
            {
                throw new ObjectDisposedException (GetType ().FullName);
            }
#endif
```

- `Visible` get/set
- `Enabled` get/st
- `HasFocus` get/set
- Mouse event handling

dodexahedron:
Just a reminder:

Don't forget to use the handy-dandy [ObjectDisposedException.ThrowIf](https://learn.microsoft.com/en-us/dotnet/api/system.objectdisposedexception.throwif?view=net-8.0) helper instead of manually constructing and throwing, too!

Better static analysis and optimization, with a little less code, so a win all around!

The code in the initial post would become:

```csharp
ObjectDisposedException.ThrowIf(WasDisposed,typeof(whateverClassThisIsIn));
```

Don't use the overload that takes an object instance, as it uses reflection to get the type name (which is all the second argument is for). It calls GetType() at run-time on whatever you passed. `typeof(SomeType)` is a compile-time constant, which avoids that and is more efficient, to boot. 🥳

dodexahedron:
Also, it's typically not thrown from property accessor methods, unless the accessor itself accesses a ***different*** disposable object that has been disposed, and it is supposed to leave the throwing of that up to _that_ object.

Methods, however, are expected to throw, ideally with a more detailed InnerException if possible/relevant.

dodexahedron:
A couple of heads ups before you write a whole bunch of code, just in case I didn't beat this particular horse enough or properly last time we did exception work. 😄

dodexahedron:
Oooo

And for methods that return IDisposables, and any types that are themselves IDisposable, it's friendly to include the `[MustDisposeResource(bool)]` attribute, with true if the caller is expected to dispose it or false if it doesn't matter who disposes it.

tig:
Here's the issue I found that led to this exercise:

https://github.com/gui-cs/Terminal.Gui/issues/3678

PR #3681, opened 2024-08-21
Fixes #3678. ContextMenu accesses disposed MenuBar.

### LINKED ISSUE #3678: `ContextMenu` accesses disposed `MenuBar`
To repo, change `Application.UngrabMouse()`:

```cs
    /// <summary>Releases the mouse grab, so mouse events will be routed to the view on which the mouse is.</summary>
    public static void UngrabMouse ()
    {
        if (MouseGrabView is null)
        {
            return;
        }

#if DEBUG_IDISPOSABLE
        ObjectDisposedException.ThrowIf (MouseGrabView.WasDisposed, typeof (View));
#endif

        if (!OnUnGrabbingMouse (MouseGrabView))
        {
            View view = MouseGrabView;
            MouseGrabView = null;
            OnUnGrabbedMouse (view);
        }
    }
```

Run unit tests.

My suggestion is to refactor it to create `_menuBar` as a `static` and use `_menuBar.Visible` and/or `_menuBar.Enabled` to manage.

BDisp:
## Fixes

- Fixes #3678

## Proposed Changes/Todos

- [x] Call UngrabMouse before disposing _menubar

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Application/Application.Mouse.cs
+++ b/Terminal.Gui/Application/Application.Mouse.cs
@@ -167,8 +167,6 @@ public static partial class Application // Mouse handling
                 View = view ?? MouseGrabView
             };
 
-            // INTENT: Just checking: This is operating on a copy of ViewPort and throwing it away.
-            // Was assignment also intended?
             if ((MouseGrabView.Viewport with { Location = Point.Empty }).Contains (viewRelativeMouseEvent.Position) is false)
             {
                 // The mouse has moved outside the bounds of the view that grabbed the mouse
@@ -182,16 +180,14 @@ public static partial class Application // Mouse handling
             }
         }
 
+        // We can combine this into the switch expression to reduce cognitive complexity even more and likely
+        // avoid one or two of these checks in the process, as well.
         WantContinuousButtonPressedView = view switch
                                           {
                                               { WantContinuousButtonPressed: true } => view,
                                               _                                     => null
                                           };
 
-        // NOTE: These two nested ifs were already an AND condition, so I at least merged them in this pass.
-        // But also, we have type-checked view already, right above.
-        // We can combine this into the switch expression to reduce cognitive complexity even more and likely
-        // avoid one or two of these checks in the process, as well.
         if (view is not Adornment
          && (view is null || view == ApplicationOverlapped.OverlappedTop)
          && Current is { Modal: false }
@@ -209,7 +205,8 @@ public static partial class Application // Mouse handling
             }
         }
 
-        // NOTE: This statement is also likely mergeable with the above stuff, too.
+        // May be null before the prior condition or the condition may set it as null.
+        // So, the checking must be outside the prior condition.
         if (view is null)
         {
             return;
--- a/Terminal.Gui/Application/Application.Mouse.cs
+++ b/Terminal.Gui/Application/Application.Mouse.cs
@@ -57,7 +57,7 @@ public static partial class Application // Mouse handling
         }
 
 #if DEBUG_IDISPOSABLE
-        ObjectDisposedException.ThrowIf (MouseGrabView.WasDisposed, typeof (View));
+        ObjectDisposedException.ThrowIf (MouseGrabView.WasDisposed, MouseGrabView);
 #endif
 
         if (!OnUnGrabbingMouse (MouseGrabView))
--- a/Terminal.Gui/Application/Application.Mouse.cs
+++ b/Terminal.Gui/Application/Application.Mouse.cs
@@ -56,6 +56,10 @@ public static partial class Application // Mouse handling
             return;
         }
 
+#if DEBUG_IDISPOSABLE
+        ObjectDisposedException.ThrowIf (MouseGrabView.WasDisposed, typeof (View));
+#endif
+
         if (!OnUnGrabbingMouse (MouseGrabView))
         {
             View view = MouseGrabView;
--- a/Terminal.Gui/Application/Application.Mouse.cs
+++ b/Terminal.Gui/Application/Application.Mouse.cs
@@ -160,13 +160,13 @@ public static partial class Application // Mouse handling
                 Position = frameLoc,
                 Flags = mouseEvent.Flags,
                 ScreenPosition = mouseEvent.Position,
-                View = MouseGrabView
+                View = view ?? MouseGrabView
             };
 
             if ((MouseGrabView.Viewport with { Location = Point.Empty }).Contains (viewRelativeMouseEvent.Position) is false)
             {
                 // The mouse has moved outside the bounds of the view that grabbed the mouse
-                MouseEnteredView?.NewMouseLeaveEvent (mouseEvent);
+                MouseGrabView?.NewMouseLeaveEvent (mouseEvent);
             }
 
             //System.Diagnostics.Debug.WriteLine ($"{nme.Flags};{nme.X};{nme.Y};{mouseGrabView}");
--- a/Terminal.Gui/Application/Application.Mouse.cs
+++ b/Terminal.Gui/Application/Application.Mouse.cs
@@ -267,7 +267,7 @@ public static partial class Application // Mouse handling
 
         //Debug.WriteLine ($"OnMouseEvent: ({a.MouseEvent.X},{a.MouseEvent.Y}) - {a.MouseEvent.Flags}");
 
-        while (view.NewMouseEvent (me) is not true || MouseGrabView is not { })
+        while (view.NewMouseEvent (me) is not true && MouseGrabView is not { })
         {
             if (view is Adornment adornmentView)
             {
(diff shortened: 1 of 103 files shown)


tig (approved):

BDisp:
@tznind is there any problem the _menuBar being disposing here for the TGD?

BDisp:
> Should be fine. Maybe some of the Undo operations might access a closed sub menu. E.g.
> 
> * open menu
> * add new entry
> * close menu
> * hit ctrl+z

Can you please test if that really happens when ctrl+z is hit? If affirmative I've to found a solution to fix it. `_menuBar` is never added to a superview, but initialized calling `BeginInit/EndInit`. Only the menus are added to superview when they are opened. Maybe subscribing to the `Application.Top.Dispose` event to dispose the `_menuBar` instead of disposing on the `MenuAllClosed` event. But this is only needed after you confirm that this may be an issue. Thanks.

Terminal.Gui/Application/Application.Mouse.cs
```diff
@@ -56,6 +56,10 @@ public static void UngrabMouse ()
             return;
         }
 
+#if DEBUG_IDISPOSABLE
+        ObjectDisposedException.ThrowIf (MouseGrabView.WasDisposed, typeof (View));
```
dodexahedron, line 60: That view should be the object responsible for throwing this exception, not a class using it.

Same exception is fine - just belongs elsewhere.

If an exception is desired here, it should be most likely be InvalidOperationException, but only if the MouseGrabView View didn't already throw ObjectDisposedException anyway. I'd argue that the need for the exception being thrown by this class is itself a bug, though.

Typically, the object that was disposed is the one who is supposed to throw ODE.

Assuming that principle is followed, and a class like this one that _owns_ a disposable object tries to _use_ that object after that object was disposed, this is how it's supposed to be handled:
 * Highest preference/best option, if possible, is to try to avoid the situation entirely by performing a safe check before trying to use the IDisposable object and creating a new one if necessary. That's one of the main uses of a `Disposed` event - letting subscribers holding a reference to that object know it has been disposed, so they can react appropriately, such as by setting their reference to it to `null`, removing it from any collections they may have put it in, or any other implementation details that the IDisposable doesn't know about and can't be expected to know about; OR
 * Next best is to try to recover from the situation, if you catch an ODE that you weren't able to avoid. Usually that just means making a new instance of the object that was disposed, if there's sufficient information to do so. OR
 * If the object trying to use the IDisposable can't do either of the above safely, it should then do one of:
    * Let the ODE bubble up. OR
    * If it caught the ODE because it has clean-up to do before things can safely continue, re-throw it as-is at the end of the catch. OR
    * If additional context is important to add at that point in the code that the stack trace wouldn't already have been sufficient for, wrap the ODE in an InvalidOperationException, which is the type of exception that is expected to be used when the object is in an invalid state (which it is, _because_ of the disposed object, which is part of its state, and thus the operation is invalid for the current state of the object)
 

All objects owning an IDisposable instance or static member field or property should do the above procedure, at minimum, as part of the contract that IDisposable is supposed to convey.

The IDisposable objects themselves are responsible for their own cleanup, part of which _is_ informing others that disposal has happened.\
That's the easiest, most timely, and most efficient way to do the first bullet (avoiding/preventing the situation entirely), and prevents any exceptions from being thrown, which saves a lot of expensive implicit stuff that happens when an exception is thrown, like generating the stack trace and system event logging and everything else that an exception involves.

At that point, if we've implemented our end of the contract properly, any consumer who doesn't avoid the exception can't blame it on us, as it is not our fault - It's _their_ bug for not respecting the contract, just like it is _currently_ our fault for not respecting our own contract.

dodexahedron (commented):
Similar commentary, but with a little more detail, to what I was talking about with @tig in #3675 about ObjectDisposedException.

Terminal.Gui/Application/Application.Mouse.cs
tig: > That view should be the object responsible for throwing this exception, not a class using it.

My bad. I must have mis-read you before, because I thought you were recommending the opposite.

BDisp:
I think I found a better solution than what I first presented. It's only worth preserving the last opened `MenuBar`. In any case, the `ContextMenu` information is always present when the object is created and assigned to the `MenuItems` property. `ContextMenu` just creates a static instance of the `MenuBar` with the information collected from the `MenuItems` property. At least now, if there is a need to access information from the last `MenuBar` of the `ContextMenu` opened, it will be available while the application is running.

Terminal.Gui/Application/Application.Mouse.cs
BDisp: @tig do you want I made any change here?

BDisp: Maybe to this:

`ObjectDisposedException.ThrowIf (MouseGrabView.WasDisposed, MouseGrabView);`

dodexahedron: > > Maybe to this, because it's passing the object that is failing?
> > `ObjectDisposedException.ThrowIf (MouseGrabView.WasDisposed, MouseGrabView);`
> 
> @dodexahedron is this more appropriate?

Not quite. It's not the exception - it's where it's being thrown.

I can pull this down to take an actual look at it in a few minutes.

Is whatever is latest currently pushed to your branch?

dodexahedron: > > That view should be the object responsible for throwing this exception, not a class using it.
> 
> My bad. I must have mis-read you before, because I thought you were recommending the opposite.

I was pretty certain there would be at least one or five reversals. 😅

That's part of why those comments took me a while to write - I had to keep going back and making sure I wasn't reversing things, which was super easy to do and I _did_ do (but therefore fixed before I posted....I think). 😆 

The pseudo-logic way I stuck it in this thread is a lot easier for _my_ eye to follow, at least, than how I babbled it out in your PR. 🤔

dodexahedron:
Cool. I'll have a look at the new hotness in a few minutes.

> It's only worth preserving the last opened MenuBar.

***Depending on other specifics***, it is either easy or easy-ish to make at least this argument against that reasoning: Why should we reconstruct objects that are pretty heavy to construct (due to all the hotkey parsing and all that sort of stuff, especially), when they could very well be needed again, possibly very soon? Just seems somewhat unconventional and kinda wasteful.

I'm sure seeing the code will help answer some of the underlying technical bits, though, so I'll come back here once I get a good look at it.......assuming I don't get distracted. But I told alexa to remind me every 30 minutes to check my focus, just in case. 😅

dodexahedron:
Oh yeah also while it's in my mind...

Something both of you might find useful is the .net Community Toolkit has a standalone component for its CommunityToolkit.Diagnostics namespace (no other dependencies), which has a TON more throw helpers and guard methods, covering a bunch more exception types as well as a lot more use cases, in a semi-fluent style, similar to how Asserts look.

And it's another code generation-based thing, so it doesn't add a runtime dependency on the CommunityToolkit.Diagnostics package nor a design or build-time dependency for consumers of the Terminal.Gui library, either. Probably one we should add and make use of, for the rather nice consistency it can help provide, just like the built-in throw helpers do for the few exception types those are implemented for. 👍 

I have something partially written up for #3675 that I'll send over to @tig when it's done that also demonstrates a much cleaner way of dealing with all those conditionals, rather than having all the preprocessor directives for them. In short, it's primarily just making use of the `[Conditional(string)]` attribute, which allows you to call a method it is placed on as if it is always there, but the compiler will skip it without the need for messy preprocessor directives if the symbol given to the attribute is not defined. That and `[DoesNotReturn]` (on a method) or [DoesNotReturnIf(bool)] (on a parameter of a method) give cleaner and slightly DRY-er code, plus great static analysis like most of the throw helpers also provide (they use those attributes for exactly that).

Anyway.... Off to the git machine!

BDisp:
The unit test error isn't belong from this PR but reproduces in the v2_develop branch.

BDisp:
@dodexahedron please don't send anymore PR to my branches that aren't directly related with the issue that the PR was created. If you want to improve and cleanup code that you think the TG needs to, then submit a PR yourself to the TG repository and not for the forked repositories. I spend a while to fix an erroneous code you submitted to my branch but that was unnecessary for me take the responsibility to worry about that. I admit that was my fault when I merged your PR, but please don't send me more unrelated PR to my branches. Sorry if I offended you but that wasn't my intention. Thanks.

Terminal.Gui/Application/Application.Mouse.cs
```diff
@@ -56,6 +54,10 @@ public static void UngrabMouse ()
             return;
         }
 
+#if DEBUG_IDISPOSABLE
```
tig, line 57: While this is technically incorrect usage of ODE, I'm a fan of having this here for now... until we can do ODE right. Thanks.

tig (approved):
Great work @BDisp

PR #3686, opened 2024-08-23
TableView: Use caption of data column if set to render label

Remo:
## Proposed Changes/Todos

I believe the `ColumnName` property should only be a fall-back for the actual caption/label/title of a column.
That would allow us to specify an internal name, like a field name or property and still have a different name shown to the end-users. It's also what is written in the DataColumn class: https://learn.microsoft.com/en-us/dotnet/api/system.data.datacolumn.caption?view=net-8.0

## Pull Request checklist:

### DIFF
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -11,7 +11,7 @@ on:
       - '**.md'
       
 jobs:
-  build_and_test:
+  build_and_test_debug:
 
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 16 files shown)


tig:
@tznind owns TableView and should be a reviewer on this.

Thanks @remo!!!

tznind:
Saw this, looks like a good simple improvement.

Agree that the ?? probably isn't needed but doesn't hurt.

Would be good to have tests though.

I can add later if you don't have effort available

tznind (approved):

Remo:
@tznind yes, it's maybe a bit verbose to have `??` since it's already happening in the DataColumn class. I've removed it and I've also added a test.

tznind (approved):

PR #3688, opened 2024-08-23
Fixes #3687. ColorPicker isn't respecting the current UI culture.

### LINKED ISSUE #3687: ColorPicker isn't respecting the current UI culture.
That why actually some unit tests are failing.

BDisp:
## Fixes

- Fixes #3687

## Proposed Changes/Todos

- [x] Replace with CurrentUICulture

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drawing/ColorStrings.cs
+++ b/Terminal.Gui/Drawing/ColorStrings.cs
@@ -21,7 +21,7 @@ public static class ColorStrings
     public static string? GetW3CColorName (Color color)
     {
         // Fetch the color name from the resource file
-        return _resourceManager.GetString ($"#{color.R:X2}{color.G:X2}{color.B:X2}", CultureInfo.CurrentCulture);
+        return _resourceManager.GetString ($"#{color.R:X2}{color.G:X2}{color.B:X2}", CultureInfo.CurrentUICulture);
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


tig (approved):
Is there a way to have unit tests validate this so these mistakes (which I made) are ALWAYS caught?

BDisp:
> Is there a way to have unit tests validate this so these mistakes (which I made) are ALWAYS caught?

You didn't make any mistakes. Who made the mistake was me when I accepted the PR, sorry. I think, due the fact no more errors were caught it meaning that all the rest is correct. The issue was the mouse event was always running while the `MouseGrabView` is null and view not null, resulting the event propagating to the upper hierarchy. My apologies.

BDisp:
> Is there a way to have unit tests validate this so these mistakes (which I made) are ALWAYS caught?

I think the failing unit test is enough. But I think there will be have problems if the user creates a custom culture other the `en` by not having available resources strings, resulting in a empty color names.

tig:
> > Is there a way to have unit tests validate this so these mistakes (which I made) are ALWAYS caught?
> 
> I think the failing unit test is enough. But I think there will be have problems if the user creates a custom culture other the `en` by not having available resources strings, resulting in a empty color names.

It did not fail for me. It fails for you because you don't use `en`.

BDisp:
> It did not fail for me. It fails for you because you don't use `en`.

For this reason I think that all unit tests that uses resource string should have a non `en` culture. I propose `pt` if it's ok with you. Also I think they must fallback to the `en` culture if they wasn't yet translated for an existing resource.

PR #3704, opened 2024-08-29
Fixes #3702. Manipulating Adornment doesn't always cause LayoutSubviews.

### LINKED ISSUE #3702: Manipulating `Adornment` doesn't always cause `LayoutSubviews`
WindowsTerminal_8yd3k3DdVs

> > > Oh. Yeah, that happens in v2_develop as well.
> > 
> > 
> > Do you think it's a bug on the `Dim.Auto (DimAutoStyle.Content)`?
> 
> We're missing SetLayoutNeeded or LayoutSubviews somewhere.

IIRC, I think I suspected it was someting to do with SubViewsNeedsLayout not getting set right. The logic of all of that is pretty complex.

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3627#issuecomment-2316306055_

BDisp:
## Fixes

- Fixes #3702

## Proposed Changes/Todos

- [x] Force layout at once before calculate all the subviews.

## Pull Request checklist:


Terminal.Gui/View/Layout/DimAuto.cs
```diff
@@ -153,6 +153,11 @@ internal override int Calculate (int location, int superviewContentSize, View us
                                                            .ToList ();
                 }
 
+                foreach (View notDependentSubView in notDependentSubViews)
+                {
+                    notDependentSubView.SetRelativeLayout (us.Frame.Size);
```
tig, line 158: This should be `us.ContentView.Size`, right?

Also, did you try putting this inside the for loop below?

tig (changes requested):

Terminal.Gui/View/Layout/DimAuto.cs
BDisp: > This should be `ContentSize`, right?
> 
I think you are right. I'll change that.

> Also, did you try putting this inside the for loop below?

Yes I did but doesn't work because it's run the layout and also calculating at same time for each subView and we need they run all the layout first for all subViews. Then the subViews already have the correct position and size to calculate the correct position and dimension for the superView.

Issue #3444, opened 2024-04-30
Re-factor `Cursor` handling per `./docfx/docs/cursor.md`


tig:
Related: 

- #3432 

See:

https://github.com/gui-cs/Terminal.Gui/blob/v2_develop/docfx/docs/cursor.md

Note, in this we need to fix `Focused` and `MostFocused` such that these unit tests are addressed:

```cs
    [Fact, Trait("BUGBUG", "Fix in Issue #3444")]
    public void Focused_NoSubviews ()
    {
        var view = new View ();
        Assert.Null (view.Focused);

        view.CanFocus = true;
        view.SetFocus ();
        Assert.True (view.HasFocus);
        Assert.Null (view.Focused); // BUGBUG: Should be view
    }

    // View.MostFocused - No subviews
    [Fact, Trait ("BUGBUG", "Fix in Issue #3444")]
    public void Most_Focused_NoSubviews ()
    {
        var view = new View ();
        Assert.Null (view.Focused);

        view.CanFocus = true;
        view.SetFocus ();
        Assert.True (view.HasFocus);
        Assert.Null (view.MostFocused); // BUGBUG: Should be view
    }
```

dodexahedron:
👍 

Remember to include the `using` for those `View`s

Issue #3408, opened 2024-04-15
There have been long-standing repaint-flicker issues in the libarary. At one point I thought I fixed them all with a fix to how `ClearWhenVisibleFalse` was in play.


tig:
There have been long-standing repaint-flicker issues in the libarary. At one point I thought I fixed them all with a fix to how `ClearWhenVisibleFalse` was in play.

However, I did something in this PR that has caused the problem to re-appear. I've spent HOURs trying to track it down unsuccessfully. 

I am not going to hold up merging this PR to fix it, because it's cosmetic, but it needs to be found and fixed. Any help appreciated.

![FLUaXgV 1](https://github.com/gui-cs/Terminal.Gui/assets/585482/15212750-f4b5-4c80-97e7-3b39839d23b6)

Here I held down Ctrl-G to repro:

![e2DfkaP 1](https://github.com/gui-cs/Terminal.Gui/assets/585482/16748f55-1a0c-41ba-b7d2-18ad265eb411)

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3323#issuecomment-2057152669_

tig:
@BDisp if you're going to look into making Clip more sophisticated, I found the PR I thought I lost where I started down that path:

https://github.com/gui-cs/Terminal.Gui/pull/2606

BDisp:
> I'm done trying to debug this flicker thing for now. So frustrating!

That's true, it's really frustrating. I have also analyzed the cause and I am almost certain that it is necessary to limit the drawing of decorations to their thickness. It could also be someone from Label at Border who is cleaning outside of his area.

tig:
> @BDisp if you're going to look into making Clip more sophisticated, I found the PR I thought I lost where I started down that path:
> 
> #2606

Nevermind. That branch does not have all the really awesome code I wrote in it. I must have never pushed. Long gone. Sad.

BDisp:
> @BDisp if you're going to look into making Clip more sophisticated, I found the PR I thought I lost where I started down that path:
> 
> #2606

I've seen it, but it has a lot of unrelated changes and I'd rather start from scratch.

tig:
> > I'm done trying to debug this flicker thing for now. So frustrating!
> 
> That's true, it's really frustrating. I have also analyzed the cause and I am almost certain that it is necessary to limit the drawing of decorations to their thickness. It could also be someone from Label at Border who is cleaning outside of his area.

[...]

PR #3705, opened 2024-08-30
Fixes #2537, #3691, #3406 - New `Overlapped` Design, mouse/keyboard sizing - Progress on #2491

### LINKED ISSUE #2483: Remove need for `TileView` - Use `LineCanvas` to draw borders and auto-join borders

### LINKED ISSUE #2484: Move all rendering out of Toplevel, FrameView, Window, etc...  into View.
Now that the new View architecture is in place, we need to update all the other views.

This Issue focuses on just cleaning up the places where these views do weird drawing things

- Toplevel
- FrameView
- Window
- Dialog

### LINKED ISSUE #2537: Add mouse/keyboard resizing of views
Right now, dragging is handled w/in TopLevel. This needs to change to be handled within Border (or in Application, but via an interface with Border). 

- [x] Define `ViewArrangement.Movable` - if set and view has a top Border, it can be dragged with mouse. Always with keyboard (although no visual indicator if no border).
- [x] Define `ViewArrangement.Sizable` - if set and a view has left/right/bottom Border it can be resized with mouse. Always with keyboard (although no visual indicator if no border).
- [ ] Determine keyboard drag/size UI
- [ ] Remove temporary hack in `Application.cs`:
image

Related: 
- #https://github.com/gui-cs/Terminal.Gui/issues/2502. 

My proposal:

### Mouse:

- Dragging `Border` top or bottom resizes the view vertically
- Dragging `Border` left or right resizes the view horizontally

### Keyboard:

- Hit `Ctrl+F5` to enable **Arrange Mode** [^whyctrlf5]
- A line will get a `◊` indicating it's active for resizing. Left/Right/Up/Down will size.
- Tab to move to next line
- Shift-Tab to move back 
- `Ctrl+F5` or `Esc` to exit **Arrange Mode** 

### Mouse:

- Dragging `Border` top moves the view
- Dragging `Border` bottom resizes the view vertically
- Dragging `Border` left or right resizes the view horizontally

### Keyboard:

- Hit `Ctrl+F5` to enable **Arrange Mode** [^whyctrlf5]
- A `◊` appears in top left corner of Border indicating **Arrange Mode**
- If `ViewArrangement.Movable` is set `Left/Right/Up/Down` will move the view 
- `Tab` and `Shift+Tab` will cycle through any sides that can be sized and "Move"
  - UNLESS: `ViewArrangement.Resizable` (all side-flags) is set. Then `◊` appears in bottom right corner of Border. 
- A side selected for sizing will get `◊` centered indicating it's active for resizing. 
- `Left/Right/Up/Down` will size as appropriate
- `Ctrl+F5` to exit **Arrange Mode** 

[^whyctrlf5]: On Windows `Alt-Space, M` enables move-with-keyboard. On Mac `Ctrl-F5` "Move focus to the window toolbar". No standard on Linux. `Ctrl-F5` isn't used by Windows Terminal. Happy for a better suggestion.

### LINKED ISSUE #3406: Fix focus and keyboard nav with `Adornments`
Right now, in #3323, mouse works but

- Focus does not
- Keyboard does not work well 

How should Focus work? 

* By default all Adornments have `CanFocus = false`. Focus & keyboard nav will only work if set to true.
* By default all Adornments have `TabStop == TabBehavior.NoStop`. Thus `Tab` or `F6` will not nav into any Adornments or subveiws of adornments. 

The internal API `GetFocusChain()` returns all of a View's Subviews PLUS the 3 adornments (Padding, Border, Margin ... in that order). 

Thus, IF `adornment.TabStop` is same as `adornment.Parent.TabStop` (and not NoStop), naving into /out of the adornment with keyboard will work. Not recommended tho.

Generally if a View wants focuable subviews in an adornment, the view needs to provide a keyboard model for naving into the adornment. See how `BeginArrangeMode` in Border works by enabling `Ctrl-F5` do put focus on the Border.

### LINKED ISSUE #3691: `Popover` - A consistent way of enabling a Subview to popup outside of a View
(Very drafty for now... just some thoughts).

This Issue is a proposal for how to build into TG v2 a consistent way for a View to have a UI element show itself outside of the View's Viewport. 

*Popover* - A `View` that that is displayed outside of the View that owns it's Viewport.
 
There are at least these use-cases of Popovers:

## Autocomplete Popup

`TextView` would like a list of autocomplete items to be displayed below the view, at the cursor position as the user types. The user can use mouse/keyboard to select an autocomplete item and the pop over disappears. 

### Current Implementation

- Adds `_popup` (a `ListView`) to `this.SuperView` 
- Uses `Visible` to show/hide
- (In #3627 I've had to change the implementation so `_popup` only gets removed/disposed when the host is removed/disposed).
 
### Current Implementation

- Does not support pop-over; The `ComboBox` needs to be sized to allow the drop-down to show if HideDropdownListOnClick is true.
- `_listview` (of type `ComboListView`) is a subview of `Combobox`
- `ComboListView` overrides `OnMouseEvent` (which is `internal`!) with a bunch of hackery to deal with when it should be displayed

### Current Implementation

- `MenuBar` is a "special" subview of `Application.Top` (Toplevel has knowledge of it).
- When `MenuBar` is activated (via keyboard/mouse/api) it creates instances of `Menu` and adds them to `Top`
- Both `MenuBar` and `Menu` have convoluted logic for sensing when the menus should close (via mouse & keyboard). 
- When an open `Menu` closes, it is removed from `Top`

### Current Implementation

- Uses a "fake" `MenuBar` that is created when `Show` is called. 
- `_menuBar` is never actually added to any views; `BeginInit`/`EndInit` are called manually. 

## Tooltip

As a user, when I hover the mouse over a `View`, I'd like a temporary popup to appear providing a "tip" regarding the View's purpose. 

### Current Implementation

- Not currently implemented

### Tenets (Unless you know better ones...)

* **There can only be one** - Just like in Highlander, there can only be one Popover visible and active to the user at a time. None of the use-cases above lead to needing to have more than one, thus this is a simplifier. [^1]
* **Its Rude to Mess With Someone Elses' Subviews** - Asking View developers to have to code defensively around the fact that some Subview they add, or some other agent, is adding/removing subviews is just bad juju. It leads to bugs in object lifecycles and requires skill to get right (e.g. adding a Subview during draw can result in `Subviews` being changed during a `foreach` iteration). In this design we will ensure there's no need to mess with another View's `Subviews`.
* **Any View can be a Popover** - The design should not require a View that "Pops Over" be coded specially for that purpose. 

### Design

At the `Application` level we'll add a peer of `Application.Top` named `Application.Popover`:

* In the run loop, after `Top.Draw()` is called, if `Popover.Visible` we'll call `Popover.Draw()`. This will ensure the Popover is always drawn over everything else.
* In keyboard handling, if `Popover.Visible && Popover.HasFocus` we'll give `Popover` a chance to handle key events. 
* In mouse handling, if `Popover.VIsible` and `Popover == Applicaiton.MouseGrabView` we'll let it have mouse events. If a click happens outside of `Popover.Frame` we set `Popover.Visible = false`.
* In focus/navigation handling, if `Popover.HasFocus`, if any other View gets focus, we'll set `Popover.Visible = false`.
* For layout, whenever `Top.LayoutSubviews` is called, we'll then call `Popover.SetRelativeLayout`
* Two APIs:
  * `Applicaton.ShowPopover (View popoverView)`
 
```cs
public static bool ShowPopover (View popoverView) 
{
   if (Popover is {})
   {
      Popover.Visible = false;
   }

   if (!popoverView.IsInitialized)
   {
      popoverView.BeginInit();
      popoverView.EndInit();
   }

   Popover = popoverView;
   Popover.Visible = true;
   Popover.SetRelativeLayout(screen);
}
```

  * `Application.HidePopover ()` 

```cs
public static void HidePopover () 
{
   if (Popover is {})
   {
      Popover.Visible = false;
   }
}
```

`View` will have no knowledge of the popover concept.

Any `View` subclass that wants a Popover, will:

- `_myPopover = new MyPopoverView();`
- Note, `_myPopover` will always have `this.SuperView is null`.
- Any time the popover should be displayed:
  - Set `_myPopover.X/Y/Width/Height` to ??? (NEED TO FIGURE THIS OUT)
  - Call `Application.ShowPopover(_myPopover)` 
- If the view wants to explicitly hide the popover, 
  - Call  `Application.HidePopover()`
  
[^1]: I've considered "As a user, when press a special key, I'd like all visible Views to show temporary popup providing a "tip" regarding the View's purpose." This could be implemented by having the container view showing a Popover who's job it was to show "tips" for each subview.

### LINKED ISSUE #3735: `Slider`: API for setting options is clunky
Currently:

```cs
    /// <summary>Causes the specified option to be set and be focused.</summary>
    public bool SetOption (int optionIndex)
    /// <summary>Causes the specified option to be un-set and be focused.</summary>
    public bool UnSetOption (int optionIndex)
```
The problem is these APIs want to change which option is focused.

As part of #3705 I want to use Slider in `ArrangementEditor` and this is causing me fits.

So I will fix this there...

### LINKED ISSUE #3741: Tiled mode: need a way to declare tiled view size dependencies
This is a PR for after #3705 is merged.

Note that when I size Tiled1, Tiled4 gets confused.

GswFB00 1

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3705#issuecomment-2356456719_

### LINKED ISSUE #3742: Navigation Scenario: Add a `TabStop` Editor
It would be useful to be able to change the `TabStop` setting on the Views in this scenario.

### LINKED ISSUE #3743: Subviews in `Adornments` that are not `ViewArrangement.Fixed` are not supported
In

#3705 

I didn't bother to test/enable what happens if a subview of an adornment use ViewArrangement.Overlapped etc....

Probably not a use-case that needs to be addressed pre-V2, but adding as an Issue to track.

tig:
## Fixes

- Partially Addresses #2491 by 
  - Finishes `ViewArrangement.Overlapped` support
  - Fixes #2537
  - Fixes #3735 
  - Fixes #3406
  - Fixes #2484 
 
## Proposed Changes/Todos

- [x] Remove temporary code in `OnDrawContent` and get z-order correct.
- [x] Enhance `Bars` scenario to illustrate Popover menu concept
- [x] Enhance `Navigation` scenario with `ArrangementEditor`; like `AdornmentEditor` makes it wasy to change view settings related to arrangement for testing/demos.
- [x] Clone `Navigation` scenario to `Arrangement` scenario - one for each 
- [x] Implement support for `ViewArrangement.Resizable`
- [x] Fix `Draw`
- [x] Remove more code from `Toplevel` - See how far I can get in removing `Toplevel` allogether without re-doing all the `Runnable` stuff (for a later PR focused on `IRunnable`).
- [x] Fix #2537
- [x] Update `docfx/docs` and other API docs

## For future PRs

- [ ] Enable `Runnable` concept
- [ ] Fix #3741
- [ ] Fix #3742
- [ ] Fix #3691
- [ ] This PR re-broke the cursor flickering issue (#3408). Will fix this in #3444
- [ ] Create a scenario that illustrates an "MDI" like app (replacing BackgroundWorkerCollection). Show how `MenuBar` and `StatusBar` can be in any View, even multiple.
- [ ] Fix #3743
- [ ] Fix #2483

## Pull Request checklist:


tig:
In making keyboard-based arrangement work, I needed to finally fix 

#3406

0vXorhO 1

tig:
Latest: Tiled mode is semi-enabled. Remaining challenges:

- [ ] Need a simple way to declare tiled view size dependencies. Note that when I size Tiled1, Tiled4 gets confused.
- [ ] Auto join of borders/drawing of titles needs fixing (#2483).

GswFB00 1

Terminal.Gui/Application/Application.Keyboard.cs
```diff
@@ -155,7 +153,7 @@ public static bool OnKeyDown (Key keyEvent)
     }
 
     /// <summary>
-    /// INTENRAL method to invoke one of the commands in <see cref="CommandImplementations"/>
+    ///     INTENRAL method to invoke one of the commands in <see cref="CommandImplementations"/>
```
BDisp, line 156: I think it's private.

BDisp (approved):

Terminal.Gui/Application/Application.Keyboard.cs
tig: Is this all you found in  your review?

tig:
Closing. This branch is now rebased with 

- #3571

PR #3710, opened 2024-08-31
Implements #3692 - First round of ConsoleDriver cleanup

### LINKED ISSUE #3692: Refactoring, cleanup, and abstraction for ConsoleDrivers
Filing an actual issue now that I've gotten to work on things that have been talked about here and there and some discretionary/opportunistic bits along the way.

This one starts with the console drivers.[^NoBullyingTheDrivers]

I'm nearly done with what I'm slapping a round 1 label on and will have a PR ready later today, hopefully.

Things that are addressed so far:

 * General code cleanup and organization.
 * Bug fixes
    * Addressing both theoretically possible bugs based on the intent and usage of the code as well as possible problems pointed out by static analysis, such as potential null reference exceptions.
    * Tweaks to certain exception throwing and handling to help identify, report, and handle what's going on and to align exception handling closer to design guidelines and conventions.
    * Preventing the conditions that lead to certain avoidable exceptions.
 * Type-safety/correctness
    * Things like generics, mostly.
    * Also cases related to nullable structs, null checks, polymorphism, and native interop.
 * Thread-safety
    * That part (so far) is minor and mostly informational - comments and such.
    * More specific concurrency work will happen later on, whether here or in a separate issue.
 * Performance
    * Various approaches to try to make things a bit leaner where I see a reasonable opportunity to do so.
    * Things like ref passing, use of ref locals, use of Spans, avoiding run-time values that are still just constants and making them compile-time constants, etc.
    * Priorities are in this order:
       1. Speed and simplicity, with basically equal priority
           * Simplicity - I'm not going to do things that are hard to maintain or understand.
           * Execution speed - Largely avoiding copies and avoidable heap operations.
       3. Memory consumption
       4. Compilation, design-time, and testing performance
           * Small but non-zero work items to reduce how heavy the solution is on dev and CI machines, overall.
       5. Compiled assembly size
           * Not really directly addressing (at least not yet anyway), but paying attention to it. 
 * Trim-friendliness
    * Eliminating uses of reflection.
    * Use of generics instead of `Type` values.
       * This one also falls under performance, and is a type-safety thing, too. Type-typed method parameters are a break in the chain for static type analysis, which is why they aren't trim-friendly.
 * Nullability context
    * Turning it on wholesale for most files I'm touching, and in targeted sections in others I'm not interested in fully migrating in this round.
       * Also using the resulting analysis to improve code in various places where it makes or could make a difference.
    * This also falls under bug fixes and type safety.
 * Interop/PInvoke
    * Migration to source-generated interop.
    * Using CsWin32 to create more robust native interop types and methods.
       * Some or potentially all generated code may get copied to permanent files since the APIs used are stable, to enable hand-tweaking and for sake of transparency and simpler debugging.
 * Some structural reorganization
    * Moved some files.
    * Split multi-type files that were worked on into one type per file.
 * Testing improvements
    * More parameterization
    * DRY-er code (largely enabled by generics and things like `MemberData`)
    * Use of Assume (from the Xunit.Assume extension package) to differentiate pre-conditions in some tests from the test itself (my god why is Xunit so terrible???).
    * Better determinism/repeatability.
       * Removing some uses of CombinatorialRandom, since that makes every test run different where it is used.
    * Using and re-using new common helper methods to provide simple, clean, and consistent generation of test cases.
    * Applying attributes such as category traits to various tests and test classes.
    * Tweaking some tests to provide better accuracy, speed, simplicity, etc
 * Probably more, but I want to get back to work now.

I've done it as a ton of mostly smaller commits, as usual, to make it easier to follow along through the commit graph. However, one of the biggest changes I made at the start is actually what prompted some of the additional work that got done in this round, and hasn't been committed to the branch yet, so the current head of https://github.com/dodexahedron/Terminal.Gui/tree/v2_d12_some_driver_cleanup_1 will not compile at the moment, though my local copy (which of course DOES have those changes) compiles and passes all tests. I just need to finish the work on it before I can commit and push that part.

PR will come when I push the commits with that work finished.

There are also a few extra bits of work I want to do before round 1 is done, beyond that, but those will come after a draft PR is submitted.

[^NoBullyingTheDrivers]: No specific reason about the code or code quality itself for picking that as a starting point for this work. It was based on the fact that they're alphabetically early in the project folder tree, they're some of the first code that runs, and the first warnings in the list when I turned analysis up a notch locally brought me to one of those files.

dodexahedron:
OK, so I'm cutting the first round of this off here.

It's ready for y'all to have a look at it, but it's not ready to merge because I need to finish committing a couple of build-related tweaks, which I'll get to when I wake up.

There are various TODO-style comments I dropped here and there in it. Most are notes for myself for future rounds of this work.

## Fixes

- Fixes #3692

## Proposed Changes/Todos

- [x] Clean up/reduce Win32 PInvokes.
- [x] Switch to using FileStream under the hood for NetWinVTConsole.
- [x] Various fixes to tests and incidentally related types.
- [x] Add some test helpers for providing drivers to test methods.
- [x] Misc cleanup, doc improvements, and optimizations.

## Pull Request checklist:

   - [x] My changes address a ton of warnings from a stricter analysis level.


dodexahedron:
Looks like there are also some i18n issues in CI for some tests that I'll need to address.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -420,6 +429,13 @@ public void FillRect (Rectangle rect, Rune rune = default)
     /// </returns>
     public bool IsValidLocation (int col, int row)
     {
+        // INTENT: Half of this is impossible and thus redundant.
+        // Screen is hard-coded to have origin at 0,0.
```
tig, line 433: But may not be 0, 0 in the future (non-full screen apps).

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -420,6 +429,13 @@ public void FillRect (Rectangle rect, Rune rune = default)
...
+        // Screen is also hard-coded to be Cols x Rows in size.
```
tig, line 434: But may not be in the future.

Terminal.Gui/View/View.cs
```diff
@@ -106,15 +106,8 @@ namespace Terminal.Gui;
 
 #endregion API Docs
 
-public partial class View : Responder, ISupportInitializeNotification
+public partial class View : Responder, ISupportInitializeNotification, INotifyPropertyChanging, INotifyPropertyChanged
```
tig, line 109: I've written this several times regarding `INotifyPropertyChanging`. Not sure if you didn't see it or don't believe me or disagree. 

`INotifyPropertyChanging` does not do what many people think it does. It does NOT prevent the change from happening, only allows you to record that it is happening.

https://github.com/gui-cs/Terminal.Gui/issues/3561#issuecomment-2194932007

https://github.com/gui-cs/Terminal.Gui/pull/3625#issuecomment-2265847460

I don't think we should be using it.

Terminal.Gui/Views/ColorPicker.cs
```diff
@@ -235,18 +247,22 @@ private void DisposeOldViews ()
...
+        SetSelectedColor (_strategy.GetColorFromBars (_bars, Style.ColorModel), false);
+    }
 
+    /// <exception cref="Exception">A delegate callback throws an exception.</exception>
     private void SetSelectedColor (Color value, bool syncBars)
     {
         if (_selectedColor != value)
         {
-            Color old = _selectedColor;
+            RaisePropertyChanging (this, new (nameof (SelectedColor)));
```
tig, line 261: See above.

Terminal.Gui/Views/ColorPicker.cs
```diff
@@ -235,18 +247,22 @@ private void DisposeOldViews ()
...
             _selectedColor = value;
+            RaisePropertyChanged (this, new (nameof (SelectedColor)));
```
tig, line 263: I do not think we should start raising INPC until we have use-cases that require it (that won't work with the existing events). 

As I said here: https://github.com/gui-cs/Terminal.Gui/pull/3625#issuecomment-2265847460




> And regarding `INotifyPropertyChanged`: It may be used extensively, and it has some good characteristics, but it has some awful ones as well. It is designed primarily for data binding scenarios and (other than some awful things like requiring strings for property names) works great for that. I will not adopt it at the core of TG except in cases where the use case is primarily a data binding case.

> In the future, once we get around to building a XAML-based or -like markup UI system, it will be trivial to add `INotifyPropertyChanged` support on top of the pattern I am showing here. In fact, I think it could even be automated.

tig (changes requested):
Loving this! See my comments.

dodexahedron:
> Loving this! See my comments.

Will have a look when I'm back home.

I'm semi-out and about for the long weekend. 🥳

Terminal.Gui/Input/Mouse.cs
```diff
@@ -125,22 +141,25 @@ public class MouseEvent
...
+    public override string ToString () => $"({Position}):{_flags}";
+
+    /// <summary>The View at the location for the mouse event.</summary>
+    public View? View { get; set; }
+
+    /// <summary>
+    ///     Gets a boolean indicating if <see cref="Flags"/> has one or both of <see cref="MouseFlags.Button1Pressed"/> or
+    ///     <see cref="MouseFlags.Button1Released"/> set.
+    /// </summary>
+    internal bool IsMouseButton1PressedOrReleased => (Unsafe.As<MouseFlags, int> (ref _flags) & BUTTON1_PRESSED_OR_RELEASED) != 0;
```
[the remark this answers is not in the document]
dodexahedron: Well, I went ahead and read your feedback and suppose I can answer a couple of the simple ones before I get back to a PC.

> So will we define one of these for all 4 mouse buttons?

Maybe. But probably not. It's a targeted optimization for a let's call it "warm" code path. That's why that bool is internal and the constant is private. If I can make the bool even less visible than internal, after more of this work is done, I will. Whether this member even remains at the end of all the work isn't a given anyway, depending on how things shape up.

Terminal.Gui/Input/Responder.cs
```diff
@@ -50,20 +48,20 @@ protected virtual void Dispose (bool disposing)
     }
 
     // TODO: v2 - nuke this
```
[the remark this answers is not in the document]
dodexahedron: That was a pre-existing comment, but yay. 🥳

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -1081,7 +1081,11 @@ public WindowsConsole.KeyEventRecord FromVKPacketToKeyEventRecord (WindowsConsol
         };
     }
 
-    public override bool IsRuneSupported (Rune rune) { return base.IsRuneSupported (rune) && rune.IsBmp; }
+    // INTENT: Is the same issue as in CursesDriver also true here?
+    // As far as I can tell from messing around with toy apps, the windows console appears able to show you any
+    // codepoint that your font supports.
```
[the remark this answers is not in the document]
dodexahedron: Cool.

I'll document it in place.

Terminal.Gui/Input/Responder.cs
```diff
@@ -3,29 +3,27 @@
 namespace Terminal.Gui;
 
 /// <summary>Responder base class implemented by objects that want to participate on keyboard and mouse input.</summary>
+[MustDisposeResource (false)]
```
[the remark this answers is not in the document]
dodexahedron: Will have a look.

Terminal.Gui/View/View.cs
dodexahedron: I think there might be a little bit of a misinterpretation of what the INotify* interfaces and events are for and do.

They're by no means a replacement for dedicated change events for specific properties, for like 100% of the reasons you mention.

As the name implies, it is a notification. It is informational _only_ and is there to be used by interested consumers regardless of what else may happen. That's also why it is intentionally not cancelable. If you need to know something happened, allowing some other unknown piece of code to hide it from you is black box behavior that is expected in cancelable change events, but very much not desirable for an observer for myriad purposes. Even simple logging uses or things like telemetry are common reasons for a consumer wanting to use IN* events.

And remember - we are not the application. We are used by an application.

Both of the INPC events and the INCC event are very useful in a lot of cases. When something needs to know _that_ something is going to change or has changed, that's what it's there for. And that's why, yes, it's used by WPF/WinUI for data binding - because it's the simplest way to achieve what that feature needs without making any further assumptions and without allowing other code to interfere with it.

Do I think we ourselves should be _consuming_ those events? No, not in almost any case I can think of.

But, should we provide it for consumers, especially since it's a trivial thing to implement and carries no baggage for us? Absolutely.

Dedicated change events are a different beast entirely and are important to have for specific functionality, and I'm glad we have the ones we do or I'd push to add them as well. In fact, if/when I eventually get to the event work I want to do, there's a decent chance there may be a few more such events I'll want to add for sake of completeness.

Not providing such a simple yet useful feature for consumers does them a disservice, if you ask me. And it costs us next to nil. 🤷‍♂️

Hopefully that makes my intentions and the purpose of all that clearer? If not, or if you have any other concerns, I'm more than happy to continue talking about it. 🙂

Terminal.Gui/Views/ColorPicker.cs
dodexahedron: Resolving this one in favor of the other one to keep it in one place.

Terminal.Gui/Views/ColorPicker.cs
dodexahedron: Resolving this one too to keep the convo in one thread.

Terminal.Gui/View/View.cs
dodexahedron: As for the whole XAML thing I've brought up before:

That's actually a perfect example of a reason TO have it - especially since events and interfaces are the kind of thing can't be bolted on by a consumer without modifying the TG source itself.

I don't think XAML functionality belongs in the core Terminal.Gui library at all for many reasons, not the least of which is it would be quite a bit of bloat for anyone not wanting/needing to use that feature. My intention for that was to write a separate library myself that uses TG as the console adapter/engine/framework that renders the markup. It's almost definitely going to end up quite similar to how WPF works, where it's markup that results in source-generated C# that is the actual compiled code.

dodexahedron: Performance-wise...

If it were some heavy thing that had a negative impact on the function or performance of the core library? Then yeah definitely I wouldn't be in favor of it.

But it really isn't.

In fact, since it operates via the core implementation of .net and uses compile-time constants, it's also pretty much the lightest-weight way to notify anything of anything. Those string names are, in .net terms, already interned at compile time, which turns into a blob in the data section of the binary at JIT time, able to be referenced as `const wchar_t* const` basically.

If a consumer takes action based on the string value, they're still comparing against those interned strings, and most likely themselves also using an implicit constant string, so even in that case it's still extremely lightweight. It's a method call passing one pointer to static data in the compiled binary.

dodexahedron: Another key point about what it does, which isn't served well by other events:

It allows an observer to know that _anything_ in the public API surface of a type implementing it has changed without needing to subscribe to every individual event, and without us needing to publish events for every single property.

There are plenty of properties across the library that don't have explicit events tied to them. How does an interested observer find out if those changed? They can't, because they have no hook into the implementation to do so.

Any means of doing it without there already being a mechanism provided by TG would require either modifying TG code or using an interceptor or other form of reflection-based manipulation of the types, which isn't fun or pretty.

dodexahedron: There's also `IObserver<T>` and `IObservable<T>`, which came along later but do the same sort of thing via subscription methods and callback delegates, which is the same idea and often actually implemented on top of events under the hood to avoid duplication.

Now, as for what a consumer can do with the events? Anything they want. And we don't need to care nor should we care or interfere. We're just letting them know, if they care.

And if they want access to the sender of the event, for whatever purpose that we also have no business caring about, it's right there in the delegate signature for the event (`object? sender`), and they are free to use or abuse things via that as they please, too.

And there's one last thing that comes to mind for now that I'll add after this comment just to keep things in chunks...

dodexahedron: OK, so last thing for now:

***Especially*** since we have no hooks/facilities for logging in TG at present, something like INPC is kind of a bare minimum thing we can do to allow a consumer to log activity from the library and their use of it, without us adding any logging facilities (though I _really_ think we should do that anyway at some point - maybe in a feature release?).

That at least gives a consumer a way of being able to log when things happen, whether caused by their explicit property accesses or by internal TG code accessing those properties. Without it, the consumer pretty much has a black box for anything not publishing a dedicated event.

Lack of logging/visibility outside of a debugger was actually a big enough gripe for me in various usages I have of TG v1 that I have an internal modified version of TG that just has a static ILogger (most often backed by NLog) bolted onto every class, so I can quickly add logging for things I need, either in debugging OR a live running instance of an application.

dodexahedron:
Test failure in CI was a case-sensitivity issue. VS created the resource file with "es-mx", but the test expected "es-MX".

Fixed.

Terminal.Gui/Input/Responder.cs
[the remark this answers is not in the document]
dodexahedron: Yeah, I imagine as I continue with further stages of this work that this will probably end up getting done as a side bonus.

Terminal.Gui/Input/Mouse.cs
```diff
@@ -99,6 +101,7 @@ public enum MouseFlags
 }
 
 // TODO: Merge MouseEvent and MouseEventEventArgs into a single class.
+// TODO: Not that.
```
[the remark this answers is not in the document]
dodexahedron: Certainly happy to chat about that. Out of scope for this PR though. Just a comment I dropped in for visibility later on. :)

dodexahedron:
Alrighty round 1 ready to go.

Many of the last several commits were largely the cleanup I mentioned earlier, like addressing notes to myself and formalizing some stuff.

Next up, I have an already partially completed branch that continues work on the driver-level stuff - in particular, making the API consistently usable and extensible internally and externally, which then will lead to more goodies in the next phases that will make life easier in TG, the tests, and for consumers, and will finally de-couple at least that area of the API from the internal implementation.

BDisp:
> (@BDisp, I imagine a lot of our translations for es-MX and pt-PT will end up being at least similar, so that might save us a little mental load whenever we do any localization work.)

Not really, they are very different. Maybe some Spanish user can do that translation someday. Why did you chose the es-MX instead of the es-ES, which is more universal?

dodexahedron:
> > (@BDisp, I imagine a lot of our translations for es-MX and pt-PT will end up being at least similar, so that might save us a little mental load whenever we do any localization work.)
> 
> Not really, they are very different. Maybe some Spanish user can do that translation someday. Why did you chose the es-MX instead of the es-ES, which is more universal?

I picked MX because that's the dialect I know. 🤷‍♂️ 

Certainly closer to es-ES than to Portuguese, of course.

It's like... More of a difference than between en-UK and en-US but less than between es-* and pt-*.

dodexahedron:
I'm going to rebase this on current v2_develop in a minute.

Stay tuned.

dodexahedron:
Rebased and ready again. 👌

dodexahedron:
Strike that. One more commit to rebase on top of.

dodexahedron:
There we go. All up to date now. All systems go for review/merge 🟢

dodexahedron:
Hm.

Not sure why the CI runner failed on that.

I'll force push again in case I missed something, since it runs just fine here. 🤔

dodexahedron:
There's broken file history and git did not deal with it well _at all._

I'm bisecting to track down where the change happened and which code should actually be in use.

dodexahedron:
Yep.

That's exactly what it is.

[This merge commit](https://github.com/gui-cs/Terminal.Gui/commit/4430fe2cc62352a97837cb3302764f2e7f324bad#diff-3a515e9cb1fd07947418d9374e3fdecefb5ef9360e41f16e61c985673a9caa85) is where things change, but it was one of the busted merges that rewrote all history up to that point.

This commit has the correct code for the method that is breaking in CI after the rebase.

During rebase, since there was a change dependent on related history, git tried to trace it back to the right point in the graph, but failed miserably.

From what I'm looking at, it actually looks like a potential git bug, triggered by the specific situation created by the disconnected graph.

To test, I rebased on an earlier commit and the proper history was chosen. Then I rebased again on HEAD of v2_develop, and it failed in exactly the same way. Fun times.

I'm going to have to manually fix it, but that's essentially just rolling back to the previous commit, grabbing the file with busted history, rolling back forward, and pasting it over the broken one, effectively restoring it to before _my_ changes. Then I can re-apply my changes and make it all happy again.

dodexahedron:
Important note about that.

If you merge one of those old branches or anything that started before them into your branch, you WILL cause these sorts of regressions and breakages, and they may not be shown as diffs. The only way to prevent that is to rebase on current code, rather than merge the old code into yours.

Unfortunately, that even means merging v2_develop into your branch, if the last common parent for an affected conflicting file was at one of those earlier points in the graph.

(In case you needed a stronger technical reason to rebase rather than merge.)

dodexahedron:
@tig:

What is the purpose/function of this?

```csharp
    private static ConsoleModifiers GetModifiers (ConsoleModifiers modifiers)
    {
        if (modifiers.HasFlag (ConsoleModifiers.Shift)
            && !modifiers.HasFlag (ConsoleModifiers.Alt)
            && !modifiers.HasFlag (ConsoleModifiers.Control))
        {
            return ConsoleModifiers.Shift;
        }

        if (modifiers == (ConsoleModifiers.Alt | ConsoleModifiers.Control))
        {
            return modifiers;
        }

        return 0;
    }
```

That's part of the code with busted history. And it is puzzling since it doesn't really seem to do what its name would suggest.

The version of that method that exists for much more of history (including up until the original base for this branch) is:

```csharp
    /// <param name="alt">The alt key.</param>
    /// <param name="control">The control key.</param>
    /// <returns>The console modifiers.</returns>
    public static ConsoleModifiers GetModifiers (bool shift, bool alt, bool control)
    {
        var modifiers = new ConsoleModifiers ();

        if (shift)
        {
            modifiers |= ConsoleModifiers.Shift;
        }

        if (alt)
        {
            modifiers |= ConsoleModifiers.Alt;
        }

        if (control)
        {
            modifiers |= ConsoleModifiers.Control;
        }

        return modifiers;
    }
```

Which actually does what it says.

BDisp:
Avoid rebasing as possible, doing it can cause a lot of headaches. Merge instead. I only need to rebase after a PR was already merged and before I try delete the local branch, I get a warning that there are unsaved changes. In this case I need to rebase to restore the remote branch history. If there is no incoming nor outcoming and I successfully delete the local branch, then everything went fine. With PR that aren't yet merged I normally merge the remote with my local branch. The GitHub `Update Branch` button actually does the merge the remote fork into the PR branch.

dodexahedron:
> Avoid rebasing as possible, doing it can cause a lot of headaches. Merge instead. I only need to rebase after a PR was already merged and before I try delete the local branch, I get a warning that there are unsaved changes. In this case I need to rebase to restore the remote branch history. If there is no incoming nor outcoming and I successfully delete the local branch, then everything went fine. With PR that aren't yet merged I normally merge the remote with my local branch. The GitHub `Update Branch` button actually does the merge the remote fork into the PR branch.

That's what caused the problem in the first place.

Don't merge outward.

If you have a problem in the rebase, it's precisely because of this kind of thing and means something broke that you're not being shown, and it needs to be fixed. A merge will not show you that. A rebase will. That's why this appeared. Otherwise, it would have silently been a regression of some pretty important code across several files.

dodexahedron:
Now...

If we didn't have that history problem from before the PR merge process was fixed, it would mostly not matter and be fine either way.

But we ***do*** have that history to contend with, until such time as there is no longer a direct link (via more merges) back to that point in time.

dodexahedron:
There is something we can do that would give us a clean starting point. But it would require outstanding work to be _rebased_ on top of that.

The only solutions to the original problem and the issues that it will continue to create occasionally all require you to rebase anything with common parents that are direct descendants of squash merges on something that has an explicit full file rewrite (not a partial diff) after that point.

dodexahedron:
Github actually also has an option at the repository level that can be set to help prevent (though not necessarily totalle eliminate) bad merges like that.

It's called something like "Require linear commit history."

It would disallow most merges that are likely to result in these sorts of problems. I suspect it would have prevented merge of the commit that I rebased this on top of, in fact.

dodexahedron:
I'll have to fix this in a little bit. I have some other stuff to attend to right now. I'll mark as draft again until I come back to it later this afternoon.

dodexahedron:
I'm going to leave the version of that method that takes and returns ConsoleModifiers in the code, since it is used. But I'm going to rename it based on what it appears to do. Feel free to hate one the name/suggest alternatives.

I'm thinking something like `FilterModifiersForScanCode`.

dodexahedron:
Fixed and confirmed repeatable results with a fresh branch from both deadd5bdb4ed2e3faa1fe6e339276ca5df4789b5 and 56d9c5948465c44939ca3e80f65c4fbdace473b5, each performed on fresh clones of the repo at those points.

dodexahedron:
There we go. That's more like it.

Yeah, it was interesting to see how, even though I never touched the ConsoleModifiers overload of that method, git decided it was the same thing in the diff and removed both.

Adding it back in along with the fact that that file has been pretty much entirely rewritten due to splitting out types to their own files should at least protect that file from that kind of problem again unless some really old branch happens to get merged over it again and a conflict isn't resolved by taking the whole file from the newer version.

dodexahedron:
> The GitHub `Update Branch` button actually does the merge the remote fork into the PR branch.

That behavior is configurable. You can have it fast-forward, merge, or rebase.

The safest and most formally "correct" thing to do to ensure linear history is to first update your fork's v2_develop branch, fetch/pull it locally on your machine to sync, and rebase your work branch on top of that.

BDisp:
> That behavior is configurable. You can have it fast-forward, merge, or rebase.
> 
For me is the best behavior.

> The safest and most formally "correct" thing to do to ensure linear history is to first update your fork's v2_develop branch, fetch/pull it locally on your machine to sync, and rebase your work branch on top of that.
> 
That is what I do, less the rebase where I use merge when the local branch wasn't yet merged into the mainline branch.

> Ensuring that you always only fast-forward the mainline branch that you branched from prevents cycles in the graph.

What matter, is the mainline branch history. When I said before that sometimes I get the warning that I have unsaved changes on my local branch that already was merged into the mainline branch, when I try to delete it, then in these cases I have to rebase to ensure that is everything well. After doing it I don't see any changes and finally I can delete my local branch. Sometimes the difference is some merge ID in my local branch that isn't equal to the mainline branch, but the changes contained on it are the same on both branches and so, it's commit ID that is conflicting and fixed by the rebase.

Terminal.Gui/ConsoleDrivers/Net/NetDriver.cs
```diff
@@ -0,0 +1,902 @@
...
+
+namespace Terminal.Gui.ConsoleDrivers.Net;
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+using static EscSeqUtils;
+
+internal sealed class NetDriver : ConsoleDriver
+{
+    private const int COLOR_BLACK = 30;
```
tig, line 16: Let's use `AnsiColorCode` for this since it now exists:  `.\Drawing\AnsiColorCode.cs`

Terminal.Gui/ConsoleDrivers/Net/NetDriver.cs
```diff
@@ -0,0 +1,902 @@
...
+    public bool IsWinPlatform { get; private set; }
+    public NetWinVTConsole? NetWinConsole { get; private set; }
+
+    public override void Refresh ()
+    {
+        UpdateScreen ();
+        UpdateCursor ();
+    }
+
+    public override void SendKeys (char keyChar, ConsoleKey key, bool shift, bool alt, bool control)
```
tig, line 42: This API should really be deprecated. 

It was poorly designed and is weird. Tests should use `Application.OnKeyDown()` to simulate input.

Terminal.Gui/ConsoleDrivers/Net/NetDriver.cs
```diff
@@ -0,0 +1,902 @@
...
+                                      "IL3050:Calling members annotated with 'RequiresDynamicCodeAttribute' may break functionality when AOT compiling.",
+                                      Justification = "<Pending>")]
+    private static readonly HashSet<int> ConsoleColorValues = new (
+                                                                   Enum.GetValues (typeof (ConsoleColor))
+                                                                       .OfType<ConsoleColor> ()
+                                                                       .Select (c => (int)c)
+                                                                  );
+
+    // Dictionary for mapping ConsoleColor values to the values used by System.Net.Console.
+    private static readonly Dictionary<ConsoleColor, int> colorMap = new ()
```
tig, line 496: See comment about AnsiColorCode.cs above.

tig (changes requested):

tig (changes requested):
There's a bunch of warnings that need to be cleaned up too.

E.g.

image

Terminal.Gui/ConsoleDrivers/Net/NetDriver.cs
dodexahedron: Good thought.

How about I do that when I get to NetDriver, specifically, though, which will probably be the next branch after the one I'm working on right now branched from this one?

That was just relocated code. No actual change.

Plus, it kinda looks like it shouldn't even be in that driver in the first place, but rather...I dunno... _somewhere_ else?

Terminal.Gui/ConsoleDrivers/Net/NetDriver.cs
dodexahedron: I support that wholeheartedly.

There's likely to be a fair amount of old stuff ripped out as I move along in these.

As in the other comment, though, how about I address that when I branch for that driver specifically, or if it naturally requires a heavy hand due to coupling or something?

Terminal.Gui/ConsoleDrivers/Net/NetDriver.cs
dodexahedron: Yep. That was the next place my eye immediately landed after seeing your first comment. That's at least somewhat redundant to stuff already in Color, as well.

But yeah - how about in the third round, since that's where I'm currently planning to hit NetDriver head-on.

dodexahedron:
> There's a bunch of warnings that need to be cleaned up too.
> 
> E.g.
> 
> image

Yep. Fully intend to deal with them. 👍

Similar to a PR series I did a while back and one you did some time after that, this one is, well, a series, and stuff like that might get spread across a couple branches, in my attempt to keep the individual PRs a bit more contained.

In fact, a fair number of them have already been addressed in files that are currently either in my current working branch or stashed for a future branch where they'll get popped off to be with their friends.

So, mind if, for this PR, I just backport the ones under the new Windows interop types and shore up the rest a bit more targeted, like that?

Especially since, aside from BOOL, PInvoke, ITrueFalseOperators, and CONSOLE_MODE, they're not actually new code in those new files - just relocated existing code that always had the problems in the warnings. They were just turned off.

And also some will be addressed by code getting ripped out or otherwise fundamentally altered. CONSOLE_MODE and anything related to it is one big example that comes to mind, because it's being replaced wholesale by a struct, which was always the end goal, but the enum is a convenient placeholder for the constants that doesn't take that much work to turn into a proper and intelligent type once more important bits are finished. Which happens to be in the branch I'm currently working on, where warnings in that file are gone and it's a green light up top now that it's _finished_ in my second branch.

That's going to happen to other things as well, for sure, before all of these are over.

Don't worry. I'm more anal about warnings than anyone here. That's why I keep slowly turning things up - to boil that frog. 😉

<details><summary>And the rest of this is just me explaining my thought process, my work processes around warnings/general code quality (hey - the issue is named about that!) and pointing out some commit comments, plus probably some repeated bits from frustratingly and stupidly allowing my laptop battery to die and then writing half as much as a less fun replacement.[^sorry][^youDontHaveToBeTHATRelieved]</summary>

I have a LOT more than these turned on in my local project and solution .user files. Man, you should see the miles of yellow my builds give. And that's just from the built-ins from using the "recommended" AnalysisLevel for .net 8 and c# 12. And then the sea of red squiggles in a lot of code files for stuff I have turned up to error level in my local settings, so a code file I have open will block a build while it's open if it has any, but allow the build if it's closed (that only care if it's open behavior is a setting in R# and I think also even VS itself now and I love it).

Amusingly, because of that, I actually resolved _more_ warnings in this code, at the higher analysis setting, despite the few new ones you see there. 😅 

So rest assured - I ain't ignoring them and fully intend to address any I cause and plenty I didn't cause, though I also may intentionally leave one in on occasion, if it points out something unrelated that really should be addressed.

Actually, I've also been un-suppressing some warnings in code that absolutely should not have been suppressed - some of which have a stated reason that is _significantly_ worse than a very legitimate build warning.
</details>

[^sorry]: Sorry. It was better and unfortunately my annoyance at myself wrecked the tone I meant for it to have, I think. My bad.

[^youDontHaveToBeTHATRelieved]: But hey, at least now you only have a short story to read instead of an epic.

dodexahedron:
Actually... Some of those changes can just be moved in the commit graph as I rebase it on the most recent head of v2_develop, so I'll do that to knock out a couple that fit well in this one anyway.
...And I'm sure I'll probably then end up fixing more of them over again since they're mostly pretty simple or back-porting some others. So you might get more quicker gratification than I was assuming was likely.

Gimme a bit. Dinner time. 🙂

dodexahedron:
Wait. Nevermind. That's you guys merging stuff on your branches. This is already based on head of v2_develop. 😅

I still will go ahead and reorder some of those commits, where they can be safely reordered, though. I'll probably get working on that in like an hour I imagine.

But do expect there to potentially be others in subsequent intermediate branches, since I'm trying to avoid writing stuff I'm not certain will still be there when I'm done, unless I have to a part of keeping changes smaller and safer. 🤷‍♂️

But again they are all meant to be resolved by the end of the series, for sure, so go right on ahead pointing them out if you like, and if it's not terribly disruptive to whatever I'm working on I'll probably address some right away anyway. But make it bleed in whichever ones I start marking as the end of the series.

Right now I'm thinking it's gonna be maybe 5 branches total, counting this one? We'll see as things evolve.

Quick anecdote about red squiggles:

I authored a mod for a little while in the earlier days after release of Starfield, then entire purpose of which was to deal with _spelling and grammar errors_.

The logo icon was the name of the game misspelled, with a red squiggle. 😅 

Made several hundred edits by the time I got bored with it and started having an affair with Cyberpunk  behind its back. Sshhh. Don't tell it.

dodexahedron:
Cherry-picked several commits that can go in this one.

Did a little bit more work on top of it, which resolves several warnings, but also adds some analysis-related stuff that will cause more, too.

Also fixed a bug for opening "stderr" in Windows, which isn't actually its own stream.

dodexahedron:
I think I'm done with this one for now. I went back to working on continuing work on the next branch, so this one is ready as far as I'm concerned.

Pretty big improvements and consolidation already happening with WindowsDriver/WindowsConsole, so far. 😃 

---

Oh and a minor informational side note:

I also loaded the code of this PR up in Rider on Windows and Ubuntu 24.04.1 to make sure things are still friendly to it, and to make sure tests also work in it as expected.

Went well. I didn't encounter any trouble, so looks like we're still good in Rider.

And I gotta say.... While I still don't care for Rider's overall environment, as it still feels perplexingly less capable than ReSharper somehow, it does work really well in the testing and debugging of testing.

The test runner, its output, the UI around testing, and especially coverage, profiling, and debugging of unit tests in this solution in particular do seem to work more reliably in Rider than it does in VS, on those two machines. 🤷‍♂️ 

---

(I had accidentally posted the above over on the issue yesterday. Meant for it to go here.)

dodexahedron:
Know what?

In light of some of the stuff I talked about in the issue for this, I'mma actually close this PR for right now, because I'm thinking a somewhat different approach is going to work a lot better, especially when giving consideration to abstraction and such.

I'll just open a new one when I have a useful checkpoint ready for merge.

PR #3711, opened 2024-08-31
Fixes #3695 - Add event handlers for Accept in ColorPicker text fields

### LINKED ISSUE #3695: `ColorPicker` should apply changes on `Accept` as well as `OnHasFocusChanged`
In #3627 the new `ColorPicker` has been a great test case because it's a compound view with a bunch of nested subviews.

It currently applies changes when a focused subview loses focus (via the new `OnHasFocusChanged` event (formerly `Leave`), and this is a nice touch.

However, it does not apply changes if the user presses ENTER after entering a value into one of the text fields. It should as this is what users expect. See how I did this in `CharMap`. All that needs to be done is to subscribe to the `Accept` event on the textfield and do the right thing.

@tznind if you choose to fix this before #3627 is ready, please do so as a PR to that branch as I've had to tweak CP and it's unit tests some...

tznind:
## Fixes

- Fixes #3695 

## Proposed Changes/Todos

Make Accept (default key Enter) in text boxes update color selection similarly to focus change events.

@tig I have removed the .NewValue if check as I'm not sure what it means on a focus changing event.   Presumably the new value is supposed to indicate whether it is entering or leaving? either way the behaviour seems correct regardless.

I've also removed the `-=` event handler removals.  The reason is because we are now using anonymous delegate `(s,_)=>` for both Accept and FocusChange.

I don't think they are needed anyway as the lifecycle of the view at that point is where they are Disposed anyway so no events should get raised and they will go out of scope and get cleaned up anyway.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/ColorPicker.cs
+++ b/Terminal.Gui/Views/ColorPicker.cs
@@ -63,7 +63,8 @@ public class ColorPicker : View
                     Y = y,
                     Width = textFieldWidth
                 };
-                tfValue.HasFocusChanged += UpdateSingleBarValueFromTextField;
+                tfValue.HasFocusChanged += (s,_)=> UpdateSingleBarValueFromTextField(s);
+                tfValue.Accept += (s, _)=>UpdateSingleBarValueFromTextField(s);
                 _textFields.Add (bar, tfValue);
             }
 
@@ -152,7 +153,8 @@ public class ColorPicker : View
         };
         _tfName.Autocomplete = auto;
 
-        _tfName.HasFocusChanged += UpdateValueFromName;
+        _tfName.HasFocusChanged += (s,_)=> UpdateValueFromName(s);
+        _tfName.Accept += (s, _) => UpdateValueFromName (s);
     }
 
     private void CreateTextField ()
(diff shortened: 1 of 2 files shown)


tig:
> @tig I have removed the .NewValue if check as I'm not sure what it means on a focus changing event. Presumably the new value is supposed to indicate whether it is entering or leaving? either way the behaviour seems correct regardless.

Correct. For `HasFocusChanging/HasFocusChanged` `e.NewValue == true` indicates what we used to think of as "enter".

Terminal.Gui/Views/ColorPicker.cs
```diff
@@ -63,7 +63,8 @@ public void ApplyStyleChanges ()
                     Y = y,
                     Width = textFieldWidth
                 };
-                tfValue.HasFocusChanged += UpdateSingleBarValueFromTextField;
+                tfValue.HasFocusChanged += (s,_)=> UpdateSingleBarValueFromTextField(s);
```
tig, line 66: I think you should check `!e.NewValue`. You only want to update the values on "leave", not "enter".

tig (changes requested):

Terminal.Gui/Views/ColorPicker.cs
tznind: Fixed in ff6fbc9d80f275032a2f2cdca000002081c53d48

PR #3716, opened 2024-09-01
Fixes #3715 - `Pos.Has (Type, out Pos)` -> `Pos.Has<T> (out Pos)`

### LINKED ISSUE #3715: `Pos.Has` should be generic like `Dim.Has<T>`

tig:
## Fixes

- Fixes #3715 
- `Dim.Fill` now takes a `Dim`, instead of an `int`
- Fixes `KeyBindingsDialog` to correctly size list if button shadow style is enabled.


BDisp (approved):
Nice.

Issue #2798, opened 2023-08-09
Add TrueColor support to CursesDriver


tig:
Dependent on #2797 

Add TrueColor support to CursesDriver

dodexahedron:
While diving down the CursesDriver rabbit hole to address other issues (#3287 covers some of it), I'm now rapidly leaning toward and half talking myself into going ahead and tackling a rework of it.

While that wouldn't be directly _targeted_ at enhancing it, such as adding more color support, I have a feeling that redoing it in the spirit of the other ConsoleDrivers will already implicitly bring at least a good portion of the necessary elements along with it, anyway, if not even straight up result in it just getting at least similar support to what the others have.

Someone either talk me out of that or push me over the edge! 😅

tig:
Remember: the fact Terminal.gui relies on curses for Mac/Linux support is dumb.

Curses just provides a huge amount of complexity for very little. 

If you're gonna dive in here, dive all the way. 🤪

Nutzzz:
Would using [notcurses](https://github.com/dankamongmen/notcurses) help? Though I don't think they have a C# wrapper yet...

BDisp:
When we don't need pinvokes for Windows, they may not be necessary for Linux/Mac either. Everything is interconnected with completely different systems and only by handling them is it possible to access certain functionalities. .NET does not fully provide all these features for all systems, especially on Linux/Mac, for obvious reasons, I think.

tig:
Curses (and noncurses) are not magic. 

They do what we already partially do, and need to do fully on Windows (to support VTS) anyway.

We need three things on each platform:

1. Something that outputs to the terminal.
2. Something that receives and maps keyboard input.
3. Something that receives and maps mouse input.

On Mac/Linux, leveraging some API that abstracts away low-level platform differences will be useful for 2 and 3. It may be curses/notcurses can be useful there. But given how these OS's have converged over time, I think we can use the bare minimum.

dodexahedron:
> When we don't need pinvokes for Windows, they may not be necessary for Linux/Mac either. Everything is interconnected with completely different systems and only by handling them is it possible to access certain functionalities. .NET does not fully provide all these features for all systems, especially on Linux/Mac, for obvious reasons, I think.

Spot on. Also, even Console itself is more capable in .net 8 than it was before, so we have existing code that can almost certainly be refactored to remove some pinvokes.

[...]

PR #3751, opened 2024-09-20
Fixes #3750 - `MouseEnter`/`Leave`

### LINKED ISSUE #3750: `MouseEnter/Leave` is broken for Views with subviews
The logic in `Application.OnMouseEvent` only supports raising `OnMouseEnter/Leave` when there's only two levels of nesting.

This can be reproduced in the `Mouse` Scenario by changing the `MouseDemo` class to derive from `Shortcut` instead of `Button`.

`Application.MouseEnteredView` can only hold one value. This makes no sense as any number of Views can have the mouse "over" them at the same time, when they are nested subviews.

I'm thinking through a fix now....

tig:
## Fixes

- Fixes #3750

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Application/Application.Mouse.cs
+++ b/Terminal.Gui/Application/Application.Mouse.cs
@@ -1,4 +1,10 @@
 #nullable enable
+<<<<<<< Updated upstream
+=======
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.VisualBasic.Syntax;
+
+>>>>>>> Stashed changes
 namespace Terminal.Gui;
 
 public static partial class Application // Mouse handling
@@ -116,8 +122,8 @@ public static partial class Application // Mouse handling
         UnGrabbedMouse?.Invoke (view, new (view));
     }
 
-    // Used by OnMouseEvent to track the last view that was clicked on.
-    internal static View? MouseEnteredView { get; set; }
+    // Used by OnMouseEvent to suppport MouseEnter and MouseLeave events
+    internal static List<View?> ViewsUnderMouse { get; } = new ();
 
     /// <summary>Event fired when a mouse move or click occurs. Coordinates are screen relative.</summary>
     /// <remarks>
@@ -139,17 +145,32 @@ public static partial class Application // Mouse handling
             return;
         }
 
+<<<<<<< Updated upstream
         var view = View.FindDeepestView (Current, mouseEvent.Position);
+=======
+        Stack<View?> viewsUnderMouse = View.GetViewsUnderMouse (mouseEvent.Position);
 
-        if (view is { })
+        View? deepestViewUnderMouse = viewsUnderMouse.TryPeek (out View? result) ? result : null;
+
+        if ((mouseEvent.Flags == MouseFlags.Button1Pressed
+             || mouseEvent.Flags == MouseFlags.Button2Pressed
+             || mouseEvent.Flags == MouseFlags.Button3Pressed
+             || mouseEvent.Flags == MouseFlags.Button4Pressed)
+            && Popover is { Visible: true } && !ApplicationNavigation.IsInHierarchy (Popover, deepestViewUnderMouse, includeAdornments: true))
+        {
+            Popover.Visible = false;
+        }
+>>>>>>> Stashed changes
+
+        if (deepestViewUnderMouse is { })
         {
 #if DEBUG_IDISPOSABLE
-            if (view.WasDisposed)
+            if (deepestViewUnderMouse.WasDisposed)
             {
-                throw new ObjectDisposedException (view.GetType ().FullName);
+                throw new ObjectDisposedException (deepestViewUnderMouse.GetType ().FullName);
             }
 #endif
-            mouseEvent.View = view;
+            mouseEvent.View = deepestViewUnderMouse;
         }
 
         MouseEvent?.Invoke (null, mouseEvent);
@@ -177,7 +198,7 @@ public static partial class Application // Mouse handling
                 Position = frameLoc,
                 Flags = mouseEvent.Flags,
                 ScreenPosition = mouseEvent.Position,
-                View = view ?? MouseGrabView
+                View = deepestViewUnderMouse ?? MouseGrabView
             };
 
             if ((MouseGrabView.Viewport with { Location = Point.Empty }).Contains (viewRelativeMouseEvent.Position) is false)
@@ -193,7 +214,7 @@ public static partial class Application // Mouse handling
             }
 
             // ReSharper disable once ConditionIsAlwaysTrueOrFalse
-            if (MouseGrabView is null && view is Adornment)
+            if (MouseGrabView is null && deepestViewUnderMouse is Adornment)
             {
                 // The view that grabbed the mouse has been disposed
                 return;
@@ -202,6 +223,7 @@ public static partial class Application // Mouse handling
 
         // We can combine this into the switch expression to reduce cognitive complexity even more and likely
         // avoid one or two of these checks in the process, as well.
+<<<<<<< Updated upstream
         WantContinuousButtonPressedView = view switch
                                           {
                                               { WantContinuousButtonPressed: true } => view,
@@ -224,17 +246,24 @@ public static partial class Application // Mouse handling
                 ApplicationOverlapped.MoveCurrent ((Toplevel)top);
             }
         }
+=======
+        WantContinuousButtonPressedView = deepestViewUnderMouse switch
+        {
+            { WantContinuousButtonPressed: true } => deepestViewUnderMouse,
+            _ => null
+        };
+>>>>>>> Stashed changes
 
         // May be null before the prior condition or the condition may set it as null.
         // So, the checking must be outside the prior condition.
-        if (view is null)
+        if (deepestViewUnderMouse is null)
         {
             return;
         }
 
         MouseEvent? me;
 
-        if (view is Adornment adornment)
+        if (deepestViewUnderMouse is Adornment adornment)
         {
             Point frameLoc = adornment.ScreenToFrame (mouseEvent.Position);
 
@@ -243,19 +272,19 @@ public static partial class Application // Mouse handling
                 Position = frameLoc,
                 Flags = mouseEvent.Flags,
                 ScreenPosition = mouseEvent.Position,
-                View = view
+                View = deepestViewUnderMouse
             };
         }
-        else if (view.ViewportToScreen (Rectangle.Empty with { Size = view.Viewport.Size }).Contains (mouseEvent.Position))
+        else if (deepestViewUnderMouse.ViewportToScreen (Rectangle.Empty with { Size = deepestViewUnderMouse.Viewport.Size }).Contains (mouseEvent.Position))
         {
-            Point viewportLocation = view.ScreenToViewport (mouseEvent.Position);
+            Point viewportLocation = deepestViewUnderMouse.ScreenToViewport (mouseEvent.Position);
 
             me = new ()
             {
                 Position = viewportLocation,
                 Flags = mouseEvent.Flags,
                 ScreenPosition = mouseEvent.Position,
-                View = view
+                View = deepestViewUnderMouse
             };
         }
         else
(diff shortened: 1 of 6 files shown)


BDisp (approved):
LGTM.

tig:
3XRyLlE 1

tig:
2Tjd3Zh 1

tig:
New GrayScale Theme.

Note highlight effects.

Note `CheckBox` is not Highlight enabled (yet).

dO17ivT 1

Latest API (this is Button.cs):

```cs
    private void OnHighlight (object sender, CancelEventArgs<HighlightStyle> args)
    {
        if (!HighlightStyle.HasFlag (HighlightStyle.Hover))
        {
            return;
        }

        ColorScheme cs = ColorScheme;

        if (cs is null)
        {
            cs = new ();
        }

        if (args.NewValue.HasFlag (HighlightStyle.Hover))
        {
            _savedColorScheme = cs;

            ColorScheme = ColorScheme.GetHighlightColorScheme ();
        }
        else
        {
            ColorScheme = _savedColorScheme;
            _savedColorScheme = default;
        }
    }
```

BDisp:
I think the shadow button shouldn't move to the right, like as it isn't moving to the bottom, when it's pressed. Only the shadow effect should be hidden when the button is pressed. In my opinion te button ias always in the same place and only the light is placed on top/left side chaining a shadow on the bottom/right. If the light is projected in others directions the shadows will be propagate accordingly.

Terminal.Gui/Drawing/Color.Formatting.cs
```diff
@@ -585,11 +589,14 @@ public static bool TryParse (ReadOnlySpan<byte> utf8Text, IFormatProvider? provi
...
-                   $"#{R:X2}{G:X2}{B:X2}";
+        string? name = ColorStrings.GetW3CColorName (this);
+
+        if (name is { })
+        {
+            return name;
+        }
+
+        return $"#{R:X2}{G:X2}{B:X2}";
     }
```
BDisp, line 600: These changes forces the the color key to use the named color. e.g. the color key "#1E90FF" represents the color name "DodgerBlue". The changes forces the color key to be also "DodgerBlue". I don't see any problems with this but if it's what you want then you also need to make the necessary changes in the Strings.resx file because there isn't no direct parse from color name to color key that represent thar start with #.

BDisp (commented):
Starting reviewing this wit some observations. If necessary I'll will add more.

tig:
> I think the shadow button shouldn't move to the right, like as it isn't moving to the bottom, when it's pressed. Only the shadow effect should be hidden when the button is pressed. In my opinion the button is always in the same place and only the light is placed on top/left side chaining a shadow on the bottom/right. If the light is projected in others directions the shadows will be propagate accordingly.

My opinion is the current effect looks great.

This was discussed here:

https://github.com/gui-cs/Terminal.Gui/issues/2144#issuecomment-2131348759

And the code has this comment:

image

It would be trivial to add a CM property to control this, but we'd still need to have a default. 

I enabled the effect by default in this PR so that it was in everone's face and we could get good debate going.

The right place to further opine on it is here:

https://github.com/gui-cs/Terminal.Gui/issues/3493

Terminal.Gui/Drawing/Color.Formatting.cs
BDisp: I agree. Then please test if this changed unit tests pass:

```cs
#nullable enable

using System.Collections;
using System.Globalization;
using System.Resources;
using Terminal.Gui.Resources;

namespace Terminal.Gui.ResourcesTests;

public class ResourceManagerTests
{
    private const string DODGER_BLUE_COLOR_KEY = "DodgerBlue";
    private const string DODGER_BLUE_COLOR_NAME = "DodgerBlue";
    private const string NO_NAMED_COLOR_KEY = "#1E80FF";
    private const string NO_NAMED_COLOR_NAME = "#1E80FF";
    private const string EXISTENT_CULTURE = "pt-PT";
    private const string NO_EXISTENT_CULTURE = "de-DE";
    private const string NO_EXISTENT_KEY = "blabla";
    private const string NO_TRANSLATED_KEY = "fdDeleteTitle";
    private const string NO_TRANSLATED_VALUE = "Delete {0}";
    private const string TRANSLATED_KEY = "ctxSelectAll";
    private const string TRANSLATED_VALUE = "_Selecionar Tudo";
    private static readonly string _stringsNoTranslatedKey = Strings.fdDeleteTitle;
    private static readonly string _stringsTranslatedKey = Strings.ctxSelectAll;
    private static readonly CultureInfo _savedCulture = CultureInfo.CurrentCulture;
    private static readonly CultureInfo _savedUICulture = CultureInfo.CurrentUICulture;

    [Fact]
    public void GetObject_Does_Not_Overflows_If_Key_Does_Not_Exist () { Assert.Null (GlobalResources.GetObject (NO_EXISTENT_KEY, CultureInfo.CurrentCulture)); }

    [Fact]
    public void GetObject_FallBack_To_Default_For_No_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (NO_EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (NO_EXISTENT_CULTURE);

        Assert.Equal (NO_TRANSLATED_VALUE, GlobalResources.GetObject (NO_TRANSLATED_KEY, CultureInfo.CurrentCulture));

        RestoreCurrentCultures ();
    }

    [Fact]
    public void GetObject_FallBack_To_Default_For_Not_Translated_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (NO_EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (NO_EXISTENT_CULTURE);

        Assert.Equal (NO_TRANSLATED_VALUE, GlobalResources.GetObject (NO_TRANSLATED_KEY, CultureInfo.CurrentCulture));

        RestoreCurrentCultures ();
    }

    [Fact]
    public void GetResourceSet_FallBack_To_Default_For_No_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (NO_EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (NO_EXISTENT_CULTURE);

        // W3CColors.GetColorNames also calls ColorStrings.GetW3CColorNames
        string [] colorNames = new W3CColors ().GetColorNames ().ToArray ();
        Assert.Contains (DODGER_BLUE_COLOR_NAME, colorNames);
        Assert.DoesNotContain (NO_TRANSLATED_VALUE, colorNames);

        RestoreCurrentCultures ();
    }

    [Fact]
    public void GetResourceSet_FallBack_To_Default_For_Not_Translated_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (EXISTENT_CULTURE);

        // These aren't already translated
        // ColorStrings.GetW3CColorNames method uses GetResourceSet method to retrieve color names
        IEnumerable<string> colorNames = ColorStrings.GetW3CColorNames ();
        Assert.NotEmpty (colorNames);

        // W3CColors.GetColorNames also calls ColorStrings.GetW3CColorNames
        colorNames = new W3CColors ().GetColorNames ().ToArray ();
        Assert.Contains (DODGER_BLUE_COLOR_NAME, colorNames);
        Assert.DoesNotContain (NO_TRANSLATED_VALUE, colorNames);

        // ColorStrings.TryParseW3CColorName method uses GetResourceSet method to retrieve a color value
        Assert.True (ColorStrings.TryParseW3CColorName (DODGER_BLUE_COLOR_NAME, out Color color));
        Assert.Equal (DODGER_BLUE_COLOR_KEY, color.ToString ());

        // W3CColors.GetColorNames also calls ColorStrings.GetW3CColorNames for no-named colors
        colorNames = new W3CColors ().GetColorNames ().ToArray ();
        Assert.DoesNotContain (NO_NAMED_COLOR_NAME, colorNames);

        // ColorStrings.TryParseW3CColorName method uses GetResourceSet method to retrieve a color value for no-named colors
        Assert.True (ColorStrings.TryParseW3CColorName (NO_NAMED_COLOR_NAME, out color));
        Assert.Equal (NO_NAMED_COLOR_KEY, color.ToString ());

        RestoreCurrentCultures ();
    }

    [Fact]
    public void GetResourceSet_With_Filter_Does_Not_Overflows_If_Key_Does_Not_Exist ()
    {
        ResourceSet value = GlobalResources.GetResourceSet (CultureInfo.CurrentCulture, true, true, d => (string)d.Key == NO_EXISTENT_KEY)!;
        Assert.NotNull (value);
        Assert.Empty (value.Cast<DictionaryEntry> ());
    }

    [Fact]
    public void GetResourceSet_Without_Filter_Does_Not_Overflows_If_Key_Does_Not_Exist ()
    {
        ResourceSet value = GlobalResources.GetResourceSet (CultureInfo.CurrentCulture, true, true)!;
        Assert.NotNull (value);
        Assert.NotEmpty (value.Cast<DictionaryEntry> ());
    }

    [Fact]
    public void GetString_Does_Not_Overflows_If_Key_Does_Not_Exist () { Assert.Null (GlobalResources.GetString (NO_EXISTENT_KEY, CultureInfo.CurrentCulture)); }

    [Fact]
    public void GetString_FallBack_To_Default_For_No_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (NO_EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (NO_EXISTENT_CULTURE);

        Assert.Equal (NO_TRANSLATED_VALUE, GlobalResources.GetString (NO_TRANSLATED_KEY, CultureInfo.CurrentCulture));

        RestoreCurrentCultures ();
    }

    [Fact]
    public void GetString_FallBack_To_Default_For_Not_Translated_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (EXISTENT_CULTURE);

        // This is really already translated
        Assert.Equal (TRANSLATED_VALUE, GlobalResources.GetString (TRANSLATED_KEY, CultureInfo.CurrentCulture));

        // These aren't already translated
        // Calling Strings.fdDeleteBody return always the invariant culture
        Assert.Equal (NO_TRANSLATED_VALUE, GlobalResources.GetString (NO_TRANSLATED_KEY, CultureInfo.CurrentCulture));

        RestoreCurrentCultures ();
    }

    [Fact]
    public void Strings_Always_FallBack_To_Default_For_No_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (NO_EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (NO_EXISTENT_CULTURE);

        Assert.Equal (NO_TRANSLATED_VALUE, _stringsNoTranslatedKey);

        RestoreCurrentCultures ();
    }

    [Fact]
    public void Strings_Always_FallBack_To_Default_For_Not_Translated_Existent_Culture_File ()
    {
        CultureInfo.CurrentCulture = new (EXISTENT_CULTURE);
        CultureInfo.CurrentUICulture = new (EXISTENT_CULTURE);

        // This is really already translated
        Assert.Equal (TRANSLATED_VALUE, _stringsTranslatedKey);

        // This isn't already translated
        Assert.Equal (NO_TRANSLATED_VALUE, _stringsNoTranslatedKey);

        RestoreCurrentCultures ();
    }

    private void RestoreCurrentCultures ()
    {
        CultureInfo.CurrentCulture = _savedCulture;
        CultureInfo.CurrentUICulture = _savedUICulture;
    }
}

```

BDisp: Changing this in the ColorStrings.cs fixes this issue:

```cs
    public static bool TryParseW3CColorName (string name, out Color color)
    {
        // Iterate through all resource entries to find the matching color name
        foreach (DictionaryEntry entry in GlobalResources.GetResourceSet (CultureInfo.CurrentUICulture, true, true)!)
        {
            if (entry.Value is string colorName && colorName.Equals (name, StringComparison.OrdinalIgnoreCase))
            {
                // Parse the key to extract the color components
                string key = entry.Key.ToString () ?? string.Empty;

                if (CheckTryParseW3CColorName (out color, key))
                {
                    return true;
                }
            }
        }

        return CheckTryParseW3CColorName (out color, name);
    }

    private static bool CheckTryParseW3CColorName (out Color color, string key)
    {
        if (key.StartsWith ("#") && key.Length == 7)
        {
            if (int.TryParse (key.Substring (1, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int r)
                && int.TryParse (key.Substring (3, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int g)
                && int.TryParse (key.Substring (5, 2), NumberStyles.HexNumber, CultureInfo.InvariantCulture, out int b))
            {
                color = new (r, g, b);

                return true;
            }
        }

        color = default (Color);

        return false;
    }

```

tig: Got it. Thank you!

tig:
> I already noted that unit tests are extremely slow in this PR.

Fixed.

BDisp:
These two images bellow refer to the same inline data in the `GetKeyChar_Shifted_Char_From_UnShifted_Char` unit tests. They show different bitwise masks. That why it's showing as not run tests. I think the reason for this is how the bits are set because the `KeyCode` is a `enum` using the `[Flags]` attribute. He may represent a combination of `KeyCode` values but mustn't have a new variant value. I think this is very strange and that's why the `Test Explorer` sees it as a new unit test. I commented the others InlineData to only show the one below.

`[InlineData ('9', ')', (KeyCode)')' | KeyCode.ShiftMask)]`

image

image

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -374,7 +374,7 @@ internal override MainLoop Init ()
                                                            );
         }
 
-        CurrentAttribute = new Attribute (ColorName.White, ColorName.Black);
+        CurrentAttribute = new Attribute (ColorName16.White, ColorName16.Black);
```
BDisp, line 377: I think you should standardize all this to use the Color instead of Colorname16. If `CursesDriver` supports true color some day, it will be more easy to deal with that.

`CurrentAttribute = new Attribute (Color.White, Color.Black);`

BDisp (approved):

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
tig: I agree, but the change you are looking at happened because I renamed ColorName to ColorName16, not because I directly changed CursesDriver.

I'd rather leave this as-is and have this get addressed in whatever PR fixes https://github.com/gui-cs/Terminal.Gui/issues/2798

PR #3756, opened 2024-09-24
Fixes some `ConfigurationManager` issues

tig:
## Fixes

- Fixes a bunch of latent CM related issues I found while stress-testing CM.

## Proposed Changes/Todos

- [x] File watcher is sometimes too quick and tries to read the file before the system has closed it.
- [x] Color serialization had some bugs.
- [x] Key deserialization should work with an separator

## Pull Request checklist:


tig:
@BDisp - Why does `Test_SerializableConfigurationProperty_Types_Added_To_JsonSerializerContext` explicitly prevent `Rune` properties from working?

Is there some reason I can't do what I've done so far in this PR?

BDisp:
> @BDisp - Why does `Test_SerializableConfigurationProperty_Types_Added_To_JsonSerializerContext` explicitly prevent `Rune` properties from working?
> 
> Is there some reason I can't do what I've done so far in this PR?

That was due the `Native AOT` complaining because `RuneJsonConverter` and `KeyJsonConverter` are already being initialized in the `JsonSerializerOptions _serializerOptions`. You don't need to add [JsonConverter(typeof(RuneJsonConverter))] on all properties. You already defined that the `Rune` struct is serializable in the `SourceGenerationContext` class. You may wish to add to the `Rune` properties the `SerializableConfigurationProperty` scope you want.

tig:
Hmmmm. 

What I found was the Rune properties were being serialized as ints and couldn't use a "<glyph>" in the json.

tig:
@BDisp you commented out the code in 

image

```
Commit 325180ae48396138ae8675f7a424b9f3307c36d5
Author: BDisp <bd.bdisp@gmail.com>
Date: Thursday, July 6, 2023 11:49 AM
Parent: e02fa1b1

Fixes #2680. Make the TextView API more extensible. (#2682)

* Fixes #2680. Make the TextView API more extensible.

* Remove unnecessary using.

* Add GetLine method.

* Change RuneCell Attribute property to ColorScheme property.

* Add LoadRuneCells method and unit test.

* Add helper method to set all the Colors.ColorSchemes with the same attribute.

* Change RuneCell to class.

* Add IEquatable<RuneCell> interface.

* Fix unit test.

* Still fixing unit test.

* Fixes #2688. ReadOnly TextView's broken scrolling after version update.

* keyModifiers must be reset after key up was been processed.

* Trying fix server unit test error.

* Prevents throw an exception if RuneCell is null.

* Still trying fix this unit test.

* Cleaning code.

* Fix when the RuneCell is null.

* Fix throwing an exception if current column position is greater than the line length.

* Fixes #2689. Autocomplete doesn't popup after typing the first character.

* Fix Used on TextField.

* Always use the original ColorScheme if RuneCell.ColorScheme is null.

* Fix Used on TextView.

* Add RuneCellEventArgs and draw colors events.

* Add two more samples to the scenario.

* Fix a bug which was causing unit tests with ColorScheme fail.

* Fix a issue when WordWrap is true by always loading the old text.

* Improves debugging in RuneCell.

* WordWrap is now preserving the ColorScheme of the unwrapped lines.

* Simplifying unit test.

* Ensures the foreground and background colors are never the same if Used is false.

* Remove nullable from the parameter.

* Merge syntax highlighting of quotes and keywords together

* Add IdxRow property into the RuneCellEventArgs.

* Fix pos calculation on windows
(where newline in Text is \r\n not \n)

* Fix events not being cleared when toggling samples.

* Change Undo and Redo to a public method.

* Changes some methods names to be more explicit.

* Call OnContentsChanged on needed methods and fix some more bugs.

* Adds InheritsPreviousColorScheme to allow LoadRuneCells uses personalized color schemes.

* Serializes and deserializes RuneCell to a .rce extension file.

* Prevents throwing if column is bigger than the line.

* Avoids create a color attribute without one of the foreground or background values. In Linux using -1 throws an exception.

* Replace SetAllAttributesBasedOn method with a ColorScheme constructor.

* Move RuneCell string extensions to TextView.cs

* Reverted parameter name from cell to rune.

* Change Row to UnwrappedPosition which provide the real unwrapped text position within the Col.

* Add brackets to Undo and Redo methods.

* Replace all the LoadXXX with Load and rely on the param type to differentiate.

* Open a file inside a using.

* Proves that the events run twice for WordWrap disabled and the enabled.

* Remove GetColumns extension for RuneCell.

* Add braces to Undo an Redo.

* Change comment.

* Add braces.

* Delete remarks tag.

* Explaining used color and ProcessInheritsPreviousColorScheme.

* Fix comment.

* Created a RuneCellTests.cs file.

* Rename to StringToLinesOfRuneCells.

* Make ToRuneCells private.

---------

Co-authored-by: Thomas <tznind@dundee.ac.uk>
Co-authored-by: Thomas Nind <31306100+tznind@users.noreply.github.com>
```

BDisp:
> This broke the behavior I had designed in to ensure serializing a Rune would output the glyph for readability.
> 
> I've reverted that, cleaned it up, and added some unit tests.

I did comment out that code when I created the `RuneCell` class. For some reason that was give me wrong values at the time.

UnitTests/Configuration/SerializableConfigurationPropertyTests.cs
```diff
@@ -46,8 +46,8 @@ public void Test_SerializableConfigurationProperty_Types_Added_To_JsonSerializer
 
         // Ensure no property has the generic JsonStringEnumConverter<>
         EnsureNoSpecifiedConverters (properties, new [] { typeof (JsonStringEnumConverter<>) });
-        // Ensure no property has the type RuneJsonConverter
-        EnsureNoSpecifiedConverters (properties, new [] { typeof (RuneJsonConverter) });
+        /// Ensure no property has the type RuneJsonConverter
+        //EnsureNoSpecifiedConverters (properties, new [] { typeof (RuneJsonConverter) });
```
BDisp, line 50: Don't forget to reenable this.

Terminal.Gui/Configuration/SettingsScope.cs
```diff
@@ -1,6 +1,8 @@
 ﻿#nullable enable
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Net.Security;
```
BDisp, line 5: These `using` can be removed.

BDisp:
@tig can you please fix the typo in the `NativeAot launchSettings.json` like below in this PR? `NativeAot` is working well with this PR. Thanks.

image

Terminal.Gui/Configuration/SourceGenerationContext.cs
```diff
@@ -20,5 +20,6 @@ namespace Terminal.Gui;
 [JsonSerializable (typeof (Dictionary<ColorName16, string>))]
 [JsonSerializable (typeof (Dictionary<string, ThemeScope>))]
 [JsonSerializable (typeof (Dictionary<string, ColorScheme>))]
+[JsonSerializable (typeof (Rune))]
```
BDisp, line 23: I think you don't need this here because what you are serialize/deserialize are the `GlyphDefinitions` class and the `[JsonSerializable (typeof (GlyphDefinitions))]` is already included.

tig: `Key.Separator` is a Rune.

BDisp: > `Key.Separator` is a Rune.

I know but there was already added `[JsonSerializable (typeof (Key))]` in the `SourceGenerationContext` class, which mean that will use `Key` for serialize/deserialize and the `RuneJsonConverter` to the conversion. Try without the `[JsonSerializable (typeof (Rune))]` and see what happens. If it fail then add it again.

PR #3768, opened 2024-09-30
Fixes #3767. Allowing any driver to request ANSI escape sequence with immediate response.

### LINKED ISSUE #3767: Allowing any driver to request ANSI escape sequence with immediate response.
It can used on any driver that can interpret ANSI escape sequence, returning immediate response or error.

BDisp:
## Fixes

- Fixes #3767

## Proposed Changes/Todos

- [x] Implemented the ExecuteAnsiRequest method which can be used by any driver.
- [x] Try request on any driver, on Windows, Linux or macOS and it will work.

## Pull Request checklist:


BDisp:
> To me his feels really brittle. Its what I was trying to explain in other thread. I can see its possible just dangerous:
> 
> * We force a 200 ms pause every time we emit an request

This is a minor problem because we are not requesting ANSI escape sequences all the time, like the one that are setting the colors when are writing to the console.

> * We don't check that the response matches the request (based on its terminator)

My bad, sorry. I created the `EscSeqUtils` and the `EscSeqReq` where I implemented the terminator from the requests reply. I was forgetting that. Thanks for show me that I was wrong.

> * The while loop just drains the console which could over-read. We should be stopping when we hit the expected terminator for the command e.g. 'c' for DAR

Done. Thanks.

> * Regardless of the state of the driver, we turn on mouse events at the end.  So this method has a byproduct of turning on mouse (even if it was off before method called)
> 
You are right. Thanks for call my attention. I added a flag into the `NetDriver` and `CursesDriver` that inform if the mouse moving report was enable or not.

> I still think we should update the existing input processing loop(s) e.g. `ProcessInputQueue` - instead of writing a bespoke process.
> 
But the immediate reply will still not prevail. When I found the current solution I never like how it's working because in some situations I wanted the response in the request method, to do some actions that depended on the response, but without success. I found it useless.

> I think we could make a blocking public facing method even with that approach. Just need to think about it some more. Maybe we can use async and user can just call `.Result` if they want to wait.

I think it won't block because I first stopped the mouse move reports which will decrease significantly the number of chars to read.
I tried the async approach but was give me unexpected results because we need to get the request replies as soo as possible.

You can use your approach as well because my PR doesn't break it, only changed from string to the class I created but still use the same behavior. The only disadvantage is that your approach only work with the `NetDriver`.

If you want to test put these line in the `Init` method of all drivers the below lines:

```cs
        var response1 = AnsiEscapeSequenceRequest.ExecuteAnsiRequest (EscSeqUtils.CSI_SendDeviceAttributes);
        var response2 = AnsiEscapeSequenceRequest.ExecuteAnsiRequest (EscSeqUtils.CSI_RequestCursorPositionReport);
        var request = EscSeqUtils.CSI_RequestCursorPositionReport;
        AnsiEscapeSequenceResponse response3 = null;
        request.ResponseReceived += (s, e) => response3 = e;
        var response4 = AnsiEscapeSequenceRequest.ExecuteAnsiRequest (request);
        Debug.Assert (response3 == response4);

```

tznind:
What is Value? I cant understand the comment.  My example has 61...

> The value expected in the response e.g.<cref>EscSeqUtils.CSI_ReportTerminalSizeInChars.Value</cref> which will have a 't' as terminator but also other different request may return the same terminator with a different value.

image

Can we also please have a helper property called `Successful` so you don't have to eg. null or whitespace the error field.

tznind:
I can confirm that this works and have added it into a new branch of the main sixel branch I am working on.

See https://github.com/tznind/gui.cs/pull/169

BDisp:
> What is Value? I cant understand the comment. My example has 61...
> 
> > The value expected in the response e.g.EscSeqUtils.CSI_ReportTerminalSizeInChars.Value which will have a 't' as terminator but also other different request may return the same terminator with a different value.
> 

Value is the number that apear at the beginning of the reply after the CSI. See https://terminalguide.namepad.de/seq/csi_st-18/ and https://terminalguide.namepad.de/seq/csi_st-15/. Both requests use the 't' as terminator and the reply also returns 't' as terminator. What distinguish each other are the Value. The prior returns 8 and the last returns 5.

> image
> 
> Can we also please have a helper property called `Successful` so you don't have to eg. null or whitespace the error field.

It's not enough to check the error with `string.IsNullOrEmpty`? If you really prefer the `Successful` property I can add it.

tznind:
Yes please either add `Successful` or have the method return bool and use `out` like `TryParse` style methods do.

I started with this ugliness:

```

        if (string.IsNullOrWhiteSpace (darResponse.Error) && !string.IsNullOrWhiteSpace (darResponse.Response))
        {
            // it worked
        }
        
```

tznind:
@tig please can you take a look at this when you get the chance?

Currently I am working on 2 sixel branches (one with this merged in). If its going in right direction I will just collapse and work on only 1 branch.

BDisp:
@tznind done in https://github.com/gui-cs/Terminal.Gui/pull/3768/commits/dfedcd88e19da7551ab3d8665fd7309d54108491. Let me know if it's need to do something more. Thanks.

tznind:
Cool thanks, although I meant for tryparse as an example of the pattern only.  It should probably still be same name e.g. TryExecuteAnsi...

BDisp:
> Cool thanks, although I meant for tryparse as an example of the pattern only. It should probably still be same name e.g. TryExecuteAnsi...

Ops I misunderstood. Changed on https://github.com/gui-cs/Terminal.Gui/pull/3768/commits/31dbae779125cefc4cfd18a09c7e09a24f6c8a0f.

tznind:
Got a strange error in `sixel-encoder-is-supported` in visual studio terminal.  When I open Images scenario and `SixelDetector` runs its ansi queries it seems to leave something bad in the input buffer that then crashes the windows driver:

```
System.ArgumentException
  HResult=0x80070057
  Message=Invalid KeyCode: Space, C is invalid. (Parameter 'value')
  Source=Terminal.Gui
  StackTrace:
   at Terminal.Gui.Key.set_KeyCode(KeyCode value) in D:\Repos\temp\gui.cs\Terminal.Gui\Input\Key.cs:line 214
   at Terminal.Gui.Key..ctor(KeyCode k) in D:\Repos\temp\gui.cs\Terminal.Gui\Input\Key.cs:line 79
   at Terminal.Gui.WindowsDriver.ProcessInput(InputRecord inputEvent) in D:\Repos\temp\gui.cs\Terminal.Gui\ConsoleDrivers\WindowsDriver.cs:line 1497
   at Terminal.Gui.WindowsMainLoop.Terminal.Gui.IMainLoopDriver.Iteration() in D:\Repos\temp\gui.cs\Terminal.Gui\ConsoleDrivers\WindowsDriver.cs:line 2249
   at Terminal.Gui.MainLoop.RunIteration() in D:\Repos\temp\gui.cs\Terminal.Gui\Application\MainLoop.cs:line 273
   at Terminal.Gui.Application.RunIteration(RunState& state, Boolean& firstIteration) in D:\Repos\temp\gui.cs\Terminal.Gui\Application\Application.Run.cs:line 528
   at Terminal.Gui.Application.RunLoop(RunState state) in D:\Repos\temp\gui.cs\Terminal.Gui\Application\Application.Run.cs:line 502
   at Terminal.Gui.Application.Run(Toplevel view, Func`2 errorHandler) in D:\Repos\temp\gui.cs\Terminal.Gui\Application\Application.Run.cs:line 383
   at UICatalog.Scenarios.Images.Main() in D:\Repos\temp\gui.cs\UICatalog\Scenarios\Images.cs:line 149
   at UICatalog.UICatalogApp.UICatalogMain(Options options) in D:\Repos\temp\gui.cs\UICatalog\UICatalog.cs:line 348
   at UICatalog.UICatalogApp.Main(String[] args) in D:\Repos\temp\gui.cs\UICatalog\UICatalog.cs:line 171
```

BDisp:
I have a fix this. But I need your opinion, if the request terminator is empty and the response is successfully, it's need to write to the error property "Terminator request is empty." or ignore it? If I write to the error the response will not return `true`. I think it's a good practice to include the terminator in the request, for the `Console.ReadKey (true)` breaks when the terminator is found. The advantage for this is to avoid reading several requests at once return an invalid response.

image

BDisp:
Terminator request is mandatory to stopping read when it's found.

tznind:
Crash is not because of lack of terminator.  You can see it is set.

Issue seems to be just that there is something left in buffer after reading in ANSII request class (Space+C)

asciicrash
_Open Images crashes after sending DAR_

tznind:
I think maybe we are terminating our ASCII processing on the 'key down' of the response terminator and the 'key up' is still coming through from driver?

tznind:
This code is already quite complicated and coupled to the specific drivers e.g. switch statement on Driver Type.

Maybe we can reuse the existing message pumps that the drivers are already running and move the new code to a Parser pattern? Each driver would then have its own pump but reuse something like this.

We wouldn't stop the pumps we would just add an extra 'filtering' stage to the existing input processing pipeline?

```csharp
/// <summary>
/// Driver agnostic parser that you can stream console output to as part of normal
/// input processing. Spots ansi escape sequenes as they pass through stream and
/// matches them to the current outstanding <see cref="AnsiEscapeSequenceRequest"/>
/// </summary>
public class AnsiRequestParser ()
{
    [CanBeNull]
    private AnsiEscapeSequenceRequest currentRequest;
    StringBuilder currentResponse = new StringBuilder ();
    private bool areInEscapeSequence = false;

    Queue<AnsiEscapeSequenceRequest> outstandingRequests = new Queue<AnsiEscapeSequenceRequest> ();


    /// <summary>
    /// Sends the given <paramref name="request"/> to the terminal or queues
    /// it if there is already an outstanding request.
    /// </summary>
    /// <param name="request"></param>
    public void QueueRequest (AnsiEscapeSequenceRequest request)
    {
        if (currentRequest == null)
        {
            SendRequest (request);
        }
        else
        {
            outstandingRequests.Enqueue (request);
        }
    }

    private void SendRequest (AnsiEscapeSequenceRequest request)
    {
        currentRequest = request;
        Console.Write ("... the request");
    }

    /// <summary>
    /// Takes the given key read from console input stream.  Returns true
    /// if input should be considered 'handled' (is part of an ongoing
    /// ANSI response code)
    /// </summary>
    /// <param name="key"></param>
    /// <returns></returns>
    public bool Process (char key)
    {
        // Existing code in this PR for parsing


        // If completes request
            currentRequest = null;
            // send next from queue
            return true;


        // We have not read an Esc so we are not in a response
        return false;
    }
}
```

tznind:
> Do you are using the updated branch after the latest fixes? When I said the crash is due the lack of terminator, I referred the one that is on the request and not in the response.

Have tried and get no difference in behaviour.

I think you are right and that it is the mixing of Console `Console.ReadKey` and win32 API.

> How can I reproduce what you are facing?

Not sure... this is why I was wanting to go for the parser approach as it is much easier to test.  And each driver can concern itself with its own message pump oddities.

Terminal.Gui/ConsoleDrivers/AnsiEscapeSequence/AnsiEscapeSequenceRequest.cs
```diff
@@ -0,0 +1,197 @@
...
+        var error = new StringBuilder ();
+        var savedIsReportingMouseMoves = false;
+        NetDriver? netDriver = null;
+        var values = new string? [] { null };
+
+        try
+        {
+            switch (Application.Driver)
+            {
+                case NetDriver:
```
tig, line 66: Please refactor this such that this driver-dependent code is in the related driver. This is really poor encpasulation.

BDisp: Done.

tznind:
Get an issue running this code with the new Windows Terminal Preview (the one that has sixel support).  Seems the response you read is just full of `\0`.

image

BDisp:
In which driver? I think I know why with my last refactor.

tznind:
> In which driver? I think I know why with my last refactor.

WindowsDriver.

It works with Windows Terminal just not the preview

BDisp:
I really don't know if we can rely on Windows Terminal Preview for now. See below. It's a regression or it's some workaround used in the current WT version that now was fixed on the WT Preview or vice versa?

WindowsTerminal_AF19s7lBSL

BDisp:
> Get an issue running this code with the new Windows Terminal Preview (the one that has sixel support). Seems the response you read is just full of `\0`.
> 
> I have tested in both [1.22.2702.0](https://github.com/microsoft/terminal/releases/tag/v1.22.2702.0) and 1.22.2362.0

The cause for this not been working is because the `Console.KeyAvailable` always returns false and thus don't read the response.

BDisp:
> I really don't know if we can rely on Windows Terminal Preview for now. See below. It's a regression or it's some workaround used in the current WT version that now was fixed on the WT Preview or vice versa? This is only happening in the `WindowsDriver`.
> 
> WindowsTerminal_AF19s7lBSL

I already fixed this in this commit https://github.com/gui-cs/Terminal.Gui/pull/3768/commits/91d31bd1c48f999cb25f1681cef6afb6915c632d.

BDisp:
The only driver that doesn't work well with sixel is the `NetDriver` because it use the same std-in and std-out and when I press Esc to exit the sixel scenario it doesn't exit. Since the `WindowsDriver` and `CursesDriver` they use their own std-in and std-out, allows the `Console.ReadKey` can read the input separately.

BDisp:
> The only driver that doesn't work well with sixel is the `NetDriver` because it use the same std-in and std-out and when I press Esc to exit the sixel scenario it doesn't exit. Since the `WindowsDriver` and `CursesDriver` they use their own std-in and std-out, allows the `Console.ReadKey` can read the input separately.

This issue seems to be fixed in https://github.com/gui-cs/Terminal.Gui/pull/3768/commits/884011e99cdb76fe330af2d80bf3ca1343350dee.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -608,6 +613,66 @@ public void OnMouseEvent (MouseEventArgs a)
...
 
+    /// <summary>
+    ///     Provide handling for the terminal start reporting mouse events.
+    /// </summary>
+    public abstract void StartReportingMouseMoves ();
+
+    /// <summary>
+    ///     Provide handling for the terminal stop reporting mouse events.
+    /// </summary>
+    public abstract void StopReportingMouseMoves ();
```
tznind, line 624: Its called `StopReportingMouseMoves` but really what we want is for the driver to ensure nothing comes over std in right? Presume even things like resize console or console gains focus will also be stopped for duration of this ANSI request class lifetime?

If so should this just be called `Suspend` or something? And have it full stop everything?

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -608,6 +613,66 @@ public void OnMouseEvent (MouseEventArgs a)
...
+
+    internal bool WriteAnsiRequestDefault (string ansi)
+    {
+        try
+        {
+            Console.Out.Write (ansi);
+            Console.Out.Flush (); // Ensure the request is sent
+
+            // Read the response from stdin (response should come back as input)
+            Thread.Sleep (100); // Allow time for the terminal to respond
```
tznind, line 641: Don't think we should have a thread sleep in a ConsoleDriver, this seems more related to the ansi processing.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -608,6 +613,66 @@ public void OnMouseEvent (MouseEventArgs a)
...
+    public abstract void StopReportingMouseMoves ();
+
+    /// <summary>
+    ///     Provide handling for the terminal write ANSI escape sequence request.
+    /// </summary>
+    /// <param name="ansiRequest">The <see cref="AnsiEscapeSequenceRequest"/> object.</param>
+    /// <returns>The request response.</returns>
+    public abstract string WriteAnsiRequest (AnsiEscapeSequenceRequest ansiRequest);
+
+    internal bool WriteAnsiRequestDefault (string ansi)
```
tznind, line 633: Can we call this WriteRaw or something? it is a method on console driver that is purely writing out the string given to std out.  That could wind up being more generic than just ansi requests e.g. I Could use it in sixel to move the draw code up to the base class.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
```diff
@@ -134,7 +134,7 @@ public bool WriteToConsole (Size size, ExtendedCharInfo [] charInfoBuffer, Coord
         return result;
     }
 
-    public bool WriteANSI (string ansi)
+    internal bool WriteANSI (string ansi)
```
tznind, line 137: This method sounds a lot like `internal bool WriteAnsiRequestDefault (string ansi)` whats the difference?

tznind:
@BDisp / @tig / @dodexahedron do you want me to continue looking at processing ANSI requests without suspending the input stream (see #3791)? 

There are 2 approaches here:

- The approach in this PR to suspend drivers and flush input while we then run the ANSI send/read (i.e. blocking), or
- Make ansi request detection a standard part of the input processing pipeline (i.e. non blocking)

I don't think there is much point in having both approaches.  I don't mind going with the blocking approach if that is the consensus - and I can close my PR.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
BDisp: I want to ensure that only mouse events are disabled over std-in but I ensure the existent one are processed before the request, because they're the biggest ansi escape sequence in std-in and take a long time to be processed. Any other events like resize or focus are processed after the request return.
Since it's only disable and enable mouse events I think the current name is correct.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
BDisp: I removed that method. But since I'm reading requests on demand we need to give to the terminal some time to process them, otherwise any key won't be available immediately.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
BDisp: I already removed that method. Now it's processed in each driver in the `WriteAnsiRequest` method, because they have different ways to handle the requests.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: I already removed the `WriteAnsiRequestDefault` method. They are different. The `WriteANSI` method call the `WindowsConsole.WriteConsole` of the kernel32.dll and the `WriteAnsiRequestDefault` was using the `Console.Out.Write`. On `WindowsDriver` we only can send requests through `WriteANSI` which is using the `WindowsConsole.WriteConsole`. If you use the `Console.Out.Write` we wont receive any response from the std-in, unless we are sending after the cleanup which restore the default console.

BDisp:
> @BDisp / @tig / @dodexahedron do you want me to continue looking at processing ANSI requests without suspending the input stream (see #3791)?
> 
Yes I want. I'm loving your work but I'm afraid that for immediate response from a request that needs to be handle in the same thread that may not exist anymore, because the scenario was closed. So, the action may run out of race. Imagine you want to update something in a scenario that is expecting a request response, but you click esc before the action is called. If you update something on a view that was already been disposed an exception will be throw. But I'm curious to know how you'll handle that.

> There are 2 approaches here:
> 
> * The approach in this PR to suspend drivers and flush input while we then run the ANSI send/read (i.e. blocking), or
> * Make ansi request detection a standard part of the input processing pipeline (i.e. non blocking)
> 
That is the behavior that most of the requests wants. I'm only suspend until the response is returned, then all events that was queued are processed.

> I don't think there is much point in having both approaches. I don't mind going with the blocking approach if that is the consensus - and I can close my PR.
> 
I prefer you don't close your PR because you may be right and I may be wrong. Maybe you find a way that could work as expected.

> There are advantages to each:
> 
> Blocking
> 
> * Immediate response from user code
> * Error handling for missed requests etc
> 
For immediate action.

> Non Blocking
> 
> * Simpler interaction with console i.e. no Thread.Sleep, Flush input etc

Nice for some requests that doesn't need immediate action, like the DAR that can be set later when the action is processed, which may not be in the same thread.

But please, continue your work. It's lovely. It may could improve the `EscSeqUtils` class.

BDisp:
@tznind please also add more requests to your scenario. You can use mine in this PR and adapt it to your PR. You can use the predefined already existents and use any new by editing the `TextField`. Do stressed tests in the scenario by constantly click the mouse on the send button and also continuously press the enter key, by press and release or maintaining it pressed. With these stressed tests I found another issues in my PR and fixed the best I know. Thanks.

Terminal.Gui/ConsoleDrivers/WindowsDriver.cs
BDisp: > If you use the Console.Out.Write we wont receive any response from the std-in, unless we are sending after the cleanup which restore the default console.

I forgot to say that this is only happening in the `Windows Terminal Preview`.

tznind:
Got it, I will create some stress tests.  Maybe I can have some numeric views for 'requests per second' or something.  That way we can really test the limits 👍 

For thismorning though I just focused on updating the steps processing breakdown in my PR description and some more unit testing for expansibility. 

Will look at stress testing later tonight hopefully

BDisp:
> Will look at stress testing later tonight hopefully

I meaning stress testing on a scenario because we can use a real console on unit tests. Thanks.

BDisp:
Another idea I'm having now is to allow the user how the get the response from the driver by an enum, like for e.g. `AnsiRequestType` with `Blocking` and `Non Blocking`. Thus the user can choose the desired type and all the drivers must provide both handlers for them. Only one can be active at a time. The `FakeDriver` can also simulate by using threads sending some requests or normal keys. Do you agree and think you can assuming this suggestion, @tznind?

tznind:
Hmn... maybe.  I think we have to wait till features are more complete to see how/if they can run side by side.

But it sounds interesting and worth exploring

tig:
Is this ready or is it dependent on the Ansi Parser work?

BDisp:
> Is this ready or is it dependent on the Ansi Parser work?

It isn't dependent of Ansi Parser, but I'm still improving the `CursesDriver` to not depending a lot of ncurses keys. It's really a refactoring that will accelerate the performance on it. Please wait a little more time, but you can review it if you didn't yet. Thanks.

tig:
Do me a favor and mark it as draft until it's ready. Thanks. 

Really good work.

BDisp:
The macOS's terminal has some limitations when it comes to true color support compared to other operating systems. The default macOS terminal (Terminal.app) only has partial support for true color. Here’s a breakdown of some macOS terminal environments and their color capabilities:

1. **Terminal.app** (default macOS terminal): It supports only a 256-color palette and does not fully support true color (24-bit colors), which can cause colors to appear as black, white, or gray.

2. **iTerm2:** This alternative terminal emulator for macOS fully supports true color and is generally a better option if you need accurate color rendering for applications that output 24-bit colors. If you need true color on macOS, I recommend trying your ncurses application in iTerm2 instead of the default Terminal.app.

3. **Environment Variables:** To use true color, ensure that the environment variable `COLORTERM` is set to `truecolor`. In ncurses, this may help ensure the proper color depth is applied if the terminal emulator supports it:
```bash
export COLORTERM=truecolor
```
Switching to iTerm2 should resolve the issue if your app relies on 24-bit color, as iTerm2 has more comprehensive support for these color depths.

BDisp:
Almost done. Only `WindowsDriver` isn't using the `EscSeqReqStatus` class and I want to normalize to be used on all drivers, except `FakeDriver` for now, maybe later with mock responses.

BDisp:
@tznind I think this is ready. If you have any difficult to adapt with your PR, let me know.  For now I let the `WindowsDriver` handling key-down and key-up because request response only is captured on key-down. Thanks.

tznind:
Can you call the public methods that are blocking `SendBlockingAnsiRequest` or something, the return is `string` so we cannot overload.  For a clearer API I think we need that distinction.

It looks like you have rewritten the curses driver substantially, does it still report the ansi escape sequences as normal? I will try and re implement the parser for curses into this new version when it is merged.

The parsers pattern is to sit immediately after the code that does read key and pre-filter those to extract the expected ansi responses while letting the remainder keys drop through (this is not 1 for 1 sometime 1 for 0 sometimes 1 for 2,3,4 etc depending on state (e.g. if it held Esc).

BDisp:
> Can you call the public methods that are blocking `SendBlockingAnsiRequest` or something, the return is `string` so we cannot overload. For a clearer API I think we need that distinction.
> 
Only to clarify that my PR doesn't blocking anymore. The request when is added to the `EscSeqRequests` it's enqueued into the `Statuses` property and write the raw string request to the console. Since the return values of our PR return differents types, I suggest to rename the `WriteAnsiRequest` abstract method to `WriteAnsiEscSeqRequests` in my PR and your named as WriteAnsiParser (you can rename for what you want) which will inform what kind of request the user may want to use. Based on the abstract method the user will call the desire type of request. But this change must me done after mine is merged and you are free to do what you think it's better to handle them. If you agree I can rename mine now, so let me know, please.

> It looks like you have rewritten the curses driver substantially, does it still report the ansi escape sequences as normal? I will try and re implement the parser for curses into this new version when it is merged.
> 
Yes of course it report ansi escape sequences as raw or possible incomplete escape sequences (see `EscSeqUtils.IncompleteCkInfos`). This is very rare but could happen, but normally the requests response will came at once in `CursesDriver` and possible with multiple requests responses. That why `EscSeqUtils` has new methods to split the string into `ConsoleKeyInfo` arrays for the `DecodeEscSeq` method retrieve the correct type of input (single key, mouse, request response, etc.).

> The parsers pattern is to sit immediately after the code that does read key and pre-filter those to extract the expected ansi responses while letting the remainder keys drop through (this is not 1 for 1 sometime 1 for 0 sometimes 1 for 2,3,4 etc depending on state (e.g. if it held Esc).

My PR always return the response by the enqueue order. For bad request that has no response I created a `retries` integer variable which will be incremented by one and when it's greater than one the it will remove the enqueued request and return an empty response. So, it only have one chance and don't take much time to be processed because it's for sure a bad response. Thus it doesn't use timers.

BDisp:
There is an issue with buffer resize on the Windows Terminal Preview version 1.22.2912.0 (see https://github.com/microsoft/terminal/issues/5094#issuecomment-2452708090).

tznind:
When you click on `Send Request' the scenario exits...


immediate-exit-on-click

BDisp:
Can you be more specific, please. I cant see which driver you're using but I already tried with all drivers and the exit issue doesn't happen.

tznind:
Windows driver using the old(ish) terminal that ships with Visual Studio

BDisp:
> Windows driver using the old(ish) terminal that ships with Visual Studio

Thanks for clarifying. I'm using the Windows Terminal Preview. I'll check but remember that conhost doesn't deal well with ANSI escape sequences and that why it not support true color. Or do you're meaning with the current stable Windows Terminal?

tznind:
> Or do you're meaning with the current stable Windows Terminal?

Not sure what version but it is old.  It looks like this on the task bar

image

tznind:
If I type while sending the ansi escape requests it drops characters periodically.

loosing-chars

BDisp:
> If I type while sending the ansi escape requests it drops characters periodically.

Thanks for reporting this. This is due to iterations not handling the keys while it's still processing others. I'll see what I can do. Does this is happen with your PR?

tznind:
No, I wrote ansiparser class specifically to hold onto all keys passed to it and return anything that isn't the response back to input stream processing.

It uses the native T type of each driver e.g. ConsoleKey so there is no stop or explicit read seperate from the normal loop in the driver.

Maybe you can take just the ansiparser class into this pr?.

BDisp:
> Maybe you can take just the ansiparser class into this pr?.

Can you please submit a PR to my branch with the ansiparser? Maybe you will handle better adapting the code. Thanks.

BDisp:
I'm realize now that the Ctrl+Shift+Home/End/PgUp/PgDown only is captured on key-up on `WindowsDriver`. I didn't know this was happening.

BDisp:
> I'm realize now that the Ctrl+Shift+Home/End/PgUp/PgDown only is captured on key-up on `WindowsDriver`. I didn't know this was happening.

Windows Terminal was using these keys internally and unbounded them solved the issue.

tznind:
With this branch I still getting exit (Esc) when trying to get response with the default terminal in Visual Studio.  Think we need to pay attention to older terminals here - if we try to send a request and an Esc goes through instead and exits users app then that is a pretty bad bug.

Cases we need to ensure work:

- older terminals
- tty / ssh / putty etc
- screen

instant-exit

BDisp:
@tznind I would like to reproduce this but this is the below what I get using the VSDebugConsole.exe.

VsDebugConsole_oHJoiNFJZC

tznind:
Wow that totally self destructed huh!  😱 

Can you try my branch out of interest,  also because you have it bomb immediately on enter scenario

BDisp:
@tznind, the missing letters when typing in TextView were not due to missed or unexecuted iterations, but rather due to some wrong decoding of some raw escape sequences. You can try typing again while the requests are running and you will see that there are practically no missing letters on all drivers. If there are, it is still due to some bug, which is sometimes very hard to detect.

BDisp:
> Wow that totally self destructed huh! 😱
> 
> Can you try my branch out of interest, also because you have it bomb immediately on enter scenario - while in contrast when i tested using your branch it only bombed on click the DAR

But this situation happen on only run the UICatalog without run the ansi escape sequence scenario. I also tried with your branch and the same situation is happening. But I think this is normal if we aren't using WT.

tznind:
>  But I think this is normal if we aren't using WT.

Wait as soon as you launch UICatalog the terminal does this (fills with escape codes)?

Which branches does it happen for you on?

- v1_release 
- v1_develop
- v2_release
- v2_develop
- ansi-parser

BDisp:
Yes in all V2 branches that are run without WT.

BDisp:
@tznind, I don't what happened but today I tried again to run against the vs console and now it work well, damn. I don't have your issue now, maybe was some bug that I already fixed.

VsDebugConsole_jhatAkyECf

tig:
Guys, I'm diving into reviewing this and I'm confused. 

Can you please explain the relationship between this PR and the ANSI Parser PR?

Is one dependent on the other?

BDisp:
> Guys, I'm diving into reviewing this and I'm confused.
> 
> Can you please explain the relationship between this PR and the ANSI Parser PR?
> 
There is no relationship between both. Only we have an idea for later the user choice one of them to send requests.

> Is one dependent on the other?

That I know of, there is no dependencies on each other, but if one is merged first the other will have much work to do to implement and adapt each other. Mine have much more driver changes which will I think will help some iterations.
The mainly diference of both is that mine PR try to get immediately the request response or none if there is a bad request and the ANSI Parser schedule the requests and wait for the response or cancel it after a timeout. So, you have to decide on how this will go on.

tznind:
These are basically 2 ways of solving the same problem

- `ansi-parser` branch adds a filter to all console input loops that picks up responses
- `escape-seq....all-drivers` allows client code to force process the input loop outside of its main iteration until it finds some input it likes (ansi code)

The first is harder to use for api of our library because it is based on events. The second is blocking so user gets answer right away but is imo harder to maintain/test and may be less stable.

@BDisp please correct me if I misunderstood your implementation.

We could try and support both or pick one. Either way it needs a ton of testing on old terminals too because if you fumble an escape code and it ends up going through input loop as a keypress then users view just closes and app exits.  Or escape codes start going into text boxes etc

Terminal.Gui/Input/Key.cs
```diff
@@ -208,13 +209,13 @@ public KeyCode KeyCode
         get => _keyCode;
         init
         {
-#if DEBUG
-            if (GetIsKeyCodeAtoZ (value) && (value & KeyCode.Space) != 0)
-            {
-                throw new ArgumentException ($"Invalid KeyCode: {value} is invalid.", nameof (value));
-            }
+//#if DEBUG
```
tig, line 212: I strongly disagree with removing this debug check.

I put it in here because it forces drivers to be correct. Please explain the situaiton where you think it's ok for a driver to provide an invalid key code.

Terminal.Gui/Input/Key.cs
```diff
@@ -288,7 +289,10 @@ public static bool GetIsKeyCodeAtoZ (KeyCode keyCode)
             return false;
         }
 
-        if ((keyCode & ~KeyCode.Space & ~KeyCode.ShiftMask) is >= KeyCode.A and <= KeyCode.Z)
+        // A to Z may have �, �, �, �, with Oem1, etc
+        ConsoleKeyInfo cki = ConsoleKeyMapping.GetConsoleKeyInfoFromKeyCode (keyCode);
```
tig, line 293: As noted in the source file GetConsoleKeyInfoFromKeyCode is a badly broken API. It does not actually do what it says it does.

In addition, even if it does something useful, it is not platform dependent. Thus it's functionality should be part of the Key class, or a helper class that goes in `./Terminal.Gui/Input`, not the drivers.

UICatalog/Scenarios/AnsiEscapeSequenceRequests.cs
```diff
@@ -0,0 +1,380 @@
+using System;
```
tig, line 1: #nullable enable

UICatalog/Scenarios/AnsiEscapeSequenceRequests.cs
```diff
@@ -0,0 +1,380 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Terminal.Gui;
+
+namespace UICatalog.Scenarios;
+
+[ScenarioMetadata ("AnsiEscapeSequenceRequest", "Ansi Escape Sequence Request")]
+[ScenarioCategory ("Ansi Escape Sequence")]
```
tig, line 10: Just make this "Tests". Avoid category expansion.

UICatalog/Scenarios/AnsiEscapeSequenceRequests.cs
```diff
@@ -0,0 +1,380 @@
...
+
+        // Run - Start the application.
+        Application.Run (appWindow);
+        bulk.View.Dispose ();
+        single.View.Dispose ();
+        appWindow.Dispose ();
+
+        // Shutdown - Calling Application.Shutdown is required.
+        Application.Shutdown ();
+    }
```
tig, line 63: Ctrl-E-C

tig (changes requested):
PR with these comments and a bunch of code clean up inbound.

Please look for new // TODO: and // QUESTION: in the code. 

This is really, really, really great stuff!!!

tig:
Review comments above and a PR:

- https://github.com/BDisp/Terminal.Gui/pull/203

BDisp:
> These are basically 2 ways of solving the same problem
> 
> * `ansi-parser` branch adds a 'pre filter' to all console input loops that picks up responses.  Independently there is a scheduler that handles sends.
> * `ansi-escape-sequence-all-drivers` allows client code to force process the input loop outside of its main iteration until it finds some input it likes (ansi code). So it sends then operates the console input loop till it finds response.
> 
This isn't now true because it doesn't force process the input loop outside of its main iteration anymore. All drivers now uses their own input loop to get the requests responses. So, any key or mouse event are always handle during the process. But initially I was using the method you described but not now.

> The first is harder to use for api of our library because it is based on events. The second is blocking so user gets answer right away but is imo harder to maintain/test and may be less stable.
> 
Doesn't blocking anymore. The request is always returned, with a valid or not valid response, but always return. I admit that can have some delay when a lot of requests are constantly sent, but I think there is an reasonable response.

> @BDisp please correct me if I misunderstood your implementation.
> 
I already done above.

> We could try and support both or pick one. Either way it needs a ton of testing on old terminals too because if you fumble an escape code and it ends up going through input loop as a keypress then users view just closes and app exits. Or escape codes start going into text boxes etc
> 
I admit that I concentrated all my unit tests to the `EscSeqUtils` class which the requests feature use because it had many bugs that I think I already fixed. For ANSI requests unit tests I have to create a mock responses for that, but at least in the UICatalog it's was full tested.

> Also needs tested with things like `ssh` and `screen` etc

I agree. But I think we can use both implementations using different classes, like now, and the user instantiate the one he want, mine with event or yours with action.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -608,6 +608,19 @@ public void OnMouseEvent (MouseEventArgs a)
...
+    /// </summary>
+    /// <param name="ansiRequest">The <see cref="AnsiEscapeSequenceRequest"/> object.</param>
+    /// <returns>The request response.</returns>
+    public abstract string WriteAnsiRequest (AnsiEscapeSequenceRequest ansiRequest);
+
+    /// <summary>
+    ///     Provide proper writing to send escape sequence recognized by the <see cref="ConsoleDriver"/>.
+    /// </summary>
+    /// <param name="ansi"></param>
+    public abstract void WriteRaw (string ansi);
```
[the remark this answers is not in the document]
BDisp: That why I made it abstract because depends of the driver. Only the `NetDriver` don't implement because the `EscSeqRequests.Add` method call the `Console.Out.Write` if the driver parameter is null.

Terminal.Gui/Input/Key.cs
BDisp: Because the `TextField` and the `TextView` wasn't processing accented characters because this method always returning false. If you type in the `WindowsDriver` for e.g. the letter '`á`' the virtual key (`ConsoleKey`) is '`A`'. Thus I think that debug check doesn't make sense, sorry.

Terminal.Gui/Input/Key.cs
BDisp: If you want to move the `GetConsoleKeyInfoFromKeyCode` to the `Key` class that ok, but I guarantee that method and others that was changed in this PR help me to much to improve this feature. I know that each device has it's own keyboard layout but the principle is the same. The english keyboard doesn't have these characters and it's need to use virtual keyboard to simulate that, but isn't like a real keyboard device. We already debated this many times and I was wrong and you was right, about when I said that the '`!` is the same as `Key.D1.WithShift` and I recognize that I was wrong because it really depends of the keyboard layout. But for 'Á' it's really the same as `((Key)'Á').WithShift` and the console key is equal to `ConsoleKey.A`.

Terminal.Gui/Input/Key.cs
tig: I need to you to be much more precise. 

Instead of making this about TextField/View, what about the `Keys` scenario?

I had 'á' working great with the simple fix I suggested yesterday and it did not require disabling this very sane debug check.

Terminal.Gui/Input/Key.cs
tig: Prior to this PR the only code that called `GetConsoleKeyInfoFromKeyCode` was fakedriver and all keyboard input was working fine with the exception of what you called out in #3807.

I provided a 4 line fix to that which did not require using  `GetConsoleKeyInfoFromKeyCode`.

When I dug into the whole vk_packet stuff and tried to debug `GetConsoleKeyInfoFromKeyCode` I found that they really didn't work at all, and did not actually do what they claimed. I do not remember the details, but it was important enough for me to put BUGBUGs in the code. I should have deleted them.

Why can't this fix just be the fix?

https://github.com/gui-cs/Terminal.Gui/issues/3807#issuecomment-2455997595

tig:
> But I think we can use both implementations using different classes, like now, and the user instantiate the one he want, mine with event or yours with action.

This makes no sense to me. There is only one "user" of these classes: Us, the maintainers of TG.

We should not have two implementations of the same thing.

BDisp:
> This makes no sense to me. There is only one "user" of these classes: Us, the maintainers of TG.
> 
> We should not have two implementations of the same thing.

But we was think about was using a public flag or enum for the user choose the type of request. If the user choose AnsiParser then the drivers will initialize the appropriate class and the returning response. So, only TG handle that.

Terminal.Gui/Input/Key.cs
BDisp: I'll revert this and try to fix possible errors.

Terminal.Gui/ConsoleDrivers/WindowsDriver/WindowsDriver.cs
```diff
@@ -1177,6 +198,73 @@ public override void SendKeys (char keyChar, ConsoleKey key, bool shift, bool al
...
+            lock (ansiRequest._responseLock)
+            {
+                ansiRequest.ResponseFromInput += (s, e) =>
+                                                 {
+                                                     Debug.Assert (s == ansiRequest);
+                                                     Debug.Assert (e == ansiRequest.Response);
+
+                                                     _waitAnsiResponse.Set ();
+                                                 };
+
```
tznind, line 223: This code is in all 3 drivers. It looks like you block 100ms to let the sent request get processed and then 'force run' the input stream for a bit, then look for a response? 

Is it possible for you to explain in some pseudocode how it is working?

This PR is enormous and has so much in it that isn't related to ansi response send/capture it's hard to understand what is going on.

I've tried to standardise input processing with AnsiParser<T> to replace NetEvents and similar code in the other drivers. But now all the drivers have huge changes in input handling its going to be difficult to retro fit.

I don't think we need 2 solutions to ansi send. I'd be happy if we just did 

- get rid of all the locking instead use BlockingCollections
- standardise the parsing of request across all 3 drivers into a single class (i.e. AnsiRequestParser). 

We can still do the blocking on response but the class that identifies responses in input streams needs to be centralised into a single class

BDisp: > This code is in all 3 drivers. It looks like you block 100ms to let the sent request get processed and then 'force run' the input stream for a bit, then look for a response?
> 
> Is it possible for you to explain in some pseudocode how it is working?
> 

It's a ManualResetEventSlim like it's used in the MainLoop that wait to be signaled. It only wait until has a response or until two retries and return immediately.

> This PR is enormous and has so much in it that isn't related to ansi response send/capture it's hard to understand what is going on.
> 
I really refactored the `CursesDriver` to leverage using non blocking input, like I did in the `WindowsDriver`.

> I've tried to standardise input processing with AnsiParser to replace NetEvents and similar code in the other drivers. But now all the drivers have huge changes in input handling its going to be difficult to retro fit.
> 
I think that isn't very difficult to implement, beside the changes.

> I don't think we need 2 solutions to ansi send. I'd be happy if we just did
> 
> * get rid of all the locking instead use BlockingCollections

I'm open on that but I need some explanation of that. I tough using the `ManualResetEventSlim` was a good idea, but I would like to know about alternatives, of course.

> * standardise the parsing of request across all 3 drivers into a single class (i.e. AnsiRequestParser).
> 
The `NetDriver` and the `CursesDriver` relies in the `EscSeqUtil.DecodeEscSeq` method because they are handling key and mouse events, but the `WindowsDriver` relies on the `WindowsConsole.ReadConsoleInput` because any other key that it isn't a normal key or a mouse event, is handled in this method.

> We can still do the blocking on response but the class that identifies responses in input streams needs to be centralised into a single class

Well due to my above explanation `NetDriver` and the `CursesDriver` relies in the `EscSeqUtil.DecodeEscSeq` and the `WindowsDriver` relies on the `WindowsConsole.ReadConsoleInput`. So, I don't know if they can be in a single class. But I understand your concern.

Terminal.Gui/ConsoleDrivers/AnsiEscapeSequence/AnsiEscapeSequenceRequest.cs
```diff
@@ -0,0 +1,129 @@
...
+{
+    internal readonly object _responseLock = new (); // Per-instance lock
+
+    /// <summary>
+    ///     Request to send e.g. see
+    ///     <see>
+    ///         <cref>EscSeqUtils.CSI_SendDeviceAttributes.Request</cref>
+    ///     </see>
+    /// </summary>
+    public required string Request { get; init; }
```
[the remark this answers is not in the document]
BDisp: I really don't have no idea what the values of CSIRequests and CSITerminators are. They may be any values.

tig:
> > This makes no sense to me. There is only one "user" of these classes: Us, the maintainers of TG.
> > We should not have two implementations of the same thing.
> 
> But we was think about was using a public flag or enum for the user choose the type of request. If the user choose AnsiParser then the drivers will initialize the appropriate class and the returning response. So, only TG handle that.

Why would we ask/force app developers to have to understand all this? Is there some clear use-case where  developer would pick one vs. the other?

BDisp:
> Why would we ask/force app developers to have to understand all this? Is there some clear use-case where developer would pick one vs. the other?

The user may understand if he want an immediate response or a later response which can be read in another thread than the current thread. That only thoughts.

Terminal.Gui/Input/Key.cs
BDisp: Reverted and working, thanks.

Terminal.Gui/Input/Key.cs
BDisp: Done.

tznind:
I have created a pull request to show how to switch to BlockingCollection and vastly simplify the queue handling that is going on in net driver:

https://github.com/BDisp/Terminal.Gui/pull/204

tig:
> > Why would we ask/force app developers to have to understand all this? Is there some clear use-case where developer would pick one vs. the other?
> 
> The user may understand if he want an immediate response or a later response which can be read in another thread than the current thread. That only thoughts.

Please provide an actual use-case, not an abstract concept. If you can't then there's no reason to provide choice.

TG apps should never care about any of this stuff. It should be completely hidden from app developers.

BDisp:
> Please provide an actual use-case, not an abstract concept. If you can't then there's no reason to provide choice.
> 
> TG apps should never care about any of this stuff. It should be completely hidden from app developers.

I'll discussed this with @tznind to see how this will be needed. For now I can't answers because we still are trying pruning this.

BDisp:
> @BDisp I have created a pull request to show how to switch to BlockingCollection and vastly simplify the queue handling that is going on in net driver:
> 
> [BDisp#204](https://github.com/BDisp/Terminal.Gui/pull/204)

Thank you very much. I did much more changes. Please say what you think about, please.

BDisp:
I already run the unit tests in Windows and Ubuntu and all passed. maybe is something with git actions server.

BDisp:
@tznind did you realized that `BlockingCollection` isn't working with `NetDriver` under WSL for mouse events?

BDisp:
> ```cs
> if (_inputQueue.TryTake (out var item, -1, _netEventsDisposed.Token))
> ```

This fix the issue, thanks so much.

Mainly in this PR the most huge changes was in the `CursesDriver`. But also the EscSeq libraries also had many bugs. Another concern I was having with this PR was avoiding blocking issues, like I was on `WindowsConsole` and `CursesDriver`. `NetDriver` don't have this problem because it check for `Console.KeyAvailable`. Now I was having blocking issue with `Take` method 😃

tznind:
Cool, the ansi parser class is pretty bulletproof and can work synchronously.

It is a state machine and driver agnostic.  So when this pr is ready I want it to take over from all the if/else and string builders in finding responses

BDisp:
@tznind the `NetDriver` it's still blocking on WSL. Can you please test this PR in WSL to confirm that is also happen to you? Thanks.

If affirmative it may explain the unit tests crashs in the git actions.

tznind:
> @tznind the NetDriver it's still blocking on WSL. Can you please test this PR in WSL to confirm that is also happen to you? Thanks.

Not blocked but it seems unstable, got this Exception this time, last time it was just that when I sent ANSI it interpretted as Esc and close

not-works


```
Process terminated. Assertion failed.equenceRequest├───────────────────────────────────────────────────────────────────┐
   at Terminal.Gui.AnsiEscapeSequenceRequestUtils.DecodeEscSeq(AnsiEscapeSequenceRequests escSeqRequests, ConsoleKeyInfo& newConsoleKeyInfo, ConsoleKey& key, ConsoleKeyInfo[] cki, ConsoleModifiers& mod, String& c1Control, String& code, String[]& values, String& terminator, Boolean& isMouse, List`1& buttonState, Point& pos, AnsiEscapeSequenceRequestStatus& seqReqStatus, Action`2 continuousButtonPressedHandler) in /mnt/d/Repos/temp/gui.cs/Terminal.Gui/ConsoleDrivers/AnsiEscapeSequence/AnsiEscapeSequenceRequestUtils.cs:line 379                                                                    ││
   at Terminal.Gui.NetEvents.ProcessRequestResponse(ConsoleKeyInfo& newConsoleKeyInfo, ConsoleKey& key, ConsoleKeyInfo[] cki, ConsoleModifiers& mod) in /mnt/d/Repos/temp/gui.cs/Terminal.Gui/ConsoleDrivers/NetDriver/NetEvents.cs:line 311  ││
   at Terminal.Gui.NetEvents.ProcessInputQueue() in /mnt/d/Repos/temp/gui.cs/Terminal.Gui/ConsoleDrivers/NetDriver/NetEvents.cs:line 183                                                                                                      ││
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)                                                                     ││
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)              ││
   at System.Threading.ThreadPoolWorkQueue.Dispatch()                                                                 ││
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()                                            ││
thomas@DESKTOP-GE0PH5S:/mnt/d/Repos/temp/gui.cs/UICatalog$                           ExpecteTerminator:lue:   
```

BDisp:
It's impossible to debug. It simply block without output any information. Sometimes using conhost I receive the error output you mentioned, but I put a breakpoint there the app hangs without give me any chance to find what's going on. On Windows this is working very well. Thus I deduce that may be some console conflict handling with any `Queque` object.

BDisp:
@tznind please can you review the changes made in the PR and I also added error handling in the scenario. Test in Windows and Linux. Thanks.

BDisp:
@tznind can you help with a situation in Linux where if I continuous press the enter key in the ansi scenario the assert fail in the method `DecodeEscSeq` at line 377 of the `AnsiEscapeSequenceRequestUtils.cs` file. On Windows this never happens. The request is always added to the `Statuses` property before sent to the console. I think that may be some lock not provided, but since this is only happens on the Linux is very hard to find where is failing. Thanks.

tznind:
> @tznind can you help with a situation in Linux where if I continuous press the enter key in the ansi scenario the assert fail in the method [...]

Are you able to reproduce on either of these branches?

tznind:ansi-parser
tznind:parser-attempt-2

Because I cannot
hold-enter-works
_Holding enter when on parser-attempt-2 branch_

BDisp:
I'm meaning in this PR in the single tab, please.

BDisp:
@tig can I do Ctrl+E+C in all drivers including console, mainloop, etc.?

BDisp:
> I'm meaning in this PR in the single tab, please.

I already fixed here https://github.com/gui-cs/Terminal.Gui/pull/3768/commits/6444cc07dd20d4f758758ea2e04ff2005f39042c. Thanks.

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -608,6 +608,19 @@ public void OnMouseEvent (MouseEventArgs a)
     /// <param name="ctrl">If <see langword="true"/> simulates the Ctrl key being pressed.</param>
     public abstract void SendKeys (char keyChar, ConsoleKey key, bool shift, bool alt, bool ctrl);
 
+    /// <summary>
+    ///     Provide handling for the terminal write ANSI escape sequence request.
+    /// </summary>
+    /// <param name="ansiRequest">The <see cref="AnsiEscapeSequenceRequest"/> object.</param>
+    /// <returns>The request response.</returns>
+    public abstract string WriteAnsiRequest (AnsiEscapeSequenceRequest ansiRequest);
```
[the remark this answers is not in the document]
BDisp: Done here https://github.com/gui-cs/Terminal.Gui/pull/3768/commits/36876b989900f6531e06ba89b3420777f37e0819. But I had to change some internal to public, but at least its work and reduced many duplicated code. Let me know what do you think, because I made as draft and may not look very beautiful.

BDisp:
NetDriver:
`AnsiParser` is loosing many responses:

Captura de ecrã 2024-11-11 012640

`AnsiEscapeSequence` get more responses valid responses, I provided a summary to inspect if there are invalid responses errors:

Captura de ecrã 2024-11-11 012849

tznind:
>  AnsiParser is loosing many responses:

Hmn doubtful, `AnsiParser<T>`  cannot 'loose' responses.  If it did, you would see Esc and keys falling through to main loop and appearing in text box/exiting window etc.

More likely the `AnsiRequestScheduler` is preventing going above 10 because there are still outstanding request.  Scheduler prevents there being more than 1 request 'in flight' at once and has various throttles on requests per second etc.

BDisp:
Sorry, I misspelled the word loose with schedule requests and so I meant about outstanding requests.
I confess that I'm not very good to interpret graphics. The only doubt I have is I'm sending requests sequentially but there greens in the middle of reds. My question is if the graph is querying with the right values.

Note:
Did you already noted that the points are being draw over the Y axis in the `GraphView`? Did this already happens before?

BDisp:
> This all looks correct to me.
> 
> Points are meant to draw over axis otherwise you cannot tell where intersection is.
> 
Understood, that make sense, otherwise we wont see anything that was finished at little elapsed time less than a second.

> Green series is drawn first then red series after (over the top). Because there is always a delay before DAR send/queue (green) and response received (red) they will not always exactly like up over one another.
> 
> tl;dr looks fine!
> 
I changed the filter on the ansi scenario and now I only see red points which mean that they are responded before draw in the graph.

> Just make sure that you can still type even at max requests I.e. and doesn't starve other input processing

I know this is the main point which it's need to have the maximum careful to avoid hangs on the UI. With typing on the `TextView` all drivers works reasonable well. Only the mouse events hangs a lot with the `NetDriver` in Windows. I know that requests response must be handled on a separated thread, it's the only solution.

BDisp:
With the latest changes it can handle much more requests per second and the `NetDriver` handles the mouse more quickly with running requests.
I think this can be reviewed to being cleaning code.

BDisp:
@tznind the green points always represent the sent requests no matter was already answered or not. Sometimes isn't showing because a red point, which I think is drawn last, draw over it. I changed my scenario to remove the answer received from the _sends variable and it's always 0. I don't know if I'm totally right.

tznind:
With WindowsDriver clicking 'SendRequest' exits (i.e. is interpretted as Esc not ansi response).

tznind:
I think we should merge https://github.com/gui-cs/Terminal.Gui/pull/3791 instead of this to start with.  The parsing and request sending code adds even more threading and locking when we already have issues with the drivers e.g. see #3812

BDisp:
> With WindowsDriver clicking 'SendRequest' exits (i.e. is interpretted as Esc not ansi response) in VisualStudio console only (i.e. not reproduce in Windows Terminal).

I can't reproduce this.

VsDebugConsole_zSMVnL0Vle

tznind:
> > With WindowsDriver clicking 'SendRequest' exits (i.e. is interpretted as Esc not ansi response) in VisualStudio console only (i.e. not reproduce in Windows Terminal).
> 
> I can't reproduce this.

This is the state of the read input at the point your code decides to send Escape key down
image

Here is the repro video:
esc


Here is a git diff patch that adds static queue of the input events
```
 git diff
diff --git a/Terminal.Gui/ConsoleDrivers/WindowsDriver/WindowsConsole.cs b/Terminal.Gui/ConsoleDrivers/WindowsDriver/WindowsConsole.cs
index b27b8fe96..5c0f620e2 100644
--- a/Terminal.Gui/ConsoleDrivers/WindowsDriver/WindowsConsole.cs
+++ b/Terminal.Gui/ConsoleDrivers/WindowsDriver/WindowsConsole.cs
@@ -57,6 +57,7 @@ internal class WindowsConsole
         return null;
     }

+    public static List<InputRecord> seen = new ();
     public InputRecord? ReadConsoleInput ()
     {
         const int BUFFER_SIZE = 1;
@@ -80,6 +81,8 @@ internal class WindowsConsole
                                       BUFFER_SIZE,
                                       out numberEventsRead);

+                    seen.Add (inputRecord);
+
                     if (inputRecord.EventType == EventType.Key)
                     {
                         KeyEventRecord keyEvent = inputRecord.KeyEvent;
```

I guess what happens is the console flushes `<esc>[?` then takes a breather for a millisecond before flushing `1;0c`

Probably your code is bailing too early.

BDisp:
I made all the steps you did here and I don't get any `OnKeyDown` event when I mouse click on the send button. What is the `_isWindowsTerminal` value when you hit the event?

tznind:
I am not surprised that you cannot repro.  Terminals can all vary a lot between versions.

This is my term exe debugger version.

It was at
`C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Platform\Debugger\VsDebugConsole.exe`


image

BDisp:
The same here but in Portuguese.

image

tznind:
Is this same exe path you see when opening DeviceManager while debugging and going to 'open exe location'?

You may have multiple versions installed

BDisp:
What the net80 version do you're using? The latest VS2022 also installed the net90. Are you running the latest updated version of my branch?

BDisp:
> Is this same exe path you see when opening DeviceManager while debugging and going to 'open exe location'?
> 
> You may have multiple versions installed

Yes, this is the same when I' debugging and going to 'open exe location'.

tznind:
Yes running latest commit i checked with git log

BDisp:
> Yes running latest commit i checked with git log

I don't understand what it's causing this behavior to you. All drivers don't send any escape sequence response to the `ProcessInput`. I already prevents that but may could some bug which I don't know yet.

tznind:
I've written more about the direction I think we should be headed here:

https://github.com/gui-cs/Terminal.Gui/issues/3692#issuecomment-2423661375

tig:
Yep!

tig:
@BDisp & @tznind I put a bunch of work into a PR that went into this PR that cleaned up a bunch of driver stuff. mostly structrual.

https://github.com/tig/Terminal.Gui/tree/BDisp-v2_3767_ansi-escape-sequence-all-drivers

By closing this PR that work was lost. 

Can you please help me get that work back and into v2_develop?

BDisp:
> @BDisp & @tznind I put a bunch of work into a PR that went into this PR that cleaned up a bunch of driver stuff. mostly structrual.
> 
> https://github.com/tig/Terminal.Gui/tree/BDisp-v2_3767_ansi-escape-sequence-all-drivers
> 
> By closing this PR that work was lost.
> 
> Can you please help me get that work back and into v2_develop?

That isn't very easy to track all the changes done because they are in a new folders and have a bunch of changes.

tznind:
I have also begun splitting ConsoleDriver into its 3 constituent parts, each is now a seperate class with base classes etc - see #3821 and #3837

- Content (state + draw)
- Out (writing to output, screen buffers, p/invoke etc)
- In (reading from input, p/invoke etc)

What would help me most would be if you could review and consider merging #3791 as it does what this branch did (allow user to send ANSI escape sequences and pick up responses).  That would cut down the diff on v2 considerably and make it easier to see the dircetion it is going while still delivering useful functionality:

- Ability to send ANSI codes
- Ability to detect sixel support

BDisp:
Please give me the opportunity to try to restore in a new branch all the fixes I made without the ANSI escape sequence, because I fixed other things and bugs that will be useful for merging into v2_develop. I think it won't be too difficult to merge it later. If I can't restore them in two days, I will say so and we will definitely forget about it. Accordingly?

PR #3787, opened 2024-10-10
Adjust ci native

tznind:
## Fixes

- Fixes #_____

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:

### DIFF
--- a/NativeAot/nuget.config
+++ b/NativeAot/nuget.config
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-
+﻿
 <configuration>
   <config>
     <add key="defaultPushSource" value="https://api.nuget.org/v3/index.json" />
@@ -9,8 +8,7 @@
     <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
     <add key="nuget" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
-    <add key="LocalPackages" value="Terminal.Gui/bin/Release" />
-    <add key="local_packages" value="./local_packages" />
+    <add key="local_packages" value="../Terminal.Gui/local_packages" />
   </packageSources>
   <packageSourceMapping>
     <packageSource key="nuget">
--- a/NativeAot/nuget.config
+++ b/NativeAot/nuget.config
@@ -1,24 +1,18 @@
-﻿
-<configuration>
+﻿<configuration>
   <config>
     <add key="defaultPushSource" value="https://api.nuget.org/v3/index.json" />
   </config>
   <packageSources>
-
-    <!--To inherit the global NuGet package sources remove the <clear/> line below -->
     <clear />
-    <add key="nuget" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
     <add key="local_packages" value="../Terminal.Gui/local_packages" />
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
   <packageSourceMapping>
+    <packageSource key="local_packages">
+      <package pattern="Terminal.Gui" />
+    </packageSource>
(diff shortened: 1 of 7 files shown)


SelfContained/SelfContained.csproj
```diff
@@ -12,14 +12,9 @@
     <DebugType>embedded</DebugType>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
-    <ProjectReference Include="..\Terminal.Gui\Terminal.Gui.csproj" />
-    <TrimmerRootAssembly Include="Terminal.Gui" />
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
BDisp, line 20: `SelfContained` also needs to use the local nuget package in release mode, because some of the trimmed feature handles in deferent way in debug and release mode.

SelfContained/nuget.config
```diff
@@ -0,0 +1,24 @@
...
+    <add key="nuget" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="local_packages" value="../Terminal.Gui/local_packages" />
+  </packageSources>
+  <packageSourceMapping>
+    <packageSource key="nuget">
+      <package pattern="*" />
+    </packageSource>
+	<packageSource key="LocalPackages">
+	  <package pattern="Terminal.Gui" />
+	</packageSource>
```
BDisp, line 19: I think this can be removed because only the `local_packages` is used in the `packageSources`.

NativeAot/nuget.config
```diff
@@ -9,10 +8,17 @@
...
     <add key="nuget" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="local_packages" value="../Terminal.Gui/local_packages" />
   </packageSources>
   <packageSourceMapping>
     <packageSource key="nuget">
       <package pattern="*" />
     </packageSource>
+	<packageSource key="LocalPackages">
+	  <package pattern="Terminal.Gui" />
+	</packageSource>
```
BDisp, line 19: I think this can be removed because only the `local_packages` is used in the `packageSources`.

BDisp (changes requested):

SelfContained/SelfContained.csproj
tznind: I've removed all conditions, that means all configurations regardless will use the same package.  Isn't that what we want?

tznind:
Should Native and Aot also be built in the debug builds?

SelfContained/SelfContained.csproj
BDisp: Does we can debug if using the release package? If we can that fine, otherwise I prefer to use the project reference.

BDisp:
> Should Native and Aot also be built in the debug CI pipelines builds?

I don't think so. Locally, normally it build well but when ran sometimes an error was throwing. In the Ci pipelines doesn't justify.

SelfContained/SelfContained.csproj
BDisp: > I've removed all conditions, that means all configurations regardless will use the same package. Isn't that what we want?

But you are referencing the project and not the package. In the `SelfContained` project it need at least in the release mode, reference the package because sometimes e may failing when running in release mode. That is important.

BDisp:
`NativeAot` project isn't compiling in release mode locally.

image

BDisp:
@tznind thanks very much for your help but I think I already fixed this. I have to use scripts for release mode but there is no way to suspend the package restore to build and pack first using project configuration.

tznind:
Cool glad you were able to sort it!

PR #3791, opened 2024-10-11
Fixes #3767 - Adds Ansi parser and scheduler.

### LINKED ISSUE #3767: Allowing any driver to request ANSI escape sequence with immediate response.

tznind:
Adds ability to send and respond to ANSI escape sequences as they come in via event callbacks.

Also adds detection of Sixel support, see `SixelSupportDetector.cs`

The purpose is to stream text from console driver as read and 'transparently' pluck out the expected responses live.

## Explanation of Ansi Parser

AnsiParser processes events from an input stream.  Each time it will take 1 `char` and either add it to `Held` or Release it.  Critically it sometimes releases more than 1 event.

Each event is a `char` which can have optional metadata attached (See `T` header below).

Consider the following use case:

> User presses Esc key then types "Hi" then hits Esc again.  While the user is typing a DAR response comes in `<esc>[0c`

image

**Here is a breakdown of how AnsiParser would deal with this:**

## Stage 1 - Consume first Esc

The first call to ProcessInput is with the `Esc` char.  This causes the parser to shift into expecting an escape code (i.e. a bracket).  Because we are assuming the Esc is a response we hold it and return empty.

image

## Stage 2 - Consume H

The next call to ProcessInput is with `H`.  We are expecting a bracket which would indicate that we are indeed entering an escape sequence.  Since we did not find one (`[`), we instead release both the `Esc` and the `H` for routine processing.

image

## Stage 3 - Consume next escape sequence

This process repeats, this time for the actual response we are expecting (i.e. ending with `c`)

image

image

image

When we reach the first character that is a `_knownTerminators` (i.e. `c`) or matches one of our `expectedResponses` terminators we will leave the `InResponse` state.

image

If the response built up in Held matches an `expectedResponses` we swallow it (Release None) and raise event
If the response built up does not match (instead it matches `_knownTerminators`) we release the whole thing back to be processed downstream.

## Stage 4 - Consume last `Esc`

Finally we consume the last `Esc` which leaves us in state `Expecting Bracket`.  Now this may be the start of a new escape sequence or it could be that the user has just pressed Esc - we will not know which till later.

BUT we don't want to sit around waiting for the rest of the escape sequence forever. Pressing Esc could be the last thing the user does and there could be no events while user sits around watiting for app to respond.

For this reason we put timestamp on state changes `StateChangedAt` - this lets the caller (e.g. driver main loop) force the Parser to release the `Esc` after a short period of time if no new input is comming:

image

```csharp
if (Parser.State == ParserState.ExpectingBracket &&
    DateTime.Now - Parser.StateChangedAt > _escTimeout)
{
    return Parser.Release ().Select (o => o.Item2);
}
```

## Why `<T>`?
I realized working exclusively in `char` and `string` made it difficult for WindowsDriver to integrate - so I have changed to `AnsiParser<T>` .  The class now deals with sequences of `char` each of which has any metadata you want (type T).  For WindowsDriver this means `AnsiResponseParser<WindowsConsole.InputRecord>`.

The parser can pull things out of the stream and later return them and we don't loose the metadata.

I will look at the other drivers, if they are dealing just in `char` with no other metadata I can see about putting a non generic one too that just wraps the generic like `TreeView` vs `TreeView<T>`

## Outstanding Issues

- [x] Key down and key up means double key presses
- [x] Esc on its own as isolated keypress needs to have a release timeout (i.e. if we dont see '[' in 50ms then we need to release that back to input stream as nothing else is coming).
- [x] deal/prevent parallel outstanding requests executing at once
- [x] Hitting Esc key twice in quick succession or users Esc+start escape sequence

## Fixes

- Fixes #3767

## Pull Request checklist:


tznind:
Key down and Key up events are a pain.  Here is what response looks like when sending a Device Attributes Request on startup in WindowsDriver.

We get keydown and then keyup for each letter.
`esc[?1;0c`

image

So we would need another 'pre step' to the parser that paired up down/up events and only passed them around in pairs (i.e. swallow both or release both).  Currently I am using `T` as `InputRecord` but now maybe it needs to be a Pair (down and up).

Any idea if this is universal behaviour for WindowsDriver? or are there cases where it only sends key downs for example?

BDisp:
> Any idea if this is universal behaviour for WindowsDriver? or are there cases where it only sends key downs for example?

I think that is a behavior for Wind32 API but would be interesting to know how that work with the others drivers, because you are sending any keystroke with the prefix 0x1b, right? If the other drivers respond with the same output, then there is a possibility that all drivers can deal with key-down and keu-up.

tznind:
Ok I have added `WindowsDriverKeyPairer` to match up inputs - good news is no changes to the parser we just make the 'metadata' for each `char` the pair and return that to the main loop.


[d642fb6](https://github.com/gui-cs/Terminal.Gui/pull/3791/commits/d642fb65757d27d0b572120bd7447bc303ae59eb)

tznind:
Key down/up is very unreliable.  Just by typing fast you get random order of down/ups (this is on v2_develop):

image

tznind:
Maybe we can ditch key up event? it seems almost everything goes of key down anyway? And 2 of the 3 drivers are just making up down/up as pairs anyway - for example NetDriver:

```csharp

                KeyCode map = MapKey (consoleKeyInfo);

                if (map == KeyCode.Null)
                {
                    break;
                }

                OnKeyDown (new Key (map));
                OnKeyUp (new Key (map));

                break;
```

tznind:
Now working with WindowsDriver and CursesDriver 🎉 

Looks like there is already multiple bespoke systems for handling these escape sequences (each driver has its own implementation) so later on it might be worth expanding the parser to do that logic too (is mouse etc).

tznind:
@BDisp I have built stress testing Scenario into this PR

You can type away while the system is sending ANSI requests.  Currently the WindowsDriver will crash under following circumstances.

Failure state for this test is:

- U becomes anything other than 1 (this means the DAR is getting 'responded' to with other strings).  Likely a symptomn of failure
- When the app crashes - I have seen this with the `'Invalid KeyCode: Space, C is invalid.` thing we saw in your PR too.

image

BDisp:
> @BDisp I have built stress testing Scenario into this PR
> 
> You can type away while the system is sending ANSI requests. Currently the WindowsDriver will crash under following circumstances.
> 
> Currently I see errors on my computer around 16 requests per second.
> 
> Failure state for this test is:
> 
> * U becomes anything other than 1 (this means the DAR is getting 'responded' to with other strings).  Likely a symptomn of failure
> * When the app crashes - I have seen this with the `'Invalid KeyCode: Space, C is invalid.` thing we saw in your PR too.
> 
I have fixed some of that failures in my PR. If I remind the `'Invalid KeyCode: Space, C is invalid.` was occurring because I wasn't breaking the response with the terminator request, reading while key is available. Other issue was having an extra char after the terminator on the `NetDriver`, which I fixed in the `NetEvents.ProcessInputQueue`. Other issue was having an extra char before the esc key, which I fixed in the `ConsoleDriver.ReadAnsiResponseDefault` method.
> 
> I will look to add a Scheduler component to this PR as really we should prevent 2+ requests being sent before response for first as come in (When terminator is same). With some timeouts incase a request is ignored by term or user specifies wrong terminator etc.

In my PR I removed the control about when terminator is same. That was causing malfunctions. See how the `EscSeqRequests` class is working now.

tznind:
Ok, but your code and mine are different so it may not be the same thing exactly.

I have captured everything that came in over std in and here is the last records up to the crash (see below)

I have color coded the responses.  Each response should be `<esc>[?1;0c` however when it gets too many requests too fast we see that one response starts coming in before the last has been completely written.

Colored in blue and red we see isolated 'bits' of a response that are interrupted by new events.

The bit that causes error is the final `c` terminator which seems to come in as `Space | C` but that is really just a symptom.  Problem is the terminal giving us responses in the wrong order.

I will update the parser internals to throttle requests based on how fast the responses are coming in (i.e. 1 outstanding request at once) - that should fix this I think.  Although the windows terminal is really the thing at fault here IMO - it should not be doing this.

image

BDisp:
Yes I see. Maybe it isn't thread safe. When you get the response in the Action method it may have mixed keys, which make it unreadable.

tznind:
Ok I have added a scheduler, its still a WIP - I think it makes mose sense to move it to ConsoleDriver rather than instantiating in scenario.  But it has 2 throttling mechanisms:

- Do not allow a send request when there is an outstanding response being waited for
- Do not send more than x requests per second (e.g. 10) for any given terminator

In both cases the request is put in holding and sent when chance next becomes available (and throttle/wait conditions are met).

image

Terminal.Gui/ConsoleDrivers/AnsiResponseParser.cs
```diff
@@ -1,7 +1,91 @@
...
 
+using System.Collections.Concurrent;
+using System.Runtime.ConstrainedExecution;
+
 namespace Terminal.Gui;
 
+public class AnsiRequestScheduler(IAnsiResponseParser parser)
+{
+    public static int sent = 0;
+    public List<AnsiEscapeSequenceRequest> Requsts = new  ();
```
BDisp, line 11: Typo change to `Requests`, please.

BDisp (changes requested):

tznind:
> Do you also prevented if no response get back at all, e.g. in the cases where the request isn't recognized and no event is send by the mainloop, waiting indefinitely?

Well there is no waiting going on.  But yes unmet requests will be forever outstanding which will also block new requests for same terminator. 

I will look at adding timeout to scheduler.  It shouldn't be too hard.

BDisp:
> Well there is no waiting going on. But yes unmet requests will be forever outstanding which will also block new requests for same terminator.
> 
I think you shouldn't restrict requests for the same terminator because there are different requests that use the same terminator like the size request 't'. That why I removed it from my PR. Requests must be read sequentially, no mater having the same terminator.

> I will look at adding timeout to scheduler. It shouldn't be too hard.

That's the disadvantage of non-blocking case because returning from the timeout doesn't mean there is no response at all later, but I really think there is no other way. I also had the same issue and I was looking from null input representing any response but there is no such of signal from all drivers. In my case was more easy because if there is no key available then the response will be empty. With the `NetDriver` if the `EscSeqRequests.Statuses` isn't empty and no response was returned, then I dequeue it.

tznind:
> I think you shouldn't restrict requests for the same terminator 

I am not restricting. Just ensuring that there are not 2 running at once. If 2 come along at same time the second Just gets queued till first gets its response answered e.g. in 20ms when the console finishes writing its reply to us.

Thinking about it, for the timeout I actually only need to evict when a new request comes in for same terminator (and old requestis past timeout).  Otherwise having outstanding request does no harm.

Another use case for timeout is the schedule queue. If user is pushing faster than requests can be dispatched by a lot and for a long time we don't just want an ever growing list outstanding. They need to be abandoned or throw an Exception.

BDisp:
Another issue I had with my PR was in `WindowsDriver` and `CursesDriver` when the mainloop is waiting for input. Even suspended them from reading the input to only using the `Console.ReadKey`, theses drivers held the first character, which is the Esc if we done a request, corrupting the response with `Invalid KeyCode: Space, C is invalid`. Thus in my PR I now check if there is available input before read and now that doesn't happen. But since you you're using non-blocking mode this shouldn't been happening, curious.

tznind:
Ok I think this is more or less feature complete.

Next I will add:

- [ ] Scheduler tests
- [ ] More parser tests
- [ ] Scenario support for more diverse requests (like yours @BDisp)

BDisp:
Do you are having unfinished requests returning by the timeout schedule? If affirmative would also be great to include in the stress tests with the yellow color.

tznind:
NetDriver is really difficult to integrate with AnsiParser.  The main loop `ProcessInputQueue` is choc full of `break` and `continue`.

Ideally I wanted to start by just having AnsiParser pre filter the input then pass on the rest to the main loop like I did with WindowsDriver but its difficult with calls like `Console.KeyAvailable` in the middle of the handling loop code (i.e. `ProcessInputQueue`).

A lot of what it is doing is detecting ansi codes (mouse, key, terminal resize etc).

I will look into whether the existing code can be replaced wholesale.  It should be possible just to have an outstanding expectation in parser for the mouse and key combinations and just have those come out as events like the rest.  

I will use a seperate branch incase things dont work out.

tznind:
Ok I have added the idea of persistent expectations.  Can think about refactoring as now have 3 collections (expected, late and persistent). I think this will let the existing mouse/screen events logic in NetEvents get nuked.  Will look at doing that in `ansi-parser-net-driver`

tznind:
Ok this now includes the NetDriver changes to have AnsiParser take over responsibility for response parsing instead of the NetEvents class

Only issues I now have revolve around some terminals sending interleaved responses 

As described here: https://github.com/gui-cs/Terminal.Gui/pull/3791#issuecomment-2425135732

We no longer have the DAR responses tripping each other because we schedule those to have 1 in 1 out.

But sometimes we see mouse getting interrupted by DAR response (in some terminals).

Easiest way to get this to trigger - if you have a console where it is an issue.  Is to run the scenario and set to 20 DAR per second then just wave mouse around frantically.

It manifest in one of two ways

- Scenario closes because 2 Esc get seen in a row (DAR interleaved with Mouse both outputting `Esc` then both outputting `[` (or some variation thereof)
- If in a text field - part of the mouse response code or DAR appearing written into text view as text

tznind:
@j4james thanks for the help on the Sixel stuff - maybe I can pick your brains on this also? 

Have you ever seen this kind of behaviour from terminals? I am pretty sure this is coming in raw like this and not a race condition in the Terminal.Gui code.

Any advice would be appreciated.

j4james:
@tznind I've not experienced that issue myself, but I could easily believe that was a bug in older versions of the conhost/conpty framework. And if you're getting `\e[1;0c` in the DA1 report, that suggest you're testing an old version of Windows Terminal, or an old version of conhost, or a third-party terminal which is relying on an old version of conhost.

Something similar was reported quite recently as a bug in Windows Terminal (https://github.com/microsoft/terminal/issues/17775), but that has since been fixed, and was identified as a regression, so I think that would only have been evident in versions 1.19 to 1.21 of Windows Terminal (which would not have returned a `\e[1;0c` report). However, that does suggest that a similar bug might have been present in older versions as well.

tznind:
Interesting, thanks for linking - that does sound similar, and indeed, I have a very old version running as default terminal for Visual Studio.

The test here is pretty extreme (spamming DAR is a bit of an artificial use case) and it doesn't reproduce on more recent versions - maybe we can live with this.

I will look into it further and try with some of the Linux consoles like xterm and mlterm I have on laptop later this week.

tznind:
@BDisp this is also ready - which do you want to merge first if we are going to have both these systems (blocking/realtime) as options?

BDisp:
> @BDisp this is also ready - which do you want to merge first if we are going to have both these systems (blocking/realtime) as options?

I prefer mine is merge first because it has all driver as non-blocking input and very raw changes. I think it would more easy to merge yours later, if you don't mind.

tznind:
ok, I also have changes to driver but I think they are more limitted

tznind:
@BDisp I have also added your combo box and controls to this PR so we can test other commands.

I also added the `StoppedExpecting` event so we can see when the parser gives up on getting a response.

run-single

BDisp:
On the request text field put a invalid request that wont response at all and see what happen. Also with a valid request maintain the enter key pressed during a long time and see if any error come up.

UICatalog/Properties/launchSettings.json
```diff
@@ -1,7 +1,8 @@
 {
   "profiles": {
     "UICatalog": {
-      "commandName": "Project"
+      "commandName": "Project",
+      "commandLineArgs": "-d NetDriver"
     },
     "UICatalog --driver NetDriver": {
```
BDisp, line 7: There is already a NetDriver for the UICatalog. This is really needed?

tznind: Reverted, this was change made in UI

Terminal.Gui/ConsoleDrivers/NetDriver.cs
```diff
@@ -1716,63 +1673,33 @@ void IMainLoopDriver.Setup (MainLoop mainLoop)
...
-            return true;
-        }
-
-        try
-        {
-            if (!_eventReadyTokenSource.IsCancellationRequested)
-            {
-                // Note: ManualResetEventSlim.Wait will wait indefinitely if the timeout is -1. The timeout is -1 when there
-                // are no timers, but there IS an idle handler waiting.
-                _eventReady.Wait (waitTimeout, _eventReadyTokenSource.Token);
```
BDisp, line 1736: Did you realized that the `EventsPending` method is always running overkilling the CPU usage with 13%, approximately? That why the `Threading` scenario it's working on `Linux` and `Windows`. It's needed to wait until it's necessary. So, the `BlockingCollection` is only blocking in the `NetInputHandler` method but the `EventsPending` method is always running. I think the `BlockingCollection` blocks in it own thread and cause the driver not responding at all hanging in `Linux`. In `Windows` may have some managed code that can surpass this issue, maybe because `NetDriver` is using Win32 API which respond to a focus event, I don't know.

tznind: Thanks for pointing this out, I have fixed in [49475b6](https://github.com/gui-cs/Terminal.Gui/pull/3791/commits/49475b6f9481e23ad0e1375edc4f9b4a80fe8217)

tznind:
Current state of this branch is that it is working and could be merged.  Only reason it is in draft is because @BDisp and I are also exploring combined approach - see https://github.com/BDisp/Terminal.Gui/pull/207

That branch is currently encountering significant issues with deadlocks, so we're treating it as an experimental effort for the time being.

tig:
@tznind I remain confused about this PR. I think you said I should merge it. 

Should I? (after you fix the conflicts of course)?

tznind:
> @tznind I remain confused about this PR. I think you said I should merge it. 
> 
> Should I? (after you fix the conflicts of course)?

The PR adds ansi code send/parse - for example to request if sixel is supported.

However to do this needs to add itself into drivers loops input processing pipelines.  

The conflicts are substantial but I will try to resolve.

tznind:
Seems like this test failure situation is non deterministic, it failed in [b0e921a](https://github.com/gui-cs/Terminal.Gui/pull/3791/commits/b0e921a51222aed3b31f87af7d37272cc8e55e55) but passed in next commit ([1480f13](https://github.com/gui-cs/Terminal.Gui/pull/3791/commits/1480f1316b76b7c4e56bdf9369ff5ff535c4abac)) which was just a rename.

tig:
image

Yeah, non-deterministic. Drives me nuts.

BDisp:
I also already faced these failures when calling `Application.Init` and `Application.Shutdown` in a unit test without using an attribute that ensures `Application.Init` is called without a previous call to `Application.Shutdown` nor `Application.Shutdown` is called without a previous call to `Application.Init`.

tig (changes requested):
Submitted a PR instead of doing a full reveiw in line...

tig:
https://github.com/tznind/gui.cs/pull/173

tznind:
Great thanks, I will look through the BUGBUG when I get a chance and fix

tznind:
Ok I think I have addressed most of the TODO / BUGBUG 

The only ones I have left were around `EscSeqUtils` which I have not modified in this PR and which are quite intimidating - code wise. 

I looked at using Pos align in scenario but I think the reason for gap is because of combo boxes nature to not pop over properly.  I couldnt see how pos align would help especially since there are multiple label/text fields per row - we dont want to just stack the views on top of each other vertically.

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
```diff
@@ -990,6 +999,14 @@ ref ConsoleKeyInfo [] cki
                 }
                 else
                 {
+                    if (cki != null)
+                    {
+                        foreach (var c in cki)
+                        {
+                            _parser.ProcessInput (c.KeyChar.ToString());
+                        }
```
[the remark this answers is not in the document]
tznind: Yeah curses is currently blowing up at

```
                 // It's request response that wasn't handled by a valid request terminator
                 System.Diagnostics.Debug.Assert (EscSeqRequests.Statuses.Count > 0);
```

I will have to review what has changed to break this.

But curses code is fighting with ansi parser because both want to build up ansi escape codes - really the old code should be removed and switched to work with the ansi parser rather than against it (e.g. get rid of `ConsoleKeyInfo []? IncompleteCkInfos ` and `public static List<EscSeqReqStatus> Statuses { get; } = [];`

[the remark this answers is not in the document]
tznind: Ok code no longer crashes in curses and is back to working as of [d063873](https://github.com/gui-cs/Terminal.Gui/pull/3791/commits/d06387325fc504a58a86e2c77bab232478589e39)

Basically before we go into `EscSeqUtils.DecodeEscSeq` we give the whole sequence to ansi parser, if it consumes it then it means there was an outstanding request and it was a valid escape code etc.

Only if parser fails to process it do we release back to the legacy code.

Note that the legacy code also deals with mouse etc so we cannot remove it entirely - although I am working towards that in my v2 drivers branch.

We will basically have lots of little parsers MouseParser, ScreenSizeChangeParser etc and they will all use regex and function on complete sequences rather than the state machine stuff - which will all be handled already up front by the a AnsiParser

BDisp:
> Ok I think I have addressed most of the TODO / BUGBUG
> 
> The only ones I have left were around `EscSeqUtils` which I have not modified in this PR and which are quite intimidating - code wise.
> 
The changes I've made with the `EscSeqUtils` allow a driver to handle with all keys, mouse and request ANSI responses. When I was done the changes in the `CursesDriver` in the closed PR it only used their methods, mainly the `DecodeEscSeq`. But of course it can be improved.

> I looked at using Pos align in scenario but I think the reason for gap is because of combo boxes nature to not pop over properly. I couldnt see how pos align would help especially since there are multiple label/text fields per row - we dont want to just stack the views on top of each other vertically.

I did some changes on my closed PR related with the scenario. Don't know if it will help you.

```cs
private View BuildSingleTab ()
{
    var w = new View
    {
        Width = Dim.Fill (),
        Height = Dim.Fill (),
        CanFocus = true
    };

    w.Padding.Thickness = new (1);

    // TODO: This hackery is why I think the EscSeqUtils class should be refactored and the CSI's made type safe.
    List<string> scrRequests = new ()
    {
        "CSI_SendDeviceAttributes",
        "CSI_ReportTerminalSizeInChars",
        "CSI_RequestCursorPositionReport",
        "CSI_SendDeviceAttributes2"
    };

    var cbRequests = new ComboBox { Width = 40, Height = 5, ReadOnly = true, Source = new ListWrapper<string> (new (scrRequests)) };
    w.Add (cbRequests);

    // TODO: Use Pos.Align and Dim.Func so these hardcoded widths aren't needed.
    var label = new Label { Y = Pos.Bottom (cbRequests) + 1, Text = "_Request:" };
    var tfRequest = new TextField { X = Pos.Left (label), Y = Pos.Bottom (label), Width = 20 };
    w.Add (label, tfRequest);

    label = new () { X = Pos.Right (tfRequest) + 1, Y = Pos.Top (tfRequest) - 1, Text = "E_xpectedResponseValue:" };
    var tfValue = new TextField { X = Pos.Left (label), Y = Pos.Bottom (label), Width = 6 };
    w.Add (label, tfValue);

    label = new () { X = Pos.Left (tfValue) + label.Text.Length, Y = Pos.Top (tfValue) - 1, Text = "_Terminator:" };
    var tfTerminator = new TextField { X = Pos.Left (label), Y = Pos.Bottom (label), Width = 4 };
    w.Add (label, tfTerminator);

    cbRequests.SelectedItemChanged += (s, e) =>
                                      {
                                          if (cbRequests.SelectedItem == -1)
                                          {
                                              return;
                                          }

                                          string selAnsiEscapeSequenceRequestName = scrRequests [cbRequests.SelectedItem];
                                          AnsiEscapeSequenceRequest selAnsiEscapeSequenceRequest = null;

                                          switch (selAnsiEscapeSequenceRequestName)
                                          {
                                              case "CSI_SendDeviceAttributes":
                                                  selAnsiEscapeSequenceRequest = AnsiEscapeSequenceRequestUtils.CSI_SendDeviceAttributes;

                                                  break;
                                              case "CSI_ReportTerminalSizeInChars":
                                                  selAnsiEscapeSequenceRequest = AnsiEscapeSequenceRequestUtils.CSI_ReportTerminalSizeInChars;

                                                  break;
                                              case "CSI_RequestCursorPositionReport":
                                                  selAnsiEscapeSequenceRequest = AnsiEscapeSequenceRequestUtils.CSI_RequestCursorPositionReport;

                                                  break;
                                              case "CSI_SendDeviceAttributes2":
                                                  selAnsiEscapeSequenceRequest = AnsiEscapeSequenceRequestUtils.CSI_SendDeviceAttributes2;

                                                  break;
                                          }

                                          tfRequest.Text = selAnsiEscapeSequenceRequest is { } ? selAnsiEscapeSequenceRequest.Request : "";

                                          tfValue.Text = selAnsiEscapeSequenceRequest is { }
                                                             ? selAnsiEscapeSequenceRequest.ExpectedResponseValue ?? ""
                                                             : "";
                                          tfTerminator.Text = selAnsiEscapeSequenceRequest is { } ? selAnsiEscapeSequenceRequest.Terminator : "";
                                      };

    // Forces raise cbRequests.SelectedItemChanged to update TextFields
    cbRequests.SelectedItem = 0;

    label = new () { Y = Pos.Bottom (tfRequest) + 2, Text = "_Response:" };
    var tvResponse = new TextView { X = Pos.Left (label), Y = Pos.Bottom (label), Width = 40, Height = 4, ReadOnly = true };
    w.Add (label, tvResponse);

    label = new () { X = Pos.Right (tvResponse) + 1, Y = Pos.Top (tvResponse) - 1, Text = "_Error:" };
    var tvError = new TextView { X = Pos.Left (label), Y = Pos.Bottom (label), Width = 40, Height = 4, ReadOnly = true };
    w.Add (label, tvError);

    label = new () { X = Pos.Right (tvError) + 1, Y = Pos.Top (tvError) - 1, Text = "_Value:" };
    var tvValue = new TextView { X = Pos.Left (label), Y = Pos.Bottom (label), Width = 6, Height = 4, ReadOnly = true };
    w.Add (label, tvValue);

    label = new () { X = Pos.Right (tvValue) + 1, Y = Pos.Top (tvValue) - 1, Text = "_Terminator:" };
    var tvTerminator = new TextView { X = Pos.Left (label), Y = Pos.Bottom (label), Width = 4, Height = 4, ReadOnly = true };
    w.Add (label, tvTerminator);

    var btnResponse = new Button { X = Pos.Center (), Y = Pos.Bottom (tvResponse) + 2, Text = "_Send Request", IsDefault = true };

    var lblSuccess = new Label { X = Pos.Center (), Y = Pos.Bottom (btnResponse) + 1 };
    w.Add (lblSuccess);

    btnResponse.Accepting += (s, e) =>
                             {
                                 var ansiEscapeSequenceRequest = new AnsiEscapeSequenceRequest
                                 {
                                     Request = tfRequest.Text,
                                     Terminator = tfTerminator.Text,
                                     ExpectedResponseValue = string.IsNullOrEmpty (tfValue.Text) ? null : tfValue.Text
                                 };

                                 bool success = Application.Driver!.TryWriteAnsiRequest (
                                                                                         Application.MainLoop!.MainLoopDriver,
                                                                                         ref ansiEscapeSequenceRequest
                                                                                        );

                                 tvResponse.Text = ansiEscapeSequenceRequest.AnsiEscapeSequenceResponse?.Response ?? "";
                                 tvError.Text = ansiEscapeSequenceRequest.AnsiEscapeSequenceResponse?.Error ?? "";
                                 tvValue.Text = ansiEscapeSequenceRequest.AnsiEscapeSequenceResponse?.ExpectedResponseValue ?? "";
                                 tvTerminator.Text = ansiEscapeSequenceRequest.AnsiEscapeSequenceResponse?.Terminator ?? "";

                                 if (success)
                                 {
                                     lblSuccess.ColorScheme = Colors.ColorSchemes ["Base"];
                                     lblSuccess.Text = "Success";
                                 }
                                 else
                                 {
                                     lblSuccess.ColorScheme = Colors.ColorSchemes ["Error"];
                                     lblSuccess.Text = "Error";
                                 }
                             };
    w.Add (btnResponse);

    w.Add (new Label { Y = Pos.Bottom (lblSuccess) + 2, Text = "Send other requests by editing the TextFields." });

    return w;
}
```

Terminal.Gui/ConsoleDrivers/CursesDriver/CursesDriver.cs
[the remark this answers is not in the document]
BDisp: If `EscSeqUtils.DecodeEscSeq` cannot recognize yet the ANSI escape sequence as valid, then the `ConsoleKeyInfo []? IncompleteCkInfos` will store it and insert at the beginning of next key get from the console until a valid ANSI escape sequence is valid. So, I recommend to check the `IncompleteCkInfos` value before discard any invalid ANSI escape sequence. I don't believe there are invalid ANSI escape sequence but only invalid requests, which can result with no response at all and thus discard the related request from the pending queue.

tznind:
> If EscSeqUtils.DecodeEscSeq cannot recognize yet the ANSI escape sequence as valid, then the ConsoleKeyInfo []? IncompleteCkInfos will store it and insert at the beginning of next key get from the console 

This is also how AnsiParser<T> works.  It uses a state machine such that no matter how split the input order is, the resolution remains the same.

I feel like we are duplicating work here, after this is merged we should work to empower the ansi parser and retire `DecodeEscSeq` and some of the other stuff in `EscSeqUtils`.

This diagram may help explain how it works (see below).  The diagram is for v2 but the parser is basically implemented in the same way for the normal `ConsoleDriver` classes 

- Parser filters input stream - everything read from console goes straight to parser
  - Sometimes parser holds onto keys (see `IHeld`)
  - Sometimes parser releases multiple keys at once
  - Always parser has 'state' i.e. in response etc
- Parser operates in the native type that each input stream is provided e.g. `InputRecord` for win and `ConsoleKeyInfo` for net
- Parser can be managed in main loop to force release at any time
  - for example if see `Esc` then `[` but its now 50 ms since see anything so is probably just user typing this sequence.
- When parser releases keys they get processed downstream as normal e.g. `Esc`, `[` would be released and processed exactly as if parser were not in the pipeline
  - This 'stream filtering' approach means parser doesnt have to raise key events etc it just realizes mistake and dropps the events back into input stream at next chance

image

BDisp:
> I feel like we are duplicating work here, after this is merged we should work to empower the ansi parser and retire `DecodeEscSeq` and some of the other stuff in `EscSeqUtils`.
> 
I agree but ensure that the `EscSeqUtils` unit tests wont fail with the `AnsiParser`.

>   * for example if see `Esc` then `[` but its now 50 ms since see anything so is probably just user typing this sequence.
> * When parser releases keys they get processed downstream as normal e.g. `Esc`, `[` would be released and processed exactly as if parser were not in the pipeline
>   
>   * This 'stream filtering' approach means parser doesnt have to raise key events etc it just realizes mistake and dropps the events back into input stream at next chance
> 
Normally check if the console has no key available then for sure it's a user typing a sequence, otherwise it's really a sequence.

> This is the class I wrote that replaces all the complex mouse handling cki stuff with a single tiny class. Because AnsiParser has already isolated the response we can just operate on the complete `string`
> 
Great.

tznind:
> Normally check if the console has no key available then for sure it's a user typing a sequence, otherwise it's really a sequence.

We are broadly agreed :) , however this is actually one of the things I did away with.  

What matters is the time the key was read at, not the available/not available state of the input pipe.  Because a pipe could get momentarily blocked or anything at any time.  

So this is why we have `public DateTime StateChangedAt { get; private set; }` so we can use realtime as a measure for bailing out of a response not the input pipeline state.

BDisp:
> We are broadly agreed :) , however this is actually one of the things I did away with.
> 
> What matters is the time the key was read at, not the available/not available state of the input pipe. Because a pipe could get momentarily blocked or anything at any time.
> 
But the user typing would not be lost, right?

> So this is why we have `public DateTime StateChangedAt { get; private set; }` so we can use realtime as a measure for bailing out of a response not the input pipeline state.

But this bailing out of a response will be what the user was really typing, right? Sorry for these silly questions, but I'm learning about these new processing.

tznind:
You are exactly right, because parser holds the full input event object (`T`), when asked to release then those events can be fed into the downstream process as normal.  So they surface as normal.

PR #3409, opened 2024-04-16
Abandoned - Fixes #3407. `Border` Subviews

tig:
## Fixes

- Fixes #3407

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:

- [ ] I've named my PR in the form of "Fixes #issue. Terse description."
- [ ] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [ ] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [ ] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [ ] My changes generate no new warnings
- [ ] I have checked my code and corrected any poor grammar or misspellings
- [ ] I conducted basic QA to assure all features are working

Issue #2307, opened 2023-01-27
Global overlay drawing system (e.g. for drag and drop)


tznind:
**Is your feature request related to a problem? Please describe.**

Sometimes you want to be able to paint arbitrary stuff over the top of other views.  For example during drag and drop.

### Use cases

The primary use case is for 'drag and drop' style effects e.g.

- Showing a representation of a file as you drag from one box to another
- Showing a tab as you drag around the screen to drop into another `TabView`

For example after dragging a tab in VisualStudio you see the title in 2 places.  Its original location and wherever the cursor is.

![shot-2023-01-27_19-49-19](https://user-images.githubusercontent.com/31306100/215184137-ae8ab1a8-3a6b-4bd2-ae01-be34c9892c8d.jpg)

But there are other potential use cases e.g.

- Holding down a button to show a tutorial overlay over the UI.

**Describe the solution you'd like**

The drawing would need to happen **last** after all other views had drawn for the loop iteration.  

When a new drawing/location is set (e.g. after the mouse moves) then the old contents would need to be invalidated (so that dragging didn't leave a 'trail').

Performance would have to be good so maybe have some kind of inbuilt throttling or back buffer.  Especially relevant if redrawing on Mouse Move events.

**Alternatives Considered**
It would probably be possible to have the rendered content be a `View` itself but this feels hacky and would likely interfere with detecting what the mouse was released over (e.g. because the animation follows the mouse)

tig:
Related:

* #2539 
* #2505 
* #2491 
* #2483 
* #2478 
*

tig:
Closing as fixed by the addition of `Popovers`.  @tznind feel free to re-open if you don't believe this can be implemented via that.

PR #3798, opened 2024-10-15
Fixes #3761, #2886, #3780, #3485, #3622, #3413, #2995 - Draw  and Layout performance/correctness

### LINKED ISSUE #2886: Address Redraw Performance Issues
some of what i am writing here is based on my assumptions on how this is implemented. this feature request is mostly a note to myself, i will investigate further and later fork and create a pull request. 

Many inputs e.g. keyboard strokes, mouse events within under one expected frame time cause the lib to refresh the screen one time each. these can create lag in some terminals to the extend of needing a couple seconds to catch up on unrealisticly high input loads e.g. smashing the keyboard on the text input field. 

mark dirty (as requiring an update), and have a threaded updater that checks within a frame time if we are dirty. only in that case emit a redrawing. 
Drawback: we may incur input lag up to one frame as well as how precise sleeping, waiting is

1) As input output is blocking we cannot proactively skip some input and render on the next as we have no guarantee when this will be
2) if input output on all or some platforms allows timeouts, we can instead schedule the update in the main thread upon reaching that timeoujt
3) instead of emitting all redrawings to tty we could also draw to a buffer and after redrawing is complete only emit the differences to tty, however this does not fix that some redrawing procedures might be computationally expensive

**Additional context**
I plan to investigate further and implement the solution suggested here myself as well as add buffering as an optional feature (alternative 3). i also will investigate alternative 2 and may only fall back to threading only if the current tty does not support timeouts. I also do plan to make all these optional features and keep the default behaviour intact

### LINKED ISSUE #2995: Auto-join of `Border` is not always working
Run Frames Demo:

image

I saw this a long time ago, but all of the discussions of "finishing" Adornments (#2994) reminded me it's broken.

### LINKED ISSUE #3413: Improve redraw performance by making `Clip` be a `Region` instead of `Rectangle`

### LINKED ISSUE #3485: Remove need to explicitly call `SetNeedsLayout` -> Have `Pos`/`Dim` use events
`Pos.Align` pushes the envelope on being the most sophisticated `Pos`/`Dim` object. The fact that it has multiple configuration properties and the use-cases mean there are interesting cases where a dev will want to modify a `PosAlign` object.

For example, if `X` is `Pos.Align`, a dev may want to `posAlign.AlignmentModes &= ^AlignmentMode.AddSpaces` without creating a new instance an assinging it to X. If they do that, there's no way for `View` to know it the state was changed to force a layout.

`View.LayoutSubviews` is (rightly, IMO) `internal` so a dev has really no way to indicate a layout is needed other than forcing `View.X` to be set with a new instance of `PosAlign`.

This is actually not a common case, so I'm not going to fix it now. However, thinking through how it would work made me realize that if `Pos` and `Dim` supported `IPropertyNotify` (or similar), and on `set_X` we subscribed, `View` would always just know layout was needed. There would be no need to ever explicitly call `SetLayoutNeeded` and would radically remove the times where calling `LayoutSubviews` was need.  

This could be a big perf win (I have not measured the # of times layout happens unnecessarily in a long time, but I did it ages ago and it's an issue). It could also be a correctness win.

Not high pri by any means, but something to noodle on.

### LINKED ISSUE #3622: Apps should never reference `ConsoleDriver`
_Originally posted by @dodexahedron in https://github.com/gui-cs/Terminal.Gui/pull/3615#discussion_r1685170103_

>  I have never cared for this method's name, either.

> Typically, an instance method on a type that has a simple verb name results in that verb being performed on the instance of the type. In other words, a `Move` instance method implies that the `ConsoleDriver` is being moved, which of course is not what's happening.

> This repositions ConsoleDriver's equivalent of cursor position without moving the actual cursor, so is really just moving a 2D index for internal operations.

> The method seems unnecessary anyway, however. The Row and Col properties are both public auto-properties. Why not make their set accessors public if the consumer can already effectively change them directly?

I semi-agree. My take is `ConsoleDriver` details should not be exposed to app code, AT ALL.

All of these operations (`AddRune`, `Move`, `GetAttribute`, ...) should be on either `View` or `Application`.

This issue is to track this.

### LINKED ISSUE #3761: Latest version of all 3 drivers call refresh constantly
**Describe the bug**

refresh-bug
_Clicking the mouse causes about 68 screen refreshes (it does eventually stop)_

**To Reproduce**
This changed recently.

1. Add a debug counter to `UpdateScreen` 
2. Click mouse

**Expected behavior**
Only redraw when something has changed.

**Screenshots**

**Additional context**

### LINKED ISSUE #3780: Content Scrolling scenario clear the contents on mouse move or key press.
@tig do you have any idea about this?

WindowsTerminal_f49ayTB37Z

tig:
## Fixes

- Fixes #3761 
- Fixes #2886
- Fixes #3780 
- Fixes #3485
- Fixes #3622
- Fixes #3413
- Fixes #2995
- Improves #3209

## Related

- #2307

## Proposed Changes/Todos

- [x] Decouple layout from draw
- [x] Refactor `View.LayoutSubviews` etc... to make more testable and clearer to use. Introduce `View.Layout()` as the primary API that builds on `View.SetRelativeLayout` and `View.LayoutSubviews`. Move from depending on `IsInitialized` to `NeedsLayout` as the signal that layout has been performed.
- [x] Refactor `View.Draw` to be more granular and use proper event patterns.
- [x] Minimize when Draw/OnDrawContent are called
- [x] Enforce that Clear rarely happens
- [x] Add `View.SetAttribute`; make `Driver.SetAttribute` internal
- [x] Add non-rectangular clip regions - #3413
- [x] Refine non-rectangular clip regions - some edge cases not working like shadowview
- [x] Undo "fix" to `Enabled` to enable disabled views to stay disabled even if superview changes; use an event instead of hacky logic.
- [x] Remove all refs to `Driver.Move/AddStr` 
- [x] Ensure drivers minimize full repaints
- [x] Test other drivers
- [x] Fix `LineCanvas` and auto-join of lines/`Border`

## Known Issues that will be fixed later

- [ ] `TabView` works, but has horrible performance and strange redraw. Unit tests are disabled. This will be fixed with https://github.com/gui-cs/Terminal.Gui/pull/3808
- [ ] `ScrollView` is non-functional. This will be fixed with https://github.com/gui-cs/Terminal.Gui/pull/3498
- [ ] Titles get messed up - This will be fixed once `Border` starts actually using subviews for the border lines and title. https://github.com/gui-cs/Terminal.Gui/pull/3409
- [ ] I have a hack in place right now that causes all views in a view hierarchy that have `SuperViewRendersLineCanvas` set to redraw anytime any of them changes. This negates the perf work I've done when `SuperViewRendersLineCanvas` is enabled. 

## Pull Request checklist:


tig:
@BDisp, @tznind, @dodexahedron  or anyone else who'd like to comment... 

As I've worked this PR, I've learned a bunch of things, and gotten clarity on other issues that have bugged me for a long time.

Before I go further, I want to share my thoughts in writing. This will solidify my thinking and enable y'all to challenge and debate it.

Some of the changes I've made, and want to continue to make are significant. While they don't fundamentally change the TG programming model, they do require changes to how devs will develop against it. Some will make porting from v1 more challenging. 

## Layout vs Draw

First, TG has always been confused about the relationship between Layout and Draw. This confusion got worse and worse over time. The addition of `View.AutoSize` made it even worse. As did more sophisticated use-cases including overlapped Views, multi-threading, and Adornments. The state of `v2_develop` is horrific in this regard. Hence #3761 etc... that lead to this PR.

The codebase is full of convoluted code that attempts to ensure things get laid out then drawn. Property `_set` methods call a combination of `SetNeedsDisplay`, `SetLayoutNeeded`, and `LayoutSubviews`. The default `OnDrawContent` forces layout because it can't assume layout has already happened. The mainloop uses brute-force-overkill to force layouts and draws and ends up calling `Consoledriver.Refresh` way too frequently. Plus `View.Clear` is called constantly.  Some Views (`TabView`) re-layout elements DURING draw!

We have to fix this at a fundamental level or the library will continue to get bogged down. 

Assertions:

* **Layout and Draw must be decoupled** - Layout code should not call draw code. Draw code should not call layout code. 
* **Layout happens first; then Draw** - This sounds obvious, but there are multiple places where the library or built-in Views either blindly do the opposite or attempt to do layout mid-draw. 

## Asynchronous Layout vs Explicit/Synchronous

Way back, the only place that called `SetRelativeLayout` and `LayoutSubviews` was the mainloop. MG's design for gui-cswas these methods were to be called each mainloop iteration. They were `internal` for this reason. Overtime, though, folks started using them more and more from within View implementations or app code. 

The names of these methods leads to confusion as well. 

Assertion:

* **Layout is Asynchronous** - In the vast majority of cases developers should not call `SetRelativeLayout` or `LayoutSuviews` directly, but should just signal layout is needed with `SetLayoutNeeded()`. We will introduce a new API (which is really just a refactor of the currently `private` `LayoutSubview()`: `public bool Layout (Size contentSize)` that first calls `SetRelativeLayout` and then `LayoutSubviews`. Devs can use this in the rare cases where an immediate layout is required.

Great, but what about cases where a dev DOES want immediate layout? Should it be implicit (as it is today) or explicit?

### Option 1

* **`Frame`, `X`, `Y`, `Width`, and `Height` are non-deterministic if `IsLayoutNeeded` is `true`** - This is a BIG one.  It means this code will fail on the 2nd assert:

```cs
  View v = new ();
  Assert.Equal (0, v.Frame.Height);
  v.Height = 1;
  Assert.Equal (1, v.Frame.Height);
```

This will not fail:

```cs
  View v = new ();
  Assert.Equal (0, v.Frame.Height);
  v.Height = 1;
  Assert.Equal (0, v.Frame.Height);
  v.Layout ();  
  Assert.Equal (1, v.Frame.Height);
```

### Option 2

* **`Frame`, `X`, `Y`, `Width`, and `Height` are non-deterministic if `IsLayoutNeeded` is `true`** 
* **Setting `X`, `Y`, `Width`, and `Height` to a `Pos/DimAbsolute` will call `Layout()` (causing `IsLayoutNeeded` to be `false`) IFF the other 3 are also `Pos/DimAbsolute`.
* **Setting `Frame`, will cause `will call `Layout()` (causing `IsLayoutNeeded` to be `false`).

The downside of this is that devs may be confused:

```cs
  View v = new ();
  Assert.Equal (0, v.Frame.Height);
  v.Height = 1;
  Assert.False (v.IsLayoutNeeded());
  Assert.Equal (1, v.Frame.Height); // works because `1` is `Dim.Absoulte(1)`

  v.Width = Dim.Percent(50);
  Assert.True (v.IsLayoutNeeded());
  Assert.Equal (0, v.Frame.Width); // still 0
```

The reality is the above code is really only valid in UNIT TESTS. No dev will ever do this IRL. 

IOW, if we don't call `SetLayout` in the `Pos/DimAbsoulte` cases, all unit tests will need to explicitly call `Layout` which is a PITA.

In this PR I started with Option 1. It felt pure. But it means I have to touch 100s of Unit tests that simply are using absolute values and add calls to `v.Layout()`. So now I'm going with Option 2.

.... More later. This is enough for now...

BDisp:
As I remember the layout was always done before drawn. When the drawn was called alone and `IsLayoutNeeded() == true` the layout was being performed first. I also agree with the option 2. The #3761 really needs to be fixed. Good work.

tznind:
Awesome!

This will definitely help with our flicker issue and bring some structure
to layout - which has been a bit unstructured and opaque.

Regarding redraw - i think we can also eliminate redraws of identical
screen content.  I did this in the sixel branch to work around the constant
refreshes and it seemed to work.  Basically on screen redraw you record the
final string sent to console and if it's the same you bail.  This could be
a 'last line of defence' against this kind of thing.

tznind:
I am having a look at TabView

The reason that it does layout in draw is because of https://github.com/gui-cs/Terminal.Gui/pull/2980

Prior to this PR there were 2 views only (content view and tab row view).  The row view drew all the characters direct from the driver e.g. scroll indicators, tabs etc.

After #2980 these all became views (`_leftScrollIndicator` etc).  But the positioning stayed where it was:

```csharp
// draw scroll indicators

// if there are more tabs to the left not visible
if (host.TabScrollOffset > 0) {
	Move (0, y);

	// indicate that
	Driver.AddRune (CM.Glyphs.LeftArrow);
```

Became

```csharp
// draw scroll indicators
// if there are more tabs to the left not visible
	if (_host.TabScrollOffset > 0) {
		_leftScrollIndicator.X = 0;
		_leftScrollIndicator.Y = y;

		// indicate that
		_leftScrollIndicator.Visible = true;
		// Ensures this is clicked instead of the first tab
		BringSubviewToFront (_leftScrollIndicator);
		_leftScrollIndicator.Draw ();
	} else {
		_leftScrollIndicator.Visible = false;
	}
```

I haven't worked with it after the change but will have a bash at creating a layout method where this happens instead.

BDisp:
I'm also having a look at this, since I implemented the new Border adornment. The drawing issue I already fixed and I'm only dealing with key navigation, right now.

tznind:
Ok awesome I will hold off then - no point in duplicating 👍 thanks!

tig:
IMO TabRowView should just be a SuperView for a set of Views that have their borders set. 

We should enable Border to support different line styles for each side by making each side be a Line that's a subview of Border. 

We have enough examples of adornments with subviews working well (and the draw perf pr makes it even better) to have confidence that Border can finally get rid of all the custom rendering and just use subviews. 

Even better would be each Tab to just be a View with one side having a thickness of 3. We'd need to have Border support the ability to set the start offset of the title.  

There should be no need for TabView  to use LineCanvas directly.

BDisp:
@tig I think this code isn't totally correct because it's avoiding the first view of the `indicies` to get focus if it contains the `_previouslyFocused`. This is avoiding the tab getting focus if a previously subview had the focus before.

```cs
    /// <summary>
    ///     INTERNAL API to restore focus to the subview that had focus before this view lost focus.
    /// </summary>
    /// <returns>
    ///     Returns true if focus was restored to a subview, false otherwise.
    /// </returns>
    internal bool RestoreFocus ()
    {
        View [] indicies = GetFocusChain (NavigationDirection.Forward, null);

        if (Focused is null && _previouslyFocused is { } && indicies.Contains (_previouslyFocused))
        {
            if (_previouslyFocused.SetFocus ())
            {
                return true;
            }

            _previouslyFocused = null;
        }

        return false;
    }

```

tznind:
> IMO TabRowView should just be a SuperView for a set of Views that have their borders set.

Yes this makes sense. And is why tab view now uses subviews. Wasn't suggesting trying to go back to manual draw just that the reason for all the draw layout was the introduction of subviews.

Current code is half way through a journey that is all

BDisp:
> > IMO TabRowView should just be a SuperView for a set of Views that have their borders set.
> 
> Yes this makes sense. And is why tab view now uses subviews. Wasn't suggesting trying to go back to manual draw just that the reason for all the draw layout was the introduction of subviews.
> 
> Current code is half way through a journey that is all

@tznind can you then please go forward with this approach?

tig:
We now have pretty sophisticated benchmarking built into UICatalog:

0E7uhiP 1

```ps
> UICatalog.exe --help

Description:
  A comprehensive sample library for Terminal.Gui

Usage:
  UICatalog [<scenario>] [options]

Arguments:
  The name of the scenario to run. [default: none]
  
Options:
  -b, --benchmark                                                 Enables benchmarking.
  -f, --file <file>                                               The file to save benchmark results to. If not specified with --benchmark, the results will be  displayed in a TableView.
  -d, --driver <CursesDriver|FakeDriver|NetDriver|WindowsDriver>  The ConsoleDriver to use.
  --version                                                       Show version information
  -?, -h, --help                                                  Show help and usage information
```

`Scenario` now has an API for providing "Demo Keys":

```cs
    public virtual List<Key> GetDemoKeyStrokes () => new List<Key> ();
```

For example, CharMap does this:

```cs
    public override List<Key> GetDemoKeyStrokes ()
    {
        var keys = new List<Key> ();

        for (int i = 0; i < 200; i++)
        {
            keys.Add (Key.CursorDown);
        }

        // Category table
        keys.Add (Key.Tab.WithShift);

        // Block elements
        keys.Add (Key.B);
        keys.Add (Key.L);

        keys.Add (Key.Tab);

        for (int i = 0; i < 200; i++)
        {
            keys.Add (Key.CursorLeft);
        }
        return keys;
    }
```

So, 

4eYS6xt 1

BDisp:
@tig I only have one test failing `public void ProcessKey_Down_Up_Right_Left_Home_End_PageDown_PageUp` because the navigation is forcing using the `RestoreFocus` method and thus avoid the `SelectedTab` being focused and always focus the `_previouslyFocused`. I understand the use of this method when called explicitly, like the unit tests does, but I don't understand why calling it in a normal navigation, unless in a `TabGroup`. Can you help me on this, please.

tig:
> @tig I only have one test failing `public void ProcessKey_Down_Up_Right_Left_Home_End_PageDown_PageUp` because the navigation is forcing using the `RestoreFocus` method and thus avoid the `SelectedTab` being focused and always focus the `_previouslyFocused`. I understand the use of this method when called explicitly, like the unit tests does, but I don't understand why calling it in a normal navigation, unless in a `TabGroup`. Can you help me on this, please.

Not entirely sure what you mean. Maybe submit your work as a PR against this PR and I can take a look?

BDisp:
Done in https://github.com/tig/Terminal.Gui/pull/41. The only way to focus the selected tab is if the view container `CanFocus` is false or if a view is currently selected in the container and pressing cursor up.

Terminal.Gui/Application/Application.Mouse.cs
```diff
@@ -249,6 +249,15 @@ internal static void RaiseMouseEvent (MouseEventArgs mouseEvent)
                 View = deepestViewUnderMouse
             };
         }
+
+        // TODO: There should be a clearner way to to this
+        // Add a timeout to ensure at least one application iteration happens after each
+        // mouse event, enabling Refresh to be called.
+        Application.AddTimeout(new (0,0,0,0, 200),
```
BDisp, line 256: I really don't understand the necessity of this call here. Any view that will handle the mouse is responsible to call `SetNeedsLayout` to recalculate Pos/Dim and `SetNeedsDisplay` or `SetSubViewNeedsDisplay` if they only needed to redraw. Calling `SetNeedsLayout` automatically force the necessity of calling `SetNeedsDisplay`. Calling `Refresh` in this `AddTimeout` will decrease the performance because will call `Refresh` more times than the necessary. Not all the mouse event need to call `Refresh` at all, unless the view needs to design effects on mouse move, where the view is responsible to set the above proprieties if needed. I think calling `Refresh` through the `RunIteration` is enough.

BDisp (approved):
I like the code cleanup you made. Very clear. Only one observation.

Terminal.Gui/Application/Application.Mouse.cs
tig: Here's the problem that needs to be solved, that this works around:

In this PR, `Daw` called only from `Refresh` which is ONLY called at the end of an iteration. (There are some exceptions where some code still calls `View.Draw` or `Applicaiton.Refresh` directly, but those are either bugs or very special cases).

`Refresh` does this:

```cs
...

        Stack<Toplevel> redrawStack = new (TopLevels);

        while  (redrawStack.Count > 0)
        {
            Toplevel? tlToDraw = redrawStack.Pop ();

            if (clear || forceRedraw)
            {
                tlToDraw.SetNeedsDisplay ();
            }

            if (!(!View.CanBeVisible (tlToDraw) || tlToDraw is { NeedsDisplay: false, SubViewNeedsDisplay: false }))
            {
                tlToDraw.Draw ();

                if (redrawStack.TryPeek (out var nexToplevel))
                {
                    nexToplevel.SetNeedsDisplay();
                }
            }
        }

        Driver?.Refresh ();
```

This logic is currently not complete, because I was still debugging something, but it essentially ensures the top-most Toplevel draws last.

The problem is code like this:

```cs
Toplevel app = new () { "app"};

Button btn = new () { Title = "_Open Dialog" );
app.Add(btn);
btn.Accepting += (s,e) => { Application.Run (dlg); };

Application.Run(app);
```

If `btn` is pressed with the hotkey (`O`) everything is fine.  

- User presses `O` down
- `Application.RaiseKeydown` is called by the driver from `MainLoop.RunIteration ()`
  - Causes `btn.Accepting` -> `Application.Run(dlg)` 
  - `Run` calls `Begin` changes focus such that `btn` no longer has focus so it and `app` get `NeedsDisplay = true`.
  - `Run` calls `RunLoop` which BLOCKS waiting for the driver to generate an event
- User releases `O` key causing the driver to unblock
- `Application.RaiseKeydown` is called by the driver from `MainLoop.RunIteration ()`
  - The iteration calls `Refresh` and draws everything correctly.

Yay.

HOWEVER, If `btn` is pressed with the MOUSE:

- User presses `btn`
- `Application.RaiseMouseEvent` is called with `Button1Pressed` by the driver from `MainLoop.RunIteration ()`
- `Application.RaiseMouseEvent` is called with `Button1Released` by the driver from `MainLoop.RunIteration ()`
- `Application.RaiseMouseEvent` is called with `Button1Clicked` by the driver from `MainLoop.RunIteration ()`
  - Causes `btn.Accepting` -> `Application.Run(dlg)` 
  - `Run` calls `Begin` changes focus such that `btn` no longer has focus so it and `app` get `NeedsDisplay = true`.
  - `Run` calls `RunLoop` which BLOCKS waiting for the driver to generate an event

The Dialog is not rendered because we're blocked waiting for a new event from the driver. 

Move the mouse and `Refresh` gets called and all is good.
  
This is really hard to debug because `WindowsDriver` generates events on focus changes and each time you step in the debugger, it re-activates the terminal generating iterations. 

Anyway, here's what we don't want to do to fix this:

- Cause Iterations to happen all the time. 
- Call `Refresh` synchronously from `Run/Begin` (see rule above about only doing draw async).

The best solution I could find was to generate a Timer event after each mouse event.

I'm very much open to other ideas.

> Calling Refresh in this AddTimeout will decrease the performance because will call Refresh more times than the necessary.

This is no longer true with this PR because Refresh is a NOP if there's no `NeedsDisplay = true` anywhere.

BDisp: I already saw that a view that call SetFocus before call Application.Run loose the focus in the Add method. I don't know why.

BDisp: > The Dialog is not rendered because we're blocked waiting for a new event from the driver.
> 
When a new modal toplevel is loaded the `Begin` method must call the `Refresh` like it also call the `PositionCursor`. This is mandatory.
```cs
        toplevel.OnLoaded ();

        Refresh ();

        if (PositionCursor ())
        {
            Driver?.UpdateCursor ();
        }

```

> Move the mouse and `Refresh` gets called and all is good.
> 
If you call the `Refresh` as above move the mouse isn't necessary.

> This is really hard to debug because `WindowsDriver` generates events on focus changes and each time you step in the debugger, it re-activates the terminal generating iterations.
> 
Use `NetDriver` or `CursesDriver` to debug it because they don't generate events on focus changes.

> Anyway, here's what we don't want to do to fix this:
> 
> * Cause Iterations to happen all the time.

It's preferable to execute all the timers and idle handlers than having the drawing occurring all the time without need, which I already saw it's now a NOP which is very useful.

> * Call `Refresh` synchronously from `Run/Begin` (see rule above about only doing draw async).
> 
I don't agree not calling the `Refresh` to draw in the `Begin` method. If the rule must be imposed then the PositionCursor method mustn't also be called. The problem you describe that it's needed to move the mouse when running a `Dialog` also occurs when you start running the UICatalog because Begin doesn't call `Refresh`.

> The best solution I could find was to generate a Timer event after each mouse event.
> 
Not needed in a regular mouse event. If the Button only respond to the click event then the pressed and release only run `MainLoop.RunIteration ()` if there are timers or idle handlers to be executed and drawing if it's needed, like the button effects.

> I'm very much open to other ideas.
> 
I'm tying to give some.

> > Calling Refresh in this AddTimeout will decrease the performance because will call Refresh more times than the necessary.
> 
> This is no longer true with this PR because Refresh is a NOP if there's no `NeedsDisplay = true` anywhere.

I know and I like the solution added. It's the best place to have this control instead of in the `MainLoop.RunIteration`. But I think the `AddTimeout` in the `RaiseMouseEvent` is unnecessary.

tig: > When a new modal toplevel is loaded the Begin method must call the Refresh. This is mandatory.

The fact you say this means you are not understanding what I've done in this PR. 

Please dive deeper and see the purity of having Refresh only happen at the end of an iteration is the way to go.

tig: Actually, I think I just found a better solution that's more generic.

Instead of doing this in the mouse handler, add an Idle event at the end of `Application.Begin`:

```cs
        // Force an Idle event so that an Iteration (and Refresh) happen.
        Application.Invoke(() => {});
```

BDisp: I now understand, sorry. I like more this approach because it don't impose any timeout and wakeup immediately the mainloop. But you also can remove the call t the `PositionCursor` in the `Begin` method because it will be called in the `RunIteration`.
Another doubt I have is when a view only call the `SetNeedsDisplay` and not the `SetNeedsLayout` will the screen been drawing?

tig: Re: cursor - I'm intentionally ignoring working on that until we get to the PR for the issue on it (mobile so can't look up issue #). 

Re: when to call SetNeedsDisplay vs SetNeedsLayout. Not sure about that yet. Been thinking a lot about it though. Would love suggestions.

BDisp: If a view only need to redraw because something changed on a fixed size (all area or on a area part) then only call `SetNeedsDisplay` is enough. If a view need to redraw because something changed the position or the size then it's needed to call `SetNeedsLayout` which will impose a mandatory `SetNeedsDisplay` call.

tig:
I renamed "SetNeedsDisplay" to "SetNeedsDraw". Sorry, not sorry.

tig:
To help visually track down situations where things are getting redrawn when they shouldn't be, I've added `ViewDiagnostics.DrawIndicator`:

g7z4wjX 1

To illustrate the usefulness of this, here's the `Navigation` scenario shown in the GIF above, with the hack enabled that causes Overalpped Views to be re-drawn constantly so that they appear correctly.

E2wVG4P 1


FWIW, the only "right" way to make overlapped views draw efficiently will be to implement non-rectangular clip regions.

For now this hack is enabled in `View.DrawSubviews`, and it only comes into play in cases like the `Navigation` Sceanario where `Overlapped` views are used.

```cs
    public void DrawSubviews ()
    {
        if (_subviews is null || !SubViewNeedsDraw)
        {
            return;
        }

#if HACK_DRAW_OVERLAPPED
        IEnumerable<View> subviewsNeedingDraw = _subviews.Where (view => (view.Visible && (view.NeedsDraw || view.SubViewNeedsDraw))
                                                                      || view.Arrangement.HasFlag (ViewArrangement.Overlapped));
#else
        IEnumerable<View> subviewsNeedingDraw = _subviews.Where (view => (view.Visible && (view.NeedsDraw || view.SubViewNeedsDraw)));
#endif

        foreach (View view in subviewsNeedingDraw)
        {
#if HACK_DRAW_OVERLAPPED
            if (view.Arrangement.HasFlag (ViewArrangement.Overlapped))
            {

                view.SetNeedsDraw ();
            }
#endif
            view.Draw ();

        }
    }
```

tig:
I've started implementing 

- #3413

- [x] Created a port of `System.Drawing.Region` that does not use any external libs and only supports ints.
- [x] Ported a bunch of unit tests for `Region` from `System.Drawing.RegionTests`.
- [x] Changed `ConsoleDriver.Clip` to be `Region` instead of `Rectangle`
- [x] Hacked `View` and unit tests to work with this, but there's currently no support for anything but the Region containing a single rect.

Terminal.Gui/Application/Application.Mouse.cs
BDisp: > If a view need to redraw because something changed the position or the size then it's needed to call `SetNeedsLayout` which will impose a mandatory `SetNeedsDisplay` call.

I forgot to mention that in this case the views under the current view must be flag to being redraw as well, but of course you already know.

tig:
lXMhxsU 1

This vid demonstrates non-rectangular clip regions doing their magic!

dodexahedron:
Loving the work on this. 👍

Planning to take a good look at the code later today.

tig:
The current state is pretty good with the exception of:

- Shadows (actually Margin in general) are broken
- ScrollView is broken (and I won't fix... we have a new one coming)
- TabView is broken. Bdisp has a pr i haven't had a chance to look at. 
- Attribute setting is brute force. Need to really clean that up because Color comparison is a heavy op. 

I have not invested as much energy in unit tests as I've liked.

tig:
This is how this all works (from `drawing.md`):

------------------

### Drawing occurs each MainLoop Iteration

The @Terminal.Gui.Application MainLoop will iterate over all Views in the view hierarchy looking for views have their @Terminal.Gui.View.NeedsDraw property set. The @Terminal.Gui.View.Draw method will be called which, in turn:

1) Sets the clip to the view's Frame.
2) Draws the @Terminal.Gui.View.Border and @Terminal.Gui.View.Padding (but NOT the Margin).
3) Sets the clip to the view's Viewport.
4) Sets the Normal color scheme.
5) Calls Draw on any @Terminal.Gui.View.Subviews.
6) Draws @Terminal.Gui.View.Text.
7) Draws any non-text content (the base View does nothing.)
8) Sets the clip back to the view's Frame.
9) Draws @Terminal.Gui.View.LineCanvas (which may have been added to by any of the steps above).
10) Draws the @Terminal.Gui.View.Border and @Terminal.Gui.View.Padding Subviews (just the subviews). (but NOT the Margin).
11) The Clip at this point excludes all Subviews NOT INCLUDING their Margins. This clip is cached so @Terminal.Gui.View.Margin can be rendered later.
12) DrawComplete is raised.
13) The current View's Frame NOT INCLUDING the Margin is excluded from the current Clip region.

Most of the steps above can be overridden by developers using the standard [Terminal.Gui cancellable event pattern](events.md). For example, the base @Terminal.Gui.View always clears the viewport. To override this, a subclass can override @Terminal.Gui.View.OnClearingViewport to simply return `true`. Or, a user of `View` can subscribe to the @Terminal.Gui.View.ClearingViewport event and set the `Cancel` argument to `true`.

Then, after the above steps have completed, the Mainloop will iterate through all views in the view hierarchy again, this time calling Draw on any @Terminal.Gui.View.Margin objects, using the cached Clip region mentioned above. This enables Margin to be transparent.

### Declaring that drawing is needed

If a View need to redraw because something changed within it's Content Area it can call @Terminal.Gui.View.SetNeedsDraw. If a View needs to be redrawn because something has changed the size of the Viewport, it can call @Terminal.Gui.View.SetNeedsLayout.

### Clipping

Clipping enables better performance by ensuring on regions of the terminal that need to be drawn actually get drawn by the @Terminal.Gui.ConsoleDriver. Terminal.Gui supports non-rectangular clip regions with @Terminal.Gui.Region. @Terminal.Gui.ConsoleDriver.Clip is the application managed clip region and is managed by @Terminal.Gui.Application. Developers cannot change this directly, but can use @Terminal.Gui.Application.ClipToScreen, @Terminal.Gui.Application.SetClip(Region), @Terminal.Gui.View.ClipToFrame, and @Terminal.Gui.ClipToViewPort.

------------------

I am now working through what it would mean to have `Border` be treated like `Margin` above: Be excluded from the Clip of a View and rendered in a 2nd pass. This *should* make the whole LineCanvas thing work much more cleanly.

UICatalog/Scenarios/SimpleDialog.cs
```diff
@@ -0,0 +1,59 @@
+#nullable enable
+using Terminal.Gui;
+
+namespace UICatalog.Scenarios;
+
+[ScenarioMetadata ("SimpleDialog", "SimpleDialog ")]
```
BDisp, line 6: I only see an "A" on the top left.

UICatalog/Scenarios/SingleBackgroundWorker.cs
```diff
@@ -9,7 +9,7 @@ namespace UICatalog.Scenarios;
 
 [ScenarioMetadata ("Single BackgroundWorker", "A single BackgroundWorker threading opening another Toplevel")]
 [ScenarioCategory ("Threading")]
-[ScenarioCategory ("Overlapped")]
+[ScenarioCategory ("Arrangement")]
 [ScenarioCategory ("Runnable")]
 public class SingleBackgroundWorker : Scenario
```
BDisp, line 14: The `MenuBar` of the Worker window isn't showing and can't be closed.

UICatalog/Scenarios/SpinnerStyles.cs
```diff
@@ -28,7 +27,7 @@ public override void Main ()
 
         foreach (Type style in typeof (SpinnerStyle).GetNestedTypes ())
         {
-            styleDict.Add (i, new KeyValuePair<string, Type> (style.Name, style));
+            styleDict.Add (i, new (style.Name, style));
```
BDisp, line 30: Not all the glyphs are rendering well. See the SoccerHeader please.

UICatalog/Scenarios/SyntaxHighlighting.cs
```diff
@@ -282,8 +282,8 @@ private void ApplySyntaxHighlighting ()
         };
 
         _textView.TextChanged += (s, e) => HighlightTextBasedOnKeywords ();
-        _textView.DrawContent += (s, e) => HighlightTextBasedOnKeywords ();
-        _textView.DrawContentComplete += (s, e) => HighlightTextBasedOnKeywords ();
+        _textView.DrawingContent += (s, e) => HighlightTextBasedOnKeywords ();
```
BDisp, line 285: The color is only rendered when we click in the `TextView` on open.

Terminal.Gui/Views/TabView.cs
```diff
@@ -19,8 +18,8 @@ public class TabView : View
     /// <summary>This sub view is the 2 or 3 line control that represents the actual tabs themselves.</summary>
     private readonly TabRowView _tabsBar;
 
-    private Tab _selectedTab;
-    private TabToRender [] _tabLocations;
+    private Tab? _selectedTab;
+    private TabToRender []? _tabLocations;
```
BDisp, line 22: `TabView` is rendering badly again during the tab select changing.

UICatalog/Scenarios/TableEditor.cs
```diff
@@ -14,7 +14,6 @@ namespace UICatalog.Scenarios;
 [ScenarioCategory ("Controls")]
 [ScenarioCategory ("Dialogs")]
 [ScenarioCategory ("Text and Formatting")]
-[ScenarioCategory ("Overlapped")]
 public class TableEditor : Scenario
```
BDisp, line 17: Sometimes after open this scenario and navigating through the lines when I press escape to exit the main UICatalog window isn't redrawn totally but only half of the window. This happening with the window maximized. I have to click to redraw it.

BDisp: Sometimes is when I open the scenario it draw well and then it stay with a black screen.

Terminal.Gui/Views/TabView.cs
BDisp: It seems you merged an old version of `TabView`, before the changes I made,

Terminal.Gui/View/View.Drawing.cs
```diff
@@ -1,280 +1,282 @@
...
+            // Clear the viewport
+            // TODO: Simplify/optimize SetAttribute system.
+            DoSetAttribute ();
+            DoClearViewport ();
+
+            // Draw the subviews
+            if (SubViewNeedsDraw)
+            {
+                DoSetAttribute ();
+                DoDrawSubviews ();
```
BDisp, line 73: Subviews are drawn before the superview itself? Is this correct? `DoDrawContent` should be at first and then the subviews. Am I wrong?

UICatalog/Scenarios/AdvancedClipping.cs
```diff
@@ -0,0 +1,166 @@
+﻿using System.Text;
+using System.Timers;
+using Terminal.Gui;
+
+namespace UICatalog.Scenarios;
+
+[ScenarioMetadata ("AdvancedClipping", "AdvancedClipping Tester")]
+[ScenarioCategory ("AdvancedClipping")]
+public class AdvancedClipping : Scenario
```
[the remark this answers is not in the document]
tig: > One question. When we click on a window in this scenario should that window brough to front?

No, not by default because they don't have Overlapped set. Use ArrangmentEditor to change them and you'll see.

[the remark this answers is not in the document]
tig: > Another question. If `Application.Top` is moveable then the background should be cleared or set it to non-moveable.

Correct. This is an outstanding bug. I plan on fixing it later.

UICatalog/Scenarios/SimpleDialog.cs
tig: Work in progress.

UICatalog/Scenarios/SingleBackgroundWorker.cs
tig: Can you  please help me debug?

UICatalog/Scenarios/SpinnerStyles.cs
tig: Please open another issue. This exists in v2_develop.

UICatalog/Scenarios/SyntaxHighlighting.cs
tig: Does this occur in v2_develop?

Terminal.Gui/Views/TabView.cs
tig: Please ignore Tabview for now.

UICatalog/Scenarios/TableEditor.cs
tig: I'll take a look.. Thanks.

Terminal.Gui/View/View.Drawing.cs
tig: Please read `drawing.md`.

With non-rectangular Clip, views "above" now get drawn first. 

Then views "below" get drawn, but don't overwrite because clipping is actually doing its job.

UICatalog/Scenarios/SyntaxHighlighting.cs
BDisp: > Does this occur in v2_develop?

No.

Terminal.Gui/View/View.Drawing.cs
BDisp: Got it and thus only the visible sides are drawn. That's really impressive and have to be well articulated. When UpdateScreen is call all contents must be already defined and no background call mustn't corrupt it, with exception of sixel, of course. Good work. Thanks.

UICatalog/Scenarios/SingleBackgroundWorker.cs
BDisp: I submitted the PR https://github.com/tig/Terminal.Gui/pull/42 to fix this.

UICatalog/Scenarios/SyntaxHighlighting.cs
tig: Fixed.

UICatalog/Scenarios/TableEditor.cs
tig: Can't repro.

tig:
@tznind, @BDisp, @dodexahedron - This is as ready as it's gonna get. Please read the first post to see all that I've done.

I'd like to get this merged tomorrow.

BDisp (approved):
Only remining the `TabView` but I think it will be fixed with the #3808.

Please can you change the file TabMouseEventArgs.cs as below? Thanks!

```cs
using System.ComponentModel;

namespace Terminal.Gui;

/// <summary>Describes a mouse event over a specific <see cref="Tab"/> in a <see cref="TabView"/>.</summary>
public class TabMouseEventArgs : HandledEventArgs
{
    /// <summary>Creates a new instance of the <see cref="TabMouseEventArgs"/> class.</summary>
    /// <param name="tab"><see cref="Tab"/> that the mouse was over when the event occurred.</param>
    /// <param name="mouseEvent">The mouse activity being reported</param>
    public TabMouseEventArgs (Tab tab, MouseEventArgs mouseEvent)
    {
        Tab = tab;
        MouseEvent = mouseEvent;
    }

    /// <summary>
    ///     Gets the actual mouse event.  Use <see cref="HandledEventArgs.Handled"/> to cancel this event and perform custom
    ///     behavior (e.g. show a context menu).
    /// </summary>
    public MouseEventArgs MouseEvent { get; }

    /// <summary>Gets the <see cref="Tab"/> (if any) that the mouse was over when the <see cref="MouseEvent"/> occurred.</summary>
    /// <remarks>This will be null if the click is after last tab or before first.</remarks>
    public Tab Tab { get; }
}

```

PR #3811, opened 2024-11-10
Fixes #2489. New `ScrollSlider`, `Scroll`, and `ScrollBar` - Adds `IScrolLBar` so any View can have scrollbars

### LINKED ISSUE #2489: Create new `ScrollBar` based on a new `Scroll` and remove `ScrollBarView/ScrollView`

tig:
This is based on @BDisp's awesome work in #3498 and #3798.

## Fixes

- Fixes #2489

## Proposed Changes/Todos

- [x] Port @bdisp's work to #3798.
- [x] Polish `ScrollSlider`
- [x] Update `Scroll` to have `ContentPosition` and `SliderPosition` ++
- [x] Update `ScrollBar` to be based on `Scroll` and just use`Button`
- [x] Update unit tests  
- [x] Rebuild `CharMap` to use the new `Scrollbar` and demonstrate: Content area with headers, full keyboard integration, scrolling with > 1 increments (horizontal - 3 columns per).
- [x] Update `Scroll` and `ScrollBar` scenarios
- [x] Fix `AllViewsTester` so it can test scrollbars 
- [x] Update `Scrolling` Scearnio
- [x] Update `ContentScrolling` scenario 
- [ ] Move `HorizontalScrollbar`/`VerticalScrollbar` out of `View` and into `IScrollBar`/`IScrollbarIHelper` classes
- [ ] Re-implement the equivalent of "keep content visible"

## Pull Request checklist:


tig:
Ax4k6Nw 1

PR #3837, opened 2024-11-19
Fixes #3692++ - Rearchitects drivers

### LINKED ISSUE #2319: Refactor `ConsoleDriver` to minimize duplicated code in all drivers

### LINKED ISSUE #2610: Refactor `ConsoleDriver`s to support ANSI sequences natively (e.g. Virtual Terminal Sequences on Windows)

### LINKED ISSUE #2803: Harden `EscSeqUtils`

### LINKED ISSUE #3312: Mouse API makes it way too hard to track button pressed

### LINKED ISSUE #3374: Not always getting `Button1Clicked` events on `Netdriver` or `CursesDriver`

### LINKED ISSUE #3404: Mouse doubleclick stops continous pressed events.
I can reproduce in the `Buttons` scenario:

- Hold the mouse perfectly still and press-release then press-hold again. Increments once, but then does  not repeat:

https://i.imgur.com/eE3BSgr.gif

Using the `Mouse` scenario the issue has to do with the doubleclick logic.

Here, I just pressed-hold-released:

image

Here, I pressed-released-pressed-hold:

image

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/3323#issuecomment-2053691080_

### LINKED ISSUE #3692: Refactoring, cleanup, and abstraction for ConsoleDrivers

### LINKED ISSUE #3892: Comprehensive Logging and Metrics (Opt in)
I have taken this on as part of #3837 

'Opt in' means you have to specifically create a logger in your host program before logging output gets created (i.e. we default to null logger).

- Log when key events happen e.g. Init/Shutdown
- Log fatal events in background threads e.g. input loop thread
- Knowing how long things like the iteration loop is taking

Add logging abstraction dependency (i.e. `ILogger`).
Add standard metrics dependency e.g. [System.Diagnostics.Metrics](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics?view=net-9.0)

We could roll our own or write a scenario in UICatalog but its better to just go with the standard.  It also makes things work for downstream customers.

### LINKED ISSUE #3916: Pressing Alt+any key closes app when using NetDriver
**Describe the bug**
Pressing Alt+any key closes app when using NetDriver  in v2_release

**To Reproduce**
1. Go to UICatalog directory
2. dotnet run --driver NetDriver
3. Press Alt+F
4. UICatalog immediately exits

**Expected behavior**
The 'File' menu should open

 - OS: Windows 11
 - Windows Terminal/Powershell

tznind:
Creates new drivers that are simpler versions of existing WindowsDriver and NetDriver.  Also introduces a shared architecture such that only subcomponents vary between driver implementations, as discussed in #3821 .

After a period of shared execution with both sets of drivers available we will fix any bugs and then retire the old drivers.

Curses driver is not tackled, I am hopeful it can be retired completely now that net input is so strong and well implemented in core dotnet framework.

## Fixes

- Fixes #3920
- Fixes #3916
- Fixes #3904
- Fixes #3892
- Fixes #3692 
- Fixes #2319
- Fixes #2803
- Fixes #2610
- Might fix #3312 
- Might fix #3404 
- Might fix #3374 

## Metrics

You can now see metrics on ant Terminal.Gui program by using its exe name and a metrics tool e.g.

```
 dotnet-counters monitor -n UICatalog --counters Terminal.Gui
```

Implementation uses [System.Diagnostics.Metrics](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics?view=net-8.0) making it integrate easily with tools such as OpenTelemetry

## Logging
You can enable logging of internals of Terminal.Gui by setting `Logging.Logger` to any `Microsoft.Extensions.Logging` compatible logging framework (nlog, serilog etc).

Naturally you must not use a console logger

For example with Serilog to a file:

```csharp
Logging.Logger = CreateLogger ();

...

private static ILogger CreateLogger ()
{
	// Configure Serilog to write logs to a file
	Log.Logger = new LoggerConfiguration ()
				 .MinimumLevel.Verbose () // Verbose includes Trace and Debug
				 .WriteTo.File ("logs/logfile.txt", rollingInterval: RollingInterval.Day)
				 .CreateLogger ();

	// Create a logger factory compatible with Microsoft.Extensions.Logging
	using var loggerFactory = LoggerFactory.Create (builder =>
						{
							builder
								.AddSerilog (dispose: true) // Integrate Serilog with ILogger
								.SetMinimumLevel (LogLevel.Trace); // Set minimum log level
						});
	// Get an ILogger instance
	return loggerFactory.CreateLogger ("Global Logger");
}
```

## Class Diagram

image

## Progress

- Keyboard
  - [x] Exit
  - [x] Typing
  - [x] Backspace
- Views
  - [x] Run/RequestStop
  - [x] MenuBar
- Mouse (see `MouseInterpreter`)
  - [x] Button 1 click
  - [x] Multi clicks
  - [x] All buttons
  - [x] Scroll wheel
- Other
  - [x] Clipboard

## Proposed Changes/Todos

Splits drivers up into logical components, simplifies and centralizes shared code patterns.

## Pull Request checklist:


tig:
Hope you're ok with me updating the title & first post...

tznind:
> Hope you're ok with me updating the title & first post...

No problem! Always appreciate a clear naming

Now we have fledgling support for
- Keyboard
- Mouse
- Screen size

Net (i.e. native dotnet) and Windows (i.e win 32 api) are the two implementations - I assume we can drop curses?

resize-etc

tznind:
Working on it but I think this is how I want to handle mouse.  It needs work but goal is to capture all the relevant info as you would on desktop and only release events when sure of end users behaviour:

This would all be internal to the state machines initially and may just surface with the existing API until our events system aligns.

```csharp
#nullable enable

namespace Terminal.Gui;

public class MouseStateManager
{
    /// <summary>
    /// Function for returning the current time. Use in unit tests to
    /// ensure repeatable tests.
    /// </summary>
    private Func<DateTime> Now { get; set; }

    /// <summary>
    /// How long to wait for a second click after the first before giving up and
    /// releasing event as a 'click'
    /// </summary>
    public TimeSpan DoubleClickThreshold { get; set; }

    /// <summary>
    /// How long to wait for a third click after the second before giving up and
    /// releasing event as a 'double click'
    /// </summary>
    public TimeSpan TripleClickThreshold { get; set; }

    /// <summary>
    /// How far between a mouse down and mouse up before it is considered a 'drag' rather
    /// than a 'click'. Console row counts for 2 units while column counts for only 1. Distance is
    /// measured in Euclidean distance.
    /// </summary>
    public double DragThreshold { get; set; }

    public MouseState CurrentState { get; private set; }

    private ButtonNarrative? [] _ongoingNarratives = new ButtonNarrative? [4];

    public MouseStateManager (
        Func<DateTime>? now = null,
        TimeSpan? doubleClickThreshold = null,
        TimeSpan? tripleClickThreshold = null,
        int dragThreshold = 5
    )
    {
        Now = now ?? (() => DateTime.Now);
        DoubleClickThreshold = doubleClickThreshold ?? TimeSpan.FromMilliseconds (500);
        TripleClickThreshold = tripleClickThreshold ?? TimeSpan.FromMilliseconds (1000);
        DragThreshold = dragThreshold;
    }


    /// <summary>
    /// 
    /// </summary>
    /// <param name="e"></param>
    /// <returns></returns>
    public IEnumerable<ButtonNarrative> UpdateState (MouseEventArgs e)
    {
        // TODO: manage transitions

        for (int i = 0; i < 4; i++)
        {
           // Update narratives

           // Release stale or naturally complete ones based on thresholds
        }
    }

    /// <summary>
    /// If user double clicks and we are waiting for a triple click
    /// we should give up after a short time and just assume no more
    /// clicks are coming. Call this method if the state for a given button
    /// has not changed in a while.
    /// </summary>
    /// <returns></returns>
    public ButtonNarrative? ReleaseState (int button)
    {
        
    }

    private void PromoteSingleToDoubleClick ()
    {

    }
    private void PromoteDoubleToTripleClick ()
    {

    }

    public static double DistanceTo (Point p1, Point p2)
    {
        int deltaX = p2.X - p1.X;
        int deltaY = p2.Y - p1.Y;
        return Math.Sqrt (deltaX * deltaX + deltaY * deltaY);
    }
}

/// <summary>
/// Describes a completed narrative e.g. 'user triple clicked'.
/// </summary>
/// <remarks>Internally we can have a double click narrative that becomes
/// a triple click narrative. But we will not release both i.e. we don't say
/// user clicked then user double-clicked then user triple clicked</remarks>
public class ButtonNarrative
{
    public int Button { get; set; }
    public int NumberOfClicks { get; set; }

    /// <summary>
    /// Mouse states during which click was generated.
    /// N = 2x<see cref="NumberOfClicks"/>
    /// </summary>
    public List<ButtonState> MouseStates { get; set; }

    /// <summary>
    /// <see langword="true"/> if distance between first mouse down and all
    /// subsequent events is greater than a given threshold.
    /// </summary>
    public bool IsDrag { get; set; }
}

public class MouseState
{
    public ButtonState[] ButtonStates = new ButtonState? [4];

    public Point Position;
}

public class ButtonState
{
    /// <summary>
    ///     When the button entered its current state.
    /// </summary>
    public DateTime At { get; set; }

    /// <summary>
    /// <see langword="true"/> if the button is currently down
    /// </summary>
    private bool Depressed { get; set; }

    /// <summary>
    /// The screen location when the mouse button entered its current state
    /// (became depressed or was released)
    /// </summary>
    public Point Position { get; set; }

    /// <summary>
    /// The <see cref="View"/> (if any) that was at the <see cref="Position"/>
    /// when the button entered its current state.
    /// </summary>
    public View? View { get; set; }

    /// <summary>
    /// True if shift was provided by the console at the time the mouse
    ///  button entered its current state.
    /// </summary>
    public bool Shift { get; set; }

    /// <summary>
    /// True if control was provided by the console at the time the mouse
    /// button entered its current state.
    /// </summary>
    public bool Ctrl { get; set; }

    /// <summary>
    /// True if alt was held down at the time the mouse
    /// button entered its current state.
    /// </summary>
    public bool Alt { get; set; }
}
```

tig:
> Working on it but I think this is how I want to handle mouse.

Wow. I think I really like this!!

Can you write a paragraph that would go in the intro of `mouse.md` that describes the overall model in plain English, using the word "narrative"? This will go a long way to helping me understand your intent, and helping ensure your thinking is clear.

tznind:
I've updated the initial post with the new diagram of how `IApplication` fits in.  Theres a lot of work to knit things together to work robustly.  Theres just so much complexity in the existing code!

tznind:
Invoke, Idle and Timeout support in v2.

I refactored into a new class with interface `ITimedEvents` that is decoupled from main loop.

Currently still same code implementation but in future we can work on it seperately from main loop.

v2-animation
_Animation running in v2 with timed events, note the exit code 0, indicating we can also exit the app!_

tznind:
@BDisp I am struggling to figure out this bug I have with the new `WindowsOutput`

It seems to render correctly to the screen but then clear again immediately or something.

You can see by uncommenting the line:

```
                // Uncomment this to see that it does render properly :/
                // Process.GetCurrentProcess ().Kill();
```

I can't figure out what is clearing it or making the terminal not show - do you have any ideas?

console-no-show

It was working correctly in a small test project, and the NetDriver style NetOutput works correctly (you can change to net driver by doing following):

```diff
-        if ( p == PlatformID.Win32NT || p == PlatformID.Win32S || p == PlatformID.Win32Windows)
-        {
-            var inputBuffer = new ConcurrentQueue<WindowsConsole.InputRecord> ();
-            var loop = new MainLoop<WindowsConsole.InputRecord> ();
-            _coordinator = new MainLoopCoordinator<WindowsConsole.InputRecord> (TimedEvents,
-                                                                                () => new WindowsInput (),
-                                                                                inputBuffer,
-                                                                                new WindowsInputProcessor (inputBuffer),
-                                                                                () => new WindowsOutput (),
-                                                                                loop);
-        }
-        else
-        {
            var inputBuffer = new ConcurrentQueue<ConsoleKeyInfo> ();
            var loop = new MainLoop<ConsoleKeyInfo> ();
            _coordinator = new MainLoopCoordinator<ConsoleKeyInfo> (TimedEvents,
                                                                    () => new NetInput (),
                                                                   inputBuffer,
                                                                   new NetInputProcessor (inputBuffer),
                                                                   () => new NetOutput (),
                                                                   loop);
-        }
```

Any help would be greatly appreciated

BDisp:
Sorry I couldn't found the cause of this issue. My laptop overheat running it. I thought the cause was due to the recursive write which doesn't allow the driver show the output, but enabling the conditional `if (s != _lastWrite)` it still doesn't output anything. I also add this conditional to avoid always clear the contents, but still remain the issue.

```cs
        if (OutputBuffer.Cols != size.Width || OutputBuffer.Rows != size.Height)
        {
            OutputBuffer.SetWindowSize (size.Width, size.Height);
        }
```

You have to see when this issue started happening to check the commit that cause it.

tznind:
Found one issue,  `WindowsConsole` which I thought was just helper functions - it actually spins up an infinite loop of its own to read console input with no delay.

I have turned it off based on bool to constructor and now you can at least boot and Esc to exit WindowsDriver and see the output! But still there is the issue of clearing even if you change the wait period to 500ms etc. 

One of the tenants of v2 is to prevent any view refreshing, pushing output, clearing screen etc ever.

Instead the view has to notify its intent e.g. via SetNeedsDisplay and then MainLoop<T> (the new loop) does it once at the end of the iteration.

9a72628d354fbc419086ba8232e987b648f60479

tznind:
Maybe the output is being written to one screen buffer but app has switched to another and it only comes back on exit the app?

BDisp:
Maybe it's using a different output handle than the original.

BDisp:
> I have WindowsOutput on one thread and WindowsInput on a different one, perhaps the output is writing to the buffer that the input one is using while also creating its own or something.

Output and input handles are separated from each other. You make ensures recreating the handles after shutdown and not during. Make sure the handles isn't changed during a normal iteration.

tznind:
Windows Output driver working :) I stopped using WindowsConsole and just copied relevant methods across to WindowsOutput instead.  I think issue was wrong handle.

tznind:
@tig how far are we along the route of making it so MenuBar and StatusBar are not 'special views' that sit outside of the hierarchy (i.e. outside of Application.Top).

I am trying to get the new main loop to work as follows:

- Nobody is allowed to 'ad hoc draw' i.e. Driver.Refresh does nothing
- Exclusively operate on NeedsDisplay / NeedsLayout
- Draw only once at the end of the main loop and only if views need it

Currently this results in menu/status bar disapearing.  This is the current 'brute force' draw approach:

```csharp
 public void Iteration ()
 {
     InputProcessor.ProcessQueue ();

     if (Application.Top != null)
     {
         bool needsDrawOrLayout = AnySubviewsNeedDrawn(Application.Top);

         if (needsDrawOrLayout)
         {
             // TODO: throttle this
             var size = Out.GetWindowSize ();

             OutputBuffer.SetWindowSize (size.Width, size.Height);

             // TODO: Test only

             Application.Top.Layout ();
             Application.Top.Draw ();

             Out.Write (OutputBuffer);
         }
     }

     TimedEvents.LockAndRunTimers ();

     TimedEvents.LockAndRunIdles ();
 }

 private bool AnySubviewsNeedDrawn (View v)
 {
     if (v.NeedsDraw || v.NeedsLayout)
     {
         return true;
     }

     foreach(var subview in v.Subviews )
     {
         if (AnySubviewsNeedDrawn (subview))
         {
             return true;
         }
     }

     return false;
 }
```

tznind:
Ok most stuff works and cpu sits at about 0% 

Now needs loads of tests and to round out a few implementations like all the various key translations, double click etc

I think I will add ILogger support so we can see things like ignored ansi codes, record where/if there is slowdown etc

Terminal.Gui/ConsoleDrivers/V2/NetInputProcessor.cs
```diff
@@ -23,7 +23,7 @@ protected override void Process (ConsoleKeyInfo consoleKeyInfo)
     /// <inheritdoc />
     protected override void ProcessAfterParsing (ConsoleKeyInfo input)
     {
-        var key = ConsoleKeyMapping.MapKey (input);
+        var key = EscSeqUtils.MapKey (input);
```
BDisp, line 26: Good choice.

tznind:
Ok I need to work a bit on my understanding of how to integrate `Application.Run`.

The users main entry thread for `Program.cs` is going to call `Application.Init` and then `Run`.

That method will 'block' but it also is the thread that processes Click events e.g. if the user wants to `Run` another `View`.

Currently I have 2 main threads
- Input Loop
- Everything else loop (output, timers, layout, draw etc)

But I'm realizing now that that is an issue for integrating with `Run`.  Because `Run` has to appear to block but still process user event code (e.g. click) and be able to put another `Run` call onto the csharp call stack.

Any insight into my understanding of things or suggestions would be welcome - especially clarifications on how it is currently working as its a bit opaque.

This is my vision for how v2 will work.  At the moment Run just blocks and loops, looking at the state of the coordinator.  But that does not work well for starting a nested Run or exiting a nested Run back into the parent Run.

image

tznind:
Ok I have got it! Looking at the existing codebase the 'Iterate/Output' loop that I have on its own thread is actually called repeatedly within `Run`.  This means that the entry thread the application has in their program is the one that the UI code is all running on.

That makes sense and is 1 less thread so I'm up for it.  Now you can open and close as many scenarios as you want.

tznind:
Have started looking at logging (using defacto standard `System.Extensions.Logging`).  I've used Serilog for UICatalog but any logging framework will be compatible in end users applications. It would be nice to throw if user tries to set a console logger but probably impractical to do reliably.

For metrics I'm less certain on the 'standard' approach.  After a couple of false starts I found the [Meter class in System.Diagnostics.Metrics](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation) which seems to be a good fit for performance logging:

To see metrics realtime for running process you can run:
```
dotnet tool update -g dotnet-counters
dotnet-counters monitor -n UICatalog --counters Terminal.Gui
```

image
_File logging in client program UICatalog using Serilog dependency + metrics monitoring via `dotnet-counters`_

tznind:
I have dialed MouseInterpreter way back to just do the existing click behaviours including repeat click.  It was getting too complicated and it doesnt actually make sense to delay reporting click to see if there is double click - so it is back to reporting click,double then triple.

tznind:
Ok this is notionally ready for first pass review

I'm onto the final phase now

- [ ] Warnings (NRT, xmldoc etc)
- [ ] Unit tests
- [ ] All the corner cases I'm forgetting about

You can run with:
image

tznind:
Once this is merged and well tested we can have an epic deleting session to remove the old drivers - if we want to that is.

tznind:
@BDisp can you see if you can see why right click context menus are not popping? I guess something about the mouse event(s) in v1 vs v2?

BDisp:
> ok looks like Released+Clicked flags together is bad. TG code wants 2 seperate events I think.
> 
> Got it, is just about order of events

Glad you found the solution but that's really the cause of the bad behavior that was causing the right click wasn't working. Of course that each escape sequence must raise separate events, for them to work as expected. Good work and Happy New Year.

tznind:
>  Of course that each escape sequence must raise separate events, for them to work as expected. Good work and Happy New Year.

Happy new year! 

There is actually no escape sequence for click so you must in fact create one manually in interpreter:

Input:
```
[<0;3;3m
[<35;3;3m
[<0;4;4M
```
We see mouse is moved with no button
Mouse is down (button 35)
Mouse is released (button is back to 0 - also we have 'M' instead of 'm;)

This must currently be translated into 4 events for Terminal.Gui right click menu to work properly:
```
ReportMousePosition
Button1Pressed
Button1Released
Button1Clicked
```

So it is not a 1 to 1 mapping between escape codes and events, click is a 'supplemental' event that we inject into the event stream.

Combining flags e.g. Released+Click as one event does not work with current codebase (which is the original thing I was doing).

tznind:
@BDisp `EscSeqUtils.MapKey` does not handle backspace.  Here is the ConsoleKeyInfo that comes in direct from console when reading with console read:

image

Apparently:
> When you press the Backspace key, the console might interpret it as sending the 127 (0x7F) code instead of 0x08 (Backspace).
> This happens because of differences in how the terminal emulator or console environment maps keyboard keys to control codes.

How is this handled in NetDriver? is MapKey the best place to put this new logic?

BDisp:
The `DEL` (127) is handled on the `EscSeqUtils.MapConsoleKeyInfo` method.

image

BDisp:
This code fixes the right button click to popup the context menu.

image

tznind:
> The DEL (127) is handled on the EscSeqUtils.MapConsoleKeyInfo method.

Great, I have added in call to this function in the new architecture, see [aa8bdd7](https://github.com/gui-cs/Terminal.Gui/pull/3837/commits/aa8bdd72499f7a95e75485acefe07eef0a3c6642)

> This code fixes the right button click to popup the context menu.

Right click works for v2 net but not v2 win, I will have to explore your suggestion further when I have some time.

tznind:
Windows right click working thanks! [6c55e11](https://github.com/gui-cs/Terminal.Gui/pull/3837/commits/6c55e117e015c15614b5fa1e902968a7182b3800)

BDisp:
> Great, I have added in call to this function in the new architecture, see [aa8bdd7](https://github.com/gui-cs/Terminal.Gui/pull/3837/commits/aa8bdd72499f7a95e75485acefe07eef0a3c6642)

Nice.

> Right click works for v2 net but not v2 win, I will have to explore your suggestion further when I have some time.

Works with `NetDriver` because it rely in the `EscSeqUtils` API and `WindowsDriver` don't, and it's needed to point the right way how to handle it. Don't forget to deal with the warning.

tznind:
I have added metrics for Redraw in v2 drivers and Trace logging for the view that is responsible for wanting itself redrawn.  This will let us fix issues e.g. this one where HexView is constantly saying it wants layout and redraw when nothing is going on in app:

image

tznind:
@BDisp @tig any feedback on this PR would be appreciated.

BDisp:
> @BDisp @tig any feedback on this PR would be appreciated.

I'm in vacancy mode sorry.

tig:
> @BDisp @tig any feedback on this PR would be appreciated.

Yes. 

My lighthouse beam is focused on other things right now. 

But it's swinging around and will be back on TG real soon now. 

I'm really excited about this PR!

tznind:
@BDisp the ` EscSeqUtils.MapConsoleKeyInfo` is broken for some keys e.g. 

```
new ConsoleKeyInfo ('}', ConsoleKey.Oem6, true, false, false), new Rune ('}') 
```
Gets returned as 
```
new ConsoleKeyInfo('}', ConsoleKey.None, true, false, false);
```

This then breaks `EscSeqUtils.MapKey` 

`MapConsoleKeyInfo` and `MapKey` are the two methods I am chaining to turn `ConsoleKeyInfo` into `Key`

For now I am using workaround:

```csharp
 /// <summary>
 /// Converts terminal raw input class <see cref="ConsoleKeyInfo"/> into
 /// common Terminal.Gui event model for keypresses (<see cref="Key"/>)
 /// </summary>
 /// <param name="input"></param>
 /// <returns></returns>
 public static Key ConsoleKeyInfoToKey (ConsoleKeyInfo input)
 {
     ConsoleKeyInfo adjustedInput = EscSeqUtils.MapConsoleKeyInfo (input);
     
     // TODO : EscSeqUtils.MapConsoleKeyInfo is wrong for e.g. '{' - it winds up clearing the Key
     //        So if the method nuked it then we should just work with the original.
     if (adjustedInput.Key == ConsoleKey.None && input.Key != ConsoleKey.None)
     {
         return EscSeqUtils.MapKey (input);
     }

     return EscSeqUtils.MapKey (adjustedInput);
 }
```

tig:
Starting my review...

Some things I noticed in pulling down and running locally. This is with "ConsoleDriverFacade(win)"

* Incorrect drawing:
image

* Window drag doesn't work
* App hotkeys don't work (Ctrl-A) for UI Catalog about box.
* Exiting UI Catalog about box causes an exception

This stuff leads me to believe this isn't really ready for merge, right? 

But you want me to review the code?

tig:
Running UICatalog "All Views Tester" -b -t 5000 --driver v2net

Shows lots of issues.

Terminal.Gui/ConsoleDrivers/V2/ApplicationV2.cs
```diff
@@ -0,0 +1,218 @@
...
+
+namespace Terminal.Gui;
+
+/// <summary>
+///     Implementation of <see cref="IApplication"/> that boots the new 'v2'
+///     main loop architecture.
+/// </summary>
+public class ApplicationV2 : ApplicationImpl
+{
+    private readonly Func<INetInput> _netInputFactory;
```
tig, line 15: Not sure I like having Application be so tightly bound to "net" and "win". What about curses/linux? Or is the idea "net" works great on linux?

Terminal.Gui/ConsoleDrivers/V2/Logging.cs
```diff
@@ -0,0 +1,52 @@
...
+///     Singleton logging instance class. Do not use console loggers
+///     with this class as it will interfere with Terminal.Gui
+///     screen output (i.e. use a file logger).
+/// </summary>
+/// <remarks>
+///     Also contains the
+///     <see cref="Meter"/> instance that should be used for internal metrics
+///     (iteration timing etc).
+/// </remarks>
+public static class Logging
```
tig, line 17: Love this concept. 

Does it have to be restricted to drivers though?

tig (changes requested):
Some comments here. 

My review was pretty superficial. I didn't actually dive deep into the tests.

What is your thinking on how we move forward with this?

It appears that unless you explicitly enable the new drivers, everything works swell.

Is the intent to merge with the new drivers "broken" and then fix them off of v2_develop?

Terminal.Gui/ConsoleDrivers/V2/ApplicationV2.cs
tznind: Yes my current thinking is that if we can get away with only the two drivers then it will make things simpler.

I think core dotnet console apis will work fast and well enough now on Linux for prime time - but we can test and see if there are distros that have performance issues that would need curses.

Application is the ugliest bit of this PR as it is where the old architectural approach (static gateway + driver monolith) meets the new (orchestrator + generics subcomponents).

I think these factories and how we boot strap new architecture will evolve here.

This needs to be good enough not perfect (my hope anyway).

Terminal.Gui/ConsoleDrivers/V2/Logging.cs
tznind: Certainly this can be used later on for other bits of the core code, it is not tied to drivers specifically.

tznind:
I would like to fix bugs.  Especially any crash bugs.

Biggest issue I noticed was with menus.  They seem to be 'special case' for Top and have explicit code in Application that interacts with them.  

I was hoping that if we could get it so any view can have menus and they work just like any other subview - then the issues might become less opaque and/or disappear. 

Let me try the all views tester.  I have run many scenarios in isolation but it's unclear what some of them are meant to do upon cursory examinations.

It would be great to have help identifying the main outstanding 'show stopper bugs' in terms of behaviour.

tznind:
I want to solve all the 'driver level' bugs but then address 'view specific' issues later e.g. HexView always reporting that it NeedsLayout/NeedsRedraw constantly true. 

The new 'main loop' is more trusting of the views reporting their state. 

For example there are no longer any 'out of band' screen Refresh allowed. Console is only flushed/updated after loop iteration and only once (and only if a View was needs draw true)

Terminal.Gui/ConsoleDrivers/V2/Logging.cs
tznind: I only put it in this Drivers folder because its where I was creating everything else, can move to root or another folder if you like.

tznind:
> Running UICatalog "All Views Tester" -b -t 5000 --driver v2net

I've not used this before,  I will have a look into it thanks.

I also see that the behaviour of app is different when launch  from UICatalog vs visual studio e.g. black colors.  Maybe something to do with configuration file/paths

image

tznind:
Looks like AllView tester is double booting.  And the LayoutEditor is constantly reporting true for redraw.

```
2025-01-30 18:40:02.825 +00:00 [INF] Main Loop Coordinator booting...
2025-01-30 18:40:02.842 +00:00 [INF] Creating NetOutput
2025-01-30 18:40:02.844 +00:00 [INF] Creating NetInput
2025-01-30 18:40:02.846 +00:00 [INF] Main Loop Coordinator booting complete
2025-01-30 18:40:02.849 +00:00 [INF] Main Loop Coordinator booting...
2025-01-30 18:40:02.849 +00:00 [INF] Creating NetOutput
2025-01-30 18:40:02.849 +00:00 [INF] Creating NetInput
2025-01-30 18:40:02.849 +00:00 [INF] Main Loop Coordinator booting complete
2025-01-30 18:40:02.942 +00:00 [INF] Run 'Window(Esc to Quit - Scenario: All Views Tester){X=0,Y=0,Width=0,Height=0}'
2025-01-30 18:40:03.002 +00:00 [VRB] Window triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.003 +00:00 [INF] Console size changes from '{Width=0, Height=0}' to {Width=120, Height=30}
2025-01-30 18:40:03.128 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.184 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.247 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.309 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.372 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.434 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.498 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.561 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.625 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.688 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.751 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.813 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.877 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:03.940 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.003 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.067 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.129 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.191 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.255 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.318 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.379 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.442 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.505 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.568 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.631 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.691 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.753 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.814 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.875 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.938 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:04.999 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.061 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.124 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.186 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.248 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.311 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.373 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.436 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.498 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.552 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.607 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.670 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.732 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.794 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.857 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.919 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:05.983 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.045 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.108 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.171 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.233 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.296 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.358 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.420 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.483 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.545 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.608 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.669 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.732 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.793 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.855 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.918 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:06.981 +00:00 [VRB] AnsiResponseParser releasing ''
2025-01-30 18:40:06.984 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.043 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.106 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.170 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.233 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.295 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.358 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.420 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.483 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.544 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.607 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.669 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.732 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.795 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.857 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.919 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:07.981 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:08.044 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:08.106 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:08.168 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:08.231 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:08.294 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:08.356 +00:00 [VRB] LayoutEditor triggered redraw (NeedsDraw=True NeedsLayout=True) 
2025-01-30 18:40:08.420 +00:00 [INF] RequestStop ''
```

tig:
> I would like to fix bugs.  Especially any crash bugs.
> 
> Biggest issue I noticed was with menus.  They seem to be 'special case' for Top and have explicit code in Application that interacts with them.  

Hence the Issues and PRs around Top and rewriting Menu. See those. 

> 
> I was hoping that if we could get it so any view can have menus and they work just like any other subview - then the issues might become less opaque and/or disappear. 

Yes. See related menu PRs.

tznind:
All views tester now working, was just missing raising init event.  Looks like configuration manager doesnt work same either as its monochrome - I will look into this
allviews

```
dotnet run -- "All Views Tester" -b -t 5000 --driver v2nwin
```

allviews

tznind:
Colors fixed, I just wasn't calling the init on config manager from the new bootstrapping Application2 class init

Terminal.Gui/ConsoleDrivers/V2/MainLoop.cs
```diff
@@ -142,6 +144,21 @@ internal void IterationImpl ()
         Logging.IterationInvokesAndTimeouts.Record (swCallbacks.Elapsed.Milliseconds);
     }
 
+    private View? _lastTop;
+    private void HackLayoutDrawIfTopChanged ()
+    {
+        // TODO: This fixes closing a modal not making its host (below) refresh
+        //       until you click.  This should not be the job of the main loop!
```
BDisp, line 151: You're right. That must be an `Application` job.

tznind: Somehow, it seems to work with the original drivers but not the v2 ones.

V2 drivers only listen to these properties (needs layout/draw).  There are no manual or triggered draws, so possibly that is why this hack is needed for now.

BDisp: I understand, but I think that if you do the same thing in the `Application.LayoutAndDraw` method you'll will obtain the same result, but I could be wrong.

tznind: In my code, layout and draw is only called if a subview of Top wants redrawn or layed out... I will look into it some more later

tznind:
I think the remaining issues are to do with layout and/or the new iteration.

I am trying to structure the new MainLoop<T> to only redraw and/or layout when a view says it needs to.  I think implementation is quite naive but I also think the existing main loop/application is quite brittle and exposed e.g. anyone can and does call Refresh, Layout etc at will.

I could use some help, for example I see lots of LayoutException that get caught later on.  Why does that happen with new loop and not old? It's hard to understand why.

tznind:
@tig my proposal for this branch would be to confirm

- Regular drivers are not broken (confirmed)
- Issues with the new drivers are primarily small and/or layout/redraw related

Then we can merge it and I can do a followup work where we log the hell out of everything in the library and I can start working on layout/draw issues.

tig:
Travelling and had a few minutes to play.... This is looking great. 

I agree with your plan for mergeing...

Some new issues I noticed:

Line Drawing pallette is corrupt.



Drag about box window broken (and crashes) on .win driver.

W0LJM3E 1

Drag scroll bar sometimes doesn't restore non-drag color:

8anNNx5 1

Numeric up/down does not auto-increment/acellerate when held.

Charmap has a bunch of issues. Getting CP info never completes. When loaded, the right hand pane doesn't show until a mouse move. 

image

tznind:
Good finds! seems I was not raising released flags on win driver.  [cc96920](https://github.com/gui-cs/Terminal.Gui/pull/3837/commits/cc96920cc308df12101015d7bce716e986a5dbdf) shoulf fix the 'mouse grab' related bugs (drag window etc)

tznind:
> Line Drawing pallette is corrupt.

Initial size is corrupt, if you drag out the bottom and right it appears as it does in normal drivers.  The checker pattern for background color is normal.  Though I note that background colour doesnt work in any of the drivers - seems like a regression, I remember it used to work.

> When loaded, the right hand pane doesn't show until a mouse move.

Hmn probably it doesn't properly implement NeedsDisplay or part of its layout doesn't trigger it.  I think there will quite a bit of this, we need to have things like view becoming visible for first time to set it as NeedsDisplay.  The old architecture had a lot more `Refresh` and so on (`Refresh` is now completely ignored - all views must play nice with the loop and the NeedsDisplay / NeedsLayout flags).

tznind:
Have fixed the Ready event not being raised in [6b9367f](https://github.com/gui-cs/Terminal.Gui/pull/3837/commits/6b9367f0a4e6626726a6af1c89f63bca541acdca).

Terminal.Gui/ConsoleDrivers/AnsiResponseParser/AnsiResponseParser.cs
```diff
@@ -1,13 +1,41 @@
...
 
+    /// <summary>
+    ///     Event raised when mouse events are detected - requires setting <see cref="HandleMouse"/> to true
+    /// </summary>
+    public event EventHandler<MouseEventArgs>? Mouse;
+
+    /// <summary>
+    ///     Event raised when keyboard event is detected (e.g. cursors) - requires setting <see cref="HandleKeyboard"/>
+    /// </summary>
+    public event Action<object, Key>? Keyboard;
```
tig, line 25: Why Action? I thought we agreed on this model for events in the library?

https://gui-cs.github.io/Terminal.GuiV2Docs/docs/events.html

Specifically:

> Common Patterns
> The primary pattern for events is the event/EventHandler idiom. We use the Action<T> idiom > sparingly. We support INotifyPropertyChanged in cases where data binding is relevant.

Happy to be educated on why Action is better here.

Terminal.Gui/ConsoleDrivers/V2/InputProcessor.cs
```diff
@@ -0,0 +1,154 @@
...
+public abstract class InputProcessor<T> : IInputProcessor
+{
+    /// <summary>
+    ///     How long after Esc has been pressed before we give up on getting an Ansi escape sequence
+    /// </summary>
+    private readonly TimeSpan _escTimeout = TimeSpan.FromMilliseconds (50);
+
+    internal AnsiResponseParser<T> Parser { get; } = new ();
+
+    public IKeyConverter<T> KeyConverter { get; }
```
tig, line 21: missing API docs?

tig (changes requested):
Looking great! Just a few asks...

Terminal.Gui/ConsoleDrivers/AnsiResponseParser/AnsiResponseParser.cs
tznind: Just an oversight, have fixed 🚀

tznind:
Merge conflict fixed

tig (approved):

tig:
I want to merge this, but we keep getting unit test failures (see https://github.com/gui-cs/Terminal.Gui/issues/3930).

likely not due to this PR, but I'm not going to merge until this passes at least once. I just restarted the action....

PR #3872, opened 2024-11-30
Adds IApplication

### LINKED ISSUE #3863: Finalize `IApplication`
## From @tig, re-framing this Issue based on @tznind's original.

This is a sub-issue of #4125. 

See these Issues for more context:

- #4318 
- #4316 
- #4319 

The work for this Issue involves moving things currently in `Application` that SHOULD go in `IApplication`. Some things SHOULD NOT go directly into `IApplication`. Eg. Mouse, Keyboard, Navigation, etc... as they are better left as encapsulated sub-objects.

In #4318, I added a bunch of TODOs that indicate what still needs to be moved to `IApplication`. Note saying "moved to `IApplication`" implies:

- Adding to `IApplication`
- Moving implementation to `ApplicationImpl`
- Implementing the `static` in `Application` that forwards to `ApplicationImpl`

The PR that fixes this issue, should 

- Ensure `Application` is a "no code" facade into IApplication. The only non-passthrough members should be the [ConfigProperties].
- Strive to continue to ensure View sub-classes have no need to ever access View.Driver. ViewBase.Driver should be `internal` and ViewBase should expose everything needed.
- Update `FakeDriver` so that it doesn't use the Application.RunIteration method at all because it indirectly still uses the legacy driver stuff.
- Change terminology: We currently use "Initialization" (e.g. Application.Initialzation.cs) when all that is really about the application lifecycle from `Init` > `Shutdown`. So "Initialization"->"Lifecycle". 
- Clarify terminology: "Application.Run" is really `Begin -> Run -> Layout/Draw -> End -> Stop` and ncludes `RunState`. Is there a better name than "Run" for all this?
- Application should be organized so that the capabilities (methods, properties, events, etc...) are grouped and ordered logically. 
  1. Lifecycle
  2. Driver
  3. Keyboard
  4. Mouse
  5. Top -> Rename to Views? Or View Hierarchy? Or something else that represents "the part of the application that starts with a top-level View". Includes cursor management (which depends on both drivers and View Hierarchy). Remember we want to get rid of the current Toplevel junk (see #2491).
  6. Run (or a better name)
  7. Multitasking - Timers, Invoke, etc...
  8. LayoutAndDraw, Screen, & ToString
  9. Navigation
  10. Popover
- Folder structure for ./Terminal.Gui/App and ./Tests/ should be updated to match new structure and lexicon.
 
## Original Issue (by @tznind)

Currently `Application` is almost completely static methods and properties.  This prevents defining alternative implementations and tests cannot Moq expectations of the static methods their members invoke.

1. Change all Application methods to not be static (adding virtual where it makes sense)
2. Implement Singleton pattern i.e. everyone accesses `Application.Instance` which is an IApplication
3. Update all call sites

This will enable adding alternative implementations to IApplication e.g. #3837 

To do this I could for example subclass Application but override certain methods, or implement IApplication from scratch.  We would then change the instance in Program.cs (i.e. at start of program)

```csharp
// This would throw unless the current instance is Shutdown or not yet Init
Application.ChangeInstance(myAltApplicationClass)
```

An alternative is to keep public API how it is and add the singleton pattern behind the static methods i.e. the static methods all forward to an internal singleton.  This would be more code to write but would look prettier for the user.

**Additional context**

tznind:
Major work in progress

Architecture is:
- Keep existing public static convention e.g. Application.Init
- Add a new singleton class `ApplicationImpl`
- Static method forwards to `ApplicationImpl` as `IApplication` singleton

Public method remains the same:

```csharp
public static partial class Application
{
[...]
    public static bool Initialized
    {
        get => ApplicationImpl.Instance.Initialized;
        set => ApplicationImpl.Instance.Initialized = value;
    }
[...]
}
```

Here is the singleton

```csharp
internal class ApplicationImpl : IApplication
{
    // Private static readonly Lazy instance of Application
    private static readonly Lazy<IApplication> lazyInstance =
        new Lazy<IApplication> (() => new ApplicationImpl ());

    // Public static property to access the instance
    public static IApplication Instance => lazyInstance.Value;

    public bool Initialized { get; set; }
```

Here is the burgeoning class diagram
image

## Fixes

- Fixes #3863 

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:


tznind:
@tig please can you confirm you are happy with this approach (static accessors) - before I go too far down this rabbithole

tznind:
RunState seems to only have a single member which is Toplevel which already exists on Application - can we get rid of this class and its events?

tig:
> RunState seems to only have a single member which is Toplevel which already exists on Application - can we get rid of this class and its events?

We support multiple calls to app.Run. Each has a runstate and Toplevel. 

Application.Top points to the runstate.toplevel that is "top" (getting input). 


So no.

tig:
> @tig / @BDisp / @dodexahedron please can you confirm you are happy with this approach (static accessors) - before I go too far down this rabbithole

I am good with it!

tig:
Please read my comments on the issue about toplevel where I justify nuking it as a concept and replacing it with separate concerns for "runnable" (IRunnable) and "overlapped". 

I've completed the overlapped work. But haven't gotten to IRunnable. I have not fully thought out IRubbable and view it just as a concept at this point. If, as you add IApplication see how it could be implemented id hug you.

tznind:
Ok so I think there's two concepts going on in Application

- Global (Global keysbindings, events etc)
- State management (boot strap driver/loop, the modal View stack etc)

I think only the state stuff needs to come out. The rest can stay.

I think that any view should be runnable.  Multiple view windows running at once should just be handled with a view e.g. a DesktopView or he'll just View base class with multiple draggable window child views

I will see how I can progress down this direction

tznind:
Going to close this in favour of  #3878 

There is too much going on in Application to tackle it all at once, I will continue with the 'first principles' approach by just porting Run/Init/Shutdown to IApplication and seeing how far that goes.

Issue #2933, opened 2023-10-25
`Cell` needs much deeper thought to truly support Unicode


tig:
I hate the design I came up with for this for v2:

```cs
/// <summary>
/// Represents a single row/column in a Terminal.Gui rendering surface
/// (e.g. <see cref="LineCanvas"/> and <see cref="ConsoleDriver"/>).
/// </summary>
public class Cell {
	/// <summary>
	/// The list of Runes to draw in this cell. If the list is empty, the cell is blank. If the list contains
	/// more than one Rune, the cell is a combining sequence.
	/// (See #2616 - Support combining sequences that don't normalize)
	/// </summary>
	public List<Rune> Runes { get; set; } = new List<Rune> ();

	/// <summary>
	/// The attributes to use when drawing the Glyph.
	/// </summary>
	public Attribute? Attribute { get; set; }

	/// <summary>
	/// Gets or sets a value indicating whether this <see cref="T:Terminal.Gui.Cell"/> has
	/// been modified since the last time it was drawn.
	/// </summary>
	public bool IsDirty { get; set; }
}
```

It forces simple code to do `cell.Runes[0]` which makes code hard to read.

I propose changing this to:

```cs
/// <summary>
/// Represents a single row/column in a Terminal.Gui rendering surface
/// (e.g. <see cref="LineCanvas"/> and <see cref="ConsoleDriver"/>).
/// </summary>
public class Cell {
	/// <summary>
	/// The character to display. If <see cref="Rune"/> is <see langword="null"/>, then <see cref="Rune"/> is ignored.
	/// </summary>
	public Rune Rune { get; set; }

	/// <summary>
	/// The combining mark for <see cref="Rune"/> that when combined makes this Cell a combining sequence that could
	/// not be normalized to a single Rune.
	/// If <see cref="CombiningMark"/> is <see langword="null"/>, then <see cref="CombiningMark"/> is ignored.
	/// </summary>
	/// <remarks>
	/// Only valid in the rare case where <see cref="Rune"/> is a combining sequence that could not be normalized to a single Rune.
	/// </remarks>
	internal Rune CombiningMark { get; set; }

	/// <summary>
	/// The attributes to use when drawing the Glyph.
	/// </summary>
	public Attribute? Attribute { get; set; }

	/// <summary>
	/// Gets or sets a value indicating whether this <see cref="T:Terminal.Gui.Cell"/> has
	/// been modified since the last time it was drawn.
	/// </summary>
[...]

PR #3877, opened 2024-12-03
Fixes #2616. Support combining sequences that don't normalize.

### LINKED ISSUE #2616: Support combining sequences that don't normalize

BDisp:
## Fixes

- Fixes #2616

## Proposed Changes/Todos

- [x] Using `WindowsDriver` only works on Windows Terminal Preview

## Pull Request checklist:


BDisp:
image

Terminal.Gui/ConsoleDrivers/ConsoleDriver.cs
```diff
@@ -282,13 +282,13 @@ public void AddRune (Rune rune)
...
                             Contents [Row, Col].Rune = rune;
 
-                            if (Col < clipRect.Right - 1)
-                            {
-                                // Invalidate cell to right so that it doesn't get drawn
-                                // TODO: Figure out if it is better to show a replacement character or ' '
-                                Contents [Row, Col + 1].Rune = Rune.ReplacementChar;
-                                Contents [Row, Col + 1].IsDirty = true;
-                            }
+                            //if (Col < clipRect.Right - 1)
```
tig: This is a pretty important case. I'm not sure why it's revoved.

tig (changes requested):

tig:
This is awesome.

Can you help figure out why charmap is doing this?

Tq8hSSN 1

BDisp:
The `WindowsDriver` it's more harder to deal because it's using a buffer instead of rows and columns like the other drivers. it's needed to add space  characters to fill the gaps generated in the columns. It's giving much more work and I couldn't yet figure out the way to fix it.

BDisp:
I think the better solution for the charmap is to use views with a width of 2 and a height of 1 for all glyphs. Thus it isn't needed to manipulate the column width of the runes. Also all the views are aligned with each other. What do you think?

tig:
> I think the better solution for the charmap is to use views with a width of 2 and a height of 1 for all glyphs. Thus it isn't needed to manipulate the column width of the runes. Also all the views are aligned with each other. What do you think?

No, that would just mask the issue. 

I'd rather us figure out how to fix Windows driver.

BDisp:
This approach and the existent one have a real issue when a combining mark is set in the `View.Text` and the previous rune don't belong to this view. For e.g. if the previous rune belong to label and it's an 'a' starting inserting an '\u0301' in the `View.Text`, the driver will write 'á' instead of "a´". So, we need to think a better solution to add combining marks. I'll try find a more appropriate solution.

tig:
> This approach and the existent one have a real issue when a combining mark is set in the `View.Text` and the previous rune don't belong to this view. For e.g. if the previous rune belong to label and it's an 'a' starting inserting an '\u0301' in the `View.Text`, the driver will write 'á' instead of "a´". So, we need to think a better solution to add combining marks. I'll try find a more appropriate solution.

This is precisely why I started down the Cell route way back. 

We need to create an abstraction that lets us keep track of what the user/dev intended and then ensure the driver (where each platform behaves differently) gets the right chars in the right spot.

tig:
This is relevant: https://github.com/gui-cs/Terminal.Gui/issues/2933

Particularly the links near end of discussion.

tznind:
If possible can you please think about how we will integrate these changes into the v2 drivers refactoring branch too.  I've copied much of this code you are changing to the WindowsOut class.

BDisp:
> If possible can you please think about how we will integrate these changes into the v2 drivers refactoring branch too. I've copied much of this code you are changing to the WindowsOutput class.

That's good, this way you can check if it's working as expected in the new v2 drivers. I haven't yet resolves all the glyphs handling which I still try to found how to deal with some situations. But when I've this almost finished I'll to see how it can be integrated. `WindowsDriver` is giving some more work and I really seriously thinking using a for loop for rows and cols in the write to console instead of a foreach, unless I figure that I can deal well to find each last column.

BDisp:
I already ran the `ShowTopLine_True_TabsOnBottom_False_With_Unicode` unit test in Linux and Mac and it pass on both. So, I've no idea about it's going on here.

```
Passed Terminal.Gui.ViewsTests.TabViewTests.RemoveTab_ChangesSelection [2 ms]
[xUnit.net 00:01:34.23]     Terminal.Gui.ViewsTests.TabViewTests.ShowTopLine_True_TabsOnBottom_False_With_Unicode [FAIL]
[xUnit.net 00:01:34.23]       Assert.Equal() Failure: Strings differ
[xUnit.net 00:01:34.23]                                          ↓ (pos 29)
[xUnit.net 00:01:34.23]       Expected: ···"──────╮    \n│Les Misérables│    \n◄       "···
 │hi2               │
 └──────────────────┘
  But Was:
 ╭──────────────╮    
 │Les Miserables│    
 ◄              ╰───╮
 │hi2               │
 └──────────────────┘



Test Run Failed.
Total tests: 10312
     Passed: 10289
     Failed: 1
    Skipped: 22
 Total time: 2.0031 Minutes
       _VSTestConsole:
         MSB4181: The "VSTestTask" task returned false but did not log an error.
     8>Done Building Project "/home/runner/work/Terminal.Gui/Terminal.Gui/UnitTests/UnitTests.csproj" (VSTest target(s)) -- FAILED.
     1>Done Building Project "/home/runner/work/Terminal.Gui/Terminal.Gui/Terminal.sln" (VSTest target(s)) -- FAILED.

```

PR #3902, opened 2025-02-02
(Maybe) Add Terminal.Gui Designer

ris-work:
Terminal.Gui Designer is special in the sense it is also a good companion tool, especially for rapid prototyping and for beginners.

## Fixes

- Fixes Unspecified, improves the life for beginners and those who love RAD


tig (approved):
Thanks!

Issue #2725, opened 2023-06-23
Intermediate heap allocations


TheTonttu:
I was wondering has anyone taken a closer look at Terminal.Gui intermediate heap allocations?

OS: Windows 10 22H2 (build 19045.3086)

VS2022 performance profiler ".NET Object Allocation Tracking" shows quite a lot of intermediate System.Rune[] allocations from Terminal.Gui.TextFormatter calls on redraw. This is mostly seen when something is updated in time intervals, e.g. progress bar or a time display.

The UI Catalog Progress demo is one example of this. Although bear in mind it seems to also have lot of func and enumerator allocations which is probably just related to how the main loop invocations and progress bar are used in that demo.

On another note, I ran the same Progress demo on v2_develop branch in hopes of seeing improvement. Sadly the heap allocations have increased drastically. This time coming from Terminal.Gui.LineCanvas which might just be due to some temporary regression.

tznind:
Is this a performance bottleneck? Can you give an example of part of the code that is slow because of this and how it might be changed to be faster?

TheTonttu:
Thanks for the reply,

I haven't had a chance to take a closer look at the code yet so I cannot pinpoint anything specific. I was mainly curious if the project has had a look at the performance and allocations now that v2 is being developed and it is dropping support for older .NET versions.

It is probably not a performance bottleneck in itself but when using multiple libraries that have similar allocation characteristics then the cumulative effect can become bottleneck at the application-level, especially if doing near real-time stuff. Although you could argue maybe GC languages aren't the best choice then but that's besides the point. I would still prefer library code strive for maximum performance wherever possible if it does not make the internals or public interfaces convoluted.

.NET 5+ introduced a lot of simple APIs surrounding spans that could be used with encoding, runes and intermediate string manipulations. There are probably also some places in general that could utilize array pooling via ArrayPool and object pooling via Microsoft.Extensions.ObjectPool or with simple shared objects.

I could take a closer look in case there are any low-hanging fruit gains to be had if that's ok?
Would probably need to add a benchmark project to the solution to verify any modification does not just make the performance worse.

tznind:
> I could take a closer look in case there are any low-hanging fruit gains to be had if that's ok?

Performance is always nice to have! And it sounds like you have some expertise in this area.  Definitely all PRs are welcome.

From a purely performance perspective, the two areas I have found issues with are:

- Repeated refreshes e.g. when resizing the console screen on windows
- TextView with large files

[...]

PR #3919, opened 2025-02-22
Reduce func allocations

TheTonttu:
## Fixes

- Improves #2725
- Eliminates major sources of `Func<,>` allocations originating from LINQ lambda outer variable captures by replacing them with equivalent loops.*

\* Measured in most likely skewed scenario described in the "Data collection scenario" section.

## Proposed Changes/Todos

- [x] Refactor `LineCanvas` to reduce majority of Func allocations
    - [x] `LineCanvas.All(...)`
    - [x] `LineCanvas.GetCellMap(...)`
    - [x] `LineCanvas.Has(...)`
- [x] Refactor `Region` to reduce majority of Func allocations
    - [x] `Region.Contains(...)`
    - [x] `Region.Intersect(...)`
- [x] Refactor `PosAlign.CalculateMinDimension(...)` to reduce majority of Func allocations  
- [x] Refactor `TextFormatter.GetRuneWidth(...)` to reduce majority of Func allocations

## Pull Request checklist

## Disclaimer

I like LINQ and the convenience it offers and use it most of time over loops when appropriate. Unfortunately when used in application hot paths the heap allocations can get out of hand due to hidden allocations from captured outer variables.

## Data collection scenario

Run in Windows Terminal v1.21.10351.0 120x30 Command Prompt tab.

1. Start UICatalog.
2. Navigate down the Categories list using arrow key down.
    - Pause for a while (~1 sec) between key presses so that the Scenarios view has time to load properly.
3. Once at the bottom of the list navigate up using the arrow key up in similar manner until back at the top of the Categories list.
4. Quit application.
5. ???
6. Profit

This does not necessarily represent a typical application use and probably skews the statistics but at least reveals some allocation issues when using similar list navigation.

Ideally there would be some sort of automation that emulates more typical application.

## Object allocation statistics

Data collected in Release build using Visual Studio Performance Profiler .NET Object Allocation Tracking.

### Before
1_before_func-allocations

### After

As a bonus, also list and enumerator allocations decreased slightly.
2_after_1_top_allocations

The main focus `Func<,>` dropped way down the allocation list.

There are still some unnecessary Func<,> allocations but the PR is already getting quite bloated. Also some of them are not worth the effort unless of course profiling other parts of the UICatalog tells otherwise. I nope'd out quite quickly after seeing `DimAuto.Calculate(...)`. 😆

2_after_2_func-allocations

### LineCanvas.All

LINQ All replaced with equivalent foreach loop.

### LineCanvas.GetCellMap

Replaced array allocation inside nested loops + LINQ lambdas with List buffer outside the nested. If IntersectionDefinition would be struct then stackalloc/ArrayPool would have made more sense.

Maybe the biggest change due to ReadOnlySpan replacing arrays in the methods further down the line, which forced to replace LINQ lambdas from those methods as well. `GetRuneForIntersects(...)` being the biggest "victim" of this change. Tried to keep it readable with local methods.

### LineCanvas.Has

Similar change as in `LineCanvas.All(...)`.

### Region.Contains

Replaced LINQ Any with foreach loop.

### Region.Intersect

Replaced list allocation and LINQ lambdas with foreach loop utilizing stackalloc buffer or pooled array depending on the amount of region rectangles. The original rectangles list is cleared and reused instead of creating new list every time. The list can still of course internally create new array if capacity is too small.

### PosAlign.CalculateMinDimension

Group list and sum list allocations with multiple loops replaced with single loop combining the group check and summing.

### TextFormatter.GetRuneWidth

Removed intermediate list allocation and just looped through the StringRuneEnumerator and summing the widths.

### DIFF
--- a/Terminal.Gui/Drawing/Region.cs
+++ b/Terminal.Gui/Drawing/Region.cs
@@ -154,14 +154,34 @@ public class Region : IDisposable
     /// <param name="x">The x-coordinate of the point.</param>
     /// <param name="y">The y-coordinate of the point.</param>
     /// <returns><c>true</c> if the point is contained within the region; otherwise, <c>false</c>.</returns>
-    public bool Contains (int x, int y) { return _rectangles.Any (r => r.Contains (x, y)); }
+    public bool Contains (int x, int y)
+    {
+        foreach (var rect in _rectangles)
+        {
+            if (rect.Contains (x, y))
+            {
+                return true;
+            }
+        }
+        return false;
+    }
 
     /// <summary>
     ///     Determines whether the specified rectangle is contained within the region.
     /// </summary>
     /// <param name="rectangle">The rectangle to check for containment.</param>
     /// <returns><c>true</c> if the rectangle is contained within the region; otherwise, <c>false</c>.</returns>
-    public bool Contains (Rectangle rectangle) { return _rectangles.Any (r => r.Contains (rectangle)); }
+    public bool Contains (Rectangle rectangle)
+    {
+        foreach (var rect in _rectangles)
+        {
+            if (rect.Contains (rectangle))
+            {
+                return true;
+            }
+        }
+        return false;
+    }
 
     /// <summary>
     ///     Returns an array of rectangles that represent the region.
--- a/Terminal.Gui/Drawing/Region.cs
+++ b/Terminal.Gui/Drawing/Region.cs
@@ -1,5 +1,7 @@
 ﻿using System.Buffers;
 
+#nullable enable
+
 /// <summary>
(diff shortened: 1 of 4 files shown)


Terminal.Gui/Drawing/Region.cs
```diff
@@ -1,4 +1,8 @@
-﻿/// <summary>
+﻿#nullable enable
```
tig, line 1: FWIW, in https://github.com/gui-cs/Terminal.Gui/pull/3886, i've rewritten Region, so this is is all moot. I'd love your help on that PR once I'm ready tho!

tig:
With:

image

Without:

image

2 sec improvement. :-)

PR #3921, opened 2025-02-25
Fixes #3918 and #3913 - `Accepting` behavior

### LINKED ISSUE #3905: When the `CanFocus` property is set to false, the `Accepting` event is executed 3 times
**To Reproduce**
1. Copy the code below
```csharp
var window = new Window();
window.Title = "test1";

var btn = new Button();
btn.Title = "test2";
btn.X = Pos.Center();
btn.Y = Pos.Center();
btn.CanFocus = false;
btn.Accepting += (_, _) => {
    Debug.WriteLine("accept");
};

window.Add(btn);

Application.Init();
Application.Run(window);
Application.Shutdown();
```
2. Run in your IDE or text editor

**Expected behavior**
The event will only be executed once

**Screenshots**
When executing the code above I have this in the Visual Studio output:
Image

If I comment out the CanFocus property code, I have this in the Visual Studio output:
Image

- OS: Windows 11 24H2 26100.2894
- Browser: Microsoft Edge
- Version: 132.0.2957.140

**Additional context**
It's in v2
Image

### LINKED ISSUE #3913: IsDefault on button makes it click itself on close
**Describe the bug**
When a View has 2 `Button` (A and B).  Both have `Accepting` handler.  One of which `IsDefault` then the default one is always fired even when clicking on the other button.

**To Reproduce**
```csharp
Application.Init ();

Window w = new Window () { Width = Dim.Fill (), Height = Dim.Fill () };

var btn1 = new Button ()
{
    Text = "A",
    IsDefault = true,
};
btn1.Accepting += (s, e) => {
                      Debug.WriteLine ("A pressed");
                      Application.RequestStop ();
                  };

var btn2 = new Button ()
{
    X = Pos.Right (btn1),
    Text = "B"
};
btn2.Accepting += (s, e) =>
                  {
                      Debug.WriteLine ("B pressed");
                      Application.RequestStop ();
                  };

w.Add (btn1);
w.Add (btn2);

Application.Run (w);
w.Dispose ();
Application.Shutdown ();

return 0;
```

When you click A you see

"Pressed A"

When you click B you see

"Pressed B"
"Pressed A"

Image

**Expected behavior**
Button should not click itself!

** Workaround **
As a workaround you can have all the buttons that are not IsDefault call e.Cancel on the accepting event.

```diff
btn2.Accepting += (s, e) =>
                  {
                      Debug.WriteLine ("B pressed");
+                    e.Cancel = true;
                      Application.RequestStop ();
                  };
```

### LINKED ISSUE #3918: Label triggers click event (V2)
Strange behavior. Depending in which order the label and the button are added to the window, clicking in the label calls the click event in the button. Am I missing something?

Example of the error happening in Windows 11:

        Application.Init();
        Window w = new Window() { Width = Dim.Fill(), Height = Dim.Fill() };
        var label = new Label() { Text = "Label: Click Here", X = 1, Y = 1 };
        var button = new Button { Text = "Button: click here", X = 1, Y = 5, IsDefault = true };
        button.Accepting += (sender, args) => { MessageBox.Query("", "clicked", "OK"); };
        
        if(true) //change to if(!true) to not see the error
            w.Add(label, button); //click error
        else
            w.Add(button, label); //no Click error

        Application.Run(w);
        w.Dispose();
        Application.Shutdown();

        return;

Image

tig:
## Fixes

- Fixes #3905
- Fixes #3913 
- Fixes #3918 

## Proposed Changes/Todos

- [x] Verify #3905 is not actually a bug.
- [x] Fix #3918 - Label should test if it actually has a HotKey or not.
- [x] Figure out #3913 - See comments in that Issue - Not fixing.

## Pull Request checklist:


UnitTests/View/ViewCommandTests.cs
```diff
@@ -124,6 +124,148 @@ public void MouseClick_Does_Not_Invoke_Accept_Command ()
...
+
+        int btnAcceptedCount = 0;
+        bool btnCancelAccepting = false;
+        var btn = new Button ()
+        {
+            Title = "Button",
+            Width = 3,
+            IsDefault = true,
+        };
+        btn.CanFocus = true;
```
BDisp, line 228: `btn.CanFocus` is always `true` here, so I suggest to change to an assertion:

`Assert.True (btn.CanFocus);`

Furthermore, the unit test doesn't really test for what the method name mentions, `Button_CanFocus_False_Raises_Accepted_Correctly` and you really wanted to mean `btn.CanFocus = false;`, which in that case the test will fail.

BDisp (changes requested):
Seems that `Button`'s `CanFocus` false isn't working as expected.

tznind (approved):
Hmn, so do I understand correctly that it is a new requirement for all button click handlers is for user to e.cancel in the handler? Otherwise any default buttons nearby may get click handler (now called Accepting) called?

We should double check there are no IsDefault issues with our existing dialogs like MessageBox and the Query one - if they have defaults we will need this new esc bit

I also think this is going to be confusing for new users. But happy to merge and see how it goes.

BDisp:
I agree with @tznind. This will be confusing for new users. My suggestion is to add a new event called `Accepted` which will be raised if the `Accepting` event wasn't been handled and so, not canceled. In this case most of the calls to the `Accepting` event will be replaced by the `Accepted` event.

UnitTests/View/ViewCommandTests.cs
tig: Correct. I forgot to update the test. The test IS named correctly. Now a comment says:

```cs
  // See: https://github.com/gui-cs/Terminal.Gui/issues/3905
  // There is still a bug here, but it's hidden by the fact that Accepting sets e.Cancel to true
  [Fact]
  public void Button_CanFocus_False_Raises_Accepted_Correctly ()
```

tig:
> Hmn, so do I understand correctly that it is a new requirement for all button click handlers is for user to e.cancel in the handler? Otherwise any default buttons nearby may get click handler (now called Accepting) called?

Correct. 

Note, previously, `e.Handed/e.Cancel` actually did either nothing or was non-deterministic. So while it was "easier to use" it was actually deeply broken.

> We should double check there are no IsDefault issues with our existing dialogs like MessageBox and the Query one - if they have defaults we will need this new esc bit

I already did this when I implemented all this.

> I also think this is going to be confusing for new users. But happy to merge and see how it goes.

The crux of this is the making `Accepted` universal and getting rid of explicit `Clicked` events. 

Things are now super consistent and deterministic, but there IS a bit of mental load required to "get it". 

I'm happy to discuss ways of mitigating this. Thots:

- Add an `Accept` event that is raised AFTER and can't be cancelled. 
- Re-introduce `Clicked`. I hate this idea.

Others?

Issue #3930, opened 2025-02-27
Tests are crashing again...


tig:
This is intermittent and non-deterministic.

![Image](https://github.com/user-attachments/assets/0bb33d45-cca9-4976-9b9a-b2c7fef776dd)

![Image](https://github.com/user-attachments/assets/5f3917c3-7b81-4858-b8fa-21bd431398d3)

I "fixed" this previously, but was not sure what I actually did to fix it.

Some test is failing in such a way that it's causing the test process to crash. Pretty sure (based on prior experience) it's NOT the test identified above as it changes from run to run.

If anyone has more experience with these things than I, I'd love help.

tig:
```
The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
The test running when the crash occurred: 
Terminal.Gui.ApplicationTests.ApplicationScreenTests.ClearScreenNextIteration_Resets_To_False_After_LayoutAndDraw
```

This seems consistent...

tig:
Reopening.

Still crashing. This is on #3952 after @BDisp's #3948 was merged:

https://github.com/gui-cs/Terminal.Gui/actions/runs/13617816995/job/38063151746?pr=3952

```
  Passed Terminal.Gui.ViewsTests.ScrollBarTests.Position_Event_Cancelables [1 ms]
[xUnit.net 00:00:07.61]     Terminal.Gui.ViewsTests.ScrollBarTests.KeepContentInAllViewport_True_False [SKIP]
[xUnit.net 00:00:07.61]       Disabled - Will put this feature in View
Error opening terminal: unknown.
The active test run was aborted. Reason: Test host process crashed : Error opening terminal: unknown.

Data collector 'Blame' message: System.IO.FileNotFoundException: Collect dump was enabled but no dump file was generated.
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.ProcessDumpUtility.GetDumpFiles(Boolean warnOnNoDumpFiles, Boolean processCrashed) in /_/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/ProcessDumpUtility.cs:line 84
   at Microsoft.TestPlatform.Extensions.BlameDataCollector.BlameCollector.SessionEndedHandler(Object sender, SessionEndEventArgs args) in /_/src/Microsoft.TestPlatform.Extensions.BlameDataCollector/BlameCollector.cs:line 534.

Test Run Aborted.
[...]

PR #3946, opened 2025-03-01
Fixes #3941 - v2win/v2net: UICatalog crash - Fixes warnings

### LINKED ISSUE #3941: v2win/v2net: UICatalog full benchmark suite crashes in Dialogs scenario
**Describe the bug**

When running the full benchmark suite with either the v2win or v2net driver the Dialogs scenario crashes. Running the Dialogs scenario alone does not crash most of the time but it is also prone to crash.

The Dialogs scenario seems to be stuck for a while before crashing. Sometimes instead of crashing it is just stuck there at least longer than my patience persists.

**To Reproduce**

Choose one of the commands and run it in Release build:
```
UICatalog.exe --driver v2win --benchmark
```
```
UICatalog.exe --driver v2net --benchmark
```

**Expected behavior**

The benchmark suite finishes without crashes.

**Screenshots**

Running the full benchmark suite with both v2 drivers in VS the debug output shows a StackOverflowException when the benchmark crashes after message flood. Of course no stack trace is available when stack overflows.

Image

After flooding for a while...

Image

Memory is flooded rapidly at the time it happens.

Image

**Desktop**
 - OS: Windows 10 22H2 (build 19045.5487)
 - CLI: Windows Terminal v1.21.10351.0 /w cmd
 - Terminal.GUI branch: v2_develop

tig:
## Fixes

- Fixes #3941
- Fixes warnings caused by Transparent PR
- Fixes warnings in other places

## Proposed Changes/Todos

- [x] See if adjusting the UI Catalog demo key pacing will fix

## Pull Request checklist:


tig:
> @tznind - I wonder if this is why we have sporadic test failures in CI, there is `All_Scenarios_Benchmark` and a stack overflow might cause the behaviour we see.

FWIW, I think I saw one failure after I implemented #3946. So I'm not declaring victory on #3930 yet.

But I think this does fix #3941

BenchmarkDotNet.Artifacts/BenchmarkRun-20250218-165908.log
```diff
@@ -0,0 +1,9 @@
+// Validating benchmarks:
```
BDisp, line 1: Are these .log files needed to belong to the repository. Isn't better to add them to the .`gitignore`?

tig: I agree. I added an exclude for `logs/` but I'll also add `*.log`

tig:
FWIW, #3930 is not related to this; still happens with the benchmark scenario test disabled.

UICatalog/Scenario.cs
```diff
@@ -149,7 +149,7 @@ public static ObservableCollection<Scenario> GetScenarios ()
     public virtual void Main () { }
 
     private const uint BENCHMARK_MAX_NATURAL_ITERATIONS = 500; // not including needed for demo keys
-    private const int BENCHMARK_KEY_PACING = 1; // Must be non-zero
+    private const int BENCHMARK_KEY_PACING = 10; // Must be non-zero
```
tznind, line 152: I think more robust fix is to look into the runstate changes that's actually happening that triggers the addition of a new timeout set of demo keys. E.g. add a flag or collection recording which demo keys have already been sent (`HashSet<Scenario>`, for example)

See also this comment https://github.com/gui-cs/Terminal.Gui/issues/3941#issuecomment-2692502887

tig: Agree. I'll merge this now and we can then do it right via #3949

PR #3949, opened 2025-03-02
Fixes 3941 - Do not repeatedly raise new sets of keystrokes when modals pop (stack overflow)

tznind:
This PR is to explore alternative to #3946

First Fix
---------
Do not repeatedly raise new sets of keystrokes when modals pop and trigger run state changes.

This fixes the infinite loop but now _with all drivers (v2 and original)_ the Scenarios hang on any scenario that fires up a modal dialog.

## Fixes

- Fixes #_____

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:

### DIFF
--- a/UICatalog/Scenario.cs
+++ b/UICatalog/Scenario.cs
@@ -229,13 +229,20 @@ public class Scenario : IDisposable
         }
     }
 
+    private HashSet<Scenario> _scenariosRun = new HashSet<Scenario> ();
     private void OnApplicationNotifyNewRunState (object? sender, RunStateEventArgs e)
     {
+        // We are just returning to the same scenario
+        if (!_scenariosRun.Add (this))
+        {
+            return;
+        }
+
         SubscribeAllSubviews (Application.Top!);
 
         _currentDemoKey = 0;
         _demoKeys = GetDemoKeyStrokes ();
-
+        
         Application.AddTimeout (
                                 new TimeSpan (0, 0, 0, 0, BENCHMARK_KEY_PACING),
                                 () =>


tznind:
@tig what is supposed to exit benchmark scenarios? is demokeys supposed to send Esc after each run or is it something else? I can see the abort code but not the 'normal' exit conditions for each scenario

tig:
There's an iteration limit.

tig:
I added the keystroke stuff later. The original idea was to just find code that was excessively causing draws. 

Then I added code to make the scenarios actually do stuff to expose more of that (demokeys).

Then I decided it would make a good benchmark framework. 

It was not all that well thought through ☺️

tig:
Oh, this made me realize : with your new driver model there ARE NO iterations if there's no need.

tznind:
There are iterations but no draws or layouts.  So only thing that fires are timed events.

But v2 use Metrics not benchmark for tracking frequency... maybe I just need to tap into the benchmark counters too.

tig:
FWIW, I'm intentionally re-running the Actions on all PRs to try to repro the test failures.  I'm letting you know so you don't get freaked out if I quickly update these branches etc...

tznind:
ok v2 with no frame limiter seems to behave, it zips through the required iterations so fast its hard to see what is going on.  There is no pause in Dialogs scenario.

WindowsDriver still runs into stop at the dialogs.

Dialogs has mix of keys - some of which should be handled in main Scenario and some get handled by the modal.  But I think that is brittle and/or not working.

It used to work because every time you get run state change it raises the current demo keys again.  This means when the dialog popped up it ran the Dialogs keys again which includes Enter (which would close the Dialog anyway)... that is my theory anyway.

Not sure whether v2 is going so fast it doesn't pop the Dialog in the first place or pops it fast enough to catch the requisite close events.

Is it possible to benchmark a specific scenario? Maybe verbose logging will tell us.

tig:
> Is it possible to benchmark a specific scenario? Maybe verbose logging will tell us.

Yea, just add the Scenario name to the UI Catalog command line.

PR #3962, opened 2025-03-06
Cleans up/Refactors `View.Subviews` 

tig:
There is a lot of cruft in how Subviews (Add/Remove etc...) work. In order to work on these, I need this cleaned up:

- #3852 
- #2994 
- #2491 

## Proposed Changes/Todos

- [x] Make all core code use `InternalSubviews` vs. `Subviews`
- [x] Reorg unit tests
- [x] Code and API doc cleanup

## Pull Request checklist:


tig:
Folks (@BDisp , @tznind , etc...)

It's ALWAYS driven me nuts that `View.Subviews` has a lowercase V (or that `SuperView` has an upper case V).

Can I fix this, please?

BDisp:
> Folks (@BDisp , @tznind , etc...)
> 
> It's ALWAYS driven me nuts that `View.Subviews` has a lowercase V (or that `SuperView` has an upper case V).
> 
> Can I fix this, please?

I like more `SubViews` and `SuperView`.

tig:
Thanks.

Here's another topic I'd like your opinion on:

Currently the events around `View.Add/Remove` are:

```cs
    /// <summary>Event fired when this view is added to another.</summary>
    public event EventHandler<SuperViewChangedEventArgs>? Added;

    /// <summary>Method invoked when a subview is being added to this view.</summary>
    /// <param name="e">Event where <see cref="ViewEventArgs.View"/> is the subview being added.</param>
    public virtual void OnAdded (SuperViewChangedEventArgs e)
    {
        View view = e.SubView;
        view.IsAdded = true;
        view.Added?.Invoke (this, e);
    }

    /// <summary>Event fired when this view is removed from another.</summary>
    public event EventHandler<SuperViewChangedEventArgs>? Removed;

    /// <summary>Method invoked when a subview is being removed from this view.</summary>
    /// <param name="e">Event args describing the subview being removed.</param>
    public virtual void OnRemoved (SuperViewChangedEventArgs e)
    {
        View view = e.SubView;
        view.IsAdded = false;
        view.Removed?.Invoke (this, e);
    }
```

Note that these are super confusing and do not match standard event patterns.

- `OnAdded` is called when a SubView is being added to `this`
- `Added` is raised ON THE SubView when the SubView has been added to `this`

etc...

There are NO overrides of either virtual method in our codebase.

The uses of the events (`Added` and `Removed`) are few. And none of them use the eventargs passed; they only use the events as a signal.  Eg. `TextField`:

```cs
    private void TextField_Added (object sender, SuperViewChangedEventArgs e)
    {
        if (Autocomplete.HostControl is null)
        {
            Autocomplete.HostControl = this;
            Autocomplete.PopupInsideContainer = false;
        }
    }
```

This can and should be simplified.

My proposal:

- Call `OnAddedChanged`/Raise `IsAddedChanged` when `this.IsAdded` changes. Make it `EventArg<bool>`. 
- Call `OnSubViewAdded`/Raise `SubViewAdded` when `this.Add` is called. This can continue to use `SuperViewChangedEventArgs`.

(Same for Remove).

thots?

BDisp:
I think that way you're duplicating the event raising twice because `IsAddedChanged` is the same as using the `SubViewAdded` . In my opinion the methods `OnXXX` should be `private void` because a derived class would not have any reason to override it and broken the event to be raised.

BDisp:
I would use the `private void OnSubViewAdded` which raise `SubViewAdded`.

tig:
> I think that way you're duplicating the event raising twice because `IsAddedChanged` is the same as using the `SubViewAdded` . In my opinion the methods `OnXXX` should be `private void` because a derived class would not have any reason to override it and broken the event to be raised.

Please read https://gui-cs.github.io/Terminal.GuiV2Docs/docs/events.html

A subclass of View may want to know when it's been added to a superview.

With my proposal it can either:

a) override `OnIsAddedChanged`

Or

b) subscribe to `IsAddedChanged`

Take a look at the latest code i pushed here and let me know what you think.

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -59,7 +59,7 @@ public ComboBox ()
         // On resize
         SubviewsLaidOut += (sender, a) => ProcessLayout ();
 
-        Added += (s, e) =>
+        IsAddedChanged += (s, e) =>
```
BDisp, line 62: Since this is a derived class it could override the `OnIsAddedChanged`.

Terminal.Gui/Views/Menu/MenuBar.cs
```diff
@@ -84,7 +84,7 @@ public MenuBar ()
         WantMousePositionReports = true;
         IsMenuOpen = false;
 
-        Added += MenuBar_Added;
+        IsAddedChanged += MenuBar_IsAddedChanged;
```
BDisp, line 87: Since this is a derived class it could override the `OnIsAddedChanged`.

Terminal.Gui/Views/TextField.cs
```diff
@@ -45,9 +45,7 @@ public TextField ()
 
         Initialized += TextField_Initialized;
 
-        Added += TextField_Added;
-
-        Removed += TextField_Removed;
+        IsAddedChanged += TextField_IsAddedChanged;
```
BDisp, line 48: Since this is a derived class it could override the `OnIsAddedChanged`.

Terminal.Gui/Views/TextView.cs
```diff
@@ -1899,7 +1899,7 @@ public TextView ()
 
         Initialized += TextView_Initialized!;
 
-        Added += TextView_Added!;
+        IsAddedChanged += TextView_IsAddedChanged!;
```
BDisp, line 1902: Since this is a derived class it could override the `OnIsAddedChanged`.

BDisp:
> A subclass of View may want to know when it's been added to a superview.
> 
> With my proposal it can either:
> 
> a) override `OnIsAddedChanged`
> 
> Or
> 
> b) subscribe to `IsAddedChanged`
> 
> Take a look at the latest code i pushed here and let me know what you think.

Understood. So I see that protect virtual methods has precedence to events. Only for to be sure.

BDisp:
A unit test that you maybe could add into the `SubviewTests.cs`:

```cs
    [Fact]
    public void IsAdded_IsAddedChanged_SubViewAdded_SubViewRemoved ()
    {
        var isAdded = false;

        View superView = new () { Id = "superView" };
        View subView = new () { Id = "subView" };

        superView.SubViewAdded += (s, e) =>
                                  {
                                      Assert.True (isAdded);
                                      Assert.True (subView.IsAdded);
                                      Assert.Equal (subView, e.SubView);
                                      Assert.Equal (superView, e.SuperView);
                                  };

        superView.SubViewRemoved += (s, e) =>
                                    {
                                        Assert.False (isAdded);
                                        Assert.False (subView.IsAdded);
                                        Assert.Equal (subView, e.SubView);
                                        Assert.Equal (superView, e.SuperView);
                                    };

        subView.IsAddedChanged += (s, e) => { isAdded = e.CurrentValue; };

        superView.Add (subView);
        Assert.True (isAdded);
        Assert.True (subView.IsAdded);
        Assert.Single (superView.Subviews);

        superView.Remove (subView);
        Assert.False (isAdded);
        Assert.False (subView.IsAdded);
        Assert.Empty (superView.Subviews);
    }
```

tig:
@tznind it's important you review this and are ok with it. Esp:

- Changing the event model for `View.Add/Remove`
- Renaming `Subview` to `SubView`

Sorry so many other files got touched. I got motivated to clean up a bunch of warnings in unit tests and the `Subview` rename too.

Thanks

Terminal.Gui/Views/ComboBox.cs
```diff
@@ -57,14 +57,14 @@ public ComboBox ()
         Initialized += (s, e) => ProcessLayout ();
 
         // On resize
-        SubviewsLaidOut += (sender, a) => ProcessLayout ();
+        SubViewsLaidOut += (sender, a) => ProcessLayout ();
 
-        Added += (s, e) =>
+        IsAddedChanged += (s, e) =>
```
tznind, line 62: Why do we need an IsAdded field? doesn't the SuperView implicitly tell you this.

If you have a SuperView you have been Added to something otherwise you have not.

Terminal.Gui/Views/Menu/MenuBar.cs
```diff
@@ -559,7 +559,7 @@ private void CloseOtherOpenedMenuBar ()
         if (Application.Top is { })
         {
             // Close others menu bar opened
-            Menu? menu = Application.Top.Subviews.FirstOrDefault (v => v is Menu m && m.Host != this && m.Host.IsMenuOpen) as Menu;
+            Menu? menu = Application.Top.SubViews.FirstOrDefault (v => v is Menu m && m.Host != this && m.Host.IsMenuOpen) as Menu;
             menu?.Host.CleanUp ();
```
tznind, line 563: We have Adornment.Parent and Menu.Host and other similar.

We should standardize on only 2 conditions

- View that is added like normal to the hierarchy, use `SuperView`
- View that is special e.g. a border, a menu, a context popup etc etc, use standard term e.g. `Parent`

Even better if we add an interface for these views `IViewSubcomponent` or something

Terminal.Gui/Views/Slider.cs
```diff
@@ -59,7 +59,7 @@ private void SetInitialProperties (
         // BUGBUG: This should not be needed - Need to ensure SetRelativeLayout gets called during EndInit
         Initialized += (s, e) => { SetContentSize (); };
 
-        SubviewLayout += (s, e) => { SetContentSize (); };
+        SubViewLayout += (s, e) => { SetContentSize (); };
```
tznind, line 62: what is difference between Layout and LaidOut?

Terminal.Gui/View/View.Hierarchy.cs
```diff
@@ -11,39 +11,77 @@ public partial class View // SuperView/SubView hierarchy management (SuperView,
...
+    internal IList<View> InternalSubViews => _subviews ?? _empty;
 
-    /// <summary>This returns a list of the subviews contained by this view.</summary>
-    /// <value>The subviews.</value>
-    public IList<View> Subviews => _subviews?.AsReadOnly () ?? _empty;
+    /// <summary>Gets the list of SubViews.</summary>
+    /// <remarks>
+    ///     Use <see cref="Add(Terminal.Gui.View?)"/> and <see cref="Remove(Terminal.Gui.View?)"/> to add or remove subviews.
+    /// </remarks>
+    public IList<View> SubViews => _subviews?.AsReadOnly () ?? _empty;
```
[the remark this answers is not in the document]
tig: Not really, without requiring uses to do this instead of `cb.SubViews[0]`.

```cs
cb.SubViews.ToArray() [0].ColorScheme.Focus,
```

Thots?

Terminal.Gui/Views/ComboBox.cs
tig: Good question. `IsAdded` has been there for a while.

@BDisp do you remember?

Terminal.Gui/View/View.Hierarchy.cs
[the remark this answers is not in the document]
tznind: Can do `cb.SubViews.ElementAt(0).ColorScheme.Focus`.  Otherwise it just looks like you can add to the list but then when you do it crashes presumably at runtime?

Terminal.Gui/Views/Menu/MenuBar.cs
tig: I agree. The menu/popup menu is being fixed in

- #2975

Terminal.Gui/Views/Slider.cs
tig: `Layout` == Layout is starting
`LaidOut` == Layout is complete

I struggled with this naming given the standard guidance on naming that says don't use "Starting/Completed".

tznind: Hmn... ok lets go with it.

Terminal.Gui/View/View.Hierarchy.cs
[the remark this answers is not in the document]
tig: Done

Terminal.Gui/Views/ComboBox.cs
BDisp: Yes I remember. It was added since many years on the v1. We all know that if a `SuperView` is null then the view wasn't added yet but there was a necessity of raise some code in the exact moment a view is added to a SuperView by raising an event. Without raising this event a view needed to check periodically, using a timer or timeout, if SuperView is not null.

BDisp: But I understand that the `IsAdded` property can be removed, since it's easy to check that. Only the virtual method and event is important.

Terminal.Gui/View/View.Hierarchy.cs
```diff
@@ -7,43 +7,81 @@ namespace Terminal.Gui;
...
+    private static readonly IReadOnlyCollection<View> _empty = [];
 
     private List<View>? _subviews; // This is null, and allocated on demand.
 
-    // Internally, we use InternalSubviews rather than subviews, as we do not expect us
-    // to make the same mistakes our users make when they poke at the Subviews.
-    internal IList<View> InternalSubviews => _subviews ?? _empty;
+    // Internally, we use InternalSubViews rather than subviews, as we do not expect us
+    // to make the same mistakes our users make when they poke at the SubViews.
+    internal IList<View> InternalSubViews => _subviews ?? [];
```
BDisp, line 16: I think that `InternalSubViews` also can be `IReadOnlyCollection` instead of `IList`.

tig: > I think that `InternalSubViews` also can be `IReadOnlyCollection` instead of `IList`.

No. It's bad that internal code uses _subviews directly. See latest push which makes all internal usages use InternalSubViews and it's def not read only.

tig:
Ok, look again.

* Nuked View.IsAdded.
* Added View.SuperViewChanged.

BDisp (approved):

PR #4013, opened 2025-04-02
Partial #2975 - Deprecate v1 `Menu` classes 

tig:
## Fixes

- Partial #2975

## Proposed Changes/Todos

- [x] Fix `UICatalog` bugs in `v2win/v2net`
- [x] Add Fluent menubarv2 tests
- [ ] Mark v1 classes as obsolete

## Pull Request checklist:


tig:
BQ2Yb5W 1

In the above the Diagnostics menu's `FlagSelector` menu item has been tweaked to do the right thing.

The other two menus exhibit the old, broken behavior.

tig:
This is fun.

WindowsTerminal_wPgdpbfiN8

tig:
And, of course "None" works:

https://i.imgur.com/gjx5yuX.png

BDisp (approved):

tznind (approved):

tig:
ERUMT17 1

New `MenuBarv2.EnableForDesign` menu and `Menus Scenario`

tznind:
I have also fixed [92aae38](https://github.com/gui-cs/Terminal.Gui/pull/4013/commits/92aae38c2e234de77bbe76108e37363770373c28) in #4029

tig:
> I have also fixed [92aae38](https://github.com/gui-cs/Terminal.Gui/pull/4013/commits/92aae38c2e234de77bbe76108e37363770373c28) in #4029 (in [58b0178](https://github.com/gui-cs/Terminal.Gui/commit/58b0178e4500402bb883b3c840604d810de661d2)) with various followups in TableView ect and delete dead code.
> 
> Only mentioning to try and avoid duplicated effort/merge conflicts

Thanks.

Please look at the fluent test changes here, as well as the changes to ApplicationV2 and drivers. I found some hard to track down issues and fixed them. 

Especially:

```cs
public class ApplicationV2Tests
{

    private ApplicationV2 NewApplicationV2 ()
    {
        var netInput = new Mock<INetInput> ();
        SetupRunInputMockMethodToBlock (netInput);
        var winInput = new Mock<IWindowsInput> ();
        SetupRunInputMockMethodToBlock (winInput);

#if DEBUG_IDISPOSABLE
        // Always set this in tests. Because this wasn't set, these tests were not catching
        // that Application.Top was not being disposed.
        View.DebugIDisposable = true;
#endif
```

I added `View.DebugIDisposable` a while back so that the `DEBUG_IDISPOSABLE` behavior could be controlled at runtime. I don't remember precisely why, but it was important :-). I also don't remember why I set it to false by default. I may try switching it to default to true to see what happens...

The crux is that if tests don't set it, then places where Views are not being disposed properly will not be caught.

AND, if another test DOES set it, since it's a static, then those same tests will weirdly fail in confusing ways.

It turns out that in the v2Application tests and in the Fluent stuff, you were not disposing Application.Top before calling Shutdown. I spent hours tracking this down yesterday.

tig:
Figured out why I added it:

I didn't want the Parallelizable tests all have to explicitly Dispose stuff.

I've just implemented what I think is a better solution:

- `View.DebugIDisposable` is now `true` by default
- We now have a global test fixture that causes any test class using the `[Collection ("Global Test Setup")]` to set `View.DebugIDisposable` to `false`.
- You can also just have the test class inherit from `ParallelizableBase` which adds that attribute.

I didn't add the attribute to all parallelizable test classes, but we should. I added it to those that seemed to be causing issues and have "run tests till failure" a bunch with no failures. We can add it to the rest over time.

tznind:
> Please look at the fluent test changes here, as well as the changes to ApplicationV2 and drivers. I found some hard to track down issues and fixed them.

One of the things I was trying to do in the v2 driver was simplify things.  The `RunState` class confused me.  I thought that:

- If you are in `Application.TopLevels` then you are running
- If you are the head of `Application.TopLevels` and/or are `Application.Top` then you are the topmost one

So what is RunState for?

But I also missed some other Init events on TopLevel views and stuff - thanks for fixing, I'm sorry it took hours to figure out.

But also if there is a way to simplify things I would be all for that 🙏 

>  I added View.DebugIDisposable a while back [...] may try switching it to default to true to see what happens...

I guess the reason to have false in production release at least is because throwing when a user missed disposing a view is a bit unfriendly of a library.  Typically if you don't dispose something you might get a memory leak at worst but normally not a hard crash.

tig:
> > Please look at the fluent test changes here, as well as the changes to ApplicationV2 and drivers. I found some hard to track down issues and fixed them.
> 
> One of the things I was trying to do in the v2 driver was simplify things. The `RunState` class confused me. I thought that:
> 
> * If you are in `Application.TopLevels` then you are running
> * If you are the head of `Application.TopLevels` and/or are `Application.Top` then you are the topmost one
> 
> So what is RunState for?

We are both working on simplifying. You made a bunch of progress on

- #2491

With the v2 stuff. But we're not really coordinated in this. I think what I just hit is just progress on that. 

Can you please re-read that Issue and maybe comment in it how you envision your V2Application stuff helping to address it. Note I'm not stuck on the concept of an `IRunnable` interface; that's just a concept to make the point that the current arch, based on `RunState` is confusing and complex.

> But also if there is a way to simplify things I would be all for that 🙏

Yep, yep, and yep!

> I guess the reason to have false in production release at least is because throwing when a user missed disposing a view is a bit unfriendly of a library. Typically if you don't dispose something you might get a memory leak at worst but normally not a hard crash.

`View.DebugIDisposable` is only defined in #DEBUG AND #DEBUG_IDISPOSABLE

BDisp:
@tig can you pay attention that popover when is closed doesn't restore the focus to the previous focused view. Also if Esc is pressed to close a popover closes the toplevel instead closing only the popover. Thanks.

BDisp:
@tig to avoid this kind of issue in the future I would suggest to add this assertion to the `View.Navigation.cs` file:

image

BDisp:
@tig just a few things I want to ask if you'll handle them in this PR:
- Menu stays always open when clicked second time. F9 work as expected closing the menu when pressing F9 second time.
- Menu radio botton aren't checked after click in the UICatalog themes.

BDisp:
You need to be more clear. I don't understand what you are suggesting.

In the UICatalog if you click in the Dark theme for i.e. the radio button wont be flagged as checked.

tig:
> You need to be more clear. I don't understand what you are suggesting.
> 
> In the UICatalog if you click in the Dark theme for i.e. the radio button wont be flagged as checked.

Ahhh. Yep, thats one of the bugs that led me to realize I had some deep design errors that needed to be corrected.

tig:
This PR is stale. Closing.

PR #4015, opened 2025-04-03
Fixes #4009 - fix tree ordering

tznind:
## Fixes

- Fixes #4009 

## Proposed Changes/Todos

Change from `Dictionary<T,Branch<T>>` to `List<Branch<T>>`.

The refreshing process now always sorts list after making pruning/additions

Tradeoffs:
- Slower
- Respects order

## Why are we tracking branches anyway?

At first look `Branch.Refresh` looks overly complicated.  You would be forgiven for asking why we don't just always create new branches instead of prune/add.

The reason is simply to preserve the expansion/collapsed state of the branch and the sub branches thereof (all the way down).

## Other Additions/Fixes

- Added a `Focus` method to the fluent API
- Made `Then` call `Stop` on `Exception` which helps avoid threading issues in crash scenarios
- Fix net key presses e.g. `Right` not `WaitIteration` which is behavioural difference vs win test driver.
- Moved `TestOutputWriter` to its own file to make it reusable

## Pull Request checklist:


Terminal.Gui/Views/TreeView/Branch.cs
```diff
@@ -27,7 +27,7 @@ public Branch (TreeView<T> tree, Branch<T> parentBranchIfAny, T model)
     ///     The children of the current branch.  This is null until the first call to <see cref="FetchChildren"/> to avoid
```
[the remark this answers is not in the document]
tig: Please use this as an opportunity to `#nullable enable` this class.

TerminalGuiFluentTesting/GuiTestContext.cs
```diff
@@ -243,7 +243,18 @@ public GuiTestContext WaitIteration (Action? a = null)
     /// <returns></returns>
     public GuiTestContext Then (Action doAction)
     {
-        doAction ();
+        try
```
tig, line 246: In #4013 I've been playing with this and it's VERY COOL!

Eg. 

https://github.com/gui-cs/Terminal.Gui/blob/01b526691a874924712ffe156278b4e60345ba27/Tests/IntegrationTests/FluentTests/MenuBarv2Tests.cs#L233-L271

However, as you'll note in that PR, I had to modify `GuiTestContext` to fix some issues. You should look at that (and comment).

Also, I think it would be great if there was a `GuiTestContext` member called `Assert` that somehow neatly integrated with `xUnit` so:

```diff
.SendKey (MenuBarv2.DefaultKey)
-.Then (() => Assert.True (Application.Popover?.GetActivePopover () is PopoverMenu))
-.Then (() => Assert.True (menuBar?.IsOpen()))
-.Then (() => Assert.Equal ("_New file", Application.Navigation?.GetFocused ()!.Title))
+.Assert.True (Application.Popover?.GetActivePopover () is PopoverMenu)
+.Assert.True (menuBar?.IsOpen())
+.Assert.Equal ("_New file", Application.Navigation?.GetFocused ()!.Title)
.ScreenShot ($"After {MenuBarv2.DefaultKey}", _out)
````

(Or roughly equivalent that is much cleaner).

File: TerminalGuiFluentTesting/GuiTestContext.cs [hunk 3]
```diff
@@ -550,4 +562,25 @@ private void SendWindowsKey (ConsoleKeyMapping.VK specialKey)
...
+    /// <summary>
+    /// Sets the input focus to the given <see cref="View"/>.
+    /// Throws <see cref="ArgumentException"/> if focus did not change due to system
+    /// constraints e.g. <paramref name="toFocus"/>
+    /// <see cref="View.CanFocus"/> is <see langword="false"/>
+    /// </summary>
+    /// <param name="toFocus"></param>
+    /// <returns></returns>
+    /// <exception cref="ArgumentException"></exception>
+    public GuiTestContext Focus (View toFocus)
```
tig, line 575: Is there some way to not have to re-implement EVERY action on `Application` in `GuiTestContext`? 

You added `Focus`. I added `SendKey`. Next we'll need more mouse, navigation, etc... methods....

Terminal.Gui/Views/TreeView/Branch.cs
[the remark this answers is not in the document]
tznind: I can take a look, but Branch is not a view itself so needs the TreeView to pass in the drawing classes so it can operate.

TerminalGuiFluentTesting/GuiTestContext.cs
tznind: > [...] I think it would be great if there was a GuiTestContext member called Assert that somehow neatly integrated with xUnit so:

Hmn should be able to do this, I originally avoided it to mean that the csproj was not bound to xunit and e.g. could be used with just nunit or other testing frameworks.

But we could probably still have that by having 2 csproj

- TerminalGuiFluentTesting
- TerminalGuiFluentTesting.Xunit

The Xunit one would have the methods you sugested implemented as extension methods.  That would make it possible to use either approach and/or add a `TerminalGuiFluentTesting.Nunit` later if we wanted.

TerminalGuiFluentTesting/GuiTestContext.cs
tznind: Hmn... its a good question.  I am not sure.

I wanted to do 'whole system testing' so the methods like Down etc actually synthesise the net/windows raw inputs - this exercises the full input processor pipe etc.

SendKey steps away from that but I think thats sensible i.e. just practical.

I will have a think but I can't think of an immediate way to expose e.g. all members of Application.

Terminal.Gui/Views/TreeView/Branch.cs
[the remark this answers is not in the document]
tig: Fair enough (assuming it CANT be a View). But imo passing an IConsoleDriver around breaks encapsulation. If there's a way we can extend the View drawing API that supports this use case
I'd love to investigate that.

Terminal.Gui/Views/TreeView/Branch.cs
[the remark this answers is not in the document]
tznind: Done

Terminal.Gui/Views/TreeView/Branch.cs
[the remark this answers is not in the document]
tznind: Now just calls relevant methods on TreeView.  No driver or ColorScheme injection.

Branch doesnt want to ve a View - we might have hundreds of thousands of branches e.g. if you ExpandAll in File System Explorer a root drive (get ready to wait on your file system loading entire planet - like 5 mins or more).  

But once the enumeration is finished though the TreeView performs fine navigating around in it.  This is because scroll offset and draw only happens for things in the viewport - there is no view layout for branches etc.  Branch remains lightweight as intended.

TerminalGuiFluentTesting/GuiTestContext.cs
tznind: Ok I've dug into this and most maintainable way to do this is with CodeGenerator

image

I have created proof of concept that iterates all AssertEqual methods in xunit and creates the relevant fluent versions in the TerminalGuiFluentTestingXunit repo:

https://github.com/tznind/gui.cs/pull/175


Alternative is to manually code these as and when we need the various overloads - which is dangerous (error prone).

Code gen has dangers of its own but typically they are detected at compile time

tznind: I think we should take the assert use case and 'how do we not replicate entire IApplication API surface as a seperate issue for now.  So i have marked this ready for review. 

The codegen is cool but might be overkill and really getting quite far from this original bug which is about branch reordering

PR #4020, opened 2025-04-05
Xunitcodegen

tznind:
## Fixes

- Fixes #4019 

This is a super set of #4015 which is when I started working on this

## Noteworthy

I had to change the package version from `4.13,5` to `4.12.0` on the CSharp and Analyzers code gen dependencies - otherwise I got warnings.  I don't see any issues doing this but am happy to discuss if others have any build issues.

UPDATE: CI wanted a different version so I set it in the props to use either which seems to work with locally and in CI now.

I created this repo to help understand how codegen works: https://github.com/tznind/CodeGenerationExample

## Proposed Changes/Todos

Adds all xunit methods (e.g. `Assert.False` to the fluent API as concatenated name e.g. `AssertFalse`)

image

## How does it work? 

We create an xunit project that is empty

- TerminalGuiFluentTestingXunit

We create a code generator project that creates code in the empty project

- TerminalGuiFluentTestingXunit.Generator

Result is that when compiled the empty project contains generated static extension methods to the fluent context that perform the asserts and shutdown safely:

image

## Pull Request checklist:

- [x] I've named my PR in the form of "Fixes #issue. Terse description."
- [x] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [x] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [x] I ran `dotnet test` before commit
- [x] I have made corresponding changes to the API documentation (using `///` style comments)
- [x] My changes generate no new warnings
- [x] I have checked my code and corrected any poor grammar or misspellings
- [x] I conducted basic QA to assure all features are working

PR #4024, opened 2025-04-07
Fixes #4022 file dialog tests and bugfix for cancellation

### LINKED ISSUE #4022: Cancel button not canceling FileDialog
**Describe the bug**
Pressing 'Cancel' button on FileDialog does not cancel the dialog.

**To Reproduce**
1. Run Editor scenario in UICatalog
2. Click on 'File', 'Save As'
3. Click 'Cancel'
4. Logic for 'Save' executes

**Expected behavior**
Dialog cancels quietly.

**Screenshots**
Image

**Build Version:**
 - 2.0.0-develop.4406

**Workaround**

The button works as expected by adding the following line in Terminal.Gui.FileDialog:
```csharp
        _btnCancel.Accepting += (s, e) =>
                                {
                                    if (e.Cancel)
                                    {
                                        return;
                                    }

                                    if (Modal)
                                    {
                                        e.Cancel = true; // Adding this makes it work
                                        Application.RequestStop ();
                                    }
                                };
```

tznind:
## Fixes

- Fixes #4022 
- Fixes # #4026

This is a super set of 

- #4020 

## Proposed Changes/Todos

- Fix current directory default to use current dir of IO abstractions (works better with tests)
- Add constructor overload fo IO abstraction to subclass SaveFileDialog
- Add fluent tests
- Fix actual bug!

## Pull Request checklist:


Tests/IntegrationTests/FluentTests/TreeViewFluentTests.cs
```diff
@@ -59,10 +64,14 @@ public void TreeView_AllowReOrdering (V2TestDriver d)
...
+                                 () =>
+                                 {
+                                     Assert.Equal (root, tv.GetObjectOnRow (0));
+                                     Assert.Equal (bike, tv.GetObjectOnRow (1));
+                                     Assert.Equal (car, tv.GetObjectOnRow (2));
+                                     Assert.Equal (lorry, tv.GetObjectOnRow (3));
+                                 })
                 .WriteOutLogs (_out);
 
         context.Stop ();
```
tig, line 77: There's an underlying problem here, that the old non-parallel unit tests have too:

Assert stops the test code. Thus shutdown never happens. This is the root of subsequent tests not being able to run when there's a failure. 

I wonder if there's a way to fix this via your codegen stuff such that if we do an `Assert` we ALWAYS also do an `Applicaton.ResetState(true)`?

tznind: The codegen wraps the method in try/catch with hard stop and rethrow so should be ok

tznind:
I've added a beautiful new method @tig 

It sends console raw input accurate Tab events till you hit a view that matches the lamda.  Or throws if it loops round completely and/or gets a null most focused.

```diff
[Theory]
[ClassData (typeof (V2TestDrivers))]
public void CancelFileDialog_UsingCancelButton_TabThenEnter (V2TestDriver d)
{
    var sd = new SaveDialog (CreateExampleFileSystem ());
    using var c = With.A (sd, 100, 20, d)
                      .ScreenShot ("Save dialog", _out)
+                      .Focus <Button>(b=> b.Text == "_Cancel")
                      .Enter ()
                      .Stop ();

    Assert.True (sd.Canceled);
}
```

tznind:
Ok got them all, as I suspected the other buttons on the page e.g. tree expand also passively trigger the IsDefault button.  Because default buttons fire whenever anything does `Accepting` regardless of whether view is closing etc.

PR #4029, opened 2025-04-13
Fixes #4027. Add collection search matcher

### LINKED ISSUE #4027: ListView collection navigator not overridable
It took as a while to understand that key strokes on a list view are always intercepted and forwarded to the collection navigator. Unfortunately this means when I want to intercept key strokes on a list view in a different way I have no possibility to do so but have to fall back to the application level key down events. 

The collection navigator also looses the actual Key information while doing so which means on the search string I can't different a 'b' from a 'B' as an example. 

If the collection navigator would be overridable I could at least return one that never gets a matching item by always returning -1 which then would return false on the `OnKeyDown` and therefore raising the KeyDown event if I understand correctly.

tznind:
## Fixes

- Fixes #4027

image

## Proposed Changes/Todos

This PR empower users to write custom navigator behaviour for `ListView`, `TableView` etc without having to 're invent the wheel'.

We move the most user servicable bits to a new class  `ICollectionNavigatorMatcher`
- Which keys trigger navigation
- How to match typed strings to objects in the collection

Also incidentally fixed:

- Fix unit test spelling of apricot

## Pull Request checklist:


Terminal.Gui/Views/FileDialogCollectionNavigator.cs
```diff
@@ -0,0 +1,21 @@
+namespace Terminal.Gui;
+
+internal class FileDialogCollectionNavigator (FileDialog fileDialog, TableView tableView) : CollectionNavigatorBase
+{
+    protected override object ElementAt (int idx)
```
tig, line 5: again, `object`???

tig (changes requested):

Terminal.Gui/Text/ICollectionNavigatorMatcher.cs
```diff
@@ -0,0 +1,26 @@
...
+    bool IsCompatibleKey (Key a);
+
+    /// <summary>
+    ///     Return true if the <paramref name="value"/> matches (e.g. starts with)
+    ///     the <paramref name="search"/> term.
+    /// </summary>
+    /// <param name="search"></param>
+    /// <param name="value"></param>
+    /// <returns></returns>
+    bool IsMatch (string search, object value);
```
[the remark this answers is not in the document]
tznind: I can look into it, but I think not.

This matcher is used by both ListView and TreeView.

I think the only way to be typesafe is to migrate a lot of classes to generics, possibly including ListView itself - which is a much larger piece of work.

Terminal.Gui/Text/IListCollectionNavigator.cs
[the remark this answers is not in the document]
tznind: The code being refactored was in this namespace to begin with.  I can move it if you want.

It is used by both TableView and ListView and TreeView (see class diagram pic in PR description)

Terminal.Gui/Text/DefaultCollectionNavigatorMatcher.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <inheritdoc/>
+    public bool IsMatch (string search, object? value) { return value?.ToString ()?.StartsWith (search, Comparer) ?? false; }
+
+    /// <summary>
+    ///     Returns true if <paramref name="a"/> is a searchable key (e.g. letters, numbers, etc) that are valid to pass
+    ///     to this class for search filtering.
+    /// </summary>
+    /// <param name="a"></param>
+    /// <returns></returns>
+    public bool IsCompatibleKey (Key a)
```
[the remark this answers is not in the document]
tznind: Done

Terminal.Gui/Text/CollectionNavigatorBase.cs
```diff
@@ -166,8 +130,7 @@ public static bool IsCompatibleKey (Key a)
     public virtual void OnSearchStringChanged (KeystrokeNavigatorEventArgs e) { SearchStringChanged?.Invoke (this, e); }
 
     /// <summary>This event is invoked when <see cref="SearchString"/>  changes. Useful for debugging.</summary>
```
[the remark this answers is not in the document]
tznind: Done

Terminal.Gui/Text/CollectionNavigatorBase.cs
```diff
@@ -145,18 +120,7 @@ public int GetNextMatchingItem (int currentIndex, char keyStruck)
...
-    /// <returns></returns>
-    public static bool IsCompatibleKey (Key a)
-    {
-        Rune rune = a.AsRune;
 
-        return rune != default (Rune) && !Rune.IsControl (rune);
-    }
 
     /// <summary>
     ///     Invoked when the <see cref="SearchString"/> changes. Useful for debugging. Invokes the
```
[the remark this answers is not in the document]
tznind: Done

Terminal.Gui/Text/IListCollectionNavigator.cs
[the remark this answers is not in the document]
tznind: I have moved to new folder `/Views/CollectionNavigation` let me know if that does not suit.

Terminal.Gui/Views/FileDialogCollectionNavigator.cs
tznind: We must deal in `object` for now.  This PR is just refactor of existing code to allow users to better modify collection navigation behaviour (add interfaces, move the most user servicable bit from a method to a 'component' subclass).

There are 4 views all using the same subsystem

|     View |               What is the `object`  |   Navigator subclass  |
|:-------------:|:------------------------------------:|:------------------------------------:|
| `TreeView<T>` |  `T`           | `CollectionNavigator` |
| `ListView` |  `object`                  | `CollectionNavigator` |
| `FileDialog` | `string`    | `FileDialogCollectionNavigator` |
| `TableView` | `string` or `object` * | `TableCollectionNavigator` |

\* Depending on if there is a `ColumnStyle` with `RepresentationGetter`

Terminal.Gui/Text/IListCollectionNavigator.cs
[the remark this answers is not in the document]
tig: I was also referring to the api docs which say it is specific to ListView.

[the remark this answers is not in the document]
tznind: Got it! Good point - and actual this interface is used by both Listview and TreeView - only TableView uses the base interface

I will update the xmldoc to indicate correct users

tznind:
Keybinding priority fix is now in place for all TreeView, TableView and ListView.

Terminal.Gui/Views/CollectionNavigation/DefaultCollectionNavigatorMatcher.cs
```diff
@@ -0,0 +1,30 @@
+﻿#nullable enable
+
+namespace Terminal.Gui;
+
+/// <summary>
+///     Default implementation of <see cref="ICollectionNavigatorMatcher"/>, performs
+///     case-insensitive (see <see cref="Comparer"/>) matching of items based on
+///     <see cref="object.ToString()"/>.
+/// </summary>
+public class DefaultCollectionNavigatorMatcher : ICollectionNavigatorMatcher
```
danielmarbach, line 10: It is interesting that this entire implementation could actually be done as default on the interface. Then the interface could declare the comparer as only getter default property and the class could have an init only member for the comparer. If my mental compilation is correct this would work

Terminal.Gui/Views/CollectionNavigation/ICollectionNavigator.cs
```diff
@@ -0,0 +1,49 @@
...
+    ///     Gets the current search string. This includes the set of keystrokes that have been pressed since the last
+    ///     unsuccessful match or after <see cref="TypingDelay"/>) milliseconds. Useful for debugging.
+    /// </summary>
+    string SearchString { get; }
+
+    /// <summary>
+    ///     Class responsible for deciding whether given entries in the collection match
+    ///     the search term the user is typing.
+    /// </summary>
+    ICollectionNavigatorMatcher Matcher { get; set; }
```
danielmarbach, line 36: Are these things mutable by default because you want to allow replacing and setting the matcher at any time or simplify mocking? From an overall component design perspective I guess it might be better to make the interface member not mutable and enforce it to be set when the class is newed up with init only members. It rarely ever makes sense to replace it after object creation

Terminal.Gui/Views/ListView.cs
```diff
@@ -807,32 +807,39 @@ public bool OnOpenSelectedItem ()
...
+            if (keys.Contains (key))
             {
                 return false;
             }
 
         }
 
+        // If the key was bound to key command, let normal KeyDown processing happen. This enables overriding the default handling.
+        // See: https://github.com/gui-cs/Terminal.Gui/issues/3950#issuecomment-2807350939
+        if (KeyBindings.TryGet (key, out _))
```
danielmarbach, line 834: Is there any potential overlap with `Command.Select` and `Command.Down` used above vs the `AllowsMarking` flag to be false that could lead to weird behavior here? (Sorry I'm not knee-deep into this code so this might be a stupid question)

Terminal.Gui/Views/CollectionNavigation/ICollectionNavigator.cs
```diff
@@ -0,0 +1,49 @@
...
+///     Navigates a collection of items using keystrokes. The keystrokes are used to build a search string. The
+///     <see cref="SearchString"/> is used to find the next item in the collection that matches the search string when
+///     <see cref="GetNextMatchingItem(int, char)"/> is called.
+///     <para>
+///         If the user types keystrokes that can't be found in the collection, the search string is cleared and the next
+///         item is found that starts with the last keystroke.
+///     </para>
+///     <para>If the user pauses keystrokes for a short time (see <see cref="TypingDelay"/>), the search string is cleared.</para>
+/// </summary>
+public interface ICollectionNavigator
```
danielmarbach, line 15: Given you have a public base class hierarchy what is the added value of those interfaces?

Terminal.Gui/Views/FileDialogCollectionNavigator.cs
```diff
@@ -0,0 +1,21 @@
...
+                                                              tableView.SelectedColumn,
+                                                              fileDialog.State?.Children [idx]
+                                                             );
+
+        if (val is null)
+        {
+            return string.Empty;
+        }
+
+        return val.ToString ().Trim ('.');
```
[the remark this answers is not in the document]
tznind: Not sure if it has a name, it's the formatting settings we have configured on project and enforced bu ReSharper CodeCleanup.

@tig could probably tell you more

Terminal.Gui/Views/CollectionNavigation/ICollectionNavigator.cs
tznind: The interface allows for the consuming views (`ListView` etc) to reference an interface rather than a concrete abstract base.

I prefer this as preference even though it means more classes - in general I just love interfaces for their flexibility later on.

So user can:

- Create custom implementation of matcher only 
- Create custom implementation of entire collection navigator 

This ensures maximum flexibility - user is not tied to the abstract base if they don't want to.

Terminal.Gui/Views/ListView.cs
tznind: Hmn good point, it actually looks like the entire `AllowsMarking` bit can go.

All it is doing is looking for specific keybindings to return `false` then we have a blanket if below it!

Good catch

tznind: Fixed in [0e29e04](https://github.com/gui-cs/Terminal.Gui/pull/4029/commits/0e29e04c44a3725d8bd83fd5026f1a92e337944d)

Terminal.Gui/Views/CollectionNavigation/ICollectionNavigator.cs
tznind: In general (and certainly for Views) we always want a default contructor and no `required` properties.

This enables support for setting stuff more freely e.g. in [Terminal Gui Desinger](https://github.com/gui-cs/TerminalGuiDesigner)

Frequently it will generate code like:

```csharp
private Terminal.Gui.MenuItem goToLineMenuItem;

private void InitializeComponent() {
    this.menuBar = new Terminal.Gui.MenuBar();
    this.Width = Dim.Fill(0);
    this.Height = Dim.Fill(0);
    this.X = 0;
    this.Y = 0;
    this.Visible = true;
    this.Arrangement = Terminal.Gui.ViewArrangement.Overlapped;
    this.CanFocus = true;
    this.ShadowStyle = Terminal.Gui.ShadowStyle.None;
    this.Modal = false;
    this.TextAlignment = Terminal.Gui.Alignment.Start;
    this.Title = "";
    this.menuBar.Width = Dim.Fill(0);
    this.menuBar.Height = 1;
    this.menuBar.X = 0;
    this.menuBar.Y = 0;
    this.menuBar.Visible = true;
    this.menuBar.Arrangement = Terminal.Gui.ViewArrangement.Fixed;
    this.menuBar.CanFocus = false;
```

So while it might be the most principled approach to make it immutable I think practicality wins out.  But happy to discuss the issue.

Any thoughts @tig  / @BDisp ?

Terminal.Gui/Views/FileDialogCollectionNavigator.cs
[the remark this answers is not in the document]
BDisp: I think that's because mono library uses these code style and Miguel want to keep it as is.

Terminal.Gui/Views/CollectionNavigation/ICollectionNavigator.cs
tznind: The most common use case I envisage for external users is:

```csharp
var lv = new ListView();
lv.KeystrokeNavigator.Matcher = new MyCustomMatcher();
```

tznind:
Turns out there was a whole bit of UICatalog that was creating a collection navigator for TableView but that already works out the box.  So I have removed the bespoke code in [d142ff1](https://github.com/gui-cs/Terminal.Gui/pull/4029/commits/d142ff1cd128f933a1e4b4f1fd1c545c96e7a46a)

I also made all the class implementations internal as the interfaces are what users should work with and ideally focus only on the matcher interface.

tig:
@tznind can you please resolve conflicts?

Terminal.Gui/Views/CollectionNavigation/CollectionNavigatorBase.cs
```diff
@@ -61,21 +37,21 @@ public int GetNextMatchingItem (int currentIndex, char keyStruck)
             var candidateState = "";
             var elapsedTime = DateTime.Now - _lastKeystroke;
 
-            Logging.Trace($"CollectionNavigator began processing '{keyStruck}', it has been {elapsedTime} since last keystroke");
+            Logging.Trace ($"CollectionNavigator began processing '{keyStruck}', it has been {elapsedTime} since last keystroke");
```
tig, line 40: Change these to `Debug`, please. IMO `Trace` should only be used for very-core, low-level things. No?

Tests/UnitTests/Views/ListViewTests.cs
```diff
@@ -1157,4 +1158,98 @@ void Lw_CollectionChanged (object sender, NotifyCollectionChangedEventArgs e)
             }
         }
     }
+    [Fact]
```
tig, line 1161: I believe these tests have no dependency on `Application` or `ConfigurationManager` and thus can be moved to parallizable? If I'm right, please do so.

Tests/UnitTests/Views/TableViewTests.cs
```diff
@@ -3387,6 +3387,48 @@ public void TableCollectionNavigator_FullRowSelect_True_False (bool fullRowSelec
         Assert.Equal (expectedRow, tableView.CollectionNavigator.GetNextMatchingItem (0, "3".ToCharArray () [0]));
     }
 
+
```
tig, line 3390: If you are not testing HotKeyBindings, and just KeyBindings, I don' think there's a need to depend on Application.Top in this case. 

It's good practice when writing unit tests to strive to simplify dependencies as much as possible and try to make them work as a parallelizable test. I do recognize that KeyBindings tends to put a wrench in that sometimes though.

Tests/UnitTests/Views/TreeViewTests.cs
tig, line 1: Same comment as above: Can the dependency on Application.Top be removed so this can move to a parallelizable test?

tig (changes requested):

Terminal.Gui/Views/FileDialogCollectionNavigator.cs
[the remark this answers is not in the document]
danielmarbach: The space between `ToString` and `()` is not a style I have seen before

Terminal.Gui/Views/CollectionNavigation/CollectionNavigatorBase.cs
tznind: Sure, I've changed to Debug in [ebecbc8](https://github.com/gui-cs/Terminal.Gui/pull/4029/commits/ebecbc8dfb1021273a6ea10ca8a7f0d077d00f9e)

Tests/UnitTests/Views/TableViewTests.cs
tig: Yes, like that. 

- Strive to build tests that don't depend on `Application` or `ConfigurationManager`.
- Put those tests that don't depend on `Application` or `ConfigurationManager` in `UnitTestsParallezable`.

The following have implicit/non-obvious dependencies on `Application`:

- `HotKeyBindings`
- Complex Navigation - Anytime, `Application.Navigation.GetFocused ()` may come into play.

Of course, once the `Applicationv2` stuff is fully done we can leverage `ChangeInstance()` to make even tests that depend on `Application` parallelizable! (Thank you!!!)

Tests/UnitTests/Views/TableViewTests.cs
```diff
@@ -105,6 +101,8 @@ public void CellEventsBackgroundFill ()
             style.ColorGetter = e => { return scheme; };
         }
 
+        // Required or won't draw properly
+        Application.Driver.Clip = new Region (tv.Frame);
```
BDisp, line 105: Staying in the `UnitTests` project there is no problem. Only in the Parallelizable project that would be a problem.

tznind: It is problem, if you remove it then test fails.  This commit I removed the Application.Top bit 

Alone removing top causes issues with repeated draws.  This line fixes that

BDisp: I think you didn't understand. What @tig said was "`Can the dependency on Application.Top be removed so this can move to a parallelizable test?`". If a test can't be run without the dependency the leave as is, otherwise remove the dependencies and move to a parallelizable test. By the fact it still need the `Application` it can't be removed, So, it can't, don't do anything 😃

tznind: Ah OK, thanks for explaining,  I will see which can be moved and which cannot

tznind:
I've removed the references to Application

BDisp:
> I've removed the references to Application

Fine, but @tig said if you could get rid of `Application` and anything that use it, like drivers, etc, to move the unit tests to the `UnitTestsParallelizable` project. If the file doesn't yet exist create it in the same folder and move all the tests in the `UnitTests` project to the `UnitTestsParallelizable` project.

tznind:
> move the unit tests to the `UnitTestsParallelizable` project. If the file doesn't yet exist create it in the same folder

Done in [639b1d1](https://github.com/gui-cs/Terminal.Gui/pull/4029/commits/639b1d1f907eb7b51cffa6e887373d67f3754c60)

I only moved the new ones - any moving or filtering of the existing tests should happen as part of a seperate PR as its nothing to do with collection navigation.

BDisp:
I see, I just remembered because @tig asked you to take advantage of moving them in this PR. But if it involves a lot of unit tests it justifies doing them in a new PR.

tig (approved):

Issue #2871, opened 2023-10-02
ConfigurationManager: `Applied` and `Updated` events are clumsy


tig:
Currently `ConfigurationManager` uses an event pattern so that classes that want to know if the config was updated (or been applied) can update their state. This lets apps update in-real-time whenever config changes.

However, this pattern means classes like `LineCanvas` have to subscribe to an event:

```cs
		public LineCanvas()
		{
			ConfigurationManager.Applied += ConfigurationManager_Applied;
		}
```

And THIS means they also have to implement `IDisposable` to correctly unsubscribe. This is what led to a big memory leak reported in #2870. IMO, if the disposable pattern can be avoided, it should be as it relies on developers remembering to do things.

I propose that in addition to having these events classes can implement an `ISupportConfigChanges` interface:

```cs
/// <summary>
/// <para>
/// Classes should implement this interface on any class that should be notified when the configuration changes.
/// This is an alternative to using the <see cref="ConfigurationManager.Updated"/> and
/// <see cref="ConfigurationManager.Applied"/> events.
/// <para>
/// When the configuration has changed, the <see cref="ConfigurationManager"/> will call the <see cref="OnUpdated"/>
/// method. When the configuration is being applied, the <see cref="OnApplied"/> method will be called.
/// 
/// </para>
/// </summary>
public interface ISupportConfigChanges {

	/// <summary>
	/// Called when the configuration has been updated.
	/// </summary>
	public void OnUpdated ();

	/// <summary>
	/// Called when the configuration is being applied.
	/// </summary>
	public void OnApplied ();
}
```

`ConfigurationManager.OnApplied()` currently looks like this:

```cs
	/// <summary>
	/// Called when an updated configuration has been applied to the  
	/// application. Fires the <see cref="Applied"/> event.
	/// </summary>
	public static void OnApplied ()
	{
		Debug.WriteLine ($"ConfigurationManager.OnApplied()");
		Applied?.Invoke (null, new ConfigurationManagerEventArgs ());
	}
```

We'll use reflection after the call to `Applied?.Invoke()` to find all instances of classes implementing `ISupportConfigChanges` and call the `OnApplied()` method on them.

I may be missing a better pattern for this, so i'd love to hear from others on this concept.

tig:
My plan for fixing this in #4062 is, roughly:

- Make `ConfigProperty` inherit from `INotifyPropertyChanged`.
[...]

PR #4062, opened 2025-04-28
Fixes #4057 - MASSIVE! Fully implements `ColorScheme`->`Scheme` + `VisualRole` + `Colors.`->`SchemeManager.` 

### LINKED ISSUE #3753: `HighlightStyle.Hover` etc... needs a different implementation
Renamed this issue from "`HighlightOnPress` and/or `WantContinousButtonPressed` has issues with nested subviews." because there are deeper issues. 

The current impl changes the View's Scheme. Instead Highlight should be a `VisualRole`. 

I am NOT fixing this in #4062 , but instead disabling Hover all together until it can be done right in a PR for this Issue.

## Orignial Post

In #3751 I refactored and fixed a bunch of issues around mouse and highlight.

I'm not going to fix this there, so I'm creating an issue for it:

In the new `Mouse` scenario below each view in the `Enter/Leave Demo` is the same class... nested subviews:

- The root
  - One in the root's padding: inPadding
  - Subview: Sub1 
  - Subview: Sub2

Right now `WantContinousButtonPressed` is true for ALL 4. But `Highlight on Press` is only enabled for the root.

Note how sub1 and sub2 are showing highlight when I press in the Viewport of the root? that should not be happening.

PQPZSes 1

### LINKED ISSUE #4016: Add events for `GetNormalColor` etc....

### LINKED ISSUE #4021: `ConfigurationManager` - It's a bug for any `SomeProperty = DefaultSomeProperty` if `SomeProperty` isn't nullable
e.g. `BorderStyle` should be nullable (any property that has a default value should be nullable).

 What if the user set the BorderStyle to something else?

```cs

    private void OnConfigurationManagerApplied (object? sender, ConfigurationManagerEventArgs e)
    {
        // BUGBUG: We should NOT do this. What if the user set the BorderStyle to something else?
        // BUGBUG: BorderStyle should be nullable (any property that has a default value should be nullable)!
        BorderStyle = DefaultBorderStyle;
    }
```

(more later)

### LINKED ISSUE #4057: Rename `ColorScheme` to `Scheme` and Migrate `Colors` to `ConfigurationManager.Themes`
As discussed in:

- #457
- #4058
- #2381

The names `ColorScheme` and `Colors` in Terminal.Gui are misleading because they imply color-only functionality, yet `Attributes` within `ColorScheme` will soon support non-color styles (e.g., Bold, Italic, per #4058). The static `Colors` dictionary (`Terminal.Gui.Colors.ColorSchemes`) causes thread-safety issues and lacks type safety, making it unsuitable for modern theme management. I propose:

- Rename `ColorScheme` to `Scheme` to reflect its role in defining both color and non-color visual styles.

- Remove `Colors` and migrate its functionality to `ConfigurationManager.Themes`, leveraging Terminal.Gui’s JSON-driven theme management system.

This issue focuses on these naming and structural changes, keeping #457 as the master issue for the broader `ColorSchemes` redesign, including per-view defaults and terminal settings (#2381). Backward compatibility for `Colors` is not required, simplifying the migration.

## Current State

- `ColorScheme` is a class defining `Attributes` (Normal, HotNormal, Focus, HotFocus, Disabled) used by views for styling.

- `Colors` is a static dictionary with string keys (e.g., "Base", "Dialog", "Menu", "Error"), which is prone to threading issues and not configurable by users.

- `ConfigurationManager.Themes` manages themes (e.g., “Light”, “Dark”) as a non-static dictionary of `ColorScheme` instances, loaded from JSON configs (e.g., `.tui/config.json`, embedded `config.json`). It supports user-defined themes, runtime overrides (e.g., `ConfigurationManager.RuntimeConfig = """{ "Theme": "Light" }"""`), and a robust API for theme management (e.g., adding, retrieving, applying themes).

- Current `ConfigurationManager.Themes` JSON structure:

```json
{
  "Dark": {
    "ColorSchemes": [
      {
        "TopLevel": {
          "Normal": {
            // Color and attribute definitions
          }
...
```

## Proposed Changes

- **Rename `ColorScheme` to `Scheme`**:

  - Update the class name to `Scheme` for clarity and brevity.

  - Refactor all references in the codebase (e.g., `View.ColorScheme`, `ConfigurationManager.Themes`, `ThemeManager`).

  - Update the `tui-config-schema.json` to use `Schemes` instead of `ColorSchemes` in theme definitions (https://gui-cs.github.io/Terminal.GuiV2Docs/schemas/tui-config-schema.json).

- **Remove `Colors` and Migrate to `ConfigurationManager.Themes`**:

  - Eliminate the static `Terminal.Gui.Colors` class and its `ColorSchemes` dictionary.

  - Migrate default schemes (Base, Dialog, Menu, Error) to `ConfigurationManager.Themes`, storing them as `Scheme` instances in the embedded `config.json` or user `.tui` configs.

  - Update `ThemeManager` to store and manage `Scheme` instances, ensuring seamless integration with the existing theme management API.

  - Example updated JSON structure:

```json
{
  "Dark": {
    "Schemes": [
      {
        "TopLevel": {
          "Normal": {
            // Color and attribute definitions
          }
...
```

- **Support Theme Inheritance**:

  - Ensure `ConfigurationManager.Themes` supports theme inheritance, where views default to their `SuperView`’s theme if unspecified, per #457’s hierarchy model.

- **Support Future Terminal Settings**:

  - Design `ConfigurationManager.Themes` to integrate with user terminal colors (e.g., `$PSStyle` or `$TERM` settings) when implemented in #2381, though this is out of scope here.

- **Documentation and Testing**:

  - Update documentation, samples (e.g., UICatalog, Example.cs), and API docs to reflect `Scheme` and `ConfigurationManager.Themes`.

  - Fix unit tests to remove `Colors` dependencies and validate `ThemeManager` integration.

## Benefits

- Clarifies that `Scheme` supports both colors and non-color attributes (aligning with #4058).

- Resolves threading issues by replacing static `Colors` with `ConfigurationManager`’s non-static, JSON-driven system.

- Enables user-defined themes via `.tui` configs, enhancing flexibility and customization.

- Supports #457’s goal of a consistent, scalable styling system and prepares for #2381’s terminal settings integration.

## Tasks

- [ ] Refactor `ColorScheme` to `Scheme` across the codebase and JSON schema.

- [ ] Remove `Terminal.Gui.Colors` and migrate its schemes to `ConfigurationManager.Themes`.

- [ ] Update documentation, samples, and UICatalog to use `Scheme` and `ConfigurationManager.Themes`.

- [ ] Fix unit tests to eliminate `Colors` dependencies and validate `ThemeManager` integration.

## Related Issues

- #4058: Add non-color attributes to `Attributes` (supporting styles like Bold, Italic).

- #457: Master issue for `ColorSchemes` redesign, including per-view defaults and terminal settings.

- #2381: Honor user terminal settings (mappings to `Scheme` attributes).

- #48: Truecolors support (relevant for advanced color handling).

## Open Questions

- Are there edge cases in `ThemeManager` that need addressing for `Scheme` migration?

- Should `ConfigurationManager.Themes` include additional validation for user-defined themes?

- Should the JSON structure for `Schemes` be flattened or simplified (e.g., remove nested array)?

Community feedback on the rename, migration approach, and `ThemeManager` integration is welcome!

### LINKED ISSUE #4059: Make Theme Manager a better Singleton for Better Testability
(Note: This applies to `ConfigurationManager` as well, but I want to start with Themes)

## Problem

The current `ThemeManager` and `Themes` structures use static classes and static fields.  
This design makes unit testing difficult because:

- Tests that interact with Themes share global state.
- Tests cannot run safely in parallel.
- It’s difficult to inject mock themes for isolation.

## Proposal

Refactor `ThemeManager` to:

- Provide a `ThemeManager.Instance` static accessor.
- Back `Instance` with an interface (`IThemes`).
- Default `Instance` to a normal implementation (`Themes`).
- Allow tests to override `Instance` with a mock/fake.

## Scope

- Refactor `ThemeManager` and `Themes` to use Singleton pattern via `Instance`.
- Introduce `IThemes` interface.
- Update usages of `Themes.Default` to `ThemeManager.Instance.Default` across the library.
- No functional changes to loading ColorSchemes yet.
- No changes to the config file, `Colors.ColorSchemes`, or deeper theming structures yet.

## Benefits

- Unit tests can inject fake Themes.
- Tests can run safely in parallel.
- Prepares the codebase for cleaner future refactoring (e.g., runtime dynamic themes, multi-theme support).

### LINKED ISSUE #4069: TextView does not allow changing of read-only text background
Non-standard bug report format because I'm not _entirely_ sure this is a bug, but would be surprised if it isn't. I've already found the offending code snippet, but want to check this is actually a bug before making a PR or anything.

In `TextView`:

https://github.com/gui-cs/Terminal.Gui/blob/b46b7781f10c13b4860242f29d2105919a118927/Terminal.Gui/Views/TextView.cs#L3997-L4009

This is in the `OnDrawReadOnlyColor()` method, which fires when drawing read-only text. This triggers the `DrawReadOnlyColor` event, which allows you to modify the attributes of drawn `Cell`s to change the color.

The problem is that this leaves no way to manually set both the foreground and background colors at the same time: If they are different, the foreground will be kept, but the background will be set to correspond with the `TextView`'s `ColorScheme`. If they are the same, you can change the background color, but then of course the text is illegible.

The other color selection methods simply check if the attribute was null, set it to a default if it was, and use it as-is otherwise. I'm not sure why that isn't the case here as well, and it doesn't feel like this is intentional, as it means the only way to change the read only text color is to derive the class and override the function, which isn't necessary for any other color choices.

### LINKED ISSUE #4072: `ConfigurationManager` - `DeepMemberwiseClone` does not do what it says
I was young and stupid when I wrote it. 

It doesn't actually clone byref values.

It mostly worked when used in apps, but this is actually the cause of some of the weird unit tests failures.

I'm working on a fix in #4062

### LINKED ISSUE #4073: `ConfigurationManager` - It's stupid that it tries to be scoped by `Application.Init`
CM is per-process. All the logic in `Application.Init` should just go away. 

I'm working on a fix for this in 

- #4062 

This will simplify things and make them deterministic.

As part of this I'm enabling module-initailziation and creating a cache of the hard-coded `[SerializableConfigurationProperty]` values to prove correctness and enable a "CM is totally disabled" mode.

### LINKED ISSUE #4084: Rename `TextStyle` `Style`
In #4071, @Error-String-Expected-Got-Nil did an excellent job!

I'm not happy with the property of `Attribute` being named `TextStyle` though. It's a minor nit, but it's needlessly verbose and should just be `Style`. Likewise, the `enum` should be `Style`.

I will fix this in #4062 unless someone has an objection.

### LINKED ISSUE #4087: Refactor `DimAuto.cs` for better readability
Currently cumbersome. I attempted a refactor like this, but it broke something. Leaving it as is for now, but opening this issue so I can save off the work and address later.

```cs
#nullable enable
using System.Diagnostics;

namespace Terminal.Gui;

/// <summary>
///     Represents a dimension that automatically sizes the view to fit all the view's Content, SubViews, and/or Text.
/// </summary>
/// <remarks>
///     <para>
///         See <see cref="DimAutoStyle"/>.
///     </para>
///     <para>
///         This is a low-level API that is typically used internally by the layout system. Use the various static
///         methods on the <see cref="Dim"/> class to create <see cref="Dim"/> objects instead.
///     </para>
/// </remarks>
/// <param name="MaximumContentDim">The maximum dimension the View's ContentSize will be fit to.</param>
/// <param name="MinimumContentDim">The minimum dimension the View's ContentSize will be constrained to.</param>
/// <param name="Style">The <see cref="DimAutoStyle"/> of the <see cref="DimAuto"/>.</param>
public record DimAuto (Dim? MaximumContentDim, Dim? MinimumContentDim, DimAutoStyle Style) : Dim
{
    /// <inheritdoc/>
    public override string ToString () { return $"Auto({Style},{MinimumContentDim},{MaximumContentDim})"; }

    /// <inheritdoc />
    internal override int GetAnchor (int size) => 0;

    internal override int Calculate(int location, int superviewContentSize, View us, Dimension dimension)
    {
        var textSize = 0;

        int autoMin = MinimumContentDim?.GetAnchor(superviewContentSize) ?? 0;
        int screenX4 = dimension == Dimension.Width ? Application.Screen.Width * 4 : Application.Screen.Height * 4;
        int autoMax = MaximumContentDim?.GetAnchor(superviewContentSize) ?? screenX4;

        Debug.WriteLineIf(autoMin > autoMax, "MinimumContentDim must be less than or equal to MaximumContentDim.");

        if (Style.FastHasFlags(DimAutoStyle.Text))
        {
            if (dimension == Dimension.Width)
            {
                if (us.TextFormatter.ConstrainToWidth is null)
                {
                    us.TextFormatter.ConstrainToSize = us.TextFormatter.FormatAndGetSize(new(int.Min(autoMax, screenX4), screenX4));
                }

                textSize = us.TextFormatter.ConstrainToWidth!.Value;
            }
            else
            {
                if (us.TextFormatter.ConstrainToHeight is null)
                {
                    textSize = us.TextFormatter.FormatAndGetSize(new(us.TextFormatter.ConstrainToWidth ?? screenX4, int.Min(autoMax, screenX4))).Height;
                    us.TextFormatter.ConstrainToHeight = textSize;
                }
                else
                {
                    textSize = us.TextFormatter.ConstrainToHeight.Value;
                }
            }
        }

        int maxCalculatedSize = CalculateMaxContentSize(us, dimension, superviewContentSize, textSize);

        // All sizes here are content-relative; ignoring adornments.
        // We take the largest of text and content.
        int max = int.Max(textSize, maxCalculatedSize);

        // And, if min: is set, it wins if larger
        max = int.Max(max, autoMin);

        // And, if max: is set, it wins if smaller
        max = int.Min(max, autoMax);

        Thickness thickness = us.GetAdornmentsThickness();

        int adornmentThickness = dimension switch
        {
            Dimension.Width => thickness.Horizontal,
            Dimension.Height => thickness.Vertical,
            Dimension.None => 0,
            _ => throw new ArgumentOutOfRangeException(nameof(dimension), dimension, null)
        };

        max += adornmentThickness;

        return max;
    }

    private int CalculateMaxContentSize(View us, Dimension dimension, int superviewContentSize, int textSize)
    {
        int maxCalculatedSize = textSize;

        if (Style.FastHasFlags(DimAutoStyle.Content))
        {
            if (us is { ContentSizeTracksViewport: false, InternalSubViews.Count: 0 })
            {
                maxCalculatedSize = dimension == Dimension.Width ? us.GetContentSize().Width : us.GetContentSize().Height;
            }
            else
            {
                List<View> includedSubViews = us.InternalSubViews.ToList();

                List<View> notDependentSubViews = GetNotDependentSubViews(includedSubViews, dimension);

                foreach (View notDependentSubView in notDependentSubViews)
                {
                    notDependentSubView.SetRelativeLayout(us.GetContentSize());
                }

                foreach (View v in notDependentSubViews)
                {
                    int size = dimension == Dimension.Width
                        ? v.X.GetAnchor(0) + v.Width!.Calculate(0, superviewContentSize, v, dimension)
                        : v.Y!.GetAnchor(0) + v.Height!.Calculate(0, superviewContentSize, v, dimension);

                    if (size > maxCalculatedSize)
                    {
                        maxCalculatedSize = size;
                    }
                }

                // Additional logic for other subview types (e.g., centered, anchored, etc.) can be added here
            }
        }

        return maxCalculatedSize;
    }

    private List<View> GetNotDependentSubViews(List<View> includedSubViews, Dimension dimension)
    {
        if (dimension == Dimension.Width)
        {
            return includedSubViews.Where(
                v => v.Width is { }
                     && (v.X is PosAbsolute or PosFunc
                         || v.Width is DimAuto or DimAbsolute or DimFunc)
                     && !v.X.Has<PosAnchorEnd>(out _)
                     && !v.X.Has<PosAlign>(out _)
                     && !v.X.Has<PosCenter>(out _)
                     && !v.Width.Has<DimFill>(out _)
                     && !v.Width.Has<DimPercent>(out _)
            ).ToList();
        }
        else
        {
            return includedSubViews.Where(
                v => v.Height is { }
                     && (v.Y is PosAbsolute or PosFunc
                         || v.Height is DimAuto or DimAbsolute or DimFunc)
                     && !v.Y.Has<PosAnchorEnd>(out _)
                     && !v.Y.Has<PosAlign>(out _)
                     && !v.Y.Has<PosCenter>(out _)
                     && !v.Height.Has<DimFill>(out _)
                     && !v.Height.Has<DimPercent>(out _)
            ).ToList();
        }
    }
}

```

### LINKED ISSUE #4089: Add `View.GetSizeRequiredForSubViews()`
In https://github.com/gui-cs/Terminal.Gui/issues/4070 @Error-String-Expected-Got-Nil  semi-requested an API to get the maximum dims of all content.

We actually do need such an API to make enabling scrolling easier.

Prototype impl:

```cs
 /// <summary>
 ///     Gets the size required for all of the View's SubViews.
 /// </summary>
 /// <returns></returns>
 public Size GetSizeRequiredForSubViews ()
 {
     int maxWidth = 0;
     int maxHeight = 0;

     // If ContentSizeTracksViewport is false and there are no subviews, use the explicitly set ContentSize
     if (!ContentSizeTracksViewport && InternalSubViews.Count == 0)
     {
         return GetContentSize ();
     }

     // Iterate through all subviews to calculate the maximum width and height
     foreach (View subView in InternalSubViews)
     {
         // Calculate the width
         if (subView.Width is { })
         {
             int subViewWidth = subView.X.GetAnchor (0) + subView.Width.Calculate (0, GetContentSize ().Width, subView, Dimension.Width);
             maxWidth = Math.Max (maxWidth, subViewWidth);
         }

         // Calculate the height
         if (subView.Height is { })
         {
             int subViewHeight = subView.Y.GetAnchor (0) + subView.Height.Calculate (0, GetContentSize ().Height, subView, Dimension.Height);
             maxHeight = Math.Max (maxHeight, subViewHeight);
         }
     }

     // Return the calculated maximum content size
     return new Size (maxWidth, maxHeight);
 }

```

tig:
***HUGE PR***

This PR introduces a significant refactoring of Terminal.Gui's Configuration Management system, addressing issue #4057 and a bunch of other related issues. The changes include:

* **Significant* refactor and upgrade of `ConfigurationManager`
  * Simplifies and cleans up naming (e.g. `SerializableConfigruationProperty` -> `ConfigProperty`).
  * Uses `[ModuleInitalizer`] to ensure that CM caches the original values of all `[ConfigProperty]`s and provides APIs for resetting those values back (`ConfigurationManager.ResetToHardCodedValues`).
  * Removes all lame/complex code that tried to tie CM to `Application.Init`. Instead CM is simply process-wide. 
  * Makes CM **OFF BY DEFAULT**. Apps must call `ConfigurationManager.Enable` to enable it. 
* Renames `ColorScheme` to `Scheme`. Enables `SchemeManager.GetCurrentSchemes()` as the API that replaces the old `ColorSchemes` static class.
* Introduces `VisualRole` to replace string-based `ColorSchemes[???]` identifiers.
* ***Changes `View.Scheme` API signficantly***
  * `View.ColorScheme` -> `View.Scheme` 
  * NEW: `View.HasScheme` to indicate whether a Scheme was explicitly set or not.
  * NEW: `Get/SetScheme` (used by `Scheme` get/set). Enable...
  * NEW: `GettingScheme`/`OnGettingScheme` cancellable event
  * NEW: `SettingScheme`/`OnSettingScheme` cancellable event
  * GONE: `Scheme` is no longer `virtual` (the above addressses).
* `HighlightStyle.Hover` is **DISABLED** in this PR. to be fixed in #3753.

Much of the vast churn in this PR was done to finally address a slew of non-deterministic unit test problems having to do with static state. There are still some non-CM related issues buried in there, but I've killed most of them here. Note: Any unit test that uses `Configuration.Enable` **MUST NOT** be in `Parallelizable`. However, if you don't call `Enable` or use `Application` it's highly likely Parallelizable will work.
  
***IMPORTANT***: If you rely on `View.Scheme` being `null` ... STOP. Instead use `HasScheme`.

## Fixes

- Fixes #4057
- Fixes #4059 
- Fixes #4072
- Fixes #4073
- Fixes #4016
- Fixes #4084
- Partially Addresses #2871
- Fixes #4021
- Fixes #4089
- Fixes #3753
- Fixes #4069
- Fixes #4087

## Proposed Changes/Todos

- [x] Rename `ColorScheme` to `Scheme`
- [x] Introduce `VisualRole` enum 
- [x] Refactor/Fix ConfigManager to be more robust and thread safe.
- [x] Refactor CM to make accessing Schemes work better.
- [x] Replace `View.GetXXXColor ()` with `View.GetAttributeForRole ()`
- [x] Add dozens of new unit tests, fix a bunch of old unit tests that were non-deterministic.
- [x] Ensure new `Scheme` and `Attribute` related APIs in `View` are fully baked and thought through.
- [x] Make `CM.Enable` more rational - Have it take `ConfigLocation`
- [x] Fix bad implementation of `View.Enable` in `View.Drawing`, built-in Views and Scenarios. 
- [x] Update `Text Styles` scenario to show more
- [x] Identify and fix why sometimes `Style` is misapplied/artifacts
- [x] On `Load()` `"Schemes": []` doesn't accept partial dict. Just overwrites dest.
- [x] Make sure above addresses #4021
- [x] Figure out how to reduce the in-memory size of the config (currently ~126kb!)

Now:
```
Start: Color size: 4 b
Start: Attribute size: 48 b
Start: Base Scheme size: 2870 b
Start: PropertyInfo size: 48 b
Start: ThemeScope (0) size: 56 b
Start: ThemeScope (1) size: 3020 b
Start: HardCoded Schemes (5) size: 2870 b
Start: Themes dictionary (1) size: 22 Kb
Enabled: Themes dictionary (1) size: 22 Kb
After Load: Themes dictionary (8) size: 47 Kb
Total Settings Size: 52 Kb
```

- [x] Fix dynamic theme re-loading in UI Catalog
- [x] Revamp `config.md` Deep Dive
- [x] Add a `Schemes` Scenario that shows all views and lets you play with applying different schemes
- [x] Refactor GetViewsUnderMouse to GetViewsUnderLocation etc... to work better
- [x] Enable the other new `VisualRole` (`Active`, `HotActive`, `Highlight`)
- [x] Ensure all API docs are up-to-date
- [x] Revamp `drawing.md` Deep Dive
- [x] Redo how `ShadowStyle` works to leverage the above. Shadows now look much better.
- [x] Re-do `Hover` with `Highlight`
- [x] Fix spurious unit test failures seen in CI/CD

## For Future PRs

- #2871
- #4086
- [ ] Validate AOT stuff still works (I may defer this).

## Pull Request checklist:


tig:
@tznind , @BDisp :For your reading pleasure (and to make it easier for you to give me feedback), here's the updated Configuration Manager deep dive:

# Configuration Management

Terminal.Gui provides persistent configuration settings via the [`ConfigurationManager`](~/api/Terminal.Gui.ConfigurationManager.yml) class.

## Configuration Lexicon and Taxonomy

| Term | Meaning |
|:-----|:--------|
| **AppSettings** | Application-specific settings stored in the application's resources. |
| **Apply** | Apply the configuration to the application; copies settings from configuration properties to corresponding `static` `[ConfigProperty]` properties. |
| **Attribute** | Defines concrete visual styling for a visual element (Foreground color, Background color, TextStyle). |
| **BackgroundColor** | A property of `Attribute` describing the background text color. |
| **Color** | Base terminal color (supports TrueColor and named values like White, Black, Cyan, etc.). |
| **ConfigProperty** | A property decorated with `[ConfigProperty]` that can be configured via the configuration system. |
| **Configuration** | A collection of settings defining application behavior and appearance. |
| **ConfigurationManager** | System that loads and manages application runtime settings from external sources. |
| **ForegroundColor** | A property of `Attribute` describing the foreground text color. |
| **Load** | Load configuration from given location(s), updating with new values. Loading doesn't apply settings automatically. |
| **Location** | Storage location for configuration (e.g., user's home directory, application directory). |
| **Reset** | Reset configuration to current values or hard-coded defaults. Does not load configuration. |
| **Scope** | Defines the context where configuration applies (Settings, Theme, or AppSettings). |
| **Scheme** | Maps `VisualRole` to `Attribute`, defining visual element appearance (color and style) based on semantic purpose. |
| **Settings** | Runtime options including both system settings and application-specific settings. |
| **Sources** | Set of locations where configuration can be stored (@Terminal.Gui.ConfigLocations enum). |
| **Style** | Property of `Attribute` for font-like hints (bold, italic, underline). |
| **Theme** | Named instance containing specific appearance settings. |
| **ThemeInheritance** | Mechanism where themes can inherit and override settings from other themes. |
| **Themes** | Collection of named Theme definitions bundling visual and layout settings. |
| **VisualRole** | Semantic role/purpose of a visual element (Normal, Focus, HotFocus, Active, Disabled, ReadOnly). |

# Fundamentals

The `ConfigurationManager` class provides a way to store and retrieve configuration settings for an application. The configuration is stored in JSON documents, which can be located in the user's home directory, the current working directory, in memory, or as a resource within the application's main assembly.

Settings are defined in JSON format, according to this schema: https://gui-cs.github.io/Terminal.GuiV2Docs/schemas/tui-config-schema.json.

Terminal.Gui library developers can define settings in code and set the default values in the Terminal.Gui assembly's resources (e.g. `Terminal.Gui.Resources.config.json`).

Terminal.Gui application developers can define settings in their apps' code and set the default values in their apps' resources (e.g. `Resources/config.json`) or by setting @Terminal.Gui.Application.RuntimeConfig to string containing JSON.

Users can change settings on a global or per-application basis by providing JSON formatted configuration files. The configuration files can be placed in at .tui folder in the user's home directory (e.g. `C:/Users/username/.tui`, or `/usr/username/.tui`) or the folder where the Terminal.Gui application was launched from (e.g. `./.tui`).

## CM is Disabled by Default

The `ConfigurationManager` class is disabled by default. To enable it, call @Terminal.Gui.ConfigurationManager.Enable() in your application's `Main` method.

```csharp
ConfigurationManager.Enable();
```

If `ConfigurationManager.Enable()` is not called (`ConfigurationManager.IsEnabled` is 'false'), all configuration settings are ignored and ConfigurationManager will effectively be a no-op. All `[ConfigurationProperty]` properties will initially be their hard-coded default values. Calling @Terminal.Gui.ConfigurationManager.Reset will reset all configuration properties back to their hard-coded default values.

Other than that, no other ConfigurationManager APIs will have any effect.

## Loading and Applying Configuration

The `ConfigurationManager` class provides a `Load` method that loads the configuration from the given location. The `Load` method does not apply the settings to the application; that happens when the `Apply` method is called.

When a configuration has been loaded, the @Terminal.Gui.ConfigurationManager.Apply method must be called to apply the settings to the application. This method uses reflection to find all static fields decorated with the `[ConfigurationProperty]` attribute and applies the settings to the corresponding properties.

```csharp
// Load the configuration from just the users home directory.
ConfigurationManager.Enable();
ConfigurationManager.Load(ConfigLocations.GlobalHome);
ConfigurationManager.Apply();
```

> [!IMPORTANT]
>  Configuration Settings Apply at the Process Level. 
> Configuration settings are applied at the process level, which means that they are applied to all applications that are part of the same process. This is due to the fact that configuration properties are defined as static fields, which are static for the process.

## Configuration Types and Scopes

Terminal.Gui supports three main configuration scopes:

### SettingsScope
System-level settings that affect Terminal.Gui behavior:
```csharp
[ConfigurationProperty (Scope = typeof (SettingsScope))]
public static int MaxSearchResults { get; set; } = 10000;
```

### ThemeScope
Visual appearance settings that can be themed:
```csharp
 [ConfigurationProperty (Scope = typeof (ThemeScope))]
 public new static LineStyle DefaultBorderStyle { get; set; } = LineStyle.Single;
```

### AppSettingsScope (default)
Application-specific settings:
```csharp
[ConfigurationProperty] // AppSettingsScope is default
public static string MyAppSetting { get; set; } = "default";
```

## Configuration Precedence

Configuration values are applied in the following order (later sources override earlier ones):

1. Hard-coded defaults (lowest precedence)
2. Terminal.Gui assembly defaults
3. Runtime configuration
4. App resources
5. App-specific user home directory
6. App-specific current directory
7. Global user home directory
8. Global current directory (highest precedence)

```mermaid
graph TD
    A[Hard-coded Defaults] --> B[Terminal.Gui Defaults]
    B --> C[Runtime Config]
    C --> D[App Resources]
    D --> E[App Home Directory]
    E --> F[App Current Directory]
    F --> G[Global Home Directory]
    G --> H[Global Current Directory]
```

## Configuration Events

The ConfigurationManager provides several events to track configuration changes:

```csharp
// Called after configuration is applied
ConfigurationManager.Applied += (sender, e) => {
    // Handle configuration changes
};

// Called when the active theme changes
ConfigurationManager.ThemeChanged += (sender, e) => {
    // Handle theme changes
};
```


## How Settings are Defined 

Application Developers define settings by decorating static properties with the `[ConfigurationProperty]` attribute.

```csharp
class MyApp
{
    [ConfigurationProperty]
    public static string MySetting { get; set; } = "Default Value";
}
```

Configuration Properties must be `public` or `internal` static properties.

The above example will define a configuration property in the `AppSettings` scope. The name of the property will be `MyApp.MySetting` and will appear in JSON as:

```json
{
    "AppSettings": {
      "MyApp.MySetting": "Default Value"
    }
}
```

`AppSettings` property names must be globally unique. To ensure this, the name of the AppSettings property is the name of the property prefixed with a period and the full name of the class that holds it. In the example above, the AppSettings property is named `MyApp.MySetting`.

Terminal.Gui library developers can use the `SettingsScope` and `ThemeScope` attributes to define settings and themes for the terminal.Gui library.

> [!IMPORTANT] App developers cannot define `SettingScope` or `ThemeScope` properties.

```csharp
    /// <summary>
    ///     Gets or sets whether <see cref="Button"/>s are shown with a shadow effect by default.
    /// </summary>
    [ConfigurationProperty (Scope = typeof (ThemeScope))]
    public static ShadowStyle DefaultShadow { get; set; } = ShadowStyle.None;
```

## Precedence

Settings are applied using the following precedence (higher precedence settings overwrite lower precedence settings):

1. Hard-coded default values in any static property decorated with the `[ConfigurationProperty]` attribute.

2. @Terminal.Gui.ConfigLocations.Default - Default settings in the Terminal.Gui assembly -- Lowest precedence.

3. @Terminal.Gui.ConfigLocations.Runtime - Settings stored in the @Terminal.Gui.ConfigurationManager.RuntimeConfig static property.

4. @Terminal.Gui.ConfigLocations.AppResources - App settings in app resources (`Resources/config.json`).

5. @Terminal.Gui.ConfigLocations.AppHome - App-specific settings in the users's home directory (`~/.tui/appname.config.json`). 

6. @Terminal.Gui.ConfigLocations.AppCurrent - App-specific settings in the directory the app was launched from (`./.tui/appname.config.json`).

7. @Terminal.Gui.ConfigLocations.GlobalHome - Global settings in the the user's home directory (`~/.tui/config.json`).

8. @Terminal.Gui.ConfigLocations.GlobalCurrent - Global settings in the directory the app was launched from (`./.tui/config.json`) --- Hightest precedence.


The [`ConfigurationManager`](~/api/Terminal.Gui.ConfigurationManager.yml) will look for configuration files in the `.tui` folder in the user's home directory (e.g. `C:/Users/username/.tui` or `/usr/username/.tui`), the folder where the Terminal.Gui application was launched from (e.g. `./.tui`), or as a resource within the Terminal.Gui application's main assembly.

Settings that will apply to all applications (global settings) reside in files named `config.json`. Settings that will apply to a specific Terminal.Gui application reside in files named `appname.config.json`, where *appname* is the assembly name of the application (e.g. `UICatalog.config.json`).

# Sample Code

The `UICatalog` application provides an example of how to use the [`ConfigurationManager`](~/api/Terminal.Gui.ConfigurationManager.yml) class to load and save configuration files. The `Configuration Editor` scenario provides an editor that allows users to edit the configuration files. UI Catalog also uses a file system watcher to detect changes to the configuration files to tell [`ConfigurationManager`](~/api/Terminal.Gui.ConfigurationManager.yml) to reload them; allowing users to change settings without having to restart the application.

# What Can Be Configured

The `ConfigurationManager` class provides the following features:

1) **Settings**. Settings are applied to the [`Application`](~/api/Terminal.Gui.Application.yml) class. Settings are accessed via the `Settings` property of [`ConfigurationManager`](~/api/Terminal.Gui.ConfigurationManager.yml). E.g. `Settings["Application.QuitKey"]`
2) **Themes**. Themes are a named collection of settings impacting how applications look. The default theme is named "Default". Two other built-in themes are provided: "Dark", and "Light". Additional themes can be defined in the configuration files. `Settings ["Themes"]` is a dictionary of theme names to theme settings.
3) **AppSettings**. Applications can use the [`ConfigurationManager`](~/api/Terminal.Gui.ConfigurationManager.yml) to store and retrieve application-specific settings.

Methods for discovering what can be configured are available in the `ConfigurationManager` class:

- Call @ConfigurationManager.GetConfigurationProperties()
- Search the source code for `[ConfigurationProperty]` 
- View `./Terminal.Gui/Resources/config.json`

For complete schema details and examples, refer to:
- Schema: https://gui-cs.github.io/Terminal.GuiV2Docs/schemas/tui-config-schema.json
- Default configuration: Terminal.Gui/Resources/config.json

## Themes

A Theme is a named collection of settings that impact the visual style of Terminal.Gui applications. The default theme is named "Default". The built-in configuration within the Terminal.Gui library defines two more themes: "Dark", and "Light". Additional themes can be defined in the configuration files. The JSON property `Theme` defines the name of the theme that will be used. If the theme is not found, the default theme will be used.

Themes support defining Schemes (a set of colors and styles that define the appearance of views) as well as various default settings for Views. Both the default color schemes and user-defined color schemes can be configured. See [Schemes](~/api/Terminal.Gui.Schemes.yml) for more information.

### Theme Configuration

Themes provide a way to bundle visual settings together. Themes can inherit from other themes:

```json
{
  "Themes": [
    "MyCustomTheme": {
      "BasedOn": "Default",
      "Colors": {
        "Normal": {
          "Foreground": "White",
          "Background": "Blue"
        }
      }
    }
  ]
}
```

### Glyphs

Themes support changing the standard set of glyphs used by views (e.g. the default indicator for [Button](~/api/Terminal.Gui.Button.yml)) and line drawing (e.g. [LineCanvas](~/api/Terminal.Gui.LineCanvas.yml)).


The value can be either a decimal number or a string. The string may be:

- A Unicode char (e.g. "☑")
- A hex value in U+ format (e.g. "U+2611")
- A hex value in UTF-16 format (e.g. "\\u2611")

```json
  "Glyphs.RightArrow": "►",
  "Glyphs.LeftArrow": "U+25C4",
  "Glyphs.DownArrow": "\\u25BC",
  "Glyphs.UpArrow": 965010
```

The `UI Catalog` application defines a `UICatalog` Theme. Look at the UI Catalog's `./Resources/config.json` file to see how to define a theme.

### Theme and Scheme Management

Terminal.Gui provides two key managers for handling visual themes and schemes:

The ThemeManager provides convenient methods for working with themes:

```csharp
// Get the currently active theme
ThemeScope currentTheme = ThemeManager.GetCurrentTheme();

// Get all available themes
Dictionary<string, ThemeScope> themes = ThemeManager.GetThemes();

// Get list of theme names
ImmutableList<string> themeNames = ThemeManager.GetThemeNames();

// Get/Set current theme name
string currentThemeName = ThemeManager.GetCurrentThemeName();
ThemeManager.Theme = "Dark"; // Switch themes

// Listen for theme changes
ThemeManager.ThemeChanged += (sender, e) => {
    // Handle theme changes
};
```

### SchemeManager

The SchemeManager handles schemes within themes. Each theme contains multiple schemes for different UI contexts:

```csharp
// Get current schemes
Dictionary<string, Scheme> schemes = SchemeManager.GetCurrentSchemes();

// Get list of scheme names
ImmutableList<string> schemeNames = SchemeManager.GetSchemeNames();

// Access specific schemes
Scheme topLevelScheme = SchemeManager.Schemes["TopLevel"];
Scheme dialogScheme = SchemeManager.Schemes["Dialog"];
Scheme menuScheme = SchemeManager.Schemes["Menu"];

// Listen for scheme changes
SchemeManager.CollectionChanged += (sender, e) => {
    // Handle scheme changes
};
```

### Built-in Schemes

The following schemes are available by default:

- **TopLevel**: Used for the application's top-level windows
- **Base**: Default scheme for most views
- **Dialog**: Used for dialogs and message boxes
- **Menu**: Used for menus and status bars
- **Error**: Used for error messages and dialogs

Each scheme defines colors and attributes for different view states (Normal, Focus, HotNormal, HotFocus, Disabled).

## Application Settings

Terminal.Gui provides several top-level application settings:

```json
{
  "Key.Separator": "+",
  "Application.ArrangeKey": "Ctrl+F5",
  "Application.Force16Colors": false,
  "Application.IsMouseDisabled": false,
  "Application.NextTabGroupKey": "F6",
  "Application.NextTabKey": "Tab",
  "Application.PrevTabGroupKey": "Shift+F6",
  "Application.PrevTabKey": "Shift+Tab",
  "Application.QuitKey": "Esc"
}
```

### View-Specific Settings

Settings that control specific view behaviors:

```json
{
  "PopoverMenu.DefaultKey": "Shift+F10",
  "FileDialog.MaxSearchResults": 10000,
  "FileDialogStyle.DefaultUseColors": false,
  "FileDialogStyle.DefaultUseUnicodeCharacters": false
}
```

### Key Bindings

> [!WARNING]
>  Configuration Manager support for key bindings is not yet implemented.

Key bindings are defined in the `KeyBindings` property of the configuration file. The value is an array of objects, each object defining a key binding. The key binding object has the following properties:

- `Key`: The key to bind to. The format is a string describing the key (e.g. "q", "Q,  "Ctrl+Q"). Function keys are specified as "F1", "F2", etc. 

## Loading And Applying Configuration

```csharp
// Enable configuration with specific locations
ConfigurationManager.Enable();
ConfigurationManager.Locations = ConfigLocations.Global | ConfigLocations.User;
ConfigurationManager.Load();
ConfigurationManager.Apply();
```

## Configuration Events

```csharp
// Configuration applied event
ConfigurationManager.Applied += (s, e) => {
    // Handle configuration changes
};

// Configuration loading event
ConfigurationManager.Loading += (s, e) => {
    Console.WriteLine($"Loading from: {e.Location}");
};
```

# Error Handling

```json
{
  "ConfigurationManager.ThrowOnJsonErrors": false
}
```

Set to `true` to throw exceptions on JSON parsing errors instead of silent failures.


# AOT Support

When using AOT compilation:

```csharp
[JsonSourceGenerationOptions(WriteIndented = true)]
[JsonSerializable(typeof(Configuration))]
internal partial class ConfigurationContext : JsonSerializerContext { }

// Enable with AOT support
ConfigurationManager.Enable(options => {
    options.JsonSerializerContext = ConfigurationContext.Default;
});
```

# Configuration File Schema

Settings are defined in JSON format, according to the schema found here:

## Schema

[!code-json[tui-config-schema.json](../schemas/tui-config-schema.json)]

## The Default Config File

To illustrate the syntax, the below is the `config.json` file found in `Terminal.Gui.dll`:

[!code-json[config.json](../../Terminal.Gui/Resources/config.json)]

tig:
Woot!!!!

WindowsTerminal_mcIXfuDhxI

tznind:
Wow that looks really nice!

tig:
There are some bugs in how `Style` is being output that might be hard to track down but this is going to enable some really clean UI.

WindowsTerminal_LvrYe8I4Lj

tig:
FWIW, I enabled `Attribute.Style` in `WindowsDriver` in this PR. 

Too many bugs in `v2win/v2net` that were slowing me down.

E.g. #4088

tig:
@BDisp , @tznind , @YourRobotOverlord, @Error-String-Expected-Got-Nil -

This PR is now good enough for you to start playing with. Not really ready to come out of Draft stage though.

@tznind - It would be great if you could pull this down and report back how painful it's going to be for you to update TGD with it.

@YourRobotOverlord - I'd love for you to pull this down and report back on how it works with your scheme-heavy app.

tznind:
> It would be great if you could pull this down and report back how painful it's going to be for you to update TGD with it.

Taking a look, shouldn't be too bad.  Theres quite basic interaction with `ColorScheme` at the moment in TGD.  Mostly there is `NamedColorScheme` which are not shared between seperate views.  Plenty of exciting opportunities.

I still have a big task updating to support the menu v2 too (see https://github.com/gui-cs/TerminalGuiDesigner/pull/312) which probably have to tackle first then onto this once its merged and can more easily pull nuget package.

tznind:
One thing I noticed is that switching Themes to another theme then back to Default does not seem to apply.

no-default

tig:
> One thing I noticed is that switching Themes to another theme then back to Default does not seem to apply.
> 

Yep. see Todos above.

> > It would be great if you could pull this down and report back how painful it's going to be for you to update TGD with it.
> 
> Taking a look, shouldn't be too bad. Theres quite basic interaction with `ColorScheme` at the moment in TGD. Mostly there is `NamedColorScheme` which are not shared between seperate views. Plenty of exciting opportunities.
> 
> I still have a big task updating to support the menu v2 too (see [gui-cs/TerminalGuiDesigner#312](https://github.com/gui-cs/TerminalGuiDesigner/pull/312)) which probably have to tackle first then onto this once its merged and can more easily pull nuget package.

One note on this: Generally, setting the View's Scheme directly anywhere but at construction (e.g. `SchemeName = "Menu"`) is not idea. Instead, folks should leverage the `Getting/SetttingScheme`, `Getting/SettingAttributeForRole` events.

tig:
> > One thing I noticed is that switching Themes to another theme then back to Default does not seem to apply.
> 
> Yep. see Todos above.
> 

Fixed in https://github.com/gui-cs/Terminal.Gui/pull/4062/commits/e42e4214a7e1f0eea4116fe2c6c008ac0b32914d

tig:
Just fixed some more CM dynamic update stuff...

tig:
WindowsTerminal_UHQqn6or49

Latest default theme.

tig:
TextView and TextField now have rational support for Enabled and ReadOnly attributes!

WindowsTerminal_hBqaIuLEKb

tig:
WindowsTerminal_IGxDDLsFqB

Still needs work, but looking pretty amazing if I say so myself...

tig:
This PR is now ready for review, @tznind, @BDisp etc...

WindowsTerminal_8mUKM36RmC

All that's really left is to 

- re-enable Hover/Highlight....
- Validate AOT stuff still works (I may defer this).

BDisp (approved):
Good work. I like it very much. The only repair I would like to express is when I press the keyboard keys down and up, on radio button or checked opened menus I can't see an indicator on what is currently selected. I know we can select by hot key, but that's the only thing I don't like in the v2 menus navigation.

tig:
> Good work. I like it very much. The only repair I would like to express is when I press the keyboard keys down and up, on radio button or checked opened menus I can't see an indicator on what is currently selected. I know we can select by hot key, but that's the only thing I don't like in the v2 menus navigation.
> 
> Pressing F9 to open the menu, then press Cursor.Right and then press Esc will exit the UICatalog instead only closing the menu.

Thanks. 

I'll be fixing those menu issues in #4013. 

I couldn't address the selection thing without fixing Schemes first.

tig:
WindowsTerminal_iTXAAkYgrW

tig:
> Only the Toplevel Disabled attribute isn't much perceptible.

Yeah. 

On this, I keep thinking "why do we have both Base and Toplevel schemes". In what world does anyone want both?

tig:
> > Yeah.
> > On this, I keep thinking "why do we have both Base and Toplevel schemes". In what world does anyone want both?
> 
> I'm only thinking of one possibility. Maybe I'm wrong but when you define a schema for `Window` and it's the `Application.Top`, all other `Window` sub-views will inherit the same schema, right? If affirmative a user may want a different schema for the `Application.Top` and other schema for the `Window` sub-views. In this case it's an advantage having the `Toplevel` schema. If it's possible to have different schemas for the same `Window` type then the `Toplevel` may be removed.

We'll decide this here (where I already wrote about this, but forgot):

- #4086

tig:
Ok, @BDisp and @tznind this is ready for final review. 

Menuv2, OptionSelector, and FlagSelector still do not highlight correctly. That will get fixed later.

It's possible that built-in-Views like TableView have scheme issues. We can fix those in subsequent PRs, but still call them out if you see 'em.

Thanks.

BDisp (approved):

tig:
@tznind Can you give me some idea of when you'll have a chance to review this?

Examples/UICatalog/Scenarios/Buttons.cs
```diff
@@ -138,7 +138,8 @@ static void DoMessage (Button button, string txt)
         var removeButton = new Button
         {
             X = 2, Y = Pos.Bottom (button) + 1,
-            ColorScheme = Colors.ColorSchemes ["Error"], Text = "Remove this button"
+            SchemeName = "Error",
```
tznind, line 141: So string is good for this,  but we should use a central const for the known core ones

Examples/UICatalog/Scenarios/Editors & Helpers/AdornmentsEditor.cs
```diff
@@ -18,6 +18,8 @@ public AdornmentsEditor ()
         ExpanderButton!.Orientation = Orientation.Horizontal;
 
         Initialized += AdornmentsEditor_Initialized;
+
```
tznind, line 21: The file path having space and ampersand in it is a bit odd. Probably not really an issue but why risk it - I'd suggest change to pascal case and use word And instead

Terminal.Gui/Application/Application.Driver.cs
```diff
@@ -26,7 +26,7 @@ public static partial class Application // Driver abstractions
     ///     Note, <see cref="Application.Init(IConsoleDriver, string)"/> will override this configuration setting if called
     ///     with either `driver` or `driverName` specified.
     /// </remarks>
-    [SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
+    [ConfigurationProperty (Scope = typeof (SettingsScope))]
```
tznind, line 29: Is the reason we use typeof instead of enum so API users can add their own scopes? I guess string with `nameof` would also work... just interested not really something i think needs changed.

Terminal.Gui/Application/Application.Mouse.cs
```diff
@@ -147,7 +150,7 @@ internal static void RaiseMouseEvent (MouseEventArgs mouseEvent)
         //Debug.Assert (mouseEvent.Position == mouseEvent.ScreenPosition);
         mouseEvent.Position = mouseEvent.ScreenPosition;
 
-        List<View?> currentViewsUnderMouse = View.GetViewsUnderMouse (mouseEvent.ScreenPosition);
+        List<View?> currentViewsUnderMouse = View.GetViewsUnderLocation (mouseEvent.ScreenPosition, ViewportSettings.TransparentMouse);
```
tznind, line 153: Nice, this is a more professional name for the method 👍 

We could change Location to 'ScreenPoint' or 'Point' to be even more specific (depending on what this function uses for its coordinate space)

Terminal.Gui/Application/Application.Run.cs
```diff
@@ -26,13 +26,13 @@ public static Key QuitKey
...
     /// <summary>Gets or sets the key to activate arranging views using the keyboard.</summary>
-    [SerializableConfigurationProperty (Scope = typeof (SettingsScope))]
+    [ConfigurationProperty (Scope = typeof (SettingsScope))]
     public static Key ArrangeKey
     {
         get => _arrangeKey;
         set
         {
-            if (_arrangeKey != value)
+            //if (_arrangeKey != value)
```
tznind, line 35: Is this meant to be commented out? The curly bracers remain...

Terminal.Gui/Configuration/SourcesManager.cs
```diff
@@ -0,0 +1,206 @@
...
+    /// <summary>INTERNAL: Loads the `config.json` file a <paramref name="filePath"/> into the specified <see cref="SettingsScope"/>.</summary>
+    /// <param name="settingsScope">The Settings Scope object that <paramref name="filePath"/> will be loaded into.</param>
+    /// <param name="filePath">Json document to update the settings with.</param>
+    /// <param name="location">The Config Location corresponding to <paramref name="filePath"/></param>
+    /// <returns><see langword="true"/> if the settingsScope was updated.</returns>
+    [RequiresUnreferencedCode ("AOT")]
+    [RequiresDynamicCode ("AOT")]
+    internal bool Load (SettingsScope? settingsScope, string filePath, ConfigLocations location)
+    {
+        string realPath = filePath.Replace ("~", Environment.GetFolderPath (Environment.SpecialFolder.UserProfile));
```
tznind, line 83: Nice, this is user friendly

tznind (approved):

tznind:
Really nice changes but a mammoth diff! I have put some comments in and it looks good.

Examples/UICatalog/Scenarios/Buttons.cs
tig:
```diff
+ SchemeName = SchemeManager.SchemesToSchemeName(Schemes.Error),
```

Terminal.Gui/Application/Application.Run.cs
tig: Yes. This will be fixed as part of fixing #2871 where `INotifyPropertyChanged` will come into play.

Terminal.Gui/Application/Application.Mouse.cs
tig: Good call

Now: 


```cs
    /// <summary>
    ///     Gets the Views that are under <paramref name="screenLocation"/>, including Adornments. The list is ordered by depth. The
    ///     deepest
    ///     View is at the end of the list (the top most View is at element 0).
    /// </summary>
    /// <param name="screenLocation">Screen-relative location.</param>
    /// <param name="excludeViewportSettingsFlags">
    ///     If set, excludes Views that have the <see cref="ViewportSettings.Transparent"/> or <see cref="ViewportSettings.TransparentMouse"/>
    ///     flags set in their ViewportSettings.
    /// </param>
    public static List<View?> GetViewsUnderLocation (in Point screenLocation, ViewportSettings excludeViewportSettingsFlags)
```

Issue #4058, opened 2025-04-26
Add Non-Color Attributes (Bold, Italic, Underline, etc.) to `Attributes`


tig:
As part of the `ColorSchemes` redesign discussed in

- #457 
 
Terminal.Gui’s `Attributes` (within `ColorScheme`, proposed to be renamed `Scheme`) currently only support colors but will soon need to include non-color styles to align with modern terminal capabilities and user expectations. These styles, inspired by PowerShell’s `$PSStyle` ANSI styles and common terminal features, include Bold, Italic, Underline, Blink, Reverse, Hidden, and Strikethrough. Adding these to `Attributes` will enable richer visual styling for Terminal.Gui views, support PowerShell-like TUI applications, and align with efforts to honor user terminal settings (#2381).

This issue focuses on extending `Attributes` to support these non-color styles, keeping #457 as the master issue for the broader `ColorSchemes` redesign, including naming changes (#4057).

**Current State**:
- `Attributes` (within `ColorScheme`) defines foreground and background colors for states like Normal, HotNormal, Focus, HotFocus, and Disabled.
- No support exists for non-color styles (e.g., Bold, Italic), limiting Terminal.Gui’s ability to render advanced terminal effects.
- PowerShell’s `$PSStyle` (PowerShell 7.2+) supports ANSI styles (Bold, Italic, Underline, Blink, Reverse, Hidden, Strikethrough), which Terminal.Gui should emulate for consistency with user terminal settings.

**Proposed Changes**:
1. **Extend `Attributes` struct**:
   - Define an `enum` for flags for non-color styles: `Bold`, `Italic`, `Underline`, `Blink`, `Reverse`, `Hidden`, `Strikethrough`.
   - Add a property for this enum to `Attribute`
 
 2. **Update Rendering**:
   - Modify Terminal.Gui’s rendering engine to apply these styles using ANSI escape sequences or terminal-specific APIs, leveraging truecolor support (#48).
   - Ensure compatibility with terminals supporting 16, 256, or 24-bit RGB colors.
   - 
3. **Integrate with `Scheme`**:
   - Update `Scheme` (or `ColorScheme`, pending #4057) to use the enhanced `Attributes` for all attributes (Normal, HotNormal, Focus, HotFocus, Disabled, Active, HotActive, ReadOnly).
   
   - Example:
     ```csharp
     var scheme = new Scheme
     {
         Normal = new Attribute { Foreground = Color.White, Bold = true },
[...]

PR #4079, opened 2025-05-12
More on #4058 - Adds `TextStyle` Scenario

tig:


ChristopheI:
Hi, I just tested this on Windows and it's not working ...
image

ChristopheI:
image

tig:
This only works with the v2win/v2net drivers. Start UI Catalog with

image

ChristopheI:
Wow ! I didn't know at all this kind of options ...
Now it's like I'm in a rocket: quick resizing, more emoji supported, 
Why v2net it's not the default ? there is some restrictions to use it ? Perhaps I miss some documentation about this ?
Thx

Error-String-Expected-Got-Nil:
It's not entirely completed yet. Notably, anything that still uses `Application.Run()` to create a new running `TopLevel` will break if used in a V2 app. And for me, Unicode support broke when I tried to use V2net with my terminal, but V2 Windows worked fine.

tznind:
> Notably, anything that still uses Application.Run() to create a new running TopLevel will break

@Error-String-Expected-Got-Nil are you able to provide minimal example please? this might be a missing routing but all relevant `Application` methods should automatically route to ApplicationV2 after doing `Application.Init` with v2.

ChristopheI:
On my side I tested v2net vs v2win (windows 10.0.22631)
I use this to set a driver (perhaps Im doing it in the wrong way):
```
        Console.OutputEncoding = Encoding.Default;
        var _forceDriver = "v2net"; // v2win
        Terminal.Gui.Application.ForceDriver = _forceDriver;
        Terminal.Gui.Application.Init(driverName: _forceDriver);
```

In v2net: more emoji working
image

In v2win:
image

Error-String-Expected-Got-Nil:
> > Notably, anything that still uses Application.Run() to create a new running TopLevel will break
> 
> @Error-String-Expected-Got-Nil are you able to provide minimal example please? this might be a missing routing but all relevant `Application` methods should automatically route to ApplicationV2 after doing `Application.Init(null,"v2")` .

Oh, I wasn't aware of that. In my case I would init an `ApplicationV2` instance and use that everywhere- I didn't realize you could init the base `Application` for V2. I just thought they were still incompatible. That would have saved me some trouble before... seems like a documentation issue, though I suppose that's not unexpected for a WIP feature.

PR #4085, opened 2025-05-14
Fixes #4080. TabGroup not always navigate correctly across groups

### LINKED ISSUE #4080: TabGroup not always navigate correctly across groups
Maybe I could be wrong but group navigation should always respect the current focused view in the current focused group.

```cs
    [Theory]
    [ClassData (typeof (V2TestDrivers))]
    public void Toplevel_TabGroup_Forward_Backward (V2TestDriver d)
    {
        var v1 = new View { Id = "v1", CanFocus = true };
        var v2 = new View { Id = "v2", CanFocus = true };
        var v3 = new View { Id = "v3", CanFocus = true };
        var v4 = new View { Id = "v4", CanFocus = true };

        using GuiTestContext c = With.A<Window> (50, 20, d)
                                     .Then (
                                            () =>
                                            {
                                                var w1 = new Window { Id = "w1" };
                                                w1.Add (v1, v2);
                                                var w2 = new Window { Id = "w2" };
                                                w2.Add (v3, v4);
                                                Toplevel top = Application.Top!;
                                                Application.Top!.Add (w1, w2);
                                            })
                                     .WaitIteration ()
                                     .Then (() => Assert.True (v3.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6)
                                     .Then (() => Assert.True (v1.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6)
                                     .Then (() => Assert.True (v3.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6.WithShift)
                                     .Then (() => Assert.True (v1.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6.WithShift)
                                     .Then (() => Assert.True (v3.HasFocus))
                                     .RaiseKeyDownEvent (Key.Tab)
                                     .Then (() => Assert.True (v4.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6)
                                     .Then (() => Assert.True (v1.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6)
                                     .Then (() => Assert.True (v4.HasFocus)) // Fails because it navigates to the previous view v3
                                     .RaiseKeyDownEvent (Key.F6.WithShift)
                                     .Then (() => Assert.True (v1.HasFocus))
                                     .RaiseKeyDownEvent (Key.Tab)
                                     .Then (() => Assert.True (v2.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6.WithShift)
                                     .Then (() => Assert.True (v4.HasFocus))
                                     .RaiseKeyDownEvent (Key.F6.WithShift)
                                     .Then (() => Assert.True (v2.HasFocus))
                                     .WriteOutLogs (_out)
                                     .Stop ();
        Assert.False (v1.HasFocus);
        Assert.False (v2.HasFocus);
        Assert.False (v3.HasFocus);
        Assert.False (v4.HasFocus);
    }
```

BDisp:
## Fixes

- Fixes #4080

## Proposed Changes/Todos

- [x] Added new method `MoveOverlappedSubViewToEnd`
- [x] Fix and added 2 more unit tests

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Application/Application.Keyboard.cs
+++ b/Terminal.Gui/Application/Application.Keyboard.cs
@@ -97,17 +97,11 @@ public static partial class Application // Keyboard handling
             }
             else
             {
-                // BUGBUG: this seems unneeded.
-                if (!KeyBindings.TryGet (key, out KeyBinding keybinding))
-                {
-                    return null;
-                }
-
                 bool? toReturn = null;
 
-                foreach (Command command in keybinding.Commands)
+                foreach (Command command in binding.Commands)
                 {
-                    toReturn = InvokeCommand (command, key, keybinding);
+                    toReturn = InvokeCommand (command, key, binding);
                 }
 
                 handled = toReturn ?? true;
--- a/Terminal.Gui/View/View.Hierarchy.cs
+++ b/Terminal.Gui/View/View.Hierarchy.cs
@@ -447,6 +447,26 @@ public partial class View // SuperView/SubView hierarchy management (SuperView, 
     #region SubViewOrdering
 
     /// <summary>
+    ///     Moves <paramref name="subview"/> to the end of the <see cref="SubViews"/> list keeping the original sorting.
+    /// </summary>
+    /// <param name="subview">The subview to move.</param>
+    public void MoveOverlappedSubViewToEnd (View subview)
+    {
+        PerformActionForSubView (
+                                 subview,
+                                 x =>
+                                 {
+                                     while (InternalSubViews!.IndexOf (x) != InternalSubViews.Count - 1)
+                                     {
+                                         var v = InternalSubViews [0];
+                                         InternalSubViews!.Remove (v);
+                                         InternalSubViews.Add (v);
+                                     }
+                                 }
+                                );
+    }
+
+    /// <summary>
     ///     Moves <paramref name="subview"/> one position towards the end of the <see cref="SubViews"/> list.
     /// </summary>
     /// <param name="subview">The subview to move.</param>
--- a/Terminal.Gui/View/View.Navigation.cs
+++ b/Terminal.Gui/View/View.Navigation.cs
@@ -61,14 +61,31 @@ public partial class View // Focus and cross-view navigation management (TabStop
             if (direction == NavigationDirection.Forward && focused == focusChain [^1] && SuperView is null)
             {
                 // We're at the top of the focus chain. Go back down the focus chain and focus the first TabGroup
-                View [] views = GetFocusChain (NavigationDirection.Forward, TabBehavior.TabGroup);
-
-                if (views.Length > 0)
+                if (focusChain.Length > 0)
                 {
-                    View [] subViews = views [0].GetFocusChain (NavigationDirection.Forward, TabBehavior.TabStop);
+                    // Get the index of the currently focused view
+                    int focusedTabGroupIndex = focusChain.IndexOf (Focused); // Will return -1 if Focused can't be found or is null
+
+                    if (focusedTabGroupIndex + 1 > focusChain.Length - 1)
+                    {
+                        focusedTabGroupIndex = 0;
+                    }
+                    else
+                    {
+                        focusedTabGroupIndex++;
+                    }
+
+                    View [] subViews = focusChain [focusedTabGroupIndex].GetFocusChain (NavigationDirection.Forward, TabBehavior.TabStop);
 
                     if (subViews.Length > 0)
                     {
+                        if (focusChain [focusedTabGroupIndex]._previouslyFocused is { }
+                            && subViews.Any (v => v == focusChain [focusedTabGroupIndex]._previouslyFocused))
+                        {
+                            return focusChain [focusedTabGroupIndex]._previouslyFocused!.SetFocus ();
+                        }
+
+                        // We have a subview that can be focused
                         if (subViews [0].SetFocus ())
                         {
                             return true;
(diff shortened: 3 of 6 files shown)


Terminal.Gui/Application/Application.Keyboard.cs
```diff
@@ -97,17 +97,11 @@ public static bool RaiseKeyDownEvent (Key key)
             }
             else
             {
-                // BUGBUG: this seems unneeded.
```
tig, line 100: Are you 100% sure this won't break anything? 

I am not, which is why I have not previously cleaned it up.

Terminal.Gui/View/View.Hierarchy.cs
```diff
@@ -446,6 +446,26 @@ public static bool IsInHierarchy (View? start, View? view, bool includeAdornment
 
     #region SubViewOrdering
 
+    /// <summary>
+    ///     Moves <paramref name="subview"/> to the end of the <see cref="SubViews"/> list keeping the original sorting.
+    /// </summary>
+    /// <param name="subview">The subview to move.</param>
+    public void MoveOverlappedSubViewToEnd (View subview)
```
tig, line 453: I don't understand the difference between this an MoveSubViewToEnd.  Is the difference that MoveSubViewToEnd changes the original sorting?

Terminal.Gui/View/View.Navigation.cs
```diff
@@ -619,7 +639,7 @@ public bool SetFocus ()
 
         if (Arrangement.HasFlag (ViewArrangement.Overlapped))
         {
-            SuperView?.MoveSubViewToEnd (this);
+            SuperView?.MoveOverlappedSubViewToEnd (this);
```
tig, line 642: I suggest you nuke MoveOverlappedSubViewToEnd and put this flag check inside MoveubViewToEnd.

Will that work?

tig (changes requested):

Terminal.Gui/Application/Application.Keyboard.cs
BDisp: I thinking it's a duplicated running always with the same returning result and thus I think it's unneeded. I didn't experienced any situation when the second call was different from prior. Also no unit test fail which I think it's secure to remove it.

Terminal.Gui/View/View.Navigation.cs
BDisp: No they are different. `MoveSubViewToEnd` is intended to change the sub-views order and `MoveOverlappedSubViewToEnd` always maintain the same order. Otherwise you haven't a way to change the sub-views order.

Terminal.Gui/View/View.Hierarchy.cs
BDisp: Yes, `MoveSubViewToEnd` is intended to change the sub-views order and `MoveOverlappedSubViewToEnd` always maintain the same order. When you want to show the current top, `MoveOverlappedSubViewToEnd` move to begin but always keeping the original sorting.

tig: I mean do this:

```cs


    /// <summary>
    ///     Moves <paramref name="subview"/> to the end of the <see cref="SubViews"/> list.
    /// </summary>
    /// <param name="subview">The subview to move.</param>
    public void MoveSubViewToEnd (View subview)
    {
        if (Arrangement.HasFlag (ViewArrangement.Overlapped))
        {
            PerformActionForSubView (
                                 subview,
                                 x =>
                                 {
                                     while (InternalSubViews!.IndexOf (x) != InternalSubViews.Count - 1)
                                     {
                                         var v = InternalSubViews [0];
                                         InternalSubViews!.Remove (v);
                                         InternalSubViews.Add (v);
                                     }
                                 }
                                );
             return;
        }
        PerformActionForSubView (
                                 subview,
                                 x =>
                                 {
                                     InternalSubViews!.Remove (x);
                                     InternalSubViews.Add (x);
                                 }
                                );
    }
```

And then, in `View.Navigation.cs` do this:

```diff
-        if (Arrangement.HasFlag (ViewArrangement.Overlapped))
-        {
            SuperView?.MoveSubViewToEnd (this);
-        }
```

tig: I don't like having a special "Move" function specific for Overlapped.

BDisp: I got it. Done. Thanks.

Terminal.Gui/ViewBase/View.Navigation.cs
```diff
@@ -640,7 +640,7 @@ public bool SetFocus ()
 
         if (Arrangement.HasFlag (ViewArrangement.Overlapped))
         {
-            SuperView?.MoveOverlappedSubViewToEnd (this);
+            SuperView?.MoveSubViewToEnd (this);
```
tig: What happens if you remove the if statement surrounding this call?

BDisp:
With the `MoveSubViewToEnd_ViewArrangement` unit test you can't see the difference and behave the same no matter is used `ViewArrangement.Fixed` or `ViewArrangement.Overlapped`. The only unit test where you can confirm that this PR really behave as expected is the `Toplevel_TabGroup_Forward_Backward` because the issue is only visible when a driver is really running.

BDisp:
This isn't working in release mode in the `NativeAot` project.

tig (approved):
Well done.!

Terminal.Gui/ViewBase/View.Navigation.cs
BDisp: Without using a driver behave the same but with a driver pressing F6, Shift+F6, Tab and Shift+Tab in the sub-views the order is break and doesn't work as expected.

PR #4092, opened 2025-05-16
Fixes #3209 - Formalize Cancellable Work Pattern and add helpers

### LINKED ISSUE #3209: Formalize the Cancellable Work Pattern as v2s core eventing model

tig:
## Fixes

- Fixes #3209

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:


tig:
This is good enough for now I think. Makes the point. 

Curious your thoughts folks.

BDisp (approved):
That's fantastic. I'm really impressed.

PR #4114, opened 2025-06-01
Adds `namespace` docs

tig:
E.g.:

image

### DIFF
--- a/docfx/apispec/namespace-app.md
+++ b/docfx/apispec/namespace-app.md
@@ -1,0 +1,27 @@
+---
+uid: Terminal.Gui.App
+summary: The `App` namespace holds @Terminal.Gui.Application and associated classes.
+---
+
+@Terminal.Gui.Application provides the main entry point and core application logic for Terminal.Gui applications. This static singleton class is responsible for initializing and shutting down the Terminal.Gui environment, managing the main event loop, handling input and output, and providing access to global application state.
+
+Typical usage involves calling `Application.Init()` to initialize the application, creating and running a `Window`, and then calling `Application.Shutdown()` to clean up resources. The class also provides methods for culture support, idle handlers, and rendering the application state as a string.
+
+## Example Usage
+
+```csharp
+Application.Init();
+var win = new Window()
+{
+    Title = $"Example App ({Application.QuitKey} to quit)"
+};
+Application.Run(win);
+win.Dispose();
+Application.Shutdown();
+```
+
+## See Also
+
+- [Logging](logging.md)
+- [Mainloop](mainloop.md)
+<!-- - [Application](application.md) -->
--- a/docfx/apispec/namespace-configuration.md
+++ b/docfx/apispec/namespace-configuration.md
@@ -1,0 +1,30 @@
+---
+uid: Terminal.Gui.Configuration
+summary: The `Configuration` namespace provides comprehensive configuration management for Terminal.Gui applications.
+---
+
+@Terminal.Gui.Configuration provides a robust system for managing application settings, themes, and runtime configuration. This namespace includes the configuration manager, property attributes, and scoping mechanisms that allow applications to persist and load settings from various sources.
+
+The configuration system supports multiple scopes (Settings, Themes, AppSettings) and sources (user directory, application directory, etc.), enabling flexible deployment and customization scenarios. It also provides theme inheritance and hot-reloading capabilities for dynamic configuration updates.
+
+## Key Components
+
+- **ConfigurationManager**: Central system for loading, applying, and managing configuration
+- **ConfigProperty**: Attribute for marking properties as configurable
+- **Scopes**: Settings, Theme, and AppSettings contexts
+- **Sources**: Multiple storage locations for configuration persistence
+
+## Example Usage
+
+```csharp
+// Mark a property as configurable
+[ConfigProperty]
+public static bool MyFeatureEnabled { get; set; } = true;
+
+// Load configuration from default sources
+ConfigurationManager.Enable(ConfigLocations.All);
+```
+
+## Deep Dive
+
+- [Configuration Management Deep Dive](~/docs/config.md) - Comprehensive configuration system documentation
--- a/docfx/apispec/namespace-drawing.md
+++ b/docfx/apispec/namespace-drawing.md
@@ -1,0 +1,28 @@
+---
+uid: Terminal.Gui.Drawing
+summary: The `Drawing` namespace provides comprehensive text rendering, color management, and visual styling for Terminal.Gui applications.
+---
+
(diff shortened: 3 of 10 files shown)


tig:
Reviewers, please review this new doc:

# Multitasking and Background Operations

_See also [Cross-platform Driver Model](drivers.md)_

Terminal.Gui applications run on a single main thread with an event loop that processes keyboard, mouse, and system events. This document explains how to properly handle background work, timers, and asynchronous operations while keeping your UI responsive.

## Threading Model

Terminal.Gui follows the standard UI toolkit pattern where **all UI operations must happen on the main thread**. Attempting to modify views or their properties from background threads will result in undefined behavior and potential crashes.

### The Golden Rule
> Always use `Application.Invoke()` to update the UI from background threads.

## Background Operations

### Using async/await (Recommended)

The preferred way to handle background work is using C#'s async/await pattern:

```csharp
private async void LoadDataButton_Clicked()
{
    loadButton.Enabled = false;
    statusLabel.Text = "Loading...";
    
    try
    {
        // This runs on a background thread
        var data = await FetchDataFromApiAsync();
        
        // This automatically returns to the main thread
        dataView.LoadData(data);
        statusLabel.Text = $"Loaded {data.Count} items";
    }
    catch (Exception ex)
    {
        statusLabel.Text = $"Error: {ex.Message}";
    }
    finally
    {
        loadButton.Enabled = true;
    }
}
```

### Using Application.Invoke()

When working with traditional threading APIs or when async/await isn't suitable:

```csharp
private void StartBackgroundWork()
{
    Task.Run(() =>
    {
        // This code runs on a background thread
        for (int i = 0; i <= 100; i++)
        {
            Thread.Sleep(50); // Simulate work
            
            // Marshal back to main thread for UI updates
            Application.Invoke(() =>
            {
                progressBar.Fraction = i / 100f;
                statusLabel.Text = $"Progress: {i}%";
            });
        }
        
        Application.Invoke(() =>
        {
            statusLabel.Text = "Complete!";
        });
    });
}
```

## Timers

Use timers for periodic updates like clocks, status refreshes, or animations:

```csharp
public class ClockView : View
{
    private Label timeLabel;
    private object timerToken;
    
    public ClockView()
    {
        timeLabel = new Label { Text = DateTime.Now.ToString("HH:mm:ss") };
        Add(timeLabel);
        
        // Update every second
        timerToken = Application.MainLoop.AddTimeout(
            TimeSpan.FromSeconds(1), 
            UpdateTime
        );
    }
    
    private bool UpdateTime()
    {
        timeLabel.Text = DateTime.Now.ToString("HH:mm:ss");
        return true; // Continue timer
    }
    
    protected override void Dispose(bool disposing)
    {
        if (disposing && timerToken != null)
        {
            Application.MainLoop.RemoveTimeout(timerToken);
        }
        base.Dispose(disposing);
    }
}
```

### Timer Best Practices

- **Always remove timers** when disposing views to prevent memory leaks
- **Return `true`** from timer callbacks to continue, `false` to stop
- **Keep timer callbacks fast** - they run on the main thread
- **Use appropriate intervals** - too frequent updates can impact performance

## Idle Processing

Idle handlers run when the application has no events to process, useful for background maintenance:

```csharp
public class AutoSaveView : View
{
    private object idleToken;
    private DateTime lastSave = DateTime.Now;
    
    public AutoSaveView()
    {
        idleToken = Application.MainLoop.AddIdle(CheckAutoSave);
    }
    
    private bool CheckAutoSave()
    {
        if (DateTime.Now - lastSave > TimeSpan.FromMinutes(5))
        {
            if (HasUnsavedChanges())
            {
                SaveDocument();
                lastSave = DateTime.Now;
            }
        }
        return true; // Continue idle processing
    }
    
    protected override void Dispose(bool disposing)
    {
        if (disposing && idleToken != null)
        {
            Application.MainLoop.RemoveIdle(idleToken);
        }
        base.Dispose(disposing);
    }
}
```

## Common Patterns

### Progress Reporting

```csharp
private async void ProcessFiles()
{
    var files = Directory.GetFiles(folderPath);
    progressBar.Fraction = 0;
    
    for (int i = 0; i < files.Length; i++)
    {
        await ProcessFileAsync(files[i]);
        
        // Update progress on main thread
        progressBar.Fraction = (float)(i + 1) / files.Length;
        statusLabel.Text = $"Processed {i + 1} of {files.Length} files";
        
        // Allow UI to update
        await Task.Yield();
    }
}
```

### Cancellation Support

```csharp
private CancellationTokenSource cancellationSource;

private async void StartLongOperation()
{
    cancellationSource = new CancellationTokenSource();
    cancelButton.Enabled = true;
    
    try
    {
        await LongRunningOperationAsync(cancellationSource.Token);
        statusLabel.Text = "Operation completed";
    }
    catch (OperationCanceledException)
    {
        statusLabel.Text = "Operation cancelled";
    }
    finally
    {
        cancelButton.Enabled = false;
    }
}

private void CancelButton_Clicked()
{
    cancellationSource?.Cancel();
}
```

### Responsive UI During Blocking Operations

```csharp
private async void ProcessLargeDataset()
{
    var data = GetLargeDataset();
    var batchSize = 100;
    
    for (int i = 0; i < data.Count; i += batchSize)
    {
        // Process a batch
        var batch = data.Skip(i).Take(batchSize);
        ProcessBatch(batch);
        
        // Update UI and yield control
        progressBar.Fraction = (float)i / data.Count;
        await Task.Yield(); // Allows UI events to process
    }
}
```

## Common Mistakes to Avoid

### ❌ Don't: Update UI from background threads
```csharp
Task.Run(() =>
{
    label.Text = "This will crash!"; // Wrong!
});
```

### ✅ Do: Use Application.Invoke()
```csharp
Task.Run(() =>
{
    Application.Invoke(() =>
    {
        label.Text = "This is safe!"; // Correct!
    });
});
```

### ❌ Don't: Forget to clean up timers
```csharp
// Memory leak - timer keeps running after view is disposed
Application.MainLoop.AddTimeout(TimeSpan.FromSeconds(1), UpdateStatus);
```

### ✅ Do: Remove timers in Dispose
```csharp
protected override void Dispose(bool disposing)
{
    if (disposing && timerToken != null)
    {
        Application.MainLoop.RemoveTimeout(timerToken);
    }
    base.Dispose(disposing);
}
```

## Performance Considerations

- **Batch UI updates** when possible instead of updating individual elements
- **Use appropriate timer intervals** - 100ms is usually the maximum useful rate
- **Yield control** in long-running operations with `await Task.Yield()`
- **Consider using `ConfigureAwait(false)`** for non-UI async operations
- **Profile your application** to identify performance bottlenecks

## See Also

- [Events](events.md) - Event handling patterns
- [Keyboard Input](keyboard.md) - Keyboard event processing
- [Mouse Input](mouse.md) - Mouse event handling
- [Configuration Management](config.md) - Application settings and state

PR #4126, opened 2025-06-04
Fixes #4050 - `Select`->`Activate`

### LINKED ISSUE #4050: Rename/Refactor `Command.Select` and `Selecting` Event to `Activate`
## Summary

The `Command.Select`, `Selecting` event, and related APIs (`OnSelecting`, `RaiseSelecting`) should be renamed to `Command.Activate`, `Activating` event, `OnActivating`, and `RaiseActivating` to improve clarity and consistency in the `Command` system. The term "select" is ambiguous for stateless views (e.g., `Button`, `MenuItemv2`, where it only sets focus) and less precise for non-list-based state changes (e.g., toggling a `CheckBox`). "Activate" better captures the action of changing a view’s state or preparing it for interaction, aligning with both stateful (e.g., `CheckBox`, `FlagSelector`, `ListView`) and stateless (e.g., `Button`, `MenuItem`) use cases while clearly distinguishing from `Accepting`.

## Motivation
The current `Command.Select` and `Selecting` terminology has several issues:
1. **Ambiguity in Stateless Views**: For views like `Button` or `MenuItem`, `Command.Select` only sets focus, which isn’t typically considered “selection,” leading to developer confusion.
2. **Imprecise for Non-List Contexts**: In `CheckBox` (toggling state) or `FlagSelector` (toggling flags), “select” doesn’t fully capture the state change, unlike list-based selection in `ListView`.
3. **Overlap with Accepting**: In views like `ListView` or `Menu`, `Selecting` and `Accepting` can feel similar (e.g., Enter triggering both), requiring clearer distinction.
4. **Propagation Needs**: `MenuBar` requires `Activating` events from subviews to manage `PopoverMenu` visibility, but the current local handling model relies on view-specific events (e.g., `SelectedMenuItemChanged`), which isn’t generalizable. Subviews should remain decoupled from superviews, avoiding knowledge of superview implementation details (e.g., `MenuBar`’s popover logic).

The `Activate` terminology addresses these by:
- Representing both state changes (e.g., toggling a `CheckBox`, selecting a `ListView` item) and preparatory actions (e.g., focusing a `Button`, navigating a `MenuItemv2`).
- Distinguishing clearly from `Accepting`, which confirms actions (e.g., executing a menu command, submitting a dialog).
- Supporting a targeted propagation model with `PropagateActivating`, enabling superviews like `MenuBar` to opt-in to subview `Activating` events without coupling subviews to superview details.

### 1. Rename `Command.Select` and Related APIs to `Activate`
- Rename `Command.Select` to `Command.Activate` in the `Command` enum.
- Rename `Selecting` event to `Activating`, `OnSelecting` to `OnActivating`, and `RaiseSelecting` to `RaiseActivating` in the `View` class and all derived classes (e.g., `Menuv2`, `MenuItemv2`, `CheckBox`, `FlagSelector`).
- Update related code:
  - Modify `SetupCommands` in `View` to use `Command.Activate`.
  - Update `Shortcut.DispatchCommand` and other command handlers to reference `Activating`.
  - Ensure all event handlers and documentation reflect the new terminology.
  - Ensure all tests are appropriately renamed and updated
 
## Related

- #2404

tig:
## Fixes

- Fixes #4050

## Proposed Changes/Todos

- [x] `Command.Select->Activate`
- [ ] Rename `Selecting` event to `Activating`, `OnSelecting` to `OnActivating`, and `RaiseSelecting` to `RaiseActivating` in the `View` class and all derived classes (e.g., `Menuv2`, `MenuItemv2`, `CheckBox`, `FlagSelector`).
- [ ] Update related code:
  - [ ] Modify `SetupCommands` in `View` to use `Command.Activate`.
  - [ ] Update `Shortcut.DispatchCommand` and other command handlers to reference `Activating`.
  - [ ] Ensure all event handlers and documentation reflect the new terminology.
- [ ] Introduce Targeted Propagation Model with `PropagateActivating`
- [ ] Update Documentation
  - [ ] Clarify `Activating` vs. `Accepting`:
  - [ ] Document propagation: Explain the targeted model with `PropagateActivating`, providing examples for `MenuBarv2` and potential `Toplevel` use cases.
  - [ ] Update view-specific docs:

### DIFF
--- a/Terminal.Gui/Input/Command.cs
+++ b/Terminal.Gui/Input/Command.cs
@@ -48,12 +48,15 @@ public enum Command
     HotKey,
 
     /// <summary>
-    ///     Selects the View or an item in the View (e.g. a list item or menu item) without necessarily accepting it.
+    ///     Activates the View or an item in the View (e.g. a list item or menu item) without necessarily accepting it.
+    ///     <para>
+    ///         In some cases, activating a View just sets focus to it, while in others it may trigger an action.
+    ///     </para>
     ///     <para>
     ///         The default implementation in <see cref="View"/> calls <see cref="View.RaiseSelecting"/>.
     ///     </para>
     /// </summary>
-    Select,
+    Activate,
 
     #endregion
 
--- a/Terminal.Gui/ViewBase/View.Command.cs
+++ b/Terminal.Gui/ViewBase/View.Command.cs
@@ -38,7 +38,7 @@ public partial class View // Command APIs
 
         // Space or single-click - Raise Selecting
         AddCommand (
-                    Command.Select,
+                    Command.Activate,
                     ctx =>
                     {
                         if (RaiseSelecting (ctx) is true)
@@ -203,9 +203,9 @@ public partial class View // Command APIs
     public event EventHandler<CommandEventArgs>? Accepting;
 
     /// <summary>
-    ///     Called when the user has performed an action (e.g. <see cref="Command.Select"/>) causing the View to change state.
+    ///     Called when the user has performed an action (e.g. <see cref="Command.Activate"/>) causing the View to change state.
     ///     Calls <see cref="OnSelecting"/> which can be cancelled; if not cancelled raises <see cref="Accepting"/>.
-    ///     event. The default <see cref="Command.Select"/> handler calls this method.
+    ///     event. The default <see cref="Command.Activate"/> handler calls this method.
     /// </summary>
     /// <remarks>
     ///     The <see cref="Selecting"/> event should be raised after the state of the View has been changed and before see
@@ -236,7 +236,7 @@ public partial class View // Command APIs
     }
 
     /// <summary>
-    ///     Called when the user has performed an action (e.g. <see cref="Command.Select"/>) causing the View to change state.
+    ///     Called when the user has performed an action (e.g. <see cref="Command.Activate"/>) causing the View to change state.
     ///     Set CommandEventArgs.Handled to <see langword="true"/> and return <see langword="true"/> to indicate the event was
     ///     handled and processing should stop.
     /// </summary>
@@ -245,7 +245,7 @@ public partial class View // Command APIs
     protected virtual bool OnSelecting (CommandEventArgs args) { return false; }
 
     /// <summary>
-    ///     Cancelable event raised when the user has performed an action (e.g. <see cref="Command.Select"/>) causing the View
+    ///     Cancelable event raised when the user has performed an action (e.g. <see cref="Command.Activate"/>) causing the View
     ///     to change state.
     ///     Set CommandEventArgs.Handled to <see langword="true"/> to indicate the event was handled and processing should
     ///     stop.
--- a/Terminal.Gui/ViewBase/View.Keyboard.cs
+++ b/Terminal.Gui/ViewBase/View.Keyboard.cs
@@ -10,7 +10,7 @@ public partial class View // Keyboard APIs
     private void SetupKeyboard ()
     {
         KeyBindings = new (this);
-        KeyBindings.Add (Key.Space, Command.Select);
+        KeyBindings.Add (Key.Space, Command.Activate);
         KeyBindings.Add (Key.Enter, Command.Accept);
 
         HotKeyBindings = new (this);
--- a/Terminal.Gui/ViewBase/View.Mouse.cs
+++ b/Terminal.Gui/ViewBase/View.Mouse.cs
@@ -13,11 +13,11 @@ public partial class View // Mouse APIs
         MouseBindings = new ();
 
         // TODO: Should the default really work with any button or just button1?
-        MouseBindings.Add (MouseFlags.Button1Clicked, Command.Select);
-        MouseBindings.Add (MouseFlags.Button2Clicked, Command.Select);
-        MouseBindings.Add (MouseFlags.Button3Clicked, Command.Select);
-        MouseBindings.Add (MouseFlags.Button4Clicked, Command.Select);
-        MouseBindings.Add (MouseFlags.Button1Clicked | MouseFlags.ButtonCtrl, Command.Select);
+        MouseBindings.Add (MouseFlags.Button1Clicked, Command.Activate);
+        MouseBindings.Add (MouseFlags.Button2Clicked, Command.Activate);
+        MouseBindings.Add (MouseFlags.Button3Clicked, Command.Activate);
+        MouseBindings.Add (MouseFlags.Button4Clicked, Command.Activate);
+        MouseBindings.Add (MouseFlags.Button1Clicked | MouseFlags.ButtonCtrl, Command.Activate);
     }
 
     /// <summary>
--- a/Terminal.Gui/Views/CharMap/CharMap.cs
+++ b/Terminal.Gui/Views/CharMap/CharMap.cs
@@ -43,7 +43,7 @@ public class CharMap : View, IDesignable
         AddCommand (Command.ScrollLeft, () => ScrollHorizontal (-1));
 
         AddCommand (Command.Accept, HandleAcceptCommand);
-        AddCommand (Command.Select, HandleSelectCommand);
+        AddCommand (Command.Activate, HandleSelectCommand);
         AddCommand (Command.Context, HandleContextCommand);
 
         KeyBindings.Add (Key.CursorUp, Command.Up);
--- a/Terminal.Gui/Views/CheckBox.cs
+++ b/Terminal.Gui/Views/CheckBox.cs
@@ -27,7 +27,7 @@ public class CheckBox : View
         CanFocus = true;
 
         // Select (Space key and single-click) - Advance state and raise Select event - DO NOT raise Accept
-        AddCommand (Command.Select, AdvanceAndSelect);
+        AddCommand (Command.Activate, AdvanceAndSelect);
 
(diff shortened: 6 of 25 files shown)


Examples/UICatalog/Scenarios/EditorsAndHelpers/DimEditor.cs
```diff
@@ -93,7 +93,7 @@ private void DimEditor_Initialized (object? sender, EventArgs e)
         };
         Add (label);
         _dimRadioGroup = new () { X = 0, Y = Pos.Bottom (label), RadioLabels = _radioItems };
-        _dimRadioGroup.SelectedItemChanged += OnRadioGroupOnSelectedItemChanged;
+        _dimRadioGroup.SelectedItemChanged += OnRadioGroupOnSelectedtemChanged;
```
BDisp: Typo `OnRadioGroupOnSelectedtemChanged` should be `OnRadioGroupOnSelectedItemChanged` or `OnRadioGroupOnSelectedChanged`.

BDisp (approved):

Examples/UICatalog/Scenarios/EditorsAndHelpers/DimEditor.cs
tig: I just fixed all these.

tig:
I'm not convinced we don't still need a `Command.Select` in ADDITION to `Command.Accept`.

Code like this in `TableView`...

```cs
        // BUGBUG: OnCellActivated is misnamed, it should be OnCellAccepted? Or is it OnCellSelected?
        // BUGBUG: Does this mean we still need Command.Select?
        AddCommand (Command.Accept, () => OnCellActivated (new (Table, SelectedColumn, SelectedRow)));
```

And TreeView:

```cs
    /// <summary>
    ///     <para>Triggers the <see cref="ObjectActivated"/> event with the <see cref="SelectedObject"/>.</para>
    ///     <para>This method also ensures that the selected object is visible.</para>
    /// </summary>
    /// <returns><see langword="true"/> if <see cref="ObjectActivated"/> was fired.</returns>
    public bool? ActivateSelectedObjectIfAny (ICommandContext commandContext)
    {
        // By default, Command.Accept calls OnAccept, so we need to call it here to ensure that the event is fired.
        if (RaiseAccepting (commandContext) == true)
        {
            return true;
        }

        T o = SelectedObject;

        if (o is { })
        {
            // TODO: Should this be cancelable?
            ObjectActivatedEventArgs<T> e = new (this, o);
            OnObjectActivated (e);

            return true;
        }

        return false;
    }
```

???

BDisp:
I think it make sense to have both. `Command.Select` doesn't mean it was accepted. So, adding `Command.Accept` will mean it was confirmed and will perform some action, like closing or opening a view, etc.... `Command.Select` also may perform some action but not definitely. Both must raise the respective events.

tznind:
>[we] still need a Command.Select in ADDITION to Command.Accept

yes table and tree view both distinguish between selecting (focusing) a cell/object and activating (opening) it.

BDisp:
Since you're dealing with `Select/Activate` it's normal a `Button` handles `Accept` when a `Shortcuts` in a `ScrollBar` are clicked?

WindowsTerminal_D6JtYpxtM8

BDisp:
There is one situation in `Terminal.Gui` which make me crazy. A simple `Accepting` event which will allow the developer to `Cancel` the default behavior (but has the `Handled` property) will repeat the event 3 times by mouse if the developer doesn't Cancel it. `UICatalog` has a bunch of `Button`'s that if they are activated by mouse will repeat the event 3 times, no matter the user press `Key.Enter` or `Key.Esc`. Despite by adding `Handled = true` fix this, there is no necessity of this extra code for a normal `Accepting` event. We can't handle mouse event as key event because only the only the view where the mouse is will be executed and doesn't not propagate to other views and with key event it can be propagate to other views that may want to handle it. I agree the leverage of keybindings, commands and context to achieve this behavior but we need to handle the mouse with much more careful.

tig:
I'm not entirely sure what you mean. But I am working through this here. 

Can you create a simple sample or unit test that illustrates?

BDisp:
It's only enough to comment the line `e.Handled = true;` in the `Generic` scenario. There are many scenarios where this isn't implemented and I think it's an unnecessary extra code. The default behavior when a Button is clicked by mouse it only raise once. If you press `Key.Enter` you'll see that this issue doesn't happens even the commented code. It's only happens with mouse. I still don't agree with the need of adding e.Handled = true on this eevnt. Probably will make much more sense in a `Selecting` event which can be canceled and if it isn't canceled an event `Accepted` not cancelable will being raise.

```cs
        button.Accepting += (s, e) =>
                            {
                                // When Accepting is handled, set e.Handled to true to prevent further processing.
                                //e.Handled = true;
                                MessageBox.ErrorQuery ("Error", "You pressed the button!", "_Ok");
                            };
```

BDisp:
I really can't repro this with the `FakeDriver` but I already found that all drivers really send a clicked event with the `Button` position when no mouse button pressed is performed. Seems that all drivers are treat mouse movement as clicks more 2 times. @tznind do you know what is causing this?

tznind:
Hmn there is various low level logging that can be turned on.  I would guess it is more in Application layer though if it affects v2 drivers also.

Especially since Handled Fixes, it implies its same event going round and round

BDisp:
That what I was think but the wrong click event came from the drivers. Do `Application` is injecting the click event into the driver?

BDisp:
Sorry, the driver has still the previous position when the `Button` was pressed. It isn't a resent event. The mouse must return true at the end of the method, otherwise it will process as a hotkey which isn't the same as a Key event.

BDisp:
The `CommandContext` `struct` could probably be converted to a `class`, thus allowing it to be more volatile and manageable.

tig:
This is ready, I think.

I'm NOT fixing `Menuv2` in this. That will be in another PR.

BDisp:
Do you know why `Trace` item it's always selected at start instead of the activated `Debug` item?

image

tig:
> Do you know why `Trace` item it's always selected at start instead of the activated `Debug` item?
> 

Focus etc... is broken in MenuV2.

tig:
> > Do you know why `Trace` item it's always selected at start instead of the activated `Debug` item?
> 
> Focus etc... is broken in MenuV2.

In my refactor of `OptionSelector` (see `SelectorBase` etc...) I removed the code that was trying to make them look right in MenuV2 because it was not correct. I'm now working on bringing that back in a more correct way.

Terminal.Gui/Views/Menu/MenuBarv2.cs
```diff
@@ -524,8 +524,7 @@ public bool EnableForDesign<TContext> (ref TContext context) where TContext : no
 
         var mutuallyExclusiveOptionsSelector = new OptionSelector
         {
-            Options = ["G_ood", "_Bad", "U_gly"],
-            SelectedItem = 0
+            Labels = ["G_ood", "_Bad", "U_gly"],
```
BDisp, line 547: In the `Dispose` method I think it must be `SuperViewChanged -= OnSuperViewChanged;`.

```cs
    /// <inheritdoc/>
    protected override void Dispose (bool disposing)
    {
        base.Dispose (disposing);

        if (disposing)
        {
            SuperViewChanged += OnSuperViewChanged;
            ConfigurationManager.Applied -= OnConfigurationManagerApplied;
        }
    }

```

tig:
I've converted this back to draft. I have discovered/remembered I have tons more to do here to make this really work.

That said, I finally got `OptionSelector` and `FlagSelector` looking and working mostly correctly when added as a `Shortcut.CommandView`, making `Menuv2` much better. There are still some issues (related to the how `Command.Accepting/Activating` propogate).

WindowsTerminal_d42hjk3ReP

BDisp:
@tig can a `MenuBarv2` having a `MenuItemv2` with an `Action` or is mandatory only using `MenuBarItemv2`?

**Edit:**
Or maybe more appropriate, can a `MenuBarv2` having a `MenuBarItemv2` with an `Action`?

tig:
IIRC my intent was both
What you describe should work. But I doubt I ever tested MenuBarv2 containing a MenuItemv2. 

The end user scenario is a MenuBar item that acts just like a StatusBar item.

BDisp:
Works like a charm. Doesn't make sense to set `Action` in a `MenuBarItemv2` because it will handle popover menus, although it's possible but the Action will be ignored. With `MenuItemv2` works perfectly. Thanks.

BDisp:
@BDisp I moved your comment to the Issue where it belongs: https://github.com/gui-cs/Terminal.Gui/issues/4170

tig:
@copilot can you fix these merge conflicts?

tig:
Examples/UICatalog/Scenarios/SendKeys.cs
Examples/UICatalog/Scenarios/Shortcuts.cs
Terminal.Gui/App/CWP/CWPPropertyHelper.cs
Terminal.Gui/ViewBase/View.Drawing.Attribute.cs
Terminal.Gui/ViewBase/View.Drawing.Scheme.cs
Terminal.Gui/ViewBase/View.Navigation.cs
Terminal.Gui/Views/CharMap/CharMap.cs
Tests/UnitTests/Views/RadioGroupTests.cs
Tests/UnitTestsParallelizable/View/Navigation/AdvanceFocusTests.cs

tig:
Abandoning this. I've backported most of what was here and have a new plan....

PR #4134, opened 2025-06-07
Fixes #4102. Add AOT version of Tests

BDisp:
## Fixes

- Fixes #4102

## Proposed Changes/Todos

- [x] Add `UnitTests.AOT` project
- [ ] Add skip AOT unit tests for `Moq` packages due to not be AOT compatible. Our concern is only to make `Terminal.Gui` AOT compatible

## Pull Request checklist:

- [x] I've named my PR in the form of "Fixes #issue. Terse description."
- [x] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [x] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [x] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [x] My changes generate no new warnings
- [x] I have checked my code and corrected any poor grammar or misspellings
- [x] I conducted basic QA to assure all features are working

PR #4142, opened 2025-06-10
Fixes #4139. Application.Run<T> isn't initializing properly by setting the Application.ForceDriver property

### LINKED ISSUE #4139: Application.Run<T> isn't initializing properly by setting the `Application.ForceDriver` property
As `Run<T>` is the only method used to start the application without needing to explicitly call `Application.Init()` it's imperative that the initialization should properly correctly if we want run an app with a certain driver than the default by setting the `Application.ForceDriver` property.

BDisp:
## Fixes

- Fixes #4139

## Proposed Changes/Todos

- [x] Add WSL: UICatalog --driver v2net to the launchSettings.json
- [x] Checking for v2 drivers in the `InternalInit` method
- [x] Using ForceDriver if the driverName is null
- [x] Ensuring the running the correct app Instance
- [x] Reset _lazyInstance on Shutdown
- [x] Avoiding run console functions when running unit tests

## Pull Request checklist:

### DIFF
--- a/Examples/UICatalog/Properties/launchSettings.json
+++ b/Examples/UICatalog/Properties/launchSettings.json
@@ -32,6 +32,13 @@
       "commandLineArgs": "dotnet UICatalog.dll --driver NetDriver",
       "distributionName": ""
     },
+    "WSL: UICatalog --driver v2net": {
+      "commandName": "Executable",
+      "executablePath": "wsl",
+      "commandLineArgs": "dotnet UICatalog.dll --driver v2net",
+      "distributionName": ""
+    },
+
     "Benchmark All": {
       "commandName": "Project",
       "commandLineArgs": "--benchmark"
--- a/Terminal.Gui/App/Application.Initialization.cs
+++ b/Terminal.Gui/App/Application.Initialization.cs
@@ -77,6 +77,17 @@ public static partial class Application // Initialization (Init/Shutdown)
             throw new InvalidOperationException ("Init has already been called and must be bracketed by Shutdown.");
         }
 
+        var driverNameOrForceDriver = driverName ?? ForceDriver;
+
+        if (driverNameOrForceDriver?.StartsWith ("v2") ?? false)
+        {
+            ApplicationImpl.ChangeInstance (new ApplicationV2 ());
+            ApplicationImpl.Instance.Init (driver, driverNameOrForceDriver);
+            Debug.Assert (Driver is { });
+
+            return;
+        }
+
         if (!calledViaRunT)
         {
             // Reset all class variables (Application is a singleton).
--- a/Terminal.Gui/App/ApplicationImpl.cs
+++ b/Terminal.Gui/App/ApplicationImpl.cs
@@ -34,7 +34,7 @@ public class ApplicationImpl : IApplication
     [RequiresDynamicCode ("AOT")]
     public virtual void Init (IConsoleDriver? driver = null, string? driverName = null)
     {
-        Application.InternalInit (driver, driverName);
+        Application.InternalInit (driver, driverName ?? Application.ForceDriver);
     }
 
     /// <summary>
@@ -85,7 +85,12 @@ public class ApplicationImpl : IApplication
         if (!Application.Initialized)
         {
             // Init() has NOT been called.
-            Application.InternalInit (driver, null, true);
+            Application.InternalInit (driver, Application.ForceDriver, true);
+        }
+
+        if (Instance is ApplicationV2)
+        {
+            return Instance.Run<T> (errorHandler, driver);
         }
 
         var top = new T ();
@@ -227,6 +232,8 @@ public class ApplicationImpl : IApplication
 
             Application.OnInitializedChanged (this, new (in init));
         }
+
+        _lazyInstance = new (() => new ApplicationImpl ());
     }
 
     /// <inheritdoc />
--- a/Terminal.Gui/Drivers/V2/ConsoleDriverFacade.cs
+++ b/Terminal.Gui/Drivers/V2/ConsoleDriverFacade.cs
@@ -64,7 +64,19 @@ internal class ConsoleDriverFacade<T> : IConsoleDriver, IConsoleDriverFacade
     }
 
     /// <summary>Gets the location and size of the terminal screen.</summary>
-    public Rectangle Screen => new (new (0, 0), _output.GetWindowSize ());
+    public Rectangle Screen
+    {
+        get
+        {
+            if (ConsoleDriver.RunningUnitTests)
+            {
+                // In unit tests, we don't have a real output, so we return an empty rectangle.
+                return Rectangle.Empty;
+            }
+
+            return new (new (0, 0), _output.GetWindowSize ());
+        }
+    }
 
     /// <summary>
     ///     Gets or sets the clip rectangle that <see cref="AddRune(Rune)"/> and <see cref="AddStr(string)"/> are subject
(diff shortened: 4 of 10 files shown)


tig:
Did you consider this?

```cs
ConfigurationManager.RuntimeConfig = """
                                     {
                                         "ForceDriver": "v2net"
                                     }
                                     """;

ConfigurationManager.Enable (ConfigLocations.Runtime);
Application.Run<MyWindow>();
```

BDisp:
> Did you consider this?
> 
> ```cs
> ConfigurationManager.RuntimeConfig = """
>                                      {
>                                          "ForceDriver": "v2net"
>                                      }
>                                      """;
> 
> ConfigurationManager.Enable (ConfigLocations.Runtime);
> Application.Run<MyWindow>();
> ```

This is a good solution for a real app but in my case I using a `launchSettings.json` file. Please see my https://github.com/gui-cs/Terminal.Gui/pull/4134#issuecomment-2962243861. Thanks.

tig:
> > Did you consider this?
> > ```cs
> > ConfigurationManager.RuntimeConfig = """
> >                                      {
> >                                          "ForceDriver": "v2net"
> >                                      }
> >                                      """;
> > 
> > ConfigurationManager.Enable (ConfigLocations.Runtime);
> > Application.Run<MyWindow>();
> > ```
> 
> This is a good solution for a real app but in my case I using a `launchSettings.json` file. Please see my [#4134 (comment)](https://github.com/gui-cs/Terminal.Gui/pull/4134#issuecomment-2962243861). Thanks.

If you want to temporarily force all apps on your machine to use a particular driver while testing things, simply put a `config.json` containing the following in your `~/.tui` directory.

```json
{
   "ForceDriver": "v2net"
}
```

Or, you could put this in `.\Terminal.Gui\Examples\Example\bin\Debug\net8.0\.tui\config.json` per-app.

BDisp:
Thanks. The correct configuration is like bellow. With this PR works well but with the `v2_develop` branch I get the error bellow. I'm still working to fix the unit tests failures.

```json
{
   "Application.ForceDriver": "v2net"
}
```

`System.ArgumentException: 'Invalid driver name: v2net. Valid names are CursesDriver, FakeDriver, NetDriver, ConsoleDriverFacade`1, WindowsDriver'`

tig:
> Valid names are

You should be able to fix this easily:

```cs
    public static List<Type?> GetDriverTypes ()
    {
        // use reflection to get the list of drivers
        List<Type?> driverTypes = new ();

        foreach (Assembly asm in AppDomain.CurrentDomain.GetAssemblies ())
        {
            foreach (Type? type in asm.GetTypes ())
            {
                if (typeof (IConsoleDriver).IsAssignableFrom (type) && !type.IsAbstract && type.IsClass)
                {
                    driverTypes.Add (type);
                }
            }
        }

        return driverTypes;
    }
```

tig:
... and if you do fix that, this code can be cleaned up in UI Catalog:

```cs
        Option<string> driverOption = new Option<string> ("--driver", "The IConsoleDriver to use.").FromAmong (
             Application.GetDriverTypes ()
                        .Where (d => !typeof (IConsoleDriverFacade).IsAssignableFrom (d))
                        .Select (d => d!.Name)
                        .Union (["v2", "v2win", "v2net"])
                        .ToArray ()
            );
```

BDisp:
I had already thought the same think but remember the v2 drivers run his own `Init` and `Run` methods. With this PR implementation I don't need to add extra code to the `UICatalog`.

tig:
> ```cs
>                         .Union (["v2", "v2win", "v2net"])
> ```

I think you are overthinking this. 

The bug here is that when we added `v2win/net` we hacked UI Catalog to work around the fact that `GetDriverTypes` didn't support them. The above code is a hack.

If you fix `GetDriverTypes` to correctly return `v2win` and `v2net` then everything will work, and the hack code in UI Catalog can be removed.

Of course, this needs to be done too:

image

BDisp:
What I did meant is that the code that follow the driver check shouldn't be used if the driver is v2 type. The v2 already have it own `Init` and `Run` implementations. Your code is fine but I need to fine tunning and flow the way it will be run. Of course that the hack code in the `UICatalog` can be removed and it will. If I'm wrong I beg sorry.

```cs
        if (Driver is null)
        {
            PlatformID p = Environment.OSVersion.Platform;

            if (string.IsNullOrEmpty (ForceDriver))
            {
                if (p == PlatformID.Win32NT || p == PlatformID.Win32S || p == PlatformID.Win32Windows)
                {
                    Driver = new WindowsDriver ();
                }
                else
                {
                    Driver = new CursesDriver ();
                }
            }
            else
            {
                List<Type?> drivers = GetDriverTypes ();
                Type? driverType = drivers.FirstOrDefault (t => t!.Name.Equals (ForceDriver, StringComparison.InvariantCultureIgnoreCase));

                if (driverType is { })
                {
                    Driver = (IConsoleDriver)Activator.CreateInstance (driverType)!;
                }
                else
                {
                    throw new ArgumentException (
                                                 $"Invalid driver name: {ForceDriver}. Valid names are {string.Join (", ", drivers.Select (t => t!.Name))}"
                                                );
                }
            }
        }

        try
        {
            MainLoop = Driver!.Init ();
            SubscribeDriverEvents ();
        }
        catch (InvalidOperationException ex)
        {
            // This is a case where the driver is unable to initialize the console.
            // This can happen if the console is already in use by another process or
            // if running in unit tests.
            // In this case, we want to throw a more specific exception.
            throw new InvalidOperationException (
                                                 "Unable to initialize the console. This can happen if the console is already in use by another process or in unit tests.",
                                                 ex
                                                );
        }

        SynchronizationContext.SetSynchronizationContext (new MainLoopSyncContext ());

        // TODO: This is probably not needed
        if (Popover.GetActivePopover () is View popover)
        {
            popover.Visible = false;
        }

        MainThreadId = Thread.CurrentThread.ManagedThreadId;
        bool init = Initialized = true;
        InitializedChanged?.Invoke (null, new (init));
    }
```

tig:
I just dove into this a little and I was incorrect on the simplest fix. It's def more complicated than I thought because v2win/net are not fully integrated. 

I'm not sure how to fully fix this, but I suggest you start by making `UICatalog.cs` more correct:

```diff
public class UICatalog
{
-    private static string _forceDriver = string.Empty;
+     private static string? _forceDriver = null;

...

    private static void UICatalogMain (UICatalogCommandLineOptions options)
    {
        // By setting _forceDriver we ensure that if the user has specified a driver on the command line, it will be used
        // regardless of what's in a config file.
-        Application.ForceDriver = (_forceDriver = (string.IsNullOrEmpty(options.Driver) ? null : options.Driver))!;
+        Application.ForceDriver = _forceDriver = options.Driver;
```

Thanks.

BDisp:
Changing the `Application.GetDriverTypes` method doesn't return what you want. We can get only one more type, the generic `ConsoleDriverFacade`1` but not `v2win` or `v2net`. I think that the alternative in the https://github.com/gui-cs/Terminal.Gui/pull/4142/commits/85d35691508a8402f4ee5e6f25a8bada5bf5409e is enough for now.

tig (approved):

tig:
Looks nice. Ready?

BDisp:
> Looks nice. Ready?

Thanks. I think I can do a little better by removing the additional code in the `UICatalog` and change `GetDriverTypes` to return a `List<string?>`. Just a few minutes, please.

BDisp:
I've finished this now. The `UICatalog` only have to call the `GetDriverTypes` without doing any hacking.

PR #4168, opened 2025-06-17
Fixes #4167. Button is always raising Accepting event 2 more times by mouse

### LINKED ISSUE #4167: Button is always raising Accepting event 2 more times by mouse
There is one situation in `Terminal.Gui` which make me crazy. A simple `Accepting` event which will allow the developer to `Cancel` the default behavior (but has the `Handled` property) will repeat the event 3 times by mouse if the developer doesn't Cancel it. `UICatalog` has a bunch of `Button`'s that if they are activated by mouse will repeat the event 3 times, no matter the user press `Key.Enter` or `Key.Esc`. Despite by adding `Handled = true` fix this, there is no necessity of this extra code for a normal `Accepting` event. We can't handle mouse event as key event because only the only the view where the mouse is will be executed and doesn't not propagate to other views and with key event it can be propagate to other views that may want to handle it. I agree the leverage of keybindings, commands and context to achieve this behavior but we need to handle the mouse with much more careful.

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/4126#issuecomment-2976658963_
            
I understand that the current implementation is correct to handle adornments and all the views under the mouse but the `Button` must be responsible to set `e.Handled = true;` after the `MouseClick`event. to prevent further processing to itself.

BDisp:
## Fixes

- Fixes #4167

## Proposed Changes/Todos

- [x] Button must always set `Handled = true` no matter what the user have done to prevent further processing to itself

## Pull Request checklist:

### DIFF
--- a/Examples/UICatalog/Scenarios/Generic.cs
+++ b/Examples/UICatalog/Scenarios/Generic.cs
@@ -27,8 +27,6 @@ public sealed class Generic : Scenario
 
         button.Accepting += (s, e) =>
                             {
-                                // When Accepting is handled, set e.Handled to true to prevent further processing.
-                                e.Handled = true;
                                 MessageBox.ErrorQuery ("Error", "You pressed the button!", "_Ok");
                             };
 
--- a/Terminal.Gui/Views/Button.cs
+++ b/Terminal.Gui/Views/Button.cs
@@ -99,6 +99,7 @@ public class Button : View, IDesignable
 
         return false;
     }
+
     private void Button_MouseClick (object sender, MouseEventArgs e)
     {
         if (e.Handled)
@@ -107,7 +108,8 @@ public class Button : View, IDesignable
         }
 
         // TODO: With https://github.com/gui-cs/Terminal.Gui/issues/3778 we won't have to pass data:
-        e.Handled = InvokeCommand<KeyBinding> (Command.HotKey, new KeyBinding ([Command.HotKey], this, data: null)) == true;
+        InvokeCommand<KeyBinding> (Command.HotKey, new KeyBinding ([Command.HotKey], this, data: null));
+        e.Handled = true;
     }
 
     private void Button_TitleChanged (object sender, EventArgs<string> e)


tig:
I'm willing to bet other Views than button exhibit the same problem. 

Also, button really shouldn't be responding to MouseClick but should use MouseBindings.

I think the real issue is in how event propagation does/doesn't work today (where Accepting is propogated to the SuperView). 

I'm rethinking this as part of the PR I'm working on. 

It might make sense, though, to accept THIS PR as a temporary fix.

BDisp:
I have this PR as draft because I also fixed the `Button` `Accepting` event raising after a `Shortcut` is clicked on a `StatusBar`. But this fix is causing v2 menu issue and I'm attempting to find the cause. I don't know if MouseBindings is a good solution but no matter what solution is used it can't be propagated to the SuperView because the mouse hit a specific location contrary to keyboard.

tig:
Please don't waste your time on this. 

I'm literally working on the same topic.

BDisp:
Reopen this again because I need their changes.

BDisp:
This code in the `View.Command.cs` it's only useful for menus and is buggy for others situations.

```cs
            if (SuperView is { })
            {
                Logging.Debug ($"{Title} ({ctx?.Source?.Title}) - Invoking Accept on SuperView ({SuperView.Title}/{SuperView.Id})...");
                return SuperView?.InvokeCommand (Command.Accept, ctx);
            }
```

tig:
Accept propagation is not just for menus. But any nested view scenario including statusbar and selectors. 

Nested view scenarios need a standard way for events to propagate up the
Hierarchy. The current implementation for Accept is problematic for sure. But it IS needed. As is a similar solution for select/activate. 

As I said before, several times, I'm literally working on fixing this all right now. 

I'm sorry that you are encountering problems but you also trying to fix the underlying issues is not helpful.

BDisp:
> I'm sorry that you are encountering problems but you also trying to fix the underlying issues is not helpful.

Sorry about any problem I'm causing but doesn't my intention. I'm only also trying to help find some correctness for this and I don't want you think I'm making something difficult for you. My intention is giving some more suggestions to help fix this.
Also making two consecutive cancelable events isn't the best choice, unless for some specific case.

Currently is:
```
Selecting -> Cancelable
Accepting -> Cancelable
```

A developer that only want to execute code when it was accepted has a real difficult to handle it unless he set `Handled = true`.

Recommended solution for the current situation is:
```
Selecting -> Cancelable
Accepted -> Not Cancelable
```

But since we need to differentiate a selection from acceptation one solution is:
For selection (using navigation keys or mouse pressed or click):
```
Selecting -> Cancelable
Selected -> Not Cancelable

```
For acceptation (using enter, space or mouse click or double click):
```
Accepting -> Cancelable
Accepted -> Not Cancelable
```

Please do not take these suggestions lightly as they are only meant to help.

tig:
Good input. I do like the idea of making both have a cancelable and non-cancelable.

tig:
Part of why I get frustrated is you doing things like reopening this issue with a comment that is disjointed and incomplete. 

It is confusing to me because I can only assume you are doing it for a good reason. But I can't infer what the reason is. Then, it turns out it's precisely what I'm thinking about in elsewhere (the Select-Activate PR). 

It would have better if you commented on the Select-Accept PR.

BDisp:
Sorry for I really didn't been more explicit when I reopen this. My intention was adding a better provisory fix for this issue until your PR is finished, which could be merge before yours. I almost done it, but I'm only struggling with the hot keys (alt+key), but still working in it. Wasn't any barely intentional.

tig:
> Sorry for I really didn't been more explicit when I reopen this. My intention was adding a better provisory fix for this issue until your PR is finished, which could be merge before yours. I almost done it, but I'm only struggling with the hot keys (alt+key), but still working in it. Wasn't any barely intentional.

What are you working on that requires a fix for this (vs a work around or just not having it work exactly right)?

BDisp:
> What are you working on that requires a fix for this (vs a work around or just not having it work exactly right)?

Not working properly is always a constant, otherwise we wouldn't need fixes. I'm just trying to make it work a little better than it already is. Then you can still improve it a lot later and possibly, if my fix is ​​successful, it can help you solve it even better. I'm trying not to make too much of a mess and using some rational logic.

BDisp:
I think this PR is ready for review. I advise you to use the modified code from the Generic scenario in the v2_develop branch to understand the problems that are causing and that are not reproducible in a unit test. You will notice that clicking the button will open the MessageBox three times instead of once. Clicking the "No action" shortcut will open the MessageBox for the button. Also clicking the "With action" shortcut will also open the MessageBox for the button before opening the MessageBox for the shortcut, the same happening if you press F5.

BDisp:
In this PR it's now possible to close an opened menu by clicking again on it.

PR #4172, opened 2025-06-19
Fixes #4171 - remove all idles and replace with zero timeouts

### LINKED ISSUE #4171: Idles are just timeouts with Timespan.Zero
We draw a distinction between timeouts and idles but in practice they both run once per iteration.

An Idle is just a timeout with TimeSpan.Zero.

All the interface distinctions around lamdas and add/remove idle can be vastly simplified by just replacing with:

```csharp
        ml.TimedEvents.AddTimeout (TimeSpan.Zero, someFunc);
```

I've started a PR that seems to suggest this is feasible to simplify `ITimedEvents` considerably.  This is required first step because I want to add the ability to have more exciting time systems e.g. also Logarithmic (for #4123 ).  `Timeout` doesn't explicitly always have to be the same time over and over.

tznind:
## Fixes

- Fixes #4171 

## Proposed Changes/Todos

- [x] Replace idles with zero duration timeouts.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/App/Application.Screen.cs
+++ b/Terminal.Gui/App/Application.Screen.cs
@@ -4,6 +4,7 @@ namespace Terminal.Gui.App;
 
 public static partial class Application // Screen related stuff
 {
+    private static readonly object _lock = new ();
     private static Rectangle? _screen;
 
     /// <summary>
@@ -18,11 +19,15 @@ public static partial class Application // Screen related stuff
     {
         get
         {
-            if (_screen == null)
+            lock (_lock)
             {
-                _screen = Driver?.Screen ?? new (new (0, 0), new (2048, 2048));
+                if (_screen == null)
+                {
+                    _screen = Driver?.Screen ?? new (new (0, 0), new (2048, 2048));
+                }
+
+                return _screen.Value;
             }
-            return _screen.Value;
         }
         set
         {
@@ -30,7 +35,11 @@ public static partial class Application // Screen related stuff
             {
                 throw new NotImplementedException ($"Screen locations other than 0, 0 are not yet supported");
             }
-            _screen = value;
+
+            lock (_lock)
+            {
+                _screen = value;
+            }
         }
     }
 
--- a/Tests/UnitTests/Application/MainLoopTests.cs
+++ b/Tests/UnitTests/Application/MainLoopTests.cs
@@ -113,9 +113,11 @@ public class MainLoopTests
 
         Assert.True (ml.TimedEvents.RemoveTimeout(a));
         Assert.False (ml.TimedEvents.RemoveTimeout (a));
-        Assert.True (ml.TimedEvents.RemoveTimeout (b));
 
-        Assert.Equal (3, functionCalled);
+        // Cannot remove b because it returned false i.e. auto removes itself
+        Assert.False (ml.TimedEvents.RemoveTimeout (b));
+
+        Assert.Equal (1, functionCalled);
     }
 
     [Fact]
@@ -483,7 +485,7 @@ public class MainLoopTests
         ml.TimedEvents.AddTimeout (TimeSpan.Zero, fnTrue);
         bool retVal = ml.TimedEvents.CheckTimers(out int waitTimeOut);
         Assert.True (retVal);
-        Assert.Equal (-1, waitTimeOut);
+        Assert.Equal (0, waitTimeOut);
     }
 
     [Fact]
@@ -576,7 +578,7 @@ public class MainLoopTests
                         new App.Timeout { Span = new (), Callback = () => true }
                        );
     }
-
+    /*
     [Theory]
     [MemberData (nameof (TestAddTimeout))]
     public void Mainloop_Invoke_Or_AddTimeout_Can_Be_Used_For_Events_Or_Actions (
@@ -659,7 +661,7 @@ public class MainLoopTests
 
         Application.Shutdown ();
     }
-
+    */
     [Fact]
     public void RemoveIdle_Function_NotCalled ()
     {
--- a/Tests/UnitTests/Application/MainLoopTests.cs
+++ b/Tests/UnitTests/Application/MainLoopTests.cs
@@ -29,19 +29,19 @@ public class MainLoopTests
 
     // Idle Handler tests
     [Fact]
-    public void AddIdle_Adds_And_Removes ()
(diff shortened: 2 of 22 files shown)


tznind:
Only test that fails is `Mainloop_Invoke_Or_AddIdle_Can_Be_Used_For_Events_Or_Actions` 

Its very complicated and hard to understand test - I've commented it out for now in the hopes that if there is a real issue here we can make a simpler test that exhibits it.

BDisp:
The Post method is normally never run in the UIThread and thus the necessity of adding to the idle field which call the wakeup method that will run in the UIThread and execute the idle.

tznind:
Yes this is also how timeouts work. They also trigger wakeup and run in ui thread

Tests/UnitTests/Application/MainLoopTests.cs
```diff
@@ -130,19 +107,21 @@ public void AddIdle_Twice_Returns_False_Called_Twice ()
...
-        Assert.True (ml.TimedEvents.RemoveIdle (fnStop));
-        Assert.False (ml.TimedEvents.RemoveIdle (fn1));
-        Assert.False (ml.TimedEvents.RemoveIdle (fn1));
 
-        Assert.Equal (2, functionCalled);
+        Assert.True (ml.TimedEvents.RemoveTimeout(a));
+        Assert.False (ml.TimedEvents.RemoveTimeout (a));
+
+        // Cannot remove b because it returned false i.e. auto removes itself
+        Assert.False (ml.TimedEvents.RemoveTimeout (b));
```
BDisp, line 118: I think this is the cause why the `Mainloop_Invoke_Or_AddTimeout_Can_Be_Used_For_Events_Or_Actions` unit test is failing. Maybe the pending idle function isn't executed because it was already removed without call `RemoveTimeout`? If it's the case it's needed to ensure that this timed event come from `Post` method should be run no matter the timed event is elapsed time is expired.

BDisp: Only the `Action a1 = StartWindow;` ran successfully the unit test. The `Action a2 = () => Application.Invoke (StartWindow);` not allowing to be run asynchronous. Probably it isn't being called properly or AddTimeout doesn't allow run non-UIThreads. The `SynchronizationContext.Post` method isn't hit.

```cs
public static IEnumerable<object []> TestAddTimeout
    {
        get
        {
            // Goes fine
            Action a1 = StartWindow;

            yield return new object [] { a1, "Click Me", "Cancel", "Pew Pew", 0, 1, 2, 3, 4 };

            // Also goes fine
            Action a2 = () => Application.Invoke (StartWindow);

            yield return new object [] { a2, "Click Me", "Cancel", "Pew Pew", 0, 1, 2, 3, 4 };
        }
    }
```

BDisp:
> Yes this is also how timeouts work. They also trigger wakeup and run in ui thread

But the problem is when a timeout run some code that will call an asynchronous code the `SynchronizationContext.Post` method isn't hit and thus no more timeout are added. That's the reason why idle handlers and timeouts are separated.

tznind:
Are you able to create a minimum repro?

> when a timeout run some code that will call an asynchronous code the SynchronizationContext.Post method isn't hit 

Surely this can be fixed by making it so that asynchronous code in triggered code calls Post? What bit of code does that for idles - I don't see it.

tznind:
Ok think I have fixed it, I just updated Invoke to resolve lamda if already on UI thread up front.

BDisp:
Comment this lines in the unit test:

```cs
// Goes fine
            Action a1 = StartWindow;

            yield return new object [] { a1, "Click Me", "Cancel", "Pew Pew", 0, 1, 2, 3, 4 };
```

and run only with this lines:

```cs
// Also goes fine
            Action a2 = () => Application.Invoke (StartWindow);

            yield return new object [] { a2, "Click Me", "Cancel", "Pew Pew", 0, 1, 2, 3, 4 };
        }
```

You will see that the `Post` method isn't hit.

tznind:
Ok they all passing now... do you still think there is a problem? can you create a scenario or test that fails on this branch but works on v2_develop?

One thing that I think we should agree is that, if there is some threading/continuation thing that works with Idles, it should also work with Timeouts.  So if it didn't in v2_develop then that needed fixed here in the merge of the two.

BDisp:
I didn't test again with the latest changes. I you already could run it without fail, that great and I don't have knowledge of other issues. Thanks.

BDisp:
I already tested and all is running well. Great job. Thanks.

tznind:
Looks like the parallel test sometimes fails in `Application.Screen` access.

I have added a lock in c588e04912a29b8a146c2b885b0e4dd19133ab5a which should help

tig:
> Looks like the parallel test sometimes fails in `Application.Screen` access.
> 
> 
> 
> I have added a lock in c588e04912a29b8a146c2b885b0e4dd19133ab5a which should help

Unless there's a way to make Application.Screen not a static (eg the v2 Instance stuff), any parallel test that relies on it being set needs to move to non-parallel.

Terminal.Gui/App/Application.Screen.cs
```diff
@@ -4,6 +4,7 @@ namespace Terminal.Gui.App;
 
 public static partial class Application // Screen related stuff
 {
+    private static readonly object _lock = new ();
```
tig, line 7: Please name this such that it's clear it is a lock for screen.

Terminal.Gui/App/Application.Screen.cs
```diff
@@ -18,19 +19,27 @@ public static Rectangle Screen
...
             }
-            return _screen.Value;
         }
         set
         {
             if (value is {} && (value.X != 0 || value.Y != 0))
             {
                 throw new NotImplementedException ($"Screen locations other than 0, 0 are not yet supported");
             }
-            _screen = value;
```
tig, line 33: Another thing we can do here is check IsInitialized. Throw an exception if false. 

This will find the parallel test that is setting it.

tig (changes requested):

tig:
Also, will you please update the multitasking.md doc to reflect these changes?

tznind:
I have done both of these changes in https://github.com/gui-cs/Terminal.Gui/pull/4173

Going to close this PR as I'm now working on #4173 which is super set of both my recent changes PRs.

Issue #4043, opened 2025-04-24
Even in release mode huge dependencies used ( Microsoft.CodeAnalysis.*)


ChristopheI:
**Describe the bug**
There is a lot of huge dependencies used: 
Microsoft.CodeAnalysis.CSharp.dll
Microsoft.CodeAnalysis.VisualBasic.dll
Microsoft.CodeAnalysis.Workspaces.dll
~15 Mo with these 3 packages

There is also:
Microsoft.CodeAnalysis.CSharp.Workspaces.dll
Microsoft.CodeAnalysis.dll
Microsoft.CodeAnalysis.VisualBasic.Workspaces.dll
More than 4 Mo with these last 3 packages

**To Reproduce**
Compile UICatalog example

**Expected behavior**
Are they necessaries ?

**Desktop :**
 - OS: Windows 11
 - Using Terminal.Gui v2.0.0-develop.4407

**Additional context**
Just before I used Terminal.Gui v2.0.0-v2-develop.679
And there was no dependency at all to any Microsoft.CodeAnalysis.* package ...

ChristopheI:
I just tested to use only these dependencies:
    
    
    
    <ItemGroup>
        <PackageReference Include="ColorHelper" />
        <PackageReference Include="JetBrains.Annotations" />
        <PackageReference Include="Microsoft.Extensions.Logging" />
        <PackageReference Include="System.IO.Abstractions" />
        <PackageReference Include="System.Text.Json" />
        <PackageReference Include="Wcwidth" />
        
        <PackageReference Include="Microsoft.SourceLink.GitHub" />
    </ItemGroup>

And the compilation is working fine (i.e. the project Termnai.Gui) in Debug and in Release

Original dependencies set:
    
    
    
    <ItemGroup>
        <PackageReference Include="ColorHelper" />
        <PackageReference Include="JetBrains.Annotations" />
        <PackageReference Include="Microsoft.CodeAnalysis" />
        <PackageReference Include="Microsoft.CodeAnalysis.Common" />
        <PackageReference Include="Microsoft.CodeAnalysis.CSharp" />
        <PackageReference Include="Microsoft.Extensions.Logging" />
        <PackageReference Include="System.IO.Abstractions" />
        <PackageReference Include="System.Text.Json" />
        <PackageReference Include="Wcwidth" />
        
        <PackageReference Include="Microsoft.SourceLink.GitHub" />
    </ItemGroup>

tig:
[...]

PR #4178, opened 2025-06-25
Cleanup Dependencies slighty

danielmarbach:
## Fixes

- Fixes dependencies in the main library

## Proposed Changes/Todos

- SourceLink should be used with PrivateAssets
- The System.Text.Json reference is unnecessary since NET8 has that included and nothing specific from v9 or higher is used
- Logging abstractions should be used instead of logging since nothing in the code depends directly on specific logging things
- Code Analysis references are also not needed in the main lib

I think overall the dependencies and the management of the dependencies needs a bit of an overhaul. For example mixing in the central package management the shippable references with ranges together with all others makes it difficult to follow what is needed and why. It might even be possible to even further improve things, but I ran out of time.

## Pull Request checklist:

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -12,7 +12,7 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="[4.11,4.12)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.11.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="[9.0.2,10)" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="[9.0.2,10)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="9.0.6" />
     <PackageVersion Include="System.IO.Abstractions" Version="[22.0.11,23)" />
     <PackageVersion Include="Wcwidth" Version="[2,3)" />
     <PackageVersion Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="[1.21.2,2)" />
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -13,7 +13,6 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.11.0" />
     <PackageVersion Include="Microsoft.Extensions.Logging" Version="[9.0.2,10)" />
     <PackageVersion Include="System.IO.Abstractions" Version="[22.0.11,23)" />
-    <PackageVersion Include="System.Text.Json" Version="[8.0.5,9)" />
     <PackageVersion Include="Wcwidth" Version="[2,3)" />
     <PackageVersion Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="[1.21.2,2)" />
     <PackageVersion Include="Serilog" Version="4.2.0" />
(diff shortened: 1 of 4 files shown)


ChristopheI:
Good point ! It will close this https://github.com/gui-cs/Terminal.Gui/issues/4043

Directory.Packages.props
```diff
@@ -11,9 +11,9 @@
     <PackageVersion Include="Microsoft.CodeAnalysis" Version="[4.11,4.12)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="[4.11,4.12)" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.11.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging" Version="[9.0.2,10)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="[9.0.2,10)" />
+    <PackageVersion Include="Microsoft.Extensions.Logging" Version="9.0.6" />
```
BDisp, line 15: This can't be like below?
```csproj
<PackageVersion Include="Microsoft.Extensions.Logging" Version="[9.0.6,10)" />

danielmarbach:
FYI when using net 8 SDK to build apparently source link is now built in and no longer necessary but I guess this can be done in a follow up PR

Directory.Packages.props
danielmarbach: It could but as I mentioned in the description I do not understand your package versioning strategy. Normally floating versions only make sense for prod dependencies and for tests etc it would make sense to pin versions and use something like dependabot or renovate to update those. Otherwise your ground moves potentially on every build

BDisp: Actually we are using this strategy of updating nuget packages without dependabot dependence. In this case, it's instructing to use the latest nuget package from the version 9.0.6 until the release version 10.

danielmarbach: Whatever works for you 👍
Feel free to push the change to this branch. It has edit rights

BDisp: I prefer to wait for @tig decision.

tig: I'm not expert enough in this to make a defniitive call. I think in principle, we want to automatically get the latest versions of dependencies that don't have known breaking changes. I think this means 9.0.x is ok, but 9.1.x is not.

danielmarbach: @tig That's unfortunately not how this works. The version range of `[9.0.2,10)` this will most likely resolve to lowest unless something is forcing a higher version. It means if nothing is forcing, for example a 9.1 it will resolve to 9.0.2. It sounds like you want "floating" versions, which would require wildcards in the version numbers, like `9.0.*`.

I wonder why you are trying to avoid something like Dependabot or renovate. The tradeoffs are non-deterministic builds. Overall, it might also make sense to evaluate what dependencies are prod, what are SemVer compatible and unlikely to break consumers of your lib etc and then cleanup things accordingly.

That being said, I think this is outside the scope of this PR. For us, what matters is to get rid of unnecessary dependencies to make our build artifacts of what we ship leaner.

danielmarbach: I slightly amended my above statement since my friend @bording rightfully pointed out I confused things 🫤

BDisp: > It sounds like you want "floating" versions, which would require wildcards in the version numbers, like `9.0.*`.

I think wildcards will not be the best solution for the lowest version because sometimes it needed for force to a specific lowest version, e.g. `9.0.*` probably will use the `9.0.0` or `9.0.1` when we really want to use equal or grater than `9.0.2` version. But I agree that this subject is outside of the scope of this PR.

PR #4193, opened 2025-07-15
Fixes 4191 - Consider width 2 chars that are not IsBmp

### LINKED ISSUE #4188: Windows drivers are forcing 16 color mode by default on some terminals
**Describe the bug**

Initially raised in https://github.com/gui-cs/Terminal.Gui/discussions/4187

Using the color:

```csharp
SetAttribute (new Attribute (new Color (232, 112, 23), new Color (112, 255, 112)));
```

Results in wrong colors for some terminals:

Image

Here are the actual colors rendered:

|     Terminal     | Foreground | Background  |
|:----------------:|:----------:|-------------|
| Powershell       | 255,0,0    | 0,255,255   |
| Command Prompt   | 231,72,86  | 97,214,214  |
| Debug Console    | 232,112,23 | 112,255,112 |
| Windows Terminal | 232,112,23 | 112,255,112 |

**To Reproduce**

```csharp
    private class MyView : View
    {
        /// <inheritdoc />
        protected override bool OnDrawingContent (DrawContext? context)
        {
            SetAttribute (new Attribute (new Color (232, 112, 23), new Color (112, 255, 112)));

            Move (0, 0);
            AddStr ("Hello");

            return true;
        }
    }
    private static int Main (string [] args)
    {
        Application.Init (null, "v2");
        var w = new Window ();

        w.Add (
               new MyView ()
               {
                   Width = 10,
                   Height = 10,
               });
        Application.Run (w);
        Application.Shutdown ();
        return 0;
```

**Expected behavior**

**Screenshots**

**Additional context**

**Set Project & Milestone**
If you have access, please don't forget to set the right Project and Milestone.

### LINKED ISSUE #4191: CJK Characters doesn't render correctly on Alacritty (v2_develop)
**Describe the bug**
As title, but this works quite well in v1 on Alacritty 

This is a continuation of https://github.com/gui-cs/Terminal.Gui/issues/2928

**To Reproduce**
1. Use Alacritty
2. In the directory, `dotnet run --project Examples/UICatalog/UICatalog.csproj` (`dotnet run --project UICatalog/UICatalog.csproj` on v1)

**Expected behavior**
Runes should be rendered correctly

**Screenshots**
Alacritty, `v2_develop` and `v2_release` (renders incorrectly):

Image

Windows Terminal, `v2_develop` and `v2_release` (seems normal):

Image

Alacritty, `v1_release` (seems normal):

Image

Windows Terminal, `v1_release` (renders incorrectly):

Image

**Additional context**
As per https://github.com/microsoft/terminal/issues/17016 and https://github.com/microsoft/terminal/pull/16825 , I can't 100% sure whether this is a Terminal.Gui problem or the corresponding terminal emulator's.

tznind:
## Fixes

- Fixes #4191
- Fixes #4188

## Proposed Changes/Todos

We now assume SupportsTrueColor is `true` unless user explicitly sets `false` themselves.

Not all width 2 characters are `Rune.IsBmp`.  The old code only ever considered `rune.GetColumns()` in the `IsBmp` pathway.

Fix is applied only to v2win

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/V2/WindowsOutput.cs
+++ b/Terminal.Gui/Drivers/V2/WindowsOutput.cs
@@ -146,7 +146,7 @@ internal partial class WindowsOutput : IConsoleOutput
 
                 outputBuffer [position].Empty = false;
 
-                var rune = buffer.Contents [row, col].Rune;
+                Rune rune = buffer.Contents [row, col].Rune;
                 int width = rune.GetColumns ();
 
                 if (rune.IsBmp)
--- a/Terminal.Gui/Drivers/V2/WindowsOutput.cs
+++ b/Terminal.Gui/Drivers/V2/WindowsOutput.cs
@@ -146,16 +146,37 @@ internal partial class WindowsOutput : IConsoleOutput
 
                 outputBuffer [position].Empty = false;
 
-                if (buffer.Contents [row, col].Rune.IsBmp)
+                var rune = buffer.Contents [row, col].Rune;
+                int width = rune.GetColumns ();
+
+                if (rune.IsBmp)
                 {
-                    outputBuffer [position].Char = (char)buffer.Contents [row, col].Rune.Value;
+                    if (width == 1)
+                    {
+                        // Single-width char, just encode first UTF-16 char
+                        outputBuffer [position].Char = (char)rune.Value;
+                    }
+                    else if (width == 2 && col + 1 < buffer.Cols)
+                    {
+                        // Double-width char: encode to UTF-16 surrogate pair and write both halves
+                        var utf16 = new char [2];
+                        rune.EncodeToUtf16 (utf16);
+                        outputBuffer [position].Char = utf16 [0];
+                        outputBuffer [position].Empty = false;
+
+                        // Write second half into next cell
+                        col++;
+                        position = row * buffer.Cols + col;
+                        outputBuffer [position].Char = utf16 [1];
+                        outputBuffer [position].Empty = false;
+                    }
                 }
                 else
                 {
                     //outputBuffer [position].Empty = true;
                     outputBuffer [position].Char = (char)Rune.ReplacementChar.Value;
 
-                    if (buffer.Contents [row, col].Rune.GetColumns () > 1 && col + 1 < buffer.Cols)
+                    if (width > 1 && col + 1 < buffer.Cols)
                     {
                         // TODO: This is a hack to deal with non-BMP and wide characters.
                         col++;
--- a/Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
+++ b/Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
@@ -344,37 +344,16 @@ internal class WindowsDriver : ConsoleDriver
 
                 _outputBuffer [position].Empty = false;
 
-                Rune rune = Contents [row, col].Rune;
-                int width = rune.GetColumns ();
-
-                if (rune.IsBmp)
+                if (Contents [row, col].Rune.IsBmp)
                 {
-                    if (width == 1)
-                    {
-                        // Single-width char, just encode first UTF-16 char
-                        _outputBuffer [position].Char = (char)rune.Value;
-                    }
-                    else if (width == 2 && col + 1 < Cols)
-                    {
-                        // Double-width char: encode to UTF-16 surrogate pair and write both halves
-                        var utf16 = new char [2];
-                        rune.EncodeToUtf16 (utf16);
-                        _outputBuffer [position].Char = utf16 [0];
-                        _outputBuffer [position].Empty = false;
-
-                        // Write second half into next cell
-                        col++;
-                        position = row * Cols + col;
-                        _outputBuffer [position].Char = utf16 [1];
-                        _outputBuffer [position].Empty = false;
-                    }
+                    _outputBuffer [position].Char = (char)Contents [row, col].Rune.Value;
                 }
                 else
                 {
                     //_outputBuffer [position].Empty = true;
                     _outputBuffer [position].Char = (char)Rune.ReplacementChar.Value;
 
-                    if (width > 1 && col + 1 < Cols)
+                    if (Contents [row, col].Rune.GetColumns () > 1 && col + 1 < Cols)
                     {
                         // TODO: This is a hack to deal with non-BMP and wide characters.
                         col++;
--- a/Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
+++ b/Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
@@ -344,16 +344,37 @@ internal class WindowsDriver : ConsoleDriver
 
                 _outputBuffer [position].Empty = false;
 
-                if (Contents [row, col].Rune.IsBmp)
+                var rune = Contents [row, col].Rune;
+                int width = rune.GetColumns ();
+
+                if (rune.IsBmp)
                 {
-                    _outputBuffer [position].Char = (char)Contents [row, col].Rune.Value;
+                    if (width == 1)
+                    {
+                        // Single-width char, just encode first UTF-16 char
+                        _outputBuffer [position].Char = (char)rune.Value;
+                    }
+                    else if (width == 2 && col + 1 < Cols)
+                    {
+                        // Double-width char: encode to UTF-16 surrogate pair and write both halves
+                        var utf16 = new char [2];
+                        rune.EncodeToUtf16 (utf16);
+                        _outputBuffer [position].Char = utf16 [0];
+                        _outputBuffer [position].Empty = false;
+
+                        // Write second half into next cell
+                        col++;
+                        position = row * Cols + col;
+                        _outputBuffer [position].Char = utf16 [1];
+                        _outputBuffer [position].Empty = false;
+                    }
                 }
                 else
                 {
                     //_outputBuffer [position].Empty = true;
                     _outputBuffer [position].Char = (char)Rune.ReplacementChar.Value;
 
-                    if (Contents [row, col].Rune.GetColumns () > 1 && col + 1 < Cols)
+                    if (width > 1 && col + 1 < Cols)
                     {
                         // TODO: This is a hack to deal with non-BMP and wide characters.
                         col++;
--- a/Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
+++ b/Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
@@ -344,7 +344,7 @@ internal class WindowsDriver : ConsoleDriver
 
                 _outputBuffer [position].Empty = false;
 
-                var rune = Contents [row, col].Rune;
+                Rune rune = Contents [row, col].Rune;
                 int width = rune.GetColumns ();
 
                 if (rune.IsBmp)
--- a/Terminal.Gui/Drivers/V2/MainLoopCoordinator.cs
+++ b/Terminal.Gui/Drivers/V2/MainLoopCoordinator.cs
@@ -25,7 +25,6 @@ internal class MainLoopCoordinator<T> : IMainLoopCoordinator
     private ConsoleDriverFacade<T> _facade;
     private Task _inputTask;
     private readonly ITimedEvents _timedEvents;
-    private readonly bool _isWindowsTerminal;
 
     private readonly SemaphoreSlim _startupSemaphore = new (0, 1);
 
@@ -61,7 +60,6 @@ internal class MainLoopCoordinator<T> : IMainLoopCoordinator
         _inputProcessor = inputProcessor;
         _outputFactory = outputFactory;
         _loop = loop;
-        _isWindowsTerminal = Environment.GetEnvironmentVariable ("WT_SESSION") is { } || Environment.GetEnvironmentVariable ("VSAPPIDNAME") != null;
     }
 
     /// <summary>
@@ -161,11 +159,6 @@ internal class MainLoopCoordinator<T> : IMainLoopCoordinator
                            _output,
                            _loop.AnsiRequestScheduler,
                            _loop.WindowSizeMonitor);
-
-            if (!_isWindowsTerminal)
-            {
-                Application.Force16Colors = _facade.Force16Colors = true;
-            }
 
             Application.Driver = _facade;
 


Terminal.Gui/Drivers/V2/WindowsOutput.cs
```diff
@@ -146,16 +146,37 @@ public void Write (IOutputBuffer buffer)
...
                 {
-                    outputBuffer [position].Char = (char)buffer.Contents [row, col].Rune.Value;
+                    if (width == 1)
+                    {
+                        // Single-width char, just encode first UTF-16 char
+                        outputBuffer [position].Char = (char)rune.Value;
+                    }
+                    else if (width == 2 && col + 1 < buffer.Cols)
+                    {
+                        // Double-width char: encode to UTF-16 surrogate pair and write both halves
```
BDisp: If `IsBmp` is `true` are considered Basic Multilingual Plane (BMP) and they aren't composed by surrogate pairs. Some BMP may have a width of 2 columns. If `IsBmp` is `false` (non-Bmp) then it's considered as surrogate pairs which some may have a width of 2 columns and some may have a width of 1 column.

Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
```diff
@@ -344,16 +344,37 @@ public override bool UpdateScreen ()
...
                 {
-                    _outputBuffer [position].Char = (char)Contents [row, col].Rune.Value;
+                    if (width == 1)
+                    {
+                        // Single-width char, just encode first UTF-16 char
+                        _outputBuffer [position].Char = (char)rune.Value;
+                    }
+                    else if (width == 2 && col + 1 < Cols)
+                    {
+                        // Double-width char: encode to UTF-16 surrogate pair and write both halves
```
BDisp: If `IsBmp` is `true` are considered Basic Multilingual Plane (BMP) and they aren't composed by surrogate pairs. Some BMP may have a width of 2 columns. If `IsBmp` is `false` (non-Bmp) then it's considered as surrogate pairs which some may have a width of 2 columns and some may have a width of 1 column.

Terminal.Gui/Drivers/V2/WindowsOutput.cs
tznind: are you suggesting new comment wording? or that code is wrong?

BDisp: The comment and the code.

In the v2_develop using cmd.exe:

image

In this PR branch using cmd.exe:

image

Don't use Windows Terminal and VSConsole because they already has virtual terminal enabled and you won't see the issue. Use cmd.exe or conhost.exe.
The Unicode Character “你” (U+4F60) is a char with a width of 2 columns but isn't a surrogate pair nor an array of chars, because `"你".Length` is equal to 1 which mean it's only a single char that occupies 2 columns.

BDisp: Also the second item from the array returned by `rune.EncodeToUtf16 (utf16)` is a null char because it's initialized with 2 elements but the char '你' has only 1 element.

tznind: How does v2net look for you on this setup?

Its clear the current code isn't right, because you can see there is that diamond question mark showing second cell of the Rune has bad data in it on v2_develop.

This issue does not affect NetDriver/v2net

BDisp: > How does v2net look for you on this setup?
> 
It looks great using v2net:

image

> Its clear the current code isn't right, because you can see there is that diamond question mark showing second cell of the Rune has bad data in it on v2_develop.
> 
The question mark is a replacement char added in the `AddRune` method after a wide code point. Each driver will handle it accordingly to their default. In my opinion it should be a null char '\0', which Win API normally ignore it. But also a replacement char is needed to indicate thar a wide code point doesn't feet in the last column, where it's already handled in the `AddRune` method.

> This issue does not affect NetDriver/v2net

This issue doesn't affect NetDriver/v2net because the runes are handled as string. When a wide code point is being handled the driver already has the right encoded chars and skip the next rune. With Win API drivers we need to pass each encoded char to each cell, for output the wide code point. Here, happens 2 different situations, First situation is if the wide code point is really a char, where it must be pass to the Win API drivers, follow by a null char '\0' which will be ignored. The second situation is if the wide codepoint is a surrogate pair, where each char must be pass to the Win API drivers in each cell. The real problem, is with surrogate pairs that weren't a wide code point but will occupy 2 cells, because each char should be pass to each cell, not fitting all the chars in these columns. In this case, there is needed to rethink about a way to workaround this.

Terminal.Gui/Drivers/WindowsDriver/WindowsDriver.cs
tznind: I will make changes only to v2 so resolving this one.

Terminal.Gui/Drivers/V2/WindowsOutput.cs
tznind: The problem with cmd.exe comes down to 'force16' colors code pathway.

```diff
    public bool WriteToConsole (Size size, WindowsConsole.ExtendedCharInfo [] charInfoBuffer, WindowsConsole.Coord bufferSize, WindowsConsole.SmallRect window, bool force16Colors)
    {

        //Debug.WriteLine ("WriteToConsole");

        //if (_screenBuffer == nint.Zero)
        //{
        //    ReadFromConsoleOutput (size, bufferSize, ref window);
        //}

        var result = false;

-        if (force16Colors)
+       if (false)
```

When I make this change on my fixed PR here for v2win then all terminals render correctly:

image
_cmd.exe (top left) using font NSimSun with Alacritty (top right) and Windows Terminal (below)_

tznind:
Ok looks like only change needed really is to get rid of the force 16 colors.

I've created PR for that here https://github.com/gui-cs/Terminal.Gui/pull/4195

tznind:
Seems the changes are a mixed bag.  My change fixes Alacritty for 16 color mode but as you said above @BDisp it results in 'gap' in conhost.

I think for now we can leave it with the changes in #4195 which is just to stop using `WriteConsoleOutputW` (16 color mode) entirely.  The number of people who will toggle 16 color mode on is going to be low.  So intersection of wide characters + 16 color mode users is miniscule.  

And it looks like different terminals are dealing with the lack of proper surrogate pair support in that API call differently.

tznind:
@BDisp this can be tested now

BDisp:
Still the same issue with v2win.

With Windows Terminal:
devenv_VmpMdXOqtJ

With Alacritty it's resizing correctly, but at exit it isn't restoring the original screen buffer correctly like Windows Terminal as well too. When VTS (Virtual Terminal Sequences) are enabled, set by the Terminal.Gui or by the terminal itself, it's needed to send escape sequences to enable alternate screen buffer (EscSeqUtils.CSI_SaveCursorAndActivateAltBufferNoBackscroll) at driver initialization and to disable it (EscSeqUtils.CSI_RestoreCursorAndRestoreAltBufferWithBackscroll) at the end. The best way to check if a terminal has VTS enabled in Windows is by querying his flag in console modes.
With cmd.exe and conhost.exe it's impossible to run it even with Force16Colors.
Another situation is when it's needed to paste surrogate pairs into a TextView or TextField, where a `System.ArgumentOutOfRangeException: 'Specified argument was out of the range of valid values. (Parameter 'value')'`will throws in the `Key` class.
That's was what I detected for now. I'll give feedback if I found more. Great work.

BDisp:
I forgot to tell that even NetDriver and v2net have enabled VTS, the escape sequences for the alternate screen buffer wont work with cmd.exe and conhost.exe.
What cmd.exe and conhost.exe supports with VTS enabled is:
- Basic ANSI colors
- Cursor movement (CSI n;m H)
- Clearing screen, text attributes
- Mouse tracking (to some extent)
- CSI ? 25 l (hide cursor) etc.

Alternate screen buffer support is only fully implemented in modern console hosts like Windows Terminal and Alacritty.

tznind:
Im using the win32 api methods for controlling the extra screen buffer.  16 color mode v2win should not be sending any terminal sequences at all on this branch I think.

    `private static extern nint CreateConsoleScreenBuffer (`

Are you saying there are terminals for windows that do not support this method call? 

The scrolling up and down issue isn't too important, I will check. But as long as when you scroll back down again it fixes itself that would probably be fine.

BDisp:
> Im using the win32 api methods for controlling the extra screen buffer. 16 color mode v2win should not be sending any terminal sequences at all on this branch I think.
> 
Although v2win doesn't enable VTS, some terminals already has VTS enabled, like Windows Terminal which doesn't support win32 API for resizing. In this case you shouldn't use and create the `_screenBuffer` but instead use the `STD_OUTPUT_HANDLE` and only using the `GetWindowSize` method. So, if VTS is enabled you should call `CSI_SaveCursorAndActivateAltBufferNoBackscroll` and `CSI_RestoreCursorAndRestoreAltBufferWithBackscroll`.
If you disable Windows Terminal by setting host windows console, you'll see that VSConsole is sending sequences, although he support VTS, but you need set the alternate screen buffer first. If you run Terminal.Gui directly in cmd.exe or conhost.exe you also will see the sending sequences.

> ```
> `private static extern nint CreateConsoleScreenBuffer (`
> ```
> 
> Are you saying there are terminals for windows that do not support this method call?
> 
Yes, as I said above Windows Terminal absolutely ignore any win32 API that does buffer resize handling, because he handle it itself.

> The scrolling up and down issue isn't too important, I will check. But as long as when you scroll back down again it fixes itself that would probably be fine.

It's very ugly. When `Terminal.Gui` is running the only scroll that must exist is the `Terminal.Gui`'s ScrollBar and not the terminal scroll, which only display corruptive buffer contents.

BDisp:
About v2net it's prepared do work well with Windows Terminal because it call `CSI_SaveCursorAndActivateAltBufferNoBackscroll` and `CSI_RestoreCursorAndRestoreAltBufferWithBackscroll`. But if you disable Windows Terminal by setting host windows console, you'll see the issue in the below gif. It's using VSConsole by debugging from VS2022. Although VSConsole has VTS enable it doesn't support, like cmd.exe and conhost.exe, alternate screen buffer. But they support win32 API buffer resize handling. In this case you should create the `_screenBuffer` to control the buffer size and setting it equal to the window size.

devenv_NkBFCvvh5g

BDisp:
About support true color in `Windows` it's just enough to verify if the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` is enabled and the build version. For Unix systems normally support it by default.
A example to get VTS:
```cs
private static bool HasVtSupport()
{
    var handle = GetStdHandle(STD_OUTPUT_HANDLE);
    return GetConsoleMode(handle, out var mode) &&
           (mode & ENABLE_VIRTUAL_TERMINAL_PROCESSING) != 0;
}
```

For v2net in Windows this isn't needed because `Terminal.Gui` already enable it internally. Maybe just verifying the build version.

BDisp:
I absolutely agree that we need to check two variables, `IsWindowsTerminal` and `IsVirtualTerminal`. I though that only `IsVirtualTerminal` was enough, but I was wrong because even with terminal with VTS enabled some supports win32 API buffer resize and other don't. What you think about?

BDisp:
In v2win beside being is ugly seeing the back buffer, the screen isn't refreshing allowing mouse handling with the corrupted buffer.

WindowsTerminal_CZzd1TMpcK

tznind:
I think we need only 2 modes

- Works with windows 10 console out of the box with no strange checkboxes enabled etc
- Works with ancient windows consoles i.e. no CSI codes just win 32 method calls

We will call these modes 'modern' and '16 color'

We don't have to detect what user is running we just assume they are running a modern console that supports true color and CSI etc.  They can set 16 color mode to revert to old style rendering.

We need 2 tables

- 1 table that shows api calls that work in modern and equivellent in 16 color mode
- 1 table that shows which consoles fit into which modes

The API users can default to first as thats the most widely supported these days and then they can let end users switch to 16 color mode using relevant CLI flags in their end user apps or config settings.

tig:
> We don't have to detect what user is running we just assume they are running a modern console that supports true color and CSI etc. They can set 16 color mode to revert to old style rendering.

I completely agree that we should not waste time trying to detect legacy terminals. We've spent way too much time on that already, and keep going in circles with edge cases and stuff changing under us. 

DEVELOPERs have little need to select modes. USERSs do. The primary way a USER should select the non-default, legacy mode is via Configuration Manager. With CM a user can set the mode on an app, user, or system basis. No new code required.

> We will call these modes 'modern' and '16 color'

I don't love these names. `modern` will no longer be "modern" in very short order. `16 color` underrepresents because other things are not supported besides true color.

I think we should call the two modes: `Default` and `Legacy`.

BDisp:
> I completely agree that we should not waste time trying to detect legacy terminals. We've spent way too much time on that already, and keep going in circles with edge cases and stuff changing under us.
>
> DEVELOPERs have little need to select modes. USERSs do. The primary way a USER should select the non-default, legacy mode is via Configuration Manager. With CM a user can set the mode on an app, user, or system basis. No new code required.
> 
I agree that if a user want legacy mode he should set it via CM, but we can't forget that `Terminal.Gui` must also provide support for it. Otherwise CM for legacy mode doesn't make any sense.

> I don't love these names. `modern` will no longer be "modern" in very short order. `16 color` underrepresents because other things are not supported besides true color.
> 
> I think we should call the two modes: `Default` and `Legacy`.

I also agree with `Default` and `Legacy` names.

BDisp:
> I absolutely agree that we need to check two variables, `IsWindowsTerminal` and `IsVirtualTerminal`. I though that only `IsVirtualTerminal` was enough, but I was wrong because even with terminal with VTS enabled some supports win32 API buffer resize and other don't. What you think about?

I was wrong about this. Only `IsVirtualTerminal` or similar check is necessary because the buffer size window code doesn't affect "Windows Terminal" which will ignore it and use his own value. On other terminal that use the win32 API buffer size window will set it accordingly. See how I've done that with v1 drivers in the PR #4202. That can be easily implemented into the v2 drivers as well.

tznind:
> See how I've done that with v1 drivers in the PR https://github.com/gui-cs/Terminal.Gui/pull/4202.  That can be easily implemented into the v2 drivers as well.

Can you send me a PR that shows this change to this branch? 

Only with changes to `WindowsOutput` if possible please.  I have handled the screen size change wholy within the `WindowsOutput` class in this PR so if you can follow that pattern in the applying of this change that would be appreciated.

See the bit in my PR where I do:

```csharp
if (_lastSize == null || _lastSize != newSize)
```

BDisp:
I can try but it's important to check if VTS is enable on WindowsInput, WindowsOutput and probably in the mainloop coordinator or facade. What the best way from each one communicate with each other?

BDisp:
The reason why resize isn't working in Windows Terminal is because with VTS enabled you must use the output handle and not the screen buffer handle which Windows Terminal handles.

tznind:
> I can try but it's important to check if VTS is enable on WindowsInput, WindowsOutput and probably in the mainloop coordinator or facade. What the best way from each one communicate with each other?

Its lifetime scoped state so I would say to use a new Singleton class that caches the answer using Lazy so its only run if called.

WindowsInput seems to be working fine, why do we need changes there too? Can we solve rendering first and then look at any input issues separately?

Why does this take changes in input and coordinator too? I dont want them knowing about windows specific stuff if at all possible - they are elegant and simple at the moment so containing changes to windows specific classes would be ideal please 🙏

BDisp:
Done in https://github.com/tznind/gui.cs/pull/179 for VTS. Legacy still not work.

BDisp:
I think you don't need to create two buffers. It's only enough to use _outputHandle (VTS) and _screenBuffer (non-VTS). The _outputHandle get a handle from STD_OUTPUT_HANDLE and _screenBuffer get a handle from CreateConsoleScreenBuffer. Do I missing something? With your code I can't avoid the extra last line in non-VTS because you are recreating again the _screenBuffer when the console is resizing, invalidating the initial configuration. For that is more correct to use the SetConsoleWindow which will call the kernel32 functions to resize the buffer, without lost the initial configuration, set on the constructor. For know I'll wait for your response because you may have some reason to do that.

tznind:
We keep two buffers in non-VT mode to reduce flicker by rendering off-screen before flipping the offscreen buffer to active (making it appear).

Without it the screen will flicker.  This is because we switched from raw buffer setting to using instead write console.

tznind:
Thre wide characters are working nicely in v2net but not v2win.

I am going to see if I can introduce a base output class that just leaves us implementing in the specific output classes methods like:

- Setup buffers etc (optional)
- Go to cursor position
- Append output

tznind:
Ok @BDisp on the plus side things are looking a lot simpler from code perspective

Down sides (v2win only)

- Wide characters are breaking completely from the point they are outputted
- Legacy mode is just a blank screen 😆 

But I think this shared base class OutputBase is going to work out nicely and vastly simplify things.  Going to leave it for now - feel free to run with it or I can pick it up some time next week.

BDisp:
@tznind when do you intend to add support to these combinations? With NetDriver all the combinations with the F1 is possible as you can see in the image.

```cs
        // Function keys with modifiers
        /*  Not currently supported
        yield return new object [] { "\u001b[1;2P", Key.F1.WithShift };
        yield return new object [] { "\u001b[1;3Q", Key.F2.WithAlt };
        yield return new object [] { "\u001b[1;5R", Key.F3.WithCtrl };
        */
```

image

tznind:
Needs patterns in AnsiParser that dont collide with other existing patterns.

Lets get comprehensive list of what is missing first

BDisp:
The patterns never colide with each other. It's only needed to add the modifiers to the patterns which in EscSeqUtils is as below. So, any number from 2 to 8 before a terminator that may be considered a key input, represent the modifier.

```cs
    /// <summary>
    ///     Gets the <see cref="ConsoleModifiers"/> from the value.
    /// </summary>
    /// <param name="value">The value.</param>
    /// <returns>The <see cref="ConsoleModifiers"/> or zero.</returns>
    public static ConsoleModifiers GetConsoleModifiers (string? value)
    {
        return value switch
        {
            "2" => ConsoleModifiers.Shift,
            "3" => ConsoleModifiers.Alt,
            "4" => ConsoleModifiers.Shift | ConsoleModifiers.Alt,
            "5" => ConsoleModifiers.Control,
            "6" => ConsoleModifiers.Shift | ConsoleModifiers.Control,
            "7" => ConsoleModifiers.Alt | ConsoleModifiers.Control,
            "8" => ConsoleModifiers.Shift | ConsoleModifiers.Alt | ConsoleModifiers.Control,
            _ => 0
        };
    }
```

tznind:
I have moved to seperate issue and fixed in 
https://github.com/gui-cs/Terminal.Gui/pull/4220

tznind:
@BDisp anything else to do on this PR or are we happy? Seems to work a lot better and be simpler code now.

BDisp:
> @BDisp anything else to do on this PR or are we happy? Seems to work a lot better and be simpler code now.

For me it's good enough. Thanks.

tig:
I'm getting married this weekend. Hence my absence. 

I'll dive back in to TG next week-ish.

tznind:
> I'm getting married this weekend. Hence my absence. 

Congratulations! 

> 
> I'll dive back in to TG next week-ish. 
> 

Haha no rush!  Enjoy the festivities

tig:
Ok... diving back in...

tig (commented):
@tznind - what is the plan for creating the final names for the v2 drivers and updating the folder structure to remove the legacy stuff etc...? Should that all be in this PR?

Terminal.Gui/Drivers/V2/OutputBase.cs
```diff
@@ -0,0 +1,163 @@
...
+        if (ConsoleDriver.RunningUnitTests)
+        {
+            return;
+        }
+
+        if (Console.WindowHeight < 1
+            || buffer.Contents.Length != buffer.Rows * buffer.Cols
+            || buffer.Rows != Console.WindowHeight)
+        {
+            //     return;
```
tig, line 22: ???

Terminal.Gui/Drivers/V2/OutputBase.cs
```diff
@@ -0,0 +1,163 @@
...
+        SetCursorVisibility (CursorVisibility.Invisible);
+
+        const int maxCharsPerRune = 2;
+        Span<char> runeBuffer = stackalloc char [maxCharsPerRune];
+
+        for (int row = top; row < rows; row++)
+        {
+            if (Console.WindowHeight < 1)
+            {
+                return;
```
tig, line 43: Why is this within the loop? It can't change while this code is running, right?

tig (approved):
This looks great to me!

PR #4197, opened 2025-07-21
Fixes #4196. Application.Begin doesn't refresh the screen at start

### LINKED ISSUE #4196: Application.Begin doesn't refresh the screen at start
With `WindowsDriver` this issue doesn't happens because a Focus event is raised when app start, forcing a screen refreshing. With v2win and v2net this issue doesn't also happens because they use a `Task.Delay` which force an iteration although no events is present. But `NetDriver` and `CursesDriver` are waiting for an event for the screen being refresh. If the user doesn't move the mouse or press any key the screen stay empty. The cause of this is when the Invoke method is called it execute immediately the action because the `Thread.CurrentThread.ManagedThreadId` is the same as `Application.MainThreadId` and return without adding the action to the `TimedEvents`.

BDisp:
## Fixes

- Fixes #4196

## Proposed Changes/Todos

- [x] Enclosing the action inside a `Task.Run`

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/App/Application.Run.cs
+++ b/Terminal.Gui/App/Application.Run.cs
@@ -160,9 +160,9 @@ public static partial class Application // Run (Begin, Run, End, Stop)
             Top = toplevel;
         }
 
-        if ((Top.Modal == false && toplevel.Modal)
-            || (Top.Modal == false && !toplevel.Modal)
-            || (Top.Modal == true && toplevel.Modal))
+        if ((Top?.Modal == false && toplevel.Modal)
+            || (Top?.Modal == false && !toplevel.Modal)
+            || (Top?.Modal == true && toplevel.Modal))
         {
             if (toplevel.Visible)
             {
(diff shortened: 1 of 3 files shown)


Terminal.Gui/App/Application.Run.cs
```diff
@@ -212,8 +212,15 @@ public static RunState Begin (Toplevel toplevel)
...
 
-        // Force an Idle event so that an Iteration (and Refresh) happen.
-        Invoke (() => { });
+        if (!ConsoleDriver.RunningUnitTests)
+        {
+            // Force an Idle event so that an Iteration (and Refresh) happen.
+            Task.Run (() =>
+                      {
+                          Invoke (() => { });
+                          Task.Delay (10).Wait ();
```
tig: Why 10ms. Why not 100? Why not 1? 

Can this be done in a way where someone reading the code understands precisely how this is working?

Writing a  // comment may work.

tig (changes requested):
Please see my question.

Terminal.Gui/App/Application.Run.cs
BDisp: The explanation is in https://github.com/gui-cs/Terminal.Gui/issues/4196#issue-3249176460, but I explained in a comment. I also decrease the delay to 1 because it's enough to be added in the timeouts.

PR #4199, opened 2025-07-21
Fixes #4198. Application.Invoke isn't wakeup the driver if idle

### LINKED ISSUE #4198: Application.Invoke isn't wakeup the driver if idle
To see this issue please run the AnimationScenario which doesn't refresh in each `Invoke` call.
This issue doesn't happen with v2win and v2net because they use a `Task.Delay` which will for force a new iteration even no events are present. With `WindowsDriver`, `NetDriver` and `CursesDriver` this is a critical issue if they are waiting for some event to wakeup. When `Application.Invoke` is called it must wakeup the driver if it's idle.

BDisp:
## Fixes

- Fixes #4198

## Proposed Changes/Todos

- [x] Call `Application.MainLoop.Wakeup` method

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/App/ApplicationImpl.cs
+++ b/Terminal.Gui/App/ApplicationImpl.cs
@@ -293,6 +293,10 @@ public class ApplicationImpl : IApplication
                                return false;
                            }
                           );
+
+        // Ensure the action is executed in the main loop
+        // Wakeup mainloop if it's waiting for events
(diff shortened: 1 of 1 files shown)


tznind:
This should only affect the legacy drivers, v2 drivers use `ApplicationV2` instead of `ApplicationImpl`

And it uses an override of `Invoke` so this code change (in base class) should be fine.

PR #4205, opened 2025-07-23
Fixes #4204. v2win and v2net aren't refreshing the Character Map correctly

### LINKED ISSUE #4204: v2win and v2net aren't refreshing the Character Map correctly
It's need to click the down button on the scroll bar to display it fully.

BDisp:
## Fixes

- Fixes #4204

## Proposed Changes/Todos

- [x] Only needed to call `Application.LayoutAndDrawImpl (true);` in the `MainThreadId`

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/V2/ApplicationV2.cs
+++ b/Terminal.Gui/Drivers/V2/ApplicationV2.cs
@@ -232,6 +232,9 @@ public class ApplicationV2 : ApplicationImpl
         if (Application.MainThreadId == Thread.CurrentThread.ManagedThreadId)
         {
             action ();
+            // Ensure the action is executed and forces LayoutAndDrawImpl
+            Application.LayoutAndDrawImpl (true);
+
             return;
         }
 


tznind (changes requested):
If i read this correct you are forcing an explicit redraw on every invoke user does?

This violates v2 design of redraw rate limiting.

There should only ever be 1 layout/redraw per iteration and it should come from the new loop I wrote.

Correct way to handle this would be for whoever wants redrawn to set their NeedsDraw flag instead.

tznind:
Oh - I see this is actually going to do just the flags because of the override on the method... 

Still i dont think every invoke should force this flag.

If user code invokes it should set its own flag and only if it really does want redrawn

BDisp:
I think the main problem is the legacy drivers already set the console size in the Init method before call the Begin method and the V2 drivers only do that in the first iteration in the Begin method.

tig:
I fixed this already. Maybe it was just in my active PR. 

I'll look later.

tig:
> I fixed this already. Maybe it was just in my active PR. 
> 
> I'll look later. 

Nope. I mis-remembered. This bug still exists in #4126.

BDisp:
@tznind has right about some bug in the redraw by not clearing correctly draw flags.

tig:
> > I fixed this already. Maybe it was just in my active PR.
> > I'll look later.
> 
> Nope. I mis-remembered. This bug still exists in #4126.

Oops. Was running an older pull on this machine. I *did* already fix this!

This is the fix, in `View.cs`:

image

tig:
I recommend this (at line 248 of View.cs):

```cs
        // Force a layout each time a View is initialized
        // See: https://github.com/gui-cs/Terminal.Gui/issues/3951
        // See: https://github.com/gui-cs/Terminal.Gui/issues/4204
        Layout ();
```

tig:
Actually, please add these comments as well:

```cs
        // Force a layout each time a View is initialized
        // See: https://github.com/gui-cs/Terminal.Gui/issues/3951
        // See: https://github.com/gui-cs/Terminal.Gui/issues/4204
        Layout ();

        // Complex layout scenarios (e.g. DimAuto and PosAlign) may require multiple layouts to be performed.
        // Thus, we call SetNeedsLayout() to ensure that the layout is performed at least once.
        SetNeedsLayout ();
```

BDisp:
Thanks for your fix. I really didn't saw that before. It really fix the v2 driver refresh but not the constantly redrawing in the legacy drivers as @tznind related. In the v2 drivers this doesn't happens because they only rely on `NeedsDraw` and `NeedsLayout` and the legacy drivers rely on `NeedsLayout` and `SubViewNeedsDraw`. So, the problem with the legacy drivers is that `SubViewNeedsDraw` is always true at the end. I'm still trying find the cause by adding `Debug.Assert` in some places to more easily see what going on.

tig:
> Thanks for your fix. I really didn't saw that before. It really fix the v2 driver refresh but not the constantly redrawing in the legacy drivers as @tznind related. In the v2 drivers this doesn't happens because they only rely on `NeedsDraw` and `NeedsLayout` and the legacy drivers rely on `NeedsLayout` and `SubViewNeedsDraw`. So, the problem with the legacy drivers is that `SubViewNeedsDraw` is always true at the end. I'm still trying find the cause by adding `Debug.Assert` in some places to more easily see what going on.

Great. But you agree the fix I posted above fixes THIS issue (#4204)? Right?

BDisp:
> Great. But you agree the fix I posted above fixes THIS issue (#4204)? Right?

Yes, absolutely.

BDisp:
The CI pass on my laptop using Windows and Linux. So I think this is a race condition.

BDisp:
> The CI pass on my laptop using Windows and Linux. So I think this is a race condition.

Running again the unit tests it really failed. I merged into this the #4209 PR which fix the SyncrhonizationContextTests.

tig:
@bdisp - Integration tests are still failing.

BDisp:
> @BDisp - Integration tests are still failing.

I run now without errors. I think that is a racing condition where a collection is being modified in the \IntegrationTests\FluentTests\MenuBarv2Tests.cs:line 454 when executing "Application.Top!.Add (menuBar);" which will modify a collection.

BDisp:
I think FluentTests are running in parallel which is causing different results sometimes.

Now passed without any change. So, what I said above has some sense.

PR #4210, opened 2025-07-25
Fixes #3951. Add DimFuncWithView with a View dependency

### LINKED ISSUE #3951: v2win/v2net: UICatalog Character Map render/layout issues
**Describe the bug**

UICatalog Character Map category jump list is hidden until the character list is scrolled.

**To Reproduce**

1. Start UICatalog
2. Open Character Map scenario
    - Right-side category list missing
4. Scroll the character list
    - Right-side list appears 

**Expected behavior**

The character map category list is displayed from the beginning similar to the legacy console drivers.

**Screenshots**

Image

**Desktop**
- OS: Windows 10 22H2 (build 19045.5487)
- CLI: Windows Terminal v1.21.10351.0 /w cmd
- Terminal.GUI branch: v2_develop

BDisp:
## Fixes

- Fixes #3951

## Proposed Changes/Todos

- [x] Added DimFuncWithView with View dependency

## Pull Request checklist:

### DIFF
--- a/Examples/UICatalog/Scenarios/CharacterMap/CharacterMap.cs
+++ b/Examples/UICatalog/Scenarios/CharacterMap/CharacterMap.cs
@@ -39,7 +39,6 @@ public class CharacterMap : Scenario
         {
             X = 0,
             Y = 1,
-            Width = Dim.Fill (Dim.Func (() => _categoryList!.Frame.Width)),
             Height = Dim.Fill (),
            // SchemeName = "Base"
 
@@ -171,6 +170,8 @@ public class CharacterMap : Scenario
             ]
         };
         top.Add (menu);
+
+        _charMap.Width = Dim.Fill (Dim.FuncWithView (v => v!.Frame.Width, _categoryList));
 
         _charMap.SelectedCodePoint = 0;
         _charMap.SetFocus ();
--- a/Terminal.Gui/App/Application.Run.cs
+++ b/Terminal.Gui/App/Application.Run.cs
@@ -189,7 +189,7 @@ public static partial class Application // Run (Begin, Run, End, Stop)
         if (!toplevel.IsInitialized)
         {
             toplevel.BeginInit ();
-            toplevel.EndInit (); // Calls Layout
+            toplevel.EndInit (); // Calls SetNeedsLayout
         }
 
         // Call ConfigurationManager Apply here to ensure all subscribers to ConfigurationManager.Applied
@@ -212,8 +212,7 @@ public static partial class Application // Run (Begin, Run, End, Stop)
 
         NotifyNewRunState?.Invoke (toplevel, new (rs));
 
-        // Force an Idle event so that an Iteration (and Refresh) happen.
-        Invoke (() => { });
+        LayoutAndDraw (true);
 
         return rs;
     }
--- a/Terminal.Gui/App/MainLoopSyncContext.cs
+++ b/Terminal.Gui/App/MainLoopSyncContext.cs
@@ -11,15 +11,22 @@ internal sealed class MainLoopSyncContext : SynchronizationContext
     public override void Post (SendOrPostCallback d, object state)
     {
         // Queue the task
-        Application.MainLoop?.TimedEvents.Add (TimeSpan.Zero,
-                                                      () =>
-                                                      {
-                                                          d (state);
+        if (ApplicationImpl.Instance.IsLegacy)
+        {
+            Application.MainLoop?.TimedEvents.Add (TimeSpan.Zero,
+                                                   () =>
+                                                   {
+                                                       d (state);
 
-                                                          return false;
-                                                      }
-                                                     );
-        Application.MainLoop?.Wakeup ();
+                                                       return false;
+                                                   }
+                                                  );
+            Application.MainLoop?.Wakeup ();
+        }
+        else
+        {
+            ApplicationImpl.Instance.Invoke (() => { d (state); });
+        }
     }
 
     //_mainLoop.Driver.Wakeup ();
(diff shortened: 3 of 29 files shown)


BDisp:
`View.EndInit` only should call `View.SetNeedsLayout` and the `Application.Begin` should call `LayoutAndDraw` and `PositionCursor`. This avoid layout event run more than once and there is no need to `Begin` call `Invoke` because in the `RunIteration` the `LayoutAndDraw` and `PositionCursor` only run after the first `Iteration` event is raised.

tig (changes requested):
This is slick. Nice work.

I think it would be better if, instead of creating a new class for `DimFuncWIthView`, you instead added an optional `View` property to `DimFunc` and added another constuctor that took a `View`. Or made the existing `DimFunc` constructor take an optional `View` parameter.

BDisp:
Done. I opted for add the optional View parameter and property because is impossible to store in a property two different types `Func<int>` and `Func<View, int>`. I was forgetting to also add this feature in the `PosFunc`. All done, I hope it's enough.

BDisp:
I'm only wasting my damn time here.

BDisp:
Finally. I think all the recursion issues covered by the unit tests are resolved.

Terminal.Gui/ViewBase/View.Drawing.cs
```diff
@@ -131,6 +153,11 @@ public void Draw (DrawContext? context = null)
 
     private void DoDrawAdornmentsSubViews ()
     {
+        if (Border?.NeedsLayout == true)
```
tig: We should never have to invoke layout from within draw. 

We must find a way to fix whatever problem you discovered that required these calls to layout adornments without having to do it from the Draw process.

Can you explain the problem that required these layout calls to be put here? Is there a unit test that fails if you remove them?

Terminal.Gui/ViewBase/View.Hierarchy.cs
```diff
@@ -558,3 +558,22 @@ private void PerformActionForSubView (View subview, Action<View> action)
 
     #endregion SubViewOrdering
 }
+
+internal static class ViewCollectionHelpers
```
tig: This should be in a separate file. One class per file.

Terminal.Gui/ViewBase/View.Layout.cs
```diff
@@ -419,7 +419,10 @@ public bool Layout (Size contentSize)
         {
             LayoutSubViews ();
 
-            // Debug.Assert(!NeedsLayout);
+            // A layout was performed so a draw is needed
+            // NeedsLayout may still be true if a dependent View still needs layout after SubViewsLaidOut event
```
tig, line 423: It is possible that no layout actually occured and thus this will unnecessarily cause redraws.

Are you sure this is needed?

Terminal.Gui/ViewBase/View.Layout.cs
```diff
@@ -635,7 +638,7 @@ internal void LayoutSubViews ()
 
         List<View> redo = new ();
 
-        foreach (View v in ordered)
+        foreach (View v in ordered.Snapshot ())
```
tig, line 641: Snapshot is very expensive. Does it rally need to be used here (and the other places where you've used it)?

tig (changes requested):
Please see my questions and requests.

This is awesome stuff!

Terminal.Gui/ViewBase/View.Layout.cs
BDisp: There are situations where an adornment was changed and if it has subviews the adornment need to set NeedsLayout to true to force draw the subviews.

Terminal.Gui/ViewBase/View.Layout.cs
BDisp: That was to fix the integration unit tests at the time and was achieved successfully the desired effect. But @tznind is also being fix the integration unit tests and I don't know if it's needed now. What I see is a great failures in the integration unit tests which causes a lot of troubles.

BDisp: As you can see this is the only branch that doesn't fail in the integration unit tests.

Terminal.Gui/ViewBase/View.Hierarchy.cs
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4210/commits/0e0337345b31c249e265d62aefc1a2724ee1c084.

Terminal.Gui/ViewBase/View.Drawing.cs
BDisp: If I recall at the time the border wasn't lay outing before draw the subviews.

BDisp: I removed it in https://github.com/gui-cs/Terminal.Gui/pull/4210/commits/552ab14fc8f4cbb9ecf3f3258841afbe99a1fc0d.

Terminal.Gui/ViewBase/View.Drawing.cs
```diff
@@ -151,6 +173,11 @@ private void DoDrawAdornmentsSubViews ()
             SetClip (saved);
         }
 
+        if (Padding?.NeedsLayout == true)
+        {
+            Padding.Layout ();
```
tig: Do not call layout from the draw loop. 

Please remove these calls to Layout. We should never do layout functions from within the draw path.

BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4210/commits/dd7c73fb2409cd1016be31827733270ce876fc26.

Terminal.Gui/ViewBase/View.Layout.cs
BDisp: I'm only using Snapshot in the places where was causing the error that the collection was changed, mainly in FluentTests.

PR #4220, opened 2025-08-07
Fixes #4221 Extra modifiers f1 to f4 in v2net

### LINKED ISSUE #4221: Add extra modifies for F1/F4
@tznind when do you intend to add support to these combinations? With NetDriver all the combinations with the F1 is possible as you can see in the image.

```cs
        // Function keys with modifiers
        /*  Not currently supported
        yield return new object [] { "\u001b[1;2P", Key.F1.WithShift };
        yield return new object [] { "\u001b[1;3Q", Key.F2.WithAlt };
        yield return new object [] { "\u001b[1;5R", Key.F3.WithCtrl };
        */
```

image

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/4193#issuecomment-3161584941_

tznind:
## Fixes

- Fixes #4221 

## Proposed Changes/Todos

F1 key can be 
`<esc>OP`

But if combined with modifier keys it can instead be:

`<esc>[1;3P`

This PR updates key parsing to support this sequence

- [x] Support Alt+F1, Shift+F1 etc

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/V2/MainLoopCoordinator.cs
+++ b/Terminal.Gui/Drivers/V2/MainLoopCoordinator.cs
@@ -25,7 +25,6 @@ internal class MainLoopCoordinator<T> : IMainLoopCoordinator
     private ConsoleDriverFacade<T> _facade;
     private Task _inputTask;
     private readonly ITimedEvents _timedEvents;
-    private readonly bool _isWindowsTerminal;
 
     private readonly SemaphoreSlim _startupSemaphore = new (0, 1);
 
@@ -61,7 +60,6 @@ internal class MainLoopCoordinator<T> : IMainLoopCoordinator
         _inputProcessor = inputProcessor;
         _outputFactory = outputFactory;
         _loop = loop;
-        _isWindowsTerminal = Environment.GetEnvironmentVariable ("WT_SESSION") is { } || Environment.GetEnvironmentVariable ("VSAPPIDNAME") != null;
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Terminal.Gui/Drivers/V2/MainLoopCoordinator.cs
```diff
@@ -25,6 +25,7 @@ internal class MainLoopCoordinator<T> : IMainLoopCoordinator
     private ConsoleDriverFacade<T> _facade;
     private Task _inputTask;
     private readonly ITimedEvents _timedEvents;
+    private readonly bool _isWindowsTerminal;
```
BDisp: It's better to check if the console support VTS. In this case the field should be `_isVirtualTerminal`. Of course with v2win you have to get this from the STD_OUTPUT_HANDLE and verify if ENABLE_VIRTUAL_TERMINAL_PROCESSING is enabled. Alacritty has VTS enabled and if you only check for `Environment.GetEnvironmentVariable ("WT_SESSION") is { } || Environment.GetEnvironmentVariable ("VSAPPIDNAME") != null;` it won't support true colors because  you'll force 16 colors. With v2net VTS is always enabled by itself, so true color is always true in Unix like system and in Windows depends of the build version. The problem is check for VTS in this class. On my PR that I reverted I did a call to WindowsOutput to get the internal field `IsVirtualTerminal`, but you probably have another better solution for this. In Windows also always check the OS version build.

tznind: This is not in diff I think, I revert that accidental commit

BDisp: > This is not in diff I think, I revert that accidental commit

I saw again and it was reverted, sorry. Good choice to assume they assume true colors at start. After start the app they are assumed or not accordingly with what the console supports.

Issue #4102, opened 2025-05-29
Add AOT version of Tests


tig:
Each of the Test projects should have a AOT version.

It should be possible to create e.g., `./Tests/UnitTests.AOT` that links to all the tests in `./Tests/UnitTests` but has project settings enabling AOT.

We should also have the same for each project in `/Examples`.

tig:
@BDisp - What are your thoughts on doing this?

BDisp:
I don't think I can handle that responsibility. I'd rather you write it yourself and I'll give you all the help I can within the limits of my knowledge, if you don't mind.

tig:
> I don't think I can handle that responsibility. I'd rather you write it yourself and I'll give you all the help I can within the limits of my knowledge, if you don't mind.

I have no idea how to do it.

BDisp:
> I have no idea how to do it.

I thought you already had some idea. In that case I will try to work out some solution. However, the most reliable test is always in release mode, because in debug mode we can't always detect incompatibilities.

tig:
My only thought is i really hope there's a way to do it where the new projects don't copy any code, but somehow link to the other non AOT test projects.

BDisp:
> We should also have the same for each project in `/Examples`.

I agree, but I would create a `ExampleWindow.cs` file with the `public class ExampleWindow : Window` in the `Example` project and link it to the `NativeAot` and `SelfContained` projects, removing duplicated code. The `NativeAot` and `SelfContained` projects should maintain as is because already handle with trimmed compilation. Only the `CommunityToolkitExample`, `Example`, and `ReactiveExample` projects may worth create a `.AOT` project for each one. What's your opinions?

BDisp:
`NetDriver` and `v2net` are rendering badly the `Button` only in `Windows`.

![Image](https://github.com/user-attachments/assets/adea4746-40c9-4908-93bc-07a5919cdf04)

![Image](https://github.com/user-attachments/assets/22a33b28-2c4b-4dd5-8ea3-d2d2023f995d)

BDisp:
But what is odd is that the output variable is correct. Thus it seems that the `Console.Write` or the `Windows Terminal` isn't rendering well the output.

![Image](https://github.com/user-attachments/assets/cd9567b3-d817-4911-ba04-9fb5e9dcbc3b)

![Image](https://github.com/user-attachments/assets/b30f9291-0b55-4e15-9594-6c8fad7639d7)

[...]

PR #4222, opened 2025-08-09
Add v2 version of fake driver

### LINKED ISSUE #3947: Replace `FakeDriver` with a `v2fake` or equivalent
@tznind: can you please fix the warnings in the driver code? You introduced a bunch.

Image

tznind:
## Fixes

- Fixes #3947 

### Requirements

Support creating `Application` and `Driver` level tests for v2.

- Driver level tests are those that use `[SetupFakeDriver]`
- Application level tests are those that use `[AutoInitShutdown]`

These tests need:
- To 'work' 
- To be able to resize 

### Implementation
To enable this workflow we first need to pull all the Dependency Injection elements out into a shared provider factory.  This lets us create classes but swap out Mocks and fakes as needed.

- Add IComponentFactory for creating system internal classes
  - Input
  - Output
  - WindowSize
  - etc

This lets us inject fakes.

- [ ] Todo 1

## Test Assumption Changes

| Test | Description |
|-----------------|----------------|
|[Can_Access_Cancel_Property_After_Run](https://github.com/gui-cs/Terminal.Gui/pull/4222/commits/67c4258e190a89af7896fe4be8ba4836fc9f295d) | Test seems to want main loop run to detect when you `Run` multiple views one after the other without first disposing the first.  This is not a sensible activity and so part of v2. |
|[Dialog_Opened_From_Another_Dialog](https://github.com/gui-cs/Terminal.Gui/pull/4222/commits/7b5324c20c0435c8a495d41cbf9466811aa8c19b)|Run causes iteration and we no longer support adhoc calls to LayoutAndDraw so the indexes these check are now staggered|

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/V2/NetOutput.cs
+++ b/Terminal.Gui/Drivers/V2/NetOutput.cs
@@ -72,31 +72,6 @@ public class NetOutput : OutputBase, IConsoleOutput
     }
 
 
-    /// <inheritdoc/>
-    protected override void Write (StringBuilder output) { Console.Out.Write (output); }
-
-    private Point? _lastCursorPosition;
-
-    /// <inheritdoc />
-    protected override void AppendOrWriteAttribute (StringBuilder output, Attribute attr, TextStyle redrawTextStyle)
-    {
-        EscSeqUtils.CSI_AppendForegroundColorRGB (
-                                                  output,
-                                                  attr.Foreground.R,
-                                                  attr.Foreground.G,
-                                                  attr.Foreground.B
-                                                 );
-
-        EscSeqUtils.CSI_AppendBackgroundColorRGB (
-                                                  output,
-                                                  attr.Background.R,
-                                                  attr.Background.G,
-                                                  attr.Background.B
-                                                 );
-
-        EscSeqUtils.CSI_AppendTextStyleChange (output, redrawTextStyle, attr.Style);
-    }
-
     /// <inheritdoc />
     protected override void Write (StringBuilder output)
     {


tznind:
This is starting to shape up.

I have a lot of tests that are just drawing blank. Any ideas @?

Ideally keeping any changes to real v2 drivers at a minimum. Lets focus firstn on understand and categorising current failures.

Tests/TerminalGuiFluentTesting/FakeDriverV2.cs
```diff
@@ -0,0 +1,156 @@
...
+
+
+        IApplication origApp = ApplicationImpl.Instance;
+
+        var sizeMonitor = new FakeSizeMonitor ();
+
+        var v2 = new ApplicationV2 (new FakeNetComponentFactory (fakeInput, _output, sizeMonitor));
+
+        ApplicationImpl.ChangeInstance (v2);
+        v2.Init (null,"v2net");
```
BDisp, line 29: The problem is here, because you're forcing creating a v2net driver instead of a v2fake driver. Thus the driver size is always empty causing drawing blank screen.

BDisp: The driver Cols and Rows aren't initialized with the output size.

tznind: > The problem is here, because you're forcing creating a v2net driver instead of a v2fake driver. Thus the driver size is always empty causing drawing blank screen.

The FakeNetConponentFactory should result in correct fake driver.

I'll look into the size initialise though - thanks 😊

BDisp: Also call `Application.LayoutAndDraw ();` in the unit test is really only calling `SetNeedsDraw` and `SetNeedsLayout` methods for the `Application.Top` and isn't really lay outing and  drawing nothing.

BDisp:
Since you now created the v2 of fake driver can you get rid of the `Mock` dependency because it isn't AOT compatible and doesn't have any source generator at compile time? You probably won't need it now and I already ask the owner but he want to get paid to do that. Thanks.

tznind:
Moq is awesome,  why would we want rid of it? Its for test only. 

If you want to run tests as AOT we can take that as seperate issue I think.

I want to focus first on getting tests passing

tznind:
Ok heres how we going to handle this.  I just fixed some tests in DialogTests

[33f5a5b](https://github.com/gui-cs/Terminal.Gui/pull/4222/commits/33f5a5b2c25e03ed3e3ab748f7f588958ad0247a)

So far there are 2 things that every test seems to want to do:

- Resize the console
- Run an iteration

Previously these were done by directly interacting with `Application` and casting to `FakeDriver` etc

I will update all call sites and standardize them to use static methods in `AutoInitShutdownAttribute`.  This will allow for the implementation to be updated/refined if needed later.

Currently they are:

```csharp
    /// <summary>
    /// 'Resizes' the application and forces layout. Only works if your test uses <see cref="AutoInitShutdownAttribute"/>
    /// </summary>
    /// <param name="size"></param>
    public static void FakeResize (Size size)
    {
        var d = (IConsoleDriverFacade)Application.Driver!;
        d.OutputBuffer.SetWindowSize (size.Width, size.Height);
        ((FakeSizeMonitor)d.WindowSizeMonitor).RaiseSizeChanging (size);

        Application.LayoutAndDrawImpl ();
    }

    /// <summary>
    /// Runs a single iteration of the main loop (layout, draw, run timed events etc.)
    /// </summary>
    public static void RunIteration ()
    {
        var a = (ApplicationV2)ApplicationImpl.Instance;
        a.Coordinator?.RunIteration ();
    }
```

Hopefully I can simplify the `FakeResize` one - it shouldn't need to get so deep into internals of app.  But I will focus on getting tests to pass first.

BDisp:
> Moq is awesome, why would we want rid of it? Its for test only.
> 
I know it's really awesome, but isn't compatible with AOT unit tests.

> If you want to run tests as AOT we can take that as seperate issue I think.
> 
There is already the issue #4102 to add AOT version of unit tests. With the PR #4134 I found that the Mock dependency isn't AOT compatible.

> I want to focus first on getting tests passing

Sure, no problem. I only want to ask what Mock does that the v2 fake driver don't, for to know what I need to do later in the PR #4134. Thanks.

tznind:
The problem with `DrawFrame_With_Negative_Positions`

Is that run iteration doesn't think theres anything changed that needs drawn
```csharp
        menu.CloseAllMenus ();
        menu.Frame = new (-1, -2, menu.Frame.Width, menu.Frame.Height);

        menu.OpenMenu ();
        AutoInitShutdownAttribute.RunIteration ();
```

This returns false
```csharp
 bool needsDrawOrLayout = AnySubViewsNeedDrawn (Application.Popover?.GetActivePopover () as View)
                          || AnySubViewsNeedDrawn (Application.Top);
```

For this test @BDisp where is the menu actually tracked? It doesn't seem to be added to anything just newed and opened.  Does it get attatched to top or something? how to see if there are any menu views (v1) that require redraw?

Tests/UnitTests/Dialogs/DialogTests.cs
```diff
@@ -928,7 +928,7 @@ public void Dialog_In_Window_With_Size_One_Button_Aligns ()
 
                           dlg.Loaded += (s, a) =>
                                         {
-                                            LayoutAndDraw ();
```
BDisp, line 934: Why only calling `LayoutAndDraw` shouldn't be enough for lay outing and drawing without running an iteration? Normally `RunIteration` is used if we want to run all the subsets at once, timeouts, timed events, layout, drawing and position cursor, etc.

Tests/UnitTests/FileServices/FileDialogTests.cs
```diff
@@ -46,9 +46,9 @@ public void DirectTyping_Allowed ()
                      );
 
         // continue typing the rest of the path
-        Send ("BOB");
+        Send ("bob");
```
BDisp: I'm seeing many unit tests changed when the v2 is the one who have to be adapted to make them being passed. Only the one that it's proved that was wrong with the v1 drivers should be fixed.

Tests/UnitTests/Dialogs/DialogTests.cs
tznind: > Why only calling `LayoutAndDraw` shouldn't be enough for lay outing and drawing without running an iteration? 

Because this is against the law in v2.

Views cannot arbitrarily decide to order this expensive activity.   If views can render when they want then it leads to performance bugs e.g. multiple renders per iterations or circular rendering. 

The new law in v2 is that only the main loop T gets to do this activity.

That way it is controlled and standardised e.g. with frame limit that prevents excess iterations speeds.

Tests/UnitTests/FileServices/FileDialogTests.cs
tznind: > I'm seeing many unit tests changed when the v2 is the one who have to be adapted to make them being passed. Only the one that it's proved that was wrong with the v1 drivers should be fixed.

I disagree.

Where there is meaningful change to test assumptions I have documented in above table the change.

I have standardised.  There are tests that want to call iteration on main loop, others want to do it direct on Application instead.  We have a new loop in v2 and it does the layout and draw.

Most of these tests are only incidentally testing main loop when really what they want is to process key input then have a particular appearance rendered after

BDisp:
> For this test @BDisp where is the menu actually tracked? It doesn't seem to be added to anything just newed and opened. Does it get attatched to top or something? how to see if there are any menu views (v1) that require redraw?

Because `Application.Run.cs` has the follow code where `ClearScreenNextIteration` is `true` when `SuperView` is null and `Application.TopLevels.Count == 1` which is setting in the `View.Layout.cs` as bellow, called from the `Menu.BeginInit` method.

Application.Run.cs:
image

View.Layout.cs:
image

Tests/UnitTests/Dialogs/DialogTests.cs
BDisp: Well we are using a fake driver where some unit tests can bypass these rules, in my opinion.

tznind: I understand your point of view but I still prefer to continue with my current approach.

It standardise tests and allows later on to swap out approach in one place only  (the iteration/resize test method).

Iteration is the thing that happens constantly in a Terminal.Gui app - so it makes sense that all tests that are comparing before/after style things use this method.

Ultimately we would love if Application was non static.  Having standardisation in unit tests brings us one step closer to this too.  Imagine if the autoinit attribute could do either v1 or v2 based on debugging option for example

BDisp: > I understand your point of view but I still prefer to continue with my current approach.
> 
I completely agree that we always have to do the way in what we believe it's more correct.

> It standardise tests and allows later on to swap out approach in one place only (the iteration/resize test method).
> 
Also agree in standard unit tests that can be run in any driver version.

> Iteration is the thing that happens constantly in a Terminal.Gui app - so it makes sense that all tests that are comparing before/after style things use this method.
> 
Disagree here because only the unit tests that call `Application.Run`, `Application<T>.Run` or `Application.Run(Toplevel)` runs `Iteration`. Only by assign an unit test with an attribute, like the `AutoInitShutdownAttribute` doesn't mean the iterations will be called, but only by call `Run` itself.

> Ultimately we would love if Application was non static. Having standardisation in unit tests brings us one step closer to this too. Imagine if the autoinit attribute could do either v1 or v2 based on debugging option for example

That would be nice, but I think the fact the `Application` is static isn't a stopper to do what you would like. Just passing the driver type to the constructor, based on the debug option, may be possible to run any driver version, v1 or v2.

BDisp:
If you want a prove that I'm correct about the unnecessary call to the `AutoInitShutdownAttribute.RunIteration` method when iteration is not needed, do the follow changes and the unit test pass:

Replace `AutoInitShutdownAttribute.RunIteration` again to `Application.LayoutAndDraw`:
```cs
    [Fact]
    [AutoInitShutdown]
    public void DrawFrame_With_Negative_Positions ()
    {
        var menu = new MenuBar
        {
            X = -1,
            Y = -1,
            Menus =
            [
                new (new MenuItem [] { new ("One", "", null), new ("Two", "", null) })
            ]
        };
        menu.Layout ();

        Assert.Equal (new (-1, -1), new Point (menu.Frame.X, menu.Frame.Y));

        Toplevel top = new ();
        Application.Begin (top);
        menu.OpenMenu ();
        Application.LayoutAndDraw ();

        var expected = @"
──────┐
 One  │
 Two  │
──────┘
";

        Rectangle pos = DriverAssert.AssertDriverContentsWithFrameAre (expected, output);
        Assert.Equal (new (0, 0, 7, 4), pos);

        menu.CloseAllMenus ();
        menu.Frame = new (-1, -2, menu.Frame.Width, menu.Frame.Height);
        menu.OpenMenu ();
        Application.LayoutAndDraw ();

        expected = @"
 One  │
 Two  │
──────┘
";

        pos = DriverAssert.AssertDriverContentsWithFrameAre (expected, output);
        Assert.Equal (new (1, 0, 7, 3), pos);

        menu.CloseAllMenus ();
        menu.Frame = new (0, 0, menu.Frame.Width, menu.Frame.Height);
        AutoInitShutdownAttribute.FakeResize(new Size(7, 5));
        menu.OpenMenu ();
        Application.LayoutAndDraw ();

        expected = @"
┌──────
│ One  
│ Two  
└──────
";

        pos = DriverAssert.AssertDriverContentsWithFrameAre (expected, output);
        Assert.Equal (new (0, 1, 7, 4), pos);

        menu.CloseAllMenus ();
        menu.Frame = new (0, 0, menu.Frame.Width, menu.Frame.Height);
        AutoInitShutdownAttribute.FakeResize(new Size(7, 3));
        menu.OpenMenu ();
        Application.LayoutAndDraw ();

        expected = @"
┌──────
│ One  
│ Two  
";

        pos = DriverAssert.AssertDriverContentsWithFrameAre (expected, output);
        Assert.Equal (new (0, 0, 7, 3), pos);
        top.Dispose ();
    }
```

Remove the unnecessary `override void LayoutAndDraw` from ApplicationV2.cs file:
```cs
    /// <inheritdoc />
    public override void LayoutAndDraw (bool forceDraw)
    {
        // No more ad-hoc drawing, you must wait for iteration to do it
        Application.Top?.SetNeedsDraw();
        Application.Top?.SetNeedsLayout ();
    }
```

tznind:
Have found how to detect the need to redraw a menu:

```diff
            bool needsDrawOrLayout = AnySubViewsNeedDrawn (Application.Popover?.GetActivePopover () as View)
                                     || AnySubViewsNeedDrawn (Application.Top)
+                                     || (Application.MouseGrabHandler.MouseGrabView != null && AnySubViewsNeedDrawn (Application.MouseGrabHandler.MouseGrabView));
```

Tests/UnitTests/FileServices/FileDialogTests.cs
BDisp: I think you didn't understood what I meant to said here. If with v1 drivers this unit test pass with "BOB" and "CSV" without hacking shift flag with `char.IsUpper (ch)` in the `Send` method, then with the v2 driver should also assert as "bob.csv". An upper case "B" with shift as true should return "Key.B.WithShift", outputing "B". But you're asserting as lower case, which is incorrect, because you're forcing upper case by setting Shift flag as true. That why I said to leave the unit tests as is, to be compatible with both driver versions, v1 and v2. Only the necessary v2 code should be implemented to allow the unit tests also being passed, but not by changing the act and assertions. I'll sent you a PR proving this.

tznind: SendKeys is hack city.  It has a signature that makes no sense.  It should simply take single arg `Key`.

ConsoleKey is driver specific part of input processing not key processing.

I have reluctantly implemented SendKey into facade where previously it did nothing.  But I have had to map to Key.

This test you cite was already hacky and full of workarounds. Specifically it sent in caps then expected driver to interpret as lower by its asserts.

BDisp: But you don't agree that these tests must have to pass as it's pass with v1 fake driver? Do you meaning v1 fake driver is wrong? I'm really confused, sorry. You have to play with the modifiers keys too. See my PR https://github.com/tznind/gui.cs/pull/183, please. Thanks.

BDisp: > This test you cite was already hacky and full of workarounds. Specifically it sent in caps then expected driver to interpret as lower by its asserts.

With the `Key` class implemented by @tig a upper letter is only considered as upper case if the shift flag is true. That the new rule in the `Key` class. I thought you already know that.

BDisp: With my PR we can now manipulate upper and lower cases to get the same path names as it's sent. But this is only to manipulate path names. For hot keys we need always to sent as upper case.

BDisp:
> Have found how to detect the need to redraw a menu:
> 
> ```diff
>             bool needsDrawOrLayout = AnySubViewsNeedDrawn (Application.Popover?.GetActivePopover () as View)
>                                      || AnySubViewsNeedDrawn (Application.Top)
> +                                     || (Application.MouseGrabHandler.MouseGrabView != null && AnySubViewsNeedDrawn (Application.MouseGrabHandler.MouseGrabView));
> ```

Nice. But when v2 needs to draw it call  `Application.LayoutAndDrawImpl (true);` and  that why I think when iteration isn't necessary, it's enough to call `Application.LayoutAndDraw ();` as before, by removing the `public override void LayoutAndDraw (bool forceDraw)` method. But it's up to you.

tznind:
@BDisp the MapConsoleKeyInfoToKeyCode that you added in SendKeys has some issues

```
ConsoleKeyMapping.MapConsoleKeyInfoToKeyCode (new ConsoleKeyInfo('f',ConsoleKey.F,false,false,false))
```

This returns `Space | F`

Which is invalid input

It crashes:

```csharp

    [Fact]
    [AutoInitShutdown]
    public void TestAutoAppend_AfterCloseKey_NoAutocomplete ()
    {
        TextField tf = GetTextFieldsInViewSuggesting ("fish");

        // f is typed and suggestion is "fish"
        Application.Driver?.SendKeys ('f', ConsoleKey.F, false, false, false);
```

Crashes like this

```
  Message: 
System.ArgumentException : Invalid KeyCode: Space, F is invalid. (Parameter 'value')

  Stack Trace: 
Key.set_KeyCode(KeyCode value) line 253
Key.ctor(KeyCode k) line 80
Key.op_Implicit(KeyCode keyCode) line 417
ConsoleDriverFacade`1.SendKeys(Char keyChar, ConsoleKey key, Boolean shift, Boolean alt, Boolean ctrl) line 399
AppendAutocompleteTests.TestAutoAppend_AfterCloseKey_NoAutocomplete() line 15
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

I have reverted its use in [d4badc0](https://github.com/gui-cs/Terminal.Gui/pull/4222/commits/d4badc07c9932c1d08f0eb757d21aac7e15926ec)

tznind:
Behold a green tick!

BDisp:
What's wrong is the test sending a 'f' instead a 'F'. What's differentiate upper from lower case is the shift flag. So, you are doing the same wrong. Only replace the test with 'F'.

BDisp:
> What's wrong is the test sending a 'f' instead a 'F'. What's differentiate upper from lower case is the shift flag. So, you are doing the same wrong. Only replace the test with 'F'.

Sorry, I was wrong by said to replace 'f' to 'F'. We need to understand what is going on here. Only `WindowsConsole.InputRecord` get the correct key and modifiers. With `ConsoleKeyInfo` in `Windows` with VTS enabled or in `Unix`, the key and modifiers flags aren't available by the input driver. They must be manipulate to return the correct key and modifiers flags, based in cap case or escape sequence. See the images bellow using with the `NetDriver`.

'f' was pressed and Key is None, KeyChar is 'f' and Modifiers is None:

Captura de ecrã 2025-08-19 115218

After internally managed and before sending to `Application`, Key is F, KeyChar is 'f' and Modifiers is None:

Captura de ecrã 2025-08-19 120512

'F' was pressed and Key is None, KeyChar is 'F' and Modifiers is None:

Captura de ecrã 2025-08-19 115319

After internally managed and before sending to `Application`, Key is F, KeyChar is 'F' and Modifiers is Shift:

Captura de ecrã 2025-08-19 120706

See that the correct Key (of `ConsoleKeyInfo`) and Modifiers is sent to the `MapKey` method. As you can see the `map` variable has always the upper case if it belong to the A-Z range. In the `Key` class the `AsRune` property return the correct `Rune` based on the Shift flag.

But I agree that when we are sending a set of string, like "Bob.csv" the output must also to be the same as "Bob.csv". What is need is to send the correct `ConsoleKeyInfo` properties before call the `MapKey` method.

BDisp:
The PR https://github.com/tznind/gui.cs/pull/184 leverage the hard coded `EscSeqUtils.MapKey` and `EscSeqUtils.MapConsoleKeyInfo` methods, which has already all or almost all the key combinations.

tznind:
I think iteration is running on more than 1 thread.

I am exploring in my `iteration-zero` branch.  Where i set the post iteration delay to zero.

We can see for any fluent test it can crash in many ways in layout and in draw.

I need to add thread id tracking to main loop to see if it is ever hit from more than 1 thread.

Critically i can trigger the error just running a single test over and over so its not knock on effects of multiple tests at once

See: https://github.com/gui-cs/Terminal.Gui/pull/4225

I will start by detecting any time iteration is on other thread id, but then also look at layout or draw.

If any of them is running on non main UI thread it is bad.

BDisp:
One way I used to resolve the fluent unit tests, is in the PR #4210.
Other issue I found with ApplicationV2 unit tests I resolve with the PR #4223, which only was failing on Ubuntu and macOS with the error "Input loop exited during startup instead of entering read loop properly (i.e. and blocking)". The problem here is the _input variable being still null when try to build the v2 driver facade.
You can see that all unit tests are passing but it's important that all main loop iterations are running in the UIThread indeed.

tig:
@tznind please fix conficts and I'll review.

tznind:
This work stands alone and is good but will be further extended in #4225  where I deal with all the secret multi threading issues it turns out there are in the fluent test.

Basically any time you do anything on the 'test' thread then you might trip over what is going on in the 'run thread' that the fluent sets up.

I fix that by removing frame limiter on the main loop so it iterates constantly.  That is still WIP though.

tig:
@tznind Please resovle conflicts.

tznind:
This is subset of iteration-zero #4225  so will close for now as that is better PR to do and has no conflicts

PR #4225, opened 2025-08-21
Fixes #3947 Adds Fake driver and fixes fluent tests (iteration-zero)

### LINKED ISSUE #3947: Replace `FakeDriver` with a `v2fake` or equivalent

tznind:
Removes iteration timer on fluent tests to reveal multi threading issues and fixes them

- Tests should no longer fail randomly
- Should complete faster (~8s for 94 tests)
- When one test fails it should no longer bring down all future tests! (_coordinator was unexpectedly null)

Note this is super set of #4222 

## Fixes

- Fixes #3947

## Proposed Changes/Todos

Support creating `Application` and `Driver` level tests for v2.

- Driver level tests are those that use `[SetupFakeDriver]`
- Application level tests are those that use `[AutoInitShutdown]`

These tests need:
- To 'work' 
- To be able to resize 

### Implementation
To enable this workflow we first need to pull all the Dependency Injection elements out into a shared provider factory.  This lets us create classes but swap out Mocks and fakes as needed.

- Add IComponentFactory for creating system internal classes
  - Input
  - Output
  - WindowSize
  - etc

## Pull Request checklist:


tznind:
This is great branch for debugging integration tests.

Basically they fail whenever any UI op e.g Add or Remove happens on non UI thread as draw trips over changes to view stack etc.

tznind:
Definetly all the issues seem to come from running operations on other threads than the one the iterations are running on e.g. Add etc. 

I seem to be able to fix it with moving things onto the main thread using invoke.

tznind:
Something is trying to load windows kernel32 in linux, not sure what

BDisp:
> Something is trying to load windows kernel32 in linux, not sure what

Some method is calling Win32 API in Linux. If it's calling inside a try/catch then it's only reporting that and doesn't fail in the unit tests, but if it isn't calling inside a try/catch then the unit test will fail.

tig:
@tznind please fix conficts

tznind:
Done

Examples/UICatalog/Scenarios/SendKeys.cs
```diff
@@ -39,7 +39,7 @@ public override void Main ()
 
         txtResult.KeyDown += (s, e) =>
                              {
-                                 rKeys += (char)e.KeyCode;
+                                 rKeys += new Rune ((uint)(e.KeyCode & ~KeyCode.AltMask & ~KeyCode.CtrlMask & ~KeyCode.ShiftMask));
```
tig: Please use `Key`, not `KeyCode`. `Key` has helpers for this. Since this is an example, we don't want to be encouraging devs to use `KeyCode`.

Terminal.Gui/Drivers/V2/InputProcessor.cs
```diff
@@ -179,6 +180,22 @@ internal bool IsValidInput (Key key, out Key result)
         if (_highSurrogate > 0 && char.IsLowSurrogate ((char)key))
         {
             result = (KeyCode)new Rune (_highSurrogate, (char)key).Value;
+
+            if (key.IsAlt)
+            {
+                result = new (result.KeyCode | KeyCode.AltMask);
```
tig: `result = result.WithAlt`

tznind: Done in [14c68e5](https://github.com/gui-cs/Terminal.Gui/pull/4225/commits/14c68e550f04bce301b1e297d93e65ea1fc76070)

Examples/UICatalog/Scenarios/SendKeys.cs
tznind: Done in [14c68e5](https://github.com/gui-cs/Terminal.Gui/pull/4225/commits/14c68e550f04bce301b1e297d93e65ea1fc76070)

tznind:
@tig have addressed feedback and fixed the warnings from xmldoc and also put each class in its own file (tidyup).

And always green ticks from CI 💪 - no more random failures

And as added bonus the fluent tests now finish in only ~9s now that frame limiter off

PR #4234, opened 2025-09-01
Fixes #4229 Support transparent foreground/background in v2

### LINKED ISSUE #4229: V2 the default background of the app is not transparent.
with latest v2, the background color of tha app, is not transparent by default, if I force to use Application.Force16Colors = true; it can be tranparent, but I lose the ability to draw sixel image. the default scheme's background should be transparent instead of set any color.

Image

Image
if we want to keep the default color scheme as currently, how can I change the background of my app to be transparent.

tznind:
## Fixes #4229 - Transparent foreground/background

This is the simplest way to get transparent background.  It uses console native background on `Color.Transparent` (0,0,0,0).

Applies to both net and win

```
dotnet run -- -d v2
```
- Fixes #4229

When the background color is Transparent we use `CSI 49m` instead of explicitly mark it 0,0,0,0 color.  Similarly for foreground color.

image

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drawing/Scheme.cs
+++ b/Terminal.Gui/Drawing/Scheme.cs
@@ -306,7 +306,7 @@ public record Scheme : IEqualityOperators<Scheme, Scheme, bool>
                                VisualRole.Focus =>
                                    GetAttributeForRoleCore (VisualRole.Normal, stack) with
                                    {
-                                       Foreground = GetAttributeForRoleCore (VisualRole.Normal, stack).Background,
+                                       Foreground = SwapTransparentForBlack(GetAttributeForRoleCore (VisualRole.Normal, stack).Background),
                                        Background = GetAttributeForRoleCore (VisualRole.Normal, stack).Foreground
                                    },
 
@@ -369,6 +369,11 @@ public record Scheme : IEqualityOperators<Scheme, Scheme, bool>
         stack.Remove (role);
 
         return result;
+    }
+
+    private Color SwapTransparentForBlack (Color c)
+    {
+        return c == Color.Transparent ? Color.Black : c;
     }
 
     /// <summary>


metasong:
thank you for this fix, it looks like a creative patch when set the background color to explicit black (default console color)!
I would like to see similar function in WindowsDriver

tig:
Based on this, I think we should consdider changing the Default theme's Base BG to black.

Like this:

WindowsTerminal_oJOpu3MFX6

In Sheme.cs:

image

tig:
BTW, this makes me uncomfortably happy.

tznind:
We could also just say if the background color has Alpha of 0 we do this.

And add a color member for Transparent  if we dont already have one.

That might be more elegant as it would let you still use black as normal and not have surprises e.g. if users terminal default color is white for bg

tznind:
`View_Resolves_Attributes_From_Scheme` is failing.  It seems using explicit `Color` (Transparent) isn't how the schemes are supposed to work.

I guess we should instead be using the `StandardColor` for Transparent instead... but those don't have alpha so.... I need to dig into it a bit more.

tznind:
Ok StandardColor being vanilla enum means you can't have alpha modelled. So all current values are just actually RGB.  And in fact when passed to Color struct, results in an A of 0!

I've changes it to uint enum and added 0xFF000000 to all values except Transparent. 

This should mean we get correct A values when mapping this enum to Color.  And hence we can correctly distinguish 0xFF000000 (black) from 0x00000000 (Transparent).

Probably will blow up a lot of unit tests though :sweat:

tznind:
So this test fails when setting background color to Transparent.  Its confusing because test seems only to be checking the Foreground but I'm changing the background - so can't understand why it would fail.

```csharp
    [Fact]
    public void View_Resolves_Attributes_From_Scheme ()
    {
        View view = new Label { SchemeName = "Base" };

        foreach (VisualRole role in Enum.GetValues<VisualRole> ())
        {
            Attribute attr = view.GetAttributeForRole (role);
            Assert.NotEqual (default, attr.Foreground); // Defensive: avoid all-defaults
        }

        view.Dispose ();
    }
```
Specifically:
```diff
        Scheme CreateBase ()
        {
            return new ()
            {
-                Normal = new (StandardColor.LightBlue, StandardColor.RaisinBlack)
+                Normal = new (StandardColor.LightBlue, StandardColor.Transparent)
            };
        }
```

tznind:
Ah I see the issue, `Focus` is probably defined as the inverse of `Normal` and hence when normal ends up with a transparent background the foreground becomes transparent and hence `default`

```csharp
// Derivation algorithm as documented
Attribute result = role switch
                   {
                       VisualRole.Focus =>
                           GetAttributeForRoleCore (VisualRole.Normal, stack) with
                           {
                               Foreground = GetAttributeForRoleCore (VisualRole.Normal, stack).Background,
                               Background = GetAttributeForRoleCore (VisualRole.Normal, stack).Foreground
                           },
```

tznind:
As simplest solution and probably most sensible I've just set this explicitly as a rule
[69b9c33](https://github.com/gui-cs/Terminal.Gui/pull/4234/commits/69b9c33cffeae001613b061ba07caefd3adfa47e)

Transparent now means 'clear color' - that means that for background your likely to get either black (default terminal color) or black + semi transparent (user has configured % transparency).  But transparent foreground is also 'clear color' but likely to be... well the inverse of the background anyway (i.e. white terminal with black text or black terminal with white text).

Terminal.Gui/Drawing/Color/Color.ColorExtensions.cs
```diff
@@ -45,7 +45,7 @@ static ColorExtensions ()
             { new Color(255, 0, 255), ColorName16.Magenta },
             { new Color(255, 255, 0), ColorName16.Yellow },
             { new Color(128, 128, 128), ColorName16.Gray },
-            { new Color(118, 118, 118), ColorName16.DarkGray },
+            { new Color(169, 169, 169), ColorName16.DarkGray },
```
BDisp, line 48: That correspond adding 20% of the opacity (118+255*0.20=118+51=169), which may be the opacity percentage used by the `StandardColor`? It's possible to comprove with a unit test the opacity percentage (20%) for all `StandardColor`?

Terminal.Gui/Drawing/Scheme.cs
```diff
@@ -306,7 +306,7 @@ private Attribute GetAttributeForRoleCore (VisualRole role, HashSet<VisualRole>
                                VisualRole.Focus =>
                                    GetAttributeForRoleCore (VisualRole.Normal, stack) with
                                    {
-                                       Foreground = GetAttributeForRoleCore (VisualRole.Normal, stack).Background,
+                                       Foreground = SwapTransparentForBlack(GetAttributeForRoleCore (VisualRole.Normal, stack).Background),
```
BDisp, line 309: Otherwise a text in a transparent foreground with a opaque background is unreadable, correct?

Terminal.Gui/Drivers/V2/OutputBase.cs
```diff
@@ -144,7 +144,44 @@ public virtual void Write (IOutputBuffer buffer)
...
-    protected abstract void AppendOrWriteAttribute (StringBuilder output, Attribute attr, TextStyle redrawTextStyle);
+    /// <summary>
+    /// Change the color being used to render text.
+    /// </summary>
+    /// <param name="output">The output stream</param>
+    /// <param name="attr">The foreground and background color to switch to</param>
+    /// <param name="redrawTextStyle">Style</param>
+    protected virtual void AppendOrWriteAttribute (StringBuilder output, Attribute attr, TextStyle redrawTextStyle)
+    {
+        if (attr.Foreground == Color.Transparent)
```
BDisp, line 155: Please, don' forget to consider the `Force16Colors` in v2net:

```cs
        if (Application.Force16Colors)
        {
            output.Append (EscSeqUtils.CSI_SetForegroundColor (attr.Foreground.GetAnsiColorCode ()));
            output.Append (EscSeqUtils.CSI_SetBackgroundColor (attr.Background.GetAnsiColorCode ()));
        }
        else
        {
            ...
        }
```

BDisp (changes requested):

tig:
@tznind see https://github.com/gui-cs/Terminal.Gui/pull/4285

Terminal.Gui/Drawing/Color/Color.cs
```diff
@@ -19,9 +19,14 @@ namespace Terminal.Gui.Drawing;
...
 [JsonConverter (typeof (ColorJsonConverter))]
 [StructLayout (LayoutKind.Explicit)]
-public readonly partial record struct Color : ISpanParsable<Color>, IUtf8SpanParsable<Color>, ISpanFormattable,
+public readonly partial record struct  Color : ISpanParsable<Color>, IUtf8SpanParsable<Color>, ISpanFormattable,
                                               IUtf8SpanFormattable, IMinMaxValue<Color>
 {
+    /// <summary>
+    /// Transparent color (0 for all elements). This is interpreted as default console color.
+    /// </summary>
+    public static readonly Color Transparent = new Color (0, 0, 0, 0);
```
mrazza, line 28: You discuss this problem elsewhere, but may I suggest naming this "None" or "ConsoleDefault"?

While the motivation for this change is, indeed, transparency, the actual behavior as implemented is to switch to the console's default color scheme which has no requirement for transparency. If you look at an application like vim you set the color (in Vim's color scheme) to NONE for this.

e.g.: `hi Normal guibg=NONE ctermbg=NONE`

HueByte:
Any updates on the topic? Working on the app and could really use the transparent background - it is a huge UI improvement

tig:
Gonna close this in favor of #4740 .

PR #4251, opened 2025-09-16
Fixes #4250. Add a v2unix driver

### LINKED ISSUE #4250: Add a v2unix driver
I would like to propose adding a new v2unix driver for Unix systems. It only use native Unix libraries. Old Curses is totally absent in this driver. `System.Console` is only used in the `Suspend` method which it's common to v2net and v2unix. Works on Linux and macOS.

BDisp:
## Fixes

- Fixes #4250

## Proposed Changes/Todos

- [x] Uses Unix native libraries

## Pull Request checklist:


Terminal.Gui/Drivers/V2/ApplicationV2.cs
```diff
@@ -92,13 +93,17 @@ private void CreateDriver (string? driverName)
...
+            _coordinator = CreateUnixSubcomponents ();
+        }
         else if (p == PlatformID.Win32NT || p == PlatformID.Win32S || p == PlatformID.Win32Windows)
         {
             _coordinator = CreateWindowsSubcomponents ();
         }
         else
         {
-            _coordinator = CreateNetSubcomponents ();
+            _coordinator = CreateUnixSubcomponents ();
```
tznind: This changes the default driver on linux/macos to the new unix one. Worth noting.

Terminal.Gui/Drivers/V2/IInputProcessor.cs
```diff
@@ -25,6 +25,11 @@ public interface IInputProcessor
     /// <summary>Event fired when a mouse event occurs.</summary>
     event EventHandler<MouseEventArgs>? MouseEvent;
 
+    /// <summary>
+    /// Gets the name of the driver associated with this input processor.
+    /// </summary>
+    string DriverName { get; init; }
```
tznind, line 31: This class isn't the right place for this, input processors care about processing input not storing info about what the current driver is.

Terminal.Gui/Drivers/V2/UnixKeyConverter.cs
```diff
@@ -0,0 +1,20 @@
...
+///     (used by <see cref="View"/> etc).
+/// </summary>
+internal class UnixKeyConverter : IKeyConverter<char>
+{
+    /// <inheritdoc />
+    public Key ToKey (char value)
+    {
+        ConsoleKeyInfo adjustedInput = EscSeqUtils.MapChar (value);
+
+        return EscSeqUtils.MapKey (adjustedInput);
```
tznind, line 18: Does operating purely on char cause issues later on for surrogate pairs? Or is ok?

tznind:
What is the benefits of these lib methods vs pure ansi?

Is it faster? More compatible?

Terminal.Gui/Drivers/V2/ApplicationV2.cs
BDisp: That was the intention. Since pure native Win32 API is the default for Windows, I though that pure native Unix library should be the default for Linux, Mac and all Unix like systems. Unless it can be proved that v2net has more performance than v2unix. If it isn't the case then v2net can be an optional cross platform alternative.

Terminal.Gui/Drivers/V2/UnixKeyConverter.cs
BDisp: Absolutely works as expected because like all the drivers surrogate pairs inputs are always considered the join of two chars (high and low surrogates). You can copy and paste by keyboard or context menu a surrogate pair in a TextField or TextView without any issue.

Terminal.Gui/Drivers/V2/IInputProcessor.cs
BDisp: I choose this place because `ConsoleDriverFacade` use the `InputProcessor` property in the `GetVersionInfo` method to get driver information. What define the driver kind is the input type, `InputRecord` for v2win, `ConsoleKeyInfo` for v2net and `char` for v2unix. What place do you suggest?

BDisp:
> What is the benefits of these lib methods vs pure ansi?
> 
I didn't understood quite well your question. I used pure ansi in all places where the other drivers use it. The lib methods used here are for leverage native unix libs in the implementation.

For e.g. to set console raw mode which allow intercept all the input keys, even Ctrl+C:
- v2win use console modes.
- v2net use `Console.TreatControlCAsInput = true;`.
- v2unix use the `EnableRawModeAndTreatControlCAsInput` method.

So, no pure ansi are used above. But you may referring something that I don't know yet.

> Is it faster? More compatible?

Honestly I don't know. Needed to test with benchmark to compare it. But it seems very faster. I also think it's more compatible than `System.Console` which not all of their objects are Unix compatible.

Terminal.Gui/Drivers/V2/InputProcessor.cs
```diff
@@ -96,7 +99,8 @@ public void OnMouseEvent (MouseEventArgs a)
     ///     Key converter for translating driver specific
     ///     <typeparamref name="T"/> class into Terminal.Gui <see cref="Key"/>.
     /// </param>
-    protected InputProcessor (ConcurrentQueue<T> inputBuffer, IKeyConverter<T> keyConverter)
+    /// <param name="driverName"></param>
+    protected InputProcessor (ConcurrentQueue<T> inputBuffer, IKeyConverter<T> keyConverter, string driverName)
```
tig: Passing the driver name is redundant. The implementation should set `InputProcesor.DriverName` in it's constructor.

Terminal.Gui/Drivers/V2/UnixInputProcessor.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Terminal.Gui.Drivers;
+
+/// <summary>
+///     Input processor for <see cref="UnixInput"/>, deals in <see cref="char"/> stream.
+/// </summary>
+internal class UnixInputProcessor : InputProcessor<char>
+{
+    /// <inheritdoc />
+    public UnixInputProcessor (ConcurrentQueue<char> inputBuffer) : base (inputBuffer, new UnixKeyConverter (), "unix") { }
```
tig: Just set `DriverName` to "unix" here in the constructor. No need to pass a string around when the KeyConverter and InputProcessor are strongly typed.

Terminal.Gui/Drivers/V2/ApplicationV2.cs
```diff
@@ -154,6 +159,29 @@ private IMainLoopCoordinator CreateNetSubcomponents ()
                                                         cf);
     }
 
+    private IMainLoopCoordinator CreateUnixSubcomponents ()
```
tig: This code should be encapsulated in each driver implementation, not in Application. Can you please figure out how to do so?

tig: Same for the WIndows and Net factory code, too, of course.

tznind: The application creates the main loop.  Main loop is boot strapped by IMainLoopCoordinator.

`IConsoleDriver` is a wrapper abstraction into this system designed for compatibility (that's why its called facade in v2)

In adding support for FakeDriver v2 I added component factory.  This functions as DI provider for the coordinator and its sub components and allows substituting out specific components as needed e.g. for mocks.

So I think this being in Application is correct for now. 

 I can look into simplifying as part of removing the old drivers

Terminal.Gui/Drivers/EscSeqUtils/EscSeqUtils.cs
```diff
@@ -1131,6 +1141,17 @@ public static ConsoleKeyInfo MapConsoleKeyInfo (ConsoleKeyInfo consoleKeyInfo)
...
+                break;
+            case uint n when n is >= 28  and <= 31:
+                key = (ConsoleKey)(char)(consoleKeyInfo.KeyChar + 24);
+
+                newConsoleKeyInfo = new (
+                                         (char)key,
+                                         key,
+                                         (consoleKeyInfo.Modifiers & ConsoleModifiers.Shift) != 0,
+                                         (consoleKeyInfo.Modifiers & ConsoleModifiers.Alt) != 0,
+                                         true);
```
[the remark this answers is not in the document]
BDisp: The console keys in the range 52-55 are valid. They represent the keys from D4 to D7 and never produce values outside the valid ConsoleKey enum range.

Terminal.Gui/Drivers/AnsiResponseParser/Keyboard/EscAsAltPattern.cs
```diff
@@ -8,7 +8,7 @@ internal class EscAsAltPattern : AnsiKeyboardParserPattern
     public EscAsAltPattern () { IsLastMinute = true; }
 
 #pragma warning disable IDE1006 // Naming Styles
-    private static readonly Regex _pattern = new (@"^\u001b([a-zA-Z0-9_])$");
+    private static readonly Regex _pattern = new (@"^\u001b([\u0001-\u001a\u001f\a-zA-Z0-9_])$");
```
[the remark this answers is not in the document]
BDisp: Fixed in the https://github.com/gui-cs/Terminal.Gui/pull/4251/commits/57f3d4646a32c7d80ac5dd045854bfdda0f50dbb.

Terminal.Gui/Drivers/V2/InputProcessor.cs
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4251/commits/f3d55b8db8ce7d732f4f0b9df79fa678d64ab45f.

Terminal.Gui/Drivers/V2/ApplicationV2.cs
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4251/commits/167d65cd1d2c7cd45ff479c3faf71d1b1549e5f7.

Encapsulated in each driver implementation is also redundant because the code is basically the same on all drivers, except the type. With a generic helper method is a better solution reducing duplicate code.

Terminal.Gui/Drivers/V2/UnixInputProcessor.cs
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4251/commits/f3d55b8db8ce7d732f4f0b9df79fa678d64ab45f.

PR #4254, opened 2025-09-22
add more parameters to TableView's renderCell function

metasong:
when override the cell rendering, we do not know which cell is rendering


tznind:
Hmn... for v2 we should probably prioritise using the events system for this. 

Add a DrawingCell event for example.

That would be preferable to override I think.

@tig thoughts?

@metasong would you be up for having a go at this? I can assist 

Benefits include:

- Future proofing for parameters (since they would be event args instead)
- No need to subclass to change behaviour 

Proposition would be to make it a cancelable event. User would set cancel to true and call relevant methods to draw themselves

tig:
I concur.

tig:
Closing as stale. TableView has been significantlyi refactored.

PR #4255, opened 2025-09-26
Fixes #4233 - CharMap rendering

### LINKED ISSUE #4233: CharMap rendering broken (differently on different drivers)
WindowsDriver:

Image

v2:

Image

v2 is better, but there're still issues.

tig:
## Fixes

- Fixes #4233

## Proposed Changes/Todos

- [x] Figure out what's going on - Apps that output codepoints that are not normally rendered, need special case logic using `UnicodeCategory`
- [x] Fix it
- [x] Fix remaining issues. Eg `Hexagram For The Creative Heaven - ䷀ U+04dc0`


tig:
There are now only ~10 codepoints that render incorrectly. I can't figure out what's the deal with them. Eg.

WindowsTerminal_nMJ4eu4B8R

tig:
Ok, I've figured out the remaining rendering issues.

There are a set of codepoints, all in the "OtherSymbol" category that unicode (and wcwidth) think are 1 column wide, but that Windows Terminal treats as 2 columns. 

The fix is to override `wcwidth` in our `Rune.GetColumns` extension thusly:

```cs
    public static int GetColumns (this Rune rune)
    {
        int value = rune.Value;

        // HACK: This is a workaround for the fact that WT treats these glyphs as 2 columns wide
        // HACK: when, unicode (and wcwidth) say they are 1.
        // HACK: See https://github.com/gui-cs/Terminal.Gui/pull/4255 for more.
        // Check I Ching symbol ranges (hardcoded for reliability)
        if ((value >= 0x2630 && value <= 0x2637) ||  // Trigrams
            (value >= 0x268A && value <= 0x268F) ||  // Monograms/Digrams
            (value >= 0x4DC0 && value <= 0x4DFF))    // Hexagrams
        {
            return 2; // Assume double-width due to Windows Terminal font rendering
        }

        // Fallback to original GetWidth for other code points
        return UnicodeCalculator.GetWidth (rune);
    }
```

I've enhanced CharMap to make it easy to only show codepoints that are interesting. 

Before the GetColumns fix:
WindowsTerminal_K1LaE4BBJL


After:
WindowsTerminal_33xELxNggI

tig:
https://github.com/microsoft/terminal/issues/19389

BDisp (approved):
Very nice.

PR #4275, opened 2025-10-06
Fixes #4274. Using Windows Host Console v2win is rendering window size badly using VSDebugConsole.exe

### LINKED ISSUE #4274: Using Windows Host Console v2win is rendering window size badly using VSDebugConsole.exe
The cause is the `Console.Out.Write (EscSeqUtils.CSI_SaveCursorAndActivateAltBufferNoBackscroll);` being called from the `WindowsOutput` constructor, while in the `WindowsDriver` is called from the `Init` method. But the v2 doesn't have that method.

With `WindowsDriver` is rendering as expected:

Image

With v2win the right vertical line is hide by the scroll bar:

Image

BDisp:
## Fixes

- Fixes #4274

## Proposed Changes/Todos

- [x] Check if it's running from VS 2022 console.

## Pull Request checklist:

### DIFF
--- a/Tests/IntegrationTests/UICatalog/ScenarioTests.cs
+++ b/Tests/IntegrationTests/UICatalog/ScenarioTests.cs
@@ -92,11 +92,6 @@ public class ScenarioTests : TestsAllViews
             {
                 Application.Iteration += OnApplicationOnIteration;
                 initialized = true;
-
-                lock (_timeoutLock)
-                {
-                    timeout = Application.AddTimeout (TimeSpan.FromMilliseconds (abortTime), ForceCloseCallback);
-                }
             }
             else
             {
@@ -127,6 +122,15 @@ public class ScenarioTests : TestsAllViews
 
         void OnApplicationOnIteration (object? s, IterationEventArgs a)
         {
+            if (iterationCount == 0)
+            {
+                // Start the timeout countdown on the first iteration
+                lock (_timeoutLock)
+                {
+                    timeout = Application.AddTimeout (TimeSpan.FromMilliseconds (abortTime), ForceCloseCallback);
+                }
+            }
+
             iterationCount++;
 
             if (Application.Initialized)


tig:
@BDisp please fix conflicts.

BDisp:
> @BDisp please fix conflicts.

Done.

tig:
Perhaps you  can figure out the nterminttent test failure?

BDisp:
> Perhaps you can figure out the nterminttent test failure?

Yes, I'll try to fix it. It's due the `Application.ClearScreenNextIteration` setting to `true` when there is only one `TopLevel`. Thus it's better to trust in  `Application.LayoutAndDraw` method. Didn't pass in the CI but pass locally. Again, some more racing condition.

tig:
Yeah, it would be awesome if you could look at it. 

I have a feeling we introduced some more static related race conditions with the PR. I just merged.  I saw one failure that was involved with colors being incorrect in a different unit test.

I wish I could spend more time on it right now, but it's really hard for me to type.

BDisp:
@tig I suspect it's the `UpdateFromJson` unit test that doesn't restore the color scheme and causing wrong assertion color in the `HideDropdownListOnClick_True_Highlight_Current_Item` unit test. Can you take a look, please?

tig:
I will in the morning. Thanks.

tig:
@BDisp I think I've found a fault in CM ResetToHardCodedDefaults and am now working on creating a small reproducable test. Just FYI.

BDisp:
> @BDisp I think I've found a fault in CM ResetToHardCodedDefaults and am now working on creating a small reproducable test. Just FYI.

Thanks. I fixed the one that always was failing when I run the `UpdateFromJson` and the `HideDropdownListOnClick_True_Highlight_Current_Item` unit tests together but probably there is more cases where the CM isn't correctly resetting.
Another sporadic failures is in the `All_Scenarios_Quit_And_Init_Shutdown_Properly` which sometimes some scenario doesn't quit the app after the elapsed time.

tig:
> > @BDisp I think I've found a fault in CM ResetToHardCodedDefaults and am now working on creating a small reproducable test. Just FYI.
> 
> Thanks. I fixed the one that always was failing when I run the `UpdateFromJson` and the `HideDropdownListOnClick_True_Highlight_Current_Item` unit tests together but probably there is more cases where the CM isn't correctly resetting.

How did you fix it? I still see it failing intermittently.

```
    HideDropdownListOnClick_True_Highlight_Current_Item [0:00.042] Failed: Assert.Equal() Failure: Values differ
Xunit.Sdk.EqualException
Expected: '0'
Actual:   '2'
   at UnitTests.DriverAssert.AssertDriverAttributesAre(String expectedLook, ITestOutputHelper output, IConsoleDriver driver, Attribute[] expectedAttributes) in C:\Users\Tig\s\gui-cs\Terminal.Gui\Tests\UnitTests\DriverAssert.cs:line 130
   at Terminal.Gui.ViewsTests.ComboBoxTests.HideDropdownListOnClick_True_Highlight_Current_Item() in C:\Users\Tig\s\gui-cs\Terminal.Gui\Tests\UnitTests\Views\ComboBoxTests.cs:line 560

Disabling CM to clean up.
     ▼                                                                          
One                                                                             
Two                              `                                               
Three                                                                           
                                                                                
                            
Unexpected Attribute at Contents[0,0] = [' ':[BrightGreen,Black,None]].
 Expected: 0 ([White,Aqua,None])
  But Was: 2 ([BrightGreen,Black,None])
```

BDisp:
I mean with the two related unit tests together. Locally I'm not see more failures related with this but I don't know why it fail in the CI. I'm using the test explorer in Windows and also testing with WSL.

Terminal.Gui/Configuration/ConfigurationManager.cs
```diff
@@ -340,6 +340,8 @@ internal static void ResetToCurrentValues ()
     [RequiresDynamicCode ("AOT")]
     internal static void ResetToHardCodedDefaults ()
     {
+        ResetToCurrentValues ();
```
tig: This does not make sense to do. Please remove this line.

Terminal.Gui/Configuration/ConfigurationManager.cs
```diff
@@ -340,6 +340,8 @@ internal static void ResetToCurrentValues ()
...
+        ThrowOnJsonErrors = true;
```
tig: ThrowOnJsonErros is false by default AND is a CM property. Thus calling LoadHardCodedDefaults should reset it to false. Please remove this line.

tig:
Please stop working on this right now. You're wasting my time by throwing invalid suggestions and code at me right while I/m in the middle of debugging it. You asked me to help... please let me  help. If I need any input or suggestions for you on this before I figure it out, I will ask. Thanks.

BDisp:
You're really stupid. If I've done something wrong, there are good ways to say so without offending. If you don't want my cooperation, say so right away, but I don't and never will tolerate any rudeness.

tig:
> You're really stupid. If I've done something wrong, there are good ways to say so without offending. If you don't want my cooperation, say so right away, but I don't and never will tolerate any rudeness.

Where in my reply did I say something that was offensive? Please be specific. For example, did *I* literally call you stupid? 

This has happened in the past where I have been working on something and you have decided to work on the same thing at the same time. I have told you, very patiently, and very clearly that it is distracting and makes the work I'm doing harder.

In this situation you did the same thing again. It is you who are being disrespectful by continuing to exhibit behavior I have begged you (politely and repeatedly) to chang.

I wasted a half hour of very valuable keyboarding time this morning trying to figure out what you meant by

> Thanks. I fixed the one that always was failing when I run the UpdateFromJson and the HideDropdownListOnClick_True_Highlight_Current_Item unit tests together but probably there is more cases where the CM isn't correctly resetting.

This was because I didn't see that you had pushed more commits to this PR after I made it clear I was working on it. 

Yes, I'm extra grumpy because my shoulder hurts like hell and the time I can spend at the keyboard is limited. But that's no excuse for you calling me stupid and believing that *I* was attacking you personally, when I was just asking you stop working on this for now.

Re-read my post above and tell me precisely which words or phrases in it were offensive or rude.

BDisp:
When I called you stupid, it was because of how you express yourself when you tell me what you think, in a somewhat rude way. Therefore, it's because of the way you address me. On the other hand, I had already submitted my edits when you told me that you had probably already discovered the CM problem. You were upset because you thought I had submitted them after receiving your information, but that wasn't the case, as you can see from the order in which I submitted them in relation to your information. That said, and because I don't usually be rude to anyone, nor do I speak the way you do, even if they offend me, I said you were being stupid for the way you spoke to me. I apologize if I offended you, but I also don't accept your way of addressing me because, naturally, you wouldn't appreciate it if I spoke to you the same way. There are many ways to offend. I'm sorry you're going through a painful time, and I respect that, but also understand that I also have feelings and some behaviors I don't accept. I wish you a speedy recovery.

Terminal.Gui/Configuration/ConfigurationManager.cs
BDisp: Done.

Terminal.Gui/Configuration/ConfigurationManager.cs
BDisp: Done.

BDisp:
All tests passed now. Cheers.

Tests/UnitTests/Views/ComboBoxTests.cs
```diff
@@ -525,7 +525,9 @@ public void HideDropdownListOnClick_True_Highlight_Current_Item ()
         Assert.True (cb.IsShow);
         Assert.Equal (-1, cb.SelectedItem);
         Assert.Equal ("", cb.Text);
-
+        Assert.False (Application.ClearScreenNextIteration);
```
tig, line 528: I think this just hides the problem. 

In my debugging of this, I have am pretty sure (but not 100%) that I've seen that CM is not in the right state after the UpdateFromJson test runs. 

Can you explain how what you did here fixes the actual bug or just hides it? Thanks.

tig (changes requested):
I have a question...

Tests/UnitTests/Views/ComboBoxTests.cs
tig: This test (after I fixed the Color.ToString bug) fails, illustrating that CM is not cleaning up correctly.

```cs
    [Fact]
    public void UpdateFromJson_DisableTrue_NewView_Normal_Attribute_Matches_Baseline_TopLevel_Scheme ()
    {
        Assert.True (IsInitialized ());
        Assert.False (IsEnabled);

        // Capture hardCoded hard-coded TopLevel scheme colors
        var hardCodedSchemes = SchemeManager.GetSchemes ();
        var hardCodedTopLevelNormalFg = hardCodedSchemes ["TopLevel"].Normal.Foreground;
        var hardCodedTopLevelNormalBg = hardCodedSchemes ["TopLevel"].Normal.Background;

        Assert.Equal (new Color (StandardColor.CadetBlue), hardCodedTopLevelNormalFg);
        Assert.Equal (new Color (StandardColor.Charcoal), hardCodedTopLevelNormalBg);

        // Mutate via UpdateFromJson (applies runtime theme), then Disable(true)
        var cmt = new ConfigurationManagerTests (output);
        cmt.UpdateFromJson ();



        // Create fresh UI and verify the view’s Normal attribute matches hardCoded scheme
        var top = new Toplevel ();

        var attr = top.GetAttributeForRole (VisualRole.Normal);

        // Use ToString so Assert.Equal shows the actual vs expected values on failure
        Assert.Equal (hardCodedTopLevelNormalFg.ToString (), attr.Foreground.ToString ());
        Assert.Equal (hardCodedTopLevelNormalBg.ToString (), attr.Background.ToString ());

        top.Dispose ();
    }
```

tig:
Note, I re-ran the Action and it failed. So there are still issues, they are just intermittent.

BDisp:
> Note, I re-ran the Action and it failed. So there are still issues, they are just intermittent.

You're right. I was lucky it didn't fail, as the error is intermittent. It does seem like CM isn't properly resetting to its original configuration after running the UpdateFromJson unit test, because I detected that the json variable configuration was present in the HideDropdownListOnClick_True_Highlight_Current_Item unit test. But the failure is intermittent, as it often doesn't fail on my machine, but after committing a change in the PR, it fails in CI.

Tests/UnitTests/Views/ComboBoxTests.cs
BDisp: > I think this just hides the problem.
> 
Initially I thought it was Application.ClearScreenNextIteration being set to true, so I added that assertion to clear up any doubts.

> In my debugging of this, I have am pretty sure (but not 100%) that I've seen that CM is not in the right state after the UpdateFromJson test runs.
> 
I confirm, I had already noticed this and informed you about it when I was running only the UpdateFromJson and HideDropdownListOnClick_True_Highlight_Current_Item unit tests at the same time.

> Can you explain how what you did here fixes the actual bug or just hides it? Thanks.

As you already know, I haven't really been able to resolve this issue because it's related to the CM, which I'd rather leave for you to resolve because you know it very well and will be able to better discover what the problem is.


> This test (after I fixed the Color.ToString bug) fails, illustrating that CM is not cleaning up correctly.

I'll take a look at it and try to help, if you want me to, of course.

tig:
Yes. And I'm getting closer to understanding the root cause.

Tests/UnitTests/Views/ComboBoxTests.cs
tig: No, I'd rather be able to just focus on it myself. I'll keep you posted on my progress.

tig:
Shit. I didn't mean to merge this.

tig:
I'm going to create a new PR based on this branch. Sorry for the screwup.

PR #4301, opened 2025-10-21
Fixes #4300 - Scrolling intermittently causes All_Scenarios_Quit_And_Init_Shutdown_Properly to fail

### LINKED ISSUE #4300: `Scrolling` intermittently causes `All_Scenarios_Quit_And_Init_Shutdown_Properly` to fail
The failing job is caused by the test IntegrationTests.UICatalog.ScenarioTests.All_Scenarios_Quit_And_Init_Shutdown_Properly with scenarioType: typeof(UICatalog.Scenarios.Scrolling). The error message is:

Scenario 'Scrolling' Failed to Quit with Esc after 2200ms and 1 iterations. Force quit.

This indicates the test is trying to quit the scenario by sending an Esc key after some time and iterations, but the scenario does not respond as expected.

https://github.com/gui-cs/Terminal.Gui/actions/runs/18664588956/job/53212828095

tig:
## Fixes

- Fixes #4300

### DIFF
--- a/Examples/UICatalog/Scenarios/Scrolling.cs
+++ b/Examples/UICatalog/Scenarios/Scrolling.cs
@@ -10,7 +10,7 @@ namespace UICatalog.Scenarios;
 [ScenarioCategory ("Tests")]
 public class Scrolling : Scenario
 {
-    private static object? _progressTimer = null;
+    private object? _progressTimer = null;
 
     public override void Main ()
     {


tig:
Three things about Scrolling that are pecular:

1) It does seem to take a bit of time to load when started compared to other Scenarios
2) The timer code was (until I refactored it) suspect
3) It uses Unloaded to tear down the timer code. It's possible that in the v2 architecture we're not cleaning up timers correctly?

tig:
To try tot get this to repro, next push will change the test timeout from 2200 to something much lower.

tig:
Oh, it uses `AllViewsView`. Locally this scenario takes 2.7 seconds to run!

image

In the Action, it normally takes 1s

```cs
Passed IntegrationTests.UICatalog.ScenarioTests.All_Scenarios_Quit_And_Init_Shutdown_Properly(scenarioType: typeof(UICatalog.Scenarios.Scrolling)) [1 s]
```

The fix is to either simplify the scenario or lenghten the timeout.

I like the fact this Scenario stresses scroll functionality with a huge View. So I'm going to extend the time out to 5 seconds.

PR #4421, opened 2025-11-28
Fixes #4004. Driver "windows" broken in conhost and cmd

### LINKED ISSUE #4004: Driver "v2win" broken in `conhost`
Hello, 

Driver "v2win" is broken in version v2.0.0-develop.4401 when used in Console Host. It happens inside de VS IDE and now it also happens running it outside the IDE in Windows Explorer.

It shows escape codes along with a wrong rendering.

Image

Remapping the glyphs does not fix the issue, so seems not related to glyphs.

WindowsDriver works.

_Originally posted by @GitHubThor in https://github.com/gui-cs/Terminal.Gui/discussions/4003_

BDisp:
## Fixes

- Fixes #4004

## Proposed Changes/Todos

- [x] Add `IsVirtualTerminal` property to the `IApplication` interface.

## Pull Request checklist:


BDisp:
The 16 colors are still not rendered very well on selecting ListView, TableView and Menu.

BDisp:
The unit test failure is due a some race condition because it pass locally in Windows and Ubuntu.

tig:
> The unit test failure is due a some race condition because it pass locally in Windows and Ubuntu.

I'm confused how you are able to do anything with v2_develop as I broke it pretty completely. 😮

Terminal.Gui/Drivers/WindowsDriver/WindowsOutput.cs
```diff
@@ -146,11 +146,16 @@ public WindowsOutput ()
...
 
-            // Force 16 colors if not in virtual terminal mode.
-            // BUGBUG: This is bad. It does not work if the app was crated without
-            // BUGBUG: Apis.
-            //ApplicationImpl.Instance.Force16Colors = true;
-
+            try
+            {
+                // Force 16 colors if not in virtual terminal mode.
+                (ApplicationImpl.Instance as ApplicationImpl)!.IsVirtualTerminal = false;
```
tig: Please re do this such that it does not depend on the static Application Instance.

Terminal.Gui/App/IApplication.cs
```diff
@@ -597,6 +597,11 @@ public TView Run<TView> (Func<Exception, bool>? errorHandler = null, string? dri
     /// </remarks>
     IClipboard? Clipboard { get; }
 
+    /// <summary>
+    ///     Gets or sets whether <see cref="Driver"/> support for virtualized terminal sequences.
+    /// </summary>
+    public bool IsVirtualTerminal { get; }
```
tig: I think this is backwards. Virtual terminal is a characteristic of the driver, and not the App.

tig (changes requested):

BDisp:
> I'm confused how you are able to do anything with v2_develop as I broke it pretty completely. 😮

There are still many unresolved issues, but the v2_develop branch isn't unusable because I can compile and run it. For example, one problem I'm facing is not being able to test Force16Colors in UICatalog because the CheckBox event isn't triggered when it's clicked. Could you be more specific about why your branch is broken? Do you mean the one from the gui-cs repository or your own?

tig:
> > I'm confused how you are able to do anything with v2_develop as I broke it pretty completely. 😮
> 
> There are still many unresolved issues, but the v2_develop branch isn't unusable because I can compile and run it. For example, one problem I'm facing is not being able to test Force16Colors in UICatalog because the CheckBox event isn't triggered when it's clicked. Could you be more specific about why your branch is broken? Do you mean the one from the gui-cs repository or your own?

In my case not even UICatalog will run. 

I screwed up and the legacy and modern Application concepts are intermixed and thus pretty much everything is broken.

Terminal.Gui/Drivers/WindowsDriver/WindowsOutput.cs
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4421/commits/c7475099fc2813be0b9c0bb1892fff3243aa2dce.

Terminal.Gui/App/IApplication.cs
BDisp: The problem is that the driver is now divided into several interfaces (IDriver, IInput, IOutput, IOutputBuffer). If any of them needs a reference to the current IDriver instance, it can only access it through the static Application class. With a workaround, I somewhat circumvented this limitation. I hope you like it. Another limitation is that you can only create a driver by name; there's no option to create it first and then pass it to the app's initialization. Even if it were possible to pass the instantiated driver, it's recreated again when the app starts. It's like a spiderweb of problems.
Done in https://github.com/gui-cs/Terminal.Gui/pull/4421/commits/c7475099fc2813be0b9c0bb1892fff3243aa2dce.

BDisp:
On my keyboard these tests fails because I need to pass shift as true. So, it depends on keyboard type and the unit tests doesn't get the right keyboard layout. I suggest to comment them to avoid only pass on supported keyboard layouts.

image

BDisp:
The `Init_KeyBindings_Are_Not_Reset` unit test is sometimes failing because it really reset when a new keyboard instance is created.

Tests/UnitTestsParallelizable/Drivers/DriverTests.cs
```diff
@@ -92,6 +92,57 @@ public void All_Drivers_LayoutAndDraw_Cross_Platform (string driverName)
 
         app.Dispose ();
     }
+
+    [Fact]
```
tig: Please put these in a separate test class (e.g. `VirtualTerminalTests`).

Tests/UnitTestsParallelizable/Drivers/OutputBaseTests.cs
```diff
@@ -0,0 +1,222 @@
...
+        Assert.False (buffer.Contents! [0, 2].IsDirty);
+
+        // Verify SetCursorPositionImpl was invoked by WriteToConsole (cursor set to a written column)
+        Assert.Equal (new Point (0, 0), output.GetCursorPosition ());
+    }
+
+    [Theory]
+    [InlineData (true)]
+    [InlineData (false)]
+    public void Write_EmitsSixelDataAndPositionsCursor (bool isVirtualTerminal)
```
tig, line 162: I think this is more of a Sixel test than a Outputbase test?

Terminal.Gui/App/Application.Lifecycle.cs
```diff
@@ -67,7 +67,13 @@ public static int? MainThreadId
 
     /// <inheritdoc cref="IDisposable.Dispose"/>
     [Obsolete ("The legacy static Application object is going away.")]
-    public static void Shutdown () => ApplicationImpl.Instance.Dispose ();
+    public static void Shutdown ()
+    {
+        ApplicationImpl.Instance.Dispose ();
+
+        // Use the static reset method to bypass the fence check
+        ApplicationImpl.ResetStateStatic ();
```
tig: What bug does this fix?

tig (changes requested):
Looking good. Please address my feedback/questions.

Tests/UnitTestsParallelizable/Drivers/OutputBaseTests.cs
BDisp: I'm testing the following code in the OuttupBase and that why I put it here.

```cs
        if (Driver is { } && IsVirtualTerminal)
        {
            foreach (SixelToRender s in Driver?.Sixel!)
            {
                if (!string.IsNullOrWhiteSpace (s.SixelData))
                {
                    SetCursorPositionImpl (s.ScreenPosition.X, s.ScreenPosition.Y);
                    Write ((StringBuilder)new (s.SixelData));
                }
            }
        }
```

Terminal.Gui/App/Application.Lifecycle.cs
BDisp: I did that change because the static Application.Force16Colors wasn't being resetting. Now I'm using a backing field for it in the DriverImpl class and so this change isn't needed anymore. Revert it in the https://github.com/gui-cs/Terminal.Gui/pull/4421/commits/5a9f41233b74da23d0a9c3c8d455bcf77946e247

Tests/UnitTestsParallelizable/Drivers/DriverTests.cs
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4421/commits/a35bc168220cf4cc62a5dcf376be1b540aa99c89.

Terminal.Gui/App/Application.Lifecycle.cs
BDisp: @tig it wasn't only for the static Application.Force16Colors but also for the static static Application.ForceDriver. The Dispose only reset the properties of the app instance but not the static properties. See the assertion failure when exit from UICatalog, Note that this only happens with the DotNetDriver. Do you want I revert this again?

image

tig: No. I do not think ResetState should reset the static config properties. They should be stable across all app instances.

BDisp: But that the point. Calling the static Shutdown should reset the static Application properties by calling ResetStateStatic at end. Otherwise they will be inconsistent.

**Edit:**
This issue only happens with "dotnet" or "fake" (since a driver name is sets) when a scenario is opened and then press ESC to exit the scenario and press ESC again to exit the UICatalog. Without opening any scenario the assertion doesn't fail.

BDisp: Never mind, I add the same workaround as the `All_Scenarios_Benchmark` unit test in the https://github.com/gui-cs/Terminal.Gui/pull/4421/commits/69c4fb3702e48d95c5381af98fb49c60ff1c5e07. For me it isn't better fix that the one did before, but you're the boss.

tig:

1) ensure they are both configproperties. By defn config props are process, not app wide 
2) completely remove force16 from Application and IApplication. Force16 is a driver characteristic, not app. 
3) define it in DriverImpl.cs and change the class name to be "Driver.Force16Colors"
4) modify DriverImpl to read the static in the DriverImpl constructor and hold the value in an instance field. 
5) Add to IDriver as GetForce16()
6) ensure ForceDriver is get only on IApplication. 

Make sense?

BDisp:
> The right solution for the two 'force' properties is:
> 
> 1. ensure they are both configproperties. By defn config props are process, not app wide
> 2. completely remove force16 from Application and IApplication. Force16 is a driver characteristic, not app.
> 3. define it in DriverImpl.cs and change the class name to be "Driver.Force16Colors"
> 4. modify DriverImpl to read the static in the DriverImpl constructor and hold the value in an instance field.
> 5. Add to IDriver as GetForce16()
> 6. ensure ForceDriver is get only on IApplication.
> 
> Make sense?

It make sense but not for this PR. I already make Force16Colors as a driver characteristic to make the necessary fixes related with the windows broken in cmd/conhost. Please, can the proposed changes be made later on another PR? For sure you've many ideas which can be leverage to do these improvements. I want to focus now in this PR with the Sixel flicker issue, if you don't mind.

tig:
> It make sense but not for this PR. I already make Force16Colors as a driver characteristic to make the necessary fixes related with the windows broken in cmd/conhost. Please, can the proposed changes be made later on another PR? For sure you've many ideas which can be leverage to do these improvements. I want to focus now in this PR with the Sixel flicker issue, if you don't mind.

no problem at all!

BDisp:
@tig I'll try to resolve the Sixel output flickering issue in another PR. I think this PR already contains the necessary changes to resolve the problem from the original post. I implemented IsVirtualTerminal, moved Sixel to IDriver, implemented IDriver and IsVirtualTerminal in OutputBase as internal, and added many unit tests. The changes made in #4447 will invalidate much of the code I implemented here. Therefore, I would suggest marking this PR as ready for review and then trying to resolve the Sixel output flickering issue in another PR. After merging, you can and should make the changes you deem necessary, but respecting the unit tests, with some adjustments, of course. Do you agree?

tig:
> @tig I'll try to resolve the Sixel output flickering issue in another PR. I think this PR already contains the necessary changes to resolve the problem from the original post. I implemented IsVirtualTerminal, moved Sixel to IDriver, implemented IDriver and IsVirtualTerminal in OutputBase as internal, and added many unit tests. The changes made in #4447 will invalidate much of the code I implemented here. Therefore, I would suggest marking this PR as ready for review and then trying to resolve the Sixel output flickering issue in another PR. After merging, you can and should make the changes you deem necessary, but respecting the unit tests, with some adjustments, of course. Do you agree?

Yep. Have you tried merging my PR just to see how hard I made it? No biggie if not, I'll get to it maybe tomorrow

tig:
Closing  - This work was merged into 

- #4448

Issue #2443, opened 2023-03-24
Add `IRunnable.Prompt` - Like `MessageBox.QueryFull` but generic


tig:
With `IRunnable` support we can make the current `MessageBox` concept more generic and make it so any View can be hosted in a popup where, when the user accepts a result is returned. 

The idea would be to add a `TResult? Prompt<TView> (TResult input)` API to `IRunnable<TResult>` and an implementation in `Runnable<TResult>` (of course based on a non-generic `object?` version in `IRunnable`/`Runnable`) that 

- Wraps `TView` using an updated version of `RunnableWrapper<TView, TResult>` that curently exists in prototype form. The updated verison would not ony add the created `TView` instance, but an Ok and Cancel button.
- Calls `Run`
- Returns the `Result`

Once this is in, we can re-write `MessageBox` to be a simple wrapper around this.

The exising `Prompt` methods should be replaced with this.

- `ColorPicker`
- `FileDialog`

## Original Description:

              MessageBox doesn't support text input does it?

![image](https://user-images.githubusercontent.com/31306100/227381309-f903c654-8142-4cbb-af83-be2ec7fbafe5.png)

_Originally posted by @tznind in https://github.com/gui-cs/Terminal.Gui/pull/2259#discussion_r1146950418_

tznind:
Here is the code I had in FileDialog2 when I was developing.  The nice UX things we want are:

- Ensure the TextField has input focus (for this I had to add the buttons to `Dialog` after the text box (i.e. not use the `Button` constructor).
- Allow default text for the field but also 'Select all' the text so that user can just start typing something new if they want.

```csharp
 private bool Prompt(string title,string defaultText, out string result)
{

            bool confirm = false;
            var btnOk = new Button("Ok"){
                IsDefault = true,
            };
            btnOk.Clicked+=(s,e)=>{
                confirm = true;
                Application.RequestStop();
            };
            var btnCancel = new Button("Cancel");
            btnCancel.Clicked+=(s,e)=>{
                confirm = false;
                Application.RequestStop();
            };

            var lbl = new Label(Strings.fdRenamePrompt);
            var tf = new TextField(defaultText){
                X = Pos.Right(lbl),
                Width = Dim.Fill(),
            };
            tf.SelectAll();

            var dlg = new Dialog(title){
                Width = Dim.Percent(50),
                Height = 4
            };
            dlg.Add(lbl);
            dlg.Add(tf);

[...]

PR #4422, opened 2025-11-28
Fixes #4419, #4148, #4408 - Toplevel is GONE - Replaced by Runnable

### LINKED ISSUE #2235: ISupportInitialize.BeginInit/EndInit should be virtual
Since View is intended to be heavily inherited to create new types of views, it would be quite convenient to be able to override `BeginInit` and `EndInit`, so that additional automatic initialization can be performed.

This is the typical pattern for interfaces that can be of use to derived classes (i.e. IDisposable and the virtual Dispose impl.).

If desired, the interface implementation can be made explicit and a protected virtual member can be exposed instead, to preserve the base class behavior even if overriden. The former is simpler, but the latter might be more solid if the base View performs critical work on its BeginInit/EndInit pair.

### LINKED ISSUE #4086: Revamp built-in Scheme's (e.g. what *is* `Toplevel` anyway?).
I am not going to tacklet this as part of 

- #4062 

Instead, we'll do it as a follow-on PR.

Recall this conversation: https://github.com/gui-cs/Terminal.Gui/issues/457#issuecomment-2832945208

Here's the latest Scheme Deep Dive that matches the latest impl in

- #4062 

# Scheme Deep Dive

See [Drawing](drawing.md) for an overview of the drawing system and [Configuration](config.md) for an overview of the configuration system.

## Scheme Overview

A Scheme is named a mapping from `VisualRole`s (e.g. `VisualRole.Focus`) to `Attribute`s, defining how a `View` should look based on its purpose (e.g. Menu or Dialog). @Terminal.Gui.SchemeManager.Schemes is a dictionary of `Scheme`s, indexed by name.

A Scheme defines how Views look based on their semantic purpose. The following schemes are supported:

| Scheme Name | Description |
|:-----|:--------|
| **Base** | The base scheme used for most Views. |
| **TopLevel** | The application Toplevel scheme; used for the Toplevel View. |
| **Dialog** | The dialog scheme; used for Dialog, MessageBox, and other views dialog-like views. |
| **Menu** | The menu scheme; used for Terminal.Gui.Menu, MenuBar, and StatusBar. |
| **Error** | The scheme for showing errors, such as in `ErrorQuery`. |

@Terminal.Gui.SchemeManager manages the set of available schemes and provides a set of convenience methods for getting the current scheme and for overriding the default values for these schemes.

```csharp
var scheme = SchemeManager.GetCurrentSchemes () ["TopLevel"];
```

[ConfigurationManager](config.md) can be used to override the default values for these schemes and add additional schemes. 

### Flexible Scheme Management in `Terminal.Gui.View`

A `View`'s appearance is primarily determined by its `Scheme`, which maps semantic `VisualRole`s (like `Normal`, `Focus`, `Disabled`) to specific `Attribute`s (foreground color, background color, and text style). `Terminal.Gui` provides a flexible system for managing these schemes:

1.  **Scheme Inheritance (Default Behavior)**:
    *   By default, if a `View` does not have a `Scheme` explicitly set, it inherits the `Scheme` from its `SuperView` (its parent in the view hierarchy).
    *   This cascading behavior allows for consistent styling across related views. If no `SuperView` has a scheme, (e.g., if the view is a top-level view), it ultimately falls back to the "Base" scheme defined in `SchemeManager.GetCurrentSchemes()`.
    *   The `GetScheme()` method implements this logic:
        *   It first checks if a scheme has been explicitly set via the `_scheme` field (see point 2).
        *   If not, and if `SchemeName` is set, it tries to resolve the scheme by name from `SchemeManager`.
        *   If still no scheme, it recursively calls `SuperView.GetScheme()`.
        *   As a final fallback, it uses `SchemeManager.GetCurrentSchemes()["Base"]!`.

2.  **Explicit Scheme Assignment**:
    *   You can directly assign a `Scheme` object to a `View` using the `View.Scheme` property (which calls `SetScheme(value)`). This overrides any inherited scheme. The `HasScheme` property will then return `true`.
    *   Alternatively, you can set the `View.SchemeName` property to the name of a scheme registered in `SchemeManager`. If `Scheme` itself hasn't been directly set, `GetScheme()` will use `SchemeName` to look up the scheme. This is useful for declarative configurations (e.g., from a JSON file).
    *   The `SetScheme(Scheme? scheme)` method updates the internal `_scheme` field. If the new scheme is different from the current one, it marks the view for redraw (`SetNeedsDraw()`) to reflect the visual change. It also handles a special case for `Border` to ensure its scheme is updated if it `HasScheme`.

3.  **Event-Driven Customization**:
    The scheme resolution and application process includes events that allow for fine-grained control and customization:

    *   **`GettingScheme` Event (`View.Scheme.cs`)**:
        *   This event is raised within `GetScheme()` *before* the default logic (inheritance, `SchemeName` lookup, or explicit `_scheme` usage) fully determines the scheme.
        *   Subscribers (which could be the `SuperView`, a `SubView`, or any other interested component) can handle this event.
        *   In the event handler, you can:
            *   **Modify the scheme**: Set `args.NewScheme` to a different `Scheme` object.
            *   **Cancel default resolution**: Set `args.Cancel = true`. If canceled, the `Scheme` provided in `args.NewScheme` (which might have been modified by the handler) is returned directly by `GetScheme()`.
        *   The `OnGettingScheme(out Scheme? scheme)` virtual method is called first, allowing derived classes to provide a scheme directly.

    *   **`SettingScheme` Event (`View.Scheme.cs`)**:
        *   This event is raised within `SetScheme(Scheme? scheme)` *before* the `_scheme` field is actually updated.
        *   Subscribers can cancel the scheme change by setting `args.Cancel = true` in the event handler.
        *   The `OnSettingScheme(in Scheme? scheme)` virtual method is called first, allowing derived classes to prevent the scheme from being set.

4.  **Retrieving and Applying Attributes for Visual Roles (`View.Attribute.cs`)**:
    Once a `View` has determined its active `Scheme` (via `GetScheme()`), it uses this scheme to get specific `Attribute`s for rendering different parts of itself based on their `VisualRole`.

    *   **`GetAttributeForRole(VisualRole role)`**:
        *   This method first retrieves the base `Attribute` for the given `role` from the `View`'s current `Scheme` (`GetScheme()!.GetAttributeForRole(role)`).
        *   It then raises the `GettingAttributeForRole` event (and calls the `OnGettingAttributeForRole` virtual method).
        *   Subscribers to `GettingAttributeForRole` can:
            *   **Modify the attribute**: Change the `args.NewValue` (which is passed by `ref` as `schemeAttribute` to the event).
            *   **Cancel default behavior**: Set `args.Cancel = true`. The (potentially modified) `args.NewValue` is then returned.
        *   Crucially, if the `View` is `Enabled == false` and the requested `role` is *not* `VisualRole.Disabled`, this method will recursively call itself to get the `Attribute` for `VisualRole.Disabled`. This ensures disabled views use their designated disabled appearance.

    *   **`SetAttributeForRole(VisualRole role)`**:
        *   This method is used to tell the `ConsoleDriver` which `Attribute` to use for subsequent drawing operations (like `AddRune` or `AddStr`).
        *   It first determines the appropriate `Attribute` for the `role` from the current `Scheme` (similar to `GetAttributeForRole`).
        *   It then raises the `SettingAttributeForRole` event (and calls `OnSettingAttributeForRole`).
        *   Subscribers can modify the `schemeAttribute` (via `args.NewValue`) or cancel the operation (`args.Cancel = true`).
        *   If not canceled, it calls `Driver.SetAttribute(schemeAttribute)`.

    *   **`SetAttribute(Attribute attribute)`**:
        *   This is a more direct way to set the driver's current attribute, bypassing the scheme and role system. It's generally preferred to use `SetAttributeForRole` to maintain consistency with the `Scheme`.

### Impact of SuperViews and SubViews via Events

*   **SuperView Influence**: A `SuperView` can subscribe to its `SubView`'s `GettingScheme` or `GettingAttributeForRole` events. This would allow a `SuperView` to dynamically alter how its children determine their schemes or specific attributes, perhaps based on the `SuperView`'s state or other application logic. For example, a container view might want all its children to adopt a slightly modified version of its own scheme under certain conditions.

*   **SubView Influence (Less Common for Scheme of Parent)**: While a `SubView` *could* subscribe to its `SuperView`'s scheme events, this is less typical for influencing the `SuperView`'s *own* scheme. It's more common for a `SubView` to react to changes in its `SuperView`'s scheme if needed, or to manage its own scheme independently.

*   **General Event Usage**: These events are powerful for scenarios where:
    *   A specific `View` instance needs a unique, dynamically calculated appearance that isn't easily captured by a static `Scheme` object.
    *   External logic needs to intercept and modify appearance decisions.
    *   Derived `View` classes want to implement custom scheme or attribute resolution logic by overriding the `On...` methods.

In summary, `Terminal.Gui` offers a layered approach to scheme management: straightforward inheritance and explicit setting for common cases, and a robust event system for advanced customization and dynamic control over how views derive and apply their visual attributes. This allows developers to achieve a wide range of visual styles and behaviors.

# Proposal 

Change the schemes to these:

| Scheme Name | Description |
|:------------|:------------|
| **Base** | The foundational scheme used for most `View` instances. All other schemes inherit from this. |
| **Modal** | Used for modal Views such as `Dialog`, `MessageBox`, and `Wizard`. |
| **Menus** | Used by `Menu`, `MenuBar`, and `StatusBar` to style command surfaces. |
| **Error** | Used to indicate error states, such as `ErrorQuery` or failed validations. |
| **Field** | Used for text-entry Views such as `TextField`, `ComboBox`, and related controls. |
| **Text** | Used for multiline editable content (e.g., `TextView`). |
| **List** | Used for list-like selectors (e.g., `ListView`, `RadioGroup`). |
| **Table** | Used for tabular data views like `TableView`. |
| **Label** | Used for non-interactive text display (`Label`, inline titles, etc.). |
| **Popover** | Used for lightweight, transient UI like hover text, help popups, or dropdowns. |

May be too many. But I wanted to get something down in writing.

### LINKED ISSUE #4148: Replace `Toplevel` with `IRunnable`
## Summary

This Issue recommends decoupling Terminal.Gui's "Runnable" concept from `Toplevel` and `ViewArrangement.Overlapped`, elevating it to a first-class interface-based abstraction. 

**Key Insight**: Analysis of the codebase reveals that **all runnable sessions are effectively modal** - they block in `Application.Run()` until stopped and capture input. The distinction between "modal" and "non-modal" in the current design is artificial:

- The `Modal` property only affects input propagation and Z-order, not the fundamental run loop behavior
- All `Toplevel`s block in `Run()` - there's no "background" runnable concept
- Non-modal `Toplevel`s (like `WizardAsView`) are just embedded views with `Modal = false`, not true sessions
- Overlapped windows are managed by `ViewArrangement.Overlapped`, not runnability

By introducing `IRunnable<TResult>`, we create a clean separation where:

- **Runnable** = Can be run as a **UI**-blocking session with `Application.Run()` and returns a result
- **Overlapped** = `ViewArrangement.Overlapped` for window management (orthogonal to runnability)
- **Embedded** = Just views, not runnable at all

## Full architecture doc is now in v2_develop at 

https://github.com/gui-cs/Terminal.Gui/blob/v2_develop/docfx/docs/runnable-architecture-proposal.md

**IMPORTANT** Read the entire thing before working on Issues related to this.

## Terminology

This proposal introduces new terminology to clarify the architecture:

| Term | Definition |
|------|------------|
| **`IRunnable`** | Base interface for Views capable of running as an independent session with `Application.Run()` without returning a result. Replaces `Toplevel` as the contract for runnable views. When an `IRunnable` is passed to `IApplication.Run`, `Run` blocks until the `IRunnable` `Stops`. |
| **`IRunnable<TResult>`** | Generic interface derived from `IRunnable` that can return a typed result. |
| **`Runnable<TResult>`** | Optional base class that implements `IRunnable<TResult>` and derives from `View`, providing default lifecycle behavior. Views can derive from this or implement `IRunnable<TResult>` directly. |
| **`TResult`** | Type parameter specifying the type of result data returned when the runnable completes (e.g., `int` for button index, `string` for file path, enum, or other complex type). `Result` is `null` if the runnable stopped without the user explicitly accepting it (ESC pressed, window closed, etc.). |
| **`Result`** | Property on `IRunnable<TResult>` that holds the typed result data. Should be set in `IsRunningChanging` handler (when `newValue = false`) **before** the runnable is popped from `RunnableSessionStack`. This allows subscribers to inspect results and optionally cancel the stop. Available after `IApplication.Run` returns. `null` indicates cancellation/non-acceptance. |
| **RunnableSession** | A running instance of an `IRunnable`. Managed by `IApplication` via `Begin()`, `Run()`, `RequestStop()`, and `End()` methods. Represented by a `RunnableSessionToken` on the `RunnableSessionStack`. |
| **`RunnableSessionToken`** | Object returned by `Begin()` that represents a running session. Wraps an `IRunnable` instance (via a `Runnable` property) and is stored in `RunnableSessionStack`. Disposed when session ends. |
| **`RunnableSessionStack`** | A stack of `RunnableSessionToken` instances, each wrapping an `IRunnable`. Tracks all running runnables in the application. Literally a `ConcurrentStack<RunnableSessionToken>`. Replaces `SessionStack` (formerly `Toplevels`). 
| **`IsRunning`** | Boolean property on `IRunnable` indicating whether the runnable is currently on the `RunnableSessionStack` (i.e., `RunnableSessionStack.Any(token => token.Runnable == this)`). Read-only, derived from stack state. Runnables are added during `IApplication.Begin` and removed in `IApplication.End`. Replaces `Toplevel.Running`. |
| **`IsRunningChanging`** | Cancellable event raised **before** an `IRunnable` is added to or removed from `RunnableSessionStack`. When transitioning to `IsRunning = true`, can be canceled to prevent starting. When transitioning to `IsRunning = false`, allows code to prevent closure (e.g., prompt to save changes) AND is the ideal place to extract `Result` before the runnable is removed from the stack. Event args (`CancelEventArgs<bool>`) provide the new state in `NewValue`. Replaces `Toplevel.Closing` and partially `Toplevel.Activate`. |
| **`IsRunningChanged`** | Non-cancellable event raised **after** a runnable has been added to or removed from `RunnableSessionStack`. Fired after `IsRunning` has changed to the new value (true = started, false = stopped). For post-state-change logic (e.g., setting focus after start, cleanup after stop). Replaces `Toplevel.Activated` and `Toplevel.Closed`. |
| **`IsInitialized`**  (`View` property) | Boolean property (on `View`) indicating whether a view has completed two-phase initialization (`View.BeginInit/View.EndInit`). From .NET's `ISupportInitialize` pattern. If the `IRunnable.IsInitialized == false`, `BeginInit` is called from `IApplication.Begin` after `IsRunning` has changed to `true`.  `EndInit` is called immediately after `BeginInit`. |
| **`Initialized`**  (`View` event) | Non-cancellable event raised as `View.EndInit()` completes. |
| **`TopRunnable`** (`IApplication` property) | Used to be `Top`, but was recently renamed to `Current` because it was confusing relative to `Toplevel`. It's precise definition in this proposal is "The `IRunnable` that is on the top of the `RunnableSessionStack` stack. And by definition, and per-implementation, this `IRunnable` is capturing all mouse and keyboard input and is thus "Modal". Note, any other `IRunnable` instances on `RunnableSessionStack` continue to be laid out, drawn, and receive iteration events; they just don't get any user input. Another interesting note: No code in the solution other than ./App, ./ViewBase, and tests reference `IApplication.Current` (an indication the previous de-coupling was successful). It also means the name of this property is not that important because it's just an implementation detail, primarily used to enable tests to not have to actually call `Run`. View has `public bool IsCurrentTop => App?.Current == this;`; thus we rename `IApplication.Current` to `IApplication.TopRunnable` and it's synonymous with `IRunnable.IsModal`. |
| **`IsModal`** | Boolean property on `IRunnable` indicating whether the `IRunnable` is at the top of the `RunnableSessionStack` (i.e., `this == app.TopRunnable` or `app.RunnableSessionStack.Peek().Runnable == this`). The `IRunnable` at the top of the stack gets all mouse/keyboard input and thus is running "modally". Read-only, derived from stack state. `IsModal` represents the concept from the end-user's perspective. |
| **`IsModalChanging`** | Cancellable event raised **before** an `IRunnable` transitions to/from the top of the `RunnableSessionStack`. When becoming modal (`newValue = true`), can be canceled to prevent activation. Event args (`CancelEventArgs<bool>`) provide the new state. Replaces `Toplevel.Activate` and `Toplevel.Deactivate`. |
| **`IsModalChanged`** | Non-cancellable event raised **after** an `IRunnable` has transitioned to/from the top of the `RunnableSessionStack`. Fired after `IsModal` has changed to the new value (true = became modal, false = no longer modal). For post-activation logic (e.g., setting focus, updating UI state). Replaces `Toplevel.Activated` and `Toplevel.Deactivated`. |
| **`End`** (`IApplication` method) | Ends a running `IRunnable` instance by removing its `RunnableSessionToken` from the `RunnableSessionStack`. `IsRunningChanging` with `newValue = false` is raised **before** the token is popped from the stack (allowing result extraction and cancellation). `IsRunningChanged` is raised **after** the `Pop` operation. Then, `RunnableSessionStack.Peek()` is called to see if another `IRunnable` instance can transition to `IApplication.TopRunnable`/`IRunnable.IsModal = true`. |
| **`ViewArrangement.Overlapped`** | Layout mode for windows that can overlap with Z-order management. Orthogonal to runnability - overlapped windows can be embedded views (not runnable) or runnable sessions. |

**Key Architectural Changes:**
- **Simplified**: One interface `IRunnable<TResult>` replaces both `Toplevel` and the artificial `Modal` property distinction
- **All sessions block**: No concept of "non-modal runnable" - if it's runnable, `Run()` blocks until `RequestStop()`
- **Type-safe results**: Generic `TResult` parameter provides compile-time type safety
- **Decoupled from layout**: Being runnable is independent of `ViewArrangement.Overlapped`
- **Consistent patterns**: All lifecycle events follow Terminal.Gui's Cancellable Work Pattern
- **Result extraction in `Stopping`**: `OnStopping()` is the correct place to extract `Result` before disposal

### Phase 0: Rename `Current` to `TopRunnable` (Optional) **DONE**

- Issue #4148

### Phase 1: Add IRunnable Support **DONE**

- Issue #4400

### Phase 2: Migrate Existing Views

- Issue #4408

1. Make `Toplevel` implement `IRunnable` (adapter pattern for compatibility)
2. Update `Dialog` to inherit from `Runnable<int>` instead of `Window`
3. Update `MessageBox` to use `Dialog.Result`
4. Update `Wizard` to inherit from `Runnable<WizardResult>`
5. Update all examples to use new `IRunnable` pattern

### Phase 3: Deprecate and Remove Toplevel

- Issue (TBD)

1. Mark `Toplevel` as `[Obsolete]`
2. Update all internal code to use `IRunnable`/`Runnable<TResult>`
3. Remove `Toplevel` class entirely (breaking change for v3)

### Phase 4: Upgrade View Initialization (Optional Enhancement)

- Issue (TBD)

1. Refactor `View.BeginInit()`/`View.EndInit()`/`Initialized` to be CWP compliant
2. This is independent of the runnable architecture but would improve consistency

### LINKED ISSUE #4149: Any `View` should be able to be used with `Application.Run`'d without it having to be a subclass of `Toplevel`

### LINKED ISSUE #4365: Migrate `UnitTests` that require keyboard/mouse input to `UnitTests.Parallelizable`
In 

- #4125 

I proved that any unit test that needs to render, even if it uses Views, can now be authored in `UnitTests.Parallelizable`.  See `UnitTests.Parallizable/Drawing`.

Now we need to move some tests that require keyboard / mouse input into Application.

It should now be possible to do this:

```cs
var app = new ApplicationImpl();
app.Init(nul, "fake");

// Do anything you used to be able to in a Unit Test that used Application

// e.g.
app.RaiseNewKeyDownEvent (...);
app.RaiseNewMouseEvent (...);

app.LayoutAndDraw();

DriverAssert.AssertDriverContentsAre (expected, _output);

app.Shutdown(); // technically not needed because there's no static state!
```

First step is to write a sample unit tests in the `UnitTests.Parallelizable` project. Try porting this one:

```cs

    [Fact]
    [AutoInitShutdown]
    public void CaptionedTextField_DoesNotOverspillViewport_Unicode ()
    {
        string caption = "Mise" + char.ConvertFromUtf32 (int.Parse ("0301", NumberStyles.HexNumber)) + "rables";

        Assert.Equal (11, caption.Length);
        Assert.Equal (10, caption.EnumerateRunes ().Sum (c => c.GetColumns ()));

        TextField tf = GetTextFieldsInView ();

        tf.Title = caption;
        Application.RaiseKeyDownEvent ('\t');
        Assert.False (tf.HasFocus);

        tf.Draw ();
        DriverAssert.AssertDriverContentsAre ("Misérables", output);
        Application.Top.Dispose ();
    }
```

### LINKED ISSUE #4408: SubIssue of #4148 - Phase 2 - Migrate `Toplevel` to implement `IRunnable` etc...
This Issue will be tackled after #4148 has been closed.

## Carefully following the proposal here:

- `./docfx/docs/runnable-architecture-proposal.md`

Phase 1 is completed. You are implementing phase 2 as described near the end.

## Work to be done to close this Issue:

1. Make Toplevel implement IRunnable (adapter pattern for compatibility)
2. Update Dialog to inherit from Runnable<int> instead of Window
3. Update MessageBox to use Dialog.Result
4. Update Wizard to inherit from Runnable<WizardResult>
5. Update all examples to use new IRunnable pattern
6. Update the new FluentExample to define POST_4148 and ensure it works
7.  Create a NEW set of parallelizable unit tests that extensively tests all of the above. **DO NOT ASSUME THE EXISTING TESTS ARE ANY GOOD**. It is ok to disable old tests you deem to be poorly written, confusing, or no longer applicable. Use `Skip` in the `[Fact]` pr `[Theory]`.
## Exit Criteria

- [ ] All XML API docs 
- [ ] All relevant deep dives in ./docfx/docs have been updated
- [ ] All tests pass
- [ ] There are no new warnings
- [ ] CI/CD passes

**IMPORTANT**: At each step in implementing this you must:

- Revisit #4148 regularly to ensure your thinking stays on track.
- Re-read ./docfx/application.md and ./docfx/View.md to ensure you still understand the nuances.
- Ensure you are following contributing.md
- Build and test to ensure all tests pass and no new warnings are introduced

### LINKED ISSUE #4419: `v2_develop` is broken - #4411 was merged prematurely
I merged #4411 prematurely. v2_devleop is broken as a result.

Working on a fix now.

tig:
## Fixes

- Fixes #4408 
- Fixes #4419 
- Fixes #4148 
- Fixes #4149
- Fixes #4086 
- Fixes #2235 
- Fixes #4365

### DIFF
--- a/Tests/UnitTestsParallelizable/ViewBase/Mouse/MouseDragTests.cs
+++ b/Tests/UnitTestsParallelizable/ViewBase/Mouse/MouseDragTests.cs
@@ -66,7 +66,7 @@ public class MouseDragTests
         Assert.Equal (10, movableView.Frame.Width);
         Assert.Equal (10, movableView.Frame.Height);
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -125,7 +125,7 @@ public class MouseDragTests
         Assert.Equal (13, movableView.Frame.X); // 10 + 3
         Assert.Equal (13, movableView.Frame.Y); // 10 + 3
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -191,7 +191,7 @@ public class MouseDragTests
         Assert.Equal (15, movableView.Frame.X);
         Assert.Equal (15, movableView.Frame.Y);
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -254,7 +254,7 @@ public class MouseDragTests
         Assert.Equal (15, resizableView.Frame.Width); // Width increased by 5
         Assert.Equal (10, resizableView.Frame.Height); // Height unchanged
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -313,7 +313,7 @@ public class MouseDragTests
         Assert.Equal (10, resizableView.Frame.Width); // Width unchanged
         Assert.Equal (15, resizableView.Frame.Height); // Height increased by 5
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -372,7 +372,7 @@ public class MouseDragTests
         Assert.Equal (13, resizableView.Frame.Width); // Width increased by 3
         Assert.Equal (10, resizableView.Frame.Height); // Height unchanged
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -431,7 +431,7 @@ public class MouseDragTests
         Assert.Equal (10, resizableView.Frame.Width); // Width unchanged
         Assert.Equal (12, resizableView.Frame.Height); // Height increased by 2
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -494,7 +494,7 @@ public class MouseDragTests
         Assert.Equal (15, resizableView.Frame.Width); // Width increased by 5
         Assert.Equal (15, resizableView.Frame.Height); // Height increased by 5
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -553,7 +553,7 @@ public class MouseDragTests
         Assert.Equal (13, resizableView.Frame.Width); // Width increased by 3
         Assert.Equal (12, resizableView.Frame.Height); // Height increased by 2
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -608,7 +608,7 @@ public class MouseDragTests
         int expectedMinWidth = resizableView.Border!.Thickness.Horizontal + resizableView.Margin!.Thickness.Right;
         Assert.True (resizableView.Frame.Width >= expectedMinWidth);
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }
@@ -658,7 +658,7 @@ public class MouseDragTests
         int expectedMinHeight = resizableView.Border!.Thickness.Vertical + resizableView.Margin!.Thickness.Bottom;
         Assert.True (resizableView.Frame.Height >= expectedMinHeight);
 
-        app.End (app.SessionStack.First ());
+        app.End (app.SessionStack!.First ());
         runnable.Dispose ();
         superView.Dispose ();
     }


tig:
I've got a bit more to do here:

- [x] Except for `TopRunningView` there should be NO coupling between IApplicaiton/ApplicationImpl and View. `TopRunnableView` is just a convivence. 
- [x] `ColorPicker.Prompt` needs updating.

Follow ons:

- [ ] With this new model, it should be possible to just make `Popover` an `IRunnable` that's transparent. This will be a follow up PR.
- [ ] Relating to Prompt, in a future PR (See #2443) we can esaily use `RunnableWrapper` to make a generic Prompt API.

PR #4435, opened 2025-12-03
Fixes #4434 - `InvokeLeakTest` 

### LINKED ISSUE #4434: `InvokeLeakTest` needs to be modernized and made more resiliant
I reproduced this failing in the debugger. I've never had it fail outside the debugger though. This is likely a perf related issue. And the test needs to be modnernized. 

I have a PR readay.

> In VS 2026 if I click on Run instead of Debug the unit test also fails.
> 
> Image
> 
> If I run from the terminal `dotnet test --filter "InvokeLeakTest"` the unit test also fails.
> 
> Image
> 
> If I run from the terminal `dotnet test .\Tests\StressTests\StressTests.csproj` the unit test also fails.
> 
> Image 

 _Originally posted by @BDisp in [#4429](https://github.com/gui-cs/Terminal.Gui/issues/4429#issuecomment-3607696902)_

tig:
## Fixes

- Fixes #4434

### DIFF
--- a/Tests/StressTests/ApplicationStressTests.cs
+++ b/Tests/StressTests/ApplicationStressTests.cs
@@ -1,41 +1,44 @@
+using System.Diagnostics;
 using Xunit.Abstractions;
+
+// ReSharper disable AccessToDisposedClosure
 
 namespace StressTests;
 
-public class ApplicationStressTests
+public class ApplicationStressTests (ITestOutputHelper output)
 {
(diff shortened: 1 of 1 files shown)


tig:
> This is great and the unit test pass. The only failure in this PR I think it's a race condition.

This has to do with a CM.SourcesManager test. I have a theory...

PR #4439, opened 2025-12-04
Fixes#4431 - Fixes `MainLoop` drawing issues + `View.Draw` issues

### LINKED ISSUE #4431: MainLoop draw logic is confused
There are a bunch of issues that lead to both determinism and perfomrance issues.

## `IDriver.Refresh` is not implemented

```cs
    /// <inheritdoc/>
    public void Refresh ()
    {
        // No need we will always draw when dirty
    }
```

This is illogical and silly. At the very least an impl is needed for tests. But in reality `LayoutAndDraw` call it.

```cs
        if (Driver is { })
        {
            Driver.Clip = new (Screen);

            View.Draw (views: tops!, neededLayout || forceRedraw);
            Driver.Clip = new (Screen);
            Driver?.Refresh ();
        }
```

`Command.Refresh` doesn't actually work as a result:

```cs

        AddCommand (
                    Command.Refresh,
                    () =>
                    {
                        App?.LayoutAndDraw (true);

                        return true;
                    }
                   );
```

## `IterationImpl` has bad logic and duplicates `LayoutAndDraw`

```cs
            bool needsDrawOrLayout = AnySubViewsNeedDrawn (App?.Popover?.GetActivePopover () as View)
                                     || AnySubViewsNeedDrawn (App?.TopRunnableView)
                                     || (App?.Mouse.MouseGrabView != null && AnySubViewsNeedDrawn (App?.Mouse.MouseGrabView));

            bool sizeChanged = SizeMonitor.Poll ();

            if (needsDrawOrLayout || sizeChanged)
            {
                Logging.Redraws.Add (1);

                App?.LayoutAndDraw (forceRedraw: true);

                Output.Write (OutputBuffer);

                Output.SetCursorVisibility (CursorVisibility.Default);
            }
```

- The logic for `needDrawOrLayout` doesn't account for other Runnable sessions. Thus when a MessageBox is shown or dismissed, for example, redraws don't happen until another iteration.

- Using `forceRedraw: true` completely negates all the wonderful logic in `View` on minimizing what gets redrawn and when. 

- This tight coupling to View within the main loop needs to be removed. Recursing the entire View hierarchy is not needed anyway (see above). Application (and by extension the main loop) must have no direct coupling with View... only IRunnable. I'm ok with this rule being broken in LayoutAndDraw for now, but anywhere else we should squash. 

- Direclty calling `Output.Write` duplicates the fact that `LayoutAndDraw` calls `driver.Refresh`... Or would if that API were correct (see above).

- There should be no reason to poll the screen size in each iteration. We have all the wiring for screen size changes to occur via events. 

- Not related, but related: The call to `SetCursorViisibilty` is a bug and it should be removed.

# There are no real good tests for this stuff today. 

Repeatedly, these issues are only noticed when running examples. 

# What's the right model?

There are these concepts:

- Iteration - An Application concept that happens repeatedly, throttled to a max # every second. Each of the things below happens each iteration. 
- Layout - A View concept that should only happen once an iteration and must happen before Draw. Depends on the screen size having been updated. 
- Draw - A View concept that also happens only once an iteration after layout. Causes the Driver back buffer (Contents) to be updated, only changing what portions need to be updated 
- Write - A Driver concept that takes Contents and writes it to the real console. 
- Cursor - A driver concept to ensure the cursor is where it should be and in the right visibility mode. 
- TimeOuts - An application concept where scheduled work can be performed. 

We long ago agreed on the above and @tznind did a great job of creating a decoupled architecture for all this in the new v2 system. I recently polished most of this being even more hard core about decoupling and the correct abstractions. This issue covers the biggest remaining piece.

tig:
## Fixes

- Fixes #4431

## Pull request overview

This PR addresses MainLoop and cursor positioning issues (#4431) by fundamentally refactoring how Terminal.Gui handles cursor positioning and drawing state management. The core fix involves separating the concept of the "Draw Cursor" (internal position tracking for drawing operations) from the "Terminal Cursor" (the visible cursor shown to users).

**Key Changes:**
- **Cursor Management Separation**: Removed the conflation between drawing position (`OutputBuffer.Col`/`Row`) and terminal cursor positioning. The terminal cursor is now solely managed by `ApplicationMainLoop.SetCursor()` based on `View.PositionCursor()`.
- **Drawing State Refactoring**: Extracted `NeedsDraw` logic into a new dedicated file (`View.NeedsDraw.cs`) with improved state management. Changed `ClearNeedsDraw()` to no longer clear the SuperView's `SubViewNeedsDraw` flag, which is now handled explicitly in the static `View.Draw()` method.
- **API Changes**: Modified `OnDrawingContent()` to require a `DrawContext?` parameter (breaking change), removed the parameterless overload, and made `NeedsDraw` a read-only property.

### DIFF
--- a/Examples/PowershellExample.ps1
+++ b/Examples/PowershellExample.ps1
@@ -1,0 +1,35 @@
+using namespace Terminal.Gui.App        
+using namespace Terminal.Gui.ViewBase
+using namespace Terminal.Gui.Views
+
+$dllFolder = "..\Terminal.Gui\bin\Debug\net8.0"
+
+# For this to work all dependent DLLs need to be in the $dllFolder folder
+# Do this first:
+#   dotnet build -c Debug /p:CopyLocalLockFileAssemblies=true
+
+Get-ChildItem $dllFolder -Filter *.dll | ForEach-Object {
+    Add-Type -Path $_.FullName -ErrorAction SilentlyContinue
+}
+
+$app = [Application]::Create()
+
+$app.Init()
+
+$win = [Window]@{
+    Title  = "Terminal.Gui in Powershell"
+    Width  = [Dim]::Fill()
+    Height = [Dim]::Fill()
+}
+
+$lbl = [Label]@{
+    Text = "Hello from PowerShell + Terminal.Gui!`nPress ESC to quit"
+    X    = [Pos]::Center()
+    Y    = [Pos]::Center()
+}
+$win.Add($lbl)
+
+$app.Run($win)
+
+$win.Dispose()
+$app.Dispose()


tig:
As part of this, I've found a bunch of issues in View.Draw and have fixed them, signifcantly bolstering unit tests.

I also updated the Transparency Scenario to show how DrawContent can/should work:

WindowsTerminal_Wmll7xFKEn

BDisp (approved):
It seems the CheckBox is already working in the menu. Now it only redraws when necessary, improving performance. Looks great.

tig:
> It seems the CheckBox is already working in the menu. Now it only redraws when necessary, improving performance. Looks great.

There are still bugs in `Shortcut` and `Menu` etc... w.r.t. complex CommandViews. Theya related to Accept/Select etc... I'm still working on that. Just FYI.

PR #4447, opened 2025-12-05
Fixes #4445 - Cleans up `Application.ForceDriver` and `Driver.Force16Colors`

### LINKED ISSUE #4445: Clean up `ForceDriver` and `Force16Colors`

1. ensure they are both configproperties. By defn config props are process, not app wide
2. completely remove force16 from Application and IApplication. Force16 is a driver characteristic, not app.
3. define it in DriverImpl.cs and change the class name to be "Driver.Force16Colors"
4. modify DriverImpl to read the static in the DriverImpl constructor and hold the value in an instance field.
5. Add to IDriver as GetForce16()
6. ensure ForceDriver is get only on IApplication.

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/4421#issuecomment-3613149621_

tig:
## Fixes

- Fixes #4445

### DIFF
--- a/Terminal.Gui/Drivers/DriverImpl.cs
+++ b/Terminal.Gui/Drivers/DriverImpl.cs
@@ -26,7 +26,7 @@ namespace Terminal.Gui.Drivers;
 ///         Applications interact with drivers through the <see cref="Application"/> class.
 ///     </para>
 /// </remarks>
-internal class DriverImpl : IDriver
+internal class DriverImpl : IDriver, IDisposable
 {
     private readonly IOutput _output;
     private readonly AnsiRequestScheduler _ansiRequestScheduler;
@@ -63,18 +63,16 @@ internal class DriverImpl : IDriver
                                      };
 
         SizeMonitor = sizeMonitor;
-
-        sizeMonitor.SizeChanged += (_, e) =>
-                                   {
-                                       SetScreenSize (e.Size!.Value.Width, e.Size.Value.Height);
-
-                                       //SizeChanged?.Invoke (this, e);
-                                   };
+        SizeMonitor.SizeChanged += OnSizeMonitorOnSizeChanged;
 
         CreateClipboard ();
+
+        Driver.Force16ColorsChanged += OnDriverOnForce16ColorsChanged;
+
     }
 
     private void OnDriverOnForce16ColorsChanged (object? _, ValueChangedEventArgs<bool> e) { Force16Colors = e.NewValue; }
+
 
     /// <inheritdoc/>
     public bool Force16Colors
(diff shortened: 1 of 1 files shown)


Terminal.Gui/Drivers/Driver.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Runtime.InteropServices;
+
+namespace Terminal.Gui.Drivers;
+
+/// <summary>
+///     Holds global driver settings.
+/// </summary>
+public sealed class Driver
+{
+    private static bool _force16Colors = false; // Resources/config.json overrides
```
BDisp, line 10: Instead of creating a new `public sealed class Driver`, why not change the `public static partial class Application` to `public sealed class Application` to hold all the global config settings, which may be more appropriate?
I think we could start considering removing all static properties and methods that don't require global configuration in the Application, thus forcing the use of instances only. It will be hard work, but sooner or later it will become necessary. This will force us to modify and move all unit tests from the non-parallelized project "UnitTests". Furthermore, I note that only Force16Colors was manipulated, not ForceDriver.

Terminal.Gui/Drivers/OutputBase.cs
```diff
@@ -5,7 +5,8 @@ namespace Terminal.Gui.Drivers;
 /// </summary>
 public abstract class OutputBase
 {
-    private CursorVisibility? _cachedCursorVisibility;
+    /// <inheritdoc cref="IOutput.Force16Colors"/>
+    public bool Force16Colors { get; set; }
```
BDisp, line 9: Would it be preferable to use the `internal IDriver? Driver { get; set; }` instead of just `Force16Colors`, as I implemented in #4421. This way, any object accessing `OutputBase` can obtain any property existing in the `IDriver`, through injection.

Terminal.Gui/Drivers/Driver.cs
tig: Right now CM has a limitation for class name. There is no way to override such that Force16Colors could be on DriverImpl and be named "Driver.Force16Colors".

This is what has held me back from moving the nav key properties to ApplicationNavigation. But they should and should be renamed to "Navigation.TabKey"

I'll open a new issue for this.

Terminal.Gui/Drivers/OutputBase.cs
tig: I forgot IOutput had a Driver property!

That said, doesn't it make more logical sense for the property to be on IOutput because that's actually where it's used?

BDisp: IOutput don't have it. Only in my PR I add it on the OutputBase as internal.

BDisp: If you only use the Force16Colors in the IOutput, yes it make sense. But imagine you also need to access to another IDriver property. You will create all properties you need for that? Another concerns is only add public properties and methods, avoiding include internal one that which also CM can access.

tig:
Closing  - This work was merged into 

- #4448

PR #4448, opened 2025-12-05
Fixes #4004 & #4445 - Merge of `Application.ForceDriver and Driver.Force16Colors` and `windows" broken in conhost and cmd`

### LINKED ISSUE #4004: Driver "v2win" broken in `conhost`

### LINKED ISSUE #4445: Clean up `ForceDriver` and `Force16Colors`

tig:
## Fixes

- Fixes #4004 
- Fixes #4445 

## Pull request overview

This PR consolidates driver management by merging `Application.ForceDriver` and `Driver.Force16Colors`, and refactors Windows console handling by renaming `IsVirtualTerminal` to `IsLegacyConsole` with inverted logic. This clarifies the architecture where legacy console mode (Windows conhost without VT support) is the exception rather than the modern terminal being "virtual".

**Key changes:**
- Moved `Force16Colors` from `Application` to `Driver` class, making it a driver-level configuration accessible via `Driver.Force16Colors` 
- Renamed `IsVirtualTerminal` to `IsLegacyConsole` with reversed logic throughout WindowsOutput
- Added Sixel rendering support with `IDriver.GetSixels()` and proper integration into the output pipeline

### DIFF
--- a/Terminal.Gui/App/ApplicationImpl.Driver.cs
+++ b/Terminal.Gui/App/ApplicationImpl.Driver.cs
@@ -145,7 +145,11 @@ internal partial class ApplicationImpl
 
     internal void SubscribeDriverEvents ()
     {
-        ArgumentNullException.ThrowIfNull (Driver);
+        if (Driver is null)
+        {
+            Logging.Error($"Driver is null");
+            return;
+        }
 
         Driver.SizeChanged += Driver_SizeChanged;
         Driver.KeyDown += Driver_KeyDown;
@@ -155,7 +159,11 @@ internal partial class ApplicationImpl
 
     internal void UnsubscribeDriverEvents ()
     {
-        ArgumentNullException.ThrowIfNull (Driver);
+        if (Driver is null)
+        {
+            Logging.Error ($"Driver is null");
+            return;
+        }
 
         Driver.SizeChanged -= Driver_SizeChanged;
         Driver.KeyDown -= Driver_KeyDown;


BDisp:
> `IsVirtualTerminal` is renamed to `IsLegacyConsole` and the logic is reversed.

That make sense, good.

BDisp:
Something is broken.

cmd_lcPHaNrBD5

BDisp:
It's something related with mouse because it only happens when I move the mouse. Perhaps some call to set cursor position using escape sequences, I don't know. Debug with F5 in VS 2026 the issues doesn't happen with windows host console.

tig:
Is this on conhost only?

tig:
Hnmmm.. I can't repo.

VsDebugConsole_Jy5JwrlN7b

BDisp:
> Is this on conhost only?

Not only, The cmd.exe too.

BDisp:
> Hnmmm.. I can't repo.
> 
> VsDebugConsole_Jy5JwrlN7b [ VsDebugConsole_Jy5JwrlN7b ](https://private-user-images.githubusercontent.com/585482/523159213-c178cabc-46d1-477d-af65-d87b3de0f4b8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjQ5NjgwNzYsIm5iZiI6MTc2NDk2Nzc3NiwicGF0aCI6Ii81ODU0ODIvNTIzMTU5MjEzLWMxNzhjYWJjLTQ2ZDEtNDc3ZC1hZjY1LWQ4N2IzZGUwZjRiOC5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUxMjA1JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MTIwNVQyMDQ5MzZaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0zYjgyZDU4ZGExYzJhZTdhZTE5YTk3MDcwMDUyYzY0YjU2NTlmMTgxMGQ5ZDhkOTVkZTdiOGZiZDgwMmRiYzAyJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.D8aeo9NdugGQ0-WB8mqC1c2a_jWJ4FM9OK_rWVyCVZY) [ ](https://private-user-images.githubusercontent.com/585482/523159213-c178cabc-46d1-477d-af65-d87b3de0f4b8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjQ5NjgwNzYsIm5iZiI6MTc2NDk2Nzc3NiwicGF0aCI6Ii81ODU0ODIvNTIzMTU5MjEzLWMxNzhjYWJjLTQ2ZDEtNDc3ZC1hZjY1LWQ4N2IzZGUwZjRiOC5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUxMjA1JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MTIwNVQyMDQ5MzZaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT0zYjgyZDU4ZGExYzJhZTdhZTE5YTk3MDcwMDUyYzY0YjU2NTlmMTgxMGQ5ZDhkOTVkZTdiOGZiZDgwMmRiYzAyJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.D8aeo9NdugGQ0-WB8mqC1c2a_jWJ4FM9OK_rWVyCVZY)

You must test with the cmd.exe alone, not with VS 2006 debugger.

You aren't test with a legacy console but with a console that supports VTSequence.

BDisp:
I think already know why. With my original PR it is also reproducing. Probably was caused by the MainLoop fix because it's when a MessageBox is opened that this happens. Do you remember something you did on that PR which may causing this?

tig:
> I think already know why. With my original PR it is also reproducing. Probably was caused by the MainLoop fix because it's when a MessageBox is opened that this happens. Do you remember something you did on that PR which may causing this?

Yeah I just reproduced it in both this branch and `:BDisp/Terminal.Gui.git v2_4004_windows-driver-fix`

I don't remember anything that might be a cause. Do you have any idea why it behaves differently when startred w/in VS vs not?

BDisp:
> Yeah I just reproduced it in both this branch and `:BDisp/Terminal.Gui.git v2_4004_windows-driver-fix`
> 
Yes.

> I don't remember anything that might be a cause. Do you have any idea why it behaves differently when startred w/in VS vs not?

I didn't realized this before because it wasn't a problem before. So, for debug it's more easy to use VS 2026, but even with Windows Host Console configured in the Windows Terminal, the VSDebugConsole always has the VTS enabled, not allowing running a legacy console indeed. When I really want to test legacy console I have to use cmd.exe or conhost.exe.
It seems that the MessageBox start do draw at column 0, perhaps due a wrong clip screen.

tig:
> > Yeah I just reproduced it in both this branch and `:BDisp/Terminal.Gui.git v2_4004_windows-driver-fix`
> 
> Yes.
> 
> > I don't remember anything that might be a cause. Do you have any idea why it behaves differently when startred w/in VS vs not?
> 
> I didn't realized this before because it wasn't a problem before. So, for debug it's more easy to use VS 2026, but even with Windows Host Console configured in the Windows Terminal, the VSDebugConsole always has the VTS enabled, not allowing running a legacy console indeed. When I really want to test legacy console I have to use cmd.exe or conhost.exe. It seems that the MessageBox start do draw at column 0, perhaps due a wrong clip screen.

Def mouse related because if I move the msgbox with keyboard it doesn't happen.

BDisp:
> Def mouse related because if I move the msgbox with keyboard it doesn't happen.

How to move by keyboard?

BDisp:
When a MessageBox is opened It's supposed to checkbox that belong to the runnable top, being triggered by pressing F7, which will set Force16Colors to true? In my original PR I never allow to set it to false if IsLegacyConsole is true. But the main issue is the checkbox being triggered when it shouldn't.

BDisp:
Only the current modal should respond to keyboard and mouse events. I think this was a regression.

tig:
> > Def mouse related because if I move the msgbox with keyboard it doesn't happen.
> 
> How to move by keyboard?

Ctrl-F5 is the arrange key.

BDisp:
> > > Def mouse related because if I move the msgbox with keyboard it doesn't happen.
> > 
> > 
> > How to move by keyboard?
> 
> Ctrl-F5 is the arrange key.

Thanks. That prove that's something with the mouse. If it was legacy console draw, the issue would also happen with keyboard. I already saw that the coordinates of the MessageBox is correct. Probably it's something that cause the runnable top to being draw excluding the MessageBox and maybe is some bad regions passed to it.

tig:
> Only the current modal should respond to keyboard and mouse events. I think this was a regression.

It has something to do with attributes and this code. What is happening is the border on the MB is changing from non-bold to bold when the mouse hovers. I've yet to reproduce it in a simpler scenario, but I'm pretty sure it has nothing to do with an IsModal, but I could be wrong.

```cs
    /// <summary>
    ///     Appends ANSI sequences for a single cell to the output.
    /// </summary>
    /// <param name="cell">The cell to append ANSI for.</param>
    /// <param name="output">The StringBuilder to append to.</param>
    /// <param name="lastAttr">The last attribute used, updated if the cell's attribute is different.</param>
    /// <param name="redrawTextStyle">The current text style for optimization.</param>
    /// <param name="maxCol">The maximum column, used for wide character handling.</param>
    /// <param name="currentCol">The current column, updated for wide characters.</param>
    protected void AppendCellAnsi (Cell cell, StringBuilder output, ref Attribute? lastAttr, ref TextStyle redrawTextStyle, int maxCol, ref int currentCol)
    {
        Attribute? attribute = cell.Attribute;

        // Add ANSI escape sequence for attribute change
        if (attribute.HasValue && attribute.Value != lastAttr)
        {
            lastAttr = attribute.Value;
            AppendOrWriteAttribute (output, attribute.Value, redrawTextStyle);
            redrawTextStyle = attribute.Value.Style;
        }

        // Add the grapheme
        string grapheme = cell.Grapheme;
        output.Append (grapheme);

        // Handle wide grapheme
        if (grapheme.GetColumns () > 1 && currentCol + 1 < maxCol)
        {
            currentCol++; // Skip next cell for wide character
        }
    }

 /// <inheritdoc/>
 protected override void AppendOrWriteAttribute (StringBuilder output, Attribute attr, TextStyle redrawTextStyle)
 {
     if (Force16Colors)
     {
         if (IsLegacyConsole)
         {
             Write (output);
             output.Clear ();
             var as16ColorInt = (ushort)((int)attr.Foreground.GetClosestNamedColor16 () | ((int)attr.Background.GetClosestNamedColor16 () << 4));
             SetConsoleTextAttribute (_screenBuffer, as16ColorInt);
         }
         ...
 }
```

When i refactored Output.Write, I introduced a bug here in the IsLegacyConsole path. 

When this call happens in OutputBase.Write, the cursor is not in the right place.

image

tig:
Yep. Can repro with Example.cs modified like this:

```cs
 public ExampleWindow ()
 {
     Arrangement = ViewArrangement.Movable | ViewArrangement.RightResizable | ViewArrangement.BottomResizable;
     X = Pos.Center ();
     Y = Pos.Center ();
     Width = Dim.Percent (80);
     Height= Dim.Percent (80);
```

So it has nothing to do with MB or IsModal. It has something to do with 

```
    /// <param name="lastAttr">The last attribute used, updated if the cell's attribute is different.</param>
```

And it has nothing to do with TextStyle itself, because If I use the 8-bit theme, where no text styles are specified it repros:

cmd_ifKWiHH6tO

BDisp:
I don't think so because legacy console never call EscSeqUtils.CSI_AppendTextStyleChange. If I only click on the border without moving the mouse the color change from cyan to yellow without corrupt the output. Only when the mouse is moved then the output is corrupted.

BDisp:
Probably some incorrect cursor position. With legacy console it's only needed to output immediately if the attribute change which it's needed to call SetConsoleTextAttribute to set the new attribute. So, after the write, the cursor probably isn't positioned in the last column.

tig:
> Probably some incorrect cursor position. With legacy console it's only needed to output immediately if the attribute change which it's needed to call SetConsoleTextAttribute to set the new attribute. So, after the write, the cursor probably isn't positioned in the last column.

Yea, i tried to figure out how put a call to `SetCursorPositionImpl` in the IsLegacy path to no avail.

For now this workaround works:

```cs
    /// <inheritdoc cref="IOutput.Write(IOutputBuffer)"/>
    public virtual void Write (IOutputBuffer buffer)
    {
        var top = 0;
        var left = 0;
        int rows = buffer.Rows;
        int cols = buffer.Cols;
        var output = new StringBuilder ();
        Attribute? redrawAttr = null;
        int lastCol = -1;

        if (IsLegacyConsole)
        {
            // BUGBUG: This is a workaround for some regression in legacy console mode where
            // BUGBUG: dirty cells are not handled correctly. Mark all cells dirty as a workaround.
            lock (buffer.Contents!)
            {
                for (var row = 0; row < buffer.Rows; row++)
                {
                    for (var c = 0; c < buffer.Cols; c++)
                    {
                        buffer.Contents [row, c].IsDirty = true;
                    }

                }
            }
        }

        SetCursorVisibility (CursorVisibility.Invisible);

        for (int row = top; row < rows; row++)
```
This basically means there's no IsDirty optimzation for when IsLegacyConsole = true. 

I'm going to merge this as-is, if you're ok with that. We can debug the problem later.

BDisp:
This is better:

```cs
    public virtual void Write (IOutputBuffer buffer)
    {
        var top = 0;
        var left = 0;
        int rows = buffer.Rows;
        int cols = buffer.Cols;
        var output = new StringBuilder ();
        Attribute? redrawAttr = null;
        int lastCol = -1;

        SetCursorVisibility (CursorVisibility.Invisible);

        for (int row = top; row < rows; row++)
        {
            var lc = lastCol > -1 ? lastCol : 0;

            if (!SetCursorPositionImpl (IsLegacyConsole ? lc : 0, row))
            {
                return;
            }

            output.Clear ();

            for (int col = left; col < cols; col++)

```

tig:
> ```cs
>  for (int row = top; row < rows; row++)
>         {
>             var lc = lastCol > -1 ? lastCol : 0;
> 
>             if (!SetCursorPositionImpl (IsLegacyConsole ? lc : 0, row))
>             {
>                 return;
>             }
> ```

Slightly but still not right:

cmd_TMlTS2l6I3

I'm going to leave my workaround in for now.

PR #4451, opened 2025-12-06
Fixes #4449. Regression in IsLegacyConsole mode where dirty cells are not handled correctly

### LINKED ISSUE #4449: Regression in `IsLegacyConsole` mode where dirty cells are not handled correctly
cmd_lcPHaNrBD5

> Probably some incorrect cursor position. With legacy console it's only needed to output immediately if the attribute change which it's needed to call SetConsoleTextAttribute to set the new attribute. So, after the write, the cursor probably isn't positioned in the last column.

Yea, i tried to figure out how put a call to `SetCursorPositionImpl` in the IsLegacy path to no avail.

For now this workaround works:

```cs
    /// <inheritdoc cref="IOutput.Write(IOutputBuffer)"/>
    public virtual void Write (IOutputBuffer buffer)
    {
        var top = 0;
        var left = 0;
        int rows = buffer.Rows;
        int cols = buffer.Cols;
        var output = new StringBuilder ();
        Attribute? redrawAttr = null;
        int lastCol = -1;

        if (IsLegacyConsole)
        {
            // BUGBUG: This is a workaround for some regression in legacy console mode where
            // BUGBUG: dirty cells are not handled correctly. Mark all cells dirty as a workaround.
            lock (buffer.Contents!)
            {
                for (var row = 0; row < buffer.Rows; row++)
                {
                    for (var c = 0; c < buffer.Cols; c++)
                    {
                        buffer.Contents [row, c].IsDirty = true;
                    }

                }
            }
        }

        SetCursorVisibility (CursorVisibility.Invisible);

        for (int row = top; row < rows; row++)
```
This basically means there's no IsDirty optimzation for when IsLegacyConsole = true. 

I'm going to merge this as-is, if you're ok with that. We can debug the problem later.

_Originally posted by @tig in https://github.com/gui-cs/Terminal.Gui/issues/4448#issuecomment-3619009386_

BDisp:
## Fixes

- Fixes #4449

## Proposed Changes/Todos

- [x] Add separated cursor position handled for IsLegacyConsole

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/OutputBase.cs
+++ b/Terminal.Gui/Drivers/OutputBase.cs
@@ -101,10 +101,7 @@ public abstract class OutputBase
                             lastCol++;
                         }
 
-                        if (IsLegacyConsole)
-                        {
-                            SetCursorPositionImpl (lastCol, row);
-                        }
+                        SetCursorPositionImpl (lastCol, row);
 
                         continue;
                     }
@@ -291,8 +288,6 @@ public abstract class OutputBase
         }
         else
         {
-            SetCursorPositionImpl (lastCol, row);
-
             // Wrap URLs with OSC 8 hyperlink sequences using the new Osc8UrlLinker
             StringBuilder processed = Osc8UrlLinker.WrapOsc8 (output);
             Write (processed);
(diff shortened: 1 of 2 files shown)


tig:
@BDisp you can ignore the failing test here. I need to update those tests to have better "test took too long" logic because sometimes the GH runners are just super slow.

Issue #4258, opened 2025-09-27
Two column glyphs break overlapped View rendering


tig:
I had fixed this ages ago, but the new driver code has somehow lost the correct behavor.

To repro, run charmap filter the view to "OtherSymbol"s. 

Open the detail dialog for any glyph.

Move it around and note how the border gets corrupted.

![Image](https://github.com/user-attachments/assets/84eb0171-1c95-42ee-b700-b194430bc530)

There used to be very-carefully-crafted logic that ensured that this didn't happen.

We need to re-fix this.

BDisp:
That seems that when a wider code point is intercepted with another code point at right, the wider code point can't be displayed because it's impossible to display half of a wider code point. Somewhere in the code it's replacing the visible part of the wider code point with the replacement code point. I think it's better to replace it with a space.

BDisp:
The coordinates sent when AddStr is called must be relative to the screen coordinates, not the dialog coordinates. Otherwise, it will compare and set the wrong cells.

This is what happened with older drivers because they properly used the IsDirty property. Then, when the entire screen was rendered, these flaws went unnoticed. Now, with the recent changes, these flaws are becoming evident again and need to be fixed.

BDisp:
In your WideGlyps scenario you're rebuilding the glyphs in the DrawingContent event and thus invalidating the arrangeableViews  IsDirty flag. What happening is that the app is draw the contents with the arrangeableViews (non-modal) which replace the wide glyph on the left with the replacement char. Then in the DrawingContent it's rebuilding the codepoints again and when a wide glyph is handled the AddStr method check if the current glyph is wide and if true invalidate the next column by setting IsDirty as false. So, in your scenario the arrangeableViews left border aren't output because IsDirty is false. But with a modal view should always the last being draw. I hope this explanation can help you some how.

tig:
> In your WideGlyps scenario you're rebuilding the glyphs in the DrawingContent event and thus invalidating the arrangeableViews IsDirty flag. What happening is that the app is draw the contents with the arrangeableViews (non-modal) which replace the wide glyph on the left with the replacement char. Then in the DrawingContent it's rebuilding the codepoints again and when a wide glyph is handled the AddStr method check if the current glyph is wide and if true invalidate the next column by setting IsDirty as false. So, in your scenario the arrangeableViews left border aren't output because IsDirty is false. But with a modal view should always the last being draw. I hope this explanation can help you some how.

Sorry, it doesn't. 

[...]

PR #4458, opened 2025-12-07
Fixes #4453. Regression in wide glyph rendering on all drivers

### LINKED ISSUE #4453: Regression in wide glyph rendering on all drivers
I believe this is caused by the fix @BDisp put in in #4451. Rats.

Image

BDisp:
## Fixes

- Fixes #4453

## Proposed Changes/Todos

- [x] Fix outputWidth not updating with wide characters.
- [x] Fix current col not being set the IsDirty property to true with wide characters because the col is incremented and was this one that is set the IsDirty property to true, when AddStr already handle that.

## Pull Request checklist:

### DIFF
--- a/Examples/UICatalog/Scenarios/WideGlyphs.cs
+++ b/Examples/UICatalog/Scenarios/WideGlyphs.cs
@@ -1,0 +1,129 @@
+#nullable enable
+
+using System.Text;
+
+namespace UICatalog.Scenarios;
+
+[ScenarioMetadata ("WideGlyphs", "Demonstrates wide glyphs with overlapped views & clipping")]
+[ScenarioCategory ("Unicode")]
+[ScenarioCategory ("Drawing")]
+
+public sealed class WideGlyphs : Scenario
+{
+    private Rune [,]? _codepoints;
+
+    public override void Main ()
+    {
+        // Init
+        Application.Init ();
+
+        // Setup - Create a top-level application window and configure it.
+        Window appWindow = new ()
+        {
+            Title = GetQuitKeyAndName (),
+            BorderStyle = LineStyle.None
+        };
+
+        // Build the array of codepoints once when subviews are laid out
+        appWindow.SubViewsLaidOut += (s, e) =>
+        {
+            View? view = s as View;
+            if (view is null)
+            {
+                return;
+            }
+
+            // Only rebuild if size changed or array is null
+            if (_codepoints is null || 
+                _codepoints.GetLength (0) != view.Viewport.Height || 
+                _codepoints.GetLength (1) != view.Viewport.Width)
+            {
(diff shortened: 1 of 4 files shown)


tig (approved):
Nice!

tig:
@BDisp this looks good to me. Ready to merge?

tig:
Actually, I just tested this locally, and it does not appear to fix the problem?

WindowsTerminal_fE14HmVoM2

What am I missing?

BDisp:
> Nice!

I've more fixes but I can include in the Pixel fix.

BDisp:
That's another bug with the borders as I already explained.

tig:
Note, to repro this, the terminal has to be just the right width to cause the border to be between the two cells an underlying two-column glyph is occupying:

image

IOW, this is most definitely #4258.

If you move the mouse over the dialog's border after this is drawn incorrectly the highlight states logic causes another iteration to redraw and it fixes it. This is why this was masked before I optimized the MainLoop.... the repeated redraws were masking this issue.

BDisp:
Then find a way to mask correctly on the first iteration. In the first iteration LineCanvas is send the wrong Dialog coordinates.

tig:
> Then find a way to mask correctly on the first iteration. In the first iteration LineCanvas is send the wrong Dialog coordinates.

I don't think so. I just wrote a scenario that proves that LineCanvas is fine. The problem has to do with what's described in #4258.

image

The two vertical lines are just `Line` views. The two boxes are Views with borders. 

The problem is in `OutputBufferImp.AddString`. A long time ago I put logic in that ensure that when a wide Rune was added, if it was 1 cell to the left of the edge of the clip rectangle it would output nonprintable. Somewhere we lost this code which is why I opened #4258. 

To fix this properly we need to re-integrate clip region checking into OutputBufferImpl.AddString.

tig:
I just added a Scenario that might help you track this down further as a PR to this PR.

tig:
Reviewing the latest OutputBufferImpl.AddString code, I now think you may be right, or almost right... It does appear to have the correct logic. 

The problem is likely that the Border drawing logic is not setting the clip correctly. Or the code in DoDrawAdornments is not doing somehing correct with the clip region.

tig:
Here's what I propose:

You have fixed the regression I caused with this fix.

That exposed that #4258 still has not been fixed.

Let's merge this fix and work on #4258 separately. 

It's an ugly bug and will likely take some deep thought to fix.

What do you think?

BDisp:
> Here's what I propose:
> 
> You have fixed the regression I caused with this fix.
> 
> That exposed that #4258 still has not been fixed.
> 
> Let's merge this fix and work on #4258 separately.
> 
> It's an ugly bug and will likely take some deep thought to fix.
> 
> What do you think?

Whatever is best for you is fine with me too.

PR #4462, opened 2025-12-07
Fixes #4258 - Glyphs drawn at mid-point of wide glyphs don't get drawn with clipping

### LINKED ISSUE #4258: Two column glyphs break overlapped View rendering

tig:
## Fixes

- Fixes #4258 

## Problem

Content drawn at odd columns (1, 3, 5, etc.) that overlap with wide glyphs (e.g., 🍎 which spans 2 columns) would not render correctly. The content would either not appear at all or render with incorrect colors.

**Example:** Border at column 1, with 🍎 at column 0
- Expected: `�┌─┐🍎` (replacement char, border, apple)
- Actual: `�─┐🍎` (missing left border `┌`)

**Affected content:**
- Border characters (corners, verticals) at odd columns
- Text characters
- Line drawing from `LineCanvas`
- Any single-width Unicode

## Root Cause

In `OutputBufferImpl.AddStr()`, when writing a wide character at column N:

1. Wrote the wide glyph to column N ✓
2. **Wrote replacement char (`�`) to column N+1** ❌
3. **Set column N+1 to `IsDirty = false`** ❌

**The bug:** When content was later drawn at column N+1:
- Content wrote correctly and marked `IsDirty = true`
- But when the view redraws, wide glyph code overwrote column N+1 again
- Set `IsDirty = false`, causing `OutputBase.Write()` to skip it
- Result: Content at column N+1 never appeared

## Fix

**Three changes to `OutputBufferImpl.AddStr()`:**

1. **Removed code writing replacement char to column N+1**
   - Let column N+1 retain whatever was drawn there
   
2. **Removed code setting column N+1 to `IsDirty = false`**
   - Let the wide glyph naturally render across both columns
   
3. **Moved `Col++` inside the lock**
   - Fixes race condition in parallel tests
   - Synchronizes cursor position with buffer access

**Why it works:**
- Wide glyphs at column N naturally render across columns N and N+1 in the terminal
- Column N+1 is left untouched - retains its grapheme, dirty state, and attribute
- If content is drawn at column N+1, it renders properly and overwrites the visual second half of the wide glyph

## Code Quality Improvements

Refactored `AddStr()` for readability (no functional changes):
- Extracted `AddGrapheme()` - per-grapheme logic
- Extracted `InvalidateOverlappedWideGlyph()` - handles overlap detection
- Extracted `WriteGraphemeByWidth()`, `WriteSingleWidthGrapheme()`, `WriteWideGrapheme()` - separates concerns by character width
- Refactored `NetOutput`, `FakeOutput`, `UnixOutput`, and `WindowsOutput` classes to support access to `Output` and added a `IDriver.GetOutput` to acess the `IOutput`. `IOutput` now has a `GetLastOutput` method.

## Testing

**New tests in `ViewDrawingClippingTests.cs`:**
- `Draw_WithBorderSubView_At_Col1_In_WideGlyph_DrawsCorrectly`
- `Draw_WithBorderSubView_At_Col3_In_WideGlyph_DrawsCorrectly`
- `All_Drivers_When_Clipped_AddStr_Glyph_On_Second_Cell_Of_Wide_Glyph_Outputs_Correctly` tests that this bug is fixed for all drivers.

**Updated test in `OutputBaseTests.cs`:**
- `Write_Virtual_Or_NonVirtual_Uses_WriteToConsole_And_Clears_Dirty_Flags_Mixed_Graphemes` - now expects correct behavior

**New test infrastructure:**
- `DriverAssert.AssertDriverOutputIs()` - verifies **Actual Driver Output** as ANSI output (not just `Contents` state).
- Currently only works with FakeDriver/FakeOutput.

### DIFF
--- a/Terminal.Gui/Drivers/DotNetDriver/NetOutput.cs
+++ b/Terminal.Gui/Drivers/DotNetDriver/NetOutput.cs
@@ -109,10 +109,9 @@ public class NetOutput : OutputBase, IOutput
     /// <inheritdoc />
     protected override void Write (StringBuilder output)
     {
+        base.Write (output);
         try
         {
-            _outputStringBuilder.Clear ();
-            _outputStringBuilder.Append (output.ToString ());
             Console.Out.Write (output);
         }
         catch (IOException)
@@ -120,11 +119,6 @@ public class NetOutput : OutputBase, IOutput
             // Not connected to a terminal; do nothing
         }
     }
-
-    private readonly StringBuilder _outputStringBuilder = new();
-
-    /// <inheritdoc />
-    public string GetLastOutput () => _outputStringBuilder.ToString ();
 
     /// <inheritdoc />
     protected override bool SetCursorPositionImpl (int col, int row)
@@ -147,7 +141,7 @@ public class NetOutput : OutputBase, IOutput
             }
             catch (Exception)
             {
-                return false;
+                return true;
             }
         }
 
--- a/Terminal.Gui/Drivers/DriverImpl.cs
+++ b/Terminal.Gui/Drivers/DriverImpl.cs
@@ -73,7 +73,10 @@ internal class DriverImpl : IDriver
     public void Init () { throw new NotSupportedException (); }
 
     /// <inheritdoc/>
-    public void Refresh () { _output.Write (OutputBuffer); }
+    public void Refresh ()
+    {
+        _output.Write (OutputBuffer);
+    }
 
     /// <inheritdoc/>
     public string? GetName () => GetInputProcessor ().DriverName?.ToLowerInvariant ();
--- a/Terminal.Gui/Drivers/FakeDriver/FakeOutput.cs
+++ b/Terminal.Gui/Drivers/FakeDriver/FakeOutput.cs
@@ -7,7 +7,7 @@ namespace Terminal.Gui.Drivers;
 /// </summary>
 public class FakeOutput : OutputBase, IOutput
 {
-    private readonly StringBuilder _outputStringBuilder = new ();
+   // private readonly StringBuilder _outputStringBuilder = new ();
     private int _cursorLeft;
     private int _cursorTop;
     private Size _consoleSize = new (80, 25);
@@ -28,8 +28,8 @@ public class FakeOutput : OutputBase, IOutput
     /// </summary>
     public IOutputBuffer? GetLastBuffer () => _lastBuffer;
 
-    /// <inheritdoc />
-    public string GetLastOutput () => _outputStringBuilder.ToString ();
+    ///// <inheritdoc cref="IOutput.GetLastOutput"/>
+    //public override string GetLastOutput () => _outputStringBuilder.ToString ();
 
     /// <inheritdoc />
     public Point GetCursorPosition ()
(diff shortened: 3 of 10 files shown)


BDisp:
I think the most correct solution is PR #4465. I'm not referring to code optimization, but rather to the most logical approach to handling wide glyphs. The implementation of region clipping is indeed one of the most important features, and it practically doesn't need to respect the hierarchy, but when wide glyphs are altered by overlapping other views, then it's necessary to respect the hierarchy. Furthermore, the margin doesn't need special treatment compared to other adornments.

tig:
I appreciate you think you are completely correct in this matter and unwilling to consider the evidence.

I believe:

1) In this PR, the tests that I crated definitively prove that the bug I was talking about has been fixed. 
2) You see a different bug. 

Please look at the tests that I created that 1000000% reproduce the drawing problem, at the Driver output point, without using any View hierarchy at all. Those tests pass once the fix is in place. They fail when the fix is not in place. 

Note how in these tests, I'm actually comparing the contents of `Contents` (which are always correct even with the buggy code in places) and `IOutput.Write` via the new  `DriverAssert.AssertDriverOutputIs`. 

Your code in #4465 breaks clipping, Shadows, and a bunch of other things because it completely un-does how Clipping works. I do not think you actually understand how Clipping works. You say " This principle has been in place since version v1", but this is not true. I reversed this principle when I implemented the new Draw flow, with clipping in this PR: 

- https://github.com/gui-cs/Terminal.Gui/pull/3798

I have reproduced the bug you are seeing by playing with the ChineseUI Scenario as you show. 

I can reproduce it without moving the MessageBox near the buttons. Just by hovering the mouse over the Ok button in the Message box. This indicates the bug is related to `HighlightStates` and wide characters. This is a DIFFERENT bug than the one I very carefully, with very clear documentation, and very clear (and very simple) unit tests, fixed in this PR.

If you can show me a simple unit test that reproduces the HighlightStates bug you seeing, without having to drag the mouse around in a Scenario, and the fix for it also fixes what my code fixes, I'll change my mind. We now have all the test infrastructure to test things like this without having to resort to user testing. Please see `DriverAssert.AssertDriverOutputIs`.

BDisp:
Another issue where glyphs disappear from the screen. The next column must be set with IsDirty as false and if any view want to draw on that cell it set IsDirty as true, otherwise you'll see that flickering.

VsDebugConsole_6FNYGcpwwG

tig:
Here's the latest test that illustrates, at the lowest level possible the bug. Without the fix this test fails. With it, it passes. 

Note it does not use anything but `IDriver.AddStr` and I've upgrade the driver infrastrcuture such that we can test the output of all 4 drivers in tests like this using `DriverAssert.AssertDriverContentsAre`.

```cs

    // Tests fix for https://github.com/gui-cs/Terminal.Gui/issues/4258
    [Theory]
    [InlineData ("fake")]
    [InlineData ("windows")]
    [InlineData ("dotnet")]
    [InlineData ("unix")]
    public void All_Drivers_When_Clipped_AddStr_Glyph_On_Second_Cell_Of_Wide_Glyph_Outputs_Correctly (string driverName)
    {
        IApplication? app = Application.Create ();
        app.Init (driverName);
        IDriver driver = app.Driver!;

        // Need to force "windows" driver to override legacy console mode for this test
        driver.IsLegacyConsole = false;
        driver.Force16Colors = false;

        driver.SetScreenSize (6, 3);

        driver!.Clip = new (driver.Screen);

        driver.Move (1, 0);
        driver.AddStr ("┌");
        driver.Move (2, 0);
        driver.AddStr ("─");
        driver.Move (3, 0);
        driver.AddStr ("┐");
        driver.Clip.Exclude (new Region (new (1, 0, 3, 1)));

        driver.Move (0, 0);
        driver.AddStr ("🍎🍎🍎🍎");


        DriverAssert.AssertDriverContentsAre (
                                              """
                                              �┌─┐🍎
                                              """,
                                              output,
                                              driver);

        driver.Refresh ();

        DriverAssert.AssertDriverOutputIs (@"\x1b[38;2;0;0;0m\x1b[48;2;0;0;0m�┌─┐🍎\x1b[38;2;255;255;255m\x1b[48;2;0;0;0m",
                                           output, driver);
    }
}
```

PR #4465, opened 2025-12-08
Fixes #4502. Shadows in overlapping views are not rendered consistently

### LINKED ISSUE #4502: Shadows in overlapping views are not rendered consistently
The issue with shadow on overlapped views is fixed in the commit https://github.com/gui-cs/Terminal.Gui/pull/4465/commits/c3bc46eea1cf942f0cda6acc802ca703c718f8dc

Before the fix:

Image

After the fix:

WindowsTerminal_6hLOnzUiPl

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/4465#issuecomment-3663249237_

BDisp:
## Fixes

- Fixes #4502

## Proposed Changes/Todos

- [ x] Reversing the hierarchy of runnables so that the current runnable is the last to be drawn fixed the margin shadow issue.

## Pull Request checklist:

### DIFF
--- a/Examples/UICatalog/Scenarios/Menus.cs
+++ b/Examples/UICatalog/Scenarios/Menus.cs
@@ -347,6 +347,8 @@ public class Menus : Scenario
             };
 
             ContextMenu.EnableForDesign (ref host);
+            Application.Popover.Register (ContextMenu);
+
             ContextMenu.Visible = false;
 
             // Demo of PopoverMenu as a context menu
--- a/Terminal.Gui/App/IPopover.cs
+++ b/Terminal.Gui/App/IPopover.cs
@@ -8,54 +8,64 @@ namespace Terminal.Gui.App;
 ///     <para>
 ///         A popover is a transient UI element that appears above other content to display contextual information or UI,
 ///         such as menus, tooltips, or dialogs.
-///         Popovers are managed by <see cref="ApplicationPopover"/> and are typically shown using
-///         <see cref="ApplicationPopover.Show"/>.
 ///     </para>
 ///     <para>
-///         Popovers are not modal; they do not block input to the rest of the application, but they do receive focus and
-///         input events while visible.
-///         When a popover is shown, it is responsible for handling its own layout and content.
+///         <b>IMPORTANT:</b> Popovers must be registered with <see cref="Application.Popover"/> using
+///         <see cref="ApplicationPopover.Register"/> before they can be shown with <see cref="ApplicationPopover.Show"/>.
 ///     </para>
 ///     <para>
+///         <b>Lifecycle:</b><br/>
+///         When registered, the popover's lifetime is managed by the application. Registered popovers are
+///         automatically disposed when <see cref="Application.Shutdown"/> is called. Call
+///         <see cref="ApplicationPopover.DeRegister"/> to manage the lifetime directly.
+///     </para>
+///     <para>
+///         <b>Visibility and Hiding:</b><br/>
 ///         Popovers are automatically hidden when:
-///         <list type="bullet">
-///             <item>The user clicks outside the popover (unless occluded by a subview of the popover).</item>
-///             <item>The user presses <see cref="Application.QuitKey"/> (typically <c>Esc</c>).</item>
-///             <item>Another popover is shown.</item>
-///         </list>
 ///     </para>
+///     <list type="bullet">
+///         <item>The user clicks outside the popover (unless clicking on a subview).</item>
+///         <item>The user presses <see cref="Application.QuitKey"/> (typically <c>Esc</c>).</item>
+///         <item>Another popover is shown.</item>
+///         <item><see cref="View.Visible"/> is set to <see langword="false"/>.</item>
+///     </list>
 ///     <para>
 ///         <b>Focus and Input:</b><br/>
-///         When visible, a popover receives focus and input events. If the user clicks outside the popover (and not on a
-///         subview),
-///         presses <see cref="Application.QuitKey"/>, or another popover is shown, the popover will be hidden
-///         automatically.
+///         Popovers are not modal but do receive focus and input events while visible.
+///         Registered popovers receive keyboard events even when not visible, enabling global hotkey support.
 ///     </para>
 ///     <para>
 ///         <b>Layout:</b><br/>
-///         When the popover becomes visible, it is automatically laid out to fill the screen by default. You can override
-///         this behavior
-///         by setting <see cref="View.Width"/> and <see cref="View.Height"/> in your derived class.
+///         When becoming visible, popovers are automatically laid out to fill the screen by default.
+///         Override <see cref="View.Width"/> and <see cref="View.Height"/> to customize size.
 ///     </para>
 ///     <para>
-///         <b>Mouse:</b><br/>
-///         Popovers are transparent to mouse events (see <see cref="ViewportSettingsFlags.TransparentMouse"/>),
-///         meaning mouse events in a popover that are not also within a subview of the popover will not be captured.
+///         <b>Mouse Events:</b><br/>
+///         Popovers use <see cref="ViewportSettingsFlags.TransparentMouse"/>, meaning mouse events
+///         outside subviews are not captured.
 ///     </para>
 ///     <para>
-///         <b>Custom Popovers:</b><br/>
-///         To create a custom popover, inherit from <see cref="PopoverBaseImpl"/> and add your own content and logic.
+///         <b>Creating Custom Popovers:</b><br/>
+///         Inherit from <see cref="PopoverBaseImpl"/> and add your own content and logic.
 ///     </para>
 /// </remarks>
 public interface IPopover
 {
     /// <summary>
-    ///     Gets or sets the <see cref="Current"/> that this Popover is associated with. If null, it is not associated with
-    ///     any Runnable and will receive all keyboard
-    ///     events from the <see cref="IApplication"/>. If set, it will only receive keyboard events the Runnable would normally
-    ///     receive.
-    ///     When <see cref="ApplicationPopover.Register"/> is called, the <see cref="Current"/> is set to the current
-    ///     <see cref="IApplication.TopRunnableView"/> if not already set.
+    ///     Gets or sets the <see cref="IRunnable"/> that this popover is associated with.
     /// </summary>
+    /// <remarks>
+    ///     <para>
+    ///         If <see langword="null"/>, the popover is not associated with any runnable and will receive all keyboard
+    ///         events from the application.
+    ///     </para>
+    ///     <para>
+    ///         If set, the popover will only receive keyboard events when the associated runnable is active.
+    ///     </para>
+    ///     <para>
+    ///         When <see cref="ApplicationPopover.Register"/> is called, this property is automatically set to
+    ///         <see cref="IApplication.TopRunnableView"/> if not already set.
+    ///     </para>
+    /// </remarks>
     IRunnable? Current { get; set; }
 }
--- a/Terminal.Gui/App/PopoverBaseImpl.cs
+++ b/Terminal.Gui/App/PopoverBaseImpl.cs
@@ -2,36 +2,36 @@
 namespace Terminal.Gui.App;
 
 /// <summary>
-///     Abstract base class for popover views in Terminal.Gui.
+///     Abstract base class for popover views in Terminal.Gui. Implements <see cref="IPopover"/>.
 /// </summary>
 /// <remarks>
 ///     <para>
-///         <b>Popover Lifecycle:</b><br/>
-///         To display a popover, use <see cref="ApplicationPopover.Show"/>. To hide a popover, either call
-///         <see cref="ApplicationPopover.Hide"/>,
-///         set <see cref="View.Visible"/> to <see langword="false"/>, or show another popover.
+///         <b>IMPORTANT:</b> Popovers must be registered with <see cref="Application.Popover"/> using
+///         <see cref="ApplicationPopover.Register"/> before they can be shown.
 ///     </para>
 ///     <para>
-///         <b>Focus and Input:</b><br/>
-///         When visible, a popover receives focus and input events. If the user clicks outside the popover (and not on a
-///         subview),
-///         presses <see cref="Application.QuitKey"/>, or another popover is shown, the popover will be hidden
-///         automatically.
+///         <b>Requirements:</b><br/>
+///         Derived classes must:
 ///     </para>
+///     <list type="bullet">
+///         <item>Set <see cref="View.ViewportSettings"/> to include <see cref="ViewportSettingsFlags.Transparent"/> and <see cref="ViewportSettingsFlags.TransparentMouse"/>.</item>
+///         <item>Add a key binding for <see cref="Command.Quit"/> (typically bound to <see cref="Application.QuitKey"/>).</item>
+///     </list>
 ///     <para>
-///         <b>Layout:</b><br/>
-///         When the popover becomes visible, it is automatically laid out to fill the screen by default. You can override
-///         this behavior
-///         by setting <see cref="View.Width"/> and <see cref="View.Height"/> in your derived class.
+///         <b>Default Behavior:</b><br/>
+///         This base class provides:
 ///     </para>
+///     <list type="bullet">
+///         <item>Fills the screen by default (<see cref="View.Width"/> = <see cref="Dim.Fill"/>, <see cref="View.Height"/> = <see cref="Dim.Fill"/>).</item>
+///         <item>Transparent viewport settings for proper mouse event handling.</item>
+///         <item>Automatic layout when becoming visible.</item>
+///         <item>Focus restoration when hidden.</item>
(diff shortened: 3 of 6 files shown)


BDisp:
This fix all the wide glyphs issue without breaks clipping.

VsDebugConsole_y3wp6QlDEv
VsDebugConsole_vFm6NsgNEE

BDisp:
This was working perfectly locally but something with the merge broke it. I'll take a look.

BDisp:
One of the advantages of this PR is that the margins are drawn during the superview and subview cycle, thus avoiding storing all the margin clipping regions and only drawing them at the end after all the runnables have been drawn.

Terminal.Gui/ViewBase/View.Drawing.cs
```diff
@@ -80,6 +77,11 @@ public void Draw (DrawContext? context = null)
         }
         Region? originalClip = GetClip ();
 
+        if (this is not Adornment && SuperView is null && Driver is { })
+        {
+            originalClip = new (new (Driver.Screen.Location, Driver.Screen.Size));
+        }
```
BDisp: This part of the code does not break any region clipping code, as it only executes if it is not a adornment and the superview is null, to allow any dialog box, a floating window, etc., to know the screen dimensions so that it can be drawn.

BDisp:
The commit https://github.com/gui-cs/Terminal.Gui/pull/4465/commits/52028d753fafd8b9748a634dc680a115502d3821 fixes an issue where the attribute of the next column of a wide glyph was not assigned, and moving a view over it displayed an incorrect attribute.

The issue:

Image

After the fix:

Image

tig:
@BDisp this PR is impressive and a great refactoring of how Draw works to make it simpler. You also get all the credit for finding a fix for the actual bug!

Nice job!!

However, the bug reported in #4466 (which I've renamed now that we know the precise issue) had nothing to do with the `View.Draw` flow. 

The unit tests that I added to my PR:

- #4486

clearly reproduce the bug without using any `View` (or even `Application`) logic. And the fix clearly fixes the problem without changing any code in anything other than `OutputBufferImpl.cs` and `OutputBase.cs`.

Thank you for finding the actual fix; I was able to use that to recreate the tests and then independently implement the same fix. 

I suggest you rename this PR to "Simplifies `View.Draw` logic" and to update the first comment to be just about that.

Cool?

BDisp:
Done. If you think I should change anything else, please let me know.

tig:
Please code review

- #4486 

Once we agree that fixes the issue (I've tested the heck out of it, but may have missed something) I'll merge it.

Then, this PR can be just about refactoring the View.Draw flow.

BDisp:
I think it's done. What I don't understand is why, even though you don't find a PR as perfect as you'd like, but one that essentially solves many of the problems that are communicated and expressed, why don't you just merge it immediately so you can make your changes later? That way, you'd avoid having to resolve many merging conflicts, etc. Anyway, I think it's become a habit and sometimes discouraging.

tig:
This breaks `DrawIndicator` for any `IRunnable` that `IsModal`:

Before this PR:
WindowsTerminal_yBl1rEzHGR

After:
WindowsTerminal_TdTEbC7BEY

tig:
This PR is not done. I have not had a chance to review it. You changed some pretty fundamental logic here, that I spent a lot of time refining. I want to review it thoroughly to make sure I understand it and to ensure nothing is broken.

BDisp:
If it's broken, it's not my fault. In my PR, it's perfect.

WindowsTerminal_A4JOxaqNTL

BDisp:
I got confused with the Arrange button. The DrawIndicator isn't showing up. I'll take a look.

tig:
Shadows are not being drawn transparent. IIRC, *this* is why I had the logic to reverse the views and the out-of-band `Margin` clipping regions.

Before:
WindowsTerminal_B71PblN1d9

After:
WindowsTerminal_Hs2QC2YQkW

BDisp:
I don't think this is actually a bug. If the view you're moving belongs to the same superview as the other view you're overlaying, it has to be added last if you want the desired effect. If that's the case, then you have to add that view to the end of the list to be the last one drawn. Otherwise, you'll be going against the logic of subviews belonging to the same superview. If you contradict this, then you'll be making patchwork solutions that could cause other issues.

BDisp:
The way to do what you want, if you want the current view that is being overlaid on the others, is to change its position in the subview index when you click on it, so that it is the last one to be drawn, thus overlaying other windows.

BDisp:
I just tested it with the v2_develop branch without merged your PR, and the DrawIndicator doesn't appear either. Therefore, this issue already existed and has nothing to do with this PR.

tig:
Re: Transparent Shadows. This test succeeds on v2_develop but fails on this branch. Perhaps you can use it to track down how to fix them in this PR.

```cs
[Fact]
public void TransparentShadow_Draws_Transparent_At_Driver_Output ()
{
    // Arrange
    IApplication app = Application.Create ();
    app.Init ("fake");
    app.Driver!.SetScreenSize (5, 3);

    // Force 16-bit colors off to get predictable RGB output
    app.Driver.Force16Colors = false;

    var superView = new Runnable
    {
        Width = Dim.Fill (),
        Height = Dim.Fill (),
        Text = "ABC".Repeat (40)!
    };
    superView.SetScheme (new (new Attribute (Color.White, Color.Blue)));
    superView.TextFormatter.WordWrap = true;

    // Create an overlapped view with transparent shadow
    var overlappedView = new View
    {
        Width = 4,
        Height = 2,
        Text = "123",
        Arrangement = ViewArrangement.Overlapped,
        ShadowStyle = ShadowStyle.Transparent
    };
    overlappedView.SetScheme (new (new Attribute (Color.Black, Color.Green)));

    superView.Add (overlappedView);

    // Act
    SessionToken? token = app.Begin (superView);
    app.LayoutAndDraw ();
    app.Driver.Refresh ();

    // Assert
    _output.WriteLine ("Actual driver contents:");
    _output.WriteLine (app.Driver.ToString ());
    _output.WriteLine ("\nActual driver output:");
    string? output = app.Driver.GetOutput ().GetLastOutput ();
    _output.WriteLine (output);

    DriverAssert.AssertDriverOutputIs ("""
                                       \x1b[38;2;0;0;0m\x1b[48;2;0;128;0m123\x1b[38;2;0;0;0m\x1b[48;2;189;189;189mA\x1b[38;2;0;0;255m\x1b[48;2;255;255;255mBC\x1b[38;2;0;0;0m\x1b[48;2;189;189;189mABC\x1b[38;2;0;0;255m\x1b[48;2;255;255;255mABCABC
                                       """, _output, app.Driver);

    // The output should contain ANSI color codes for the transparent shadow
    // which will have dimmed colors compared to the original
    Assert.Contains ("\x1b[38;2;", output); // Should have RGB foreground color codes
    Assert.Contains ("\x1b[48;2;", output); // Should have RGB background color codes

    // Verify driver contents show the background text in shadow areas
    int shadowX = overlappedView.Frame.X + overlappedView.Frame.Width;
    int shadowY = overlappedView.Frame.Y + overlappedView.Frame.Height;

    Cell shadowCell = app.Driver.Contents! [shadowY, shadowX];
    _output.WriteLine ($"\nShadow cell at [{shadowY},{shadowX}]: Grapheme='{shadowCell.Grapheme}', Attr={shadowCell.Attribute}");

    // The grapheme should be from background text
    Assert.NotEqual (string.Empty, shadowCell.Grapheme);
    Assert.Contains (shadowCell.Grapheme, "ABC"); // Should be one of the background characters

    // Cleanup
    if (token is { })
    {
        app.End (token);
    }

    superView.Dispose ();
    app.Dispose ();
}
```

tig:
> I just tested it with the v2_develop branch without merged your PR, and the DrawIndicator doesn't appear either. Therefore, this issue already existed and has nothing to do with this PR.

Cool! Thanks for checking.

BDisp:
@tig I renamed ShadowStyletests.cs file to ShadowTests.cs. If you look closely, it contains the tests that already existed plus the ones I added. If you don't mind, can I delete it?

tig:
> @tig I renamed ShadowStyletests.cs file to ShadowTests.cs. If you look closely, it contains the tests that already existed plus the ones I added. If you don't mind, can I delete it?

Yea, it needs to be renamed. Make sure you put the new transparent shadow test in there:

`TransparentShadow_Draws_Transparent_At_Driver_Output`

BDisp:
> > @tig I renamed ShadowStyletests.cs file to ShadowTests.cs. If you look closely, it contains the tests that already existed plus the ones I added. If you don't mind, can I delete it?
> 
> Yea, it needs to be renamed. Make sure you put the new transparent shadow test in there:
> 
> `TransparentShadow_Draws_Transparent_At_Driver_Output`

Sure. Regarding testing, it's not easy to pinpoint the error exactly with ANSI. That's why I prefer using the AssertDriverAttributesAre method. Since you created this new method, you should be more familiar with how to troubleshoot the error.

tig:
> > > @tig I renamed ShadowStyletests.cs file to ShadowTests.cs. If you look closely, it contains the tests that already existed plus the ones I added. If you don't mind, can I delete it?
> > 
> > 
> > Yea, it needs to be renamed. Make sure you put the new transparent shadow test in there:
> > `TransparentShadow_Draws_Transparent_At_Driver_Output`
> 
> Sure. Regarding testing, it's not easy to pinpoint the error exactly with ANSI. That's why I prefer using the AssertDriverAttributesAre method. Since you created this new method, you should be more familiar with how to troubleshoot the error.

AssertDriverAttributesAre only tests `Contents`. Because this use-case requires clipping to work propertly, we must test `Output`.

BDisp:
Another problem that might be present in this unit test is that you're only using one schema. I think it's assuming Normal, but internally, depending on whether the view has focus, or has some HotNormal attribute, etc., the application changes the attributes. It's not easy to decipher ANSI colors without consulting a table, and frankly, without documenting the color value of each ANSI code, I won't be able to find the flaw.

tig:
> Another problem that might be present in this unit test is that you're only using one schema. I think it's assuming Normal, but internally, depending on whether the view has focus, or has some HotNormal attribute, etc., the application changes the attributes. It's not easy to decipher ANSI colors without consulting a table, and frankly, without documenting the color value of each ANSI code, I won't be able to find the flaw.

You can ignore attributes here. Just the Graphmeme's matter. 

```cs
[Fact]
public void TransparentShadow_Draws_Transparent_At_Driver_Output ()
{
    // Arrange
    app.Init ("fake");
    app.Driver!.SetScreenSize (2, 1);
    app.Driver.Force16Colors = true;

    using Runnable superView = new ();
    superView.Width = Dim.Fill ();
    superView.Height = Dim.Fill ();
    superView.Text = "AB";
    superView.TextFormatter.WordWrap = true;
    superView.SetScheme (new (new Attribute (Color.Black, Color.White)));

    // Create view with transparent shadow
    View viewWithShadow = new ()
    {
        Width = Dim.Auto (),
        Height = Dim.Auto (),
        Text = "*",
        ShadowStyle = ShadowStyle.Transparent
    };
    // Make it so the margin is only on the right for simplicity
    viewWithShadow.Margin!.Thickness = new (0, 0, 1, 0);
    viewWithShadow.SetScheme (new (new Attribute (Color.Black, Color.White)));

    superView.Add (viewWithShadow);

    // Act
    app.Begin (superView);
    app.LayoutAndDraw ();
    app.Driver.Refresh ();

    // Assert
    _output.WriteLine ("Actual driver contents:");
    _output.WriteLine (app.Driver.ToString ());
    _output.WriteLine ("\nActual driver output:");
    string? output = app.Driver.GetOutput ().GetLastOutput ();
    _output.WriteLine (output);

    DriverAssert.AssertDriverOutputIs ("""
                                       \x1b[30m\x1b[107m*\x1b[90m\x1b[100mB
                                       """, _output, app.Driver);

}
```

Output:

```
Actual driver contents:
*B

Actual driver output:
\x1b[30m\x1b[107m*\x1b[90m\x1b[100mB
```

This makes it super obvious. If the `B` at the end is a ` `, the transparent shadow isn't working.

BDisp:
For subviews that are always on top and overlapping other views, the original code must be used at the end of each runnable. Similarly, with region clipping, only including the regions to be drawn and excluding the regions not to be drawn is irrelevant in the order in which they are drawn. For the same reason, it is irrelevant whether the content of the superview should be drawn before the subviews. So, now I recognize that with region clipping it works either way. I think I've introduced improvements and bug fixes. All the turbulence and instability observed was due to the bug with wide glyphs not rendering correctly, but I think I've helped improve it. That said, I think there's nothing else to fix and in that case it's ready to be reviewed. Thanks

tig (changes requested):
I'd appreciate it if you'd incorporate the latest simplified version of the test I put in the comments. Thanks.

BDisp:
@tig I tried reverting to the original logic (drawing from the most exposed views) but the problem reported in the comment https://github.com/gui-cs/Terminal.Gui/pull/4465#issuecomment-3648431414 is occurring. If you can resolve this using the original logic, I don't mind if this PR is reverted.

tig:
> @tig I tried reverting to the original logic (drawing from the most exposed views) but the problem reported in the comment [#4465 (comment)](https://github.com/gui-cs/Terminal.Gui/pull/4465#issuecomment-3648431414) is occurring. If you can resolve this using the original logic, I don't mind if this PR is reverted.

I'd like to get a low-level repro of what that comment shows. There's a real bug there somewhere. I'll see if I can find some time today to do so. I did some quick hacking this morning but wasn't able to come up with anything super clear. I'll be using this PR to play:

- #4490

I think there's some really good refactoring you've done in this PR, as well as some really good new tests. I'd rather not lose that work!

BDisp:
This is the unit test that test that issue. It's colors attribute that the only way to detect the issue. The 🦮 glyph is in a separated view with a different scheme to facilitate the test, which have the attribute 2 index.

https://github.com/BDisp/Terminal.Gui/blob/49cd6783024536e42cc82a2de2dbb74958ccd077/Tests/UnitTestsParallelizable/ViewBase/Adornment/BorderArrangementTests.cs#L95-L262

BDisp:
the shadows with overlapped  views is already fixed in this PR.

WindowsTerminal_5NnuLJCKUs

BDisp:
Thanks to you, this scenario where I implemented vertical transparency with two columns helped me a lot in fixing bugs.


WindowsTerminal_8Jy1uPlGuI

BDisp:
@tig, would you object if I created a new constructor for the Thickness class that accepts a Thickness object as a parameter? It becomes useful when we only want to make a copy without having to pass all sides. It would be like below:

```cs
    /// <summary>
    ///     Initializes a new instance of the <see cref="Thickness"/> class that is a copy of another <see cref="Thickness"/>.
    /// </summary>
    /// <param name="thickness"></param>
    public Thickness (Thickness thickness) { _sides = new (thickness.Left, thickness.Top, thickness.Right, thickness.Bottom); }
```

tig:
> @tig, would you object if I created a new constructor for the Thickness class that accepts a Thickness object as a parameter? It becomes useful when we only want to make a copy without having to pass all sides. It would be like below:
> 
> ```cs
>     /// <summary>
>     ///     Initializes a new instance of the <see cref="Thickness"/> class that is a copy of another <see cref="Thickness"/>.
>     /// </summary>
>     /// <param name="thickness"></param>
>     public Thickness (Thickness thickness) { _sides = new (thickness.Left, thickness.Top, thickness.Right, thickness.Bottom); }
> ```

Since Thickness is a record struct, you don't need a copy constructor. Record structs automatically provide value-based equality and copying semantics. You can simply copy a Thickness instance like this:

```cs
Thickness original = new (1, 2, 3, 4);
Thickness copy = original; // This creates a copy
```

BDisp:
The issue with shadow on overlapped views is fixed in the commit https://github.com/gui-cs/Terminal.Gui/pull/4465/commits/c3bc46eea1cf942f0cda6acc802ca703c718f8dc

WindowsTerminal_6hLOnzUiPl

tig:
What issue are you referring to?

BDisp:
> What issue are you referring to?

Please see the issue #4502. It draws over the shadow of the view that has the focus. It might be a little difficult to notice because the color black doesn't show up on top of another dark color, but if you look closely you'll see it. I remember you posting something similar before, but I can't remember where.

tig:
Got it. Nice job!

Should we update the title of this PR to match?

BDisp:
> Got it. Nice job!
> 
> Should we update the title of this PR to match?

Thanks. I also open the issue #4505 and I don't know if want I include the fix in this PR or not. If you don't agree, what title do you suggest?

tig:
> > Got it. Nice job!
> > Should we update the title of this PR to match?
> 
> Thanks. I also open the issue #4505 and I don't know if want I include the fix in this PR or not. If you don't agree, what title do you suggest?

If you think the code in this PR fixes #4505, then the title of this PR should match that. E.g "Fixes #4505 - ....".

BDisp:
> If you think the code in this PR fixes #4505, then the title of this PR should match that. E.g "Fixes #4505 - ....".

Thanks. You're right. I'll submit a separated PR then.

PR #4472, opened 2025-12-09
Fixes #4471, #4474, #3714, #2588 - MASSIVE: Refactors Mouse, Adds ANSI driver, Enables Input Injection

### LINKED ISSUE #2588: ProcessButtonClick strips any modifier keys from mouse events
**Describe the bug**

There seems to be a bug with WindowsDriver where shift/alt/ctrl are no longer passed through to mouse handlers. Culprit seems to be `ProcessButtonClick` method.

image

### LINKED ISSUE #3714: `MouseEvent`/`MouseClick` can't be externally controlled - was `Slider` does not seem to respect click handlers
**Describe the bug**
Not sure exactly what is going on but I am trying to stop clicks going through to SliderView in TGD.  This approach seems to work for other views.

I tried looking at gui.cs\UnitTests\Views\SliderTests.cs` for mouse tests but couldn't see any to adapt for repro.

```csharp
if (subView.GetType().IsGenericType(typeof(Slider<>)))
{
    // TODO: Does not seem to work
    subView.MouseEvent += (s, e) => SuppressNativeClickEvents(s, e,true);
    subView.MouseClick += (s, e) => SuppressNativeClickEvents(s,e, true);
}

...

private void SuppressNativeClickEvents(object? sender, MouseEventEventArgs obj, bool alsoSuppressClick = false)
{
    if (alsoSuppressClick)
    {
        obj.Handled = true;
    }
    else
    {
        // Suppress everything except single click (selection)
        obj.Handled = obj.MouseEvent.Flags != MouseFlags.Button1Clicked;
    }
}
```

### LINKED ISSUE #4308: View Arrangement with Keyboard not fully implemented
Looks like I didn't finish this...

If you enter arrange mode with the keyboard (CTRL-F5) the only time it currently works is if `View.Arrangement` is `ViewArrangement.Resizable`. If, for example, it's `ViewArrangement.LeftResizable` (what I want for the splitter in `FileDialog` to fix #4305) and the user presses CTRL-F5 nothing happens. 

This method needs to be flushed out:

```cs
/// <summary>
/// Sets button visibility for keyboard arrangement mode
/// </summary>
private void SetVisibilityForKeyboardMode ()
{
    if (_moveButton != null)
    {
        _moveButton.Visible = true;
    }

    if (_allSizeButton != null)
    {
        _allSizeButton.Visible = true;
    }
}
```

Here's the mouse version to illustrate:

```cs
 /// <summary>
 /// Sets button visibility based on the specified mouse arrangement mode
 /// </summary>
 private void SetVisibilityForMouseMode (ViewArrangement arrangement)
 {
     switch (arrangement)
     {
         case ViewArrangement.Movable:
             SetVisibleButton (_moveButton);
             break;

         case ViewArrangement.RightResizable | ViewArrangement.BottomResizable:
         case ViewArrangement.Resizable:
             SetVisibleButton (_rightSizeButton);
             SetVisibleButton (_bottomSizeButton);
             if (_allSizeButton != null)
             {
                 _allSizeButton.X = Pos.AnchorEnd ();
                 _allSizeButton.Y = Pos.AnchorEnd ();
                 _allSizeButton.Visible = true;
             }
             break;

         case ViewArrangement.LeftResizable:
             SetVisibleButton (_leftSizeButton);
             break;

         case ViewArrangement.RightResizable:
             SetVisibleButton (_rightSizeButton);
             break;

         case ViewArrangement.TopResizable:
             SetVisibleButton (_topSizeButton);
             break;

         case ViewArrangement.BottomResizable:
             SetVisibleButton (_bottomSizeButton);
             break;

         case ViewArrangement.LeftResizable | ViewArrangement.BottomResizable:
             SetVisibleButton (_leftSizeButton);
             SetVisibleButton (_bottomSizeButton);
             if (_allSizeButton != null)
             {
                 _allSizeButton.X = 0;
                 _allSizeButton.Y = Pos.AnchorEnd ();
                 _allSizeButton.Visible = true;
             }
             break;

         case ViewArrangement.LeftResizable | ViewArrangement.TopResizable:
             SetVisibleButton (_leftSizeButton);
             SetVisibleButton (_topSizeButton);
             break;

         case ViewArrangement.RightResizable | ViewArrangement.TopResizable:
             SetVisibleButton (_rightSizeButton);
             SetVisibleButton (_topSizeButton);
             if (_allSizeButton != null)
             {
                 _allSizeButton.X = Pos.AnchorEnd ();
                 _allSizeButton.Y = 0;
                 _allSizeButton.Visible = true;
             }
             break;
     }
 }
```

### LINKED ISSUE #4444: `WordWrap` for `View.Text` does not always re-wrap on layout/size changes
I'll expand when I get time, but I just noticed it.

### LINKED ISSUE #4471: Mouse support is seriously flawed and needs to be re-engineered
Original title: **`NumericUpDown` and other `WantContinuousButtonPressed = true` cases are not repeating**

I thought I had the mouse stuff sorted with

- #4469 

But apparently not. Not sure where I broke this, and confused as to why unit tests aren't finding it because I swear I built a bunch.

To repro run the NumericUpDown or DatePicker scenarios and hold the buttons down. No repeat.

# Issue

The way mouse events flow from the driver -> application -> view, and then all the hokey logic in `View.NewMouseEvent` is a mess. It barely works, can't be easily tested, and needs to be completely re done to make it right.

### LINKED ISSUE #4474: Duplicate Mouse Click Events on Double-Click
When a user double-clicks the mouse, v2_devellop raises duplicate click and double-click events. The drivers send 6 events instead of the expected 3-4:

**Current (Incorrect) Behavior:**
```
Button1Pressed
Button1Released
Button1Clicked        ← First duplicate
Button1Pressed
Button1Released
Button1DoubleClicked  ← Second duplicate
```

**Expected Behavior:**
```
Button1Pressed
Button1Released
Button1Released
Button1DoubleClicked
```

## Root Cause

The mouse event processing pipeline has **two** layers that both attempt to synthesize click events from press/release events:

1. **Platform Drivers** (WindowsDriver, CursesDriver, NetDriver) - Some drivers (especially WindowsDriver) synthesize their own `Clicked` and `DoubleClicked` events
2. **MouseInterpreter** - Always synthesizes `Clicked` and `DoubleClicked` events from `Pressed`/`Released` pairs

This causes duplication because both layers are doing the same work.

### Current Flow

```
┌──────────────────┐
│  Platform Driver │
│  (Windows/Unix)  │
└────────┬─────────┘
         │ Sends: Pressed, Released, [Clicked], [DoubleClicked]
         ↓
┌──────────────────┐
│ MouseInterpreter │ ← Process(MouseEventArgs e)
│                  │   - yield return e;  (passes through original)
│                  │   - Tracks button state
│                  │   - Synthesizes Clicked/DoubleClicked from Pressed/Released
└────────┬─────────┘
         │ Result: DUPLICATE events
         ↓
┌──────────────────┐
│   Application    │
│   (receives 6    │
│    events)       │
└──────────────────┘
```

## Code Location

The duplication logic is split between:

### 1. MouseInterpreter (Terminal.Gui/Drivers/MouseInterpreter.cs)

```csharp
public IEnumerable<MouseEventArgs> Process (MouseEventArgs e)
{
    yield return e;  // ← Passes through driver events (including any Clicked/DoubleClicked)

    // For each mouse button
    for (var i = 0; i < 4; i++)
    {
        _buttonStates [i].UpdateState (e, out int? numClicks);

        if (numClicks.HasValue)
        {
            yield return RaiseClick (i, numClicks.Value, e);  // ← ALSO synthesizes Clicked/DoubleClicked
        }
    }
}
```

### 2. MouseButtonStateEx (Terminal.Gui/Drivers/MouseButtonStateEx.cs)

```csharp
public void UpdateState (MouseEventArgs e, out int? numClicks)
{
    // ...
    if (Pressed)
    {
        // Click released
        numClicks = ++_consecutiveClicks;  // ← Synthesizes click on release
    }
    // ...
}
```

### 3. Platform Drivers

The drivers may also synthesize these events:

- **WindowsDriver**: Receives `DOUBLE_CLICK` flag from Windows API and translates to `Button1DoubleClicked`
- **CursesDriver**: Only sends Pressed/Released (no native double-click detection)
- **NetDriver**: Only sends Pressed/Released via ANSI sequences (no native double-click detection)

### Windows Console API
Native events from OS:
1. `MOUSE_BUTTON_PRESSED` → `Button1Pressed`
2. `MOUSE_BUTTON_RELEASED` → `Button1Released`
3. `DOUBLE_CLICK` flag → `Button1DoubleClicked`
4. `MOUSE_BUTTON_RELEASED` → `Button1Released`

### Unix/Linux (ncurses)
Native events from ncurses:
1. `BUTTON1_PRESSED` → `Button1Pressed`
2. `BUTTON1_RELEASED` → `Button1Released`
3. `BUTTON1_PRESSED` → `Button1Pressed` (second press)
4. `BUTTON1_RELEASED` → `Button1Released`

Note: ncurses does NOT detect double-clicks natively.

### .NET Console (ANSI/VT)
Native events from ANSI escape sequences:
1. `CSI < 0;x;y M` → `Button1Pressed`
2. `CSI < 0;x;y m` → `Button1Released`
3. `CSI < 0;x;y M` → `Button1Pressed` (second press)
4. `CSI < 0;x;y m` → `Button1Released`

Note: ANSI sequences do NOT include click/double-click information.

## Impact

This affects any code that handles mouse clicks:

```csharp
view.MouseEvent += (s, e) =>
{
    if (e.Flags.HasFlag(MouseFlags.Button1Clicked))
    {
        // This fires TWICE on a double-click!
        DoSomething();
    }
    
    if (e.Flags.HasFlag(MouseFlags.Button1DoubleClicked))
    {
        // This fires TWICE on a double-click!
        DoSomethingElse();
    }
};
```

Or when using mouse bindings:
```csharp
MouseBindings.Add(MouseFlags.Button1Clicked, Command.Activate);
// Command.Activate gets invoked twice on double-click
```

## Proposed Solution

**Option 1: Drivers Send Only Pressed/Released (Recommended)**

Have all drivers send ONLY `Pressed` and `Released` events, and let `MouseInterpreter` handle click synthesis uniformly:

1. Strip out any `Clicked`/`DoubleClicked`/`TripleClicked` event generation from all drivers
2. Drivers only send: `Pressed`, `Released`, `Moved`, `Wheel*` events
3. `MouseInterpreter` synthesizes all `Clicked`/`DoubleClicked`/`TripleClicked` events

**Pros:**
- Consistent behavior across all platforms
- Single source of truth for click detection
- Easier to maintain and test

**Cons:**
- Loses native platform double-click detection on Windows (but we can use platform timing)

**Option 2: Skip Processing of Pre-Synthesized Events**

Modify `MouseInterpreter.Process()` to skip processing if the driver already sent click events:

```csharp
public IEnumerable<MouseEventArgs> Process (MouseEventArgs e)
{
    yield return e;

    // Don't re-process if driver already sent a click/double-click event
    if (e.IsSingleDoubleOrTripleClicked)
    {
        yield break;
    }

    // For each mouse button
    for (var i = 0; i < 4; i++)
    {
        _buttonStates [i].UpdateState (e, out int? numClicks);
        // ...
    }
}
```

**Pros:**
- Preserves native platform double-click detection on Windows
- Less invasive change

**Cons:**
- Inconsistent behavior across platforms
- More complex logic
- Drivers and MouseInterpreter stay coupled

## Recommendation

**Option 1** is recommended because:
1. It provides consistent behavior across all platforms
2. It simplifies the codebase (single source of truth)
3. It's easier to test and maintain
4. The MouseInterpreter already has the infrastructure for accurate click detection (timing, position tracking)

### Future Consideration: Windows Driver ANSI Mouse Support

The Windows driver currently uses native Win32 API (`ReadConsoleInput()`) for mouse input regardless of the `IsLegacyConsole` setting, while Unix/Linux drivers use ANSI escape sequences. This architectural difference complicates cross-platform testing and maintenance.

**Potential Benefits of ANSI Mouse Input on Windows:**
- **Unified codebase**: Same mouse handling logic across all platforms
- **Simplified testing**: Mock ANSI sequences instead of Win32 `INPUT_RECORD` structs  
- **Consistency**: Already using ANSI for keyboard input when `IsLegacyConsole = false`
- **Standards-based**: VT100/xterm mouse protocol is well-documented

**Trade-offs:**
- ANSI parsing overhead (string/regex vs. binary structs) - likely negligible for typical mouse event rates (<100/sec)
- Loss of native Windows double-click timing - MouseInterpreter already handles this uniformly
- Requires Windows 10+ with VT support enabled (already a requirement for non-legacy mode)

This change would complement Option 1 by ensuring all drivers provide identical mouse event streams (Pressed/Released only), with MouseInterpreter as the sole authority for click synthesis. Worth investigating after resolving the current duplication issue.

### LINKED ISSUE #4484: Refactor `FakeDriver` to be a cross-platform "pure-ANSI" driver that also works great in tests
It just dawned on me that none of the drivers, other than `unix` use ANSI sequences for both keyboard input.

But, `unix` is platform dependent (on Unix). We need a pure-ANSI driver, especially for testing.

The idea is to refactor `FakeDriver` to be this driver:

- Rename it to "AnsiDriver"
- Change it's `TInputRecord` to be of type `char` instead of `ConsoleKeyInfo`
- Implement it as a full driver that uses pure ANSI for all input and output.
- Put in guards that let it run even if `Console.Write` etc... will fault if there's no real console and ensure it has a cache of what it's written, and supports `ITestableInput` (this is already mostly done in `dotnet`).

Related: 

- #4483

### LINKED ISSUE #4520: `View`'s `ScrollBar`s should be sized within `Padding`
Today, 

```cs
private void ConfigureVerticalScrollBar (ScrollBar scrollBar)
{
    scrollBar.X = Pos.AnchorEnd ();

    scrollBar.Height = Dim.Fill (
                                 Dim.Func (
                                           _ =>
                                           {
                                               if (_horizontalScrollBar.IsValueCreated)
                                               {
                                                   return _horizontalScrollBar.Value.Visible ? 1 : 0;
                                               }

                                               return 0;
                                           }));
}
```

Instead `.Height` should use the Padding's thickness:

```cs
    scrollBar.Height = Dim.Fill (
                                 Dim.Func (
                                           _ =>
                                           {
                                               int fill = 0;
                                               if (_horizontalScrollBar.IsValueCreated)
                                               {
                                                   fill =_horizontalScrollBar.Value.Visible ? 1 : 0;
                                               }

                                               return fill + Padding!.Thickness.Bottom;
                                           }));
```

To illustrate, here's a View with a Padding.Thickness.Bottom = 5 

Before:

Image

After:

Image

tig:
This PR goes deep in redesigning and re-implementing the mouse support in Terminal.Gui. It addresses a large number of deep design/legacy flaws. It is a major change that will:

- Make building View subclasses and apps that do simple mouse things extremely straightforward
- Make mouse actions such as double-click, drag, highlight, repeat easy to understand and deal with
- Make terminology consistent and sensible

One of the crux changes here is how we deal with multi-click (including repeat). I studied all of the top UI frameworks and how they deal with multi-click and have borrowed the idea of a click count from AppKit/Cocoa. We will still have the ability to have mouse commands bound to e.g. `MouseFlags.LeftButtonDoubleClick` but the underlying system uses `mouseEvent.ClickCount`  to track. This means that apps can have their own logic for dealing with multi-click with the low level `View.MouseEvent` or by looking in the `CommandContext`. 

## Fixes

- Fixes #4471
- Fixes #4474 
- Fixes #3714 
- Fixes #2588 
- Fixes #4484
- Fixes #4308 
- Fixes #4518 
- Fixes #4444 
- Fixes #4520
- Renamed `MouseEventArgs` to `Mouse` for consistency with `Key`
- Replaced button naming (`Button1*` → `LeftButton*`, `Button3*` → `RightButton*`)
- Renamed/refactored mouse state properties (`WantContinuousButtonPressed` → `MouseHoldRepeat`, `WantMousePositionReports` → `MousePositionTracking`, `HighlightStates` → `MouseHighlightStates`)
- Renamed test driver references (`fake` → `ANSI`, `TestDriver.Fake` → `DriverRegistry.Names.ANSI`)
- Updated fluent testing infrastructure (`GuiTestContext` → `TestContext`, `EnqueueKeyEvent` → `KeyDown`)
- Removed `KeyUp` event/handler infrastructure

# Problems with Old Design

**Problem:** Weak definition of models for each stage between Drivers, Application, and View
- Some v1 legacy still exists in how `IDriver/IInputProcessor` implementations work relative to ANSI (the preferred model that only WindowsDriver breaks).
- The role of Application (specifically `MouseImpl`) is unclear
- Who and what is responsible for what is unclear. E.g. who is reponsible for hold-repeat logic? Who tracks multi-clicks? 

**Problem:** Confusing logic to convert `Pressed` → `Clicked` in multiple places

**Current Behavior:**
- Driver emits `Pressed` and `Released`
- View converts `Pressed` → `Clicked` before binding lookup
- Multiple conversion points, easy to miss

**Problem:** Applications manually track double-click timing

**Problem:** Lexicon & Taxonomy is Inconsistent and Confusing
- `MouseEventArgs` is borrowed from `System.Console` but is different. It should be named something distinctive. I like the simplicity of `Key` and will rename `MouseEventArgs` to `Mouse`.
- `Button3Pressed` ect makes devs work too hard to remember which button is which. The names in the `MouseFlags` enums came from Windows. All modern frameworks use real names like `RightButtonPressed`. All these will be renamed.
- Names like `HighlightStates`, `WantContinousButtonPress`, and `WantMousePositionReports` are too long and don't really convey purpose (esp with new design). 
- Existing classes like `MouseButtonStateEx` are poorly and confusingly named.

### DIFF
--- a/Tests/IntegrationTests/FluentTests/NavigationTests.cs
+++ b/Tests/IntegrationTests/FluentTests/NavigationTests.cs
@@ -9,14 +9,9 @@ public class NavigationTests (ITestOutputHelper outputHelper) : TestsAllDrivers
 {
     private readonly TextWriter? _out = new TestOutputWriter (outputHelper);
 
-    [Theory]
-    [MemberData (nameof (GetAllDriverNames))]
-    public void Runnable_TabGroup_Forward_Backward (string d)
+    [Fact]
+    public void Runnable_TabGroup_Forward_Backward ()
     {
-        if (d == "windows")
-        {
-            return; // Something is causing this test to fail in the GH runner
-        }
         var v1 = new View { Id = "v1", CanFocus = true };
         var v2 = new View { Id = "v2", CanFocus = true };
         var v3 = new View { Id = "v3", CanFocus = true };
@@ -24,7 +19,7 @@ public class NavigationTests (ITestOutputHelper outputHelper) : TestsAllDrivers
         var v5 = new View { Id = "v5", CanFocus = true };
         var v6 = new View { Id = "v6", CanFocus = true };
 
-        using TestContext c = With.A<Window> (50, 20, d,  _out)
+        using TestContext c = With.A<Window> (50, 20, "ansi",  _out)
                                      .Then ((app) =>
                                             {
                                                 var w1 = new Window { Id = "w1" };


tig:
@bdisp / @tznind -

If you are going to work on new PRs I recommend using this branch. It involves major changes and merges will be challenging. 

There are still a lot of issues in mouse handling but it's very usable and (imho) is darn cool. 

Solves a ton of deep problems.

tig:
@BDisp, @tznind - This PR is now ready for review.

There are probably some regressions introduced but given I've expanded the test coverage substantially the only way to find them is through code review and usage. So I'd like to merge this into v2_develop asap.

Please pull it down locally and play!

Thanks!

BDisp (approved):

tig:
Fixed. 

Some Dialog functionality has issues but this is salable again.

tig:
One of the things I wanted to enable with 

- #4518 
 
was making it so Dialogs could be sized such that their content could be scrollable if sized below the natural size of the content. Done! 

Demoing here with Wizard:

WindowsTerminal_pdRcGSza6Z

PR #4486, opened 2025-12-12
Fixes #4466 - `FillRect` Corrupts Wide Characters When Overlapping

### LINKED ISSUE #4466: `FillRect` Corrupts Wide Characters When Overlapping
Original Issue:
Image

When `OutputBufferImpl.FillRect()` overwrites cells containing wide characters (e.g., Chinese text), it directly sets cell contents without properly handling wide character invalidation. This causes visual corruption when UI elements like `MessageBox` borders are drawn over wide character content.

## Reproduction
1. Display Chinese text (e.g., "你好呀") in a view
2. Draw a MessageBox that overlaps the text
3. Move mouse over the MessageBox border (triggers HighlightState changes and redraws)
4. Observe corrupted rendering where wide characters are partially overwritten

See the ChineseUI scenario in UICatalog for a concrete example.

## Root Cause
**Without Fix:**
- `FillRect()` directly assigns cell values without calling `AddStr()`
- Wide characters that get partially overwritten are not invalidated
- Adjacent cells are not marked dirty
- The second column of a wide character retains stale data

**In `OutputBufferImpl.FillRect()`:**
```csharp
#else
Contents [r, c] = new ()
{
    Grapheme = rune != default (Rune) ? rune.ToString () : " ",
    Attribute = CurrentAttribute, 
    IsDirty = true
};
#endif
```

tig:
## Fixes

- Fixes #4466

## Solution 

**In `OutputBufferImpl.FillRect()`:**

```cs
                    // We could call AddGrapheme here, but that would acquire the lock again.
                    // So we inline the logic instead.
                    SetAttributeAndDirty (c, r);
                    InvalidateOverlappedWideGlyph (c, r);
                    string grapheme = rune != default (Rune) ? rune.ToString () : " ";
                    WriteGraphemeByWidth (c, r, grapheme, grapheme.GetColumns (), clipBounds);
```

## Testing
Created comprehensive test suite in `OutputBufferWideCharTests.cs` that tests the issue at a low level. These tests fail if the fix above is not applied.
- `FillRect_OverwritesWideChar_InvalidatesProperly` - Core bug test
- `FillRect_OverwritesSecondColumnOfWideChar_InvalidatesWideChar` - Partial overwrite
- `ChineseUI_MessageBox_Over_WideChars` - Full scenario reproduction
- `FillRect_WideChar_AtBufferBoundary` - Edge case handling
- `OutputBase_Write_WideChar_MarksCellsDirty` - OutputBase fix validation
- `FillRect_WithSpaces_OverWideChars` - Clearing operations
- `FillRect_SingleWidthChars_WorksCorrectly` - Baseline behavior

### DIFF
--- a/Terminal.Gui/Drivers/OutputBase.cs
+++ b/Terminal.Gui/Drivers/OutputBase.cs
@@ -1,4 +1,3 @@
-//#define FIXES_4466
 using System.Collections.Concurrent;
 
 namespace Terminal.Gui.Drivers;


Terminal.Gui/Drivers/OutputBase.cs
```diff
@@ -131,7 +131,8 @@ public virtual void Write (IOutputBuffer buffer)
 
                     if (col != lastCol)
                     {
-                        // Was a wide grapheme and so mark clean next cell
+                        // Was a wide grapheme so mark clean next cell
+                        // See https://github.com/gui-cs/Terminal.Gui/issues/4466
                         buffer.Contents [row, col].IsDirty = false;
```
BDisp, line 135: I don't even know how you found that out. Anyway.

tig: You get the credit. I saw your changes in the other PR and used them to create dedicated tests.

BDisp: Remember that attribute issues can only be tested using DriverAssert specifically, otherwise they won't be visible to the naked eye.

BDisp (approved):

PR #4490, opened 2025-12-14
Fixes #4492, #4480 - Transparent shadows cause underlying wide glyph rendering issues

### LINKED ISSUE #4480: Shadow transparent doesn't support drawing wide glyphs

### LINKED ISSUE #4492: Transparent shadows cause underlying wide glyph rendering issues
devenv_flZe40Hnx6

May be fixed by #4490

Potentially related:

- #4491

tig:
## Fixes 

- Fixes #4492 
- Fixes #4480
- Adds extensive, low-level, unit tests documenting the fix and behavior
- Takes good stuff from #4465
  - *NOTE* the changes to `View.Draw` and `ShadowView` have nothing to do with fixing these issues; they are just good upgrades authored by @bdisp.
- Refactors `Margin.ShadowWidth/Height` to `Margin.ShadowSize` and removes unnecessary coupling with `View`

devenv_flZe40Hnx6

Potentially related:

- #4491

### DIFF
--- a/Terminal.Gui/ViewBase/Adornment/ShadowView.cs
+++ b/Terminal.Gui/ViewBase/Adornment/ShadowView.cs
@@ -155,7 +155,7 @@ internal class ShadowView : View
 
     // BUGBUG: This will never really work completely right by looking at an underlying cell and trying
     // BUGBUG: to do transparency by adjusting colors. Instead, it might be possible to use the A in argb for this.
-    // BUGBUG: See 
+    // BUGBUG: See https://github.com/gui-cs/Terminal.Gui/issues/4491
     private Attribute GetAttributeUnderLocation (Point location)
     {
         if (SuperView is not Adornment


tig:
devenv_flZe40Hnx6

BDisp:
That's the issue, because it's shading an area that doesn't belong to the margin. To achieve the same effect with the original logic, you have to exclude from the clearing the area of ​​the margin that isn't occupied by the ShadowView.

WindowsTerminal_soxn7AP9ly

tig:
This test reproduces the issue at the `Shadow` level. 

```cs
  [Fact]
  public void TransparentShadow_OverWide_Draws_Transparent_At_Driver_Output ()
  {
      // Arrange
      app.Init ("fake");
      app.Driver!.SetScreenSize (2, 3);
      app.Driver.Force16Colors = true;

      using Runnable superView = new ();
      superView.Width = Dim.Fill ();
      superView.Height = Dim.Fill ();
      superView.Text = "🍎🍎🍎🍎";
      superView.TextFormatter.WordWrap = true;
      superView.SetScheme (new (new Attribute (Color.Black, Color.White)));

      // Create view with transparent shadow
      View viewWithShadow = new ()
      {
          Width = Dim.Auto (),
          Height = Dim.Auto (),
          Text = "*",
          ShadowStyle = ShadowStyle.Transparent
      };
      // Make it so the margin is only on the bottom for simplicity
      viewWithShadow.Margin!.Thickness = new (0, 0, 0, 1);
      viewWithShadow.SetScheme (new (new Attribute (Color.Black, Color.White)));

      superView.Add (viewWithShadow);

      // Act
      app.Begin (superView);
      app.LayoutAndDraw ();
      app.Driver.Refresh ();

      // Assert
      _output.WriteLine ("Actual driver contents:");
      _output.WriteLine (app.Driver.ToString ());
      _output.WriteLine ("\nActual driver output:");
      string? output = app.Driver.GetOutput ().GetLastOutput ();
      _output.WriteLine (output);

      DriverAssert.AssertDriverOutputIs ("""
                                         \x1b[30m\x1b[107m*\x1b[90m\x1b[103m \x1b[97m\x1b[40m \x1b[90m\x1b[100m \x1b[97m\x1b[40m🍎
                                         """, _output, app.Driver);
  }
```

This `OutputBufferImpl.cs` fix causes the test to pass:

```cs
            if (printableGraphemeWidth > 1)
            {
                // Skip the second column of a wide character
                // See issue: https://github.com/gui-cs/Terminal.Gui/issues/4492
                if (Clip.Contains (Col, Row))
                {
                    // IMPORTANT: We do NOT modify column N+1's IsDirty or Attribute here.
                    // See: https://github.com/gui-cs/Terminal.Gui/issues/4258
                    Contents [Row, Col].Attribute = CurrentAttribute;
                }

                // Advance cursor again for wide character
                Col++;
            }
```

Visually this can be reproduced in `WideGlyphs` like this (as well in `Chinese UI` per above):

Before fix:
image

After fix:
image

I want to create an even lower level version of this test before moving on ....

tig:
Here it is :

```cs
    [Fact]
    public void AddStr_WideGlyph_Second_Column_Attribute_Outputs_Correctly ()
    {
        // This test verifies the fix for issue #4258 by checking the actual driver output
        // This mimics what happens when TransparentShadow redraws a wide glyph from ScreenContents
        // WITHOUT line 194, column N+1's attribute doesn't get set, causing wrong colors in output
        // See: OutputBufferImpl.cs line ~196 (Contents [Row, Col].Attribute = CurrentAttribute;)
        using IDriver driver = CreateFakeDriver ();
        driver.SetScreenSize (3, 1);
        driver.Force16Colors = true;

        // Step 1: Draw initial content - a wide glyph at column 1 with white-on-black
        driver.CurrentAttribute = new Attribute (Color.White, Color.Black);
        driver.Move (1, 0);
        driver.AddStr ("🍎X");  // Wide glyph at columns 1-2, 'X' at column 3 doesn't exist (off screen)

        // At this point:
        // - Column 0: space (default) with white-on-black  
        // - Column 1: 🍎 with white-on-black
        // - Column 2: (part of 🍎) with white-on-black (from initial ClearContents)

        // Step 2: Now redraw the SAME wide glyph at column 1 but with a DIFFERENT attribute (red-on-yellow)
        // This simulates what transparent shadow does - it redraws what's underneath with a dimmed attribute
        driver.CurrentAttribute = new Attribute (Color.BrightRed, Color.BrightYellow);
        driver.Move (1, 0);
        driver.AddStr ("🍎");

        // Verify internal state
        Assert.Equal ("🍎", driver.Contents! [0, 1].Grapheme);
        Assert.Equal (new Attribute (Color.BrightRed, Color.BrightYellow), driver.Contents [0, 1].Attribute);

        // THIS is the critical assertion - column 2's attribute MUST be red-on-yellow
        // WITHOUT line 194: column 2 retains white-on-black
        // WITH line 194: column 2 gets red-on-yellow
        Assert.Equal (new Attribute (Color.BrightRed, Color.BrightYellow), driver.Contents [0, 2].Attribute);

        driver.Refresh ();

        // Expected output:
        // Column 0: space with white-on-black
        // Columns 1-2: 🍎 with red-on-yellow (both columns must have same attribute!)
        //
        // WITHOUT line 196, the output would be:
        // \x1b[97m\x1b[40m  (white-on-black for column 0)
        // \x1b[91m\x1b[103m🍎 (red-on-yellow starts at column 1)
        // \x1b[97m\x1b[40m (WRONG! Attribute changes mid-glyph because column 2 still has white-on-black)
        //
        // WITH line 196, the output is:
        // \x1b[97m\x1b[40m  (white-on-black for column 0)
        // \x1b[91m\x1b[103m🍎 (red-on-yellow for both columns 1 and 2)
        DriverAssert.AssertDriverOutputIs (
            "\x1b[97m\x1b[40m \x1b[91m\x1b[103m🍎",
            output,
            driver);
    }
```

tig:
At the end of the day, `ShadowView` is pretty hacky. In fact, it will never really work completely right by looking at an underlying cell and trying to do transparency by ajusting colors. Instead it might be possible to use the `A` in `argb` for this. But that's for another day (I'll open an issue).

BDisp:
I have a question that I'm somewhat unsure about. When we explicitly define the margin thickness, does it then have to be increased by the ShadowWidth and ShadowHeight if the ShadowStyle is different from ShadowStyle.None, or do we have to consider it as the total thickness, including the ShadowWidth and ShadowHeight?

Terminal.Gui/Drivers/OutputBufferImpl.cs
```diff
@@ -185,8 +185,17 @@ private void AddGrapheme (string grapheme)
...
             {
                 // Skip the second column of a wide character
-                // IMPORTANT: We do NOT modify column N+1's IsDirty or Attribute here.
-                // See: https://github.com/gui-cs/Terminal.Gui/issues/4258
+                // See issue: https://github.com/gui-cs/Terminal.Gui/issues/4492
+                // Test: AddStr_WideGlyph_Second_Column_Attribute_Outputs_Correctly
+                // Test: AddStr_WideGlyph_Second_Column_Attribute_Set_When_In_Clip
+                if (Clip.Contains (Col, Row))
+                {
+                    // IMPORTANT: We do NOT modify column N+1's IsDirty or Attribute here.
```
BDisp, line 193: Attribute should be set, to be more explicitly as I did in #4465.

tig:
> I have a question that I'm somewhat unsure about. When we explicitly define the margin thickness, does it then have to be increased by the ShadowWidth and ShadowHeight if the ShadowStyle is different from ShadowStyle.None, or do we have to consider it as the total thickness, including the ShadowWidth and ShadowHeight?

Similar to View.LineStyle, if someone else changes Margin.Thickness and they have ShadowStyle set, they it's their bad for not realizing they are fighting with themselves.  Having a settable `ShadowSize` is a cool feature and was useful in debugging, but it will be very rarely used and thus no need to spend more time on it.

Terminal.Gui/Drivers/OutputBufferImpl.cs
tig: Not sure what you mean. This code looks identical to what you currently have there.

Image

BDisp:
> Similar to View.LineStyle, if someone else changes Margin.Thickness and they have ShadowStyle set, they it's their bad for not realizing they are fighting with themselves. Having a settable `ShadowSize` is a cool feature and was useful in debugging, but it will be very rarely used and thus no need to spend more time on it.

I don't entirely agree because this is controllable. When the thickness is being defined, the developer will have to take the ShadowSize into account, and thus we will only have to accommodate the ShadowView within the defined thickness. It's only necessary to comply with the rule controlled by the API.

tig:
> > Similar to View.LineStyle, if someone else changes Margin.Thickness and they have ShadowStyle set, they it's their bad for not realizing they are fighting with themselves. Having a settable `ShadowSize` is a cool feature and was useful in debugging, but it will be very rarely used and thus no need to spend more time on it.
> 
> I don't entirely agree because this is controllable. When the thickness is being defined, the developer will have to take the ShadowSize into account, and thus we will only have to accommodate the ShadowView within the defined thickness. It's only necessary to comply with the rule controlled by the API.

You are not wrong. But it's a very minor issue and I don't want to add more complexity for such a very rare use case. 

Any other objects to me merging this?

BDisp:
I don't even need to comment on anything. If this is teamwork, then I don't understand anything. You know I was working on this in another PR, you didn't speak up, you just copied it, and now it's not even worth commenting on, because it would be a waste of time and a lot of patience to put up with all this.

tig:
> you didn't speak up

I commented regularly as I proceeded. I created indivisal Issues for each of the core issues. Eg

image

> you just copied it

I gave you credit very visibly and thanked you multiple times for not only providing good fixes, but for making improvements to View.Draw.

Low level unit tests that clearly demonstrate the issue at its essence were required. You continued to believe the problem was at the View level. You were 100% certain and unwilling to be open. E.g:

image

I had a strong feeling (but couldn't prove it) from the start it was at the IOutputBuffer/IOutput level. Instead of continuing to argue with you using words, I broke the problem down further and focused on proving I was correct (or incorrect) from the bottom up. Code talks.

Sorry you are frustrated. But, in the end, this WAS a collaboration, and in the end, the code base is in far better shape now than it was before.

BDisp:
> This `OutputBufferImpl.cs` fix causes the test to pass:
> 
> ```cs
>             if (printableGraphemeWidth > 1)
>             {
>                 // Skip the second column of a wide character
>                 // See issue: https://github.com/gui-cs/Terminal.Gui/issues/4492
>                 if (Clip.Contains (Col, Row))
>                 {
>                     // IMPORTANT: We do NOT modify column N+1's IsDirty or Attribute here.
>                     // See: https://github.com/gui-cs/Terminal.Gui/issues/4258
>                     Contents [Row, Col].Attribute = CurrentAttribute;
>                 }
> 
>                 // Advance cursor again for wide character
>                 Col++;
>             }
> ```

I thought you had already seen this correction in my PR #4465. After all, you didn't copy it correctly.

https://github.com/BDisp/Terminal.Gui/blob/942f0ce3cb2fc7039d8134857bf0b20c078d9ff5/Terminal.Gui/Drivers/OutputBufferImpl.cs#L186-L200

tig:
> > This `OutputBufferImpl.cs` fix causes the test to pass:
> > ```cs
> >             if (printableGraphemeWidth > 1)
> >             {
> >                 // Skip the second column of a wide character
> >                 // See issue: https://github.com/gui-cs/Terminal.Gui/issues/4492
> >                 if (Clip.Contains (Col, Row))
> >                 {
> >                     // IMPORTANT: We do NOT modify column N+1's IsDirty or Attribute here.
> >                     // See: https://github.com/gui-cs/Terminal.Gui/issues/4258
> >                     Contents [Row, Col].Attribute = CurrentAttribute;
> >                 }
> > 
> >                 // Advance cursor again for wide character
> >                 Col++;
> >             }
> > ```
> 
> I thought you had already seen this correction in my PR #4465. After all, you didn't copy it correctly.
> 
> https://github.com/BDisp/Terminal.Gui/blob/942f0ce3cb2fc7039d8134857bf0b20c078d9ff5/Terminal.Gui/Drivers/OutputBufferImpl.cs#L186-L200

Sorry if I'm being dense, but I don't see any difference in the code. What is in this branch:
image

What is in the branch you are pointing to:
image

BDisp:
What I meant was that I had already made this change to my PR a few days ago, and only today did you bring it up, as if it were something new.

tig:
> What I meant was that I had already made this change to my PR a few days ago, and only today did you bring it up, as if it were something new.

Sorry you didn't read all my messages. As soon as I saw your fix and tried it and then I posted this:

https://github.com/gui-cs/Terminal.Gui/pull/4465#issuecomment-3648474932

In it I very clearly state that I appreciated you finding the actual fix. 

I'm not evil and doing bad things. I'm trying really hard to be productive and to communicate well. You didn't read what I wrote.

BDisp:
Yes, I read that a few days ago. I know you're not bad and you try your best. Maybe I'm having an off day, sorry.

PR #4494, opened 2025-12-14
Fixes some bugs in the `Scenario/EditorsAndHelpers` and backports `WideGlyphs`

tig:
I found I was fighting with these bugs, moving between branches. But merging these into v2_develop this will make stuff easier.

### DIFF
--- a/Examples/UICatalog/Scenarios/EditorsAndHelpers/EditorBase.cs
+++ b/Examples/UICatalog/Scenarios/EditorsAndHelpers/EditorBase.cs
@@ -163,6 +163,7 @@ public abstract class EditorBase : View
         if (disposing && App is {})
         {
             App.Navigation!.FocusedChanged -= NavigationOnFocusedChanged;
+            App.Mouse.MouseEvent -= ApplicationOnMouseEvent;
         }
 
         base.Dispose (disposing);


Examples/UICatalog/Scenarios/EditorsAndHelpers/AdornmentEditor.cs
```diff
@@ -52,16 +50,19 @@ public Adornment? AdornmentToEdit
 
             if (IsInitialized)
             {
+                _topEdit!.Value = _adornment.Thickness.Top;
                 _topEdit!.Value = _adornment.Thickness.Top;
```
BDisp: These two lines are repeated.

Examples/UICatalog/Scenarios/ShadowStyles.cs
```diff
@@ -96,4 +113,10 @@ public override void Main ()
         Application.Shutdown ();
 
     }
+
+    private void ButtonOnAccepting (object sender, CommandEventArgs e)
+    {
+        MessageBox.Query((sender as View).App, "Hello", "ok");
```
BDisp: The "ok" here refers to the message, not the button as it should.

PR #4498, opened 2025-12-15
Fixes #4497  - makes replacement char conifgurable

### LINKED ISSUE #4497: Add `Glyphs.ReplacementChar` to enable config
```cs
    /// <summary>Unicode replacement character; overrides `Rune.ReplacementChar`. Defaults to ' ' (Space).</summary>
    [ConfigurationProperty (Scope = typeof (ThemeScope))]
    public static Rune ReplacementChar { get; set; } = (Rune)' ';
```

tig:
## Fixes

- Fixes #4497

## Proposed Changes/Todos

- [x] Add property
- [x] Update usages
- [x] FIgure out how to enable parallel tests to override config without actually setting `Glyphs.ReplacementChar`, which is a static.

### DIFF
--- a/Examples/UICatalog/Resources/config.json
+++ b/Examples/UICatalog/Resources/config.json
@@ -9,6 +9,7 @@
   "Themes": [
     {
       "Hot Dog Stand": {
+        "Glyphs.WideGlyphReplacement": "①",
         "Schemes": [
           {
             "Runnable": {
@@ -134,8 +135,7 @@
       }
     },
     {
-      "UI Catalog Theme": {
-        "Glyphs.ReplacementChar": " ",
+      "UI Catalog": {
         "Window.DefaultShadow": "Transparent",
         "Button.DefaultShadow": "None",
         "CheckBox.DefaultHighlightStates": "In, Pressed, PressedOutside",
--- a/Examples/UICatalog/Scenarios/WideGlyphs.cs
+++ b/Examples/UICatalog/Scenarios/WideGlyphs.cs
@@ -34,10 +34,6 @@ public sealed class WideGlyphs : Scenario
             AutoSelectAdornments = false,
             ShowViewIdentifier = true
         };
-        adornmentsEditor.ExpanderButton!.Accepting += (sender, args) =>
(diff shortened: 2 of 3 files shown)


tig:
This change is it points out that no existing test covers this code in `OutputBufferImpl`:

image

Tests/UnitTestsParallelizable/ViewBase/Draw/ViewDrawingClippingTests.cs
```diff
@@ -738,19 +740,21 @@ public void Draw_WithBorderSubView_At_Col3_In_WideGlyph_DrawsCorrectly ()
             Height = 3
         };
 
+        driver.GetOutputBuffer ().SetReplacementChars ((Rune)'①', (Rune)'②');
```
BDisp: I understand your intention, but you will never be able to visualize glyph ② in an output. At least you are now certain of that. 😃 
In a real terminal, it's best not to use either of these two glyphs, as either one could appear at the output.

Terminal.Gui/Drivers/OutputBufferImpl.cs
```diff
@@ -291,6 +291,8 @@ private void WriteWideGrapheme (int col, int row, string grapheme)
             // Mark second column as replacement to indicate partial overlap
             if (col + 1 < Cols)
             {
+                // NOTE: This is dead code and apparently never called.
+                // NOTE: See AddStr_WideChar_FirstColumnOutsideClip_SecondColumnInside_CurrentBehavior
```
BDisp: Yes, you're right, because that piece of code will never be executed since it's already been cleared by ClearViewport and will only contain a space.

Tests/UnitTestsParallelizable/ViewBase/Draw/ViewDrawingClippingTests.cs
tig: Yep. For tests only.

tig:
Based on that, I've done some renaming and simplification here.

PR #4499, opened 2025-12-16
Fixes #4489 - Requires non-null `IApplication` in MessageBox API

### LINKED ISSUE #4489: null option on MessageBox.Query causes application silent exit
## Describe the bug

When using the new Application.Create pattern then messagebox throws when passed null.

This manifests as instantly exiting the application with no exception exposed in IDE.

## To Reproduce

```csharp
using Terminal.Gui.App;
using Terminal.Gui.Views;

using (var app = Application.Create().Init())
{
        MessageBox.Query(null,"Hello", "You clicked me!", "OK");
}
```

The solution for end user is to always provide app.  But its optional property from its API and docs.  The biggest problem is that theres no indication that this is what caused app to exit i.e. no Exception is raised or it is silently recaught elsewhere in terminal gui.

```diff
using Terminal.Gui.App;
using Terminal.Gui.Views;

using (var app = Application.Create().Init())
{
-        MessageBox.Query(null,"Hello", "You clicked me!", "OK");
+        MessageBox.Query(app,"Hello", "You clicked me!", "OK");
}
```

tig:
All MessageBox.Query and ErrorQuery overloads now require a non-null IApplication parameter instead of a nullable one. XML docs and references are updated to match. Minor code cleanups include improved button list initialization, simplified null checks, clarified variable names, and dialog construction refactoring. Core logic remains unchanged.

## Fixes

- Fixes #4489


tig:
@tznind you good with this?

PR #4583, opened 2026-01-19
Fixes #4582. Ansi driver isn't encoding correctly characters with accent

### LINKED ISSUE #4582: Ansi driver isn't encoding correctly characters with accent
> I found another issue that only happens with the ansi driver. Type `á` or any letter with accent in a TextView or TextField and the ReadFile function will get �, which is what is outputted. At least with me is what I get out.
> This only happens on Windows. 

 _Originally posted by @BDisp in [#4573](https://github.com/gui-cs/Terminal.Gui/issues/4573#issuecomment-3765845817)_

BDisp:
## Fixes

- Fixes #4582

## Proposed Changes/Todos

- [x] Using GetConsoleCP API to encode to the current code page.

## Pull Request checklist:

### DIFF
--- a/Tests/UnitTestsParallelizable/Input/InputInjectorTests.cs
+++ b/Tests/UnitTestsParallelizable/Input/InputInjectorTests.cs
@@ -121,6 +121,152 @@ public class InputInjectorTests (ITestOutputHelper output)
 
     #endregion
 
+    #region InjectKey Tests - Pipeline Mode
+
+    [Fact]
+    public async Task InjectKey_PipelineMode_MultipleKeys_RaisesAllEvents ()
+    {
+        // Arrange
+        VirtualTimeProvider timeProvider = new ();
+        timeProvider.SetTime (new DateTime (2025, 1, 1, 12, 0, 0));
+
+        using IApplication app = Application.Create (timeProvider);
+        app.Init (DriverRegistry.Names.ANSI);
+
+        List<Key> receivedKeys = [];
+        app.Keyboard.KeyDown += (_, key) => receivedKeys.Add (key);
+
+        // Explicit Pipeline mode
+        TestInputSource testSource = new (timeProvider);
+        InputInjector injector = new (app.Driver?.GetInputProcessor ()!, timeProvider, testSource);
+
+        InputInjectionOptions options = new ()
+        {
+            Mode = InputInjectionMode.Pipeline,
+            AutoProcess = true,
+            TimeProvider = timeProvider
+        };
+
+        // Act
+        injector.InjectKey (Key.A, options);
+        injector.InjectKey (Key.B, options);
+        injector.InjectKey (Key.C, options);
+
+        // BUGBUG: This is a hack; we need to figure out how to enable this without delay
+        await Task.Delay (50); // Allow some time for processing
+        injector.ProcessQueue ();
+
+        // Assert - Should raise exactly 3 KeyDown events
+        Assert.Equal (3, receivedKeys.Count);
+        Assert.Equal (Key.A, receivedKeys [0]);
+        Assert.Equal (Key.B, receivedKeys [1]);
+        Assert.Equal (Key.C, receivedKeys [2]);
+    }
+
+    [Fact]
+    public async Task InjectKey_Pipeline_AccentedKeys_RaisesAllEvents ()
+    {
+        // Arrange
+        VirtualTimeProvider timeProvider = new ();
+        timeProvider.SetTime (new DateTime (2025, 1, 1, 12, 0, 0));
+
+        using IApplication app = Application.Create (timeProvider);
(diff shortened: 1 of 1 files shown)


BDisp:
> It's be really awesome if you tried using https://gui-cs.github.io/Terminal.Gui/docs/input-injection.html to build a test that tests all drivers for accented input.
> 
> It should totally be possible now that we have this.
> 
> ```cs
> // Explicit Pipeline mode
> IInputInjector injector = app.GetInputInjector();
> InputInjectionOptions options = new() { Mode = InputInjectionMode.Pipeline };
> 
> // This tests: Key.F1 → "\x1b[OP" → Parser → Key.F1
> injector.InjectKey(Key.F1, options);
> ```

I would like to create a test for all accented inputs when you can get the following simple test in pipeline mode to pass without failure, or indicate another way to test pipeline mode. If I debug each input, the KeyDown event is possibly invoked and the test passes, but if I just run the test, it always fails. It's possible this is related to the ANSI driver failing to process some responses to ANSI requests, I'm not sure.

```cs
    [Fact]
    public void InjectKey_PipelineMode_MultipleKeys_RaisesAllEvents ()
    {
        // Arrange
        app.Init (DriverRegistry.Names.ANSI);

        List<Key> receivedKeys = [];
        app.Keyboard.KeyDown += (_, key) => receivedKeys.Add (key);

        // Explicit Pipeline mode
        IInputInjector injector = app.GetInputInjector ();
        InputInjectionOptions options = new () { Mode = InputInjectionMode.Pipeline };

        // Act
        injector.InjectKey (Key.A, options);
        injector.InjectKey (Key.B, options);
        injector.InjectKey (Key.C, options);

        // Assert - Should raise exactly 3 KeyDown events
        Assert.Equal (3, receivedKeys.Count);
        Assert.Equal (Key.A, receivedKeys [0]);
        Assert.Equal (Key.B, receivedKeys [1]);
        Assert.Equal (Key.C, receivedKeys [2]);
    }
```

tig:
This should work but...

- The `Thread.Delay` is a hack that should not be necessary
- I forgot that `Pipeline` goes ALMOST to the top of the pipline, but not all the way. It injects the input at the `IInput.Peek/Read` layer with a 'test queue'. This means that it DOES NOT simulate what happens when a driver actually tries to read from the OS input stream. Thus the test below always succeeds, which is bogus.

```cs
[Fact]
public void InjectKey_PipelineMode_Accented_Char ()
{
    // Arrange
    VirtualTimeProvider timeProvider = new ();
    timeProvider.SetTime (new DateTime (2025, 1, 1, 12, 0, 0));

    using IApplication app = Application.Create (timeProvider);
    app.Init (DriverRegistry.Names.ANSI);

    List<Key> receivedKeys = [];
    app.Keyboard.KeyDown += (_, key) => receivedKeys.Add (key);

    // Explicit Pipeline mode
    TestInputSource testSource = new (timeProvider);
    InputInjector injector = new (app.Driver?.GetInputProcessor ()!, timeProvider, testSource);

    InputInjectionOptions options = new ()
    {
        Mode = InputInjectionMode.Pipeline,
        AutoProcess = false,
        TimeProvider = timeProvider
    };

    // Act
    injector.InjectKey ('á', options);

    Thread.Sleep (100);
    injector.ProcessQueue ();

    Assert.Equal ('á', receivedKeys [0]);
}
```

I'll keep working on making Pileline mode go even deeper as a separate issue because I really want us to be able to test the whole pileline. My idea is to allow an `IInput` implementation to get passed a file handle that will be used instead of `GetStdHandle (STD_INPUT_HANDLE);`.

In the meantime, can you test out #4586 and verify that it does indeed fix the accented char issue? I tried to integrate the fix you built in #4583, but I'm not 100% sure I did it right.

tig:
@BDisp I see you are still working on trying to get Inject to work. Did you see this:

> I forgot that Pipeline goes ALMOST to the top of the pipline, but not all the way. It injects the input at the IInput.Peek/Read layer with a 'test queue'. This means that it DOES NOT simulate what happens when a driver actually tries to read from the OS input stream. Thus the test below always succeeds, which is bogus.

BDisp:
> @BDisp I see you are still working on trying to get Inject to work. Did you see this:
> 
> > I forgot that Pipeline goes ALMOST to the top of the pipline, but not all the way. It injects the input at the IInput.Peek/Read layer with a 'test queue'. This means that it DOES NOT simulate what happens when a driver actually tries to read from the OS input stream. Thus the test below always succeeds, which is bogus.

Yes, I saw it and responded here https://github.com/gui-cs/Terminal.Gui/pull/4586#issuecomment-3770197642. Therefore, only by using a real terminal with the ansi driver would the problem with accented characters be verified.

BDisp:
I I think I've finished this, and I don't think the macOS issue is related to any problem with this PR.

PR #4586, opened 2026-01-19
Fixes multiple ANSI driver issues

### LINKED ISSUE #4584: Run cmd.exe with -d ansi we only see escape sequences output
Image

### LINKED ISSUE #4587: `InputProcessor` is reporting `ignored unrecognized response` when it shouldn't
We see these log messages when using the mouse to run a new `IRunnable` (e.g. a `Dialog`).

```cs
[12:59:08 VRB] [InputProcessorImpl] [RaiseMouseEventParsed] 08.436:{X=54,Y=21}:LeftButtonPressed::
[12:59:08 VRB] [InputProcessorImpl] [RaiseSyntheticMouseEvent] 08.436:{X=54,Y=21}:LeftButtonPressed::
[12:59:08 VRB] [InputProcessorImpl] [RaiseSyntheticMouseEvent] 08.436:{X=54,Y=21}:LeftButtonPressed::
[12:59:08 VRB] [InputProcessorImpl] [RaiseMouseEventParsed] 08.577:{X=54,Y=21}:LeftButtonReleased::
[12:59:08 VRB] [InputProcessorImpl] [RaiseSyntheticMouseEvent] 08.577:{X=54,Y=21}:LeftButtonReleased::
[12:59:08 VRB] [InputProcessorImpl] [RaiseSyntheticMouseEvent] 08.577:{X=54,Y=21}:LeftButtonReleased::
[12:59:08 VRB] [InputProcessorImpl] [RaiseSyntheticMouseEvent] 08.577:{X=54,Y=21}:LeftButtonClicked::
[12:59:08 VRB] [InputProcessorImpl] [.ctor] InputProcessorImpl swallowed incomplete mouse sequence: '[<0;55;22m'
[12:59:08 VRB] [AnsiResponseParserBase] [TryLastMinuteSequences] AnsiResponseParser last minute swallowed '[<0;55;22m'

```

All drivers that use ANSI/VT mouse sequences (`ESC[<Cb;Cx;Cy;M/m`) will have this problem because:

1. Mouse press (`M`) happens in parent `Runnable`
2. Mouse release (`M`) happens in parent `Runnable` 
3. A synthetic `click` event is raised by `InputProcessor`
4. Some code responds to the click and does a `Run` on a new runnable.
5. The new Runnable starts
6. A mouse release (`m`) arrives in new Runnable context
7. `InputProcessor` reports no matching press = orphaned release event = logged as "unrecognized"
 
It is unclear why this is happening.

It seems benign, but is an indication that something is not right.

tig:
## Fixes

- Fixes #4584 
- Fixes #4583
- Fixes #4587
 
- I can no longer reproduce the "blank screen after exiting a Scenario" in cmd.exe
- Refactors AnsiInput/AnsiOutput to more readable and easier to debug

### DIFF
--- a/Terminal.Gui/Drivers/WindowsDriver/WindowsVTOutputHelper.cs
+++ b/Terminal.Gui/Drivers/WindowsDriver/WindowsVTOutputHelper.cs
@@ -116,7 +116,7 @@ internal sealed class WindowsVTOutputHelper : IDisposable
             }
 
             IsEnabled = true;
-            Logging.Information ($"Windows VTS output mode enabled successfully. Mode: 0x{newMode:X} (was 0x{_originalConsoleMode:X})");
+            //Logging.Information ($"Windows VTS output mode enabled successfully. Mode: 0x{newMode:X} (was 0x{_originalConsoleMode:X})");
 
             return true;
         }
@@ -142,7 +142,7 @@ internal sealed class WindowsVTOutputHelper : IDisposable
         {
             SetConsoleMode (OutputHandle, _originalConsoleMode);
             IsEnabled = false;
-            Logging.Information ("Windows console mode restored.");
+            //Logging.Information ("Windows console mode restored.");
         }
         catch (Exception ex)
         {
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiInput.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiInput.cs
@@ -162,7 +162,7 @@ public class AnsiInput : InputImpl<char>, ITestableInput<char>
             // NOTE: Output operations (alternate buffer, cursor visibility, mouse events)
             // NOTE: are handled by ANSIOutput, not here. ANSIInput only handles input.
 
-            Logging.Information ("ANSIInput initialized successfully");
+            //Logging.Information ("ANSIInput initialized successfully");
         }
         catch (Exception ex)
         {
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
@@ -104,13 +104,14 @@ public class AnsiOutput : OutputBase, IOutput
             Write (EscSeqUtils.CSI_ClearScreen (EscSeqUtils.ClearScreenOptions.EntireScreen));
             Write (EscSeqUtils.CSI_SetCursorPosition (1, 1)); // Move to top-left
             Write (EscSeqUtils.CSI_HideCursor);
+            // TODO: Move Input related CSI sequences to AnsiInput
             Write (EscSeqUtils.CSI_EnableMouseEvents);
 
             // Flush to ensure all sequences are sent
             // NOTE: Default implementation of Flush does nothing.
             Console.Out.Flush ();
 
-            Logging.Information ("ANSIOutput initialized successfully");
+            //Logging.Information ("ANSIOutput initialized successfully");
 
             // Note: Size will be queried via ANSI by ANSISizeMonitor.Initialize()
             // Don't use Console.WindowWidth/Height here as it may reflect the main buffer,
@@ -329,6 +330,7 @@ public class AnsiOutput : OutputBase, IOutput
         try
         {
             // Restore terminal state: disable mouse, restore buffer, show cursor
+            // TODO: Move Input related CSI sequences to AnsiInput
             Write (EscSeqUtils.CSI_DisableMouseEvents);
             Write (EscSeqUtils.CSI_RestoreCursorAndRestoreAltBufferWithBackscroll);
             Write (EscSeqUtils.CSI_ShowCursor);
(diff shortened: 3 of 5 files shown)


BDisp:
With cmd.exe, it continues to display ANSI sequences that remain visible even after exiting the app.

image

tig:
> With cmd.exe, it continues to display ANSI sequences that remain visible even after exiting the app.
> 
> image

I did 

1) `Win-R` `cmd.exe`
2) `dotnet run -p C:/Users/Tig/s/gui-cs/Terminal.Gui/Examples/UICatalog -d ansi`

and got this:

cmd_jlcNDC2MSc

What steps are you following?

tig:
Ah, if I do this I see ...


```
C:\Users\Tig>s\gui-cs\Terminal.Gui\Examples\UICatalog\bin\Debug\net10.0\UICatalog.exe -d ansi
[?1003l[?1015l[?1006l[?1049l[?25h[?1003l[?1015l[?1006l[?1049l[?25h[?1003l[?1015l[?1006l[?1049l[?25h
C:\Users\Tig>
```

What's the difference between running `dotnet run` and the `exe` directly?

BDisp:
> Ah, if I do this I see ...
> 
> ```
> C:\Users\Tig>s\gui-cs\Terminal.Gui\Examples\UICatalog\bin\Debug\net10.0\UICatalog.exe -d ansi
> �[?1003l�[?1015l�[?1006l�[?1049l�[?25h�[?1003l�[?1015l�[?1006l�[?1049l�[?25h�[?1003l�[?1015l�[?1006l�[?1049l�[?25h
> C:\Users\Tig>
> ```
> 
> What's the difference between running `dotnet run` and the `exe` directly?

I have no idea why they behave differently, but normally when I test in cmd.exe I always use the executable (.exe) which is what should be considered because we are practically testing the final product. With my fix this doesn't happen, but you have to compare to find out the difference.

BDisp:
But I can give you a hint. Notice that the mouse commands relate to the driver disposal, as they contain the letter 'l' at the end. In this case, you need to correct it in the Dispose file or something similar.

tig:
> 1049l

Your fix moves the emitting of the cleanup esc sequences to the Input.Dispose from Output.Dispose. This seems to fix it, but actually highlights a deeper issue:

When an `IApplication` instance is disposed, the order in which `IOutput` and `IInput` are disposed is timing dependent. 

The fix for this will require some refactoring of the driver lifecyle to ensure the order is not timing related. 

So, what I recommend for now is we leave this branch as-is, and open a new issue for what I describe above, with the fix-criteria being that the cleanup esc sequences are never visible (because we will ensure output will get disposed last).

Terminal.Gui/Drivers/WindowsDriver/WindowsVTOutputHelper.cs
```diff
@@ -0,0 +1,170 @@
...
+
+    private uint _originalConsoleMode;
+    private bool _disposed;
+
+    /// <summary>
+    ///     Gets whether VTS output mode was successfully enabled.
+    /// </summary>
+    public bool IsEnabled
+    {
+        get => field;
```
BDisp: It's only enough `get;` without the field.

BDisp:
> > 1049l
> 
> Your fix moves the emitting of the cleanup esc sequences to the Input.Dispose from Output.Dispose. This seems to fix it, but actually highlights a deeper issue:
> 
> When an `IApplication` instance is disposed, the order in which `IOutput` and `IInput` are disposed is timing dependent.
> 
> The fix for this will require some refactoring of the driver lifecyle to ensure the order is not timing related.
> 
> So, what I recommend for now is we leave this branch as-is, and open a new issue for what I describe above, with the fix-criteria being that the cleanup esc sequences are never visible (because we will ensure output will get disposed last).

I found the bug. You don't need to move to the AnsiInput.Dispose but only remove the line `_output.Dispose ();` from the `Dispose` method of the `DriverImpl` class, which is causing calling twice the AnsiOutput.Dispose method. The first call really does what it's expected but the second call cause the ANSI sequences being outputted to the console.

tig:
I verified the accented char bug is fixed

image

BDisp:
> I verified the accented char bug is fixed
> 
> image

Just to clarify, the unit test for accented characters always passes in any driver. The problem with the ANSI driver was that ReadFile uses code page settings and therefore did not recognize accented characters as valid, replacing them with the replacement character.

tig:
> > I verified the accented char bug is fixed
> > image
> 
> Just to clarify, the unit test for accented characters always passes in any driver. The problem with the ANSI driver was that ReadFile uses code page settings and therefore did not recognize accented characters as valid, replacing them with the replacement character.

Yep. I think we're on the same (code) page ;-).

I'm going to merge this now.

BDisp:
The problem is that you're using ReadFile, but it has to be that way. If you were using ReadConsoleInput, this encoding problem wouldn't occur. But it's been resolved anyway.

Issue #3822, opened 2024-11-18
`TableView` - Upgrade to use Content Area for scrolling


tig:
Once #3820 is done, which fixes

- #2489 

we will want to upgrade `TableView` to support `ScrollBar`. There are two approaches:

1) Upgrade `TableView` to utilize the virtual content scrolling now built into `View` via `Viewport`, `ViewportSettings`, and `SetContentSize()`. 
2) Leave the existing scroll logic in place, and program `View.HorizontalScrollBar/VerticalScrollBar` manually.

The former is a far better solution and should not be too hard to accomplish.

tznind:
TableView horizontal scrolling is by table column not characters.  Does the new ScrollBar support 'virtual' viewport units which do not have to match directly to runes?

tig:
Or, just do 2).

However, I'm quite confident if you do 1) you'll end up removing a TON of code from the class...

tig:
Oh, and I just added this to `ViewportSettings`. Previously you wouldn't need to set this for what Tableview wants, but now you'll have to.

```cs
    /// <summary>
    ///     If set and <see cref="View.Viewport"/><c>.Width</c> is greater than <see cref="View.GetContentSize ()"/>
    ///     <c>.Width</c> <see cref="View.Viewport"/><c>.X</c> can be negative.
    ///     <para>
    ///         When not set, <see cref="View.Viewport"/><c>.X</c> will be constrained to non-negative values when
    ///         <see cref="View.Viewport"/><c>.Width</c> is greater than <see cref="View.GetContentSize ()"/>
    ///         <c>.Width</c>, preventing
    ///         scrolling beyond the left of the Viewport.
    ///     </para>
    ///     <para>
    ///         This can be useful in infinite scrolling scenarios.
    ///     </para>
    /// </summary>
    AllowNegativeXWhenWidthGreaterThanContentWidth = 16,

    /// <summary>
    ///     If set and <see cref="View.Viewport"/><c>.Height</c> is greater than <see cref="View.GetContentSize ()"/>
    ///     <c>.Height</c> <see cref="View.Viewport"/><c>.Y</c> can be negative.
    ///     <para>
    ///         When not set, <see cref="View.Viewport"/><c>.Y</c> will be constrained to non-negative values when
    ///         <see cref="View.Viewport"/><c>.Height</c> is greater than <see cref="View.GetContentSize ()"/>
    ///         <c>.Height</c>, preventing
[...]

PR #4628, opened 2026-01-25
TableView can handle Scrollbars and Viewport

BerndK-Str:
## Fixes

- #3822

## Proposed Changes/Todos

Please be patient, this is my first PR to this project. I don't expect that you accept this PR.
I just wanted to see proper scrollbars when using a TableView and want to share my results.
I have expected to 'switch' a setting to have this happen, as the base classes does already support Scrollbars and Viewport.
Now I know why you didn't do that till now. The whole concept of navigating and drawing is not compatible to Viewport.
I did my best to do it as smooth as possible, however I'm quite sure that I did some things not in the right way. This 'framework' is quite complex so I probably I did not understand all concepts on this first shot.
Anyways, perhaps you might find some useful code snippets here, if not, also fine for me. Especially the test-app is not needed in production code, but perhaps helpful for testing.

Don't know if you wish to keep the old behaviour, so I tried to use a flag: 'UseScrollbars' that should be able to switch between the two drawing concepts.

If you might find this code useful, I could go on an add more testing and documentation if needed. At this point where I don't know if this heads in the right direction, I did only what was needed for making this functional.

PS:
also want to mention that I'm not sure if it is really needed that hovering with the mouse over the window should trigger Full Redraw-Updates via View.NewMouseEnterEvent and NewMouseLeaveEvent!? I would expect that if a view needs redraw it should catch/override those events and in case trigger a redraw. But again most likely I can not see the whole picture here, but perhaps here is a chance for performance improvement, as a full redraw is 'expensive'. 

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/App/Application.Lifecycle.cs
+++ b/Terminal.Gui/App/Application.Lifecycle.cs
@@ -126,7 +126,7 @@ public static partial class Application // Lifecycle (Init/Shutdown)
         //Debug.Fail ("Application.Create() called");
         ApplicationImpl.MarkInstanceBasedModelUsed ();
 
-        IApplication app = new ApplicationImpl (timeProvider ?? new SystemTimeProvider (), false);
+        IApplication app = new ApplicationImpl (timeProvider ?? new SystemTimeProvider ());
         RaiseInstanceCreated (app);
 
         return app;
--- a/Terminal.Gui/App/ApplicationImpl.Driver.cs
+++ b/Terminal.Gui/App/ApplicationImpl.Driver.cs
@@ -46,6 +46,7 @@ internal partial class ApplicationImpl
             {
                 throw new ArgumentException ($"Driver '{driverName}' is not registered in DriverRegistry.");
             }
+
             // Determine which driver to use
             if (!string.IsNullOrEmpty (driverName) && DriverRegistry.TryGetDriver (driverName, out DriverRegistry.DriverDescriptor? descriptor))
             {
@@ -74,18 +75,22 @@ internal partial class ApplicationImpl
                     Coordinator = CreateSubcomponents (() => new WindowsComponentFactory ());
 
                     break;
+
                 case DriverRegistry.Names.DOTNET:
                     Coordinator = CreateSubcomponents (() => new NetComponentFactory ());
 
                     break;
+
                 case DriverRegistry.Names.UNIX:
                     Coordinator = CreateSubcomponents (() => new UnixComponentFactory ());
 
                     break;
+
                 case DriverRegistry.Names.ANSI:
                     Coordinator = CreateSubcomponents (() => new AnsiComponentFactory ());
 
                     break;
+
                 default:
                     throw new InvalidOperationException ($"Unknown driver name: {_driverName}");
             }
@@ -100,7 +105,7 @@ internal partial class ApplicationImpl
             throw new InvalidOperationException ("Driver was null even after booting MainLoopCoordinator");
         }
 
-        Driver.Force16Colors = Terminal.Gui.Drivers.Driver.Force16Colors;
+        Driver.Force16Colors = Drivers.Driver.Force16Colors;
     }
 
     private readonly IComponentFactory? _componentFactory;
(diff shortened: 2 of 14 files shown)


tig:
Sweet!!! I pulled it down locally and you are def. on the right track with this. 

> Don't know if you wish to keep the old behaviour, so I tried to use a flag: 'UseScrollbars' that should be able to switch between the two drawing concepts.

IMO, the new behavior should be the default and any old code that tried to do scrollbars natively should be removed. I actually don't know the TableView code that well, FWIW.

> also want to mention that I'm not sure if it is really needed that hovering with the mouse over the window should trigger Full Redraw-Updates via View.NewMouseEnterEvent and NewMouseLeaveEvent!? I would expect that if a view needs redraw it should catch/override those events and in case trigger a redraw. But again most likely I can not see the whole picture here, but perhaps here is a chance for performance improvement, as a full redraw is 'expensive'.

Would you mind pointing at the exact code you're talking about?

tig:
Looking more, my opinion is:

- Nuke `UseScrollBars` and just make your updated code the core behavior. Do not set `HorizontalScrollBar.AutoShow` or `VerticalScrollBar.AutoShow` by default; if a developer wants the scrollbars they can set `tv.HorizontalScrollBar.AutoShow = true` to enable. This is how `ListView` works.

tig:
@BerndK why did you close this? I'd love to keep working on it with you.

BerndK-Str:
Sorry, was my fault. Accidently I used my main branch 'v2_develop' to create the PR. I wanted to clean that up by renaming the branch. I did it in the web-UI to not loose the connection/ref to the PR - that was the plan, but they had other plans.
Renaming did 'remove' that branch, and that did automatically close the PR ;-)
I will come back with a new (and clean) PR...

BerndK-Str:
I'm sorry, I don't have much time right now. I have done all the functionality (also with nice scrolling).
No there are a ton of tests that needs to be modified.
Did I get you right? The idea was that all is based on ViewPort (not the optional version). This means that there will be no more 'RowOffset' nor 'ColumnOffset' (but ViewPort.X , Y). Correct?
I plan to create the new PR on Saturday.

tznind:
Be aware that width of content area of a TableView is indeterminate.

This is for performance.

H scroll should always be in columns, never Runes or Cells

V scroll is always 1 per row in the table (because TableView requires all rows to be single line in height).

The drawing system takes the current row/cell offset and then draws content till it can't fit any more on the screen.

This means only the visible portion of the table needs to be evaluated - instead of requiring that the entire tables content be resolved into a giant scrollable area.

This allows for very large tables and/or slow rendering getters without things falling over

BerndK-Str:
> H scroll should always be in columns, never Runes or Cells

Yes select columns (not sure about columns vs cells?), if a large cell cannot be drawn completely, there will be also soft scroll to reveal the content.

>V scroll is always 1 per row in the table (because TableView requires all rows to be single line in height).

sure. There is soft scroll, to reveal the header lines in case they are outside the viewport

>The drawing system takes the current row/cell offset and then draws content till it can't fit any more on the screen.

sure. The idea is that there will be no more row/cell offset, but ViewportX/Y!?
The viewport will change 'per column' when navigating with the keys (left/right).

> This means only the visible portion of the table needs to be evaluated - instead of requiring that the entire tables content be resolved into a giant scrollable area.
> This allows for very large tables and/or slow rendering getters without things falling over

The first shot will do the (full) content calculation only one time and caching the results.
We might add a viewport based content calculation (only use visible rows to caclulate H space needed) in a second step.
Not sure how the user experience will be if the column widhts are jumping around while scrolling. This is surprising in the current implementation, however, I understand the performance topic.

tznind:
Can you do as subclass? So that old implementation remains usable please.

The table abstraction layer allows for very big tables i.e. pagination.

So currently user can e.g. have an api response with tokens and tell model there are 1M rows.  Then in delegates just has to fetch the row offsets requested.

Enumerating full table is good for normal tables but want to support this use case still.

tznind:
Theres also the tree table view scenrio that puts a tree view into table form

There the row count changea every time you expand or collapse a node

I.e. dynamic tables are implicitly supported with the current approach.

If you want animations in cells etc

I guess im saying i dont want to loose the current versatility

BerndK-Str:
My first Idea was to provide both implementations in parallel.
Tig said, we should go for Viewport only.
Now I did a ton of changes, a mixed implementation is not possible (or would be not maintainable).
However, we are just talking about a small fraction - the V-Size calculation. There will be a way to do this in both ways.
As mentioned, I want to finish step one. Then we will tackle that issue too and find some solution.

BerndK-Str:
There is no need to accept the first step (will be created on Saturday or Fri evening). But we should first check if this heads to rigth direction. Then continue the discussion with a common understanding of the codebase.
Limiting the calculation to visible rows should be no big prob, just we need to find a good trigger to do so.
Then we can also look at TreeView scenario too.

PR #4649, opened 2026-01-29
Fixes #4645. ListView should have a SourceChanged event

### LINKED ISSUE #4645: `ListView` should have a `SourceChanged` event
I have a case in OCGV where the code would be much cleaner if there were an event for this.

BDisp:
## Fixes

- Fixes #4645

## Proposed Changes/Todos

- [x] Add ListView SourceChanged event

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/ListView/ListView.cs
+++ b/Terminal.Gui/Views/ListView/ListView.cs
@@ -129,11 +129,8 @@ public partial class ListView : View, IDesignable, IValue<int?>
 
             SelectedItem = null;
             _lastSelectedItem = null;
-
-            if (!OnSourceChanged ())
-            {
-                SourceChanged?.Invoke (this, EventArgs.Empty);
-            }
+            OnSourceChanged ();
+            SourceChanged?.Invoke (this, EventArgs.Empty);
             SetNeedsDraw ();
         }
     }
@@ -164,13 +161,9 @@ public partial class ListView : View, IDesignable, IValue<int?>
     }
 
     /// <summary>
-    ///     Determine if the <see cref="SourceChanged"/> event should be raised.
+    ///     Called when the <see cref="Source"/> property is changed. Invokes the <see cref="SourceChanged"/> event.
     /// </summary>
-    /// <returns>
-    ///     If <see langword="true"/> the event will be raised, otherwise <see langword="false"/>.
-    ///     The default implementation returns <see langword="false"/>.
-    /// </returns>
-    protected virtual bool OnSourceChanged () => false;
+    protected virtual void OnSourceChanged () { }
 
     /// <summary>
     ///     Event to raise when the <see cref="Source"/> data is changed.


Terminal.Gui/Views/ListView/ListView.cs
```diff
@@ -158,6 +159,16 @@ private void SourceOnCollectionChanged (object? sender, NotifyCollectionChangedE
         OnCollectionChanged (e);
     }
 
+    /// <summary>
+    ///     Call the event to raises the <see cref="SourceChanged"/>.
+    /// </summary>
+    protected virtual void OnSourceChanged () => SourceChanged?.Invoke (this, EventArgs.Empty);
```
tig: This does not follow the CWP pattern. All new events we add must folow the CWP.

tig (changes requested):

Terminal.Gui/Views/ListView/ListView.cs
BDisp: But you said:

> My use case doesn't require anything other than knowing the Source has changed. So, unless you have a use case that requires more, I'd use EventArgs.Empty.
>
> I also see no need for a pre-event (SourceChanging).

Why I need to follow the CWP pattern?

BDisp: I think you're meant this https://github.com/gui-cs/Terminal.Gui/pull/4649/commits/33da176b287d49a701152d068abddff6c0e506b0.

Terminal.Gui/Views/ListView/ListView.cs
```diff
@@ -158,6 +163,20 @@ private void SourceOnCollectionChanged (object? sender, NotifyCollectionChangedE
...
     }
 
+    /// <summary>
+    ///     Determine if the <see cref="SourceChanged"/> event should be raised.
+    /// </summary>
+    /// <returns>
+    ///     If <see langword="true"/> the event will be raised, otherwise <see langword="false"/>.
+    ///     The default implementation returns <see langword="false"/>.
+    /// </returns>
+    protected virtual bool OnSourceChanged () => false;
```
tig: This should be `void`. It can't be cancelled.

BDisp: So, my initial implementation was correct, right?

tig: > So, my initial implementation was correct, right?

No, you were invoking the event from the `OnSourceChanged` method in your inital implemetation. That was incorrect.

The pattern for a non-cancellable (after) CWP event is simple:

From the place where the event should be raised:

- Call `OnXXX`.
- Invoke `XXX`.

tig: Here's a good example (except this one does convey changed state, which is not needed here):

```cs
    private void RaiseViewportChangedEvent (Rectangle oldViewport)
    {
        if (Cursor.IsVisible)
        {
            // Adjust the cursor if visible
            int deltaX = oldViewport.X - Viewport.X;
            int deltaY = oldViewport.Y - Viewport.Y;

            Point currentCursorPos = ScreenToViewport (Cursor.Position!.Value);

            SetCursor (Cursor with { Position = ViewportToScreen (new Point (currentCursorPos.X + deltaX, currentCursorPos.Y + deltaY)) });
        }

        DrawEventArgs args = new (IsInitialized ? Viewport : Rectangle.Empty, oldViewport, null);
        OnViewportChanged (args);
        ViewportChanged?.Invoke (this, args);
    }

    /// <summary>
    ///     Fired when the <see cref="Viewport"/> changes. This event is fired after the <see cref="Viewport"/> has been
    ///     updated.
    /// </summary>
    public event EventHandler<DrawEventArgs>? ViewportChanged;

    /// <summary>
    ///     Called when the <see cref="Viewport"/> changes. Invokes the <see cref="ViewportChanged"/> event.
    /// </summary>
    /// <param name="e"></param>
    protected virtual void OnViewportChanged (DrawEventArgs e) { }
```

BDisp: Thanks for remind me that. Sorry about that. Done in https://github.com/gui-cs/Terminal.Gui/pull/4649/commits/552468888082a747bd6747e91514cb68b09edd4b.

PR #4673, opened 2026-02-02
Fixes #4672. OptionSelector is not behaving properly with keyboard.

### LINKED ISSUE #4672: OptionSelector is not behaving properly with keyboard.
**Describing some shortcomings of the OptionSelector, considering an unfocused OptionSelector with 3 options [0,1,2] and with item 1 selected:**
When the OptionSelector is focused, the cursor is not highlighted on the selected item, but rather on the first or last item, depending on whether the Tab key or the Shift+Tab keys were pressed.
- When the OptionSelector focus is lost, the cursor is not restored to the selected item, because it doesn't make sense as if it were a normal view..
- Pressing Tab or Shift+Tab does not move the focus to the next or previous view outside the OptionSelector, forcing you to press all the way to the last option or all the way to the first option in order to move the focus outside the OptionSelector.

BDisp:
## Fixes

- Fixes #4672

## Proposed Changes/Todos

- [x] Restoring cursor to value when OptionSelector gains or lose focus.
- [x] Pressing Tab or Shift+Tab on the OptionSelector moves the focus to the next or previous view outside it.

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Views/Selectors/SelectorBase.cs
+++ b/Terminal.Gui/Views/Selectors/SelectorBase.cs
@@ -45,8 +45,7 @@ public abstract class SelectorBase : View, IOrientation, IValue<int?>
 
     private bool MoveNext (Command command)
     {
-        if ((command == Command.Down && Orientation == Orientation.Horizontal)
-            || (command == Command.Right && Orientation == Orientation.Vertical))
+        if ((command == Command.Down && Orientation == Orientation.Horizontal) || (command == Command.Right && Orientation == Orientation.Vertical))
         {
             return false;
         }
@@ -59,6 +58,12 @@ public abstract class SelectorBase : View, IOrientation, IValue<int?>
         }
         else
         {
+            if (Styles.HasFlag (SelectorStyles.ShowValue))
+            {
+                _valueField?.SetFocus ();
+
+                return true;
+            }
             active = 0;
         }
         SubViews.OfType<CheckBox> ().ToArray ().ElementAt (active).SetFocus ();
@@ -68,21 +73,37 @@ public abstract class SelectorBase : View, IOrientation, IValue<int?>
 
     private bool MovePrevious (Command command)
     {
-        if ((command == Command.Up && Orientation == Orientation.Horizontal)
-            || (command == Command.Left && Orientation == Orientation.Vertical))
+        if ((command == Command.Up && Orientation == Orientation.Horizontal) || (command == Command.Left && Orientation == Orientation.Vertical))
         {
             return false;
         }
 
         int active = SubViews.OfType<CheckBox> ().ToArray ().IndexOf (Focused);
 
-        if (active > 0)
+        switch (active)
         {
-            active--;
-        }
-        else
-        {
-            active = SubViews.OfType<CheckBox> ().Count () - 1;
+            case -1 when Styles.HasFlag (SelectorStyles.ShowValue):
+                active = SubViews.OfType<CheckBox> ().Count () - 1;
+
+                break;
+
+            case > 0:
+                active--;
+
+                break;
+
+            default:
+            {
+                if (Styles.HasFlag (SelectorStyles.ShowValue))
+                {
+                    _valueField?.SetFocus ();
+
+                    return true;
+                }
+                active = SubViews.OfType<CheckBox> ().Count () - 1;
+
+                break;
+            }
         }
         SubViews.OfType<CheckBox> ().ToArray ().ElementAt (active).SetFocus ();
 
@@ -111,9 +132,7 @@ public abstract class SelectorBase : View, IOrientation, IValue<int?>
 
     private bool? HandleAcceptCommand (ICommandContext? ctx)
     {
-        if (!DoubleClickAccepts
-            && ctx?.Binding is MouseBinding mouseBinding
-            && mouseBinding.MouseEvent!.Flags.HasFlag (MouseFlags.LeftButtonDoubleClicked))
+        if (!DoubleClickAccepts && ctx?.Binding is MouseBinding mouseBinding && mouseBinding.MouseEvent!.Flags.HasFlag (MouseFlags.LeftButtonDoubleClicked))
         {
             return false;
         }
@@ -354,7 +373,6 @@ public abstract class SelectorBase : View, IOrientation, IValue<int?>
             MouseHighlightStates = DefaultMouseHighlightStates,
             TabStop = TabBehavior.NoStop
         };
-
 
         return checkbox;
     }
--- a/Terminal.Gui/Views/Selectors/OptionSelector.cs
+++ b/Terminal.Gui/Views/Selectors/OptionSelector.cs
@@ -11,7 +11,7 @@ namespace Terminal.Gui.Views;
 //  HotKey - Do NOT Restore Focus. Advance Active. Do NOT Accept.
 //  Item HotKey - Do NOT Focus item. If item is not active, make Active. Do NOT Accept.
 // Focused:
-//  Space key - If focused item is Active, move focus to and Acivate next. Else, Activate current. Do NOT Accept.
+//  Space key - If focused item is Active, move focus to and Activate next. Else, Activate current. Do NOT Accept.
 //  Enter key - Activate and Accept the focused item.
 //  HotKey - Restore Focus. Advance Active. Do NOT Accept.
 //  Item HotKey - If item is not active, make Active. Do NOT Accept.
@@ -34,7 +34,7 @@ public class OptionSelector : SelectorBase, IDesignable
     /// <inheritdoc/>
     protected override bool OnHandlingHotKey (CommandEventArgs args)
     {
-        if (base.OnHandlingHotKey (args) is true)
+        if (base.OnHandlingHotKey (args))
         {
             return true;
         }
@@ -65,7 +65,7 @@ public class OptionSelector : SelectorBase, IDesignable
     /// <inheritdoc/>
     protected override bool OnActivating (CommandEventArgs args)
     {
-        if (base.OnActivating (args) is true)
+        if (base.OnActivating (args))
         {
             return true;
         }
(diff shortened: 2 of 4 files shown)


BDisp:
I don't know why the CI runners weren't working.

tig:
Just tried re-running them and it appears it was a temporary gh outage.

tig:
https://www.githubstatus.com/incidents/xwn6hjps36ty

tig:
There's a far simpler way to change the focus behavior:

Change `TabStop` to `NoStop`:

```cs
  /// <summary>
  ///     INTERNAL: Creates a checkbox subview
  /// </summary>
  protected CheckBox CreateCheckBox (string label, int value)
  {
      var checkbox = new CheckBox
      {
          CanFocus = true,
          Title = label,
          Id = label,
          Data = value,
          MouseHighlightStates = DefaultMouseHighlightStates,
          TabStop = TabBehavior.NoStop
      };

      return checkbox;
  }
```

Then, add KeyBindings for up/down:

```cs
 KeyBindings.ReplaceCommands (Key.CursorDown, Command.Down);
 KeyBindings.ReplaceCommands (Key.CursorUp, Command.Up);

 AddCommand (Command.Down, () =>
                           {
                               int active = SubViews.OfType<CheckBox> ().ToArray ().IndexOf (Focused);

                               if (active < SubViews.OfType<CheckBox> ().Count () - 1)
                               {
                                   active++;
                               }
                               else
                               {
                                   active = 0;
                               }
                               SubViews.OfType<CheckBox> ().ToArray ().ElementAt (active).SetFocus ();
                               return true;
                           });

 AddCommand (Command.Up, () =>
                           {
                               int active = SubViews.OfType<CheckBox> ().ToArray ().IndexOf (Focused);

                               if (active > 0)
                               {
                                   active--;
                               }
                               else
                               {
                                   active = SubViews.OfType<CheckBox> ().Count () - 1;
                               }
                               SubViews.OfType<CheckBox> ().ToArray ().ElementAt (active).SetFocus ();
                               return true;
                           });
```

Note you'll also need to ensure up/down do the right thing when SelectorStyles.ShowValue is enabled as that adds a `TextField` as a subview.

Terminal.Gui/Views/Selectors/SelectorBase.cs
```diff
@@ -30,6 +30,51 @@ protected SelectorBase ()
...
+        AddCommand (Command.Down, () => MoveDownRight ());
+        AddCommand (Command.Right, () => MoveDownRight ());
+
+        AddCommand (Command.Up, () => MoveUpLeft ());
+        AddCommand (Command.Left, () => MoveUpLeft ());
+    }
+
+    private bool MoveDownRight ()
+    {
+        int active = SubViews.OfType<CheckBox> ().ToArray ().IndexOf (Focused);
```
tig, line 53: These should be smart about orientation. If orientation == vertical, left/right should advance focus (return false). If orientation == horiziontal, up/down should advance focus (return false). Make sense?

Terminal.Gui/Views/Selectors/SelectorBase.cs
```diff
@@ -30,6 +30,51 @@ protected SelectorBase ()
...
+        KeyBindings.ReplaceCommands (Key.CursorLeft, Command.Left);
+
+        AddCommand (Command.Down, () => MoveDownRight ());
+        AddCommand (Command.Right, () => MoveDownRight ());
+
+        AddCommand (Command.Up, () => MoveUpLeft ());
+        AddCommand (Command.Left, () => MoveUpLeft ());
+    }
+
+    private bool MoveDownRight ()
```
tig: Rename these to "MoveNext" and "MovePrevious". Then, in the AddCommand calls, do the check for orientation and return false if it doesn't match.

tig:
I've pushed a PR to your PR:

- Some code cleanup
- I addressed `SelectorStyles.ShowValue` which you did not address.
- I re-wrote your complex unit test that tested multiple things in one test, with focused unit tests that test specific functionality
- I do not agree with the `FocusedItem_IsResetToValueAfterLoseFocus` behavior and have removed it and the test. 

```cs
  [Fact]
  public void FocusedItem_IsResetToValueAfterLoseFocus ()
  {
      var container = new View { CanFocus = true };
      var optionSelector = new OptionSelector ();
      List<string> options = ["Option1", "Option2", "Option3"];
      optionSelector.Labels = options;
      var view = new View { CanFocus = true };
      container.Add (optionSelector, view);

      // Set focus to optionSelector
      optionSelector.SetFocus ();

      // Set Option2
      optionSelector.FocusedItem = 1;
      container.Layout ();

      // Verify current assertions
      Assert.Equal (0, optionSelector.Value);
      Assert.Equal (1, optionSelector.FocusedItem);

      // Move focus away
      view.SetFocus ();

      // Cursor should reset to Value
      Assert.Equal (0, optionSelector.FocusedItem);
  }
```

If a user has moved the Focused Item within an Option or FlagSelector and focuses elsewhere, and then focuses back, the focused item should remain the same.

BDisp:
> I do not agree with the FocusedItem_IsResetToValueAfterLoseFocus behavior and have removed it and the test.
>
> If a user has moved the Focused Item within an Option or FlagSelector and focuses elsewhere, and then focuses back, the focused item should remain the same.

I completely disagree. I agree with the case where superviews are aware of the functionalities of their respective subviews, such as Runnable, Frame, etc., as well as with FlagSelector, which allows multiple selections and is indifferent to which item is focused when it is focused again. But with OptionSelector, the case is different because it's a view that only allows a single selection. I confess that the functionality it had of keeping the view that was focused before losing focus highlighted, when it was focused again, leads me to believe that this view is the one accepted by OptionSelector, because highlighting the last focused view has much more visibility than the point inside the circle.
That being said, I will accept your PR because no matter how many arguments I present, you will never agree with my opinion, and possibly that of many others. But I want to state that I do so very reluctantly.

tig:
> > I do not agree with the FocusedItem_IsResetToValueAfterLoseFocus behavior and have removed it and the test.
> > If a user has moved the Focused Item within an Option or FlagSelector and focuses elsewhere, and then focuses back, the focused item should remain the same.
> 
> I completely disagree. I agree with the case where superviews are aware of the functionalities of their respective subviews, such as Runnable, Frame, etc., as well as with FlagSelector, which allows multiple selections and is indifferent to which item is focused when it is focused again. But with OptionSelector, the case is different because it's a view that only allows a single selection. I confess that the functionality it had of keeping the view that was focused before losing focus highlighted, when it was focused again, leads me to believe that this view is the one accepted by OptionSelector, because highlighting the last focused view has much more visibility than the point inside the circle. That being said, I will accept your PR because no matter how many arguments I present, you will never agree with my opinion, and possibly that of many others. But I want to state that I do so very reluctantly.

 I will continue to debate and argue using data and facts. If I discover something I was wrong about I will admit it and change my stance. 

I just wrote several apps to see how other UI platforms deal with this. We are both right and wrong.

HTML - There is only one concept: Active. A user cannot select one radio button and move to another without selecting the other. Tabbing to/from the radio group returns focus to the selected item.

WinForms - Same as HTML.

WPF - The selected and focused items are distinct (like OptionSelector). 

1. A user can use the keyboard to move from item to item without changing which is selected.  
2. Pressing tab MOVES BETWEEN THE radio buttons until the end is reached and then moves to the next control.
3. Tabbing into the radio group causes the first or last radio button to gain focus depending on if the other control is before or after the radio group in the tab order.

RadioFocusTestWpf_xFjm226QNM

So:

WPF acts as I think OptionSelector should regarding 2. I disagreed with you on this, but agreed to keep it your way before.

Regarding point 3: The code before this PR acted as WPF does. Now it doesn't

Based on this data what do you suggest?

BDisp:
> So:
> 
> WPF acts as I think OptionSelector should regarding 2. I disagreed with you on this, but agreed to keep it your way before.
> 
The fact that WPF acts in its own way doesn't mean they're right. Maybe they don't even bother to do it. Maybe they have an option if pressing Tab should navigate through options or move to next control.
Regarding point 2, I think you agree with me that it doesn't make sense to press Tab to navigate between radio buttons to move to another control if a radio button has 40 options.

> Regarding point 3: The code before this PR acted as WPF does. Now it doesn't
> 
Regarding point 3, I maintain my opinion of keeping the accepted item as focused, due to the fact that it is a single-selection control, but the divergence of opinions is not so serious on point 3 and is acceptable, but regarding point 2, it's good that you changed your mind because it didn't make the slightest sense.

> Based on this data what do you suggest?

Of course, if there were some kind of poll where people actually participated and gave their opinion, we would have broader support for you to make a decision. Considering that you agreed with my amendment regarding point 2, I end the discussion on my part.

BDisp:
In WPF, the KeyboardNavigation.DirectionalNavigation property controls this behavior for radio buttons.
By default, radio buttons in WPF use arrow keys to move between options within a group, and Tab moves to the next control outside the group. This is the standard Windows behavior.
If you want Tab to navigate through individual radio buttons instead, you can set:

```xml
<StackPanel KeyboardNavigation.DirectionalNavulation="Continue">
    <RadioButton Content="Option 1" />
    <RadioButton Content="Option 2" />
    <RadioButton Content="Option 3" />
</StackPanel>
```

The DirectionalNavigation property has these values:

Contained (default for most containers) - Arrow keys navigate within the group, Tab moves out
Continue - Tab navigates through all children sequentially
Cycle - Navigation cycles within the container
Once - Container gets focus as a whole, then Tab moves out
None - No directional navigation

For radio buttons specifically, you typically want the default behavior where users arrow-key between options in a group and Tab to move between different controls, as this matches Windows UI conventions.

tig:
I remembered why I wanted up/down to nav out: 

When an option or flag selector is in a menu it must behave this way. 

This is also why ListView does the same thing. 

Sigh.

BDisp:
I imagined it was because of the menu that you needed this behavior, but for that you have to override it in the menu's derived class and not affect the base class, or else create a property specifying how it should behave in a given implementation.

PR #4620, opened 2026-01-23
Fixes #4473 - Re-engineer`Command` system and Fix `Menu`s

tig:
## Fixes

- Fixes #4473
- Fixes #4624
- Fixes #3972

## Overview

Re-engineers the Terminal.Gui v2 command system for correctness, clarity, and composability. See [`docfx/docs/command.md`](docfx/docs/command.md) for the complete deep dive.

### What Changed

**Core infrastructure** (`View.Command.cs`):
- `CommandRouting` enum (`Direct`, `BubblingUp`, `DispatchingDown`, `Bridged`) replaces ad-hoc boolean flags
- Immutable `CommandContext` (`readonly record struct`) with `WithCommand`/`WithRouting`
- `CommandOutcome` enum with `bool?` conversion shims (migration deferred — see below)
- `GetDispatchTarget` / `ConsumeDispatch` virtuals for declarative composite-view dispatch
- `CommandBridge` for cross-boundary routing (weak refs, one-way)
- Route tracing via `Trace.CommandEnabled` (zero overhead in Release builds)
- `_lastDispatchOccurred` reset at handler entry to prevent stale-flag bugs

**View-specific**:
- **Shortcut**: Relay dispatch (`ConsumeDispatch=false`) with deferred completion via `CommandView_Activated`
- **OptionSelector/FlagSelector**: Consume dispatch (`ConsumeDispatch=true`), context-dependent targets
- **MenuBar**: Consume dispatch + `OnActivating` toggle logic for activation/deactivation
- **MenuBarItem**: `CommandBridge.Connect` for PopoverMenu, `Bridged` guard, `IsInitialized` guard
- **Menu**: `OnActivating` override dispatches to focused MenuItem
- **Button**: `IAcceptTarget`, Space/Enter → `Command.Accept`, does NOT raise Activating

### Design

See [`plans/finalizing-command-system.md`](plans/finalizing-command-system.md) for:
- Architecture overview and design invariants
- Dispatch pattern summary (relay vs. consume)
- Phase B deviations from original plan (7 documented)
- All completed work with commit references

See [`plans/menubar-reengineer.md`](plans/menubar-reengineer.md) for MenuBar bug fixes and test migration.

### Tests

| Suite | Passed | Failed | Skipped |
|-------|--------|--------|---------|
| UnitTestsParallelizable | 14,045 | 0 | 23 |
| UnitTests | ~1,001 | 0 | ~22 |

[...]

Issue #4485, opened 2025-12-12
Toggle is on by default and is horrible (TableView)


tznind:
## Describe the bug

Clicking around in the table results in leaving wierd selected cells at random.

This is the 'toggle' system.

## To Reproduce

![Image](https://github.com/user-attachments/assets/1e1bb6d8-eacc-457d-a6a7-d3e94e80bff7)

```csharp

using System.Data;
using Terminal.Gui.App;
using Terminal.Gui.ViewBase;
using Terminal.Gui.Views;



using (var app = Application.Create().Init())
{
    var w = new Window();

    var tv = new TableView()
    {
        Width = Dim.Fill(),
        Height = Dim.Fill()
    };

    DataTable dt = new DataTable();
    dt.Columns.Add("ID", typeof(int));
    dt.Columns.Add("Name", typeof(string));
    dt.Columns.Add("Age", typeof(int));

    dt.Rows.Add(1, "Alice", 30);
    dt.Rows.Add(2, "Bob", 25);
    dt.Rows.Add(3, "Charlie", 35);


    tv.Table = new DataTableSource(dt);
    w.Add(tv);

    app.Run(w);
}
```

2. Expected behavior: (describe what should happen)

Should work like datagrid/excel/spreadsheet when used out of the box

3. Actual behavior: (describe what actually happens)

What looks like graphical artifacts are left as you click in cells - this is it turns out them being 'toggled'

You cannot turn off toggling except by subclassing TableView and overriding the 

```
class t2:TableView
{
    protected override void OnCellToggled (CellToggledEventArgs args)
    {
       // Deliberately do nothing     
        args.Cancel = true;
    }
}
```

## Environment

Please run the following commands in your terminal and paste the output:

### OS Information

**Windows (PowerShell):**
```powershell
"OS: $(Get-CimInstance Win32_OperatingSystem | Select-Object -ExpandProperty Caption) $(Get-CimInstance Win32_OperatingSystem | Select-Object -ExpandProperty Version)"
```

**macOS/Linux:**
```bash
echo "OS: $(uname -s) $(uname -r)"
```

**Output:**
```
(paste output here)
```

### Terminal Information

**Windows Terminal:**
```powershell
"Terminal: Windows Terminal $(Get-AppxPackage -Name Microsoft.WindowsTerminal | Select-Object -ExpandProperty Version)"
```

**Other terminals:**
```bash
echo $TERM
```

**Output:**
```
(paste output here)
```

[...]

PR #4691, opened 2026-02-06
Fixes #3822 - Upgrades `TableView` to use Content Area for scrolling

### LINKED ISSUE #3822: `TableView` - Upgrade to use Content Area for scrolling

BerndK-Str:
## Fixes
- Fixes #3822 

## Original Postng

Hi again.
I'm back with my TableView that can handle the Viewport and the scrollbars.
I have two commits, the first has still the Flag to opt in to the new behaviour:
Tig said, we should have this as the normal and only behaviour, so I removed the switches and made it the default.
This is not final! I thought it would be simpler. I don't understand all the concepts in the library, however, i gave my best.
I placed some warnings where we should revisit. I also placed some ToDos, especially for some tests where I was not sure how to fix them. I stopped at test 'TableView_ExpandLastColumn_False', the following ones have not been touched.
Please do a first review, if we are on the right track.
Please try to answer some questions in the Unit Tests.

I also added a new Scenario TableViewTest

If you feel that we can continue, I will try to fix the remaining tests.
In the next step the plan is also to support using only the visible rows for calculation to be able to handle large data, probably using a boolean property for that.

Please also note that I don't understand two things:
- there is no redraw/event triggered if I modify e.g. tableView.Style.ShowHorizontalHeaderUnderline - however it works fine to modify the checkbox in new scenario TableViewTest - probably because of next topic:
- hovering over a chackbox does trigger a redraw of the checkbox - it also triggers a redraw of the whole window!!! So the TableView is also redrawn. Is this by intention? I temporary added a simple DrawCounter in the testview to show this behavoir.


tig:
Might help: https://gui-cs.github.io/Terminal.Gui/api/Terminal.Gui.ViewBase.ViewDiagnosticFlags.html

tznind:
> I also added a new Scenario TableViewTest

See also the TableEditor scenario it shows various use cases for TableView including for navigating entire file system

Ignore the 'sticky' highlighting that goes on with the mouse - it is bug #4485 

Screenshot 2026-02-06 193242

BerndK-Str:
Hello,

Again:
Should we continue on this path? Is it what you expect?
Is there a chance of accepting this?

the Draw diagnostic is cool, however still have the questions:
Why is there a full redraw of the whole window, if only a checkbox is hovered?
Is there any concept, when changing  tableView.Style.Xxxx - should trigger redraw of table?

You posted a picture of the treeview/table combination!? Do you think that my implementation fails here? I can't see a difference when comparing to main branch.

There are unit tests where I need help to understand (I added comments there):
- ScrollRight_WithoutSmoothScrolling
- ScrollIndicators (is commented out)
- TableView_ExpandLastColumn_False

My menus are dead when using the mouse, not sure if #4620 will fix this?

Examples/UICatalog/Scenarios/TableViewTest.cs
```diff
@@ -0,0 +1,137 @@
...
+using Terminal.Gui.Views;
+
+namespace UICatalog.Scenarios;
+
+[ScenarioMetadata ("TableViewTest", "Demonstrates and tests TableView.")]
+[ScenarioCategory ("TableView")]
+[ScenarioCategory ("Controls")]
+[ScenarioCategory ("Dialogs")]
+
+public class TableViewTest : Scenario
```
tznind, line 13: The TableEditor scenario already does all of this and more... is there a reason to add this?

Redraw counter is the new bit from what I can see.

Can you update table editor instead of making new scenario?

Terminal.Gui/Views/FileDialogs/FileDialog.cs
```diff
@@ -1093,7 +1093,10 @@ private void PushState (FileDialogState newState, bool addCurrentStateToHistory,
                 _history.ClearForward ();
             }
 
-            _tableView.RowOffset = 0;
+            if (_tableView.Viewport.Y != 0)
+            {
+                _tableView.Viewport = _tableView.Viewport with {Y = 0};
```
tznind, line 1095: Is this the new way of programatically setting the scroll? 

@tig im not familiar with viewport but is this normally how you would do 'scroll to top'?

Terminal.Gui/Views/TableView/TableView.Drawing.cs
```diff
@@ -30,36 +54,51 @@ protected override bool OnDrawingContent (DrawContext context)
                 │ArithmeticComparator│chi       │Healthboard│Interpretation│Labnumber│
                 └────────────────────┴──────────┴───────────┴──────────────┴─────────┘
             */
+
             if (Style.ShowHorizontalHeaderOverline)
             {
-                RenderHeaderOverline (line, Viewport.Width, columnsToRender);
-                line++;
+                if ((Viewport.Y <= line || Style.AlwaysShowHeaders) && line < Viewport.Height)
```
tznind: I would combine both conditionals in new method

ShouldShowHeaderOverline 

With comments explaining the booleam logic mlre simply?

Terminal.Gui/Views/TableView/TableView.Drawing.cs
```diff
@@ -89,10 +128,12 @@ protected override bool OnDrawingContent (DrawContext context)
     ///     <see cref="ColumnStyle.ColorGetter"/>. For changing the content that is rendered use
     ///     <see cref="ColumnStyle.RepresentationGetter"/>.
     /// </summary>
+    /// <param name="ypos">the x value where to render the cell (absolute value in context of GetContentSize)</param>
+    /// <param name="xpos">the y value where to render the cell (absolute value in context of GetContentSize)</param>
     /// <param name="cellAttribute"></param>
     /// <param name="render"></param>
     /// <param name="isPrimaryCell"></param>
-    protected virtual void RenderCell (Attribute cellAttribute, string render, bool isPrimaryCell)
+    protected virtual void RenderCell (int xpos, int ypos, Attribute cellAttribute, string render, bool isPrimaryCell)
```
tznind: Are these screen coordinates or view relative?

Terminal.Gui/Views/TableView/TableView.Drawing.cs
```diff
@@ -157,20 +198,57 @@ private void RenderBottomLine (int row, int availableWidth, ColumnToRender [] co
...
     }
 
-    private void RenderHeaderMidline (int row, ColumnToRender [] columnsToRender)
+    /// <summary>
+    /// renders the text but does clipping if needed
+    /// </summary>
+    /// <param name="col">col is absolute, see GetContentSize</param>
+    /// <param name="row"></param>
+    /// <param name="text"></param>
+    private void RenderStr (int col, int row, string text)
```
tznind: TruncateOrPad is already doing similar stuff to this... can you explain more about this method?

Terminal.Gui/Views/TableView/TableView.cs
```diff
@@ -38,10 +38,65 @@ public TableView ()
...
+                        var oldSelecteCol = SelectedColumn;
+                        var oldViewportX = Viewport.X;
+                        var result = ChangeSelectionByOffsetWithReturn (1, 0);
+
+                        if (oldSelecteCol == SelectedColumn && Viewport.X < MaxViewPort().X)
+                        {
+                            var maxViewPort = MaxViewPort ();
+                            Viewport = Viewport with {X = Math.Min(oldViewportX + 1, maxViewPort.X)};
+                        }
+                        return result;
```
tznind: These should continue to have the clean code single line format.  I.e. either update the method ChangeSelectionOffsetWithReturn or add a new one  but please dont make big lamdas in command binding area

Terminal.Gui/Views/TableView/TableView.cs
```diff
@@ -349,15 +430,34 @@ private record TableViewSelectionSnapshot (int SelectedColumn, int SelectedRow,
...
-        ChangeSelectionByOffset (0, Viewport.Height - GetHeaderHeightIfAny (), extend);
+        var oldSelectedRow = SelectedRow;
+        ChangeSelectionByOffset (0, Viewport.Height /* - CurrentHeaderHeightVisible ()*/, extend);
+
+        //after scrolling the cells, also scroll to lower line
+        var remainingJump = Viewport.Height - (SelectedRow - oldSelectedRow);
+        var maxViewPort = MaxViewPort();
+        if (remainingJump > 0 && Viewport.Y < maxViewPort.Y)
+        {
+            Viewport = Viewport with {Y = Math.Min (Viewport.Y + remainingJump, maxViewPort.Y)};
```
tznind, line 467: These methods previously were 1 line, now complicated and it looks like duplicated code with the Command scrolling.

Can you consolidate this code?

tznind (changes requested):

tznind:
Great work.  Ive left some comments based only on looking at code. I need to test it out in TableEditor.

tznind:
Btw F9 opens menus - handy while mouse menus are broken

tznind:
So have tested with the big table (first scenario in TableEditor).

On your branch scrolling with mouse wheel is slow and soon lags vs v2_develop.  Can you confirm you also see this behaviour?

v2_develop the scrolling is very smooth up and down table.

Maybe to do with the new measuring code?

tznind:
UPDATE: have raised this as a seperate bug.


Also for some reason opening the char map (File->OpenCharacterMap)  scenario causes entire app to close 😕 

It crashes at RepresentationGetter so may just be that you are enumerating something the other doesn't until it scrolls into view...

The error is around:
```
new Rune ((uint)o).ToString ()
o is 55299
```
So bombing in Run constructor I think which means its likely unrelated to your code and if you scroll down far enough on v2_develop it will also crash.

```
System.ArgumentOutOfRangeException
  HResult=0x80131502
  Message=Specified argument was out of the range of valid values. (Parameter 'value')
  Source=System.Private.CoreLib
  StackTrace:
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument)
   at System.Text.Rune..ctor(UInt32 value)
   at UICatalog.Scenarios.TableEditor.<>c.<BuildUnicodeMap>b__13_0(Object o) in D:\Repos\temp\gui.cs\Examples\UICatalog\Scenarios\TableEditor.cs:line 568
   at Terminal.Gui.Views.ColumnStyle.GetRepresentation(Object value) in D:\Repos\temp\gui.cs\Terminal.Gui\Views\TableView\ColumnStyle.cs:line 107
   at Terminal.Gui.Views.TableView.GetRepresentation(Object value, ColumnStyle colStyle) in D:\Repos\temp\gui.cs\Terminal.Gui\Views\TableView\TableView.cs:line 922
   at Terminal.Gui.Views.TableView.CalculateMaxCellWidth(Int32 col, ColumnStyle colStyle) in D:\Repos\temp\gui.cs\Terminal.Gui\Views\TableView\TableView.cs:line 659
   at Terminal.Gui.Views.TableView.CalculateContentSize() in D:\Repos\temp\gui.cs\Terminal.Gui\Views\TableView\TableView.cs:line 828
   at Terminal.Gui.Views.TableView.RefreshContentSize() in D:\Repos\temp\gui.cs\Terminal.Gui\Views\TableView\TableView.Navigation.cs:line 37
   at Terminal.Gui.Views.TableView.set_Table(ITableSource value) in D:\Repos\temp\gui.cs\Terminal.Gui\Views\TableView\TableView.Navigation.cs:line 25
   at UICatalog.Scenarios.TableEditor.SetTable(DataTable dataTable) in D:\Repos\temp\gui.cs\Examples\UICatalog\Scenarios\TableEditor.cs:line 948
   at UICatalog.Scenarios.TableEditor.OpenUnicodeMap() in D:\Repos\temp\gui.cs\Examples\UICatalog\Scenarios\TableEditor.cs:line 823
   at Terminal.Gui.Views.Shortcut.DispatchCommand(ICommandContext commandContext) in D:\Repos\temp\gui.cs\Terminal.Gui\Views\Shortcut.cs:line 307
   at Terminal.Gui.Views.MenuItem.DispatchCommand(ICommandContext commandContext) in D:\Repos\temp\gui.cs\Terminal.Gui\Views\Menu\MenuItem.cs:line 142
   at Terminal.Gui.ViewBase.View.InvokeCommand(Command command, IInputBinding binding) in D:\Repos\temp\gui.cs\Terminal.Gui\ViewBase\View.Command.cs:line 476
   at Terminal.Gui.ViewBase.View.InvokeCommands(Command[] commands, IInputBinding binding) in D:\Repos\temp\gui.cs\Terminal.Gui\ViewBase\View.Command.cs:line 442
   at Terminal.Gui.ViewBase.View.InvokeCommandsBoundToKey(Key key) in D:\Repos\temp\gui.cs\Terminal.Gui\ViewBase\View.Keyboard.cs:line 599
   at Terminal.Gui.ViewBase.View.NewKeyDownEvent(Key key) in D:\Repos\temp\gui.cs\Terminal.Gui\ViewBase\View.Keyboard.cs:line 462
   at Terminal.Gui.ViewBase.View.NewKeyDownEvent(Key key) in D:\Repos\temp\gui.cs\Terminal.Gui\ViewBase\View.Keyboard.cs:line 448
   at Terminal.Gui.ViewBase.View.NewKeyDownEvent(Key key) in D:\Repos\temp\gui.cs\Terminal.Gui\ViewBase\View.Keyboard.cs:line 448
   at Terminal.Gui.App.ApplicationPopover.DispatchKeyDown(Key key) in D:\Repos\temp\gui.cs\Terminal.Gui\App\ApplicationPopover.cs:line 204
   at Terminal.Gui.App.KeyboardImpl.RaiseKeyDownEvent(Key key) in D:\Repos\temp\gui.cs\Terminal.Gui\App\Keyboard\KeyboardImpl.cs:line 167
   at Terminal.Gui.App.ApplicationImpl.Driver_KeyDown(Object sender, Key e) in D:\Repos\temp\gui.cs\Terminal.Gui\App\ApplicationImpl.Driver.cs:line 198
   at Terminal.Gui.Drivers.DriverImpl.<.ctor>b__0_0(Object s, Key e) in D:\Repos\temp\gui.cs\Terminal.Gui\Drivers\DriverImpl.cs:line 50
   at Terminal.Gui.Drivers.InputProcessorImpl`1.RaiseKeyDownEvent(Key a) in D:\Repos\temp\gui.cs\Terminal.Gui\Drivers\Input\InputProcessorImpl.cs:line 238
   at Terminal.Gui.Drivers.InputProcessorImpl`1.ProcessAfterParsing(TInputRecord input) in D:\Repos\temp\gui.cs\Terminal.Gui\Drivers\Input\InputProcessorImpl.cs:line 165
   at Terminal.Gui.Drivers.WindowsInputProcessor.Process(InputRecord inputEvent) in D:\Repos\temp\gui.cs\Terminal.Gui\Drivers\WindowsDriver\WindowsInputProcessor.cs:line 47
   at Terminal.Gui.Drivers.InputProcessorImpl`1.ProcessQueue() in D:\Repos\temp\gui.cs\Terminal.Gui\Drivers\Input\InputProcessorImpl.cs:line 102
   at Terminal.Gui.App.ApplicationMainLoop`1.IterationImpl() in D:\Repos\temp\gui.cs\Terminal.Gui\App\MainLoop\ApplicationMainLoop.cs:line 138
   at Terminal.Gui.App.ApplicationMainLoop`1.Iteration() in D:\Repos\temp\gui.cs\Terminal.Gui\App\MainLoop\ApplicationMainLoop.cs:line 122
   at Terminal.Gui.App.MainLoopCoordinator`1.RunIteration() in D:\Repos\temp\gui.cs\Terminal.Gui\App\MainLoop\MainLoopCoordinator.cs:line 97
   at Terminal.Gui.App.ApplicationImpl.RunLoop(IRunnable runnable, Func`2 errorHandler) in D:\Repos\temp\gui.cs\Terminal.Gui\App\ApplicationImpl.Run.cs:line 266
   at Terminal.Gui.App.ApplicationImpl.Run(IRunnable runnable, Func`2 errorHandler) in D:\Repos\temp\gui.cs\Terminal.Gui\App\ApplicationImpl.Run.cs:line 237
   at UICatalog.Scenarios.TableEditor.Main() in D:\Repos\temp\gui.cs\Examples\UICatalog\Scenarios\TableEditor.cs:line 320
   at UICatalog.Runner.RunScenario(String scenarioName, Boolean benchmark) in D:\Repos\temp\gui.cs\Examples\UICatalog\Runner.cs:line 61
   at UICatalog.Runner.RunInteractive[T](Boolean enableConfigWatcher) in D:\Repos\temp\gui.cs\Examples\UICatalog\Runner.cs:line 270
   at UICatalog.UICatalog.UICatalogMain(UICatalogCommandLineOptions options) in D:\Repos\temp\gui.cs\Examples\UICatalog\UICatalog.cs:line 318
   at UICatalog.UICatalog.Main(String[] args) in D:\Repos\temp\gui.cs\Examples\UICatalog\UICatalog.cs:line 205

```

Examples/UICatalog/Scenarios/TableViewTest.cs
BerndK-Str: Sure if you like I will remove the new scenario when all done. That's fine.
However, I like it to do my tests for now.

Terminal.Gui/Views/FileDialogs/FileDialog.cs
BerndK-Str: There is no more need for RowOffset, so I will remove it.
Using the Viewport has the same effect.
I can imagine that there might be a more convinient method to set just X/Y!? As this concept is global, there might be other use cases too.
Perhaps there is already such a feature and I didn't see it.

Terminal.Gui/Views/TableView/TableView.Drawing.cs
BerndK-Str: There is a comment: 'the x value where to render the cell (absolute value in context of GetContentSize)',
perhaps this is not precise. I think it is view relative, but respecting Viewport.X/Y.

BDisp: That's why these parameters are unnecessary. Since it's view-relative, the layout will take care of defining the correct location.

Terminal.Gui/Views/TableView/TableView.Drawing.cs
BerndK-Str: There are 3 similar checks, as they use local variables, it felt more natural to move to a local method 'ShouldRenderNextHeaderLine()'. I added a comment.

Terminal.Gui/Views/TableView/TableView.Drawing.cs
BerndK-Str: @BDisp Correct! When using Move before calling RenderCell, we can omit these params.

Terminal.Gui/Views/TableView/TableView.Drawing.cs
BerndK-Str: That method did limit the string to only visible ViewPort. As this is obviously already done by the driver, I removed that method and just call AddStr now.

Terminal.Gui/Views/TableView/TableView.cs
BerndK-Str: Yep, you are right, the former code looks better.
But my code does handle 2 things: a) handle selection b) moving the viewport (in rare cases, e.g. to smoth scroll when having a big cell that cannot be drawn in small viewport)
When 'merging' these two aspects to one method, that will violate single responsibility principle - handling selection and doing some scrolling.
As no user needs to use/modify that, the benefit of simplifying this is quite low - not sure if it is simpler when multiple emthods introduced.
However, it is not my code, please feel free to refactor that, if you find a better solution.

Terminal.Gui/Views/TableView/TableView.cs
BerndK-Str: Same as previous finding.

tznind:
Thanks for the tidying and responses to comments.  The biggest problem however remains the performance e.g. scrolling with mouse wheel in the big example table.

I don't think measuring all the content is viable.  Is there another way to support viewport but keep the existing measuring code in place?  

We don't need to scroll horizontally within a cell, if the cell fits it renders, if not it is pushed to next area i.e. off screen.  Thus the column widths are dynamic based on the visible content.  This is really nice for most use cases.

BerndK-Str:
@tznind You are absolutely right! We also need to look a the performance.
As I already said, I will tackle that after the general stuff is working.
So currently I am working on that, especially after your comment while testing the big table in TableEditor.
Stay tuned...

BerndK-Str:
Performance should be quite OK now! (introduced caching of ColumnsToRender)
Even when Flag 'UseAllRowsForContentCalculation' is true - performance is pretty well.
For me it does look almost OK now. Perhaps we need to look into some edge cases...
But this should be a basis that should work for most use cases.

BerndK-Str:
Can you please review the latest changes on performance, then we can try to fix all unit tests.

BDisp:
Many of the unit tests are failing unnecessarily. I advise different handling for when scrollbars are enabled and disabled. I see that Style.SmoothHorizontalScrolling is not working the same way, even with scrollbars disabled. When scrollbars are enabled, I understand that the viewport should be used, but when scrollbars are disabled, the original behavior could be maintained without using the viewport.
After enabling Scrollbars Auto, unchecking it again does not disable scrollbars.

BerndK-Str:
@BDisp Thanks for the input.
I looked at the SmoothScrolling:
It seems that my implementation does only smoothscrolling - always jump to the next cell and move that into the view.
The original implementation does the same when SmoothScrolling is on - right?
When off, it jumps 'page-wise' when scrolling to right, but single cell when scroll to left???
What is the wanted behaviour for smootscrolling off? Try to scroll as many cells that can be fully shown, at least one?

Regarding the scrollbars: I just set AutoShow to true and back to false. The rest is done by the scrollbars-implementation, not my code.
Shall I actively hide them in the TableEditor-Test?

Your idea was a different handling if scrollbars are visible or not? From my understanding, the scrollbars are just showing where in the grid we are, provides a mouse-option for quick navigation, but the navigation (especially with cursor keys) and look should be identical!? Please specify what other behaviour you have in mind.

BDisp:
You should observe the errors that the unit tests are returning and compare the current behavior with the previous one. As @tznind pointed out, it's possible to maintain the original functionalities if both scrollbars are disabled and use the new functionalities if one of the scrollbars is active.

Regarding ScrollRight_SmoothScrolling, if it's true and the cursor is in the rightmost visible column in the viewport, pressing the right arrow key moves the cursor to the next column and only the first column is removed from the viewport, assuming the currently selected column has roughly the same dimensions as the previous columns. If ScrollRight_SmoothScrolling is false and the cursor is in the rightmost visible column in the viewport, pressing the right arrow key moves the cursor to the next column and all previous columns are removed from the viewport, making the current column the first column in the viewport. Take a look at the ScrollRight_SmoothScrolling and ScrollRight_WithoutSmoothScrolling unit tests, and you'll better understand my explanation.

```
ScrollRight_SmoothScrolling:

│A│B│C│
│1│2│3│

│B│C│D│
│2│3│4│
```

```
ScrollRight_WithoutSmoothScrolling:

│A│B│C│
│1│2│3│

│D│E│F│
│4│5│6│
```

Scroll bars are a great advantage for observing point-by-point (xy) details, especially in columns larger than the viewport size. On the other hand, when we want to navigate faster using only the navigation keys, point-by-point scrolling makes navigation much slower, especially in huge tables. Therefore, my suggestion would be to create a new property, perhaps SmoothXYScrolling, that differentiates whether the navigation is point-by-point (XY location) or column-by-column/row-by-row. If scroll bars are active, the SmoothXYScrolling property should be set to true, otherwise, it can be true or false, depending on the user's preference. You might also want to set it to false even when scroll bars are active, so the scrolling would be column-by-column/row-by-row. This is debatable. We could even use the SmoothXYScrolling property to decide the general behavior, whether scroll bars are used or not.

@tznind is right about not exaggerating the size of the lambdas, and since the ones you modified already contained methods, you can include all the code logic inside those methods, instead of including it in the lambdas and then calling the methods.

BDisp:
Please review the previous comment where I made some changes, namely the property being called SmoothXYScrolling, indicating whether it should be considered point-to-point navigation (X and Y locations) or column-to-column/row-to-row.

Sorry, regarding the anomaly of the scrollbars not being disabled when set to false, this is due to the fact that you are using the AutoShow property, which by itself does not disable the scrollbars.

BerndK-Str:
Hi @BDisp 
I think we're completely talking past each other - sorry on weak communication skills.
I asked, if you think that my current implementation is correct when smoothscrolling on on? I assume 'yes' what do you think?

I think I completely understand your given example (also I did in my post), and yes, the sample you described is not working/implemented right now in my branch - it is on my list.
Prior implementing this, my question was about left scrolling. The implementation on main branch lookt to me like this:
ScrollLeft_WithoutSmoothScrolling:
│D│E│F│
│4│5│6│

│C│D│E│
│4│5│6│

but I expected?
│A│B│C│
│1│2│3│

Unfortunately I only found a test for right, not for left!

So shall the left scroll work in the same way as right scroll? I sssume so, but the current main branch does not. So after confirming or answering this question, I will implement that accordingly.

Regarding the Lambda - if you want to mix Selection and Viewport in one method, you are free to refactor that at any time. I personally don't see a benefit for now, and will leave it as is, as long as it works as expected.

No one has asked for smooth-scrolling on a basis of a single char/x-pos. My current implementation uses scroll by Column and by row (as it was before). No matter if the Scrollbars are visible or not. Again, my idea is, that the navigation is independant from the scrollbars. If they are visible, they might guide the user and enable the use of mouse - but the keys behave the same in either way.

We can discuss zu add SmoothXYScrolling (horizontal scroll per char/xpos not by column) later.

BDisp:
> Hi @BDisp I think we're completely talking past each other - sorry on weak communication skills. I asked, if you think that my current implementation is correct when smoothscrolling on on? I assume 'yes' what do you think?
> 
You're right, sorry. Using the keys draws column by column and row by row. Now, without SmoothScrolling, you just need to apply the previous functionality. @tznind's opinion would be helpful, since he created TableView, TabView, TreeView, etc., and he knows better than I do.

> I think I completely understand your given example (also I did in my post), and yes, the sample you described is not working/implemented right now in my branch - it is on my list. Prior implementing this, my question was about left scrolling. The implementation on main branch lookt to me like this: ScrollLeft_WithoutSmoothScrolling: │D│E│F│ │4│5│6│
> 
> │C│D│E│ │4│5│6│
> 
> but I expected? │A│B│C│ │1│2│3│
> 
> Unfortunately I only found a test for right, not for left!
> 
> So shall the left scroll work in the same way as right scroll? I sssume so, but the current main branch does not. So after confirming or answering this question, I will implement that accordingly.
> 
That's right, because I hadn't realized there wasn't a unit test for the left arrow key. You're correct in your observation, as I think it should follow the same criteria for the right arrow key.

```
ScrollLeft_WithoutSmoothScrolling:
│D│E│F│
│4│5│6│

│A│B│C│
│1│2│3│
```

> Regarding the Lambda - if you want to mix Selection and Viewport in one method, you are free to refactor that at any time. I personally don't see a benefit for now, and will leave it as is, as long as it works as expected.
> 
No problem, this isn't anything serious either.

> No one has asked for smooth-scrolling on a basis of a single char/x-pos. My current implementation uses scroll by Column and by row (as it was before). No matter if the Scrollbars are visible or not. Again, my idea is, that the navigation is independant from the scrollbars. If they are visible, they might guide the user and enable the use of mouse - but the keys behave the same in either way.
> 
If the keyboard and mouse are indeed working as before, regardless of whether the scroll bars are being used or not, then you should ensure that all unit tests pass successfully. I think you agree with me.

> We can discuss zu add SmoothXYScrolling (horizontal scroll per char/xpos not by column) later.

This idea no longer makes sense if what I mentioned in the previous point is true, wouldn't you agree?

I suggest you also add a specific option to show or hide the scroll bars.

```cs
            ("Scrollbars Auto", () => tableView.HorizontalScrollBar.AutoShow, b => { tableView.HorizontalScrollBar.AutoShow = b; tableView.VerticalScrollBar.AutoShow = b; }),
            ("Show Scrollbars", () => tableView.HorizontalScrollBar.Visible, b => { tableView.HorizontalScrollBar.Visible = b; tableView.VerticalScrollBar.Visible = b; }),
```

If it is useful to you, I present the unit tests related to handling the left key to test SmoothScrolling.

```cs
    [Fact]
    [SetupFakeApplication]
    public void ScrollLeft_SmoothScrolling ()
    {
        var tableView = new TableView ()
        {
            App = ApplicationImpl.Instance
        };
        tableView.BeginInit ();
        tableView.EndInit ();

        tableView.SchemeName = "Runnable";
        tableView.LayoutSubViews ();

        // 3 columns are visible
        tableView.Viewport = new (0, 0, 7, 5);
        tableView.Style.ShowHorizontalHeaderUnderline = false;
        tableView.Style.ShowHorizontalHeaderOverline = false;
        tableView.Style.AlwaysShowHeaders = true;
        tableView.Style.SmoothHorizontalScrolling = true;

        var dt = new DataTable ();
        dt.Columns.Add ("A");
        dt.Columns.Add ("B");
        dt.Columns.Add ("C");
        dt.Columns.Add ("D");
        dt.Columns.Add ("E");
        dt.Columns.Add ("F");

        dt.Rows.Add (1, 2, 3, 4, 5, 6);

        tableView.Table = new DataTableSource (dt);

        // select column D which is just off the right of the screen
        tableView.ColumnOffset = 3; // column D
        tableView.SelectedColumn = 3; // column D

        tableView.Draw ();

        var expected =
            $@"
│D│E│F│
│4│5│6│";

        DriverAssert.AssertDriverContentsAre (expected, output);

        // Scroll left
        tableView.NewKeyDownEvent (new () { KeyCode = KeyCode.CursorLeft });
        tableView.SetClipToScreen ();
        tableView.Draw ();

        expected =
            $@"
│C│D│E│
│3│4│5│";

        DriverAssert.AssertDriverContentsAre (expected, output);
    }

    [Fact]
    [SetupFakeApplication]
    public void ScrollLeft_WithoutSmoothScrolling ()
    {
        var tableView = new TableView ()
        {
            App = ApplicationImpl.Instance
        };

        tableView.BeginInit ();
        tableView.EndInit ();
        tableView.SchemeName = "Runnable";

        // 3 columns are visible
        tableView.Viewport = new (0, 0, 7, 5);
        tableView.Style.ShowHorizontalHeaderUnderline = false;
        tableView.Style.ShowHorizontalHeaderOverline = false;
        tableView.Style.AlwaysShowHeaders = true;
        tableView.Style.SmoothHorizontalScrolling = false;

        var dt = new DataTable ();
        dt.Columns.Add ("A");
        dt.Columns.Add ("B");
        dt.Columns.Add ("C");
        dt.Columns.Add ("D");
        dt.Columns.Add ("E");
        dt.Columns.Add ("F");

        dt.Rows.Add (1, 2, 3, 4, 5, 6);

        tableView.Table = new DataTableSource (dt);

        // select column D which is just off the right of the screen
        tableView.ColumnOffset = 3; // column D
        tableView.SelectedColumn = 3; // column D

        tableView.Draw ();

        var expected =
            $@"
│D│E│F│
│4│5│6│";

        DriverAssert.AssertDriverContentsAre (expected, output);

        // Scroll left
        tableView.NewKeyDownEvent (new () { KeyCode = KeyCode.CursorLeft });
        tableView.SetClipToScreen ();
        tableView.Draw ();

        expected =
            $@"
│A│B│C│
│1│2│3│";

        DriverAssert.AssertDriverContentsAre (expected, output);
    }
```

BDisp:
Just to remind you, although you probably already know this, the two codes below don't work the same way. `ColumnOffset` considers the entire area occupied by the selected column, while `Viewport.X` only considers the X location, ignoring the column boundaries.

```
tableView.ColumnOffset = 3; // column D
│D│E│F│
│4│5│6│
```

```
tableView.Viewport = tableView.Viewport with { X = 3 }; // column D ignoring the column boundaries
B│C│D│E
2│3│4│5
```

BerndK-Str:
Perfect. Seems that we are on the very same page now.

Sure I will add Scrollbar.Visible
Sure I will implement also a test for Scrolling left.

BerndK-Str:
Some more topics to discuss:
- ShowHorizontalScrollIndicators are no more in use-> R.I.P. for this property?
- RowOffset and ColumnOffset are not needed anymore (as the only truth is Viewport.X/Y)! However we might give them a second life as calculated properties for convenience. Not sure what the precise meaning would be? We have a Header which is now scrolled (instead of popping away). Should the RowOffset just be the number of rows not visible?
- refresh/recalc: as the RecalculateContentSize was 'expensive', I did it only when setting the table or manually calling RefreshContentSize (perhaps this name will change). Now when it is again cheap (as long as UseAllRowsForContentCalculation is false), we might do it on every Draw()? Currently it is done on ViewPortChange which means on scrolling and on Size change.

BerndK-Str:
You mentioned "columnOffset != Viewport.X"
for sure I know this :-)
That's why I brought up the discussion on re-activate them (ColumnOffset/RowOffset).
However not really sure if there is a real use case for that?
Besides moving the Viewport, there is also SelectedRow/Cell which will also bring the cell into view (if I'm right). Edit: nope, this won't move the Viewport, i tested it.
Is there really a use case to define the amount of hidden cells (except unit tests)?

tznind:
I really like the way you can new viewport scroll

We need a new menu item for show/hide scrollbars. 

It seems that the set max width on column no longer live updates after changes.  One thing to note I found that on v2_develop I have to accept with spacebar on Ok or the event gets lost.  I think this is just part of @tig ongoing refactoring of the broken accept stuff.

- Open big example
- Press F9 (sometimes twice) to open menu
- Go to column options and select max width
- Enter 2
- Tab to the Ok
- Press Space

Expect to see the column update but instead it stays same.

I think thers a method like update table to match styles or apply style changes or something if you need to respond.  Or you might be out of sync with v2_develop and just need to sync your branch?

Screenshot 2026-02-11 182319

BDisp:
> Some more topics to discuss:
> 
> * ShowHorizontalScrollIndicators are no more in use-> R.I.P. for this property?

It no longer exists. I think setting only the Visible property is sufficient.

> * RowOffset and ColumnOffset are not needed anymore (as the only truth is Viewport.X/Y)! However we might give them a second life as calculated properties for convenience. Not sure what the precise meaning would be? We have a Header which is now scrolled (instead of popping away). Should the RowOffset just be the number of rows not visible?

I think you understood the difference between ColumnOffset and Viewport.X and noticed the difference in output between them in the unit test. If you can achieve the same goal as ColumnOffset and RowOffset, which return the area including its boundaries, using only Viewport.X and Viewport.Y, then there's no problem in getting rid of those properties. If you agree to maintain the original functionality when scrollbars are not used, then you should also agree that the unit tests should pass without any changes. @tznind if you agree to keep the new output functionality on the screen, then it's not worth debating this issue further, and it's just a matter of adapting the unit tests to the new functionality. Please let us know your opinion. Thanks.

> * refresh/recalc: as the RecalculateContentSize was 'expensive', I did it only when setting the table or manually calling RefreshContentSize (perhaps this name will change). Now when it is again cheap (as long as UseAllRowsForContentCalculation is false), we might do it on every Draw()? Currently it is done on ViewPortChange which means on scrolling and on Size change.

It is not advisable to make any changes within the Draw() method, because it is one of the last phases of an iteration (Layout, Draw, and PositionCursor). That said, it is advisable to use the OnViewPortChange method.

BerndK-Str:
@tznind Please read my initial PR Comment again. There you can find the explanation why setting some ColumnStyle props sometimes works sometimes not. These are simple POCOs, so they do not trigger any refresh at all.
Sometimes it works because of 'everything is recalculated on redraw', that is exactly what I mentioned in my comment 40 min before. I consider to change that, however it does not feel right, because even if a Draw would recalc everything (there is no real need for that), the remaining question is: "Why is there a redraw for the table if you use some menu (let's assume that the menu does not cover the table)?"
So that is not deterministic, it is 'works by chance/random'.

tznind:
> Sometimes it works because of 'everything is recalculated on redraw', 

In your code it only updates when you scroll down far enough to trigger recompute of content

Can replicate by setting size to 2 and then scrolling down.  Once cursor leaves bottom row onto next it takes effect.

The Scenario code shows how it works its not POCO exactly.  You should tie into the Update code:

```csharp
        if (accepted)
        {
            try
            {
                setter (style, int.Parse (tf.Text));
            }
            catch (Exception ex)
            {
                MessageBox.ErrorQuery (_tableView.App!, "Failed to set", ex.Message, "Ok");
            }

            _tableView!.Update ();
        }
```


Update is the way for user to tell you stuff has changed

BDisp:
> You mentioned "columnOffset != Viewport.X" for sure I know this :-) That's why I brought up the discussion on re-activate them (ColumnOffset/RowOffset).

Excellent.

> However not really sure if there is a real use case for that? Besides moving the Viewport, there is also SelectedRow/Cell which will also bring the cell into view (if I'm right). Edit: nope, this won't move the Viewport, i tested it. Is there really a use case to define the amount of hidden cells (except unit tests)?

You'd have to look at the original code to see how it actually works. One suggestion would be to make ColumnOffset and RowOffset private fields, _columnOffset and _rowOffset, and when you actually want to use their values, assign them to Viewport.X and Viewport.Y, respectively. To use their functionality, then create a public method, perhaps `int GetViewportX(int column)` and `int GetViewportY(int row)`, to return the entire area including the boundaries. I think this might be necessary if a user wants to scroll programmatically.

BerndK-Str:
@BDisp 
> It is not advisable to make any changes within the Draw() method, because it is one of the last phases of an iteration (Layout, Draw, and PositionCursor). That said, it is advisable to use the OnViewPortChange method.

I'm not talking on 'making changes' during the draw. The problem is that some things (mostly TableView.Style (class TableStyle) and also ColumnStyle) are POCOs, see prior comment. So changing them has no effect at all.
@tznind In the old code every such setting is being respected in 'CalculateViewport' (the name was misleading), this method did calculate the columns to be rendered. It also did all the needed CalculateMaxCellWidth on every redraw. Not really good for performance.
In my current implementation, as mentioned it is done on ViewportChanged -> the reason why it is currently visible after a scroll.
As mentioned I consider to also call all the calculation on any redraw. This would cover all these imperfect things. But we need to understand that this is not a good solution! a) because of performance (not that important here) but b) This 'covering the faults' will only work on Redraw. But a bigger problem is the Redraw itself! Why do we need a redraw of a table when I click a Checkbox far away from that table??? If this wil get fixed in the future - then the unneccessary redraw will not happen anymore, so the function that now rely on will then break.

BerndK-Str:
@BDisp 

> You'd have to look at the original code to see how it actually works. One suggestion would be to make ColumnOffset and RowOffset private fields, _columnOffset and _rowOffset, and when you actually want to use their values, assign them to Viewport.X and Viewport.Y, respectively. To use their functionality, then create a public method, perhaps int GetViewportX(int column) and int GetViewportY(int row), to return the entire area including the boundaries. I think this might be necessary if a user wants to scroll programmatically.

You might imagin that I really looked at the old code, so at that time (without respecting a viewport) you needed these props to 'emulate' your own 'viewport'.
Now when using the built-in Viewport, there is absolutely no need to use these props any more, as you can see in my implementation, I don't need them at all.
The question I want to raise: 'Would a user be sad if there is no more RowOffset?"
Is the user fine with scrolling using the Viewport?
I think there is a method Viewport.Offset(0,1) which would scroll one line down (same as pressing the down arrow of the scrollbox)

tznind:
I don't mind loosing the offset properties or even the scroll indicators.

But you need to hook into `Update` to invalidate any cached data bout viewport size and/or styles.

BDisp:
The difference I'm talking about is the difference between these outputs.

```
tableView.ColumnOffset = 3; // column D
│D│E│F│
│4│5│6│

tableView.Viewport = tableView.Viewport with { X = 3 }; // column D ignoring the column boundaries
B│C│D│E
2│3│4│5
```

The second output is missing the left limit of column |B| and the right limit of column E|.

For me, it's indifferent how the output is presented when only the keys are used without the scroll bars.

But @tznind has already commented, which I appreciate, and I'll end this discussion on this specific subject.

One method you can use more reliably is OnSubViewsLaidOut, which is invoked after the layout is complete and before redrawing. You can save the last value of the Viewport and compare if there has been a change using the GetContentSize method. In this case, you can update any data based on whether there has actually been a change.

BerndK-Str:
@BDisp 
You are right. There is no doubt that Viewport.X=3 is different from ColumnOffset=3 - no discussion needed.
Exactly that is the reason why I wanted to bring back that Property back. The setter would then do the magical Viewport.X = Columns[3].X or something like this.
That has nothing to do with the keys or the scrollbar!

BerndK-Str:
Sorry have to quit now, will be back on Friday...

BDisp:
I receive the following error in both the TableEditor and TableViewTest scenarios when I scroll to the last row and maximize the terminal.

image

BerndK-Str:
I'm pretty happy with the current version.
RowOffset/ColumnOffset works now similar to before
Update will always respect changes
SmothScroll off seems to work now left and right

BerndK-Str:
Hi,
are there any showstoppers left?
What is needed to accept/close this PR?

BDisp:
> Hi, are there any showstoppers left? What is needed to accept/close this PR?

With the unit tests failing, it will be impossible to accept it, I think.

BerndK-Str:
@BDisp 
I fixed some unit tests that I had not on my list (TreeView in combo with TableView)
However there are 2 tests that still are not green:
- CanTabOutOfTableViewUsingCursor_Down : can't see why this should not work. I've tested it in scenario TableViewTest - works as expected!?
- ToKey_NumPadOperators_ReturnsExpectedKeyCode : I don't think that I changed something in this area

Please assist!

one other test did also fail (don't remember the nam), but worked when run again.

Terminal.Gui/Views/TableView/TableView.cs
```diff
@@ -38,10 +38,65 @@ public TableView ()
...
+        AddCommand (Command.Down,
+                    () =>
+                    {
+                        if (Table != null && SelectedRow >= Table.Rows - 1)
+                        {
+                            if (Viewport.Y < GetContentSize ().Height - Viewport.Height)
+                            {
+                                Viewport = Viewport with {Y = Viewport.Y + 1};
+
+                                return true;
```
BDisp: The CanTabOutOfTableViewUsingCursor_Down unit test fail because the Command.Down is always returning true even if it's already in the last line, returning always true, preventing propagating the Command.Down to the runnable top, which will focus the next view. The below code pass.

```cs
        AddCommand (Command.Down,
                    () =>
                    {
                        if (Table != null && SelectedRow > Table.Rows - 1)
                        {
                            if (Viewport.Y < GetContentSize ().Height - Viewport.Height)
                            {
                                Viewport = Viewport with { Y = Viewport.Y + 1 };

                                return true;
                            }

                            return false;
                        }

                        return ChangeSelectionByOffsetWithReturn (0, 1);
                    });
```

BDisp (changes requested):

BerndK-Str:
@BDisp 
Thanks for pointing to this problem. As mentioned I've tested this and has worked.
My code was correct, the root cause is the unit test, there was no viewport set.
Btw. I'm not a fan of focussing the next element using up/down. It does not feel good that you can't scroll to the end of the list be keep down pressed. Typically a user should use TAB to focus elements. But that is just my personal opinion, probably there are also reasons to do as is.

BDisp:
> @BDisp Thanks for pointing to this problem. As mentioned I've tested this and has worked. My code was correct, the root cause is the unit test, there was no viewport set.

Good find. So this method was always buggy because the tableView was always an empty size.

> Btw. I'm not a fan of focussing the next element using up/down. It does not feel good that you can't scroll to the end of the list be keep down pressed. Typically a user should use TAB to focus elements.

I'm also not a fan of this behavior when a view handles multiple items that can be navigated vertically and horizontally. I only agree with it in views that don't handle item navigation.
TextView is an exceptional case because we cannot use Tab or Shift+Tab to navigate because they are used for space tabbing. In Linux, it's also not possible to use Ctrl+Tab or Shift+-Ctrl+Tab because Ctrl is not captured when Tab is used, although there are other alternatives like Ctrl+N and Ctrl+P, with Tab and Shift+Tab being the default in general.

> But that is just my personal opinion, probably there are also reasons to do as is.

There was an intention to allow all views to focus on the next view if the down/right key was pressed and to focus on the previous view if the up/left key was pressed. The drawback is that when the user navigates quickly to select a specific item, other views are unintentionally focused, forcing them to go back again.

tig:
There are unresolved comments above. Is this really ready for my review?

BerndK-Str:
@tig 
I think we solved all the problems/discussions.
Some things related to the drawing pipeline / updates, are not specific for the table and should not be part of this PR.
Please point out the topics that are unclear?

Examples/UICatalog/Scenarios/ProcessTable.cs
```diff
@@ -47,8 +47,8 @@ public override void Main ()
 
     private void CreateProcessTable ()
     {
-        int ro = _tableView.RowOffset;
-        int co = _tableView.ColumnOffset;
+        //int ro = _tableView.RowOffset;
```
tig: Remove commented out code.

Terminal.Gui/Views/FileDialogs/FileDialog.cs
tig: Yes it is.

Terminal.Gui/Views/TableView/TableStyle.cs
```diff
@@ -67,6 +67,7 @@ public class TableStyle
     ///     True to render a arrows on the right/left of the table when there are more column(s) that can be scrolled to.
     ///     Requires <see cref="ShowHorizontalHeaderUnderline"/> to be true. Defaults to true
     /// </summary>
+#warning a candidate to remove
```
tig: This needs to be resolved.

Terminal.Gui/Views/TableView/TableView.cs
tig: Just create private functions for each command. The one above would be `private bool? HandleRight(ICommandContext? ctx)`.

tig (changes requested):
There are several requests in the code review to be addressed. thanks. This is awesome work!

BerndK-Str:
What about 'AutoShow'? Is it removed completely, is there any alternative?

tig:
> What about 'AutoShow'? Is it removed completely, is there any alternative?

See https://gui-cs.github.io/Terminal.Gui/api/Terminal.Gui.ViewBase.ViewportSettingsFlags.html and `HasVerticalScrollBar`

BerndK-Str:
@tig Hopefully fixed all the latest findings.

BerndK:
I don‘t feel responsible for the failing test?!

tig:
Yep. That's a known intermittent failure.

BerndK-Str:
@tig Can I do something?
Do you expect me to resolve the 3 new discussions from Sunday (one of 3 has no resolve button? 
However I did the changes already.
Sorry I am not really used to that workflow. I would expect you to resolve that after review the changes!?

Examples/UICatalog/Scenarios/ProcessTable.cs
tig: @BerndK-Str this needs to be addressed. Thanks.

Terminal.Gui/Views/TableView/TableStyle.cs
tig: @BerndK-Str this needs to be addressed. Thanks.

Terminal.Gui/Views/TableView/TableView.cs
tig: @BerndK-Str this needs to be addressed. Thanks.

tig:
I just directly tagged you in the review comments where things need to be addressed. Thanks.

Examples/UICatalog/Scenarios/ProcessTable.cs
BerndK-Str: fixed in commit of sunday , 7bb436a0

Terminal.Gui/Views/TableView/TableStyle.cs
BerndK-Str: fixed in commit of sunday , https://github.com/gui-cs/Terminal.Gui/commit/7bb436a0418cd88ec4a4f7e961d017093e71a1c2

BerndK-Str:
@tig 
From my understanding I already fixed all open topics in commit of sunday , 7bb436a
However I did not resolve or comment the requests, I thought you would do it. Now I tried to close all requests.

Examples/UICatalog/Scenarios/TableViewTest.cs
tig: @BerndK-Str please remove this scenario per tznind

tig (changes requested):
Unless you really disagree with @tznind please remove the added scenario.

BerndK-Str:
I want to use this (TableViewTest) for debugging and testing (especially as long as the mouse does not work on the menu). There are already 5 TableView testviews registered. So I can't see why another one is a no go. However, feel free to remove it now or later, I personally do not want to delete that now.

tig:
Great work!!!

tig:
Odd windows parallel test failure I've not seen before 
Rerunning to see if it passes. Can't reproduce locally.

Issue #4473, opened 2025-12-09
`Menu`, `Shortcut`, etc... are broken - Introduce `Command` Bubbling


tig:
## Summary

We currently have what I call command propagation for `Command.Accept` in place today where `RaiseAccepting` special cases any View with `IsDefaultView` (e.g. `Button`) and if not handled, propagates the command up the View hierarchy until a `SuperView` handles the command. This is the root of why the `Accepting` event handlers must indicate with `e.Handled = true` if they handled the command. The reason we have this is to support both default handling (e.g. `Button.IsDefault`) and to allow compound views such as `Shortcut` to be able to handle and override the behaviors of what subviews do when the user "accepts" them.

The current code in `View.RaiseAccepting` (in View.Command.cs):

```cs
        // Accept is a special case where if the event is not handled above, the event is
        //  - Invoked on any peer-View with IsDefault == true
        //  - bubbled up the SuperView hierarchy.
        if (!args.Handled)
        {
            // If there's an IsDefault peer view in SubViews, try it
            View? isDefaultView = SuperView?.InternalSubViews.FirstOrDefault (v => v is Button { IsDefault: true });

            if (isDefaultView != this && isDefaultView is Button { IsDefault: true } button)
            {
                Logging.Debug ($"{Title} ({ctx?.Source?.Title}) - InvokeCommand on Default View ({isDefaultView.Title})");
                bool? handled = isDefaultView.InvokeCommand (Command.Accept, ctx);

                if (handled == true)
                {
                    return true;
                }
            }

            if (SuperView is { })
            {
                return SuperView?.InvokeCommand (Command.Accept, ctx);
            }
        }
```

Note: We just renamed `Select` to `Activate` because:

- Representing both state changes (e.g., toggling a `CheckBox`, selecting a `ListView` item) and preparatory actions (e.g., focusing a `Button`, navigating a `MenuItemv2`).
- Distinguishing clearly from `Accepting`, which confirms actions (e.g., executing a menu command, submitting a dialog).
- Supporting a targeted propagation model with `PropagateActivating`, enabling superviews like `MenuBar` to opt-in to subview `Activating` events without coupling subviews to superview details.

### Problems

- This code one-offs `Command.Accept`
- `Command.Accept` is actually not the right command for this in many cases. Support for `Activating` (previously `Selecting`) is required too.
- It puts the onus of specifying behavior on subviews, requiring them to know their superview behavior. 

[...]

PR #4695, opened 2026-02-06
Fixes a crash in TableEditor's OpenTreeExample.

iosdevzone:
## Fixes

On Windows, the OpenTreeExample in the TableEditor scenario crashes with an unhandled `IOException` if a mapped network drive is offline.

## Proposed Changes/Todos

Modifying the loop that adds the available drives on the system to filter out drives that are not in the ready state solves the problem. (See attached commit)

Also removed unused `seen` variable.

## Pull Request checklist:

  - Not applicable. There is no existing issue for this crash and writing one for such a simple fix seems unnecessary. 
    - Not applicable. No API changes made 
  * Fix was tested on Windows to verify it fixes the crash
  * Also tested on macOS to verify that it did not change behavior on that system.

### DIFF
--- a/Examples/UICatalog/Scenarios/EditorsAndHelpers/AllViewsView.cs
+++ b/Examples/UICatalog/Scenarios/EditorsAndHelpers/AllViewsView.cs
@@ -3,21 +3,20 @@ namespace UICatalog.Scenarios;
 
 public class AllViewsView : View
 {
-    private const int MAX_VIEW_FRAME_HEIGHT = 25;
+    private const int MAX_VIEW_FRAME_HEIGHT = 30;
 
     public AllViewsView ()
     {
         CanFocus = true;
         BorderStyle = LineStyle.Heavy;
         Arrangement = ViewArrangement.Resizable;
-        HorizontalScrollBar.AutoShow = false;
-        VerticalScrollBar.AutoShow = true;
+        ViewportSettings |= ViewportSettingsFlags.HasVerticalScrollBar;
 
-        SubViewsLaidOut += (sender, _) =>
+        ViewportChanged += (sender, _) =>
                            {
                                if (sender is View sendingView)
                                {
-                                  sendingView.SetContentSize (new Size (sendingView.Viewport.Width, sendingView.GetHeightRequiredForSubViews ()));
+                                   sendingView.SetContentSize (new Size (sendingView.Viewport.Width, sendingView.GetHeightRequiredForSubViews ()));
                                }
                            };
 
@@ -26,8 +25,7 @@ public class AllViewsView : View
         AddCommand (Command.PageUp, () => ScrollVertical (-SubViews.OfType<FrameView> ().First ().Frame.Height));
         AddCommand (Command.PageDown, () => ScrollVertical (SubViews.OfType<FrameView> ().First ().Frame.Height));
 
-        AddCommand (
-                    Command.Start,
+        AddCommand (Command.Start,
                     () =>
                     {
                         Viewport = Viewport with { Y = 0 };
@@ -35,8 +33,7 @@ public class AllViewsView : View
                         return true;
                     });
 
-        AddCommand (
-                    Command.End,
+        AddCommand (Command.End,
                     () =>
                     {
                         Viewport = Viewport with { Y = GetContentSize ().Height };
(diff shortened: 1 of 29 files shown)


tznind:
Nice, does this error occur in FileDialog too? Its part of main lib so would be even more important fix

iosdevzone:
@tznind I'll try to take a look at that.

BDisp:
Perhaps searching the entire solution for "Environment.GetLogicalDrives" and making the changes accordingly.

iosdevzone:
@BDisp yeah, looking at that now. Looks like `GetLogicalDrives` and `DriveInfo` are only used in the examples. The `TreeViewFileSystem` scenario does crash in a similar manner. Couldn't get the `FileDialog` example to crash yet.

iosdevzone:
Whoops! My last comment thinking that all the concrete file system code was limited to Examples is incorrect.

There is some code in `DefaultTreeRootGetter` in `FileDialogStype.cs` that uses `GetLogicalDrives`. 

I'm not sure if this would lead to a similar crash. Perhaps someone more familiar with your codebase could take a peek?

BDisp:
> Whoops! My last comment thinking that all the concrete file system code was limited to Examples is incorrect.
> 
> There is some code in `DefaultTreeRootGetter` in `FileDialogStype.cs` that uses `GetLogicalDrives`.
> 
> I'm not sure if this would lead to a similar crash. Perhaps someone more familiar with your codebase could take a peek?

I think it's only worth searching for "Environment.GetLogicalDrives". I can't reproduce the exception because I don't have the same situation as you. What's strange is that the FileDialogExamples.cs scenario doesn't throw the same exception, perhaps because it's not initializing a DirectoryInfo. In FileDialogStyle, "Environment.GetLogicalDrives" is only in the XML doc and it's not being used directly, but rather through `_fileSystem.Directory.GetLogicalDrives()`.

iosdevzone:
@tznind @BDisp Sorry for the delay in getting back to this, those pesky day jobs have a habit of getting in the way of open source work ;-) 

Seaching for `DriveInfo` is useful because if the loop was not filtered the behavior could end up being the same as using `GetLogicalDrives`. As it turned out, I did not find any places where this was true.

It looks like `FileDialog` is OK because it catches exceptions in its tree builder (see [FileSystemTreeBuidler.cs:54-63](https://github.com/gui-cs/Terminal.Gui/blob/eba9bb63c85f2d22089ed485648e716648b8d9cd/Terminal.Gui/FileServices/FileSystemTreeBuilder.cs#L50-L59)). That being said, the tree display within `FileDialog` seems buggy as clicking on the tree display causes the file dialog to exit, but that is outside the scope of this PR. (It happens on my Mac too which does not have the same issue with network disks. If I get the time, I'll create a separate Issue for this.)

That leaves the `TreeViewFileSystem` scenario. I'll fix that one and add it to this PR.

iosdevzone:
I've added a fix for the crash in the File System Explorer example. In this case, I chose to leave the offline disks in the display and handle the exception if and when the user chooses one of the offline disks.

If this solution is not to your liking, you could just apply filtering to where the drive list is initialized (see [TreeViewFileSystem.cs:385](https://github.com/gui-cs/Terminal.Gui/blob/eba9bb63c85f2d22089ed485648e716648b8d9cd/Examples/UICatalog/Scenarios/TreeViewFileSystem.cs#L385)) from: 
```csharp
IEnumerable<IDirectoryInfo> rootDirs = DriveInfo.GetDrives ().Select (d => fs.DirectoryInfo.New (d.RootDirectory.FullName));
```
to:
```csharp
IEnumerable<IDirectoryInfo> rootDirs = DriveInfo.GetDrives ()
                                        .Where (di => di is not { DriveType: DriveType.Network, IsReady: false })
                                        .Select (d => fs.DirectoryInfo.New (d.RootDirectory.FullName));
```

At this point, unless there are specific changes you would like me to make, I am not intending to do any more work on this PR.

tig:
The treeview issue is same as #4473 and will be fixed as part of the fix for that.

iosdevzone:
> The treeview issue is same as #4473 and will be fixed as part of the fix for that.

@tig Thank you for letting me know. I was thinking that I might look into; this saves me wasting my time!

tig (approved):

PR #4709, opened 2026-02-09
Fixes #4689. Fragile DimAutoStyle.Content unit test

### LINKED ISSUE #4689: Fragile DimAutoStyle.Content unit test
What exactly is the following unit test testing? It seems fragile to me. The intention was to test with a subview, but in reality it is ignored.

```cs
    [Fact]
    public void DimAutoStyle_Content_IgnoresSubViews_When_ContentSize_Is_Set ()
    {
        var view = new View ();

        var subview = new View
        {
            Frame = new (50, 50, 1, 1)
        };
        view.SetContentSize (new (10, 5));

        Dim dim = Auto (DimAutoStyle.Content);

        int calculatedWidth = dim.Calculate (0, 100, view, Dimension.Width);

        Assert.Equal (10, calculatedWidth);
    }
```

BDisp:
## Fixes

- Fixes #4689

## Proposed Changes/Todos

- [x] Adapt the test with current API behavior
- [x] Add more Pos unit tests within the Dialog with containers.

## Pull Request checklist:


Tests/UnitTestsParallelizable/Views/DialogTests.cs
```diff
@@ -1176,6 +1176,360 @@ public void Dialog_Draws_EnableForDesign ()
...
+
+        using Dialog dialog = new ();
+
+        dialog.Driver = driver;
+        dialog.X = 0;
+        dialog.Y = 0;
+        dialog.BorderStyle = LineStyle.Single;
+        dialog.ShadowStyle = ShadowStyle.None;
+        dialog.Title = "Dialog";
+        dialog.AddButton (new Button { Text = "Cancel" });
```
tig: I recommend you remove the dialog buttons from these
Tests. They do not have anything to do with what's being tested and just make the tests more fragile to changes we may make to Button or Dialog in the future.

Tests/UnitTestsParallelizable/Views/DialogTests.cs
```diff
@@ -1176,6 +1176,360 @@ public void Dialog_Draws_EnableForDesign ()
         DriverAssert.AssertDriverContentsAre (expected, output, driver);
     }
 
+    [Theory]
+    [MemberData (nameof (PosData), MemberType = typeof (DialogTests))]
+    public void Dialog_Draws_SubView_With_SubViews_WithDifferentPosTypes (Pos pos, string expected)
```
tig, line 1181: These tests, without the dialog buttons test nothing specific to Dialog. 

If there are not equivalent View.LayoutTests for these use cases they should be moved there. If there are already tests for this, they are not needed.

tig (changes requested):
The Dialog tests do not actually test anything specific to Dialog. They are just view layout tests. 

I think they should be renamed and moved (and simplified to not use any View subclasses).

BDisp:
@tig, can you tell me if the `GetContentSize()` method should actually return a size smaller than the Viewport size? I have serious doubts about this and would greatly appreciate clarification. Thank you.

Here's why:

Conceptually:

- The viewport is the visible window into the content
- The content size is what you're looking at through that window
- If content is smaller than the viewport, it would just sit inside the viewport with empty space around it—there's no need to track a separate "content size" in that case

Practically:

- When contentSize < viewport: No scrolling needed, content fits entirely within view
- When contentSize ≥ viewport: Scrolling is needed to see all content
- When contentSize == viewport: Content exactly fills the viewport (no scrolling)

Considering that the _contentSize is smaller than the Viewport will result in incorrect calculations when obtaining the true size of a view, passing incorrect size parameters to Pos and Dim.

tig:
> @tig, can you tell me if the `GetContentSize()` method should actually return a size smaller than the Viewport size? I have serious doubts about this and would greatly appreciate clarification. Thank you.
> 
> Here's why:
> 
> Conceptually:
> 
> * The viewport is the visible window into the content
> * The content size is what you're looking at through that window
> * If content is smaller than the viewport, it would just sit inside the viewport with empty space around it—there's no need to track a separate "content size" in that case
> 
> Practically:
> 
> * When contentSize < viewport: No scrolling needed, content fits entirely within view
> * When contentSize ≥ viewport: Scrolling is needed to see all content
> * When contentSize == viewport: Content exactly fills the viewport (no scrolling)
> 
> Considering that the _contentSize is smaller than the Viewport will result in incorrect calculations when obtaining the true size of a view, passing incorrect size parameters to Pos and Dim.

`GetContentSize` will return either `Viewport.Size` or whatever was set by `SetContentSize`. 

```cs
    public Size GetContentSize () => _contentSize ?? Viewport.Size;
```

Layout is completely decoupled from Viewport. If there is any code in the Layout engine that sets or gets `Viewport` that is a bug. 

> Considering that the _contentSize is smaller than the Viewport will result in incorrect calculations when obtaining the true size of a view, passing incorrect size parameters to Pos and Dim.

I believe that is correct. IOW, A view should use set Height or Width to DimAuto it is doing something that causes `_contentSize < Viewport.Size`.

Tests/UnitTestsParallelizable/Views/DialogTests.cs
BDisp: I already removed the buttons and I want to keep the tests because my intention is to test specially with Dialog.
Done in https://github.com/gui-cs/Terminal.Gui/pull/4709/commits/1b30fed533a8b554c85744fe9ad0197b586ec470.

Tests/UnitTestsParallelizable/Views/DialogTests.cs
BDisp: I'm sorry to disagree with your opinion. The unit tests in question are not for testing Dim.Auto in a simple view, which is probably working correctly, but intentionally to test it in a Dialog. As you can see, you also included some tests in DialogTests.cs to confirm some doubts. I agree and will remove the buttons, as they are not necessary for this and will serve to see if the automatic resizing is working well even without the buttons. Therefore, I insist on keeping the unit tests using the Dialog as the SuperView. I hope you understand my choice.

I also opened the issue #4710 reporting a bug that was primarily affecting `Pos.Center`, and the fix is ​​available at https://github.com/gui-cs/Terminal.Gui/pull/4709/commits/0dad0d0da63ba5326678c711168c0350e5363bc3.

BDisp:
> Layout is completely decoupled from Viewport. If there is any code in the Layout engine that sets or gets `Viewport` that is a bug.

I completely agree, and it wasn't even my intention to do that.

> I believe that is correct. IOW, A view should use set Height or Width to DimAuto it is doing something that causes `_contentSize < Viewport.Size`.

Thanks for clarifying that. The content size can indeed be smaller than the viewport size if a view requires it. The problem I was having was due to issue #4710, which has since been fixed in commit https://github.com/gui-cs/Terminal.Gui/pull/4709/commits/0dad0d0da63ba5326678c711168c0350e5363bc3. Thanks.

BDisp:
> The Dialog tests do not actually test anything specific to Dialog. They are just view layout tests.
> 
> I think they should be renamed and moved (and simplified to not use any View subclasses).

The intention was to purposefully test with Dialog and a container that uses Dim.Auto, which could be a class derived from View, reusable in other views, where I was encountering some inaccuracies. With a simple view, there are already unit tests that I know work well.

Issue #4522, opened 2025-12-30
`NeedsLayout` is buggy


tig:
The implementation and design of `View.NeedsLayout` and related APIs (e.g. `SetNeedsLayout`) is buggy:

On one hand, there are places where the layout (and Content area) code creates false-positives, setting `NeedsLayout` to true, when, in fact, no layout is needed. E.g. See `PosDimSet`.

On the other hand, there are places where calls to `Layout` have been added to work around cases where the next Application iteration *should* cause a layout, but doesn't; leading someone (often me) to add a `Layout` call. Some of these are marked with `// BUGBUG:`s but not all. We should view ANY call to `Layout` directly as a bug.

We should revamp this to be more deterministic and clear. One place to start is to remove the setter for `NeedsLayout` completely. This comment in `View.Layout.cs` is bogus:

```cs
 #region NeedsLayout

 // We expose no setter for this to ensure that the ONLY place it's changed is in SetNeedsLayout

 // BUGBUG: The above statement is misleading. There are still cases internally where this property
 // BUGBUG: is being set directly without calling SetNeedsLayout. We should remove the setter completely.

 /// <summary>
 ///     Indicates the View's Frame or the layout of the View's subviews (including Adornments) have
 ///     changed since the last time the View was laid out.
 /// </summary>
 /// <remarks>
 ///     <para>
 ///         Used to prevent <see cref="Layout()"/> from needlessly computing
 ///         layout.
 ///     </para>
 /// </remarks>
 /// <value>
 ///     <see langword="true"/> if layout is needed.
 /// </value>
 public bool NeedsLayout { get; private set; } = true;
```

Any code that sets `NeedsLayout` directly should be viewed as suspicious:

<img width="777" height="180" alt="Image" src="https://github.com/user-attachments/assets/a57bdd78-485a-4461-8d90-c4008b6fac33" />

Here's a analysis I had the AIs create. I have not reviewed it for accuracy, so treat it with grains of salt:

# NeedsLayout System Analysis - Latent Bugs and Issues

**Date:** 2025-01-XX  
**Project:** Terminal.Gui  
**Scope:** View.Layout.cs NeedsLayout mechanism and related APIs

---

## Table of Contents

1. [Executive Summary](#executive-summary)
2. [NeedsLayout Flow Overview](#needslayout-flow-overview)
3. [Latent Bugs Identified](#latent-bugs-identified)
4. [Test Coverage Analysis](#test-coverage-analysis)
[...]

PR #4723, opened 2026-02-18
Fixes #4726 etc... - Addresses several issues with `TextView`

### LINKED ISSUE #4722: TextView: Why doesn't TextView behave the same way as other editors with Key.Tab.WithShift?
Notepad++, all versions of Visual Studio, VSCode, etc., contain the functionality to remove the '\t' character when pressing Shift+Tab, if the character '\t' exists in the current cursor location. TextView also previously had this feature, and now I discover that it has been removed and that pressing Shift+Tab moves the focus to the previous view. Many users who are used to this functionality will notice a different behavior than they are used to. I know that pressing Key.Backspace has the same effect. I also know that Windows Notepad assumes a Key.Tab.WithShift as just Key.Tab, which is ridiculous.

Of course, this functionality would only be valid if `TabKeyAddsTab` is true, otherwise, it will have the default behavior of focusing on the next view (Tab) or the previous view (Shift-Tab).

I would like the previous behavior to be restored, and if anyone else supports this, please react. Thanks.

### LINKED ISSUE #4724: TextView: Command.RightEnd/Right/Left is not drawing the correct cursor position in the TextView viewport.
In the "Text Input Controls" scenario, pressing the End key in the TextView does not display the current cursor position.

### LINKED ISSUE #4725: TextView: Inserting Tab into TextView when WordWrap & TabKeyAddsTab Exhibits Odd Behavior and Crashes
## To Reproduce

1. Run UICatalog => Text Input Controls scenario.
2. Check "Word Wrap" and "Tab Key Adds Tab"

Behavior 1:
- Press 'Ctrl+End', 'Enter', 'Tab'
- Immediate crash.

Behavior 2:
- Press 'Ctrl+End', 'Tab', 'Right', 'Right'
- Cursor moves one column at a time as if there are spaces.
- Pressing 'Backspace' or 'Delete' anywhere in these "spaces" causes immediate crash.

Image

Behavior 3:
- Press 'Tab' in the middle of a line of text.
- 4 spaces are inserted.
- Pressing 'Delete' or 'Backspace' removes incorrect characters from the text.

Image

All functions work correctly if Word Wrap is false

### OS Information

This repros on Windows and Linux.

### LINKED ISSUE #4726: TextView: Master Issue - Inserting Tabs in TextView Causes Mouse to Incorrectly Position Cursor
## Describe the bug
When clicking to reposition the cursor, the cursor position is offset relative to the number of inserted tabs if clicking *after* an inserted tab.

## To Reproduce

1. Run the Text Input Controls scenario.
2. Insert a tab into the TextView.
3. Use the mouse to click a position after the tabstop.

Expected behavior:
Cursor should move to clicked position.

Actual behavior:
Cursor moves to a position offset to the left of the clicked position.
The amount offset seems to be related to the number of tabstops inserted.

Image

### OS Information

Repros on Windows 11 and Debian 13

### LINKED ISSUE #4727: TextView: Command.Down/Up in TextView should have the default behavior of focusing on the next and previous view.
Regardless of whether the TabKeyAddsTab property is true or false.

### LINKED ISSUE #4819: TextView: Should pressing CursorLeft with the cursor at location 1 in the viewport display the glyph on the left in TextField|TextView?
This would allow us to perceive whether the beginning of the text has already been reached or not. `CursorRight` already advances 1 column when the cursor is in the second-to-last column of the viewport, and this would make it consistent. The IDE indents half a glyph, which is possible because it deals with pixels and not columns. Notepad++ indents one word. My suggestion would be to indent only 1 column.
I would appreciate your opinion on whether you agree or disagree, as I will only move forward with this idea if there is consensus.

BDisp:
## Fixes

- Fixes #4726

And all these sub-issues:

- Fixes #4722
- Fixes #4724
- Fixes #4725
- Fixes #4727
- Fixes #4819

## Proposed Changes/Todos

- [x] Provides Shift+Tab functionality to remove the '\t' character from the current cursor location as before.

## Pull Request checklist:


Terminal.Gui/Views/TextInput/TextView/TextView.Keyboard.cs
```diff
@@ -47,16 +47,13 @@ public bool EnterKeyAddsLine
 
     /// <summary>
```
YourRobotOverlord, line 48: I know this is nit-picky, but I think this can be clarified;
Setting this to false doesn't necessarily mean it will get ignored, i.e., the developer decides to handle the key, or the SuperView uses it for navigation.
Setting this does not necessarily have any direct effect on navigation, i.e., the developer has overridden NextTabKey/PreviousTabKey, so I don't think it should be explicitly stated here.

For example:
```csharp
    /// <summary>
    ///     Gets or sets whether <see cref="Key.Tab"/> inserts a tab character (<c>\t</c>) into the <see cref="TextView"/>.
    /// </summary>
    /// <remarks>
    /// <para>
    /// Default: <see langword="true"/>
    /// </para>
    /// <para>
    /// If <see cref="TabKeyAddsTab"/> is set to <see langword="true"/>, and the user presses <see cref="Key.Tab"/>,
    ///     a tab character is inserted at the current cursor position; if Shift+<see cref="Key.Tab"/> is pressed,
    ///     a tab character is removed from the current cursor position if one exists.
    /// </para>
    /// <para>
    /// If <see cref="TabKeyAddsTab"/> is set to <see langword="false"/>, the <see cref="IKeyboard.KeyDown"/> event
    ///     is bubbled up to the SuperView.
    /// </para>
    /// </remarks>
```

BDisp: I also thought a lot about the behavior if Shift+Tab wasn't handled, if TabKeyAddsTab is true. I thought about returning false so that the superview would use it if necessary. But I also thought about the possibility of the user holding down the Shift+Tab keys to remove all existing tabs. If all tabs were removed while still holding down the Shift+Tab keys, the focus would go to the previous views, whose behavior would also not be desirable for the user. So I decided to return true for these cases. If TabKeyAddsTab is false, this problem would not exist because Tab and Shift+Tab would have the default behavior.
Therefore, I think it's debatable and I will make the necessary changes if desired.

BDisp: Sorry @YourRobotOverlord. I thought you wanted the Shift + Tab functionality even when TabKeyAddsTab was true and unhandled. After reading more carefully, I agree with your suggestion, except for including "IKeyboard.KeyDown" in the XML doc. But I agree that pressing the down arrow key on the last line and the up arrow key on the first line could focus the next and previous views, regardless of whether TabKeyAddsTab is true or false, thus having the same behavior as most views.
Do you agree with the commit https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/63b7a4c7c3f2eab58ac4e0c8ba5839fc96330cc6?

BDisp: The desired behavior for the Down/Up keys when the cursor is on the last/first line to move the focus to the next/previous view is in the commit https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/a5a7d22432fe2ded2ef7b9f6667c21d38c4434b4.

YourRobotOverlord: > Do you agree with the commit 63b7a4c?

I do.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
```diff
@@ -2724,5 +2734,81 @@ public void Command_Up_MovesFocusToPreviousView_WhenCursorIsAtFirstLine_TabKeyAd
...
+        runnable.Add (textView);
+
+        app.Begin (runnable);
+
+        Assert.Equal (4, textView.TabWidth);
+
+        textView.SetRelativeLayout (new Size (100, 100));
+
+        // Click at position "0" stay in first
+        Assert.True (textView.NewMouseEvent (new Mouse { Position = new Point (1, 0), Flags = MouseFlags.LeftButtonClicked }));
```
tig: Please refactor this test to be multiple tests that each test the smallest unit of functionality possible. 

- Testing that NewMouseEvent returns true is not the point of this test. Do not assert for that here. 
- Each of these // Click At ... sections tests something different and should be in individual unit tests or Theory should be used.

I ask this, because it becomes very hard to debug things in tests that are not super-focused... I think we've learned this the hard way and I'd really like us to try to avoid it for new tests. 

Thanks.

tig (changes requested):
Great stuff. It is my fault this is broken, and I feel bad about that and am appreciative of you taking it on. 

See my requests.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
BDisp: Replaced with `Theory` test with two lines which cover all possible situations based on text in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/d0d9cf079231ebdc29919dc6a7ed95a7a95ada20.

Terminal.Gui/Views/TextInput/TextView/TextView.cs
```diff
@@ -198,7 +198,7 @@ public void PositionCursor ()
                         // Calculate columns to next tab stop
                         // Tab stops are at multiples of TabWidth (0, 4, 8, 12, ...)
                         // If we're at visual column col, advance to next tab stop
-                        cols = TabWidth - col % TabWidth;
+                        cols = TabWidth;
```
tig: This was very intentional. Almost all pure text editors treat "tabs" as tab-stops, like a type writer. TabWidth defines how many columns each tab-stop takes. If TabWidth is 4, the first char after a "tab" will go at column 4, 8, 16, ....  That's what this code does.

TextView didnt used to work this way. In my recent PR I refactored it to do so. It was a lot of work and I believe the right model for TUI apps. Please change this behavior back in this PR. Thanks.

BDisp: I'm not here to ruin anyone's work. If I changed it, it's because it doesn't work correctly, as it did before. You understand that your solution always calculates correctly, especially when just pressing Tab in the first column. In the "Text input Controls" scenario, go to the last column, press TAB, and you will see, as in the image, that it will calculate (4-(57%4))=4-1=3, instead of 4, which is effectively its width. At least you could run unit tests to confirm if your solution is correct. If this behavior was intentional, I don't even know how it's possible to corroborate this solution without actually having unit tests that confirm if this behavior is correct, not just by inserting Tab at the beginning of the text, but randomly anywhere. Do you really want me to make this change? If so, then you will have to fix the failing unit tests, because I don't know how to patch them.

image

BDisp: Sorry @tig. After @YourRobotOverlord's explanation, I finally understood how the Tab logic works. I'll try to correct everything that's wrong.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
```diff
@@ -2426,5 +2428,406 @@ public void Command_HotKey_SetsFocus ()
...
+        // Simulate a key event with Used = false
+        Assert.True (textView.NewKeyDownEvent (Key.A));
+
+        // Text should be overwritten at the insertion point with the typed key
+        Assert.Equal (expected, textView.Text);
+        Assert.Equal (new Point (col + 1, 0), textView.InsertionPoint);
+    }
+
+    [Fact]
+    public void Key_Insert_InvokeCommand_ToggleOverwrite_And_Toggles_Used_Property ()
```
tig, line 2669: These tests are mis-named. They do not use InvokeCommand but NewKeyDownEvent. My recommendation is to use InvokeCommand for all these tests and rename them to be `Command_<cmdname>_...`. This is the pattern I've been using elsewhere. For tests that actually test KeyDown fn, the pattern is `KeyDown_<key>_...`. 

These tests do not test key mapping, but what commands do, and thus should use InvokeCommand, IMO.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
```diff
@@ -2426,5 +2428,406 @@ public void Command_HotKey_SetsFocus ()
...
+
+        Assert.True (textView.NewKeyDownEvent (Key.InsertChar));
+        Assert.False (textView.Used);
+
+        Assert.True (textView.NewKeyDownEvent (Key.InsertChar));
+        Assert.True (textView.Used);
+    }
+
+    [Fact]
+    public void Command_RightEnd_Draws_Cursor_At_End_Of_Line ()
```
tig: Same. Either make the test use InvokeCommand and not NewKeyDownEvent or rename it.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
```diff
@@ -2426,5 +2428,406 @@ public void Command_HotKey_SetsFocus ()
...
+                       │        │
+                       │        │
+                       └────────┘
+                       """;
+
+        DriverAssert.AssertDriverContentsAre (expected, output, app.Driver);
+    }
+
+    [Fact]
+    public void Command_Right_Draws_Cursor_At_Start_Of_NextLine ()
```
tig: Same.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
```diff
@@ -2426,5 +2428,406 @@ public void Command_HotKey_SetsFocus ()
...
+                       ┌────────┐
+                       │2 Testin│
+                       └────────┘
+                       """;
+
+        DriverAssert.AssertDriverContentsAre (expected, output, app.Driver);
+    }
+
+    [Fact]
+    public void Command_Left_Draws_Cursor_At_End_Of_PreviousLine ()
```
tig: Same.

tig:
I accidentally marked "approved", but meant "request changes". Thanks.

Terminal.Gui/Views/TextInput/TextView/TextView.Movement.cs
```diff
@@ -33,10 +41,13 @@ public bool MoveHome ()
...
     /// <param name="idx">
-    ///     Row that should be displayed at the top or Column that should be displayed at the left, if the value
+    ///     The row that should be displayed at the top or the column that should be displayed at the left, if the value
     ///     is negative it will be reset to zero
     /// </param>
     /// <param name="isRow">If true (default) the <paramref name="idx"/> is a row, column otherwise.</param>
+    /// <remarks>
+    ///     The <see cref="CurrentRow"/> and <see cref="CurrentColumn"/> will not be changed by this method.
+    /// </remarks>
     public void ScrollTo (int idx, bool isRow = true)
```
BDisp: @tig please tell me if you agree with this change where only the `Viewport` will be modified, excluding `CurrentRow` and `CurrentColumn`. It only scrolls and does not modify its position at the cursor level.

BDisp: Another option is to use the `Point` type, allowing to scroll to the desired row and column at once. What do you think?

tig: This is a bizzare API for sure as defined.

What usages of it exist within the Solution today? 

If there are none, I suggest you just simplify and nuke the API from orbit. If someone ever comes back and says they really need it, we can add it then.

BDisp: So I don't know why you didn't think of this when you redesigned the `TextView`. I'll leave it as it is and not waste any more time, because I need to focus on solving the Tab functionality issue.

BDisp: If the `InsertionPoint` property truly performs the same procedure as the `Scroll` method, then I will remove it and check if there are credible unit tests to test the same effect. Is that okay with you?

BDisp:
Normally, when the CI test execution is aborted on macOS, the following message appears. Could it be related to the ANSI escape sequence calls that are in the log? I'm not sure.


Blame: Dumping 8492 - dotnet
┌──────────────────────────────────────────────────────────────────────────────┐│                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              │└──────────────────────────────────────────────────────────────────────────────┘┌──────────────────────────────────────────────────────────────────────────────┐│                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              ││                                                                              │└──────────────────────────────────────────────────────────────────────────────┘25l                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1049l1049h25l1003h1015h1006h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1003l1015l1006l1049l25h1049h25l1003h1015h1006h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1003l1015l1006l1049l25h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1049l1049l1049h25l1003h1015h1006h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1003l1015l1006l1049l25h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1049l1049l1049h25l1003h1015h1006h1049h25l1003h1015h1006h1003l1015l1006l1049l25h1003l1015l1006l1049l25h1049l1049h25l1003h1015h1006h1003l1015l1006l1049l25h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1049l1049h25l1003h1015h1006h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1003l1015l1006l1049l25h1049h25l1003h1015h1006h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1003l1015l1006l1049l25h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1049l1049l1049h25l1003h1015h1006h1003l1015l1006l1049l25h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1049l1049h25l1003h1015h1006h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1003l1015l1006l1049l25h1049h25l1003h1015h1006h1003l1015l1006l1049l25h1049l1049l                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1049h25l1003h1015h1006h                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                1003l1015l1006l1049l25hreatedump] Gathering state for process 8492 
[createdump] Writing full dump to file /Users/runner/work/Terminal.Gui/Terminal.Gui/Tests/UnitTestsParallelizable/logs/UnitTestsParallelizable/macOS/dotnet_8492_20260227T184200_hangdump.dmp
[createdump] Written 6496687000 bytes (396526 pages) to core file
[createdump] Target process is alive
[createdump] Dump successfully written in 35860ms
The active test run was aborted. Reason: Test host process crashed
Data collector 'Blame' message: VSTEST_DUMP_PATH is specified. Dump files will be saved in: /Users/runner/work/Terminal.Gui/Terminal.Gui/Tests/UnitTestsParallelizable/logs/UnitTestsParallelizable/macOS/, and won't be added to attachments..
Data collector 'Blame' message: The specified inactivity time of 4 minutes has elapsed. Collecting hang dumps from testhost and its child processes.
Data collector 'Blame' message: VSTEST_DUMP_PATH is specified. Dump files will be saved in: /Users/runner/work/Terminal.Gui/Terminal.Gui/Tests/UnitTestsParallelizable/logs/UnitTestsParallelizable/macOS/, and won't be added to attachments..

Test Run Aborted.
Total tests: Unknown
     Passed: 2520
    Skipped: 2
 Total time: 5.8146 Minutes

The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
The test running when the crash occurred: 
TextTests.TextFormatterJustificationTests.Draw_Text_Justification
TextTests.TextFormatterTests.Draw_Horizontal_Justified

This test may, or may not be the source of the crash.
     1>Project "/Users/runner/work/Terminal.Gui/Terminal.Gui/Tests/UnitTestsParallelizable/UnitTests.Parallelizable.csproj" on node 1 (VSTest target(s)).
     1>_VSTestConsole:
         MSB4181: The "VSTestTask" task returned false but did not log an error.
     1>Done Building Project "/Users/runner/work/Terminal.Gui/Terminal.Gui/Tests/UnitTestsParallelizable/UnitTests.Parallelizable.csproj" (VSTest target(s)) -- FAILED.

Build FAILED.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:05:49.51

Attachments:
  /Users/runner/work/Terminal.Gui/Terminal.Gui/Tests/UnitTestsParallelizable/TestResults/7486e056-b52e-4ff1-b6aa-c57b9bb2d64c/Sequence_753af5ceaf3d4c62bdbfca2db0b889f2.xml
Error: Process completed with exit code 1.

Terminal.Gui/Views/TextInput/TextView/TextView.Movement.cs
BDisp: @tig I've investigated the ScrollTo method further and it's indeed widely used by the OnMouseEvent method in the TextView.Mouse file, where it helps adjust the viewport's location without changing the cursor's position, unlike InsertionPoint, which effectively changes the cursor's position. This behavior isn't what's desired when simply scrolling the viewport, whether wheeling or making a selection with the mouse.

I had the idea of ​​using InsertionPoint as a replacement for the public CurrentRow and CurrentColumn properties, which could simply be private fields that collect the necessary values ​​and would be deployed by assigning their values ​​to InsertionPoint. But this involves a lot of work because it's not as simple as changing an object's name and automatically updating the code. This would involve manually changing all references to CurrentRow and CurrentColumn. If you insist on this, then I advise opening a new issue for that purpose.

BDisp: What I ended up doing with the `ScrollTo` method was preventing scrolling from exceeding the viewport size. If it's not allowed with the keyboard, it shouldn't be allowed with the mouse either, unless the flag that allows it is activated. But that flag doesn't make sense in a text editor because only the editable part should be visible, not the other part. I think this should meet your needs. The change is in the commit https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/6697ac8222e00a4444fef9d40236e0770a3b3a59.

tig:
Hey @BDisp ,  how's work on this going?

BDisp:
> Hey @BDisp , how's work on this going?

I'm still working on it. It's giving me a ton of trouble, but it's already bearing fruit. I also confess that I'm not always working on it so I can unwind a bit. I hope to submit some commits regarding the tab stop processing soon.

tig:
Thanks. Let me know how I can help. 

IMO (and id appreciate your opinion on this) the only big things blocking us from releasing v2.0.0 are 

- The Adornments/Tab style work I'm working on now
- TextView 

Pretty exciting!

BDisp:
I've already made a lot of progress, but I haven't submitted it yet because it's not quite how I'd like it to be, especially the mouse selection. But I promise to dedicate more time to finishing it.

BDisp:
@tig I need to call `base.EndInit` before performing some initial configurations based on the availability of the `Viewport` property values ​​only when the `IsInitialized` property is true. This used to work properly. I think there was some change that no longer makes these values ​​available, because now the `Viewport` is an empty rectangle. Before, there was a pre-calculation of the layout where these values ​​were correctly available. The virtual `EndInit` method is fundamental for adjusting certain aspects of the view, but without initial values, not even from the `SuperView`, it's useless.

Using the modified code, nothing is displayed when I open the scenario Editor. If I use the original code, the scenario Editor displays the content, but some unit tests fail.

The commit https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/ede15b2e8133344aada6a5f593c15749f260c4d8 resolves this issue. I hope there are no further problems despite the unit tests passing. Thank you.

tig:
> @tig I need to call `base.EndInit` before performing some initial configurations based on the availability of the `Viewport` property values ​​only when the `IsInitialized` property is true. This used to work properly. I think there was some change that no longer makes these values ​​available, because now the `Viewport` is an empty rectangle. Before, there was a pre-calculation of the layout where these values ​​were correctly available. The virtual `EndInit` method is fundamental for adjusting certain aspects of the view, but without initial values, not even from the `SuperView`, it's useless.
> 
> Using the modified code, nothing is displayed when I open the scenario Editor. If I use the original code, the scenario Editor displays the content, but some unit tests fail.
> 
> The commit [ede15b2](https://github.com/gui-cs/Terminal.Gui/commit/ede15b2e8133344aada6a5f593c15749f260c4d8) resolves this issue. I hope there are no further problems despite the unit tests passing. Thank you.

One of the reasons we used to always use EndInit or Intialzizd was Adornments were lazy created. This is no longer the case.

I have been successful in moving away from depending on initialzied.

BDisp:
> One of the reasons we used to always use EndInit or Intialzizd was Adornments were lazy created. This is no longer the case.
> 
> I have been successful in moving away from depending on initialzied.

I find it very useful, especially for configurations that are only necessary at startup.

BDisp:
Guys, I have a question and I'd really like your opinion on it. Normally, terminals use the shift key for internal manipulation when the mouse is used simultaneously. TextView uses the LeftButtonPressed+Shift combination to copy text by simply clicking at the beginning of the text and then clicking at the end of the text you want to copy, always with the shift key pressed. This only works in WindowsDriver.

I have 3 questions I'd like you to answer, please:

1. Completely remove this feature from the API.
2. Replace the Shift key with the Control key.
3. Replace the Shift key with the Alt key.

Which one do you prefer?

YourRobotOverlord:
I would say 1 (or leave it in place where it works).
If the user is expecting selection modifiers, then moving it to something like Ctrl just collides with multiselect and you just end up with that behavior not working as expected.
Also, in my experience, Shift-select always means "Everything in between" and Ctrl-select always means "Multi-select". Changing the modifier to Ctrl would fly in the face of this convention, on Windows, at least. I'm not super familiar with the convention on Linux or Mac.

YourRobotOverlord:
What I found:

**Linux:**
- Shift-click: Select range (same as Windows)
- Ctrl-click: Multi-select (same as Windows)


**macOS:**
- Shift-click: Select range (same as Windows)
- Cmd-click (Command key): Multi-select (instead of Ctrl)
- Ctrl-click: Opens context menu (right-click alternative)

This reinforces my opinion, Option 1 or leave it working on Windows only, but with a goal to get it working everywhere at some point.

BDisp:
This feature no longer works in WindowsDriver using Windows Terminal. It now behaves the same way as Linux and macOS. And what's your opinion on replacing it with the Alt key?

tig:
It frustrates me that WT uses shift-press to select and ha no way of disabling that. 

With the kitty protocol the ansi driver could support shift-press on all platforms but Windows (with WT). Sigh. 

Whatever you pick, it should be the same used by table view and ListView.

YourRobotOverlord:
> And what's your opinion on replacing it with the Alt key?

My personal opinion is that it would be better to leave it out rather than support it in a non-standard way.
Just so I understand it better, what is the reason using the Shift modifier only works on Windows?

YourRobotOverlord:
> It frustrates me that WT uses shift-press to select and ha no way of disabling that.

Ah, I see; when using WT, shift-click is hijacked by the terminal and the selection is done by WT, not by TG. That *is* lame.

tig:
We should extend the configurable keybindings so mousebindings (and combos) are supported. 

Add Command.StartSelection. 

Then, on Windows the default binding would be Key.Alt + MouseFlags.LeftButtonPressed. 

On all others the default binding would be Key.Shift + MouseFlags.LeftButtonPressed. 

Eventually someone will want rectangular select. We'd define Command.StartRectangleSelection. And:

On Windows the default binding would be Key.Alt|Key.Ctrl + MouseFlags.LeftButtonPressed. 

On all others the default binding would be Key.Shift|Key.Ctrl + MouseFlags.LeftButtonPressed.

tig:
We don't have to do all that in this PR tho. Just use alt+down for all platforms for now.

BDisp:
Since I've known Linux and Mac, I've noticed that terminal emulators always take over the shift key when using the mouse. Therefore, I don't recommend using the shift key with the mouse, even in combination with other keys. The same behavior now occurs with Windows Terminal because it uses the conPTY emulator, aligning with the Unix system, but it has the advantage that you can export the selected text by right-clicking and selecting "Export text". Below are GIFs that confirm the standard behavior of using the shift key with the mouse in Linux and Mac.

VirtualBoxVM_QomkkO0YMG

Kapture 2026-04-06 at 17 42 52

tig:
> Since I've known Linux and Mac, I've noticed that terminal emulators always take over the shift key when using the mouse. Therefore, I don't recommend using the shift key with the mouse, even in combination with other keys. The same behavior now occurs with Windows Terminal because it uses the conPTY emulator, aligning with the Unix system, but it has the advantage that you can export the selected text by right-clicking and selecting "Export text". Below are GIFs that confirm the standard behavior of using the shift key with the mouse in Linux and Mac.

I didn't realize that. That just frustrates me ;-). So rude.

At least ghostty lets you change this behavior:

```
Ghostty has a mouse-shift-capture config option. Set it to always in your config file to make shift+drag always
  get forwarded to the running program instead of triggering selection:

  mouse-shift-capture = always
```

So, yes, we should default to Alt-Select on all platforms based on this.

BDisp:
@tig the last PR #4953 merged is causing the save dialog showing when I open the Editor scenario.

image

tig:
> @tig the last PR #4953 merged is causing the save dialog showing when I open the Editor scenario.
> 

Create an Issue please.

tig:
@BDisp how's progress on this? Can you give an estimate for when you think it'll be ready? 

This, and the Filedialog bug you recently reported are really the only things preventing release of v2.0.0. as far as I can see...

Let me know if/how I can help!

BDisp:
> @BDisp how's progress on this? Can you give an estimate for when you think it'll be ready?
> 
We already know that in programming it's difficult to make estimates, but I'll try to finish it today. I confess that I lost time with the problems I had with the kitty keyboard regarding the shifted keys and deduplication of some keys, which delayed me even more, but it was necessary.

> This, and the Filedialog bug you recently reported are really the only things preventing release of v2.0.0. as far as I can see...
> 
I'm changing all calls to rune measurement in TextFormatter.cs to measure graphemes because it's much more accurate for text. I'm only leaving the measurement of runes related to hotkeys unchanged because they are indeed of the Rune type. With the correction of the tab stop measurement, it's necessary to know the existing width up to the current grapheme before continuing its measurement, as there's no other way, to my knowledge. It's no longer enough to measure each grapheme individually. The entire text must be measured. It seems I've already achieved this, but, as expected, the unit tests need to be adjusted correctly and without bugs.

> Let me know if/how I can help!

Of course you can, and I even appreciate it. I'm having problems with scrollbars where, even when returning the correct location, it changes when the other scrollbar becomes visible. Adjusting the scrollbar's location should only be done when the viewport size is greater than or equal to the content size, or when the scrollbar itself is moved by the user. See confirmation of this behavior in the GIF below. The horizontal scrollbar doesn't move to the end because the vertical scrollbar becomes visible, scrambling the location. There should be a mechanism in the scrollbar where they inspect each other, waiting to see what the other will do before taking the most appropriate action.


WindowsTerminal_EB8wOxyYtY

tig:
Are all your latest changes pushed to this branch?

BDisp:
Only the ones that have already been properly tested are included. I'm still testing what I currently have, and I prefer to submit commits after they've been properly tested. But I think this is reproducible with the current PR. If not, please let me know. Thank you for your attention.

BDisp:
@tig this is ready for review. Thanks for your patient.

Terminal.Gui/Views/TextInput/TextField/TextField.Drawing.cs
```diff
@@ -32,12 +32,13 @@ protected override bool OnDrawingContent (DrawContext? context)
         SetAttribute (GetAttributeForRole (VisualRole.Normal));
         Move (0, 0);
 
-        int p = ScrollOffset;
-        var col = 0;
+        _ = TextModel.CursorColumn (_text, ScrollOffset, 0, out List<int> glyphWidths, out _);
```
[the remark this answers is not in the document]
BDisp: Your reasoning is wrong in this case because the CursorColumn method is returning the glyphWidths, which will be used in the GetColumnWidthsBeforeStart method to calculate the startIndex, represented by the viewportX variable. Therefore, it is necessary to use ScrollOffset, which, although it can represent column widths larger than the startIndex, is performing its function correctly.

Terminal.Gui/Views/TextInput/TextView/TextView.Mouse.cs
```diff
@@ -11,14 +11,16 @@ public partial class TextView
...
     /// <inheritdoc/>
     protected override bool OnMouseEvent (Mouse mouse)
     {
         if (mouse is { IsSingleDoubleOrTripleClicked: false, IsPressed: false, IsReleased: false, IsWheel: false }
-            && !mouse.Flags.FastHasFlags (MouseFlags.LeftButtonPressed | MouseFlags.PositionReport)
-            && !mouse.Flags.FastHasFlags (MouseFlags.LeftButtonPressed | MouseFlags.Shift)
-            && !mouse.Flags.FastHasFlags (MouseFlags.LeftButtonDoubleClicked | MouseFlags.Shift)
+            && !mouse.Flags.HasFlag (MouseFlags.LeftButtonPressed | MouseFlags.PositionReport)
+            && !mouse.Flags.HasFlag (MouseFlags.LeftButtonPressed | MouseFlags.Alt)
+            && !mouse.Flags.HasFlag (MouseFlags.LeftButtonDoubleClicked | MouseFlags.Alt)
```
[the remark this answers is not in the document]
BDisp: This was a decision made, as per the comment at https://github.com/gui-cs/Terminal.Gui/pull/4723#issuecomment-4233987465, due to the fact that many terminals do not replicate the mouse event used in conjunction with the Shift key being pressed for their internal use.

Terminal.Gui/Views/TextInput/TextView/TextView.Drawing.cs
```diff
@@ -26,11 +26,13 @@ protected override bool OnDrawingContent (DrawContext? context)
         {
             List<Cell> line = _model.GetLine (idxRow);
             int lineRuneCount = line.Count;
-            var col = 0;
+            _ = TextModel.CursorColumn (TextModel.CellsToStringList (line), Viewport.X, TabWidth, out List<int> glyphWidths, out _);
```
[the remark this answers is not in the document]
BDisp: The `end` parameter of the `TextModel.CursorColumn` method does not necessarily have to be the current column containing the cursor location, nor does it necessarily have to be an index within the string list range. As previously mentioned, in this case the method is returning `glyphWidths`, which will be used in the `TextModel.GetColumnWidthsBeforeStart` method to return the `startIndex` represented by the `viewportX` variable. Note that the value of `Viewport.X` will return different `startIndex` values ​​for each line of text, depending on the width of its columns. Therefore, the call is performing its function correctly.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
```diff
@@ -2185,6 +2185,34 @@ public void ContentSize_Updates_When_Text_Is_Edited ()
         Assert.True (tv.VerticalScrollBar.Visible, "Vertical scrollbar should be visible when content exceeds viewport height");
     }
 
+    [Fact]
+    public void ContentSize_Updates_When_Text_Is_Delected ()
```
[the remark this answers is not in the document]
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/164e1176bd75cb5712511728013d4090d7f63509.

Terminal.Gui/Views/TextInput/TextView/TextView.cs
```diff
@@ -224,56 +224,52 @@ public void PositionCursor ()
         }
 
         List<Cell> line = _model.GetLine (CurrentRow);
-        var col = 0;
 
-        if (line.Count > 0)
+        // Calculate absolute cursor position and store each glyph width
+        int cursorColumn = TextModel.CursorColumn (TextModel.CellsToStringList (line), CurrentColumn, TabWidth, out List<int> glyphWidths, out _);
+        var colsWidth = 0;
+
```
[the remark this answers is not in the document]
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/426001b270bcb01ccb5491552fe976f4e2776a80.

Terminal.Gui/Views/TextInput/TextView/TextView.Scrolling.cs
```diff
@@ -57,6 +57,11 @@ private void UpdateContentSize ()
         int contentWidth = WordWrap ? Viewport.Width : _model.GetMaxVisibleLine (0, _model.Count, TabWidth);
 
         SetContentSize (new Size (contentWidth, contentHeight));
+        if (GetContentSize () == new Size (contentWidth + 1, contentHeight))
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/d225935855310a0e7d5dc2d06ecb41497ee9e66d.

Tests/UnitTestsParallelizable/Views/TextViewTests.cs
```diff
@@ -2426,6 +2450,1260 @@ public void Command_HotKey_SetsFocus ()
...
+        DriverAssert.AssertDriverContentsAre (expected, output, app.Driver);
+
+        textView.PositionCursor ();
+
+        // Cursor is always screen relative
+        Assert.Equal (new Point (1, 1), textView.Cursor.Position!.Value);
+    }
+
+    [Fact]
+    public void InvokeCommand_PageDown_PageUp_With_ViewportHeight_GraterThanOne_Draws_Cursor_Correctly ()
```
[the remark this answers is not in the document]
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/48c38ce9b1f88cb13dc7bf8da3b02689abb94999.

Terminal.Gui/Views/TextInput/TextView/TextView.Movement.cs
```diff
@@ -7,7 +7,12 @@ public bool MoveEnd ()
     {
         CurrentRow = _model.Count - 1;
         List<Cell> line = GetCurrentLine ();
-        CurrentColumn = line.Count;
+        CurrentColumn = line.Count + 1;
+
+        if (CurrentRow >= Viewport.Y + Viewport.Height || CurrentColumn > Viewport.X + Viewport.Width)
+        {
+            SetNeedsDraw ();
+        }
```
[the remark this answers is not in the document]
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/8b0463e64511998966a02af618768d67e18640cd.

Tests/UnitTestsParallelizable/Views/TextModelTests.cs
```diff
@@ -736,6 +767,222 @@ public void CalculateLeftColumn_HandlesTabs ()
...
+    public void CalculateLeftColumn_Tab_ReturnsCorrectColumn ()
+    {
+        // Setup: "hello\tworld" 
+        // Visual columns: h(0), e(1), l(2), l(3), o(4), \t(switches to 8), w(8), o(9), r(10), l(11), d(12)
+        var text = "hello\\tworld";
+        List<string> list = [];
+
+        foreach (string s in GraphemeHelper.GetGraphemes (text))
+        {
+            list.Add (s);
```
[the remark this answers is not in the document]
BDisp: Done in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/9eb1d4ad8a7d94f7c85b9ec02aa611d6bc56e499.

tig:
I have some tweaks I'm working on. 

First one is ReadOnly sets CanFocus false which prevents context menu from working.

Second is CM Editor has related bugs.

Third, which I have no idea of: CM Editor is super slow. The same json files loaded in Editor are not slow. I don't understand this yet.

BDisp:
> I have some tweaks I'm working on.
> 
> First one is ReadOnly sets CanFocus false which prevents context menu from working.
> 
I think there's some confusion about how CanFocus behaves when ReadOnly = true in TextField and TextView, because it shouldn't restrict CanFocus from being true since it's not the same as Enable = false. That said, when ReadOnly is true it only prevents editing but not selecting or copying text, restricting the ContextMenu. When Enable is false then yes, CanFocus is false and can never be focused or interacted with by either the keyboard or the mouse.

> Second is CM Editor has related bugs.
> 
> Third, which I have no idea of: CM Editor is super slow. The same json files loaded in Editor are not slow. I don't understand this yet.

I noticed that after seeing your message. Could it be slow because you're opening multiple files at the same time, even though only one is visible at a time? Perhaps it's because you're using `TabView`. Maybe try using a different view, like a menu, and select one at a time to see if you experience the same slowness.

tig:
Some changes pushed. Now working on the CME issue, which I've analyzed and undestand the root causes of. it's fundamentally a perf issue made worse by the use of Tabs in CME. I'm now working through how to fix these perf issues, which will hopefully have a dramatic perf impact for all TextView scenarios.

tig:
# TextView Scroll Performance Fix

## Problem Statement

The `ConfigurationEditor` UICatalog scenario scrolls significantly slower than the `Editor` scenario, even though both use `TextView` with `WordWrap = false`. The root cause is a set of O(N×L) operations in `TextView` that run on every scroll/layout cycle, where N = total lines and L = average line length.

### Why ConfigurationEditor is worse than Editor

| Factor | Editor | ConfigurationEditor |
|--------|--------|---------------------|
| View hierarchy | 1 level (`Window → TV`) | 3 levels (`Window → Tabs → Tab → TV`) |
| Layout cascades | Infrequent | Frequent (deeper nesting → more `OnSubViewsLaidOut` calls) |
| `ReadOnly` | `false` | `true` (extra per-cell work in drawing) |
| `ContentsChanged` handler | Lightweight status bar update | Calls `IsDirty` → O(all cells) comparison |

Both scenarios suffer from the same core `TextView` bottlenecks, but ConfigurationEditor triggers the expensive paths more frequently due to deeper view hierarchy causing more layout passes.

---

## Root Causes (ordered by impact)

### RC1: `GetMaxVisibleLine()` scans ALL lines on every content-changed scroll

**File:** `TextModel.cs:65-82`, called from `UpdateContentSize()` (`TextView.Scrolling.cs:49-66`)

```csharp
int contentWidth = WordWrap ? Viewport.Width : _model.GetMaxVisibleLine (0, _model.Count, TabWidth);
```

`GetMaxVisibleLine` iterates every line in the model, calling `CellsToStringList()` + `CursorColumn()` on each. This is **O(N×L)** per call. It runs whenever `contentMayHaveChanged` is true (which includes `NeedsDraw`).

**Fix:** Cache the max visible line width in `TextModel`. Invalidate only when lines are added, removed, or modified — not on scroll.

### RC2: `OnSubViewsLaidOut` calls `WrapTextModel()` + `UpdateContentSize()` unconditionally

**File:** `TextView.cs:157-166`

```csharp
protected override void OnSubViewsLaidOut (LayoutEventArgs args)
{
    base.OnSubViewsLaidOut (args);
    WrapTextModel ();       // O(N×L) full rewrap even if width unchanged
    UpdateContentSize ();   // O(N×L) full line scan (see RC1)
}
```

**Fix:** Guard `WrapTextModel()` with a width-change check. Guard `UpdateContentSize()` with a dirty flag — only recalculate when content actually changed.

### RC3: `CellsToStringList()` allocates a new `List<string>` on every call

**File:** `TextModel.cs:578-588`

Called per-line in drawing, per-line in `GetMaxVisibleLine`, in `AdjustViewport`, in `PositionCursor`. Creates GC pressure from thousands of small allocations per frame.

**Fix:** This is a deeper refactor (caching or span-based). Defer to a follow-up. Focus on eliminating the call sites via RC1/RC2 fixes first.

### RC4: `AdjustViewport()` computes 3 redundant `DisplaySize`/`CursorColumn` calls

**File:** `TextView.Scrolling.cs:91-94`

Three separate calls each allocate a list and scan the current line. These could be consolidated.

**Fix:** Compute once, reuse results. Lower priority since it's O(L) not O(N×L).

### RC5: `PositionCursor()` called 2-4× per position change

`CurrentRow` and `CurrentColumn` setters each call `PositionCursor()`. When both change (common), it runs twice. `SetWrapModel`/`UpdateWrapModel` pairs trigger 4 calls total.

**Fix:** Use backing fields directly and call `PositionCursor()` once. Lower priority.

---

## Implementation Plan

### Phase 1: Add passing tests for existing behavior

Add tests that verify current correct behavior of the code paths we'll be modifying. These tests MUST pass before and after our changes.

**File:** `Tests/UnitTestsParallelizable/Views/TextViewPerformanceTests.cs`

Tests to add:
1. `GetMaxVisibleLine_Returns_Correct_Width_For_Mixed_Lines` — verifies `GetMaxVisibleLine` returns the width of the longest line
2. `UpdateContentSize_Sets_Correct_Width_And_Height` — verifies content size reflects model dimensions
3. `OnSubViewsLaidOut_Calls_UpdateContentSize` — verifies content size is updated after layout
4. `ContentSize_Correct_After_Text_Change` — verifies content size updates when text changes
5. `ContentSize_Correct_After_Line_Append` — verifies content size updates when lines are appended programmatically
6. `WrapTextModel_Only_Wraps_When_WordWrap_Enabled` — verifies `WrapTextModel()` is a no-op when `WordWrap = false`
7. `Viewport_Scrolling_Does_Not_Change_ContentSize_When_Content_Unchanged` — verifies that pure scrolling (changing Viewport.Y) does not alter content size

### Phase 2: Add failing tests that expose the performance issue

These tests document the expected optimized behavior and will FAIL against current code.

**File:** `Tests/UnitTestsParallelizable/Views/TextViewPerformanceTests.cs`

Tests to add:
1. `GetMaxVisibleLine_Is_Not_Called_On_Pure_Scroll` — instruments/counts calls; verifies `GetMaxVisibleLine` is NOT called when viewport changes without content changes
2. `UpdateContentSize_Skips_GetMaxVisibleLine_When_Content_Unchanged` — verifies that scrolling without editing doesn't trigger the full line scan
3. `OnSubViewsLaidOut_Skips_WrapTextModel_When_Width_Unchanged` — verifies that `WrapTextModel` is not called when viewport width hasn't changed
4. `ContentSize_Width_Cached_Between_Scrolls` — verifies content size width remains stable between scroll operations without recalculating

### Phase 3: Implement the performance fixes

#### Fix 1: Cache max line width in TextModel

**File:** `TextModel.cs`

- Add `_cachedMaxWidth` field and `_maxWidthDirty` flag
- `GetMaxVisibleLine()` returns cached value when not dirty
- Invalidate cache in `AddLine()`, `RemoveLine()`, `ReplaceLine()`, `LoadString()`, `LoadFile()`, and any method that modifies `_lines`
- On single-line edits, do incremental update: compare new line width vs cached max

#### Fix 2: Guard `OnSubViewsLaidOut` with width-change detection

**File:** `TextView.cs`

- Add `_lastWrapWidth` field
- In `OnSubViewsLaidOut`, only call `WrapTextModel()` if `Viewport.Width != _lastWrapWidth`
- Only call `UpdateContentSize()` if a dirty flag is set (content changed or wrap changed)
- Update `_lastWrapWidth` after wrapping

#### Fix 3: Add content-dirty flag to skip `UpdateContentSize` on scroll

**File:** `TextView.Scrolling.cs`

- Add `_contentSizeDirty` flag
- Set it to `true` in all text-modification code paths (where `OnContentsChanged` is called, in `LoadString`, etc.)
- In `UpdateContentSize()`, early-return if `!_contentSizeDirty`
- Reset flag after recalculating
- Always set dirty on first use (`!Used`)

---

## Verification Steps

1. All existing tests in `TextViewScrollingTests`, `TextViewTests`, `TextModelTests`, and other `TextView*Tests` must continue to pass
2. Phase 1 tests pass before AND after changes
3. Phase 2 tests fail before changes, pass after
4. Build with no new warnings
5. Manual verification: open ConfigurationEditor scenario and scroll — should be noticeably faster
6. Manual verification: open Editor scenario — should be at least as fast as before
7. Manual verification: edit text in both scenarios — content size must still update correctly

---

## Files Modified

| File | Change |
|------|--------|
| `Terminal.Gui/Views/TextInput/TextModel.cs` | Add max-width caching with dirty flag |
| `Terminal.Gui/Views/TextInput/TextView/TextView.cs` | Guard `OnSubViewsLaidOut` with width check |
| `Terminal.Gui/Views/TextInput/TextView/TextView.Scrolling.cs` | Add content-dirty flag, skip redundant `UpdateContentSize` |
| `Tests/UnitTestsParallelizable/Views/TextViewPerformanceTests.cs` | New test file with Phase 1 + Phase 2 tests |

## Out of Scope (follow-up)

- `CellsToStringList` allocation optimization (needs deeper refactor)
- `AdjustViewport` consolidation of redundant `DisplaySize` calls
- `PositionCursor` batching
- `IsDirty` O(N) optimization (only affects editing, not scrolling)
- `GetModelColFromWrappedLines` linear scan per-cell during word-wrap drawing

tig:
@bdisp - Here's something else i noticed, with WordWrap = true:

Scroll down so there's text above the viewport, with an insertion point visible.

Type something.

Scrolling happens, moving the content around.

Perhaps my perf fixes will fix this since they will redcue the layout events, but maybe not. 

Probably best to wait to work on this until I'm done with the perf work.

tig:
> @BDisp - Here's something else i noticed, with WordWrap = true:
> 
> Scroll down so there's text above the viewport, with an insertion point visible.
> 
> Type something.
> 
> Scrolling happens, moving the content around.
> 
> Perhaps my perf fixes will fix this since they will redcue the layout events, but maybe not.
> 
> Probably best to wait to work on this until I'm done with the perf work.

I've got the perf work described above done. It makes a signifcant impact in the Editor scenario, but has not speed up CME, so I'm going to dive deeper.

I did NOT fix the bug described above.

I'll push my changes in a few minutes and then re-attack the perf issue again.

BDisp:
I've already found the bug. Could you please wait until after I submit the fix?

BDisp:
> I've already found the bug. Could you please wait until after I submit the fix?

I already pushed the fix in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/fee2df316382b068f508a13016c925553bd760db.

tig:
What you did made it better, but it's still moving content around as the user types.

I pushed my updates.

BDisp:
If you mean is moving down it's normal because using word wrap always adjust the text in the viewport.

tig:
WindowsTerminal_qyUruutAiM

but hold on - looks like i didnt actually merge your latest changes!!

tig:
you fixed it. All good!!

YourRobotOverlord:
@BDisp I'm noticing a strange behavior; if you type any non-repeating keys, e.g. "1234", they work fine.
If you repeat the same keys, every second repeated keypress is ignored, e.g. "11223344" adds "1234", "111222333444" adds "11223344", etc. to the `TextField`. 
This is the same for any keys I've tried - space, backspace, delete, etc., and it is also affecting `TextView`.
I'm not sure if it is caused by this PR, but I haven't noticed it before.

YourRobotOverlord:
Hmm, this might actually pre-date this pr. I'll investigate more in the morning.

tig:
I'm pretty sure the perf issue is not really a TextView bug, but related to 

- #4522 

I want to fix it separately from this PR.

BDisp:
> @BDisp I'm noticing a strange behavior; if you type any non-repeating keys, e.g. "1234", they work fine. If you repeat the same keys, every second repeated keypress is ignored, e.g. "11223344" adds "1234", "111222333444" adds "11223344", etc. to the `TextField`. This is the same for any keys I've tried - space, backspace, delete, etc., and it is also affecting `TextView`. I'm not sure if it is caused by this PR, but I haven't noticed it before.

I can't reproduce this but I fixed a bug in `TextField` where mouse clicking in a position greater than zero in a empty text the `ScrollOffset` was set with  negative value in https://github.com/gui-cs/Terminal.Gui/pull/4723/commits/7aa97f9c3289943e1fe21327c09a52e85a6a3b63. But this was only causing throwing an exception on the `DisplaySize` method.

YourRobotOverlord:
> I can't reproduce this ...

I wonder if it is down to something different in our machine setup, I'm using the ANSI driver under Win11. It happens everywhere in TG now, not just 'TextView', but I've traced it down to PR #4949:
```csharp
    protected override bool ShouldSuppressFallbackKeyDown (Key key)
    {
        if (string.IsNullOrEmpty (_pendingPrintableSuppression))
        {
            _pendingPrintableSuppression = key.GetPrintableText (); // <= this

            return false;
        }
```
I'll open a new issue for this.

BDisp:
> > I can't reproduce this ...
> 
> I wonder if it is down to something different in our machine setup. It happens everywhere in TG now, not just 'TextView', but I've traced it down to PR #4949:
> 
> ```cs
>     protected override bool ShouldSuppressFallbackKeyDown (Key key)
>     {
>         if (string.IsNullOrEmpty (_pendingPrintableSuppression))
>         {
>             _pendingPrintableSuppression = key.GetPrintableText (); // <= this
> 
>             return false;
>         }
> ```
> 
> I'll open a new issue for this.

But that was needed to fix my PT keyboard layout where some keys was printing twice, «»çÇºª, because they don't appear as kitty sequence. It's needed to understand why this happen. Can you please share what sequence you receive when you type 1. In my machine returns "\u001b[49;;49u". Thanks.

Lets discuss this in #4976

tig:
@BDisp Ready?

BDisp:
> @BDisp Ready?

From my side, yes.

tig:
Really nice work on this. I know this is a lot of work well done

PR #4740, opened 2026-02-21
Fixes #4229 Support transparent foreground/background in v2

### LINKED ISSUE #4229: V2 the default background of the app is not transparent.

HueByte:
When a color is Color.Transparent (alpha=0), the rendering pipeline emits ANSI CSI 39m/49m (default foreground/background) instead of explicit RGB values. This allows the terminal's native background (including transparency, acrylic, and wallpaper effects) to show through.

Changes:
- Add Color.Transparent static field (ARGB 0,0,0,0)
- Add StandardColor.Transparent enum value
- Change StandardColor to uint base type with 0xFF alpha on all values to correctly distinguish Black (0xFF000000) from Transparent (0x00000000)
- Simplify StandardColors.GetArgb() since alpha is now embedded in values
- Add CSI_AppendResetForegroundColor/BackgroundColor to EscSeqUtils
- Move AppendOrWriteAttribute from abstract to virtual in OutputBase with transparency checks in both Force16Colors and true-color paths
- Remove duplicate AppendOrWriteAttribute overrides from NetOutput, UnixOutput, and AnsiOutput (they now inherit from OutputBase)
- Simplify WindowsOutput to delegate to base for non-legacy console
- Handle Transparent in Scheme.cs Focus inversion to prevent invisible text
- Based on #4234 

## Fixes

Fixes #4229 

## Pull Request checklist:

### DIFF
--- a/Examples/UICatalog/Scenarios/TableEditor.cs
+++ b/Examples/UICatalog/Scenarios/TableEditor.cs
@@ -809,13 +809,17 @@ public class TableEditor : Scenario
                                                           { "FileSize", GetHumanReadableFileSize }
                                                       });
 
-        HashSet<string> seen = new ();
-
+        bool isWindows = OperatingSystem.IsWindows ();
         try
         {
-            foreach (string path in Environment.GetLogicalDrives ())
+            foreach (DriveInfo di in DriveInfo.GetDrives ())
             {
-                tree.AddObject (new DirectoryInfo (path));
+                if (isWindows && di is { DriveType: DriveType.Network, IsReady: false })
+                {
+                    continue;
+                }
+
+                tree.AddObject (new DirectoryInfo (di.Name));
             }
         }
         catch (Exception e)
--- a/Examples/UICatalog/Scenarios/TreeViewFileSystem.cs
+++ b/Examples/UICatalog/Scenarios/TreeViewFileSystem.cs
@@ -490,23 +490,42 @@ public class TreeViewFileSystem : Scenario
                 _fileInfo = value;
                 StringBuilder? sb = null;
 
-                if (_fileInfo is IFileInfo f)
+                try
                 {
-                    Title = $"{_iconProvider.GetIconWithOptionalSpace (f)}{f.Name}".Trim ();
-                    sb = new StringBuilder ();
-                    sb.AppendLine ($"Path:\n {f.FullName}\n");
-                    sb.AppendLine ($"Size:\n {f.Length:N0} bytes\n");
-                    sb.AppendLine ($"Modified:\n {f.LastWriteTime}\n");
-                    sb.AppendLine ($"Created:\n {f.CreationTime}");
-                }
+                    if (_fileInfo is IFileInfo f)
+                    {
+                        Title = $"{_iconProvider.GetIconWithOptionalSpace (f)}{f.Name}".Trim ();
+                        sb = new StringBuilder ();
+                        sb.AppendLine ($"Path:\n {f.FullName}\n");
+                        sb.AppendLine ($"Size:\n {f.Length:N0} bytes\n");
+                        sb.AppendLine ($"Modified:\n {f.LastWriteTime}\n");
+                        sb.AppendLine ($"Created:\n {f.CreationTime}");
+                    }
 
-                if (_fileInfo is IDirectoryInfo dir)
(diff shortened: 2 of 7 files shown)


HueByte:
Made this for to mostly support transparent themes in my app, looks good so far
image

tig (commented):
This is fantastic! It makes me feel warm inside ;-).

I'm going to pull this down locally and take a very close look asap. This is a pretty fundamental concept and it's important we get it right. My instincts, having tried several times to address this, are that you have nailed it, but I want to make sure.

Most likely, I'll submit a PR to this PR... 

In the meantime, this needs unit tests....

Thanks again!

HueByte:
Sounds good @tig , and thank you! 
I'll add some unit tests

Terminal.Gui/Drawing/Color/Color.cs
```diff
@@ -41,6 +41,17 @@ namespace Terminal.Gui.Drawing;
 public readonly partial record struct Color : ISpanParsable<Color>, IUtf8SpanParsable<Color>, ISpanFormattable,
                                               IUtf8SpanFormattable, IMinMaxValue<Color>
 {
+    /// <summary>
```
mrazza, line 44: I'll mention for consideration that the framing of "transparent" here seems incorrect and inconsistent with other terminal apps. The actual behavior is to set the color to terminal defaults _not_ to enable transparency. Insofar as the default terminal is transparent, sure, but there should be no expectation on the part of Terminal.Gui that this is the case.

Vim communicates this setting as "NONE"; presumably conveying that the application is not applying any color and allowing the terminal application to decide. (e.g. `hi Normal guibg=NONE ctermbg=NONE`)

(fwiw, I'm excited to revert my equivalent change in my fork of TGUI for something from main so ty)

tig: I completely agree. My bad that I have continued to refer to this as transparency... that's lazy.

@HueByte -  What are your thoughts on this? Alternatives to "Transparent":

- "None"
- "Passthrough"
- "Default"
- "FullAlpha"
- "Clear"
- "TerminalDefined"

I wish "Default" weren't such an overloaded term because as @mrazza says "he actual behavior is to set the color to terminal defaults ...".

HueByte: I think the "None" fits best, or maybe "System" as it shows user/user's system background in the end
I'm good with any of those tbh

mrazza: +1 to None or TerminalDefined

YourRobotOverlord: +1 for None. I feel like it conveys the intention best

HueByte: *None* it is, pushed the change.
Feel free to resolve comment if it looks right

Terminal.Gui/Drawing/Color/Color.cs
```diff
@@ -41,6 +41,16 @@ namespace Terminal.Gui.Drawing;
...
+    ///     No color (alpha = 0). When used in an <see cref="Attribute"/>, the terminal's default
+    ///     foreground or background color is used instead of an explicit color. This allows the terminal's native
+    ///     background (including any transparency or acrylic effects) to show through.
+    /// </summary>
+    /// <remarks>
+    ///     Uses ARGB 0x00FFFFFF (alpha=0, RGB=white) as the sentinel value. This ensures it is distinguishable from
+    ///     <c>default(Color)</c> which has all bytes zeroed.
+    /// </remarks>
+    public static readonly Color None = new (255, 255, 255, 0);
```
mrazza, line 53: nit: missing blank line between fields

tig (approved):

tig:
@HueByte it looks like this broke building. you probably didn't test building in Release?

HueByte:
Hmm, I think I did @tig, but I was in hurry this morning so might've missed something 
Ima take a look and fix whatever is wrong with it now, noticed you did merge so maybe something got overridden?

PR #4769, opened 2026-02-26
Fixes #4759 - Space HotKeyBinding swallowed by focused view

### LINKED ISSUE #4759: Default View KeyBindings blocking HotKeyBinding
I'm not sure what the ideal behavior here is so I was between opening an issue and a discussion for this.

Prior to https://github.com/gui-cs/Terminal.Gui/pull/4620 I had a HotKeyBinding for `Key.Space` in a View a few levels down from the root `Runnable`. This HotKey would trigger independent of what views were in focus or whether or not the view with the hotkey was visible. Only if a view explicitly captured the keystroke (e.g. a TextField via `OnKeyDownNotHandled` or similar) would the HotKey not fire. This was ideal.

However, now, `Key.Space` can, effectively, _not_ be used as a global HotKey as every few captures it and fires the `Activate` event. It would be one thing if the view actually had an action to take with the `Activate` event (in that, it set the `CommandContext.Handled` to true). However, it appears to capture the `Key.Space` ahead of the HotKeyBinding no matter what.

Is this expected?

A solution that works is to unbind `Key.Space` in _every view in the app_ but this is annoying and challenging when you simply want to re-use an existing View.

This seems to apply no matter how the HotKey is set:
```
App!.Keyboard.KeyBindings.AddApp(Key.Space, this, Command.HotKey);

// or

HotKeyBindings.Add(Key.Space, Command.HotKey);
```

Note that _non-Space_ keys work correctly with either method.

This is trivially reproducable by modifying `KeyBindingsTests.Focused_HotKey_Application_All_Work` to use `Key.Space` instead of `Key.A` (don't forget to modify `KeyBindingTests.ScopedKeyBindingView.EndInit` as well).

For some reason `Key.Enter` *does not* have the same behavior --- HotKeyBindings work despite also being bound to a Command (`Command.Accept`) by default in the View (in `View.Keyboard.cs` - `SetupKeyboard`)

tig:
## Summary

Fixes #4759: HotKeyBindings for `Key.Space` were swallowed by the focused view's `Command.Activate` KeyBinding before the HotKey dispatch could run.

### Why Space didn't work as a HotKeyBinding when a view was focused

The dispatch order in `NewKeyDownEvent` (`View.Keyboard.cs:443`):

1. Recurse into **Focused** subview (line 457)
2. `RaiseKeyDown` (line 463)
3. `InvokeCommandsBoundToKey` — **KeyBindings** on this view (line 471)
4. `InvokeCommandsBoundToHotKey` — **HotKeyBindings** on this view and subviews (line 476)
5. `RaiseKeyDownNotHandled` (line 484)

`SetupKeyboard` binds `Key.Space` → `Command.Activate` in every View's **KeyBindings**. `DefaultActivateHandler` unconditionally returned `true` (line 512), so the key was consumed at step 3 and HotKey dispatch at step 4 never ran.

### Why Enter didn't have the same problem

`Key.Enter` → `Command.Accept` is also bound in `SetupKeyboard`, but `DefaultAcceptHandler` returns `false` for plain Views (non-`IAcceptTarget`, no `DefaultAcceptView`, no bubble config) at line 314. This let the key "leak" through to step 4 where HotKeyBindings fire. **This asymmetry was unintentional.**

### The fix

`DefaultActivateHandler` now mirrors `DefaultAcceptHandler`'s return logic: it returns `_lastDispatchOccurred || activateWillBubble` instead of unconditional `true`. Plain views with no dispatch target and no bubble config return `false`, allowing key propagation to continue to `InvokeCommandsBoundToHotKey`.

### The bogus Space rejection in AddKeyBindingsForHotKey

`AddKeyBindingsForHotKey` (`View.Keyboard.cs:134`) explicitly rejected `Key.Space` with an `ArgumentException`. However, `HotKeyBindings.Add(Key.Space, ...)` bypassed this validation entirely. The Space rejection was overly protective — Space is a valid printable key and the `Rune.IsControl` check already guards against actual control characters. **This restriction has been removed.**

## Changes

- **`View.Command.cs`**: `DefaultActivateHandler` returns `_lastDispatchOccurred || activateWillBubble` instead of unconditional `true`
- **`View.Keyboard.cs`**: Removed `baseKey == Key.Space` from `AddKeyBindingsForHotKey` guard
- **New test** `Space_HotKey_Fires_When_Sibling_View_Is_Focused`: Replicates @mrazza's exact scenario — HotKeyBinding for Space on a non-focusable view nested several levels deep, with a separate focused sibling that has only default bindings
- **New test** `Enter_HotKey_Fires_When_Sibling_View_Is_Focused`: Proves the Enter/Space symmetry now works correctly
- **New test** `AddKeyBindingsForHotKey_Accepts_Space`: Verifies Space is accepted by `AddKeyBindingsForHotKey`
- **Updated test** `InvokeCommands_Returns_False_If_No_Command_Handled`: Was encoding the broken behavior (expected `true` when no command was handled)
- **Updated test** `Command_Activate_Calls_SetFocusedOption`: Side effect still works via `OnActivated`, return value correctly reflects unhandled status
- **`ScopedKeyBindingView.ExtraHotKey`**: Supports adding HotKeyBindings via `HotKeyBindings.Add`
- **Code style cleanup** of `HotKeyTests.cs` and `KeyBindingsTests.cs`

## Checklist

- [x] Remove `baseKey == Key.Space` from the `AddKeyBindingsForHotKey` guard
- [x] Update `Set_Throws_With_Invalid_Key` to remove `KeyCode.Space` from invalid keys
- [x] Convert `AddKeyBindingsForHotKey_Rejects_Space` to a positive test
- [x] Add failing test replicating mrazza's exact scenario (nested view, sibling focused)
- [x] Add passing test proving Enter asymmetry
- [x] Fix the dispatch priority so HotKeyBindings aren't swallowed by focused view KeyBindings

## Test plan

- [x] `Space_HotKey_Fires_When_Sibling_View_Is_Focused` passes
- [x] `Enter_HotKey_Fires_When_Sibling_View_Is_Focused` passes
- [x] All 15285 tests pass (14279 parallel + 1006 non-parallel)
- [x] Enter behavior is not regressed

### DIFF
--- a/Tests/UnitTestsParallelizable/ViewBase/Keyboard/KeyBindingsTests.cs
+++ b/Tests/UnitTestsParallelizable/ViewBase/Keyboard/KeyBindingsTests.cs
@@ -177,80 +177,44 @@ public class KeyBindingsTests
 
     // Claude - Opus 4.5
     /// <summary>
-    ///     Replicates issue #4759: A View several levels down has a HotKeyBinding for Key.Space.
-    ///     A *different* focused view (plain View with default Space→Activate binding) swallows
-    ///     the key via DefaultActivateHandler returning true, preventing the HotKey from firing.
-    ///     This test mirrors @mrazza's scenario: the HotKey should fire regardless of which
-    ///     view is focused, unless that view explicitly handles the key.
+    ///     Validates that unhandled default KeyBinding commands (Activate for Space, Accept for Enter)
+    ///     do not block HotKey dispatch. When a focused plain View has a default binding for a key but
+    ///     doesn't genuinely handle the command (no dispatch target, no bubble config), the key should
+    ///     propagate to <see cref="View.InvokeCommandsBoundToHotKey"/> so that HotKeyBindings on
+    ///     sibling/ancestor views can fire. See issue #4759.
     /// </summary>
-    [Fact]
-    public void Space_HotKey_Fires_When_Sibling_View_Is_Focused ()
+    [Theory]
+    [InlineData ("Space")] // Space → Command.Activate (via SetupKeyboard)
+    [InlineData ("Enter")] // Enter → Command.Accept (via SetupKeyboard)
+    public void Unhandled_Default_KeyBinding_Does_Not_Block_HotKey (string keyName)
     {
+        Key key = keyName == "Space" ? Key.Space : Key.Enter;
+
         IApplication app = Application.Create ();
         app.Begin (new Runnable<bool> { CanFocus = true });
 
-        // The view with the Space HotKeyBinding — a few levels down, like mrazza's scenario
-        View container = new () { CanFocus = false };
-        View innerContainer = new () { CanFocus = false };
+        // A view with a HotKeyBinding for the key under test
         View hotKeyView = new () { CanFocus = false };
         var hotKeyFired = false;
         hotKeyView.HandlingHotKey += (_, _) => hotKeyFired = true;
-        hotKeyView.HotKeyBindings.Add (Key.Space, Command.HotKey);
-
-        container.Add (innerContainer);
-        innerContainer.Add (hotKeyView);
-        app!.TopRunnableView!.Add (container);
-
-        // A separate focusable view (sibling) — plain View with default Space→Activate binding
-        View focusableView = new () { CanFocus = true };
-        app!.TopRunnableView!.Add (focusableView);
-
-        // With no view focused, Space HotKey should fire
-        hotKeyFired = false;
-        app.Keyboard.RaiseKeyDownEvent (Key.Space);
-        Assert.True (hotKeyFired);
-
-        // Focus the sibling view — it has default Space→Command.Activate from SetupKeyboard
-        focusableView.SetFocus ();
-        Assert.True (focusableView.HasFocus);
-
-        // Space HotKey should STILL fire — the focused view's Activate is unhandled/default
-        // BUG: DefaultActivateHandler returns true unconditionally, swallowing Space
-        // before InvokeCommandsBoundToHotKey runs on the SuperView hierarchy
-        hotKeyFired = false;
-        app.Keyboard.RaiseKeyDownEvent (Key.Space);
-        Assert.True (hotKeyFired);
-    }
-
-    // Claude - Opus 4.5
-    /// <summary>
-    ///     Verifies that Enter (Command.Accept) as a HotKeyBinding works even when another
-    ///     view is focused. This is the asymmetry noted in #4759: DefaultAcceptHandler returns
-    ///     false for plain Views, so Enter leaks through to HotKey dispatch, while Space doesn't.
-    /// </summary>
-    [Fact]
-    public void Enter_HotKey_Fires_When_Sibling_View_Is_Focused ()
-    {
-        IApplication app = Application.Create ();
-        app.Begin (new Runnable<bool> { CanFocus = true });
-
-        // View with Enter HotKeyBinding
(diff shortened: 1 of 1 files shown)


tig:
@mrazza Please verify these changes address #4759 for you. 

I attempted to build tests that replicated what you were doing...

Thanks!!

Terminal.Gui/ViewBase/View.Keyboard.cs
mrazza, line 123: I imagine this should be obvious... or, at least, clear after reading the comments... but I really don't understand why I'd use this (`AddKeyBindingsForHotKey`) over `HotKeyBindings.Add`. How should I think about this?

Tests/UnitTestsParallelizable/ViewBase/Keyboard/HotKeyTests.cs
```diff
@@ -60,35 +58,46 @@ public void AddKeyBindingsForHotKey_Sets (KeyCode key)
         }
     }
 
+    // Claude - Opus 4.5
+    [Fact]
+    public void AddKeyBindingsForHotKey_Accepts_Space ()
```
mrazza, line 63: nit: may be worth either commenting or renaming this so the reason it exists is clearer

to be honest, I don't actually know why this test was added other than the fact that you _used_ to have a check on _not_ accepting Space... but seems strange to write a test validating the non-existence of a specific feature you removed. You _could_ make this a theory where it validates you can AddKeyBindignsForHotKey with any value? Just seems strange to pick Space specifically although I may be missing something.

Tests/UnitTestsParallelizable/ViewBase/Keyboard/KeyBindingsTests.cs
```diff
@@ -151,40 +150,120 @@ public void HotKey_Raises_HandlingHotKey ()
...
+    // Claude - Opus 4.5
+    /// <summary>
+    ///     Replicates issue #4759: A View several levels down has a HotKeyBinding for Key.Space.
+    ///     A *different* focused view (plain View with default Space→Activate binding) swallows
+    ///     the key via DefaultActivateHandler returning true, preventing the HotKey from firing.
+    ///     This test mirrors @mrazza's scenario: the HotKey should fire regardless of which
+    ///     view is focused, unless that view explicitly handles the key.
+    /// </summary>
+    [Fact]
+    public void Space_HotKey_Fires_When_Sibling_View_Is_Focused ()
```
mrazza: So these tests certainly mirror what I'm doing. However, they feel like odd tests. The behavior we're actually trying to validate is that unhandled Accept/Activate commands do not swallow (allow propagation) of the binding to a HotKey.

I would consider making these tests more aligned with the actual behavior that they're validating rather than testing the specific problem I ran into.

(although, tbf, I'm flattered to have a test explicitly added for my use-case)

Same with Enter below.

Terminal.Gui/ViewBase/View.Keyboard.cs
tig: IIRC, I always intended to hide `KeyBindings`, `HotkeyBindings`, and `MouseBindings` behind an API.

But that obv never happened. If you don't use `AddKeyBindingsForHotKey` you don't get automatic modifier support (which is not relevant for `Space`).

Tests/UnitTestsParallelizable/ViewBase/Keyboard/HotKeyTests.cs
tig: Added a comment.

PR #4779, opened 2026-03-02
Fixes #4780 - Wide char continuation cell not flushed when attribute changes

### LINKED ISSUE #4780: Wide character continuation cell attribute not flushed when changed
## Description

When a wide character (2-column glyph) has its attribute changed between frames — for example, transitioning from normal to selected/highlighted — only the first cell updates visually. The second (continuation) cell retains the old attribute appearance.

## Steps to Reproduce

1. Display a wide character (e.g., ☀ U+2600, any CJK character, or emoji) in a View
2. Change the attribute on that cell between frames (e.g., toggle selection highlighting)
3. Observe that only the first cell of the wide character reflects the new attribute; the second cell keeps the old one

## Expected Behavior

Both cells of a wide character should reflect the current attribute.

## Root Cause

In \OutputBufferImpl.AddGrapheme()\, when processing a wide character, the continuation cell (column N+1) has its \Attribute\ set to \CurrentAttribute\ but \IsDirty\ is not set to \	rue\. This was intentional per the fix for #4258 (overlapping View border corruption), but it prevents legitimate attribute changes from being flushed to the terminal.

## Proposed Fix

Mark the continuation cell dirty only when the attribute actually changes (\Contents[Row, Col].Attribute != CurrentAttribute\). This preserves the #4258 fix while ensuring attribute updates are flushed.

See PR #4779.

akrantz:
## Problem

When a wide character's attribute changes between frames (e.g., from normal to selected/highlighted), the continuation cell (column N+1) has its attribute updated but is never marked dirty, so the change is never flushed to the terminal. This causes the second cell of wide characters to visually retain the old attribute.

For example, in a hex viewer with a character panel, selecting a wide emoji (☀ U+2600, 2 cells wide) on the bytes side should cross-highlight both cells on the character side, but only the first cell updates.

## Root Cause

The fix for #4258 (overlapping View border corruption) stopped marking the continuation cell as dirty. However, this also prevents legitimate attribute-only changes from being flushed.

## Fix

Mark the continuation cell dirty **only when the attribute actually changes**. This preserves the #4258 fix (same attribute = not marked dirty = no unnecessary invalidation) while ensuring attribute changes are flushed.

Fixes #4780

## Testing

- All 12 AddRune tests pass
- All 75 Wide glyph tests pass
- All 10 OutputBuffer tests pass


akrantz:
@tig @BDisp Looks like the test runner crashed running the Mac test. Should this just be re-run?

tig:
> @tig @BDisp Looks like the test runner crashed running the Mac test. Should this just be re-run?

Longstanding non-deterministic macos test failure. It can be ignored.

I'd really love to have a unit test added to test what you fixed here. 

Do you think you could add one using the technique found in `AddStr_WideGlyph_Second_Column_Attribute_Outputs_Correctly ()`?

akrantz:
Added the unit test.

akrantz:
@tig @BDisp I think approval is needed to run with the latest update.

PR #4829, opened 2026-03-11
Fixes #4183 - BREAKING CHANGE - Finishes `Adornments` and Rewrites `TabView`

### LINKED ISSUE #2994: Master Issue: Finish Adornments

### LINKED ISSUE #3365: v2 - Null reference in SelectedTab setter during `RemoveTab` call
**Describe the bug**

I am updating designer from v2 pre.250 to v2 pre.710.

It seems there has been some changes to TabView, I am now getting a null reference in RemoveTab (within the setter on SelectedTab).

null-ref-tab-select

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=Terminal.Gui
  StackTrace:>	Terminal.Gui.dll!Terminal.Gui.TabView.SelectedTab.set(Terminal.Gui.Tab value) Line 501	C#
 	Terminal.Gui.dll!Terminal.Gui.TabView.RemoveTab(Terminal.Gui.Tab tab) Line 698	C#
 	TerminalGuiDesigner.dll!TerminalGuiDesigner.TabViewExtensions.ReOrderTabs(Terminal.Gui.TabView tabView, Terminal.Gui.Tab[] newOrder) Line 58	C#
 	TerminalGuiDesigner.dll!TerminalGuiDesigner.Operations.TabOperations.MoveTabOperation..ctor.AnonymousMethod__0_1(Terminal.Gui.TabView v, Terminal.Gui.Tab[] a) Line 23	C#
 	TerminalGuiDesigner.dll!TerminalGuiDesigner.Operations.Generics.MoveOperation<Terminal.Gui.TabView, Terminal.Gui.Tab>.DoImpl() Line 108	C#
 	TerminalGuiDesigner.dll!TerminalGuiDesigner.Operations.Operation.Do() Line 48	C#
 	UnitTests.dll!UnitTests.Operations.MoveTabOperationTests.TestMoveTab_DoUndo.AnonymousMethod__0(TerminalGuiDesigner.Design d, Terminal.Gui.TabView v) Line 57	C#
 	UnitTests.dll!UnitTests.Tests.RoundTrip<Terminal.Gui.Window, Terminal.Gui.TabView>(System.Action<TerminalGuiDesigner.Design, Terminal.Gui.TabView> adjust, out Terminal.Gui.TabView viewOut, string caller) Line 105	C#
 	UnitTests.dll!UnitTests.Operations.MoveTabOperationTests.TestMoveTab_DoUndo(int idxToMove, int adjustment, int expectedNewIndex, bool expectPossible) Line 41	C#


```

**To Reproduce**
Run `TestMoveTab_DoUndo` in TGD

**Expected behavior**
No crash

### LINKED ISSUE #3407: TabView: Refactor `Border` to use subviews for title and lines, supporting a `Tab` style
## Related:

- #2814 

## Requires

- #2995 
- #3014
- #2114
- #2814 

## Will enable

- #2537

# Design

## Subviews

```cs
public Label TitleLabel { get; set; };
public Button CloseButton { get ; set; }
public Line Top { get; set; }
public Line Left { get; set; }
public Line Right{ get; set; }
public Line Bottom{ get; set; }
```

## Layout

This also indicates the order the views should be added; The label & button must be added after the lines so that they occlude them.

```cs
Top.X = Pos.Function (Thickness.Left/ 2);
Top.Y = Pos.Function (Thickness.Top / 2);
Top.Width = Dim.Fill () - Dim.Function (Thickness.Right / 2);

Left.X = Pos.Left (Top);
Left.Y = Pos.Top (Top);
Left.Height = Dim.Fill () - Dim.Function (Thickness.Bottom / 2);

Right.X = Pos.Right (Top);
Right.Y = Pos.Top (Top);
Right.Height = Dim.Fill () - Dim.Function (Thickness.Bottom / 2);

Bottom.X = Pos.Left (Top);
Bottom.Y= Pos.Bottom (Left);
Bottom.Width = Dim.Fill () - Dim.Function (Thickness.Right / 2);

TitleLabel.X = Pos.Right (Left) + 1;
TitleLabelY = Pos.Top (Top);

CloseButton .X = Pos.Left (Right) - 1;
CloseButton .Y = Pos.Top (Top);
```

## Example usage:

```cs
view.Border.TitleLabel.X = Pos.Center (); // centers title (default is X = 1)
view.Border.TitleLabel.Visible = false; // Turn off the Title.
````

### LINKED ISSUE #3875: Add left and right tab sides with alignments to the TabView.
`TabView` lacks on not allowing tabs on left and on right and would be great to implement them with the alignments addition.

Just implement them improving the current API.

Only the one I want to implement.

**Additional context**

WindowsTerminal_AOY4ZHrtWX

### LINKED ISSUE #4183: Refactor/rewrite `TabView` to use nested subviews
These built-in `Views` have been introduced in v2 or re-written in v2 to remove massive amounts of complex custom code and complexity by leveraging the fact that v2 has better infrastructure for compound/nested views.

- `Menuv2` etc...
- `Shortcut`
- `StatusBar`
- `OptionSelector` (RadioGroup)
- `FlagSelector`
- `ColorPicker`
- `ScrollBar`
- `DatePicker`
- `NumericUpDown`

These `Views` remain using mostly the v1-legacy model and thus are good candidates for cleanup. Listed in priority order based on how I perceive how much bang-for-the-buck there is in re-working them. 

- `TabView`
- `ComboBox`
- `Slider`
- `TableView`
- `TreeView`

I think we should tackle `TabView` because:

- `TabView` is a great control for testing the library. It has a lot of capabilities that stretch the library. The v1 version was internally very complex in order to work around limitations in TG. In theory, v2 should enable making this control a lot simpler. 
- Like `Shortcut`, `StatusBar`, and `Menuv2` it is a container for arbitrary other Views. This means it can both take advantage of and prove out things like `Command` propagation, Content Scrolling,, `KeyBindings`, `MouseBindings`, `Adornments`, `ViewArrangement`, and so forth.
- It is a bug-farm. The code is full of duplicative code, complex switch statements, and complex overrides and interactions that can and should be simplified.

I had a PR going (#3832) where I experimented with this but gave up to focus on other things. In it, I suggested a design incorporating things like this:

- Content scrolling - The TabRowView can/should just be a horizontally scrolling View where each Tab is just a subview. All the complex TabView logic for renering the Tabs and supporting scrolling can go away.
- Layout - Dim.Auto, deferred layout, etc... should reduce the need for a lot of the internal complexity. Especially the complex layout math scattered around.
- Drawing - There should be no need for TabView to draw lines directly. Each element can just be a View and leverage `Border` an automatic line joins currently enabled by `SuperViewRendersLineCanvas`
- Navigation - How a user navs with the keyboard across Tabs, TabRowView, and the tab content is complex. The new v2 Nav system should make this easier.

I think we should revisit this and I hope someone besides me finds the motivation to do so!

### LINKED ISSUE #4858: Region.Combine: zero-width/zero-height rectangles corrupt Union results
## Bug

Zero-width and zero-height rectangles (e.g., `{X=0, Y=0, Width=0, Height=2}`) are not filtered out during `Region.Combine` operations. This causes the sweep-line algorithm in `MergeRectangles` to produce incorrect results, expanding rectangles to cover areas they should not.

### Observed Symptom

In `View.Drawing.AddDrawnRegionForAdornment`, the `exclusion` region for a `Border` adornment incorrectly covers cells that were never drawn. A rectangle that should be `{X=2, Y=1, W=2, H=1}` (row 1 only) becomes `{X=2, Y=0, W=2, H=2}` (rows 0-1), causing transparency artifacts.

### Reproduction Path

A `TabView` subview with `TabOffset = -1` causes its Border's `LineCanvas` to render one column to the left of the view's origin at `X = -1`. When `DoRenderLineCanvas` returns the drawn region, it includes a rectangle at `{X=-1, Y=0, W=1, H=2}`. This region is then intersected with the adornment's frame `{X=0, Y=0, W=4, H=2}`, which clips it to `{X=0, Y=0, W=0, H=2}` - a zero-width rectangle.

This zero-width rectangle is kept because `Rectangle.IsEmpty` returns `false` (it only returns `true` when **all four fields** are zero). The rectangle then enters the Union operation, where it poisons the sweep-line merge.

### Root Cause

Two sites in `Region.cs` use `Rectangle.IsEmpty` to filter degenerate rectangles, but `.IsEmpty` is insufficient:

**Site 1 - `MergeRectangles`:** A zero-width rect like `{0,0,0,2}` has `Left == Right == 0`, generating start and end events at the same X. The sort order places the end event first, but since the interval isn't in `activeCounts` yet, the decrement is silently dropped. The start event then creates a phantom +1 that propagates through subsequent X groups, corrupting the output.

**Site 2 - `CombineInternal` Intersect case:** `Rectangle.Intersect` can produce zero-area results when input rectangles share only an edge. These pass the `!intersected.IsEmpty` check and are stored in the region.

### Fix

Both sites should check for positive area instead of relying on `Rectangle.IsEmpty`:

**Site 1 - `MergeRectangles`:**
```csharp
// Before:
if (r.IsEmpty) { continue; }
// After:
if (r.Width <= 0 || r.Height <= 0) { continue; }
```

**Site 2 - `CombineInternal` Intersect case:**
```csharp
// Before:
if (!intersected.IsEmpty)
// After:
if (intersected.Width > 0 && intersected.Height > 0)
```

Six new tests in `ZeroAreaRectangleTests.cs` cover the bug at both levels (`MergeRectangles`, `Intersect`, and end-to-end `Union`).

tig:
## Overview

This PR completes the v2 Adornments implementation (#2994) by replacing the old `TabView` (~1,500 lines across 6 files of manual drawing code) with a new `Tabs` class built entirely on the v2 infrastructure: `BorderSettings.Tab`, `SuperViewRendersLineCanvas`, overlapped `ViewArrangement`, and the `IValue<View?>` pattern.

WindowsTerminal_OUfqSsjg41

### The Old TabView Problem

The original `TabView` contained ~800 lines of manual `LineCanvas` drawing in `TabRow.OnRenderingLineCanvas()`, hand-computing line positions, corners, and intersections that the LineCanvas auto-join system was designed to handle automatically. It also had manual viewport calculation, custom mouse/keyboard handling outside the Command system, a separate `TabStyle` configuration class, and a `Tab.View`/`Tab.DisplayText` split.

### What Changed

**New `Tabs` class** — Any `View` can be a tab. Add views via `View.Add()` and the container automatically configures each with `BorderSettings.Tab`, `ViewArrangement.Overlapped`, and a `Border.Thickness` derived from `TabDepth` and `TabSide`. The selected tab is the focused one, reported via `IValue<View?>.Value`. Zero custom line drawing — the entire visual (tab headers, separator lines, T-junctions, corners) is produced by standard `Border` rendering + `LineCanvas` auto-joins.

**`BorderSettings.Tab`** — A new border mode that draws a tab-style header on one side of any View's border. Configured via `Border.TabSide`, `Border.TabOffset`, and `Border.TabLength`. This is a general-purpose capability — any View can have a tab border, not just views inside a `Tabs` container.

**`BorderView` tab rendering pipeline** — New code in `BorderView` handles the tab header drawing: computing header rectangles, drawing separator lines with gaps for the selected tab, rendering the title inside the tab header via `TabTitleView`, and managing line canvas intersections. All done through the existing `Border`/`LineCanvas` infrastructure.

**Four-side tab support** — `TabSide` supports `Top`, `Bottom`, `Left`, and `Right`. Tab depth is configurable via `TabDepth`.

**Scroll buttons** — When tabs overflow the available space, `ScrollButton` indicators (◄►/▲▼) appear on the separator line. `ScrollOffset` provides programmatic scrolling with clamping.

**`ScrollButton` control** — Extracted from `ScrollBar` into a standalone 1×1 button view with directional glyphs and mouse-hold repeat. Used by both `ScrollBar` and `Tabs`.

**Mouse hold repeat fix** — `MouseHoldRepeaterImpl.RaiseMouseIsHeldDownTick` now stops the repeat timer when the view becomes invisible or disabled, preventing runaway timers when scroll buttons are hidden.

### Drawing Pipeline Changes

- `View.Drawing.cs` was split into `View.Drawing.cs`, `View.Drawing.Adornments.cs`, and `View.Drawing.LineCanvas.cs` for clarity
- Adornment SubView LineCanvas auto-join pipeline was fixed (#4855): `DoDrawAdornmentsSubViews` now runs before `DoRenderLineCanvas` so merged lines are present when the LC renders
- `Region.Combine` now filters zero-width/zero-height rectangles (#4858) that corrupted Union results

### Breaking Changes

- `TabView`, `Tab`, `TabRow`, `TabStyle`, `TabChangedEventArgs`, `TabMouseEventArgs` are all deleted
- Replaced by `Tabs` (the container) — tabs are just `View` instances added via `Add()`
- `InsertTab(index, view)` for positional insertion; `TabCollection` for logical-order enumeration
- Selection via `Value` property (`IValue<View?>`) instead of `SelectedTab`/`SelectedTabIndex`

### Scenarios

- `GraphViewExample` refactored to use `Tabs` — each graph type is a tab instead of cycling via Ctrl+G
- New `TabsExample` scenario replacing the old `TabViewExample`
- `AdornmentsEditor`, `Notepad`, `ConfigurationEditor`, and other scenarios updated

## Fixes

- Fixes #4855 - **DONE**
- Adds `BorderSettings.Tab` which enables any View's border to have top/left/right/bottom "tab" like borders - **DONE**
- Fixes #4858 - **DONE**
- Fixes #3407 - **DONE**
- Fixes #2994 - **DONE**
- Fixes #4183
- Fixes #3875
- Fixes #3365

### DIFF
--- a/Terminal.Gui/ViewBase/View.Content.cs
+++ b/Terminal.Gui/ViewBase/View.Content.cs
@@ -102,16 +102,26 @@ public partial class View
         GetAutoWidth () - GetAdornmentsThickness ().Horizontal;
 
     /// <summary>
-    ///     Gets the natural (auto-size) width of the view determined by Dim.Auto, which is the minimum of the size required to
-    ///     fit all subviews and the container size.
+    ///     Gets the natural (auto-size) width of the view as calculated by <see cref="Dim.Auto ()"/>.
     /// </summary>
+    /// <remarks>
+    ///     The returned width is the full auto-calculated width for the view, including adornment thickness.
+    ///     Unlike <see cref="GetWidthRequiredForSubViews ()"/>, this value is not content-only.
+    ///     The calculation may also respect minimum and maximum content constraints applied by the auto dimension logic
+    ///     before adornments are added.
+    /// </remarks>
     /// <returns></returns>
     public int GetAutoWidth () => Dim.Auto ().Calculate (0, GetContainerSize ().Width, this, Dimension.Width);
 
     /// <summary>
-    ///     Gets the natural (auto-size) height of the view determined by Dim.Auto, which is the minimum of the size required
-    ///     to fit all subviews and the container size.
+    ///     Gets the natural (auto-size) height of the view as calculated by <see cref="Dim.Auto ()"/>.
     /// </summary>
+    /// <remarks>
+    ///     The returned height is the full auto-calculated height for the view, including adornment thickness.
+    ///     Unlike <see cref="GetHeightRequiredForSubViews ()"/>, this value is not content-only.
+    ///     The calculation may also respect minimum and maximum content constraints applied by the auto dimension logic
+    ///     before adornments are added.
+    /// </remarks>
     /// <returns></returns>
     public int GetAutoHeight () => Dim.Auto ().Calculate (0, GetContainerSize ().Height, this, Dimension.Height);
 


tig:
@copilot expand the TabCompositionTests to cover all 4 sides and tab thicknesses of 1, 2, 3.

tig:
This is close...

WindowsTerminal_Rfcywx2ISa

tig:
I'm currently working on making keyboard nav work properly, BTW. Other than that, which is goign to require some refactoring of both teh Tabs code and Adornment related Navigation code, this is def worth y'all reviewing.

BDisp (approved):
Looks good to me. Incredible.

YourRobotOverlord:
> ...this is def worth y'all reviewing.

Sorry it's taking a minute, I'm just wrapping my head around how it all works.

Terminal.Gui/Views/Tabs.cs
```diff
@@ -0,0 +1,886 @@
...
+
+        // Configure the view as a tab
+        view.TabStop = TabBehavior.TabStop;
+        view.CanFocus = true;
+        view.BorderStyle = _tabLineStyle;
+        view.Border.Settings = BorderSettings.Tab | BorderSettings.Title;
+        view.Border.TabSide = _tabSide;
+        view.Arrangement = ViewArrangement.Overlapped;
+        view.Width = Dim.Fill ();
+        view.Height = Dim.Fill ();
```
YourRobotOverlord, line 380: Nit: non-obvious side effect, maybe mention in the docs that any dimensions the user set will be overridden. It is obvious that this is going to happen, but I can foresee users reusing the same view (in a dialog for example) where the dimensions are **not** overridden.
Actually, I think all the things that might be set by the user then overridden here should be mentioned in the documentation; here it only mentions that it sets Border.Settings and Arrangement.

tig:
> This is great! I did notice a couple of things while testing it:
> 
> 1. Tabs overlap container border when scrolling:
> 
> Image

This is a bug in adornment linecanvas rendering that I hope to fix in this PR. However, it's unlikely that any real uses will enable a border to `Tabs`... The Scenario does it so that I can resize the View while testing...

> 2. Tab rendering breaks when setting Line Style to None then back to another style:

Good catch. Thanks!

BDisp:
Why not consider placing a T at the top of the tabs instead of a space appearing, and using the same procedure for other types of sides?

tig:
> Why not consider placing a T at the top of the tabs instead of a space appearing, and using the same procedure for other types of sides?

Three reasons:

1) There's no "round" T and thus it breaks the look of the tab header by changing what was round to square. I think this looks better. 

2) In the future I plan on enabling .TabAttribute which will let each tab have a distinct line and text attribute (just the header). In my experience prototyping this, it looks like crap if the header isn't distinct. 

3) Auto line join doesn't support it. We'd have to have custom logic to make it happen.

BDisp:
> Three reasons:
> 
> 1. There's no "round" T and thus it breaks the look of the tab header by changing what was round to square. I think this looks better.

I know it doesn't have one, because that's what I found when I worked on it. Using only square unions (┬, ┴, ├, ┤), with their respective single, double, and dotted line versions, it doesn't look so awful. Personally, I don't like seeing that space, but it's a matter of personal choice.

> 2. In the future I plan on enabling .TabAttribute which will let each tab have a distinct line and text attribute (just the header). In my experience prototyping this, it looks like crap if the header isn't distinct.

I also fully agree, allowing for configurability. Personally, I don't like seeing that space, but it's a matter of personal choice.

> 3. Auto line join doesn't support it. We'd have to have custom logic to make it happen.

I vote that the corresponding square join be assigned by default, as the remaining curved lines always give the impression of a curved edge. Using a square join between two curved edges is practically imperceptible, unlike the existing space.

tig:
I've added full keyboard nav support and refactored to improve the look and feel, as well as maintainability of the code. 

Pull this down locally and play with it please. 

WindowsTerminal_4oDcnlIlpU

I plan on merging this later today.l

YourRobotOverlord:
I love how configurable these are!

I also feel that the Spacing parameter is a wonderful option for @BDisp's concerns:
> Why not consider placing a T at the top of the tabs instead of a space appearing, and using the same procedure for other types of sides?

image

I did notice one unexpected behavior where the Title is not visible and becomes un-clickable when the Side is set to Right or Bottom:
WindowsTerminal_Kp5O3ZW6mi

BDisp:
I love this very much too.

image

tig:
> I did notice one unexpected behavior where the Title is not visible and becomes un-clickable when the Side is set to Right or Bottom:

 [ WindowsTerminal_Kp5O3ZW6mi ](https://private-user-images.githubusercontent.com/47068588/575457678-fc166f25-efc8-4fa6-9f93-838a174f6b39.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzU3MDUzMTksIm5iZiI6MTc3NTcwNTAxOSwicGF0aCI6Ii80NzA2ODU4OC81NzU0NTc2NzgtZmMxNjZmMjUtZWZjOC00ZmE2LTlmOTMtODM4YTE3NGY2YjM5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA0MDklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNDA5VDAzMjMzOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTc1NDU4ZDFhYTk2MTFhMTU1NGY0OTZlMjhiY2Q3MWZlNTY3MWVlYmY0YTFhZjMyZDEzYWU3ZTgyNWVkOWYyNzEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.Vgc4ie7D7BVqCny5v1osYuPancxScZ1fxdX1GJ4LBbk) [ ](https://private-user-images.githubusercontent.com/47068588/575457678-fc166f25-efc8-4fa6-9f93-838a174f6b39.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzU3MDUzMTksIm5iZiI6MTc3NTcwNTAxOSwicGF0aCI6Ii80NzA2ODU4OC81NzU0NTc2NzgtZmMxNjZmMjUtZWZjOC00ZmE2LTlmOTMtODM4YTE3NGY2YjM5LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA0MDklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNDA5VDAzMjMzOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTc1NDU4ZDFhYTk2MTFhMTU1NGY0OTZlMjhiY2Q3MWZlNTY3MWVlYmY0YTFhZjMyZDEzYWU3ZTgyNWVkOWYyNzEmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.Vgc4ie7D7BVqCny5v1osYuPancxScZ1fxdX1GJ4LBbk)

I can repro this. Happens whenever a subview of the tab exceeds the viewport. Eg.

WindowsTerminal_SS4tMD6DR6

Issue raised here: https://github.com/gui-cs/Terminal.Gui/issues/4915

PR #4863, opened 2026-03-27
Partially Fixes #4522 - Stale `ContentSize` capture

### LINKED ISSUE #4522: `NeedsLayout` is buggy

tig:
## Fixes

- Partially Addresses #4522

### DIFF
--- a/Examples/UICatalog/Scenarios/DropDownListExample.cs
+++ b/Examples/UICatalog/Scenarios/DropDownListExample.cs
@@ -7,6 +7,17 @@ namespace UICatalog.Scenarios;
 [ScenarioCategory ("Controls")]
 public class DropDownListExample : Scenario
 {
+    private enum DayOfWeekShort
+    {
+        Monday,
+        Tuesday,
+        Wednesday,
+        Thursday,
+        Friday,
+        Saturday,
+        Sunday
+    }
+
     private IApplication? _app;
     private EventLog? _eventLog;
 
@@ -87,7 +98,7 @@ public class DropDownListExample : Scenario
             Title = "Select Color",
             Source = new ListWrapper<string> (colors),
             ReadOnly = true,
-            Text = "Red"
+            Value = colors [0]
         };
 
         readOnlyDropDown.ValueChanged += (s, e) => { _eventLog?.Log ($"ReadOnly: ValueChanged - OldValue: '{e.OldValue}', NewValue: '{e.NewValue}'"); };
@@ -117,8 +128,18 @@ public class DropDownListExample : Scenario
                                               _eventLog?.Log ($"Editable: ValueChanging - CurrentValue: '{e.CurrentValue}', NewValue: '{e.NewValue}'");
                                           };
 
+        // Enum-typed dropdown (DropDownList<TEnum>) example
+        Label enumLabel = new () { X = 1, Y = Pos.Bottom (editableDropDown) + 1, Text = "Enum Mode (DropDownList<TEnum>):" };
+
+        DropDownList<DayOfWeekShort> enumDropDown = new ()
+        {
+            X = Pos.Right (enumLabel) + 1, Y = Pos.Top (enumLabel), Title = "Select Day", Value = DayOfWeekShort.Monday
+        };
+
+        enumDropDown.ValueChanged += (s, e) => { _eventLog?.Log ($"Enum: ValueChanged - NewValue: '{e.Value}'"); };
+
         // Different positions example
-        Label positionLabel = new () { X = 1, Y = Pos.Bottom (editableDropDown) + 1, Text = "Different Positions (with Borders and Resizable right sides):" };
+        Label positionLabel = new () { X = 1, Y = Pos.Bottom (enumDropDown) + 1, Text = "Different Positions (with Borders and Resizable right sides):" };
 
         DropDownList topLeftDropDown = new ()
         {
@@ -183,6 +204,8 @@ public class DropDownListExample : Scenario
                           readOnlyDropDown,
                           editableLabel,
                           editableDropDown,
+                          enumLabel,
+                          enumDropDown,
                           positionLabel,
                           topLeftDropDown,
                           topRightDropDown,
--- a/Terminal.Gui/App/Mouse/ApplicationMouse.cs
+++ b/Terminal.Gui/App/Mouse/ApplicationMouse.cs
@@ -44,6 +44,15 @@ internal class ApplicationMouse : IMouse, IDisposable
     private IPopoverView? _dismissedByMousePress;
 
     /// <summary>
+    ///     The active popover at the time of the most recent mouse press. Used to detect when a popover
+    ///     closes programmatically during the press → release → click cycle (e.g., when a ListView item
+    ///     selection hides the popover on Released). If this was set during Pressed but the popover is no
+    ///     longer active at Clicked time, the Clicked event is suppressed to prevent it from leaking to
+    ///     views below the now-hidden popover.
+    /// </summary>
+    private IPopoverView? _activePopoverAtPress;
+
+    /// <summary>
     ///     Tracks whether <see cref="RaiseMouseEvent"/> is currently executing within the
     ///     dismiss recursion, to avoid clearing <see cref="_dismissedByMousePress"/> prematurely.
     /// </summary>
@@ -96,6 +105,23 @@ internal class ApplicationMouse : IMouse, IDisposable
 
         if (mouseEvent.Handled)
         {
+            return;
+        }
+
+        // Record the active popover when a press starts so we can detect if it closes mid-cycle.
+        if (mouseEvent.IsPressed)
+        {
+            _activePopoverAtPress = App?.Popovers?.GetActivePopover ();
+        }
+
+        // Suppress Clicked events that were synthesized after a popover closed during the same
+        // press → release → click cycle. Without this, selecting an item in a popover (which
+        // hides the popover on Released) causes the Clicked event to leak to views below.
+        if (mouseEvent.IsSingleDoubleOrTripleClicked && _activePopoverAtPress is { } && App?.Popovers?.GetActivePopover () != _activePopoverAtPress)
+        {
+            Trace.Mouse ("app", mouseEvent.Flags, mouseEvent.ScreenPosition, "Popovers", "Suppressing Clicked - popover closed mid-cycle");
+            _activePopoverAtPress = null;
+
             return;
         }
 
--- a/Terminal.Gui/Views/DropDownListTEnum.cs
+++ b/Terminal.Gui/Views/DropDownListTEnum.cs
@@ -1,0 +1,70 @@
+using System.Collections.ObjectModel;
+
+namespace Terminal.Gui.Views;
+
+/// <summary>
+///     A type-safe dropdown control for selecting a single value from an enum.
+///     Provides the same interface as <see cref="OptionSelector{TEnum}"/> but rendered as a compact dropdown list.
+/// </summary>
+/// <typeparam name="TEnum">The enum type to display and select from.</typeparam>
+/// <remarks>
+///     <para>
+///         <see cref="DropDownList{TEnum}"/> and <see cref="OptionSelector{TEnum}"/> are interchangeable when the data
+///         source is an enum. Both expose a typed <see cref="Value"/> property and a typed
+///         <see cref="ValueChanged"/> event. They differ only in appearance: <see cref="DropDownList{TEnum}"/> is a
+///         compact single-line dropdown, while <see cref="OptionSelector{TEnum}"/> shows all options as checkboxes.
+///     </para>
+///     <para>
+///         The dropdown source is automatically populated from <typeparamref name="TEnum"/> values in the constructor.
+///     </para>
+///     <para>
+///         <b>Usage Example:</b>
+///     </para>
+///     <code>
+///         var dropdown = new DropDownList&lt;DayOfWeek&gt; ();
+///         dropdown.Value = DayOfWeek.Monday;
+///         dropdown.ValueChanged += (s, e) => Console.WriteLine ($"Selected: {e.Value}");
+///     </code>
+/// </remarks>
+public sealed class DropDownList<TEnum> : DropDownList, IValue where TEnum : struct, Enum
+{
+    /// <summary>
+    ///     Initializes a new instance of the <see cref="DropDownList{TEnum}"/> class.
+    ///     Automatically populates the dropdown source from <typeparamref name="TEnum"/> values.
+    /// </summary>
+    public DropDownList ()
+    {
+        Source = new ListWrapper<string> (new ObservableCollection<string> (Enum.GetValues<TEnum> ().Select (enumValue => enumValue.ToString ())));
+
+        base.ValueChanged += OnBaseValueChanged;
+    }
+
+    private void OnBaseValueChanged (object? sender, ValueChangedEventArgs<string?> e)
+    {
+        TEnum? newValue = null;
+
+        if (!string.IsNullOrEmpty (e.NewValue) && Enum.TryParse (e.NewValue, out TEnum parsed))
+        {
+            newValue = parsed;
+        }
+
+        ValueChanged?.Invoke (this, new EventArgs<TEnum?> (newValue));
+    }
+
+    /// <summary>
+    ///     Gets or sets the currently selected enum value.
+    /// </summary>
+    public new TEnum? Value
+    {
+        get => !string.IsNullOrEmpty (Text) && Enum.TryParse (Text, out TEnum result) ? result : null;
+        set => base.Value = value?.ToString ();
+    }
+
+    /// <summary>
+    ///     Raised when the selected value changes. Provides the new value as <typeparamref name="TEnum"/>?.
+    /// </summary>
+    public new event EventHandler<EventArgs<TEnum?>>? ValueChanged;
+
+    /// <inheritdoc/>
+    object? IValue.GetValue () => Value;
+}
--- a/Tests/UnitTestsParallelizable/Views/DropDownListTests.cs
+++ b/Tests/UnitTestsParallelizable/Views/DropDownListTests.cs
@@ -1,8 +1,10 @@
 using System.Collections.ObjectModel;
+using Terminal.Gui.Tracing;
+using UnitTests;
 
 namespace ViewsTests;
 
-public class DropDownListTests
+public class DropDownListTests (ITestOutputHelper output)
 {
     [Fact]
     public void AltDown_OpensDropdown ()
@@ -691,7 +693,237 @@ public class DropDownListTests
 
     // Claude - Opus 4.5
 
+    /// <summary>
+    ///     Regression test: clicking on a list item in the popover should close the popover
+    ///     and select the item, but the synthesized Click event must NOT flow through to views underneath.
+    ///     The bug: Released on the ListView fires Activate → popover closes. The subsequent synthesized
+    ///     Clicked event then finds no popover and falls through to whatever view is below.
+    /// </summary>
+    [Fact]
+    public void ClickListItem_DoesNotActivateViewBelow ()
+    {
+        // Claude - Opus 4.6
+        using IDisposable logging = TestLogging.Verbose (output, TraceCategory.Mouse | TraceCategory.Command);
+        using IApplication app = Application.Create ();
+        app.Init (DriverRegistry.Names.ANSI);
+        app.Driver!.SetScreenSize (40, 15);
+
+        using Runnable top = new ();
+        SessionToken? token = app.Begin (top);
+
+        // DropDownList at the top of the screen
+        ObservableCollection<string> items = ["Apple", "Banana", "Cherry"];
+
+        DropDownList dropdown = new ()
+        {
+            X = 0,
+            Y = 0,
+            Width = 20,
+            Source = new ListWrapper<string> (items),
+            ReadOnly = true
+        };
+
+        top.Add (dropdown);
+        top.Layout ();
+
+        // Give focus to the dropdown and open the popover via key injection
+        dropdown.SetFocus ();
+        app.InjectKey (Key.F4);
+
+        IPopoverView? popover = FindDropDownPopover (app);
+        Assert.NotNull (popover);
+        Assert.True (popover.Visible, "Popover should be open");
+
+        // Track whether ANY mouse Click event leaks to the Runnable after the popover closes.
+        // The Clicked event at the popover item position should NOT reach views below.
+        var runnableReceivedClick = false;
+
+        top.MouseEvent += (_, e) =>
+                          {
+                              if (e.Flags.HasFlag (MouseFlags.LeftButtonClicked))
+                              {
+                                  runnableReceivedClick = true;
+                              }
+                          };
+
+        // Click on the third item ("Cherry") at screen Y=3
+        // (screen Y=1 = item 0 "Apple", Y=2 = item 1 "Banana", Y=3 = item 2 "Cherry")
+        Point clickPos = new (2, 3);
+
+        DateTime baseTime = new (2025, 1, 1, 12, 0, 0);
+        InputInjectionOptions options = new () { Mode = InputInjectionMode.Direct };
+        IInputInjector injector = app.GetInputInjector ();
+
+        // Simulate click via input injection: Pressed → Released (system synthesizes Click)
+        injector.InjectMouse (new Mouse { ScreenPosition = clickPos, Flags = MouseFlags.LeftButtonPressed, Timestamp = baseTime }, options);
+
+        injector.InjectMouse (
+                              new Mouse { ScreenPosition = clickPos, Flags = MouseFlags.LeftButtonReleased, Timestamp = baseTime.AddMilliseconds (50) },
+                              options);
+
+        // The popover should have closed and the item should be selected
+        Assert.False (popover.Visible, "Popover should be closed after clicking a list item");
+        Assert.Equal ("Cherry", dropdown.Text);
+
+        // CRITICAL: The synthesized Clicked event should NOT leak through to the Runnable
+        Assert.False (runnableReceivedClick, "Runnable should NOT receive the Clicked event that was synthesized after the popover closed");
+
+        app.End (token!);
+    }
+
     // Helper to find the DropDownList popover (excludes the context menu popover)
     private static IPopoverView? FindDropDownPopover (IApplication app) =>
         app.Popovers?.Popovers.OfType<Popover<ListView, string?>> ().FirstOrDefault ();
+}
+
+public class DropDownListGenericTests
+{
+    private enum Season
+    {
+        Spring,
+        Summer,
+        Autumn,
+        Winter
+    }
+
+    [Fact]
+    public void Constructor_PopulatesSourceFromEnum ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+
+        Assert.NotNull (dropdown.Source);
+        Assert.Equal (4, dropdown.Source.Count);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void Constructor_SourceContainsEnumNames ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+
+        System.Collections.IList items = dropdown.Source!.ToList ();
+        List<string?> names = items.Cast<object?> ().Select (i => i?.ToString ()).ToList ();
+        Assert.Contains ("Spring", names);
+        Assert.Contains ("Summer", names);
+        Assert.Contains ("Autumn", names);
+        Assert.Contains ("Winter", names);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void Value_Get_ReturnsNullWhenTextIsEmpty ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+
+        Assert.Null (dropdown.Value);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void Value_Set_UpdatesText ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+
+        dropdown.Value = Season.Summer;
+
+        Assert.Equal ("Summer", dropdown.Text);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void Value_Get_ReturnsCorrectEnum ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+        dropdown.Text = "Autumn";
+
+        Assert.Equal (Season.Autumn, dropdown.Value);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void Value_SetNull_ClearsText ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+        dropdown.Value = Season.Winter;
+
+        dropdown.Value = null;
+
+        Assert.Equal (string.Empty, dropdown.Text);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void ValueChanged_RaisedWithTypedEnum ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+
+        Season? receivedValue = null;
+        dropdown.ValueChanged += (_, e) => { receivedValue = e.Value; };
+
+        dropdown.Value = Season.Spring;
+
+        Assert.Equal (Season.Spring, receivedValue);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void ValueChanged_RaisedNullWhenTextCleared ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+        dropdown.Value = Season.Winter;
+
+        Season? receivedValue = Season.Winter; // set to non-null sentinel
+        dropdown.ValueChanged += (_, e) => { receivedValue = e.Value; };
+
+        dropdown.Value = null;
+
+        Assert.Null (receivedValue);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void GetValue_ReturnsBoxedEnum ()
+    {
+        // Copilot
+        DropDownList<Season> dropdown = new ();
+        dropdown.Value = Season.Autumn;
+
+        object? boxed = ((IValue)dropdown).GetValue ();
+
+        Assert.Equal (Season.Autumn, boxed);
+
+        dropdown.Dispose ();
+    }
+
+    [Fact]
+    public void IsInterchangeableWithOptionSelector_SameValueInterface ()
+    {
+        // Copilot - demonstrates that both implement IValue with the same TEnum type
+        DropDownList<Season> dropDown = new ();
+        OptionSelector<Season> optionSelector = new ();
+
+        dropDown.Value = Season.Summer;
+        optionSelector.Value = Season.Summer;
+
+        Assert.Equal (optionSelector.Value, dropDown.Value);
+
+        dropDown.Dispose ();
+        optionSelector.Dispose ();
+    }
 }
--- a/plans/fix-popover-click-event-leak.md
+++ b/plans/fix-popover-click-event-leak.md
@@ -1,0 +1,155 @@
+# Fix: Popover Click Event Leaks to Views Below After Item Selection
+
+## Problem
+
+When a user clicks on a list item in a DropDownList popover (or any popover that closes on `Released`), the synthesized `Clicked` event leaks through to views underneath.
+
+### Root Cause
+
+The mouse event cycle is `Pressed -> Released -> Clicked` (where `Clicked` is synthesized from `Press+Release`). The bug occurs because:
+
+1. **Pressed** at (x,y) -> dispatched to ListView inside visible popover -> not consumed (no Press binding)
+2. **Released** at (x,y) -> dispatched to ListView -> fires `Command.Activate` via mouse binding -> `CommandBridge` routes to `DropDownList.OnActivated` -> **popover closes** (`_listPopover.Visible = false`)
+3. **Clicked** (synthesized) at (x,y) -> popover is now **hidden** -> `GetViewsUnderLocation` no longer finds the popover -> event falls through to Runnable and its SubViews below
+
+The dismiss-by-clicking-outside mechanism (`ApplicationMouse.cs:108-129`) handles the case where a click **outside** a popover dismisses it and intentionally recurses the event to views below. But when a popover closes **internally** (via `Visible = false` during `Released` processing), there is no equivalent guard to suppress the subsequent `Clicked` event.
+
+### Evidence
+
+Trace output from the failing test `ClickListItem_DoesNotActivateViewBelow`:
+
+```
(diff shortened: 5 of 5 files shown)


duskfold:
@tig Thanks for looking into this! I tested your PR against our dashboard app (Dialog with buttons, TextViews, checkboxes, scrollable content).

Your stale ContentSize re-read fix is correct and necessary. However, a second bug remained: repeated maximize/restore cycles caused Dialog content to shrink by 2px per cycle.

I traced it with instrumented builds and found the root cause is in `View.ScrollBars.cs`: the `VisibleChanged` handlers use relative `Padding.Thickness +1/-1` adjustments that compound during layout re-entry. Each maximize/restore cycle, `ShowHide` fires multiple times, and the padding ratchets upward permanently.

I've opened #4864 with the complete fix (builds on your changes + adds the scrollbar idempotency fix + tests). The trace data and full analysis are in that PR description.

BDisp (approved):

PR #4864, opened 2026-03-28
Fix scrollbar padding accumulation + DimAuto content size drift (#4522)

duskfold:
Builds on top of @tig's #4863 (stale ContentSize re-read). That fix is correct and necessary, but a second bug remained: repeated maximize/restore cycles caused Dialog content to shrink by 2px per cycle until the content area collapsed.

## Root cause (traced with instrumented builds)

Two bugs working together:

### 1. ScrollBar padding accumulation (View.ScrollBars.cs)

The VisibleChanged handlers use relative Padding.Thickness +1/-1 adjustments. During layout re-entry (which happens on every resize), ShowHide fires multiple times per visibility change. Each fire adds another +1, and the -1 on hide doesn't fully compensate because the baseline has already shifted.

Trace data showing Padding.Bottom ratcheting before fix:

    Cycle 1: Padding.Bottom = 3 -> 6
    Cycle 2: Padding.Bottom = 6 -> 8  
    Cycle 3: Padding.Bottom = 8 -> 10
    Cycle 4: Padding.Bottom = 10 -> 12

Fix: Track whether scrollbar padding has been applied via boolean flags (_verticalScrollBarPaddingApplied / _horizontalScrollBarPaddingApplied). The handlers become idempotent: +1 only fires when transitioning from not-applied to applied, -1 only when transitioning back.

Trace data after fix:

    Every cycle: Padding.Bottom = 3 -> 4 -> 3 (stable)

### 2. _minimumSubViewsSize floor in non-DimAuto dialogs (DialogTResult.cs)

UpdateSizes unconditionally floors ContentSize at _minimumSubViewsSize. For fixed-size dialogs (non-DimAuto), this captures the high-water mark from the maximized state and never shrinks back, inflating ContentSize beyond the actual Viewport on restore.

Fix: Only use _minimumSubViewsSize as a floor for DimAuto dialogs (which need to grow to fit content). Fixed-size dialogs use Viewport size directly.

## Tests

Added ScrollBarPaddingAccumulationTests.cs with two tests:
- Dialog_Resize_Cycles_Do_Not_Accumulate_Padding: 5 maximize/restore cycles, asserts Padding and Viewport are stable
- ScrollBar_Visibility_Toggle_Is_Idempotent: 10 layout passes, asserts Padding stays constant

All 7 existing StaleContentSizeCaptureTests continue to pass.

## How we found it

Instrumented DialogTResult.cs and AdornmentImpl.cs with file-based tracing (Padding.Thickness setter logged every change with full stack traces). The stack traces showed ScrollBar.ShowHide -> ConfigureVerticalScrollBarEvents/ConfigureHorizontalScrollBarEvents -> Padding.Thickness += 1 firing during PaddingView.OnParentFrameChanged on every layout re-entry cycle.

### DIFF
--- a/Terminal.Gui/ViewBase/View.ScrollBars.cs
+++ b/Terminal.Gui/ViewBase/View.ScrollBars.cs
@@ -2,6 +2,12 @@ namespace Terminal.Gui.ViewBase;
 
 public partial class View
 {
+    // Track whether scrollbar padding has been applied to prevent accumulation
+    // during layout re-entry. The VisibleChanged handlers use +1/-1 relative
+    // adjustments, which compound when the event fires multiple times per cycle.
+    private bool _verticalScrollBarPaddingApplied;
+    private bool _horizontalScrollBarPaddingApplied;
+
     private Lazy<ScrollBar> _horizontalScrollBar = null!;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


BDisp (approved):

tig:
Thanks. I fixed this in #4862.

PR #4886, opened 2026-04-05
Fixes #4885. View.GetScheme() falls back gracefully instead of throwing when SchemeName is not found

### LINKED ISSUE #4885: Feature: View.GetScheme() should fall back gracefully instead of throwing when SchemeName is not found

When a `View` has `SchemeName` set to a custom scheme name and the active theme doesn't define that name, `View.GetScheme()` throws a `KeyNotFoundException`. This happens in common real-world situations:

- A developer registers a custom scheme and sets `SchemeName` on their views, then the user switches to a built-in theme that doesn't include the custom scheme.
- An end-user edits their config file and makes a typo in a scheme name.
- An end-user deletes a custom scheme from their config, or upgrades to a version of the library where a scheme was renamed.
- I found this to be an issue because I define several custom schemes for specific views - ListViews and ScrollBars are a prime example - and if I don't use a custom theme, my app crashes.

There is also an asymmetry: a view *without* `SchemeName` already falls back gracefully (`SuperView → Base`), but a view *with* `SchemeName` throws immediately if the name isn't found.

Reproduction:

```csharp
SchemeManager.AddScheme("MyScheme", new Scheme { Normal = new Attribute(Color.Cyan, Color.Black) });
view.SchemeName = "MyScheme";

// Later — switch to a theme that doesn't define "MyScheme":
ThemeManager.Theme = "Dark";
ConfigurationManager.Apply();

view.GetScheme(); // KeyNotFoundException
```

This is also discussed in #4457.

When `SchemeName` is set but the named scheme is not found in the active theme, `View.GetScheme()` should fall back gracefully through the following chain (consistent with how a view without `SchemeName` already behaves):

1. Named scheme — if found in the active theme, use it
2. `SuperView.GetScheme()` — inherit from the parent view (recursive)
3. `"Base"` scheme from the active theme
4. Hard-coded `"Base"` — always available regardless of config state

A warning should be written to the diagnostic log when a fallback occurs, so developers can detect misconfigured `SchemeName` values without a crash.

New API addition to `SchemeManager`:

```csharp
// Non-throwing counterpart to GetScheme(string).
// Returns false if the scheme is missing or CM is not in a resolvable state.
public static bool TryGetScheme(string schemeName, [NotNullWhen(true)] out Scheme? scheme)
```

`GetScheme(string)` is left unchanged to preserve backward compatibility.

- **Throw with a better message** — still crashes the app; not acceptable for a theme switch.
- **Return `null` from `GetScheme()`** — would be a breaking API change and propagate null-handling throughout the codebase.
- **Validate `SchemeName` at assignment time** — the scheme may not exist yet, or may be removed later; validation at set-time is the wrong layer.

**Additional context**

See discussion #4457 for the original proposal and proof-of-concept.

YourRobotOverlord:
Fixes #4885.

## Summary

When `View.SchemeName` is set to a name that doesn't exist in the active theme, `View.GetScheme()` previously threw a `KeyNotFoundException`. This PR implements a graceful fallback chain so the view degrades silently instead of crashing.

## Fallback Chain

If the named scheme is not found, resolution continues through:

1. Named scheme — if found in the active theme, use it
2. `SuperView.GetScheme()` — inherit from the parent view (recursive)
3. `"Base"` scheme from the active theme
4. Hard-coded `"Base"` — always available regardless of config state

A `Logging.Warning` is emitted when a fallback occurs so developers can diagnose misconfigured `SchemeName` values.

### `SchemeManager` — new `TryGetScheme` method

`csharp
public static bool TryGetScheme(string schemeName, [NotNullWhen(true)] out Scheme? scheme)
`

Fully exception-safe: handles an uninitialised `ConfigurationManager` and a theme with no `Schemes` property, returning `false` in both cases. `GetScheme(string)` is left unchanged (backward compat — it still throws).

### `View.GetScheme()` — use fallback chain

Replaced the direct `SchemeManager.GetScheme(SchemeName)` call (which throws) with `TryGetScheme` + `ResolveFallbackScheme()`.

### UICatalog scenario — *Theme Fallback*

New demo scenario (`Examples/UICatalog/Scenarios/ThemeFallback.cs`) showing:
- A view with `SchemeName = "CustomHighlight"` (exists on Default, blinking yellow/blue)
- A view with `SchemeName = "NonExistentScheme"` (never exists, always falls back)
- A theme selector so you can watch the fallback chain activate live when switching themes

## Tests

7 new tests across two files — all in `UnitTestsParallelizable`:

| File | Tests added |
|---|---|
| `SchemeManagerTests.cs` | `TryGetScheme` happy path, missing scheme, CM-not-initialised |
| `SchemeTests.cs` | Fallback to SuperView, fallback to Base, no regression, deep chain |

All 15,163 existing tests pass; no new warnings.

### DIFF
--- a/Examples/UICatalog/Scenarios/ThemeFallback.cs
+++ b/Examples/UICatalog/Scenarios/ThemeFallback.cs
@@ -4,14 +4,21 @@ namespace UICatalog.Scenarios;
 
 /// <summary>
 ///     Demonstrates the SchemeName fallback chain introduced in v2.
-///
 ///     When a view's <see cref="View.SchemeName"/> is not found in the active theme, the view no longer
-///     throws a <see cref="KeyNotFoundException"/>. Instead it walks the fallback chain:
+///     throws a <see cref="KeyNotFoundException"/>. Instead, it walks the fallback chain:
 ///     <list type="number">
-///         <item><description>Named scheme (if found in current theme)</description></item>
-///         <item><description>SuperView's scheme (recursive)</description></item>
-///         <item><description>"Base" scheme from the current theme</description></item>
-///         <item><description>Hard-coded "Base" scheme (always present)</description></item>
+///         <item>
+///             <description>Named scheme (if found in current theme)</description>
+///         </item>
+///         <item>
+///             <description>SuperView's scheme (recursive)</description>
+///         </item>
+///         <item>
+///             <description>"Base" scheme from the current theme</description>
+///         </item>
+///         <item>
+///             <description>Hard-coded "Base" scheme (always present)</description>
+///         </item>
 ///     </list>
 /// </summary>
 [ScenarioMetadata ("Theme Fallback", "Demonstrates graceful SchemeName fallback when a named scheme is missing from the active theme.")]
@@ -19,8 +26,8 @@ namespace UICatalog.Scenarios;
 [ScenarioCategory ("Configuration")]
 public sealed class ThemeFallback : Scenario
 {
-    private const string CustomSchemeName = "CustomHighlight";
-    private const string MissingSchemeName = "NonExistentScheme";
+    private const string CUSTOM_SCHEME_NAME = "CustomHighlight";
+    private const string MISSING_SCHEME_NAME = "NonExistentScheme";
 
     public override void Main ()
     {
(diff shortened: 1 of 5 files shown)


YourRobotOverlord:
WindowsTerminal_VwyqmNRxk2

PR #4921, opened 2026-04-10
Fixes #4920. Re-enter RunLoop when End is cancelled by IsRunningChanging

### LINKED ISSUE #4920: StopRequested not reset when IsRunningChanging cancels a stop, causing freeze/crash
## Summary

When `OnIsRunningChanging` (or an `IsRunningChanging` event handler) returns `true` to veto a stop, `StopRequested` is never reset and `ApplicationImpl.Run()` returns early - leaving the runnable with `IsRunning = true`, `StopRequested = true`, still on the `SessionStack`, but **no event loop running**. The app then freezes or crashes.

## Steps to Reproduce

Override `OnIsRunningChanging` in a `Dialog` subclass to show a confirmation `MessageBox` when stopping:

```csharp
protected override bool OnIsRunningChanging(bool oldIsRunning, bool newIsRunning)
{
    if (!newIsRunning)
    {
        int? choice = MessageBox.Query(App, \"Confirm\", \"Really close?\", \"Stay\", \"Leave\");
        if (choice == 0) return true; // veto the stop
    }
    return base.OnIsRunningChanging(oldIsRunning, newIsRunning);
}
```

Run the dialog, trigger a stop (Esc / Cancel button), then click **Stay**. The dialog freezes - it is no longer interactive.

## Root Cause

```
RequestStop(runnable)
  → runnable.StopRequested = true          // 1. StopRequested is set

RunLoop exits (StopRequested == true)       // 2. loop condition false

Run() finally block:
  End(token)
    → RaiseIsRunningChanging(true, false)
        → OnIsRunningChanging returns true  // 3. stop vetoed
    → End() returns early                  // 4. IsRunning stays true, StopRequested stays true

Run() returns to caller                    // 5. BUG: no loop running
```

After 5.: `IsRunning=true`, `StopRequested=true`, on `SessionStack`, no message loop → freeze/crash.

## Expected Behaviour

When `End()` is cancelled by an `IsRunningChanging` handler, `Run()` should re-enter `RunLoop` so the runnable stays interactive.

## Actual Behaviour

`Run()` returns to the caller, leaving the runnable in an irrecoverable inconsistent state.

## Workaround

Pre-validate before calling `RequestStop()` (as `Editor.cs` does in `CanCloseFile()`), avoiding the `OnIsRunningChanging` cancellation path entirely. This is not the advertised pattern for the API, however.

## Environment

- Terminal.Gui `develop` @ `d979cbd47`
- Affects any `Runnable` subclass that vetoes stopping via `OnIsRunningChanging` or the `IsRunningChanging` event

YourRobotOverlord:
Fixes #4920.

## Problem

See the linked issue. When `OnIsRunningChanging` (or an `IsRunningChanging` event handler) vetoes a stop, `ApplicationImpl.Run()` returned early leaving the runnable with `IsRunning=true`, `StopRequested=true`, and no event loop active — a freeze or crash.

## Fix

In `ApplicationImpl.Run()`, wrap the `RunLoop / End` block in a `while(true)` loop and only break when `End()` succeeds (i.e. `runnable.IsRunning == false`). When `End` is cancelled, reset `StopRequested=false` and re-enter `RunLoop`:

```csharp
// BEFORE
try
{
    RunLoop(runnable, errorHandler);
}
finally
{
    End(token);
}

// AFTER
while (true)
{
    try
    {
        RunLoop(runnable, errorHandler);
    }
    finally
    {
        End(token);
    }

    if (!runnable.IsRunning)
        break;

    // End was vetoed — reset StopRequested so RunLoop can re-enter correctly
    runnable.StopRequested = false;
}
```

## Why This Is Safe

| Scenario | Before | After |
|---|---|---|
| Normal stop | ✅ works | ✅ same — `IsRunning` becomes false after `End`, loop breaks |
| Exception in `RunLoop` | ✅ `End` runs in `finally`, exception propagates | ✅ same |
| Stop vetoed by `IsRunningChanging` | ❌ `Run()` returns early; runnable dead | ✅ `RunLoop` re-entered; runnable stays interactive |

## Changes

| File | Change |
|---|---|
| `Terminal.Gui/App/ApplicationImpl.Run.cs` | Wrap `try/finally` in `while(true)` loop |
| `Tests/UnitTestsParallelizable/Application/ApplicationImplTests.cs` | Regression test `Run_WhenStopCancelledByIsRunningChanging_RunLoopResumes` |
| `Examples/UICatalog/Scenarios/IsRunningChangingCancelDemo.cs` | New UICatalog scenario demonstrating the `OnIsRunningChanging` veto pattern |


Examples/UICatalog/Scenarios/IsRunningChangingCancelDemo.cs
```diff
@@ -0,0 +1,155 @@
+#nullable enable
+
```
tig: Given the great unit tests you provided, a new scenario is not needed. Please delete it.

tig (changes requested):
Lovely! See my request.

tig (approved):

PR #4925, opened 2026-04-10
Fixes #4924. Setting Shortcut.Key to Key.Empty now removes the key binding

### LINKED ISSUE #4924: Setting Shortcut.Key to Key.Empty does not remove the key binding
## Bug Description

Setting `Shortcut.Key = Key.Empty` after assigning a non-empty key does not remove the previously-registered key binding. The old binding remains active, so the prior key still triggers the shortcut''s action.

### Root Cause

`Shortcut.UpdateKeyBindings` contained an early-return guard:

```csharp
if (!Key.IsValid)
{
    return;
}
```

`Key.IsValid` returns `false` for `Key.Empty` (defined as `this != Empty && NoAlt.NoShift.NoCtrl != Empty`). As a result, when `Key` is set to `Key.Empty`, the method returned immediately without removing the old key binding. The old binding was left registered and continued firing.

### Reproduction

The following unit tests demonstrate the bug — **they fail before the fix and pass after**:

```csharp
// Copilot
[Fact]
public void Setting_Key_To_Empty_Removes_KeyBinding ()
{
    IApplication app = Application.Create ();
    Runnable<bool> runnable = new ();
    app.Begin (runnable);

    var action = 0;

    Shortcut shortcut = new ()
    {
        Key = Key.F10,
        Title = "Test",
        Action = () => action++
    };

    runnable.Add (shortcut);

    app.Keyboard.RaiseKeyDownEvent (Key.F10);
    Assert.Equal (1, action);

    shortcut.Key = Key.Empty;

    app.Keyboard.RaiseKeyDownEvent (Key.F10);
    Assert.Equal (1, action); // Should not have incremented; old binding must be gone
}

// Copilot
[Fact]
public void Setting_Key_To_Empty_Removes_AppLevel_KeyBinding ()
{
    IApplication app = Application.Create ();
    Runnable<bool> runnable = new ();
    app.Begin (runnable);

    var action = 0;

    Shortcut shortcut = new ()
    {
        App = app,
        Key = Key.F10,
        Title = "Test",
        BindKeyToApplication = true,
        Action = () => action++
    };

    runnable.Add (shortcut);

    app.Keyboard.RaiseKeyDownEvent (Key.F10);
    Assert.Equal (1, action);

    shortcut.Key = Key.Empty;

    app.Keyboard.RaiseKeyDownEvent (Key.F10);
    Assert.Equal (1, action); // Should not have incremented; old app binding must be gone
}
```

### Fix

Remove the `if (!Key.IsValid) return;` guard from `UpdateKeyBindings`. Instead, guard only the _add_ path with `if (Key != Key.Empty)`, so the old binding is always removed but `Key.Empty` is never added as a new binding:

```csharp
private void UpdateKeyBindings (Key oldKey)
{
    if (BindKeyToApplication)
    {
        if (oldKey != Key.Empty)
        {
            App?.Keyboard.KeyBindings.Remove (oldKey);
        }

        App?.Keyboard.KeyBindings.Remove (Key);

        if (Key != Key.Empty)
        {
            App?.Keyboard.KeyBindings.AddApp (Key, this, Command.HotKey);
        }
    }
    else
    {
        if (oldKey != Key.Empty)
        {
            HotKeyBindings.Remove (oldKey);
        }

        HotKeyBindings.Remove (Key);

        if (Key != Key.Empty)
        {
            HotKeyBindings.Add (Key, Command.HotKey);
        }
    }
}
```

This is also needed by the `Keys` UICatalog scenario which adds a "Disable Quit Key" shortcut that sets `quitShortcut.Key = Key.Empty` at runtime.

tig:
Fixes #4924

## Problem

`Shortcut.UpdateKeyBindings` had an early-return guard:

```csharp
if (!Key.IsValid)
{
    return;
}
```

`Key.IsValid` is `false` for `Key.Empty`, so setting `shortcut.Key = Key.Empty` was a no-op — the previously-registered binding was never removed. The old key kept firing the action.

## Fix

- Remove the `if (!Key.IsValid) return` guard entirely.
- Guard only the *add* path with `if (Key != Key.Empty)`, so:
  - The old binding is **always** removed.
  - `Key.Empty` is **never** registered as a new binding.

## Changes

| File | Change |
|------|--------|
| `Terminal.Gui/Views/Shortcut.cs` | Remove invalid early-return; add `Key.Empty` guards on add paths |
| `Tests/…/ShortcutTests.KeyDown.cs` | Two new failing-before/passing-after unit tests |
| `Examples/UICatalog/Scenarios/Keys.cs` | Add "Disable Quit Key" shortcut to demo the fix; use `StatusBar` with `Dim.Fill(statusBar)` for list views |
| `Terminal.Gui/Drivers/AnsiDriver/AnsiInput.cs` | Expose `KeyboardFlags` as a public property |
| `Terminal.Gui/Drivers/AnsiDriver/AnsiComponentFactory.cs` | Include kitty keyboard state in driver name |
| `Terminal.Gui/Drivers/DriverImpl.cs` | Make `KittyKeyboardProtocol` public |
| `Terminal.Gui/Drivers/IDriver.cs` | TODO comment for capabilities extension |
| `Tests/…/InitTests.cs` | Update assertion for new driver name format |
| `Tests/…/AllDriverTests.cs` | Update assertion for new driver name format |

## Tests

Two new tests in `ShortcutTests.KeyDown.cs`:
- `Setting_Key_To_Empty_Removes_KeyBinding` — view-scoped binding
- `Setting_Key_To_Empty_Removes_AppLevel_KeyBinding` — app-scoped (`BindKeyToApplication = true`)

Both fail on the unfixed code and pass with the fix.

### DIFF
--- a/Terminal.Gui/Views/Shortcut.cs
+++ b/Terminal.Gui/Views/Shortcut.cs
@@ -805,11 +805,6 @@ public class Shortcut : View, IOrientation, IDesignable
 
     private void UpdateKeyBindings (Key oldKey)
     {
-        //if (!Key.IsValid)
-        //{
-        //    return;
-        //}
-
         if (BindKeyToApplication)
         {
             if (oldKey != Key.Empty)
--- a/Terminal.Gui/Views/Shortcut.cs
+++ b/Terminal.Gui/Views/Shortcut.cs
@@ -805,10 +805,10 @@ public class Shortcut : View, IOrientation, IDesignable
 
     private void UpdateKeyBindings (Key oldKey)
     {
-        if (!Key.IsValid)
-        {
-            return;
-        }
+        //if (!Key.IsValid)
+        //{
+        //    return;
+        //}
 
         if (BindKeyToApplication)
         {
@@ -819,8 +819,11 @@ public class Shortcut : View, IOrientation, IDesignable
 
             App?.Keyboard.KeyBindings.Remove (Key);
 
-            // Use the form of Add that provides target since this is an app-level hotkey
-            App?.Keyboard.KeyBindings.AddApp (Key, this, Command.HotKey);
+            if (Key != Key.Empty)
+            {
+                // Use the form of Add that provides target since this is an app-level hotkey
+                App?.Keyboard.KeyBindings.AddApp (Key, this, Command.HotKey);
+            }
         }
         else
         {
@@ -830,7 +833,11 @@ public class Shortcut : View, IOrientation, IDesignable
             }
 
             HotKeyBindings.Remove (Key);
-            HotKeyBindings.Add (Key, Command.HotKey);
+
+            if (Key != Key.Empty)
+            {
+                HotKeyBindings.Add (Key, Command.HotKey);
+            }
         }
     }
 
--- a/Tests/UnitTestsParallelizable/Views/ShortcutTests.KeyDown.cs
+++ b/Tests/UnitTestsParallelizable/Views/ShortcutTests.KeyDown.cs
@@ -5,6 +5,67 @@ namespace ViewsTests;
 [TestSubject (typeof (Shortcut))]
 public partial class ShortcutTests
 {
+    // Copilot
+    /// <summary>
+    ///     Verifies that setting <see cref="Shortcut.Key"/> to <see cref="Key.Empty"/> removes the key binding,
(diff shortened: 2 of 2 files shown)


BDisp (approved):
Great. Now assigning a command to a key is always guaranteed, as long as the key is not empty.

Issue #2302, opened 2023-01-23
Support ANSI Operating System Commands (OSCs)


tig:
See https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC

```
OSC (Operating system commands)
Note. These codes may ends with ‘ESC\’ (two symbols - ESC and BackSlash) or ‘BELL’ (symbol with code \x07, same as ‘^a’ in *nix). For simplifying, endings in the following table marked as ‘ST’.

Sequence	Description
ESC ] 0..2 ; “txt” ST	Set console window title to txt.
ConEmu specific OSC
Sequence	Description
ESC ] 9 ; 1 ; ms ST	Sleep. ms - number, milliseconds.
ESC ] 9 ; 2 ; ”txt“ ST	Show GUI MessageBox ( txt ) for any purposes.
ESC ] 9 ; 3 ; ”txt“ ST	Change ConEmu Tab to txt. Set empty string to return original Tab text.
ESC ] 9 ; 4 ; st ; pr ST	Set progress state on Windows 7 taskbar and ConEmu title. When st is 0: remove progress. When st is 1: set progress value to pr (number, 0-100). When st is 2: set error state in progress on Windows 7 taskbar, pr is optional. When st is 3: set indeterminate state. When st is 4: set paused state, pr is optional.
ESC ] 9 ; 5 ST	Wait for Enter/Space/Esc. Set environment variable “ConEmuWaitKey” to “ENTER”/”SPACE”/”ESC” on exit.
ESC ] 9 ; 6 ; ”txt“ ST	Execute GuiMacro ( txt ). Set EnvVar “ConEmuMacroResult” on exit.
ESC ] 9 ; 7 ; ”cmd“ ST	Run some process with arguments.
ESC ] 9 ; 8 ; “env” ST	Output value of environment variable.
ESC ] 9 ; 9 ; “cwd” ST	Inform ConEmu about shell current working directory.
ESC ] 9 ; 10 ST	Request xterm keyboard and output emulation.
ESC ] 9 ; 10 ; n ST	When n is 0 turn off xterm keyboard and output emulation. When n is 1 turn on xterm keyboard and output emulation. When n is 2 turn off xterm output emulation. When n is 3 turn on xterm output emulation.
ESC ] 9 ; 11; “txt” ST	Just a ‘comment’, skip it.
ESC ] 9 ; 12 ST	Let ConEmu treat current cursor position as prompt start. Useful with PS1.
[...]

PR #4930, opened 2026-04-10
Fixes comment on 2302. Add support for OSC 9;4 progress indicator

YourRobotOverlord:
## Summary
- add opt-in ProgressBar.MirrorToTerminal support for OSC 9;4 terminal progress mirroring
- add reusable OSC 9;4 helpers in EscSeqUtils and unit coverage for determinate, indeterminate, hidden, legacy, disable, and dispose behavior
- update the UICatalog ProgressBar Styles scenario to demonstrate terminal mirroring and hiding the selected progress bar

## Validation
- dotnet build --no-restore
- dotnet test --project Tests\UnitTestsParallelizable --no-build
- dotnet test --project Tests\UnitTests.NonParallelizable --no-build
- dotnet test --project Tests\UnitTestsParallelizable --no-build --filter-class "*ProgressBarTests"
- dotnet test --project Tests\UnitTestsParallelizable --no-build --filter-class "*EscSeqUtilsTests"

## Notes
- Tests\UnitTests.Legacy reported zero discovered tests in this environment, so UnitTests.NonParallelizable was used as the remaining repo test surface.


YourRobotOverlord:
This addresses the specific note from #2302:

@tig
> ESC ] 9 ; 4 ; st ; pr ST works on Windows Terminal - We should add an option to ProgressBar to have it use this!!!

Play with it when you have time and let me know if I should carry on.

WindowsTerminal_H86ijIJgVT

WindowsTerminal_ZzkdrKBSaT

tig:
I love it.

But I think the complexity w/on Progress.MirrorToTerminal should be moved into DriverImpl behind one or more new methods on IDriver. E.g. Build a helper class like KittyKeyboardDetector called ProgressIndicator. Then on IDriver:

```cs
public ProgressIndicator? ProgressIndicator {get;set;}
```

I don' think `OSC 9;4` support is detectable, but if it is, MainLoopCoordinator would instantiate this only if support was detected. If it's not detectiable then, DriverImpl would just instantiate it at Init.

ProgressBar could do:

```cs
App?.Driver.ProgressIndicator?.Clear ();
App?.Driver.ProgressIndicator?.SetValue (42);
App?.Driver.ProgressIndicator?.SeError (99);
App?.Dirver.ProgressIndicator?.SetPaused(3); 
App?.Driver.ProgressIndicator?.SetIndeterminate();
```

Or similar...

thots?

YourRobotOverlord:
> But I think the complexity w/on Progress.MirrorToTerminal should be moved into DriverImpl...

This makes sense, I'll make the changes.

> I don' think OSC 9;4 support is detectable...

I believe you are correct, but I'll do some research to verify this.

YourRobotOverlord:
tl/dr: No standardized detection; just go for it on non-dumb TTY output streams.
<details><summary>Deep Research Details</summary>
<p>

# OSC 9;4 Progress Indicators in Terminal Emulators

## Executive summary

OSC 9;4 (“ConEmu progress”) is a non-standard but increasingly adopted Operating System Command escape sequence for reporting a task’s progress to a terminal emulator so the terminal can surface progress in UI chrome (tab header, window title area, dock/taskbar, etc.). The canonical published specification originates from ConEmu and defines `ESC ] 9 ; 4 ; st ; pr ST`, with `st` selecting a state and `pr` (usually 0–100) representing percentage.

As of April 11, 2026, the strongest evidence of *explicit* (user-visible) OSC 9;4 support exists for Windows Terminal, iTerm2, Ghostty, and VTE-based terminals such as GNOME Terminal (via VTE). Windows Terminal’s documentation requires v1.6+ and describes progress display both in the tab header and Windows taskbar. iTerm2 documents the protocol and also provides a capability-reporting mechanism that can, in principle, be used for runtime detection. Ghostty documents and ships progress support (and a configuration toggle to ignore OSC 9;4).

“Detect support at runtime” is only partially solvable in a robust, universal way. There is no cross-terminal, standardized query/response for “do you support OSC 9;4?” (unlike some other terminal extensions that have explicit probes). In practice, you must combine: terminal- or ecosystem-specific feature reporting (best case: iTerm2), environment-variable heuristics, safeguards (TTY-only, output stream choice, strict termination), conservative allowlists/denylists, and user overrides.

Key interop risks are real and have already impacted mainstream tooling: OSC 9 historically also meant “desktop notification” in iTerm2, and terminals that implement OSC 9 notifications may misinterpret progress sequences as notification requests in some versions/configurations; pytest, for example, disabled its OSC 9;4-based progress plugin by default in 9.0.2 (except on Windows) due to compatibility issues, and added special casing because iTerm2 generated desktop notifications instead of the intended progress UI.

## OSC 9;4 semantics and the main compatibility traps

The ConEmu-published OSC 9;4 format is:

- `ESC ] 9 ; 4 ; st ; pr ST`
- `ST` is a string terminator; ConEmu’s documentation explicitly notes termination can be either `ESC \` or BEL (`\x07`) in its tables.
- State meanings in the ConEmu spec (most widely copied) are:
  - `st=0`: remove/hide progress
  - `st=1`: set progress value `pr` (0–100)
  - `st=2`: error state (with optional `pr`)
  - `st=3`: indeterminate
  - `st=4`: paused (with optional `pr`)

Windows Terminal documents essentially the same state model but describes UI outcomes specifically for Windows (tab header ring + taskbar progress). iTerm2 also documents OSC 9;4, but with a slightly different naming for state `4` (“warning”), highlighting that state semantics are not perfectly consistent across implementations.

Two interop problems dominate in real deployments:

1. **OSC 9 namespace collision with notifications**  
   iTerm2 historically uses OSC 9 for “Post a notification”. ConEmu repurposed OSC 9 subcommands (including 9;4) for extensions. This creates ambiguous parsing and, for some terminals, visible side effects (spurious notifications) when an app starts emitting OSC 9;4. pytest’s changelog records that its terminal progress plugin was “automatically disabled when iTerm2 is detected” because iTerm2 generated desktop notifications. kitty’s changelog documents a defensive response: it discards OSC 9 notifications that start with `4;` specifically because software is using it for “progress reporting,” implying kitty is prioritizing avoiding misbehavior rather than rendering progress UI. Ghostty’s documentation explicitly calls out the OSC 9 collision and describes how it distinguishes ConEmu sub-IDs versus notification sequences.

2. **Terminator differences (BEL vs ST) and strict parsing**  
   In theory, many OSC sequences accept either BEL or ST terminators (xterm tradition; also documented in multiple places). However, field reports indicate that at least some VTE-based setups behave differently: a Stack Overflow answer by Egmont Koblinger notes VTE “insists” on the official ST terminator (`ESC \`) rather than BEL for this sequence, and a Unix & Linux answer demonstrates a working test loop using `ESC \` on GNOME Terminal with VTE 0.80.x.  
   Practical implication: **for maximum compatibility, prefer `ESC \` as your terminator** when emitting OSC 9;4, even if Windows-centric examples use BEL.

## Support landscape and compatibility matrix

The table below reflects evidence located in primary documentation, upstream code/diffs, changelogs, or issue trackers retrieved up to April 11, 2026. Versions are often unspecified unless the source provides them; treat “unknown” as “no solid primary evidence located,” not necessarily “unsupported.”

Terminal emulator | Family / platform | OSC 9;4 support status | What “support” means here | Key caveats
-- | -- | -- | -- | --
Windows Terminal | Windows | Explicit | Renders progress ring in tab header and shows progress in Windows taskbar | Requires Windows Terminal v1.6+; examples primarily show BEL terminator
ConEmu | Windows | Explicit (canonical spec) | Implements and documents the sequence; integrates with Windows 7 taskbar and ConEmu title | Canonical state definitions; uses ST concept (BEL or ESC \)
iTerm2 | macOS | Explicit | Documents OSC 9;4 and shows a UI progress bar; also provides feature reporting | OSC 9 also used for notifications; older versions may misinterpret 9;4 as notifications (real-world reports)
Ghostty | Windows/macOS/Linux | Explicit | Shows graphical progress bars; provides config toggle to ignore OSC 9;4 | Progress style can be disabled (progress-style); some tools observe auto-hide/timeout behavior if not cleared
VTE (library) | Linux/Unix (GTK terminal widgets) | Explicit (parser + term properties) | VTE exposes progress term properties and documents that before v0.82 they were set via OSC 9;4 (ConEmu progress) | Front-end UI varies; evidence suggests at least some VTE front-ends require ESC \ terminator
GNOME Terminal | VTE-based | Explicit (observed working) | Demonstrated to show a progress indicator in tab UI on GNOME Terminal 3.56.2 with VTE 0.80.3 | May require multiple tabs to notice; relies on VTE behavior/terminator
Ptyxis | GNOME/VTE ecosystem | Explicit (design + intent) | GNOME developer blog describes adopting the ConEmu-originated progress escape for UI surfacing | Detailed UI behavior may evolve; still VTE ecosystem
Konsole | KDE | Partial | Implements “show/hide” determinate progress in tab bar; other states incomplete | MR explicitly: error/indeterminate/paused not implemented; code shows state 3 currently hides; states 2/4 TODO
WezTerm | Cross-platform | Partial | Parses OSC 9;4, tracks state, exposes via Lua; triggers tab bar re-render hooks | Docs: “progress information is not directly used by wezterm”; state 4 not supported; “nightly builds only” for API
Kitty | Cross-platform | No (defensive ignore) | Does not document rendering progress; instead discards OSC 9 notifications beginning with 4; to avoid misinterpreting progress as notifications | This reduces spurious notifications but does not imply a progress UI; third-party claims of support are inconsistent
Alacritty | Cross-platform | No | Issue labeled “wontfix” for implementing OSC 9;4 support | If emitted, likely ignored as unknown OSC; still, explicit support not planned
mintty | Windows (Cygwin/MSYS2) | No (different protocol exists) | Documents a different progress mechanism using %q sequences (CSI-style), not OSC 9;4 | If you need progress in mintty, OSC 9;4 is not the documented way
PuTTY | Windows (SSH client + emulator) | Unknown / no direct evidence | PuTTY supports some xterm-like title OSC, and explicitly rejects dangerous OSC features for security | No primary evidence of OSC 9;4 support located in this research set
Terminal.app (macOS) | macOS | Unknown / no direct evidence | No primary Apple documentation of OSC 9;4 progress support located here | Treat as unsupported unless you test and confirm
xterm | X11 | Unknown / no direct evidence | No primary xterm documentation for OSC 9;4 progress located here | Likely ignored if unknown; do not assume
rxvt / urxvt | X11 | Unknown / no direct evidence | No primary urxvt/rxvt source for OSC 9;4 progress located here | Do not assume
st | X11 (suckless) | Unknown / no direct evidence | st upstream log shows OSC colour work, but nothing here about OSC 9;4 | Many features require patches; treat as unknown
Rio | Cross-platform | Explicit (claimed by project changelog) | Project changelog lists “OSC 9;4 Progress Bar Support” | Exact UI and state coverage depends on version/config; changelog is primary for the project
tmux (multiplexer) | Unix | Not a terminal; can block OSC | By default, tmux may filter OSC sequences; apps may need DCS passthrough wrapping and user config | Requires allow-passthrough in tmux 3.2+ (per ecosystem reports); implementation varies
Zellij (multiplexer) | Cross-platform | Blocks / doesn’t handle (reported) | Reported that OSC 9;4 doesn’t work inside Zellij even when underlying terminal supports it | If you want OSC 9;4 inside multiplexers, you need multiplexer-specific handling

## Runtime detection and negotiation

### What is realistically detectable

There is **no universal terminal capability query** for OSC 9;4 comparable to how terminfo provides low-level cursor/color capabilities. Your options are:

- **Terminal-provided feature reporting (best case)**: iTerm2 defines a feature reporting mechanism that includes a PROGRESS capability and can be published via the `TERM_FEATURES` environment variable or queried via `OSC 1337 ; Capabilities`.  
- **Environment-variable heuristics (common case)**: infer the terminal family from well-known environment variables (e.g., `WT_SESSION` for Windows Terminal, `VTE_VERSION` for VTE-based terminals), while accounting for inheritance/false positives.  
- **Conservative emit-and-ignore (fallback case)**: emit OSC 9;4 only in safe situations (TTY, correct terminator, rate-limited), assuming unsupported terminals ignore unknown OSC. This is broadly true but not risk-free because OSC 9 collisions can cause visible side effects in certain terminals/versions (notifications).

### Recommended strategy stack (prioritized) and trade-offs

**Use multiple layers and stop early when you have high confidence.** A practical ordering that minimizes user harm:

1. **Hard safety gates (always)**
   - Only emit when the output stream is a TTY (`isatty`). pytest explicitly avoids emitting escape codes when `TERM=dumb`, and disables when not appropriate.  
   - Prefer emitting to **stderr** (so stdout can remain machine-readable / pipe-friendly), but still check `stderr.isatty()`. This is a standard CLI robustness pattern (not unique to OSC 9;4).  
   - Always clear progress (`st=0`) on normal exit, and best-effort clear on signals, to avoid “stuck” UI indicators; real bugs exist when apps fail to clear.

2. **Explicit feature negotiation when available**
   - **If iTerm2 feature reporting is available**, treat that as authoritative. It is explicitly designed to expose PROGRESS support.  
   Trade-off: requires parsing a terminal-specific protocol and (if querying) reading a response from the controlling terminal safely.

3. **High-confidence allowlist heuristics**
   - Windows Terminal: presence of `WT_SESSION` is a common heuristic, and Windows Terminal issues discuss populating/propagating this variable.  
     Trade-off: `WT_SESSION` can be inherited by child processes (e.g., launching VS Code from Windows Terminal) and thus produce false positives.  
   - VTE-based: presence of `VTE_VERSION` is widely used in the ecosystem to detect VTE terminals; community guidance for OSC 9;4 specifically suggests a threshold around 0.79.x (numeric 7900).  
     Trade-off: `VTE_VERSION` is not guaranteed to be exported in all distributions/shell init configurations.  
   - Konsole: Konsole has upstream code implementing progress parsing (partial).  
     Trade-off: environment detection for Konsole is less standardized; also implementation is partial.  
   - Ghostty: can be treated as supporting OSC 9;4 by design.  
     Trade-off: you still cannot universally detect Ghostty without heuristics; prefer user override if uncertain.

4. **Denylist known-bad combinations**
   - iTerm2 versions that treat OSC 9;4 as notifications: pytest explicitly special-cases iTerm2 detection and disables its progress plugin there.  
   - kitty: official changelog indicates it discards OSC 9 notifications starting with `4;`, which avoids spurious notifications but also implies “no progress UI”; you may still emit safely but don’t expect visible results.  
   - Alacritty: explicitly not implementing support (wontfix).

5. **User override controls (strongly recommended)**
   - Provide `MYAPP_OSC_PROGRESS=1|0` (or similar) to let users force enable/disable when heuristics are wrong, terminals are embedded, or multiplexers intervene. This mirrors what many tools have done after encountering terminal variability (e.g., pytest’s plugin enable/disable flags).

### Detection flow diagram

```mermaid
flowchart TD
  A[Start] --> B{Is output stream a TTY?}
  B -- No --> Z[Do not emit OSC 9;4; use text fallback]
  B -- Yes --> C{TERM == dumb?}
  C -- Yes --> Z
  C -- No --> D{iTerm2 feature reporting available?}
  D -- Yes --> E[Check TERM_FEATURES / query OSC 1337 Capabilities]
  E --> F{PROGRESS supported?}
  F -- Yes --> P[Enable OSC 9;4 emission]
  F -- No --> Z
  D -- No --> G{High-confidence terminal allowlist?}
  G -- Yes --> P
  G -- No --> H{Known-bad terminal/version?}
  H -- Yes --> Z
  H -- No --> I[Default off OR emit in “safe no-op” mode with user opt-in]
```

## Emission code snippets, probes, and reproducible tests

### Minimal OSC 9;4 emitters (prefer ST terminator)

Bash (ST terminator `ESC \`):

```bash
# Determinate 50%
printf '\033]9;4;1;50\033\\'

# Indeterminate
printf '\033]9;4;3\033\\'

# Error state (no percent)
printf '\033]9;4;2\033\\'

# Clear
printf '\033]9;4;0\033\\'
```

This aligns with the canonical ConEmu shape (`... ST`) and avoids BEL-vs-ST pitfalls observed in VTE contexts.

Python (write to stderr, TTY-gated):

```python
import os, sys

ESC = "\x1b"
ST  = "\x1b\\"

def osc_94(st: int, pr: int | None = None) -> str:
    if pr is None:
        return f"{ESC}]9;4;{st}{ST}"
    return f"{ESC}]9;4;{st};{pr}{ST}"

def set_progress(st: int, pr: int | None = None) -> None:
    if not sys.stderr.isatty():
        return
    sys.stderr.write(osc_94(st, pr))
    sys.stderr.flush()

# Demo
set_progress(3)
```

State semantics (including which states allow optional `pr`) are documented by ConEmu.

C# (Console app) (write to stderr, ST terminator):

```csharp
using System;

static class OscProgress
{
    private const string ESC = "\x1b";
    private const string ST  = "\x1b\\";
    
    public static void Set(int state, int? percent = null)
    {
        if (Console.IsErrorRedirected) return;

        if (percent is null)
            Console.Error.Write($"{ESC}]9;4;{state}{ST}");
        else
            Console.Error.Write($"{ESC}]9;4;{state};{percent}{ST}");
    }

    public static void Clear() => Set(0);
}
```

Windows Terminal’s examples demonstrate the same OSC 9;4 structure (they use BEL in examples), and ConEmu’s documentation defines the state model.

### Reproducible probe loops

Bash “fill 0→100 then clear” (works in terminals that actually render progress UI):

```bash
for i in $(seq 0 5 100); do
  printf '\033]9;4;1;%s\033\\' "$i"
  printf '%s ' "$i"
  sleep 0.2
done
printf '\033]9;4;0\033\\\n'
```

A Unix & Linux answer reports this style of loop working in GNOME Terminal with VTE 0.80.x and notes it may be visible in the tab UI (often clearer with multiple tabs).

Python “state sweep” to see how a terminal maps states:

```python
import sys, time

ESC = "\x1b"
ST  = "\x1b\\"

def emit(s: str):
    if sys.stderr.isatty():
        sys.stderr.write(s)
        sys.stderr.flush()

tests = [
    f"{ESC}]9;4;1;25{ST}",
    f"{ESC}]9;4;1;75{ST}",
    f"{ESC}]9;4;3{ST}",
    f"{ESC}]9;4;2{ST}",
    f"{ESC}]9;4;0{ST}",
]

for t in tests:
    emit(t)
    time.sleep(1.5)
```

This helps you empirically discover state mapping inconsistencies (e.g., “paused” vs “warning”) across terminals; those inconsistencies are documented between ConEmu and iTerm2.

### Runtime probe that yields an actual response (iTerm2 feature reporting)

If you choose to query iTerm2’s `OSC 1337;Capabilities`, the terminal replies with a capabilities report string; this is an actual request/response protocol (unlike OSC 9;4 itself).

A safe implementation requires:
- writing the query to the controlling TTY,
- reading the response with a timeout,
- keeping the terminal in a sane input mode.

Because the exact encoding of the `FeatureString` must follow iTerm2’s spec, treat this as terminal-specific plumbing you should feature-flag and heavily test.

## Fallback UX patterns and risk analysis

### Fallback UX patterns when OSC 9;4 is unsupported or blocked

A robust CLI typically provides a layered approach:

- **In-band textual progress (always available)**: single-line progress bar/spinner updated with carriage returns, suitable for any terminal and for log capture (where appropriate). This avoids reliance on terminal UI chrome.
- **“Quiet” mode for non-TTY**: when output is redirected, emit periodic textual milestones instead of ANSI control sequences (pytest explicitly avoids escape emission in `TERM=dumb` and, in general, when unsuitable).
- **User-controlled toggles**: `--progress=auto|always|never`, mirroring how pytest exposes the progress plugin enable/disable and how real-world tools have had to respond to compatibility surprises.
- **Always clear-on-exit**: send `st=0` at completion, and register signal/atexit handlers to clear on interrupts; bugs where tools fail to clear are observable in the wild.

### Multiplexer considerations (tmux, Zellij)

Even if the underlying emulator supports OSC 9;4, multiplexers can block or transform OSC sequences.

- tmux is widely reported to intercept OSC sequences; ecosystem requests recommend wrapping OSC in tmux’s DCS passthrough format and enabling `allow-passthrough` (tmux 3.2+).  
- Zellij users have reported OSC 9;4 not working inside sessions even when it works outside.  

Practical UX guidance: if you detect multiplexer environments (e.g., `$TMUX`), either (a) disable OSC 9;4 by default unless you can safely passthrough, or (b) provide an explicit “enable multiplexer passthrough” mode and document the required user configuration.

### Security and side effects

Emitting OSC sequences is part of the broader class of “terminal escape sequences,” which have a history of security-relevant issues (e.g., denial-of-service vectors and terminal emulation parsing vulnerabilities). A security research write-up from CyberArk describes finding multiple vulnerabilities across terminals by abusing terminal escape characters and related behavior.

Key risks specific to OSC 9;4 usage patterns:

- **Untrusted output injection**: If your app prints content from untrusted sources (logs, remote text, file contents) to the user’s terminal, those contents can include escape sequences; a malicious actor could cause misleading UI, spammy notifications, or worse in some terminal environments.  
- **Side effects from namespace collisions**: Because OSC 9 is also used for notifications in iTerm2 (and historically in other stacks), OSC 9;4 can cause unintended desktop-notification spam in some configurations/versions, motivating tool-level mitigations (pytest) and terminal-level mitigations (kitty discarding `4;` prefixed notifications).  
- **False positives in detection**: Environment variables such as `WT_SESSION` can be inherited into other hosts (e.g., VS Code launched from Windows Terminal), leading to erroneous “Windows Terminal detected” decisions.  
- **State ambiguity and portability**: Community tooling notes that state `4` is ambiguous across terminals (“paused” vs “warning”), so mapping semantic meaning onto state `4` is risky if you need consistent UX.  

Mitigations that work in practice:

- Emit OSC 9;4 only when you have a TTY and are reasonably confident in support (or the user has opted in).  
- Default to **ST (`ESC \`) terminator** to reduce strict-parsing failures.  
- Provide a “strip OSC 9;4” sanitizer for log capture and testing harnesses; multiple libraries now implement sanitization utilities, reflecting real operational need.  

## Timeline and adoption notes

This timeline is limited to date/versions explicitly supported by sources retrieved in this research pass (others remain unspecified).

- ConEmu published the OSC 9;4 progress protocol and documents integration with Windows 7 taskbar progress and ConEmu title UI (indicating Windows 7–era origins).  
- Windows Terminal supports OSC 9;4 as of v1.6+ (documentation dated September 28, 2023).  
- VTE’s introspection documentation shows `vte.progress.*` term properties “New in version 0.80” and notes that before 0.82 those term properties could be set by OSC 9;4 (indicating OSC 9;4 support landed by the 0.80 timeframe).  
- Konsole’s upstream merge request to “Support ConEmu progress OSC” was opened December 20, 2024; the diff shows determinate progress wired up, while other states are TODO.  
- WezTerm’s changelog in its “Continuous/Nightly” section notes “Preliminary support for ConEmu style progress escape sequences” (issue #6581), and the Lua API docs for `pane:get_progress()` are marked “Nightly Builds Only,” with state `4` not supported.  
- kitty’s 0.38.0 changelog entry (December 15, 2024) documents discarding OSC 9 notifications that start with `4;` to avoid collisions with progress-reporting software, a compatibility-focused stance rather than explicit progress rendering.  
- Ghostty 1.2.0 release notes (September 15, 2025) describe graphical progress bars and future plans to expand where progress is shown; Ghostty 1.3.1 release notes (March 13, 2026) add a configuration option to disable OSC 9;4 progress bars (`progress-style`).  
- pytest 9.0.0 (November 5, 2025) added OSC 9;4-based terminal progress support, then disabled it by default in 9.0.2 (except on Windows) due to terminal compatibility issues; it also records special casing around iTerm2 (notifications vs progress).
</p>
</details>

YourRobotOverlord:
WindowsTerminal_nxWYSDSMw6

kovidgoyal:
Your information about kitty is out of date. kitty uses OSC 9;4 for progress since version 0.40.1 which is over a year old.

YourRobotOverlord:
> Your information about kitty is out of date. kitty uses OSC 9;4 for progress since version 0.40.1 which is over a year old.

🤦‍♂️Good job, ChatGippity Deep Research...

Good to know, thanks!

YourRobotOverlord:
I put this back to draft, I'm going to add it as an option on `SpinnerView` as well. Seeing the spinner used in the Ai example made me think of this.
It'll just set indeterminate mode on/off with the spinner if enabled.

tig (approved):
I did some code cleanup (which I love doing on PRs because it forces me to look carefully at code). 

I also made the AI example use the new indicator. 

Really great addition to the library. Thank you!

YourRobotOverlord:
@tig I'm really second guessing the property name `UseProgressIdicator`, especially when considering the context of the `ProgressBar`. It seems like it would enable/disable the indicator on the bar itself and may be confusing af.
It should be clear that it is the terminal progress indicator, the way your `TerminalTitle` is clear.
Thoughts?

tig:
I think your instincts are right. Wanna submit a pr?

YourRobotOverlord:
> Wanna submit a pr?

You bet! `SyncTerminalProgressIndicator` seems clearer - albeit a bit long - but open to suggestions.

tig:
SyncWithTerminal

Progress is obvious.

PR #4957, opened 2026-04-13
Fixes #4956 - CPR response misinterpreted as F3

### LINKED ISSUE #4956: SaveDialog is showing when open the Editor scenario
@tig the last PR #4953 merged is causing the save dialog showing when I open the Editor scenario.

image

_Originally posted by @BDisp in https://github.com/gui-cs/Terminal.Gui/issues/4723#issuecomment-4238559310_

tig:
## Fixes

- Fixes #4956 

## Summary
- Adds a failing regression test for #4956.
- After #4953 switched the inline-mode cursor query from DECXCPR (`CSI ?6n`) to standard CPR (`CSI 6n`), the response `ESC[1;1R` (cursor at row 1) is consumed by `CsiCursorPattern` as `Key.F3` with modifier — because `ESC[1;<n>R` is xterm's legitimate encoding for modified F3 (e.g. Shift+F3 = `ESC[1;2R`).
- In `AnsiResponseParserBase.HandleHeldContent`, keyboard patterns are matched **before** registered expected responses, so a pending CPR request loses the race. The CPR callback never fires and an F3 key event is dispatched instead — which UICatalog's Editor and Notepad scenarios bind to Save, producing a spurious Save on startup.
- The old DECXCPR response `ESC[?1;1R` didn't hit this because the leading `?` prevented the keyboard regex from matching.

This PR **only adds the failing test** so the regression is captured before the fix lands.

## Test plan
- [ ] `dotnet test --project Tests/UnitTestsParallelizable` — new test `ExpectedCprResponse_IsNotMisinterpretedAsF3` fails on `develop`.
- [ ] Once the fix lands, the test passes.

Refs #4956. Regression from #4953.

### DIFF
--- a/Terminal.Gui/Drivers/AnsiHandling/AnsiResponseParserBase.cs
+++ b/Terminal.Gui/Drivers/AnsiHandling/AnsiResponseParserBase.cs
@@ -330,24 +330,13 @@ internal abstract class AnsiResponseParserBase (IHeld heldContent, ITimeProvider
                 return false;
             }
 
-            if (HandleKeyboard)
-            {
-                AnsiKeyboardParserPattern? pattern = _keyboardParser.IsKeyboard (cur);
-
-                if (pattern != null)
-                {
-                    // See https://github.com/gui-cs/Terminal.Gui/issues/4587#issuecomment-3770132337 for why
-                    // we call ResetState first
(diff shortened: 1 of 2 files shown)


BDisp (approved):
Good find!

PR #4977, opened 2026-04-16
Fixes #4976. Preserve repeated ANSI printable keys

### LINKED ISSUE #4976: Regression: repeated ANSI printable keys drop every second identical character
<details><summary>Copilot Session</summary><p>61bf61e5-930c-4a8d-bf2c-723e6f1dc0ae</p></details>

Regression after #4949: repeated printable keys on ANSI input drop every second identical character.

## Symptoms

In Windows Terminal / ANSI input:

- `123412341234` registers all keys
- `111222333444` registers as `11223344`

This appears to be a regression in the printable fallback suppression added to avoid duplicate kitty + legacy printable input.

## Minimum repro

Add this test to `Tests\UnitTestsParallelizable\Drivers\AnsiHandling\KittyKeyboardPipelineTests.cs`:

```csharp
// Copilot
[Fact]
public void Pipeline_RepeatedLegacyPrintableInput_DoesNotDropRepeatedCharacters ()
{
    (List<Key> down, List<Key> up) = InjectRawSequence ("111222333444");

    Assert.Equal (12, down.Count);
    Assert.Equal ("111222333444", string.Concat (down.Select (key => key.GetPrintableText ())));
    Assert.Empty (up);
}
```

Run:

```bash
dotnet test --project Tests/UnitTestsParallelizable --filter-class "*KittyKeyboardPipelineTests"
```

## Actual result before fix

The new test fails:

```text
Expected: 12
Actual:   8
```

## Suspected cause

`Terminal.Gui\Drivers\AnsiDriver\AnsiInputProcessor.ShouldSuppressFallbackKeyDown` was changed in #4949 to arm `_pendingPrintableSuppression` even when no parsed ANSI/kitty printable event is pending.

That makes pure fallback printable input behave like this:

1. first `1` passes and arms suppression
2. second `1` is suppressed
3. third `1` passes and re-arms suppression

Alternating keys do not reproduce because the suppression string does not match the next key.

## Expected behavior

Fallback printable input should only be suppressed when it is the duplicate legacy-printable half of a mixed kitty + legacy event, not during ordinary repeated printable input.

YourRobotOverlord:
<details><summary>Copilot Session</summary><p>61bf61e5-930c-4a8d-bf2c-723e6f1dc0ae</p></details>

## Summary

- preserve repeated ANSI legacy printable input on non-kitty terminals
- restore legacy glyph deduplication for kitty-enabled terminals where some PT glyphs still arrive as plain printable input
- propagate kitty capability state into `AnsiInputProcessor`
- replace flawed PT glyph coverage that assumed `«»çÇºª` always arrive as CSI-u

## Side-effects

- **Non-kitty terminals:** repeated printable keys like `111222` are no longer dropped
- **Kitty-enabled terminals:** fallback self-arming is restored only in kitty mode, so legacy-only duplicated glyphs like `«»çÇºª` can be deduped again
- **Watch point:** if a kitty-enabled terminal emits some other printable key only as legacy input instead of CSI-u, every second identical fallback character for that key may still be suppressed
- **Startup edge:** kitty mode is applied after startup detection, so input received before detection completes still behaves like non-kitty mode

## Testing

- `dotnet test --project Tests\UnitTestsParallelizable --filter-class "*KittyKeyboardPipelineTests"`
- `dotnet build --no-restore`
- `dotnet test --project Tests\UnitTestsParallelizable --no-build`
- `dotnet test --project Tests\UnitTests.NonParallelizable --no-build`

### DIFF
--- a/Terminal.Gui/App/MainLoop/MainLoopCoordinator.cs
+++ b/Terminal.Gui/App/MainLoop/MainLoopCoordinator.cs
@@ -223,16 +223,27 @@ internal class MainLoopCoordinator<TInputRecord> : IMainLoopCoordinator where TI
 
             kittyKeyboardDetector.Detect (result =>
                                           {
-                                              if (!result.IsSupported)
-                                              {
-                                                  Trace.Lifecycle (app?.MainThreadId?.ToString (), "KittyKeyboard", "Kitty keyboard mode not enabled");
+                                               if (!result.IsSupported)
+                                               {
+                                                   if (_inputProcessor is AnsiInputProcessor unsupportedAnsiInputProcessor)
+                                                   {
+                                                       unsupportedAnsiInputProcessor.SetKittyKeyboardEnabled (false);
+                                                   }
 
-                                                  return;
-                                              }
+                                                   Trace.Lifecycle (app?.MainThreadId?.ToString (), "KittyKeyboard", "Kitty keyboard mode not enabled");
 
-                                              // Kitty is supported. Store the capabilities and set the flags we care about.
-                                              _driver?.SetKittyKeyboardCapabilities (result);
-                                              kittyKeyboardDetector.Enable (EscSeqUtils.KittyKeyboardRequestedFlags);
+                                                   return;
+                                               }
+
+                                               // Kitty is supported. Store the capabilities and set the flags we care about.
+                                               _driver?.SetKittyKeyboardCapabilities (result);
+
+                                               if (_inputProcessor is AnsiInputProcessor supportedAnsiInputProcessor)
+                                               {
+                                                   supportedAnsiInputProcessor.SetKittyKeyboardEnabled (true);
+                                               }
+
+                                               kittyKeyboardDetector.Enable (EscSeqUtils.KittyKeyboardRequestedFlags);
 
                                               Trace.Lifecycle (app?.MainThreadId?.ToString (),
                                                                "KittyKeyboard",
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiInputProcessor.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiInputProcessor.cs
@@ -83,8 +83,6 @@ public class AnsiInputProcessor : InputProcessorImpl<char>
     {
         if (string.IsNullOrEmpty (_pendingPrintableSuppression))
         {
-            _pendingPrintableSuppression = key.GetPrintableText ();
-
             return false;
         }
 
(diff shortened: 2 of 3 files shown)


BDisp:
It's back to the previous problem, as the video demonstrates. The problem is that this glyph deduplication only happens with kitty enabled, and they appear in isolation, not as CSI-u. Therefore, the original test is flawed because it assumes that the glyphs «»çÇºª appear in the input as, for example, $"\x1b[«u", but in reality, they only appear as «. Therefore, there needs to be a mechanism that passes, perhaps as a parameter, which terminal type is in use and acts accordingly.

WindowsTerminal_rtSITLAyJ1

YourRobotOverlord:
<details><summary>Copilot Session</summary><p>61bf61e5-930c-4a8d-bf2c-723e6f1dc0ae</p></details>

Updated PR for your scenario.

What changed:
- `AnsiInputProcessor` now restores fallback self-arming only when kitty mode is enabled
- `MainLoopCoordinator` propagates kitty capability state into the ANSI input processor
- PT glyph tests no longer assume `«»çÇºª` arrive as CSI-u; coverage now includes kitty-enabled terminals where those glyphs still arrive as plain legacy input

Side-effects:
- **Non-kitty terminals:** repeated printable keys like `111222` are no longer dropped
- **Kitty-enabled terminals:** legacy-only duplicated glyphs like `«»çÇºª` can be deduped again
- **Watch point:** if a kitty-enabled terminal emits some other printable key only as legacy input instead of CSI-u, every second identical fallback character for that key may still be suppressed
- **Startup edge:** input received before kitty detection completes still behaves like non-kitty mode

BDisp:
This is working well in my machine for legacy or kitty support. Well done. Thanks.

Is this ready for review?

BDisp (approved):
Well done. Thanks for this.

tig:
@YourRobotOverlord Regarding merging approved PRs.

My rule is that I am always the final approver for PRs and no merge should happen if I have not explicitly approved it. Other contribs like you and @BDisp can approve PRs, but you should NOT view that as permission to merge.

Cool?

tig (approved):

YourRobotOverlord:
> ... you should NOT view that as permission to merge.
> 
> Cool?

100% understood, and I am aware of this requirement; hopefully I did not do this somewhere by mistake?

tig:
> > ... you should NOT view that as permission to merge.
> > Cool?
> 
> 100% understood, and I am aware of this requirement; hopefully I did not do this somewhere by mistake?

Nope! I just wanted to make sure.

BDisp:
Now that @tig has approved it, you can merge it as you wish.

PR #4978, opened 2026-04-16
Fixes #4975. Clarify MessageBox default-button docs and signatures

### LINKED ISSUE #4975: MessageBox docs and signatures are misleading about default button behavior
[MessageBox.Query(IApplication app, string title, string message, int defaultButton, params string[] buttons) API documents defaultButton as the 0-based index of the default button.

However, the current implementation does not apply defaultButton when constructing the dialog buttons. Buttons are created from text only, and later the code attempts to focus the first button with IsDefault == true, but no button is marked default from the provided index.

As a result, changing defaultButton does not affect which button is default/focused.

Expected:
The button at defaultButton should be marked as default and receive the default focus/accept behavior.

Actual:
defaultButton is ignored, so dialog focus/default behavior falls back to internal layout behavior.

YourRobotOverlord:
<details><summary>Copilot Session</summary><p>894256a0-c83e-487e-9244-a975e8cfca7b</p></details>
Fixes #4975.

## Root cause
`MessageBox` intentionally uses the last added button as the default button. That behavior comes from the dialog button flow: `MessageBox` creates buttons from text, assigns them through `dialog.Buttons = ...`, and `Dialog.AddButton` makes the last added button the dialog's `DefaultAcceptView` and marks it `IsDefault`.

The misleading part was the API surface and docs. Some `MessageBox.Query` / `MessageBox.ErrorQuery` signatures and XML docs still implied callers could choose a default button with a `defaultButton` parameter, even though intended runtime behavior was to keep the last button as default.

## Fix
- remove misleading `defaultButton` overloads from `MessageBox.Query` / `MessageBox.ErrorQuery`
- update `MessageBox` docs to state that the last button is the default button
- update affected examples and tests to match the intended API and behavior
- keep regression coverage focused on the last-button-default behavior

## Validation
- `dotnet build --no-restore`
- `dotnet test --project Tests\UnitTestsParallelizable --no-build --filter-class "*MessageBoxTests"`
- `dotnet test --project Tests\UnitTestsParallelizable --no-build`
- `dotnet test --project Tests\UnitTests.NonParallelizable --no-build`

### DIFF
--- a/Examples/UICatalog/Scenarios/ChineseUI.cs
+++ b/Examples/UICatalog/Scenarios/ChineseUI.cs
@@ -35,16 +35,15 @@ public class ChineseUI : Scenario
 
         btn.Accepting += (s, _) =>
                       {
-                          int? result = MessageBox.Query (
-                                                          (s as View)?.App!,
-                                                          "Confirm",
-                                                         "Are you sure you want to quit ui?",
-                                                         0,
-                                                         "Yes",
-                                                         "No"
-                                                        );
+                           int? result = MessageBox.Query (
+                                                           (s as View)?.App!,
+                                                           "Confirm",
+                                                           "Are you sure you want to quit ui?",
+                                                           "No",
+                                                           "Yes"
+                                                         );
 
-                          if (result == 0)
+                          if (result == 1)
                           {
                               win.RequestStop ();
                           }
--- a/Examples/UICatalog/Scenarios/MessageBoxes.cs
+++ b/Examples/UICatalog/Scenarios/MessageBoxes.cs
@@ -94,27 +94,6 @@ public class MessageBoxes : Scenario
             Width = Dim.Width (label),
             Height = 1,
             TextAlignment = Alignment.End,
-            Text = "_Default Button:"
-        };
-        frame.Add (label);
-
-        TextField defaultButtonEdit = new ()
-        {
-            X = Pos.Right (label) + 1,
-            Y = Pos.Top (label),
-            Width = 5,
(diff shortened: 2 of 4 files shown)


tig (commented):
It was my intent when I removed the default button param from Query was that the LAST button added would always be default. And that devs would have no control over this.

The idea being:

- It drives UI consistency for users for MessageBoxes. Users can count on the last button always being defult.
- If a dev really wants a different behavior, they should not be using MessageBox (or Prompt), but roling their own bespoke Dialog.

I'm happy to debate this! But I don't want to just accept reverting back to the old behavior without doing so.

YourRobotOverlord:
> * It drives UI consistency for users for MessageBoxes. Users can count on the last button always being defult.

This makes sense, and I don't disagree at all. I thought it was broken since the docs still describe the old behavior.
Shall I close as by-design? And maybe do something with those signatures so they don't cause confusion in the docs?

BDisp:
If the intention is indeed to use the last default button, then this needs to be well documented, and all references to the defaultButton parameters must be removed to avoid misleading the user.

tig:
> > * It drives UI consistency for users for MessageBoxes. Users can count on the last button always being defult.
> 
> This makes sense, and I don't disagree at all. I thought it was broken since the docs still describe the old behavior. Shall I close as by-design? And maybe do something with those signatures so they don't cause confusion in the docs?

How about we revise the issue and this PR to just be about the docs being wrong/misleading and fixing that?

tig (approved):

PR #5045, opened 2026-04-21
Fixes #5043 - `View.ScrollBars` now only show if they fit - Fixes `View.Viewport` bug

### LINKED ISSUE #5043: `View.ScrollBar` should no make scrollbars visible if `Padding` can't fit
I had an epiphany in the middle of the night...

I have seen two cases where Views that set `ViewportSettings.HasScrollBars` had weird layout issues that confused me.

In one case, it was in `TreeView` (#5038), and in stepping through the debugger to diagnose, when I got into `View.SetViewport` I noticed `Padding.Thickness` had thickness, but `Viewport` was `Empty`.

That's a bug. 

`View.Scrollbars` is making the scrollbar visible even if `Frame/Viewport` is `(0,0,0,0)`.  This means it's making `Padding.Thickness` have either `0,0,1,0` or `0,0,1,1` when the Frame/Viewport is 0 in that dimension. Which makes no sense and causes things to break.

The fix for View.ScrollBar is to put a check to not enable the SB if this happens.

But we should also put an assert or at least a Trace Error in `Adornment.set_Thickness` that checks this.

### LINKED ISSUE #5046: `Dim.Auto` restricts size to 4x screen size
```cs
...
        Size screenSize = us.App?.Screen.Size ?? new Size (2048, 2048);
        int autoMin = MinimumContentDim?.GetAnchor (superviewContentSize) ?? 0;
        int screenX4 = dimension == Dimension.Width ? screenSize.Width * 4 : screenSize.Height * 4;
        int autoMax = MaximumContentDim?.GetAnchor (superviewContentSize) ?? screenX4;
```

this should be

```cs
        Size screenSize = us.App?.Screen.Size ?? new Size (2048, 2048);
        int autoMin = MinimumContentDim?.GetAnchor (superviewContentSize) ?? 0;
        int autoMax = MaximumContentDim?.GetAnchor (superviewContentSize) ?? int.MaxValue;
```

This is why `Scrolling` only shows some of the Views.

tig:
## Fixes

- Fixes #5043
- Fixes #5046 

## Pull request overview

This PR targets issue #5043 by preventing built-in scrollbars from becoming visible when there isn’t enough space to render them without consuming the entire content area, and by addressing a `Viewport`/adornment interaction that could incorrectly grow `Frame` when re-applying the current `Viewport`.

**Changes:**
- Add scrollbar visibility guards (and a minimum-size rule) so scrollbars only show when they can fit.
- Adjust `View.SetViewport` behavior and add viewport/adornment regression tests for the “adornments consume viewport” scenario.
- Update `Shortcut`/Menu example and unit test code to accommodate the behavior changes.

### DIFF
--- a/Terminal.Gui/ViewBase/View.Content.cs
+++ b/Terminal.Gui/ViewBase/View.Content.cs
@@ -523,9 +523,12 @@ public partial class View
 
         Size newSize = new (viewport.Size.Width + thickness.Horizontal, viewport.Size.Height + thickness.Vertical);
 
-        if (newSize == Frame.Size || viewport.Size.Width - thickness.Horizontal < 0 || viewport.Size.Height - thickness.Vertical < 0)
+        // Detect the "adornments consume entire viewport" no-op: when the requested viewport size equals
+        // the current Viewport size (which may already be clamped to zero when Frame is smaller than adornments).
+        // In this case updating Frame would incorrectly grow it, so treat it as a no-op.
+        if (newSize == Frame.Size || viewport.Size == Viewport.Size)
         {
-            // The change is not changing the Frame, or it is invalid, so we don't need to update it.
+            // The change is not changing the Frame, or the adornments consume the entire Frame, so we don't need to update it.
             // Just call SetNeedsLayout to update the layout.
             if (_viewportLocation != viewport.Location)
             {
--- a/Terminal.Gui/ViewBase/View.Content.cs
+++ b/Terminal.Gui/ViewBase/View.Content.cs
@@ -521,11 +521,13 @@ public partial class View
 
         Thickness thickness = GetAdornmentsThickness ();
 
+        Rectangle inside = thickness.GetInside (Viewport);
+
         Size newSize = new (viewport.Size.Width + thickness.Horizontal, viewport.Size.Height + thickness.Vertical);
 
-        if (newSize == Frame.Size)
+        if (newSize == Frame.Size || inside.Width < 0 || inside.Height < 0)
         {
-            // The change is not changing the Frame, so we don't need to update it.
+            // The change is not changing the Frame, or it is invalid, so we don't need to update it.
             // Just call SetNeedsLayout to update the layout.
             if (_viewportLocation != viewport.Location)
             {
--- a/Terminal.Gui/ViewBase/View.Content.cs
+++ b/Terminal.Gui/ViewBase/View.Content.cs
@@ -1,6 +1,4 @@
-﻿using System.Diagnostics;
-
-namespace Terminal.Gui.ViewBase;
+﻿namespace Terminal.Gui.ViewBase;
 
 public partial class View
 {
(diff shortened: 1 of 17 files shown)


Terminal.Gui/ViewBase/View.Content.cs
```diff
@@ -523,9 +523,9 @@ private void SetViewport (Rectangle viewport)
 
         Size newSize = new (viewport.Size.Width + thickness.Horizontal, viewport.Size.Height + thickness.Vertical);
 
-        if (newSize == Frame.Size)
+        if (newSize == Frame.Size || viewport.Size.Width - thickness.Horizontal < 0 || viewport.Size.Height - thickness.Vertical < 0)
         {
-            // The change is not changing the Frame, so we don't need to update it.
+            // The change is not changing the Frame, or it is invalid, so we don't need to update it.
             // Just call SetNeedsLayout to update the layout.
             if (_viewportLocation != viewport.Location)
```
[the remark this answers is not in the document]
tig: @copilot explicitly detect the “adornments consume entire viewport” no-op case

Terminal.Gui/ViewBase/View.ScrollBars.cs
```diff
@@ -144,6 +144,21 @@ private void ConfigureVerticalScrollBarEvents (ScrollBar scrollBar)
...
 
+        scrollBar.VisibleChanging += (_, args) =>
+                                     {
+                                         if (!args.NewValue)
+                                         {
+                                             return;
+                                         }
+                                         int width = Frame.Size.Width - GetAdornmentsThickness ().Horizontal;
+
+                                         if (width < 1 || Viewport.Height < 2)
```
BDisp, line 155: Negative Width and Height are now possible to be assigned?

Terminal.Gui/ViewBase/View.ScrollBars.cs
```diff
@@ -166,6 +181,21 @@ private void ConfigureHorizontalScrollBarEvents (ScrollBar scrollBar)
...
 
+        scrollBar.VisibleChanging += (_, args) =>
+                                     {
+                                         if (!args.NewValue)
+                                         {
+                                             return;
+                                         }
+                                         int height = Frame.Size.Height - GetAdornmentsThickness ().Vertical;
+
+                                         if (Viewport.Width < 2 || height < 1)
```
BDisp, line 192: Negative Width and Height are now possible to be assigned?

BDisp (approved):
That sounds good to me. I just have two questions that I'm a little unsure about.

Terminal.Gui/ViewBase/View.ScrollBars.cs
tig: No, but `width < 1` is just slightly safer than `width == 0`.

BDisp: > No, but `width < 1` is just slightly safer than `width == 0`.

But I was referring to `Viewport.Height < 2` because that can never happen, right?

Terminal.Gui/ViewBase/View.ScrollBars.cs
BDisp: This `Viewport.Width < 2` can never happen, right?

PR #5059, opened 2026-04-22
Fixes #5057. Pressing `Alt-T` in a `TextField` causes a `t` to be entered

### LINKED ISSUE #5057: Pressing `Alt-T` in a `TextField` causes a `t` to be entered
Image

BDisp:
## Fixes

- Fixes #5057

## Proposed Changes/Todos

- [x] Invalidade AssociatedKey if Ctrl or Alt exist

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/AnsiHandling/KittyKeyboardPattern.cs
+++ b/Terminal.Gui/Drivers/AnsiHandling/KittyKeyboardPattern.cs
@@ -128,7 +128,17 @@ public class KittyKeyboardPattern : AnsiKeyboardParserPattern
             (key, modifierField) = NormalizeShiftedPrintableKey (key, modifierField);
         }
 
-        return string.IsNullOrEmpty (modifierField) ? key : ApplyModifiersAndEventType (modifierField, key);
+        if (!string.IsNullOrEmpty (modifierField))
+        {
+            key = ApplyModifiersAndEventType (modifierField, key);
+        }
+
+        if ((key.IsAlt || key.IsCtrl) && !string.IsNullOrEmpty (key.AssociatedText))
+        {
+            key = new Key (key) { AssociatedText = string.Empty };
+        }
+
+        return key;
     }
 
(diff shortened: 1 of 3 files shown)


tig:
I've already fixed this as part of #5062

BDisp:
> I've already fixed this as part of #5062

The fix would only be in KittyKeyboardPattern.cs and not in TextField, TextView, or any other view. I hope you reconsider this because if there's another view that needs to handle Alt+Letter, you'll have to repeat the same code everywhere.

tig:
First, my apologies for being so terse and closing this abruptly. I was in a hurry. Then I was with my wife and pissed at myself for not taking the time to explain and discuss with you.

tig:
I kinda like the idea of having associated text be available even if it's an alt press. 

If you study my fix you'll see the fixed impls (textfield, textview, hexview) all failed to check for alt/ctrl properly before. 

They had bugs that were latent because before there was no concept of associated character. But with that they were blindly consuming alt keys. 

So in think my fix is Good (capital G).

Your fix is at a much lower level, but prevents the key info from passing all the way up the stack. It doesn't negate my fix, but just means those classes are now defensive in a case where they probably don't have to be. 

So the point I'd like to debate with you is whether it is correct for associatedchar to be carried with ALL keystrokes or not. I asset it is because that's consistent. 

What are your thoughts?

BDisp:
No problem. Although my solution suppress the associated key you still have access through the Key.KeyCode. In my opinion the Key.AsGrapheme should only have a value when it's a printable key. So the Alt+keys or Ctrl+keys shouldn't be printable keys but only shortcuts. Can you please say the cases where you need them on the AsGrapheme and on the AsRune instead of having them on the KeyCode?

tig:
No, I can't. 

Let's go with both fixes. Mine makes the implementations more robust. 

Yours makes the pipeline correct.

tig (approved):

PR #5063, opened 2026-04-23
Fixes #5065. Pressing Shift-Alt-T in a TextField causes a t to be entered

### LINKED ISSUE #5065: Pressing Shift-Alt-T in a TextField causes a t to be entered
This question keeps coming up even when using the Shift key with Alt-T.
@tig I have a proposal to make. Since you made a point of stating that you would like to keep the `AssociatedText` regardless of whether the Alt or Ctrl keys are pressed, I think this is possible. We just have to control the value to be assigned to the `AsGrapheme` and `AsRune` properties. If either the Alt or Ctrl keys are pressed, `AsGrapheme` will be empty and `AsRune` will be zero. In this way, if the `AssociatedText` contains any value from the input via kitty, it can be preserved and used for other purposes that may be useful later. Do you agree with my proposal?

BDisp:
## Fixes

- Fixes #5065

## Proposed Changes/Todos

- [x] I would like to add the kitty addition test lines to reinforce the Key values. Thanks

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/AnsiHandling/KittyKeyboardPattern.cs
+++ b/Terminal.Gui/Drivers/AnsiHandling/KittyKeyboardPattern.cs
@@ -201,10 +201,9 @@ public class KittyKeyboardPattern : AnsiKeyboardParserPattern
         }
 
         int explicitModifiers = encodedModifiers - 1;
-        int explicitKeyboardModifiers = explicitModifiers & 0b111;
         bool isRelease = parts.Length > 1 && parts [1] == "3";
 
-        if (explicitKeyboardModifiers != 0 || isRelease)
+        if (isRelease)
         {
             return modifierField;
         }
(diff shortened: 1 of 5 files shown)


BDisp:
I don't think anything else needs to be changed. As far as I'm concerned, it's ready for review. Thanks.

BDisp:
@tig I think you've been having some difficulty getting certain key combinations to work correctly when pressing keys, as some implicit combinations weren't being processed and others weren't being preserved when pressing more keys.

tig (approved):

BDisp:
Still having a bug. I'll fix it.

BDisp:
I think that all ok now with press and release events.

BDisp:
@tig you approved before I made a last commit and so I don't know if I can merge this. I'll wait for you say something. Thanks.

tig (approved):
nicely done!

PR #5101, opened 2026-04-28
Fixes #5088 - Margin Shadow sizing wrong

### LINKED ISSUE #5088: ShadowStyle reports itself as Transparent even when no Margin
## Describe the bug

Default for ShadowStyle is ShadowStyle.Transparent even when there is no Margin.

Setting this property to the value it already has (i.e. that you just read) results in a shadow appearing and Margin being created.

This means that if you code gen based on this property you get e.g.

```
this.menuBar.ShadowStyle = Terminal.Gui.ViewBase.ShadowStyles.Transparent;
```

## Workaround
My current workaround for this is to inspect the View to determine if it really does have a shadow or not

```csharp
using System;
using System.Reflection;
using System.Text;
using Terminal.Gui.ViewBase;

namespace TerminalGuiDesigner.ToCode
{
    /// <summary>
    /// <para>
    /// Handles the fact that ShadowStyle is a magic property.
    /// </para>
    /// <para>
    /// Get its value will return e.g. ShadowStyles.Transparent even on a new instance but
    /// if you set its property to ShadowStyles.Transparent then a margin will appear i.e.
    /// it has side effects.
    /// </para>
    /// <para>
    /// Class handles it by looking at the View and investigating its Margin to see if a real
    /// shadow is in effect or not.
    /// </para>
    /// </summary>
    internal class ShadowStyleProperty : Property
    {
        public ShadowStyleProperty(Design design, PropertyInfo property) : base(design, property)
        {
        }

        public override object? GetValue()
        {
            if (!HasMargin())
            {
                return ShadowStyles.None;
            }

            return base.GetValue();
        }

        private bool HasMargin()
        {
            return Design.View.Margin != null
                && (Design.View.Margin.Thickness.Left != 0
                     || Design.View.Margin.Thickness.Top != 0
                     || Design.View.Margin.Thickness.Right != 0
                     || Design.View.Margin.Thickness.Bottom != 0);
        }
    }
}
```

tig:
## Summary

`Margin.ShadowStyle` getter was returning inherited values from the SuperView's shadow style, causing `View.ShadowStyle` to report a non-null value (e.g., `Transparent`) even when no shadow was ever set on the view. Writing that value back triggered side effects: creating a `MarginView` and adding shadow thickness where none should exist. This broke round-trip code generation (read a property, write it back) as reported in #5088.

Additionally, setting `ShadowStyle` to `null` did not reset `ShadowSize` on the `MarginView`, leaving stale dimensions that could cause inconsistent state.

## Fixes

- Fixes #5088

### `Terminal.Gui/ViewBase/Adornment/Margin.cs`

1. **Removed SuperView inheritance from getter** — `ShadowStyle` getter now returns the locally-set `field` value only, instead of falling back to `Parent?.SuperView?.ShadowStyle`. A view's shadow is a per-view visual, not an inherited style.
2. **Reset `ShadowSize` when clearing shadow** — The `null` branch of the setter now resets `MarginView.ShadowSize` to `Size.Empty` after calling `SetShadow(null)`, eliminating stale shadow dimensions.

### `Tests/UnitTestsParallelizable/ViewBase/Adornment/ShadowTests.cs`

1. **Added** `ShadowStyle_Getter_Does_Not_Inherit_From_SuperView` — Proves that a subview inside a Window with a shadow does not inherit the shadow style, and that round-tripping the property is a safe no-op.
2. **Added** `Setting_ShadowStyle_Null_Resets_ShadowSize` — Proves that clearing the shadow resets `ShadowSize` to `Size.Empty`.
3. **Updated** `Changing_ShadowStyle_Correctly_Set_ShadowWidth_ShadowHeight_Thickness` — Updated assertion to expect `Size.Empty` after `ShadowStyle = null` (was incorrectly asserting the stale `{2,2}`).


tig:
@tznind can you please confirm this fixes it for you?

tznind:
Looks good. I would have to build and link locally to fully confirm but dont have time at the moment.

Suggest release as code looks right and ill confirm when i next bump versions

tznind (approved):

PR #5102, opened 2026-04-28
More FileDialog tweaks

tig:
These are some fixes from code review comments that I forgot to push with #5090 .


BDisp (approved):
It might be worthwhile to correct those warning on line R36 in the FileDialog.cs file.

Issue #5108, opened 2026-04-28
Hero image needs tweaks


tig:
<img width="567" height="537" alt="Image" src="https://github.com/user-attachments/assets/d5f0785d-16e5-42fd-9b3e-0dc4ff6bc8e6" />

## "30+ Views" is meaningless; should represent "Stars"

<img width="281" height="56" alt="Image" src="https://github.com/user-attachments/assets/637d5d6c-b7a5-463d-a37a-b25eaf7e4cb4" />

tig:
@livlign interested in doing another turn?

livlign:
@tig I'd love to.
Quick question: the stars are volatile, would you prefer to add an approximate number (+10.9k stars) or an exact number at the moment (10.9k)?

tig:
Something like "+13k Stars". 

Also "+1.7M Downloads". 🤪

PR #5121, opened 2026-04-29
docs: tweak hero gif stats per #5108

livlign:
Refs #5108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 Cross-platform UI toolkit for building sophisticated terminal UI (TUI) applications on Windows, macOS, and Linux/Unix.
 
-![Terminal.Gui — cross-platform TUI toolkit for .NET. Build full-featured terminal UIs with menus, forms, tables, charts, wizards and file dialogs. 30+ stars, Windows / macOS / Linux, MIT-licensed.](docfx/images/hero.gif)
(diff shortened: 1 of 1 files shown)


livlign:
I replace "'17 since" with "+1.7M  downloads" to keep the layout intact with 4 items
Also render the gif with higher quality, the file size is larger but in turn we get better details (gradient, text sharpness)

tig (approved):
You rock. Thanks.

tznind:
I wonder if we could use 

https://github.com/arthurdick/term-to-svg

For the screen capture.  It allegedly creates crisp svg and smaller file sizes than pure gif.

This gif is 14mb which is quite a lot

livlign:
@tznind Thanks for sharing that. An interesting repo. 
I think it could be a significant reduce if we present the Terminal only. 
However, our hero also has other parts like the text, the stats with gradient, light effect, and animation too. 
What I did is put all of them to a HTML file, then export GIF from it. With these steps, replace the svg version with the current demo-video (55s, .mp4) could not help much to reduce the gif size.

tznind:
Interesting, yes thats a good point.  And thanks for sharing your methodology.

I wonder if the repo only supports 16 colors and/or if all the underline and bold escape codes are supported.

I think you could probably take the output svg and make a composite svg of your background image and then the other svg anination.

I might have a go for TerminalGuiDesigner once Im through all the qa.

Ill let you know if it looks like a goer.

PR #5124, opened 2026-04-29
Catch TypeLoadException in config property host scan

duskfold:
## Summary
- catch `TypeLoadException` while scanning loaded assemblies for configuration property hosts
- keep the existing behavior of skipping assemblies that cannot be reflected safely

## Why
`ConfigProperty.Initialize()` scans `AppDomain.CurrentDomain.GetAssemblies()` and calls `assembly.GetTypes()`.
Under some .NET 10 test and coverage runner combinations, that call can throw `TypeLoadException` instead of `ReflectionTypeLoadException`.
A real example is `Microsoft.TestPlatform.CoreUtilities`, where loading certain custom attributes can trip a `MemberNotNullWhenAttribute` conflict.

Today Terminal.Gui skips `ReflectionTypeLoadException` and `BadImageFormatException`, but not `TypeLoadException`, so module initialization can fail early with a `TypeInitializationException` before tests or headless apps even get moving.

This change makes the scan treat `TypeLoadException` the same way as the other assembly-load reflection failures it already tolerates.

## Validation
- `dotnet build --configuration Debug --no-restore`
- `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter-class '*ConfigPropertyTests'`
- `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter-class '*ConfigPropertyHostTypesTests'`
- `dotnet test --project Tests/UnitTests.NonParallelizable --no-build --filter-class '*ConfigurationMangerTests'`

### DIFF
--- a/Terminal.Gui/Configuration/ConfigProperty.cs
+++ b/Terminal.Gui/Configuration/ConfigProperty.cs
@@ -554,6 +554,10 @@ public class ConfigProperty
             {
                 continue;
             }
+            catch (TypeLoadException)
+            {
+                continue;
+            }
         }
     }
 


tig (commented):
Thank you. 

Do you have a repro case for this? It'd be wonderful to have a unit test that fails illustrating this problem.

duskfold:
We originally carried this patch because we saw a real test-environment startup failure around `ConfigProperty.Initialize()` in our .NET 10 test and coverage flow.

As of May 3, 2026, we’re no longer able to reproduce that failure on the current SDK and test-host stack, and our downstream repos now pass on stock `Terminal.Gui 2.0.1` without this patch.

Since I can’t provide a current end-to-end repro that meets the review bar, I’m going to close this PR rather than ask maintainers to merge a speculative hardening change.

If a concrete repro resurfaces with exact SDK and package versions plus a minimal failing test, I can reopen this or submit a fresh PR with that evidence.

duskfold:
I updated the original head branch with a deterministic upstream proof for the actual failure shape, but GitHub would not reopen this closed PR through either GraphQL or the REST issue state path. I opened #5293 on the same branch with the corrected proof and validation details so the review can continue there.

PR #5128, opened 2026-04-30
Fixes #5075. TableView ShowVerticalCellLines renders incorrectly with custom ColorGetter

### LINKED ISSUE #5075: `TableView` `Style.ShowVerticalCellLines` renders incorrectly
If Style.ShowVerticalCellLines is false, and a custom `ColorGetter` is used (as in `FileDialog`), where the lines would be drawn is drawn incorrectly.

Image

I was playing with different FileDialog table styles here'.

harder:
- Fixes #5075

## Summary

When `Style.ShowVerticalCellLines` is `false` and a `ColumnStyle.ColorGetter` is in use (as in `FileDialog`), the 1-char gap between cells retained the row scheme's attribute. This produced visible "stripes" of the row scheme color cutting through custom-colored cells.

## Root cause

`TruncateOrPad` intentionally renders each cell to `(Width - 1)` chars, leaving 1 char "for cell boundary". With `ShowVerticalCellLines = true` that gap is filled by the VLine glyph in `RenderSeparator`. With `ShowVerticalCellLines = false`, `RenderSeparator` was overdrawing it with a space using `rowScheme.Normal/Disabled` — so cells with a custom `ColorGetter` had a `rowScheme`-colored stripe between them.

## Fix

Two coordinated changes in `Terminal.Gui/Views/TableView/TableView.Drawing.cs`:

1. **`RenderRow`** — after drawing each cell, when `!ShowVerticalCellLines && SeparatorSymbol == ' '`, explicitly fill the 1-char right-edge gap with the cell's own `cellColor`. Viewport-clipped to handle horizontal scroll.
2. **`RenderSeparator`** — skip drawing when the separator would be an invisible space and lines are off (`RenderRow` now owns the gap).

No behavior change for `ShowVerticalCellLines = true` (VLines still drawn), non-default `SeparatorSymbol`, `FullRowSelect`, or cells without a custom `ColorGetter`.

Also flips `FileDialog`'s `TableView` to `ShowVerticalCellLines = false` so it now uses the repaired rendering path — its custom file-type colors paint cleanly across the row.

## Tests

Added `ShowVerticalCellLines_False_WithCustomColorGetter_PreservesCellColorAtSeparator` in `Tests/UnitTestsParallelizable/Views/TableViewTests.cs` — sets up a 2-column table with a custom red `ColorGetter`, draws, and asserts both the cells and the separator gap between them carry the custom red attribute. Test fails on the unfixed code and passes after the fix.

## Test plan

- [x] New regression test passes
- [x] All 102 existing TableView tests pass
- [x] Full `UnitTestsParallelizable` suite passes (16,529 tests)
- [x] `UnitTests.NonParallelizable` suite passes (63 tests)

```bash
git fetch copilot fix-5075-table-vertical-cell-lines-color
git checkout copilot/fix-5075-table-vertical-cell-lines-color
```

### DIFF
--- a/Terminal.Gui/Views/FileDialogs/FileDialog.cs
+++ b/Terminal.Gui/Views/FileDialogs/FileDialog.cs
@@ -200,7 +200,7 @@ public partial class FileDialog : Dialog, IDesignable
         _tableViewContainer.Add (_tableView);
 
         _tableView.Style.ShowHorizontalHeaderOverline = false;
-        _tableView.Style.ShowVerticalCellLines = true;
+        _tableView.Style.ShowVerticalCellLines = false;
         _tableView.Style.ShowVerticalHeaderLines = false;
         _tableView.Style.AlwaysShowHeaders = true;
         _tableView.Style.ShowHorizontalHeaderUnderline = false;
--- a/Terminal.Gui/Views/TableView/TableView.Drawing.cs
+++ b/Terminal.Gui/Views/TableView/TableView.Drawing.cs
@@ -434,6 +434,23 @@ public partial class TableView
             {
                 RenderSeparator (current.X + current.Width - 1, row, false);
             }
+
+            // When vertical cell lines are not rendered AND the separator symbol is the default
+            // invisible space, extend the cell's background color into the 1-char gap at the
+            // cell's right edge. The cell render itself only fills (Width - 1) chars (TruncateOrPad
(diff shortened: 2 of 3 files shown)


tig (changes requested):
Would you please modify file dialog to use this setting as part of this?

harder:
Good catch. Updated the TableView in FileDialog to ShowVerticalCellLines = false, so it now uses this setting. Does that look right?

PR #5129, opened 2026-04-30
Fixes #5068. TableView appends ellipsis indicator when cell content is truncated

### LINKED ISSUE #5068: TableView: Add optional ellipsis indicator when cell content is truncated
When a TableView cell value exceeds the available column width, TruncateOrPad() silently clips the text with no visual indicator:
```
  // Current behavior (80-char terminal):
  │A Very Long Description That Gets Cu│

  // Desired behavior:
  │A Very Long Description That Gets C…│
```

The TruncateOrPad method in TableView.cs takes characters until width is consumed, but adds no ellipsis:
```
  return new string (representation.TakeWhile (c =>
      (availableHorizontalSpace -= ((Rune)c).GetColumns ()) > 0).ToArray ());
```

 **Proposal**: Add a TruncationIndicator property to ColumnStyle (defaulting to "…" or Glyphs.Ellipsis) that is appended when content is truncated. When set to null or empty, preserve current silent-clip behavior for backward compatibility.

```
  public class ColumnStyle
  {
      /// <summary>Character(s) appended when cell content is truncated.
      /// Defaults to "…". Set to null to disable.</summary>
      public string? TruncationIndicator { get; set; } = "…";
  }
```

**Workaround**: Consumers can pre-truncate values in their data source lambdas (which is what I'm currently doing via a TuiHelpers.Truncate() utility). However, I believe that framework-level handling is safer and would be a nice benefit, especially since Terminal.Gui already knows the actual cell width rules.  

Consumer workarounds can be error-prone - I found that naive string.Length truncation looked easy, but then I ran into issues with CJK/emoji/full-width glyphs that made text appear to bleed across visual boundaries and had to make it more complex. 

 **Environment**: Terminal.Gui 2.0.0-rc.4, .NET 10, cross-platform

harder:
- Fixes #5068

## Summary

Adds a `TruncationIndicator` property to `ColumnStyle` so a `TableView` cell (or column header) renders a visible marker when its representation does not fit the available column width. Previously content was silently clipped, leaving no visual cue that the value continued past the cell boundary.

- New `ColumnStyle.TruncationIndicator` property defaults to `Glyphs.HorizontalEllipsis.ToString()` (`"…"`).
- Set to `null` or empty string to preserve the previous silent-clip behavior.
- `TruncateOrPad` reserves space for the indicator and appends it after the last fully-fitting grapheme cluster, so surrogate pairs / ZWJ emoji / combining marks remain intact.
- If the indicator itself would not leave room for at least one cell of content (very narrow column or oversize indicator), the value is silently clipped — no oversize result is produced.

## Tests

Added 10 parallelizable tests in `TableViewTests` covering: default value, default ellipsis appended on overflow, null/empty/null-ColumnStyle silent-clip fallbacks, custom multi-column indicator (`"..."`), indicator wider than the column budget, indicator equal to the budget, no-truncation case, and grapheme-aware truncation with the indicator (🎉 emoji).

All existing TableView tests continue to pass. Full `UnitTests.Parallelizable` run: 16,548 passed / 17 skipped / 0 failed.

```
git fetch copilot fix-5068-tableview-truncation-indicator
git checkout copilot/fix-5068-tableview-truncation-indicator
```

### DIFF
--- a/Terminal.Gui/Views/TableView/ColumnStyle.cs
+++ b/Terminal.Gui/Views/TableView/ColumnStyle.cs
@@ -63,6 +63,22 @@ public class ColumnStyle
     public bool Visible { get => MaxWidth >= 0 && field; set; } = true;
 
     /// <summary>
+    ///     Gets or sets the string appended to a cell value (or column header) when its rendered representation does
+    ///     not fit in the available column width. Defaults to <see cref="Glyphs.HorizontalEllipsis"/> ("…").
+    /// </summary>
+    /// <remarks>
+    ///     <para>
+    ///         Set to <see langword="null"/> or an empty string to suppress the indicator and silently clip the value at
+    ///         the available width.
+    ///     </para>
+    ///     <para>
+    ///         If the indicator itself is wider than the available column width the value is silently clipped (no
+    ///         indicator is drawn).
+    ///     </para>
+    /// </remarks>
+    public string? TruncationIndicator { get; set; } = Glyphs.HorizontalEllipsis.ToString ();
+
+    /// <summary>
     ///     Returns the alignment for the cell based on <paramref name="cellValue"/> and <see cref="AlignmentGetter"/>/
     ///     <see cref="Alignment"/>
     /// </summary>
--- a/Terminal.Gui/Views/TableView/TableView.Drawing.cs
+++ b/Terminal.Gui/Views/TableView/TableView.Drawing.cs
@@ -565,6 +565,14 @@ public partial class TableView
         // if value is too wide, truncate by grapheme cluster to avoid splitting surrogate pairs
         if (representation.GetColumns () >= availableHorizontalSpace)
         {
+            string indicator = colStyle?.TruncationIndicator ?? string.Empty;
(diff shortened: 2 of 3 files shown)


tig (approved):

BDisp:
The unit test `ProcessQueue_ReleasesStaleEscapeSequences_AfterTimeout` failed on macOS. I suspect it's not an isolated failure because when this happens, the number of unit tests tested is unknown and doesn't indicate any specific failure with any particular unit test. This time, the test result indicates that the number of unit tests tested was 16565 and 1 failed. I tested it on my Intel MacBook Pro and indeed, during the processing of the unit tests, a dialog box appeared twice asking if I allowed access to the Desktop folder the first time and if I allowed access to the Documents folder the second time. I suspect that the omission of these permissions prevents macOS from ending the unit test above within the time limit. Hence my suspicion that it is indeed a bug, perhaps caused by some unit test trying to access the aforementioned folders.
But I may be wrong because it is also true that this does not always happen, which makes it very difficult to identify its cause.

tig:
@harder can you fix the merge conflicts please?

PR #5131, opened 2026-04-30
Fixes #5130. Fix OutputBufferImpl race condition on Contents reference

### LINKED ISSUE #5130: Memory corruption - `ClearContents` replaces the `Contents` array reference outside any lock, then locks on the new object
Is anyone else experiencing weird crashes in the debugger? I've been seeing this for a few weeks on my main dev machine (x64), but never see it on either my real arm64 laptop or in Parallels on my macbook pro.

I suspect I may have something odd going on with this machine...

This only happens when I'm running a TG app in the debugger w/in VS 2026 (Insiders).

* Sometimes whatever app I'm debugging (typically UICatalog) just exits with no indication of what happened in the debugger.
* Sometimes I get an exception like this:

Image

* Sometimes there's a call stack (like the above); somtimes not. 

When there is an exception caught, it's almost always in/around `StringExtensions.GetColumns` or something to do with `IDriver.Contents`.

Originally, I figured it had something to do with someone passing null to code that wasn't prepared for it (e.g using `foo!.Property` or calling into a code with `#nullable disable`. This is why I've been forcing tons of clean up of nullabilty stuff. But that's had no effect. 

Here's a test that proves the issue:

```cs
using System.Collections.Concurrent;

namespace DriverTests.Output;

/// <summary>
///     Proves the race condition in <see cref="OutputBufferImpl"/> where
///     <see cref="OutputBufferImpl.ClearContents()"/> replaces the <c>Contents</c>
///     reference while <see cref="OutputBufferImpl.AddStr"/> is concurrently iterating.
///     See: https://github.com/gui-cs/Terminal.Gui/issues/5130
/// </summary>
public class OutputBufferImplConcurrencyTests
{
    /// <summary>
    ///     Concurrently calls <see cref="OutputBufferImpl.AddStr"/> and
    ///     <see cref="OutputBufferImpl.ClearContents()"/> from separate threads.
    ///     With the current broken <c>lock (Contents)</c> pattern this reproduces
    ///     crashes (AccessViolationException, NullReferenceException, or
    ///     IndexOutOfRangeException) within a few hundred iterations.
    /// </summary>
    [Fact]
    public void AddStr_And_ClearContents_Concurrent_DoesNotThrow ()
    {
        // Arrange
        OutputBufferImpl buffer = new ();
        buffer.SetSize (80, 25);

        ConcurrentBag<Exception> exceptions = [];
        const int ITERATIONS = 5_000;
        CancellationTokenSource cts = new ();

        // Writer thread — continuously writes strings
        Thread writer = new (() =>
                              {
                                  try
                                  {
                                      for (var i = 0; i < ITERATIONS && !cts.IsCancellationRequested; i++)
                                      {
                                          try
                                          {
                                              buffer.Move (0, i % 25);
                                              buffer.AddStr (new string ('A', 80));
                                          }
                                          catch (Exception ex) when (ex is not OutOfMemoryException)
                                          {
                                              exceptions.Add (ex);

                                              break;
                                          }
                                      }
                                  }
                                  catch (Exception ex)
                                  {
                                      exceptions.Add (ex);
                                  }
                              }) { IsBackground = true };

        // Resizer/clearer thread — continuously replaces Contents
        Thread clearer = new (() =>
                               {
                                   try
                                   {
                                       for (var i = 0; i < ITERATIONS && !cts.IsCancellationRequested; i++)
                                       {
                                           try
                                           {
                                               // Alternate sizes to maximize chance of IndexOutOfRange
                                               buffer.SetSize (i % 2 == 0 ? 80 : 40, i % 2 == 0 ? 25 : 10);
                                           }
                                           catch (Exception ex) when (ex is not OutOfMemoryException)
                                           {
                                               exceptions.Add (ex);

                                               break;
                                           }
                                       }
                                   }
                                   catch (Exception ex)
                                   {
                                       exceptions.Add (ex);
                                   }
                               }) { IsBackground = true };

        // Act
        writer.Start ();
        clearer.Start ();

        writer.Join (TimeSpan.FromSeconds (10));
        clearer.Join (TimeSpan.FromSeconds (10));
        cts.Cancel ();

        // Assert — with the bug present, we expect exceptions to have been collected.
        // Once fixed this test should pass with zero exceptions.
        Assert.True (exceptions.IsEmpty,
                     $"Caught {exceptions.Count} exception(s) during concurrent access. "
                     + $"First: {exceptions.FirstOrDefault ()?.GetType ().Name}: {exceptions.FirstOrDefault ()?.Message}");
    }
}

```

tig:
## Problem

`OutputBufferImpl.ClearContents` replaces the `Contents` array reference **outside** any lock, then locks on the **new** object. Meanwhile, `AddGrapheme` and `FillRect` lock on whatever `Contents` points to at call time. When both run concurrently, they can lock on different objects (or the same new object after the swap), providing **no mutual exclusion**. This corrupts `Cell` data and causes `AccessViolationException` in native Wcwidth, `NullReferenceException`, or `IndexOutOfRangeException`.

## Root Cause

Locking on `Contents` itself is fundamentally broken because `ClearContents` **replaces** the reference. A `lock(obj)` only provides mutual exclusion when all threads lock on the **same** object instance.

## Fix

Introduces a stable `readonly object _contentsLock` that is never replaced. All code that mutates `Contents`, `DirtyLines`, `Clip`, or `_urlMap` now operates inside `lock (_contentsLock)`.

### Changes to `OutputBufferImpl.cs`:

1. **Added `_contentsLock`** — a `private readonly object` that is never replaced
2. **`ClearContents(bool)`** — all mutations (Contents, Clip, DirtyLines, _urlMap, cell init loop) moved inside `lock (_contentsLock)`
3. **`AddGrapheme`** — replaced `lock (Contents)` with `lock (_contentsLock)`; moved Clip access and null guard inside the lock
4. **`FillRect(Rectangle, Rune)`** — replaced `lock (Contents!)` with `lock (_contentsLock)`; added null guard inside
5. **`SetSize`** — sets backing fields `_cols`/`_rows` directly and calls `ClearContents` once (previously triggered it 3x via property setters that each call `ClearContents`)

### New tests (`OutputBufferImplConcurrencyTests.cs`):

- `AddStr_And_ClearContents_Concurrent_DoesNotThrow`
- `FillRect_And_ClearContents_Concurrent_DoesNotThrow`
- `Move_AddStr_And_ClearContents_Concurrent_DoesNotThrow`
- `AddStr_FillRect_And_ClearContents_ThreeWay_Concurrent_DoesNotThrow`

### New benchmark (`OutputBufferBenchmark.cs`):

Measures `AddStr`, `FillRect`, `ClearContents`, `SetSize`, and `TypicalDrawCycle` performance.

## Benchmark Results (Before → After, 80×25)

| Method | Before | After | Change |
|--------|--------|-------|--------|
| AddStr_FullRow | 31.48 μs | 29.91 μs | −5% |
| AddStr_AllRows | 776.68 μs | 762.46 μs | −2% |
| FillRect_FullScreen | 666.49 μs | 674.71 μs | ~same |
| ClearContents | 75.76 μs | 76.36 μs | ~same |
| **SetSize** | **225.99 μs** | **75.62 μs** | **−67%** |
| TypicalDrawCycle | 859.06 μs | 834.68 μs | −3% |

`SetSize` is 67% faster because the redundant double `ClearContents` call was eliminated. All other operations are within noise. Zero change in memory allocations.

## Verification

- All 4 concurrency tests pass (previously would crash within a few hundred iterations)
- 16,542 parallelizable tests pass
- 63 non-parallelizable tests pass
- No performance regression

Fixes #5130


Terminal.Gui/Drivers/Output/OutputBufferImpl.cs
```diff
@@ -177,24 +175,173 @@ public void AddStr (string str)
...
+                    WriteGraphemeByWidth (c, r, grapheme, grapheme.GetColumns (), clipBounds);
+                }
+            }
+        }
+    }
+
+    /// <inheritdoc/>
+    public void FillRect (Rectangle rect, char rune)
+    {
+        for (int y = rect.Top; y < rect.Top + rect.Height; y++)
```
BDisp: Doesn't this method also need to be protected as thread-safe with a `lock`?

BDisp (approved):
Looks good. Only one question in comment.

Terminal.Gui/Drivers/Output/OutputBufferImpl.cs
tig: great catch!

tig:
None of this has fixed the issue for me. 

@BDisp have you seen the kinds of crashes i describe above?

I can get various crashes to occur with just a few minutes of playing. Sometimes at UICatalog startup.

BDisp:
> None of this has fixed the issue for me.
> 
> @BDisp have you seen the kinds of crashes i describe above?
> 
> I can get various crashes to occur with just a few minutes of playing. Sometimes at UICatalog startup.

I played around with virtually all the scenarios in UICatalog and didn't experience any crashes.

BDisp:
But in the `develop` branch, I'm noticing a significant slowdown because the `TableView` is taking too long in the `CalculateMaxCellWidth` method. Perhaps you could merge this PR because I don't see that slowdown with it.

BDisp:
Forget what I said. Even now, with this PR, I can verify this slowness. I suspect it's due to an incomplete compilation using some objects from a previous compilation. This slowness is truly frustrating.

tig:
I'm going to uninstall VS2026 and delete all its old folders.  Since you have not seen these crashes, and i don't see them on other machines, I'm going to assume they are install related. 

That said, this PR has some good thread-safety improvements, so I guess that's good.

PR #5178, opened 2026-05-05
fix: Region.XOR computes correct symmetric difference (#5167)

### LINKED ISSUE #5167: `Region.XOR` differences against the already-mutated state, producing non-XOR results
## Summary

```csharp
// Region.cs ~ line 217–222
this.Combine   (region, RegionOp.Difference);   // this becomes (this − region)
region.Combine (this,   RegionOp.Difference);   // region becomes (region − (this − region))
this.Combine   (region, RegionOp.Union);
```

The second `Combine` operates on the already-mutated `this`. The result is `(this − region) ∪ (region − (this − region))`, **not** the symmetric difference `(this ∖ region) ∪ (region ∖ this)`.

## File

`Terminal.Gui/Drawing/Region.cs:217-222`

## Why it matters

Public, named geometric operation that's wrong. Anyone using `Region.XOR` to compute a redraw delta gets phantom or missing cells.

## Proposed fix

Snapshot both operands before mutating either:

```csharp
Region originalThis   = new (this);
Region originalRegion = new (region);

this.Combine   (originalRegion, RegionOp.Difference);  // (this ∖ region)
Region rightOnly = new (originalRegion);
rightOnly.Combine (originalThis, RegionOp.Difference); // (region ∖ this)
this.Combine (rightOnly, RegionOp.Union);
```

## Proposed unit test

```csharp
// Claude - Opus 4.7
[Fact]
public void Region_XOR_ComputesSymmetricDifference ()
{
    // Two overlapping 2×2 boxes; intersection is the single cell (1,1).
    //
    //   A: (0,0) (1,0)        B:        (1,1) (2,1)
    //      (0,1) (1,1)                  (1,2) (2,2)
    //
    //   XOR(A,B) = A ∖ B  ∪  B ∖ A
    //            = { (0,0),(1,0),(0,1),(1,2),(2,1),(2,2) } - that is, EVERY cell
    //              of A and B except the shared (1,1).
    Region a = new (new Rectangle (0, 0, 2, 2));
    Region b = new (new Rectangle (1, 1, 2, 2));

    // Act
    a.Combine (b, RegionOp.XOR);

    // Assert — every "only-in-one" cell is in, the shared cell is out.
    Assert.True (a.Contains (0, 0));
    Assert.True (a.Contains (1, 0));
    Assert.True (a.Contains (0, 1));
    Assert.True (a.Contains (1, 2));
    Assert.True (a.Contains (2, 1));
    Assert.True (a.Contains (2, 2));

    Assert.False (a.Contains (1, 1)); // <-- this is the assertion that fails today
}
```

## Context

Surfaced in the v2.0.0 RC core review (`plans/v2-rc-core-review.md`, item #7).

tig:
## Summary

Fixes #5167. The `RegionOp.XOR` branch in `Region.Combine` mutated `this` via `Exclude (region)` and then computed `(region - this)` against the already-mutated `this`, producing a non-symmetric result. The fix snapshots both operands first and computes `(thisOriginal \ regionOriginal) ∪ (regionOriginal \ thisOriginal)` so the symmetric difference is correctly preserved.

A regression test in `Tests/UnitTestsParallelizable/Drawing/Region/RegionClassTests.cs` covers the canonical case where two unit-overlapping 2x2 rectangles XOR'd must exclude the single intersecting cell.

## Test plan

- [ ] `dotnet build --no-restore`
- [ ] `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter "FullyQualifiedName~Region"`
- [ ] Verify `Combine_XOR_ProducesSymmetricDifference` fails on `develop` and passes with this change.

> Note: this environment does not have the .NET SDK installed, so the local `dotnet build`/`dotnet test` commands could not be run. The fix and test were verified by tracing `SubtractRectangle` over the test inputs by hand: for `A=(0,0,2,2)`, `B=(1,1,2,2)`, the buggy code includes cell `(1,1)` in the result while the fixed code excludes it (the load-bearing assertion). CI must confirm the build and test pass.

---

### DIFF
--- a/Terminal.Gui/Drawing/Region.cs
+++ b/Terminal.Gui/Drawing/Region.cs
@@ -215,9 +215,24 @@ public class Region
                 break;
 
             case RegionOp.XOR:
-                Exclude (region);
-                region.Combine (this, RegionOp.Difference);
-                _rectangles.AddRange (region._rectangles);
+
+                // Snapshot both operands before mutating either, so the symmetric
+                // difference is computed as (thisOriginal \ regionOriginal) ∪
+                // (regionOriginal \ thisOriginal) rather than against a partially
+                // mutated `this`.
+                Region thisOriginal = Clone ();
+                Region regionOriginal = region.Clone ();
+
+                // Reuse `thisOriginal` to hold (thisOriginal \ regionOriginal).
+                thisOriginal.Combine (regionOriginal, RegionOp.Difference);
+
+                // Reuse a fresh clone of the unmodified `this` for (regionOriginal \ thisOriginal).
+                Region regionMinusThis = region.Clone ();
+                regionMinusThis.Combine (Clone (), RegionOp.Difference);
+
+                _rectangles.Clear ();
+                _rectangles.AddRange (thisOriginal._rectangles);
+                _rectangles.AddRange (regionMinusThis._rectangles);
 
                 break;
 
--- a/Tests/UnitTestsParallelizable/Drawing/Region/RegionClassTests.cs
+++ b/Tests/UnitTestsParallelizable/Drawing/Region/RegionClassTests.cs
@@ -1103,4 +1103,30 @@ public class RegionClassTests
         Assert.Null (exception);
     }
 
+    // Claude - Opus 4.7
+    [Fact]
+    public void Combine_XOR_ProducesSymmetricDifference ()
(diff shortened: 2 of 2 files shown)


BDisp (approved):
I love this implementation.

PR #5179, opened 2026-05-05
fix: Key.Equals honors GetHashCode contract; drops Handled from identity (#5170)

### LINKED ISSUE #5170: `Key.Equals` checks `Handled` but `GetHashCode` does not — breaks `KeyBindings` dictionary lookups
## Summary

```csharp
// Key.cs ~ line 695
public override bool Equals (object? obj)
{
    if (obj is Key other)
    {
        return other._keyCode == _keyCode && other.Handled == Handled;
    }
    return false;
}

public override int GetHashCode () => _keyCode.GetHashCode ();
```

`Equals` compares both `_keyCode` and `Handled`; `GetHashCode` mixes only `_keyCode`. This violates the .NET contract: equal objects must have equal hash codes (the *contrapositive* — but the asymmetry causes real-world breakage in both directions).

## File

`Terminal.Gui/Input/Keyboard/Key.cs:699,708`

## Why it matters

`KeyBindings` is `ConcurrentDictionary<Key, KeyBinding>`. Mutating `Handled` on a `Key` between binding insertion and lookup produces a key that's "equal" to the stored entry but lookups fail (or return wrong entries). It also poisons every other set/dictionary built on `Key`.

## Proposed fix

Drop `Handled` from `Equals`. `Handled` is per-event state, not part of binding identity.

## Proposed unit test

```csharp
// Claude - Opus 4.7
[Fact]
public void Key_Equals_GetHashCode_HonourTheNetContract ()
{
    // Two Keys with the same KeyCode, differing only in Handled.
    Key a = new (KeyCode.F1) { Handled = false };
    Key b = new (KeyCode.F1) { Handled = true };

    // Per docs, a and b represent the same binding identity.
    Assert.True (a.Equals (b),
                 "Two keys with the same KeyCode should be equal regardless of Handled.");

    // .NET contract: if Equals -> GetHashCode must be equal.
    Assert.Equal (a.GetHashCode (), b.GetHashCode ());

    // And the dictionary use case must work.
    ConcurrentDictionary<Key, string> bindings = new ();
    bindings [a] = "binding-A";

    Assert.True (bindings.TryGetValue (b, out string? found),
                 "Lookup with an equal-but-Handled-flipped Key must find the binding.");
    Assert.Equal ("binding-A", found);
}
```

The first assertion may flip depending on which fix we choose; the *third* (dictionary lookup) is the load-bearing one — it fails today.

## Context

Surfaced in the v2.0.0 RC core review (`plans/v2-rc-core-review.md`, item #16).

tig:
## Summary

Fixes #5170.

- `Key.Equals(object?)` previously compared `_keyCode` **and** `Handled`, while `GetHashCode()` returned only `_keyCode.GetHashCode()`. This violated the .NET contract that equal objects must have equal hash codes, and broke `KeyBindings` (a `ConcurrentDictionary<Key, KeyBinding>`) when `Handled` flipped between storage and lookup.
- `Handled` is per-event state, not part of binding identity, so it is removed from the equality comparison. `IEquatable<Key>.Equals(Key?)` delegates to the same `Equals(object?)` logic and is now consistent. `KeyEqualityComparer` already used `KeyCode`-only equality and required no change.
- Added a regression test class `KeyEqualityTests` that exercises the load-bearing dictionary lookup case plus direct `Equals`/`GetHashCode` assertions across differing `Handled` values.

### Tests updated to match the corrected contract

Two existing tests in `Tests/UnitTestsParallelizable/Input/Keyboard/KeyTests.cs` encoded the buggy behavior and have been updated. The original assertions were wrong because they required `Equals` to return `false` for two `Key` instances with the same `KeyCode` but differing `Handled`, while `GetHashCode` returned the same value for both — a direct violation of `object.Equals`/`GetHashCode` consistency requirements:

- `Equals_ShouldReturnTrue_WhenEqual` — removed the assertion that flipping `Handled` makes two otherwise-equal keys unequal.
- `Equals_Handled_Changed_ShouldReturnFalse_WhenNotEqual` → renamed to `Equals_Handled_Changed_ShouldReturnTrue_WhenHandledDiffers` and inverted to assert the correct behavior.

## Test plan

- [ ] `dotnet build` succeeds with no new warnings
- [ ] `dotnet test --project Tests/UnitTestsParallelizable --filter "FullyQualifiedName~Key"` — all Key tests pass, including the three new `KeyEqualityTests`
- [ ] Confirm the new `ConcurrentDictionary_Lookup_Succeeds_WhenHandledDiffers` test fails on `develop` without the fix (demonstrates the regression coverage)
- [ ] Verify `KeyBindings` lookup behavior is unaffected for keys with matching `Handled` state

> Note: dotnet was unavailable in the worktree where this PR was assembled, so local `dotnet build` / `dotnet test` could not be run. The change is small and well-scoped (one method body, plus tests). Please verify CI passes before merging.

---

### DIFF
--- a/Terminal.Gui/Input/Keyboard/Key.cs
+++ b/Terminal.Gui/Input/Keyboard/Key.cs
@@ -692,11 +692,16 @@ public class Key : EventArgs, IEquatable<Key>
     public static implicit operator string (Key key) => key.ToString ();
 
     /// <inheritdoc/>
+    /// <remarks>
+    ///     Two <see cref="Key"/> instances are considered equal when their <see cref="KeyCode"/> values match.
+    ///     <see cref="Handled"/> is per-event state and is intentionally excluded from identity so that
+    ///     <see cref="Equals(object?)"/> remains consistent with <see cref="GetHashCode"/>.
+    /// </remarks>
     public override bool Equals (object? obj)
     {
         if (obj is Key other)
         {
-            return other._keyCode == _keyCode && other.Handled == Handled;
+            return other._keyCode == _keyCode;
         }
 
         return false;
--- a/Tests/UnitTestsParallelizable/Input/Keyboard/KeyEqualityTests.cs
+++ b/Tests/UnitTestsParallelizable/Input/Keyboard/KeyEqualityTests.cs
@@ -1,0 +1,39 @@
+// Claude - Opus 4.7
+using System.Collections.Concurrent;
+
+namespace InputTests;
+
+public class KeyEqualityTests
+{
+    [Fact]
+    public void Equals_ReturnsTrue_WhenHandledDiffers ()
+    {
+        Key a = new (KeyCode.F1) { Handled = false };
+        Key b = new (KeyCode.F1) { Handled = true };
+
+        Assert.True (a.Equals (b));
+    }
+
+    [Fact]
+    public void GetHashCode_IsEqual_WhenHandledDiffers ()
+    {
+        Key a = new (KeyCode.F1) { Handled = false };
+        Key b = new (KeyCode.F1) { Handled = true };
+
+        Assert.Equal (a.GetHashCode (), b.GetHashCode ());
+    }
+
+    [Fact]
+    public void ConcurrentDictionary_Lookup_Succeeds_WhenHandledDiffers ()
+    {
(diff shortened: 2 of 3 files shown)


BDisp (approved):
That's the correct behavior. Very good.

PR #5182, opened 2026-05-05
fix: AnsiOutput.Dispose parks cursor on last row of inline region (#5166)

### LINKED ISSUE #5166: `AnsiOutput.Dispose` parks cursor one row past the inline region (`lastInlineRow` off-by-one)
## Summary

When inline mode is in use, `AnsiOutput.Dispose()` calculates the final cursor row as `appScreen.Y + appScreen.Height` — that's one row *past* the last drawn row. The cursor is repositioned outside the inline region on shutdown, scrolling the host terminal and leaving stray content below the app.

## File

`Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs` — `Dispose()`, in the inline-mode branch where `lastInlineRow` is computed.

## Why it matters

Inline-hosted apps are the most "polite" form factor (don't take over the screen, leave shell history intact). This bug actively undoes that politeness on every shutdown.

## Proposed fix

```csharp
int lastInlineRow = appScreen.Y + appScreen.Height - 1;   // was: appScreen.Y + appScreen.Height
```

(`Height` is exclusive; `Y + Height` is the row immediately *after* the region.)

## Proposed unit test

```csharp
// Claude - Opus 4.7
[Theory]
// Claude - Opus 4.7
[InlineData (0,  1)]
[InlineData (3,  5)]
[InlineData (10, 1)]
public void AnsiOutput_InlineDispose_ParksCursorOnLastRowOfRegion (int regionY, int regionHeight)
{
    // Arrange
    Rectangle appScreen = new (0, regionY, 80, regionHeight);
    RecordingAnsiStream stream = new ();
    AnsiOutput output = new (stream) { InlineMode = true };
    output.SetAppScreen (appScreen);

    // Act
    output.Dispose ();

    // Assert — the last cursor-position write must target the LAST row of the region,
    // not one row past it.
    int expectedAnsiRow = (regionY + regionHeight - 1) + 1; // ANSI is 1-indexed
    Assert.Contains (
        stream.WrittenStrings,
        s => s.Contains (CSI + expectedAnsiRow + ";1H"));

    int forbiddenAnsiRow = (regionY + regionHeight) + 1;
    Assert.DoesNotContain (
        stream.WrittenStrings,
        s => s.Contains (CSI + forbiddenAnsiRow + ";1H"));
}
```

## Context

Surfaced in the v2.0.0 RC core review (`plans/v2-rc-core-review.md`, item #6).

tig:
## Summary

Fixes #5166.

In inline mode, `AnsiOutput.Dispose()` computed the final cursor row as `appScreen.Y + appScreen.Height` — one row *past* the inline region. The cursor was repositioned outside the region on shutdown, scrolling the host terminal and leaving stray content below the app.

`Rectangle.Height` is exclusive; the 1-indexed last row of the region is `Y + Height - 1`.

## Changes

  - Extract the row calculation into an internal static helper `GetInlineLastRow(Rectangle)` that returns `appScreen.Y + appScreen.Height - 1` (the fix).
  - Remove the `_platform == AnsiPlatform.Degraded` early-return in `Dispose()`. Each `Write` call already short-circuits the platform-specific output path in degraded mode; the early-return only suppressed the buffered last-output that tests rely on to observe shutdown sequences. Real terminal behavior is unchanged.
- `Tests/UnitTestsParallelizable/Drivers/AnsiDriver/AnsiOutputInlineDisposeTests.cs` (new)
  - Theory test verifying `GetInlineLastRow` returns `Y + Height - 1` (and not `Y + Height`) across multiple region offsets/heights.
  - Theory test that drives `AnsiOutput.Dispose()` in inline mode with a known `appScreen` rectangle and asserts via `GetLastOutput()` that the emitted `CSI ... H` cursor-positioning sequence targets the 1-indexed last row of the region — and explicitly does not target one row past it.

## Test plan

- [x] `dotnet build Tests/UnitTestsParallelizable` succeeds.
- [x] `dotnet test --project Tests/UnitTestsParallelizable --filter-class "*AnsiOutputInlineDisposeTests*"` — all 10 cases pass with the fix applied.
- [x] Reverted the helper to the buggy `appScreen.Y + appScreen.Height` and re-ran: all 10 cases fail (`Assert.Equal()` mismatch on the row), confirming the test demonstrates the bug.
- [x] Restored the fix and re-ran: all 10 cases pass.
- [x] All other existing `*AnsiOutput*` tests in `UnitTestsParallelizable` still pass.

---

### DIFF
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
@@ -359,15 +359,35 @@ public class AnsiOutput : OutputBase, IOutput
         }
     }
 
+    /// <summary>
+    ///     Computes the 1-indexed terminal row of the last row of the inline application region.
+    /// </summary>
+    /// <remarks>
+    ///     <para>
+    ///         <see cref="Rectangle.Height"/> is exclusive: <c>Y + Height</c> is the row immediately
+    ///         <em>after</em> the region. The last row inside the region is <c>Y + Height - 1</c>.
+    ///         ANSI cursor-positioning sequences are 1-indexed, but <see cref="Rectangle.Y"/> is already
+    ///         the 1-indexed terminal row offset for the inline region, so no additional adjustment is
+    ///         required.
+    ///     </para>
+    ///     <para>
+    ///         Used by <see cref="Dispose"/> to park the cursor on the final row of the inline region
+    ///         on shutdown. See issue #5166.
+    ///     </para>
+    /// </remarks>
+    /// <param name="appScreen">The inline application region rectangle.</param>
+    /// <returns>The 1-indexed terminal row of the last row of the region.</returns>
+    internal static int GetInlineLastRow (Rectangle appScreen) => appScreen.Y + appScreen.Height - 1;
+
     /// <inheritdoc/>
     public void Dispose ()
     {
         try
         {
-            if (_platform == AnsiPlatform.Degraded)
-            {
-                return;
-            }
+            // Note: each Write below short-circuits the platform-specific output path when
+            // <see cref="_platform"/> is <see cref="AnsiPlatform.Degraded"/>, but still
+            // appends to the buffered last-output (see <see cref="OutputBase.Write(StringBuilder)"/>),
+            // which keeps shutdown sequences observable to tests.
 
             // Restore terminal state: disable mouse, reset attributes, show cursor
             Write (EscSeqUtils.CSI_DisableMouseEvents);
@@ -380,7 +400,7 @@ public class AnsiOutput : OutputBase, IOutput
                 // Position to the last row of the inline region (guaranteed to exist),
                 // then write \n to scroll the terminal if at the bottom edge.
                 Rectangle appScreen = AppScreenGetter?.Invoke () ?? default;
-                int lastInlineRow = appScreen.Y + appScreen.Height; // 1-indexed last row
+                int lastInlineRow = GetInlineLastRow (appScreen);
                 Write (EscSeqUtils.CSI_SetCursorPosition (lastInlineRow, 1));
                 Write ("\n");
             }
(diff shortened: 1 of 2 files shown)


tig:
This needs user-testing of Inline apps like mdv, ai, OCGV, etc to validate. I think the +1 was intentional.

BDisp:
I suddenly remembered that in Unix you need to use +1 because the cursor is 1-based, while in Windows it's zero-based. Could that be the reason?

tig:
Good catch — flagging this as draft pending your testing. @BDisp's note about Unix 1-based vs Windows 0-based cursor coordinates is a real platform distinction my PR didn't account for: I changed `appScreen.Y + appScreen.Height` to `appScreen.Y + appScreen.Height - 1` uniformly across platforms, but if the original `+1` was carrying the 1-based-cursor offset for Unix only, my "fix" is wrong on Unix and right on Windows (or vice versa).

The clean way to resolve this is probably to look at how `SetCursorPositionImpl` (around `AnsiOutput.cs:316-324`) does its 0→1 indexing — if it's already adding `+1` there for ANSI's 1-based protocol, then the original `appScreen.Y + appScreen.Height` was already accounting for that and `-1` is over-correcting. I'll defer to the inline-app testing you mention; happy to redo the fix once you've nailed down which side has the off-by-one.

Marking draft so it doesn't sit in your review queue. Close it if you'd rather start over.

---

tig:
Closing — on reflection (with @BDisp's and Copilot's coordinate notes) my "off-by-one fix" was the regression, not the original code.

`IApplication.Screen.Y` is 0-indexed (`AnsiSizeMonitor` stores `row - 1`); ANSI cursor positioning is 1-indexed. For an inline region at `Y=10, Height=5`, the last row in 0-indexed coords is `14`, which in 1-indexed ANSI is `15`. The original `appScreen.Y + appScreen.Height = 15` lands on the last row of the region. My change to `Y + Height - 1 = 14` parks the cursor one row *above* the last row.

The +1 was intentional. Closing #5182 unmerged. I'll let issue #5166 stand for whoever wants to verify with the inline apps you mentioned (mdv, ai, OCGV) — there may still be a real bug there but it isn't what this PR fixed.

---

Issue #5168, opened 2026-05-04
`Region.DrawOuterBoundary` draws regions one cell too tall and wide (off-by-one in line lengths)


tig:
## Summary

`Region.DrawOuterBoundary` already self-flags this in a `// BUGBUG` block: rendered boundaries are 1 pixel too tall and wide. The grid allocation is correct (`bounds.Width + 1, bounds.Height + 1`), but the line-length computations add a stray `+1`:

```csharp
// e.g. Region.cs ~ line 1116 / 1130
int length = x - startX + 1;   // should be `x - startX`
```

The grid iteration already includes the endpoint, so the additional `+1` overshoots by one cell.

## File

`Terminal.Gui/Drawing/Region.cs:986-1031` (BUGBUG comment), with the bad lengths at ~1116 and ~1130.

## Why it matters

Visible mis-render. This is the kind of bug that any user with a `Border`/`Frame`/region-based decoration will see immediately.

## Proposed fix

Drop the `+1` on both length computations.

## Proposed unit test

`Tests/UnitTestsParallelizable`.

```csharp
// Claude - Opus 4.7
[Fact]
public void DrawOuterBoundary_DoesNotOvershootRegionBounds ()
{
    // Arrange — a 5×5 region at (10, 10)
    Rectangle bounds = new (10, 10, 5, 5);
    Region region = new (bounds);
    OutputBufferImpl buffer = new ();
    buffer.SetSize (40, 40);

    // Act
    region.DrawOuterBoundary (buffer, Glyphs.HLine, Glyphs.VLine);

    // Assert — no boundary glyph may be drawn outside [bounds.X, bounds.Right) × [bounds.Y, bounds.Bottom)
    for (int row = 0; row < buffer.Rows; row++)
    {
        for (int col = 0; col < buffer.Cols; col++)
        {
            if (buffer.Contents! [row, col].IsDirty)
            {
                Assert.True (col >= bounds.X && col < bounds.Right,
                             $"Cell ({col},{row}) is outside region.X..Right but was drawn (overshoot).");
                Assert.True (row >= bounds.Y && row < bounds.Bottom,
                             $"Cell ({col},{row}) is outside region.Y..Bottom but was drawn (overshoot).");
            }
        }
    }

    // And confirm the boundary actually lands on the corners (not just shrunk to nothing)
    Assert.True (buffer.Contents! [10, 10].IsDirty,        "Top-left corner not drawn.");
[...]

PR #5183, opened 2026-05-05
refactor: remove Region.DrawOuterBoundary debug API (#5168)

tig:
## Summary

This PR replaces the original "fix" approach for #5168 with a deletion of the buggy API.

Per maintainer @tig: *"I actually think `DrawOuterBoundary` should be deleted. It was added as a debug API while I was developing `Region`."*

- Removes `Region.DrawOuterBoundary` (and its `BUGBUG` comment block) from `Terminal.Gui/Drawing/Region.cs`. The shared helpers it used (`GetBounds`, `DrawBoundaries`) are kept — they have other callers.
- Confirmed zero internal callers of `DrawOuterBoundary` across the framework via grep. The only consumer was the `RegionScenario` UICatalog demo.
- Refactors `Examples/UICatalog/Scenarios/RegionScenario.cs` to iterate `Region.GetRectangles()` and add the four edges of each rectangle directly to the `LineCanvas`. Shared edges between adjacent rectangles overlap; `LineCanvas`'s line-style joining handles the visual result, matching how `Border`/`Adornment` already draw. A short comment notes this is a demo simplification.
- Deletes `Tests/UnitTestsParallelizable/Drawing/Region/DrawOuterBoundaryTests.cs` (30+ structural smoke tests). None used `DriverAssert` or verified visual output, so nothing of value is lost.

This supersedes the earlier fix attempt on this branch (the branch was reset to `origin/develop` and rebuilt as a single coherent commit).

## Test plan

- [x] `dotnet build Terminal.Gui /p:DisableGitVersionTask=true` — succeeds, zero new warnings
- [x] `dotnet build Examples/UICatalog/UICatalog.csproj /p:DisableGitVersionTask=true` — `RegionScenario` compiles
- [x] `dotnet test Tests/UnitTestsParallelizable --filter-method "*Region*"` — all 160 remaining Region tests pass; the 30+ deleted `DrawOuterBoundaryTests` are gone
- [ ] Maintainer eyeball of the updated `RegionScenario` "Outer Boundary" draw mode in UICatalog (demo-only behaviour change)


tig:
I actually think DrawOuterBoundary should be deleted. It was added as a debug API while I was developoing Region.

tig:
That matches what I found independently. Solution-wide grep for `DrawOuterBoundary`:

| Caller | File | Notes |
|---|---|---|
| (none in `Terminal.Gui/`) | — | No production code in the library uses it. |
| `RegionScenario.DrawTools` | `Examples/UICatalog/Scenarios/RegionScenario.cs:110` | Demo / debug viewer only. |
| 30+ tests | `Tests/UnitTestsParallelizable/Drawing/Region/DrawOuterBoundaryTests.cs` | All structural smoke tests — every single one only asserts `canvas.GetCellMap()` is not empty, which is why a self-flagged BUGBUG sat shipping. None of them would have caught the off-by-one. |

Want me to convert this PR to *delete* `DrawOuterBoundary` and the accompanying test file, plus inline whatever the `RegionScenario` demo needs (or refactor it to use `LineCanvas` directly)? I'd rather not just delete the demo's call without giving it an equivalent.

If yes, I'll re-spin the PR. If you'd rather take it over yourself, close this one and I'll close the issue (#5168) wontfix.

---

BDisp:
It really seems obvious that `DrawOuterBoundary` was calculating incorrectly, and the unit tests in question were removed. But are there unit tests that prove how it currently behaves, such as filling and drawing outer boundaries?

BDisp (approved):
Very well. The answer to my question is in PR #5178.

PR #5184, opened 2026-05-05
fix: ApplicationImpl.Begin raises SessionBegun after state changes (#5162)

### LINKED ISSUE #5162: CWP violation: `ApplicationImpl.Begin()` raises `SessionBegun` before setting `IsRunning`/`IsModal`
## Summary

`ApplicationImpl.Begin()` invokes the `SessionBegun` event **before** setting `IsRunning(true)` and `IsModal(true)` on the runnable. Per [`.claude/rules/cwp-pattern.md`](../blob/develop/.claude/rules/cwp-pattern.md), all work must complete before the notification, so subscribers must see the runnable in its post-state. Today they see it as not-yet-running, not-yet-modal.

## File

`Terminal.Gui/App/ApplicationImpl.Run.cs` — inside the `Begin()` critical section, around the `SessionBegun?.Invoke(...)` call. State setters `runnable.SetIsRunning(true)` and `runnable.SetIsModal(true)` immediately follow.

## Why it matters

This locks in an incorrect contract for v2.0. Any subscriber that reads `runnable.IsRunning` or `runnable.IsModal` from a `SessionBegun` handler will see `false`. The whole point of CWP is "raise after the state change," and this is the entry point.

## Proposed fix

Move `SessionBegun?.Invoke(this, new SessionTokenEventArgs(token));` after `runnable.SetIsRunning(true)` and `runnable.SetIsModal(true)` (or release the lock and raise the event after the critical section, mirroring the existing pattern for the `IsModalChanged`/`IsRunningChanged` events).

## Proposed unit test

`Tests/UnitTests.NonParallelizable` (uses application lifecycle).

```csharp
// Claude - Opus 4.7
[Fact]
public void Begin_RaisesSessionBegun_AfterIsRunningAndIsModalAreTrue ()
{
    // Arrange
    ApplicationImpl app = new ();
    app.Init ();
    Toplevel runnable = new ();
    bool? observedIsRunning = null;
    bool? observedIsModal = null;

    app.SessionBegun += (_, args) =>
                        {
                            observedIsRunning = args.SessionToken.Runnable!.IsRunning;
                            observedIsModal = args.SessionToken.Runnable!.IsModal;
                        };

    // Act
    SessionToken? token = app.Begin (runnable);

    // Assert — CWP requires the state change to be complete before the event fires
    Assert.True (observedIsRunning, "SessionBegun observed IsRunning=false; CWP requires state change before notification.");
    Assert.True (observedIsModal,   "SessionBegun observed IsModal=false; CWP requires state change before notification.");

    // Cleanup
    app.End (token!);
    app.Shutdown ();
}
```

This test will fail today (the booleans are read as `false`) and pass once the event is moved after the setters.

## Context

Surfaced in the v2.0.0 RC core review (`plans/v2-rc-core-review.md`, item #1 in the verification ledger).

tig:
Fixes #5162.

## Summary

- `ApplicationImpl.Begin` violated the Cancellable Workflow Pattern (CWP): it invoked the `SessionBegun` event inside the session-stack lock and **before** calling `SetIsRunning(true)` / `SetIsModal(true)` on the runnable. Subscribers observed a `SessionToken` whose `Runnable.IsRunning` and `IsModal` were still `false`.
- Move the state mutations (`SetIsRunning(true)`, `SetIsModal(true)`, `previousTop?.SetIsModal(false)`) inside the lock, and raise `SessionBegun` **after** the lock is released. This mirrors the existing symmetric `End()` / `SessionEnded` pattern.
- Add a regression test in `Tests/UnitTests.NonParallelizable` that subscribes to `SessionBegun` and asserts `IsRunning == true` and `IsModal == true` on the event's `SessionToken.Runnable`.

The test was empirically verified to FAIL on `develop` (with the original ordering) and PASS after the fix. All 14 existing `BeginEnd` tests and the broader `ApplicationTests` suite (548 tests) continue to pass.

## Test plan

- [x] New test `SessionBegunCwpTests.SessionBegun_Raised_After_IsRunning_And_IsModal_Set_True` fails on unfixed code (observed `IsRunning == false`)
- [x] New test passes after the fix
- [x] `dotnet build` succeeds
- [x] Existing `BeginEndTests` (14 tests) pass
- [x] Broader `ApplicationTests` namespace in `UnitTestsParallelizable` passes (548/548 + 2 pre-existing skips)

---

### DIFF
--- a/Terminal.Gui/App/ApplicationImpl.Run.cs
+++ b/Terminal.Gui/App/ApplicationImpl.Run.cs
@@ -151,7 +151,7 @@ internal partial class ApplicationImpl
             TopRunnable = runnable;
 
             // Update cached state atomically - IsRunning and IsModal are now consistent
-            SessionBegun?.Invoke (this, new SessionTokenEventArgs (token));
+            // (CWP: state mutations happen BEFORE the SessionBegun notification below)
             runnable.SetIsRunning (true);
             runnable.SetIsModal (true);
 
@@ -160,6 +160,10 @@ internal partial class ApplicationImpl
         }
 
         // END CRITICAL SECTION - IsRunning/IsModal now thread-safe
+
+        // CWP: Raise SessionBegun AFTER all state mutations so subscribers
+        // observe a fully-consistent session token (IsRunning/IsModal == true).
+        SessionBegun?.Invoke (this, new SessionTokenEventArgs (token));
 
         // Fire events AFTER lock released (avoid deadlocks in event handlers)
         previousTop?.RaiseIsModalChangedEvent (false);
--- a/Tests/UnitTests.NonParallelizable/Application/SessionBegunCwpTests.cs
+++ b/Tests/UnitTests.NonParallelizable/Application/SessionBegunCwpTests.cs
@@ -1,0 +1,58 @@
+#nullable enable
+namespace UnitTests.NonParallelizable.ApplicationTests;
+
+/// <summary>
+///     Regression tests for issue #5162. Ensures that <see cref="IApplication.SessionBegun"/>
+///     is raised AFTER all session state mutations, per the Cancellable Workflow Pattern (CWP):
+///     subscribers must observe a fully-consistent <see cref="SessionToken"/> with
+///     <see cref="IRunnable.IsRunning"/> and <see cref="IRunnable.IsModal"/> already set to
+///     <see langword="true"/>.
+/// </summary>
+public class SessionBegunCwpTests
+{
+    // Claude - Opus 4.7
+    [Fact]
+    public void SessionBegun_Raised_After_IsRunning_And_IsModal_Set_True ()
+    {
+        ApplicationImpl.ResetModelUsageTracking ();
+
+        IApplication app = Application.Create ();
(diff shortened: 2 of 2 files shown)


tig:
Closing this PR — on reflection the issue framing was wrong.

`IsRunning` is documented (`IRunnable.cs:80`) as a cached mirror of "is on `SessionStack`," and `SessionBegun`'s XML doc only promises "Raised when `Begin(IRunnable)` has been called and has created a new `SessionToken`." The contract doesn't promise post-`SetIsRunning` ordering — it's a token-creation hook, not a "now running" notification. Subscribers who want post-state-change should listen to `IsRunningChanged`.

The original placement before `SetIsRunning` is consistent with that weak contract. The asymmetry with `End()`/`SessionEnded` (which fires last) is real but reflects an actual design distinction: "session created" vs. "session ended/disposed." This PR's fix re-asserts a stronger contract the public API never promised.

Closing #5184 unmerged. Issue #5162 is being closed wontfix in favor of a small doc-only PR that tightens the XML on `SessionBegun`/`SessionEnded` to make the hook semantics explicit.

---

PR #5185, opened 2026-05-05
fix: ApplicationImpl.Invoke throws NotInitializedException after Dispose (#5163)

### LINKED ISSUE #5163: `ApplicationImpl.Invoke()` thread-affinity fast path races with `Dispose`
## Summary

`ApplicationImpl.Invoke()` runs an unsynchronized fast path:

```csharp
if (TopRunnableView is IRunnable { IsRunning: true } && MainThreadId == Thread.CurrentThread.ManagedThreadId)
{
    action.Invoke ();
    return;
}
```

Both `MainThreadId` and `TopRunnableView` can change between the check and the invocation (e.g. another thread is in `Dispose()`/`ResetState()`). The action then runs on a stale main thread, mutating UI state from a non-UI thread, or against a stopped queue.

## File

`Terminal.Gui/App/ApplicationImpl.Run.cs` — both `Invoke(Action<IApplication>?)` and `Invoke(Action)` overloads.

## Why it matters

UI state corruption is a tail-end behaviour that's hard to reproduce but easy to ship by accident. Locking down the contract pre-2.0 is much cheaper than chasing flake reports later.

## Proposed fix

Either remove the fast path (always go through `TimedEvents.Add`) or take the lock, snapshot `MainThreadId` to a local, and, after `Dispose`, throw `NotInitializedException` so callers cannot silently drop work.

## Proposed unit test

`Tests/UnitTests.NonParallelizable` (touches application lifecycle).

```csharp
// Claude - Opus 4.7
[Fact]
public void Invoke_AfterDispose_DoesNotSilentlyDropAction ()
{
    // Arrange
    ApplicationImpl app = new ();
    app.Init ();
    app.Shutdown ();              // app is now disposed; MainThreadId is reset
    bool ran = false;

    // Act / Assert — calling Invoke on a disposed app must be observable to the caller.
    Assert.Throws<NotInitializedException> (() => app.Invoke (() => ran = true));
    Assert.False (ran, "Action should not have executed after Dispose.");
}
```

A targeted race test (deliberately disposing on another thread mid-`Invoke`) is also reasonable but inherently flaky — the contract test above is enough to drive the right fix.

## Context

Surfaced in the v2.0.0 RC core review (`plans/v2-rc-core-review.md`, item #3).

tig:
Fixes #5163.

## Summary

- Both `IApplication.Invoke(Action)` and `IApplication.Invoke(Action<IApplication>)` overloads previously dropped the action silently when called before `Init()` or after `Dispose()`: the thread-affinity fast path skips (`MainThreadId` is null and there is no top runnable) and the `TimedEvents.Add` fallback queues onto a stopped timer queue.
- Add an `Initialized` guard at the top of both overloads so the pre-Init / post-Dispose case fails loudly with `NotInitializedException`. This matches the existing guard already used by `Run(IRunnable, …)` in the same partial.
- Add `<exception>` documentation to the `IApplication.Invoke` interface contracts so the new behavior is part of the spec.
- Add four regression tests in `Tests/UnitTests.NonParallelizable/Application/ApplicationInvokeLifecycleTests.cs` (one per overload, before-Init and after-Dispose).

## Out of scope

The deeper thread-affinity race in the fast-path between reads of `TopRunnableView` / `MainThreadId` and concurrent shutdown is **intentionally not addressed** here. Issue #5163 calls that out as a redesign; this PR ships only the narrow, testable contract change (loud failure post-Dispose).

## Test plan

- [x] Added `ApplicationInvokeLifecycleTests` (4 tests) covering both overloads, pre-Init and post-Dispose.
- [x] Confirmed the new tests **fail on `develop`** (production fix stashed) — current code drops the action silently.
- [x] Confirmed the new tests **pass with the fix**.
- [x] Ran all `*Invoke*` and `ApplicationImplTests` tests in both `UnitTests.NonParallelizable` and `UnitTestsParallelizable` — no regressions (16 + 103 + 13 passing).
- [x] No new build warnings.

---

### DIFF
--- a/Terminal.Gui/App/ApplicationImpl.Run.cs
+++ b/Terminal.Gui/App/ApplicationImpl.Run.cs
@@ -57,7 +57,7 @@ internal partial class ApplicationImpl
     {
         if (!Initialized)
         {
-            throw new NotInitializedException (@"Init must be called before Invoke.");
+            throw new NotInitializedException (nameof (Invoke));
         }
 
         // If we are already on the main UI thread
@@ -80,9 +80,11 @@ internal partial class ApplicationImpl
     /// <inheritdoc/>
     public void Invoke (Action action)
     {
+        ArgumentNullException.ThrowIfNull (action);
+
         if (!Initialized)
         {
-            throw new NotInitializedException (@"Init must be called before Invoke.");
+            throw new NotInitializedException (nameof (Invoke));
         }
 
         // If we are already on the main UI thread
--- a/Terminal.Gui/App/ApplicationImpl.Run.cs
+++ b/Terminal.Gui/App/ApplicationImpl.Run.cs
@@ -55,6 +55,11 @@ internal partial class ApplicationImpl
     /// <inheritdoc/>
     public void Invoke (Action<IApplication>? action)
     {
+        if (!Initialized)
+        {
+            throw new NotInitializedException (@"Init must be called before Invoke.");
+        }
+
         // If we are already on the main UI thread
         if (TopRunnableView is IRunnable { IsRunning: true } && MainThreadId == Thread.CurrentThread.ManagedThreadId)
         {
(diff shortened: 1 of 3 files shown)


BDisp (approved):
Excellent. This treatment is very useful, as some background task might call the Invoke method after the app has been terminated.

PR #5186, opened 2026-05-05
fix: UnixRawModeHelper safe restore on crash + valid-state guard (#5164)

### LINKED ISSUE #5164: Unix raw mode left enabled if input thread crashes before driver `Dispose()`
## Summary

`UnixRawModeHelper.TryEnable()` saves termios and switches the terminal into raw mode, but `Restore()` is only called from `Dispose()`. If the input thread dies on an unhandled exception, or if the process exits abnormally before the driver disposes, the user's shell is left in raw mode — line discipline broken, no echo, no signals — until they `reset(1)`.

## File

`Terminal.Gui/Drivers/UnixHelpers/UnixRawModeHelper.cs`

## Why it matters

For 2.0 this is the single most user-visible "the library wrecked my terminal" failure mode. Every embedded user crashes their app at some point during development; the floor experience must be "the shell still works."

## Proposed fix

Belt-and-braces:

1. Hook `AppDomain.CurrentDomain.UnhandledException` and `Console.CancelKeyPress` to call `Restore()`.
2. Add a finalizer (or `CriticalFinalizerObject` derivative) that calls `Restore()` as a last resort.
3. Add a `_haveSavedTermios` flag so a crashed `TryEnable()` doesn't write garbage termios on `Restore()`.

## Proposed unit test

`Tests/UnitTestsParallelizable` (Linux only; gated by `OSPlatform.Linux` / `OSPlatform.OSX`).

The cleanest test mocks the native `tcsetattr`/`tcgetattr` syscalls behind an interface and asserts `Restore()` is invoked from the unhandled-exception path:

```csharp
// Claude - Opus 4.7
[Fact]
[Trait ("Category", "Unix")]
public void RawMode_RestoredOnUnhandledException ()
{
    // Skip on non-Unix
    if (!RuntimeInformation.IsOSPlatform (OSPlatform.Linux) && !RuntimeInformation.IsOSPlatform (OSPlatform.OSX))
    {
        return;
    }

    // Arrange — fake termios syscalls so we can observe Restore()
    var fakeTermios = new FakeTermiosSyscalls ();
    UnixRawModeHelper helper = new (fakeTermios);
    helper.TryEnable ();

    // Act — simulate the input thread throwing without ever reaching Dispose
    AppDomain.CurrentDomain.UnhandledException += (_, _) => { /* default-handled */ };
    try
    {
        Task.Run (() => throw new InvalidOperationException ("simulated input-thread death")).Wait ();
    }
    catch
    { /* swallow for the test */ }

    // Give the unhandled-exception hook a moment to fire
    Thread.Sleep (50);

    // Assert
    Assert.True (fakeTermios.RestoreCalled, "Restore() must run when the input thread dies before Dispose.");
}
```

If introducing a syscall abstraction is too invasive, an alternative integration test would `Process.Start()` a tiny consumer that calls `TryEnable()` then throws, then assert from the parent that `tcgetattr` of the controlling tty matches the original termios.

## Context

Surfaced in the v2.0.0 RC core review (`plans/v2-rc-core-review.md`, item #4).

tig:
Fixes #5164.

## Summary

When the input thread crashes before the driver `Dispose()` runs (or the process is aborted), the Unix terminal is left in raw mode and the user's shell becomes unusable. This PR adds three safety nets to `UnixRawModeHelper` so the original `termios` is restored on any abnormal exit path, plus a valid-state guard so a failed `TryEnable()` can never cause a later `Restore()` to write garbage back to the terminal.

- **Finalizer** — last-resort restore if `Dispose()` is bypassed (e.g. fatal native crash, immediate process abort). Skipped unless a valid `termios` snapshot was captured. Wrapped in `try/catch` because finalizers must never throw.
- **`AppDomain.ProcessExit` hook** — registered only after a successful `TryEnable()`, unregistered in `Dispose()` so the helper is not leaked for the lifetime of the process.
- **`Console.CancelKeyPress` hook** — so Ctrl+C also restores cleanly. Hook registration is wrapped because some hosts (redirected stdin) can throw on subscribe; the `ProcessExit` hook still applies in that case.
- **`_haveSavedTermios` guard** — `Restore()` is a hard no-op unless `tcgetattr` previously succeeded. Prevents writing an uninitialized `termios` struct back to the terminal in any failure path.

`Dispose()` now also calls `GC.SuppressFinalize(this)` since cleanup ran deterministically.

## Test plan

Tests added to `Tests/UnitTestsParallelizable` (gated `[Trait("Category","Unix")]`, but observable on any platform — the platform gate inside `TryEnable` short-circuits non-Unix runs to a "never enabled" state, which is exactly the condition the safety guards must protect against):

- [x] `Restore_WithoutTryEnable_IsNoOp_AndDoesNotThrow` — verifies the `_haveSavedTermios` guard.
- [x] `Restore_IsIdempotent` — verifies repeated `Restore()` calls are safe.
- [x] `Dispose_WithoutTryEnable_IsNoOp_AndDoesNotThrow` — including double-Dispose.
- [x] `Type_HasFinalizer_AsLastResortRestore` — reflection-asserts the finalizer is declared on `UnixRawModeHelper` (the contract for the last-resort restore path).
- [x] `TryEnable_OnNonUnix_ReturnsFalse_AndLeavesNoSavedTermios` — verifies the failure path does not poison `Restore()`.
- [x] `TryEnable_WhenSucceeds_HooksProcessExit_AndDisposeUnhooks` — only meaningful on a real tty; otherwise short-returns.

**Verified that tests demonstrate a real change vs `develop`:** flipped the production fix off (restored `develop` version of `UnixRawModeHelper.cs`), 2 of 6 tests fail (finalizer assertion + the `_haveSavedTermios` reflection check). With the fix in place all 6 pass on Linux .NET 10.

### What is integration-only

The end-to-end "input thread throws while raw mode is active, process aborts, terminal is restored" scenario requires a real PTY and cannot run as a unit test. The reflection / state-guard tests cover the observable contract that issue #5164 depends on.

---

### DIFF
--- a/Terminal.Gui/Drivers/UnixHelpers/UnixRawModeHelper.cs
+++ b/Terminal.Gui/Drivers/UnixHelpers/UnixRawModeHelper.cs
@@ -24,7 +24,11 @@ namespace Terminal.Gui.Drivers;
 ///         To guarantee the terminal is restored even when the input thread crashes or the
 ///         process exits abnormally, this class registers a process-exit hook, a
 ///         <see cref="Console.CancelKeyPress"/> handler, and a finalizer that all call
-///         <see cref="Restore"/>. See issue #5164.
+///         <see cref="Restore"/>. See issue #5164. Note that disabling ISIG means a keyboard
+///         Ctrl+C is delivered as a 0x03 byte rather than SIGINT, so the
+///         <see cref="Console.CancelKeyPress"/> hook is unlikely to fire while raw mode is
+///         active; the primary safety nets are <see cref="AppDomain.ProcessExit"/> and the
+///         finalizer.
 ///     </para>
 /// </remarks>
 internal sealed class UnixRawModeHelper : IDisposable
@@ -54,7 +58,15 @@ internal sealed class UnixRawModeHelper : IDisposable
 
         try
         {
-            tcsetattr (STDIN_FILENO, TCSANOW, ref _originalTermios);
+            int result = tcsetattr (STDIN_FILENO, TCSANOW, ref _originalTermios);
+
+            if (result != 0)
+            {
+                int errno = Marshal.GetLastWin32Error ();
+
+                // Best-effort log only; finalizers must never throw.
+                Logging.Warning ($"tcsetattr failed during finalizer (errno={errno}). Terminal may still be in raw mode.");
+            }
         }
         catch
         {
@@ -168,7 +180,18 @@ internal sealed class UnixRawModeHelper : IDisposable
 
         try
         {
-            tcsetattr (STDIN_FILENO, TCSANOW, ref _originalTermios);
+            int result = tcsetattr (STDIN_FILENO, TCSANOW, ref _originalTermios);
+
+            if (result != 0)
+            {
+                int errno = Marshal.GetLastWin32Error ();
+                Logging.Warning ($"tcsetattr failed during restore (errno={errno}). Terminal may still be in raw mode.");
+
+                // Leave IsRawModeEnabled set to true so callers (and the finalizer) know
+                // the terminal was not successfully restored.
+                return;
+            }
+
             IsRawModeEnabled = false;
             Logging.Information ("Unix terminal settings restored.");
         }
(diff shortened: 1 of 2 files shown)


BDisp (approved):
Great. In extreme cases, it always provides a way to obtain information about the terminal's status.

PR #5187, opened 2026-05-05
fix: AnsiOutput.Dispose flushes cleanup ANSI sequences (#5165)

### LINKED ISSUE #5165: `AnsiOutput.Dispose()` does not flush after writing cleanup ANSI sequences
## Summary

`AnsiOutput.Dispose()` writes the standard cleanup choreography — disable mouse, reset SGR, leave alternate buffer, restore cursor — but does not flush before returning. On a slow stdout (or under load) the process can exit with cleanup bytes still in the kernel write buffer; the host terminal is left with mouse capture on, attributes corrupt, and the alt-buffer still active.

## File

`Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs` — the `Dispose()` cleanup block.

## Why it matters

Same severity class as the Unix raw-mode issue: a 2.0 user whose app crashes (or exits cleanly under heavy stdout load) is left looking at a wrecked terminal and blames the library.

## Proposed fix

Call the native flush after each critical write — or, simpler, once at the end of the `finally` block in `Dispose()`. Make sure the flush is on the same write path used by the cleanup writes (raw `write(2)` rather than the buffered `Console` API, depending on driver).

## Proposed unit test

`Tests/UnitTestsParallelizable` — uses a recording stream that the driver writes to.

```csharp
// Claude - Opus 4.7
[Fact]
public void AnsiOutput_Dispose_FlushesCleanupSequences ()
{
    // Arrange — back the driver with a stream that records Flush() calls and write order
    RecordingStream recorder = new ();
    AnsiOutput output = new (recorder);
    output.Initialize ();

    // Act
    output.Dispose ();

    // Assert — every cleanup write must be observable AFTER Dispose returns
    Assert.Contains (AnsiSequences.DisableMouseTracking, recorder.WrittenBytes);
    Assert.Contains (AnsiSequences.SgrReset,             recorder.WrittenBytes);
    Assert.Contains (AnsiSequences.LeaveAlternateBuffer, recorder.WrittenBytes);

    // The last operation before Dispose returns must be a flush; otherwise the host terminal
    // can be left with mouse on / SGR set / alt-buffer active.
    Assert.True (recorder.LastOpWasFlush,
                 "AnsiOutput.Dispose returned without flushing pending cleanup sequences.");
}
```

`RecordingStream` is a tiny test fixture with a byte buffer plus `LastOpWasFlush` and `FlushCount` properties.

## Context

Surfaced in the v2.0.0 RC core review (`plans/v2-rc-core-review.md`, item #5).

tig:
## Summary

Fixes #5165.

`AnsiOutput.Dispose()` wrote its cleanup choreography (mouse-off, SGR reset, alt-buffer restore / inline cursor park, show cursor) but **did not flush** the underlying native stdout before returning. On the Windows VT and Unix raw-write paths, buffered cleanup bytes can be discarded when the process exits or when `WindowsVTOutputHelper.Dispose` closes the console handle, leaving the host terminal in a broken state (mouse tracking still on, SGR colors stuck, alt buffer not restored).

The constructor already calls `AnsiTerminalHelper.FlushNative` after writing init sequences for exactly this reason. The fix mirrors that call at the end of `Dispose`'s cleanup block so every shutdown sequence reaches the terminal before `Dispose` returns.

This PR is stacked on #5182 (which removed the `Degraded` early-return in `Dispose`, making cleanup writes observable to tests). Once #5182 merges, this PR can target `develop` directly with a clean diff.

## Changes

  - Add `AnsiTerminalHelper.FlushNative (_platform)` at the end of the `Dispose` cleanup block, after `CSI_ShowCursor` is written. Inside the same `try` so the existing `catch` swallows any platform-specific flush failure during shutdown. Mirrors the constructor's flush-after-init pattern.
  - `FlushNative` is a no-op when `_platform == AnsiPlatform.Degraded`, so headless test environments are unaffected.
- `Tests/UnitTestsParallelizable/Drivers/AnsiDriver/AnsiOutputDisposeFlushTests.cs` (new)
  - `Dispose_FullScreen_WritesAllCleanupSequences` — every `FullScreen` cleanup sequence (`CSI_DisableMouseEvents`, `CSI_ResetAttributes`, `CSI_RestoreCursorAndRestoreAltBufferWithBackscroll`, `CSI_ShowCursor`) is observable via `GetLastOutput()` after `Dispose`.
  - `Dispose_Inline_WritesAllCleanupSequences` — every `Inline` cleanup sequence including the cursor-park `CSI_SetCursorPosition (Y + Height - 1, 1)` and trailing newline is observable.
  - `Dispose_FullScreen_EmitsCleanupSequencesInExpectedOrder` — sequences appear in documented choreography order (mouse-off &lt; SGR reset &lt; alt-buffer restore &lt; show cursor).

The native flush itself is a no-op in `AnsiPlatform.Degraded` (the only platform in headless test environments), so the closest practical proxy is to assert that every cleanup sequence is observable in the captured output. The bug shows up at process exit on real terminals where the native handle's buffer is dropped before the bytes drain.

## Test plan

- [x] `dotnet build Tests/UnitTestsParallelizable` succeeds.
- [x] `dotnet test --project Tests/UnitTestsParallelizable -- --filter-method "*AnsiOutputDisposeFlushTests*"` — all 3 tests pass.
- [x] All 38 existing tests under `Tests/UnitTestsParallelizable/Drivers/AnsiDriver/` still pass.
- [x] On `origin/develop` (without #5182), `Dispose`'s `Degraded` early-return drops every cleanup write, so the visibility tests would fail there — the choreography is invisible until #5182 lands.

---

### DIFF
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
@@ -411,6 +411,14 @@ public class AnsiOutput : OutputBase, IOutput
             }
 
             Write (EscSeqUtils.CSI_ShowCursor);
+
+            // Flush the native stdout to ensure all cleanup sequences (mouse-off, SGR reset,
+            // alt-buffer restore, cursor show, etc.) reach the terminal before we return.
+            // Mirrors the FlushNative call in the constructor after init sequences are written.
+            // Without this, buffered cleanup bytes may be discarded when the process exits or
+            // when the underlying handle is closed by <see cref="WindowsVTOutputHelper.Dispose"/>.
+            // See issue #5165.
+            AnsiTerminalHelper.FlushNative (_platform);
         }
         catch
         {
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
@@ -359,15 +359,35 @@ public class AnsiOutput : OutputBase, IOutput
         }
     }
 
+    /// <summary>
+    ///     Computes the 1-indexed terminal row of the last row of the inline application region.
+    /// </summary>
+    /// <remarks>
+    ///     <para>
+    ///         <see cref="Rectangle.Height"/> is exclusive: <c>Y + Height</c> is the row immediately
+    ///         <em>after</em> the region. The last row inside the region is <c>Y + Height - 1</c>.
+    ///         ANSI cursor-positioning sequences are 1-indexed, but <see cref="Rectangle.Y"/> is already
+    ///         the 1-indexed terminal row offset for the inline region, so no additional adjustment is
+    ///         required.
+    ///     </para>
+    ///     <para>
+    ///         Used by <see cref="Dispose"/> to park the cursor on the final row of the inline region
+    ///         on shutdown. See issue #5166.
+    ///     </para>
+    /// </remarks>
+    /// <param name="appScreen">The inline application region rectangle.</param>
+    /// <returns>The 1-indexed terminal row of the last row of the region.</returns>
+    internal static int GetInlineLastRow (Rectangle appScreen) => appScreen.Y + appScreen.Height - 1;
+
     /// <inheritdoc/>
     public void Dispose ()
     {
         try
         {
-            if (_platform == AnsiPlatform.Degraded)
-            {
-                return;
-            }
+            // Note: each Write below short-circuits the platform-specific output path when
+            // <see cref="_platform"/> is <see cref="AnsiPlatform.Degraded"/>, but still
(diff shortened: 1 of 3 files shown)


tig:
Closing — this PR was stacked on #5182 (just closed), so it inherits the wrong cursor-coordinate baseline. The Copilot review also caught a real test gap I'd missed: the test uses `GetLastOutput()`, which is populated by `Write` *before* any native flush, and `FlushNative` is a no-op in `Degraded`-mode platforms (which is what the test runs on), so the assertion passes even without the new `FlushNative` call. The test wasn't actually exercising the fix.

The flush issue from #5165 is still real, but a proper fail-before/pass-after test for it needs a mockable platform layer that doesn't currently exist (the no-op-in-degraded-mode design means we can't observe the call from a unit test). I'd rather close this and re-spin a focused PR if/when someone adds an `IFlushSink` or similar abstraction.

Closing unmerged. Issue #5165 left open as the tracker.

---

PR #5232, opened 2026-05-07
 Fixes #5140. TableView: support CollectionNavigator = null to disable type-to-search

### LINKED ISSUE #5140: TableView should provide a supported way to disable typing-based row navigation
TableView uses CollectionNavigator to intercept printable keys and jump to the next matching row. That behavior is useful in many contexts, but it creates a problem for applications that want to use single-letter shortcuts while focus remains inside the table.

In those apps, unbound printable keys are consumed by TableView for incremental navigation instead of bubbling through normal key handling. That makes table-local shortcuts and application shortcuts harder to implement.

The current source comment for CollectionNavigator says “Set to null to disable this feature,” which suggests there is already an intended opt-out. But if that path is not actually safe/supported in practice, then the API contract is unclear.

**What would help:**
  1. A first-class, documented opt-out such as EnableTypeToSearch = false, or
  2. Making CollectionNavigator = null fully supported and null-safe, as the property comment implies.

Expected behavior when disabled:
  - Printable keys that are not otherwise bound should not be consumed for incremental row navigation.
  - Those keys should be available to normal KeyDown / command handling.
  - Accepted / Activating, cursor navigation, selection, and mouse behavior should continue to work normally.

**Why this matters:**
TableView is often used as the main interaction surface in terminal apps. In that role, developers may want keys like q, r, s, p, /, etc. to trigger app actions without forcing focus out of the table first. A documented opt-out would make TableView easier to integrate into keyboard-driven apps.

Bonus: the docs should be updated so the behavior and the supported disable mechanism are explicit.

References
 - #4027 seemed to refer to the same general issue for collection-navigation interception in ListView, but I don't think it resolved it for TableView.

harder:
## Summary

Implements **option 2** from #5140 (preferred by @tig, @tznind, and @BDisp): make `TableView.CollectionNavigator = null` fully supported and null-safe, matching what the property's existing XML doc already promised ("Set to null to disable this feature"). No new API surface.

When `CollectionNavigator` is `null`, printable keys that aren't otherwise bound bubble through normal key handling instead of being consumed for incremental row navigation — letting apps use single-letter shortcuts (`q`, `r`, `/`, …) while focus remains in the table.

## Changes

- `Terminal.Gui/Views/TableView/TableView.cs` — change `CollectionNavigator` to `ICollectionNavigator?` and expand the XML `<summary>` to document the disabled-state behavior.
- `Terminal.Gui/Views/TableView/TableView.Navigation.cs` — early-return `false` from `CycleToNextTableEntryBeginningWith` when `CollectionNavigator is null`. Returning false lets `OnKeyDownNotHandled` propagate the key instead of consuming it.
- `Tests/UnitTestsParallelizable/Views/TableViewTests.cs` — two new tests covering the null path (key not consumed, no NRE on hotkey).
- `Tests/UnitTestsParallelizable/Views/TableViewLegacyTests.cs` — single `!` to silence a new CS8602 in an existing test now that the property is nullable.

## What this does NOT change

- Default behavior is unchanged: `CollectionNavigator` is still constructed in `TableView`'s constructor, so type-to-search works out of the box.
- No new public API; no behavior change for apps that don't opt out.

## Test plan

- [x] `dotnet build` — clean, no new warnings
- [x] `dotnet test --project Tests/UnitTestsParallelizable` — 16,794 passed / 0 failed / 17 skipped
- [x] All 120 existing `TableView*` tests still pass
- [x] New tests verify: `CollectionNavigator = null` disables type-to-search and lets the key bubble;
`CollectionNavigator = null` + `HotKey` does not throw

### DIFF
--- a/Terminal.Gui/Views/TableView/TableView.Navigation.cs
+++ b/Terminal.Gui/Views/TableView/TableView.Navigation.cs
@@ -101,6 +101,11 @@ public partial class TableView
 
     private bool CycleToNextTableEntryBeginningWith (Key key)
     {
+        // Type-to-search disabled
+        if (CollectionNavigator is null)
+        {
+            return false;
+        }
         int row = _cursorRow;
 
         // There is a multi select going on and not just for the current row
--- a/Terminal.Gui/Views/TableView/TableView.cs
+++ b/Terminal.Gui/Views/TableView/TableView.cs
@@ -165,8 +165,12 @@ public partial class TableView : View, IValue<TableSelection?>, IDesignable
         }
     }
 
-    /// <summary>Navigator for cycling the selected item in the table by typing. Set to null to disable this feature.</summary>
-    public ICollectionNavigator CollectionNavigator { get; set; }
+    /// <summary>
+    ///     Navigator for cycling the selected row in the table by typing. Set to <see langword="null"/> to disable
+    ///     type-to-search navigation; printable keys not otherwise bound will then bubble through normal key handling
+    ///     instead of being consumed for incremental row navigation.
+    /// </summary>
+    public ICollectionNavigator? CollectionNavigator { get; set; }
 
     /// <summary>
     ///     The maximum number of characters to render in any given column.  This prevents one long column from pushing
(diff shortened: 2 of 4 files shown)


tznind:
Could add a menu option for it in UICatalog example project.

Would make it more discoverable as a feature and end user testable

tig (approved):
lgtm

BDisp (approved):

PR #5233, opened 2026-05-07
Fixes #5126. TableView: don't raise Activating when click is outside any cell

### LINKED ISSUE #5126: Clicking outside of a row in TableView triggers Activating
## Describe the bug

The Activating event fires on click outside of row area of TableView

Image

## To Reproduce

Create a Dim.Fill TableView with 1 or more rows
Click in screen outside of any rows

2. Expected behavior: (describe what should happen)
Nothing

3. Actual behavior: (describe what actually happens)
Activating fires

harder:
## Fixes

- Fixes #5126

`TableView` raised the `Activating` event on every left-click inside its bounds, even when the click landed in the header area, below the last data row, or in horizontal whitespace to the right of the last rendered column. Hit-testing was only performed inside `OnActivated`, by which time the event had already fired with no real cell selected, surprising subscribers.

Two things conspired to produce the bug:

1. `OnActivated` checked for a null `ScreenToCell` result and returned early - but `Activating` had already been raised before `OnActivated` was reached.
2. `ScreenToCell`'s column lookup uses `cellInfos.LastOrDefault(c => c.X <= clientX)` with no right-edge check, so far-right X positions map onto the last column. This is masked by default (`Style.ExpandLastColumn = true` makes the last column fill remaining viewport width), but when `ExpandLastColumn = false` there is visible whitespace to the right of the table and clicks there were still treated as cell hits.

## Proposed Changes/Todos

- [x] Override `OnActivating` in `TableView` and cancel (return `true`) when a left-click does not land on a cell
- [x] Extract shared mouse-hit logic into a `TryGetMouseCellHit` helper so `OnActivating` and `OnActivated` use the same check
- [x] Add `IsXWithinAnyRenderedColumn` and apply it inside `TryGetMouseCellHit` so clicks past the right edge of the last rendered column are also rejected. The check runs only on the activation path — `ScreenToCell`'s public API behavior is unchanged
- [x] Keep the existing keyboard activation path (Space → `ToggleExtend`) untouched - only mouse activations outside cells are cancelled
- [x] Add regression tests in `TableViewTests`:
- `Click_OutsideRows_DoesNotRaise_Activating` (click below last row)
- `Click_OnRow_Raises_Activating` (positive case — valid cell still fires)
- `Click_RightOfLastColumn_DoesNotRaise_Activating` (with
    `ExpandLastColumn = false` so real whitespace exists)
- `Click_OnHeader_DoesNotRaise_Activating`
- `Click_EmptyTable_DoesNotRaise_Activating` (`Table` is null)
- `Click_TableWithZeroRows_DoesNotRaise_Activating` (`Table` set, `Rows = 0`)
- [x] Verified all 16,798 tests in `UnitTestsParallelizable` still pass

### Key change in `TableView.cs`:

```csharp
/// <inheritdoc />
protected override bool OnActivating (CommandEventArgs args)
{
    // Cancel Activate when a mouse click lands outside any cell (e.g. below the
    // last row, in the header area, or right of the last rendered column).
    // Without this, the Activating event would fire even though no cell was
    // actually selected, surprising subscribers.
    if (!TryGetMouseCellHit (args.Context, out Point? hit))
    {
        return false;
    }

    if (hit is null)
    {
        return true;
    }

    return false;
}

private bool IsXWithinAnyRenderedColumn (int clientX)
{
    int viewportX = clientX + Viewport.X;

    foreach (ColumnToRender col in NonHiddenCellInfos ())
    {
        if (viewportX >= col.X && viewportX < col.X + col.Width)
        {
            return true;
        }
    }

    return false;
}
```

### To pull down this PR locally:

git fetch copilot fix-5126-tableview-activating-outside-rows
git checkout copilot/fix-5126-tableview-activating-outside-rows

### DIFF
--- a/Terminal.Gui/Views/TableView/TableView.cs
+++ b/Terminal.Gui/Views/TableView/TableView.cs
@@ -239,6 +239,25 @@ public partial class TableView : View, IValue<TableSelection?>, IDesignable
     }
 
     /// <inheritdoc />
+    protected override bool OnActivating (CommandEventArgs args)
+    {
+        // Cancel Activate when a mouse click lands outside any cell (e.g. below the
+        // last row or in the header area). Without this, the Activating event would
+        // fire even though no cell was actually selected, surprising subscribers.
+        if (!TryGetMouseCellHit (args.Context, out Point? hit))
+        {
+            return false;
+        }
+
+        if (hit is null)
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    /// <inheritdoc />
     protected override void OnActivated (ICommandContext? ctx)
     {
         if (ctx?.Binding is KeyBinding { Key: { } } keyBinding && keyBinding.Key == Key.Space)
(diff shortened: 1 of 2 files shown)


tig (approved):
lgtm. thanks!

PR #5254, opened 2026-05-08
Fixes #4892. OSC 8 hyperlink rendering does not clean up

harder:
Fixes #4892

## Summary

Two distinct paths leak OSC 8 hyperlink state into adjacent rendering, producing the visual artifact reported in #4892 (text near a `Link` rendered as if it were part of the hyperlink).

### 1. Stale entries in the cell URL map (`OutputBufferImpl.SetAttributeAndDirty`)

`SetAttributeAndDirty` previously only *wrote* to `_urlMap` when `CurrentUrl` was non-empty and never *cleared* existing entries. When a cell that previously carried a URL was overdrawn by content with no `CurrentUrl` set — for example a parent view filling its content area, or a `Link` whose displayed text shrank - the cell became dirty but the stale URL persisted in `_urlMap`. On the next render that cell was wrapped in OSC 8 sequences, so unrelated content rendered as a hyperlink.

Fix: when `CurrentUrl` is null/empty, remove the entry for that cell from `_urlMap`.

### 2. OSC 8 close skipped at row boundary (`OutputBase.Write`)

When dirty cells with a URL are flushed mid-row because a clean cell follows, `WriteToConsole` writes the OSC 8 *start* sequence + cell content to the terminal and clears `outputStringBuilder`, but `_lastUrl` stays set. If the row ends without another flush, the existing `if (outputStringBuilder.Length <= 0) { continue; }` early-`continue` skipped the end-of-row OSC 8 close. The hyperlink stayed open in the terminal across the row boundary, so subsequent rows rendered as part of the link.

Fix: at end of row, if `_lastUrl` is still set, emit the OSC 8 close even when nothing else is buffered. Legacy-console path is unchanged (no OSC 8 emitted).

## Tests

Three regression tests added in `Tests/UnitTestsParallelizable/Drivers/Output/OutputBaseTests.cs`:
- `Write_UrlFollowedByCleanCells_ClosesHyperlinkAtRowEnd` — covers fix #2
- `AddStr_NoCurrentUrl_ClearsStaleUrlMapping` — covers fix #1
- `AddStr_DifferentUrl_OverwritesUrlMapping` — guards the URL-clear path against breaking re-assignment when `CurrentUrl` changes

All 16,847 parallelizable tests and 73 non-parallelizable tests pass.

# To pull down this PR locally:
git remote add copilot https://github.com/harder/Terminal.Gui.git
git fetch copilot fix/4892-osc8-link-cleanup
git checkout copilot/fix/4892-osc8-link-cleanup

PR #5276, opened 2026-05-08
Fixes #5275. Clear stale TextView OSC 8 hyperlinks

### LINKED ISSUE #5275: TextView plain-text OSC 8 hyperlinks are not fully cleared after scroll or delete-all
## Describe the bug

`TextView` auto-detects plain-text URLs and emits OSC 8 hyperlinks for them, but hyperlink state is not always cleared correctly after redraws.

In terminals that support OSC 8 hover underlines:
 - after hovering a link and scrolling with the mouse wheel, the underline can appear on the wrong text and move with the mouse cursor instead of staying on the actual link
 - on Windows Terminal, after selecting all text and deleting it, the deleted links can still appear hoverable at their former screen locations

 ## To Reproduce

1. Open the UICatalog and go to the TextView > Editor scenario.
Select a link to make the underline appear, and then scroll up or down using the mousesheel. The underline moves with the mouse cursor, away from the actual link.

2.  Select all text > delete. The underlines for the deleted links still appear when I mouse over their former location, even though the text is gone.

Expected behavior:
 - Hover underline appears only on the actual URL text.
 - Scrolling does not leave hyperlink underline metadata on unrelated text.
 - After deleting all text, no hyperlink hover regions remain.

Additional context
   - The bug appears consistently in terminals with OSC 8 hover underline support.
   - I reproduced the scroll/hover behavior in both Windows Terminal and Warp on macOS.
   - I reproduced the delete-all stale-hover behavior in Windows Terminal.
   - This appears related to OSC 8 hyperlink cleanup, but it affects the plain-text URL auto-linking path in TextView, not just explicit Link view rendering.

harder:
## Summary

Fixes the stale OSC 8 hyperlink state in `TextView` when links come from plain-text URL auto-detection rather than explicit `Link` views.

- Fixes #5275

## Changes

- Split output-buffer URL tracking so explicit links and auto-detected plain-text URLs are synchronized independently.
- Reused `Osc8UrlLinker` URL-range detection to map auto-detected URLs into per-cell output state before rendering.
- Updated row rendering to emit OSC 8 close sequences when a row previously had hyperlinks but no longer does after redraw, scroll, or delete-all.
- Added regression coverage for overwriting auto-detected URLs with plain text and spaces.
- Preserved the explicit-link path fixed by #5254; this addresses a separate pre-existing bug in the plain-text auto-linking path.

## Testing

- `dotnet build --no-restore`
- `dotnet test --project Tests/UnitTestsParallelizable --no-build`

```bash
git fetch copilot harder/issue-5275-textview-osc8
git checkout copilot/harder/issue-5275-textview-osc8

### DIFF
--- a/Terminal.Gui/Drivers/AnsiHandling/Osc8UrlLinker.cs
+++ b/Terminal.Gui/Drivers/AnsiHandling/Osc8UrlLinker.cs
@@ -3,6 +3,8 @@ namespace Terminal.Gui.Drivers;
 
 internal static class Osc8UrlLinker
 {
+    internal readonly record struct UrlRange (int Start, int Length, string Url);
+
     internal readonly struct Options
     {
         internal readonly string [] _allowedSchemes;
@@ -28,6 +30,11 @@ internal static class Osc8UrlLinker
     internal static StringBuilder WrapOsc8 (StringBuilder input)
     {
         return WrapOsc8 (input, _defaultOptions);
+    }
+
+    internal static List<UrlRange> FindUrls (string input)
+    {
+        return FindUrls (input, _defaultOptions);
     }
 
     internal static StringBuilder WrapOsc8 (StringBuilder input, Options options)
@@ -101,6 +108,76 @@ internal static class Osc8UrlLinker
         }
 
         return result;
+    }
+
+    private static List<UrlRange> FindUrls (string input, Options options)
+    {
+        List<UrlRange> ranges = [];
+        ReadOnlySpan<char> span = input.AsSpan ();
+        ReadOnlySpan<char> delimiter = "://".AsSpan ();
+        int i = 0;
+
+        while (i < span.Length)
+        {
+            int rel = span.Slice (i).IndexOf (delimiter, StringComparison.Ordinal);
+            if (rel < 0)
+            {
+                break;
+            }
+
+            int delimAt = i + rel;
+            int schemeEnd = delimAt;
+            int schemeStart = schemeEnd - 1;
+
+            while (schemeStart >= 0 && char.IsLetter (span [schemeStart]))
+            {
+                schemeStart--;
+            }
+
+            schemeStart++;
+
+            if (schemeStart < 0 || schemeStart >= schemeEnd)
+            {
+                i = delimAt + delimiter.Length;
+                continue;
+            }
+
+            ReadOnlySpan<char> scheme = span.Slice (schemeStart, schemeEnd - schemeStart);
+            if (!IsAllowedScheme (scheme, options))
+            {
+                i = delimAt + delimiter.Length;
+                continue;
+            }
+
+            int urlStart = schemeStart;
+            int j = delimAt + delimiter.Length;
+
+            while (j < span.Length && !IsUrlTerminator (span [j]))
+            {
+                j++;
+            }
+
+            int urlEnd = TrimTrailingPunctuation (span, urlStart, j);
+            if (urlEnd <= (delimAt + delimiter.Length))
+            {
+                i = j;
+                continue;
+            }
+
+            string candidate = span.Slice (urlStart, urlEnd - urlStart).ToString ();
+
+            Uri? _;
+            if (options._validateWithUri && !IsValidUrl (candidate, options, out _))
+            {
+                i = j;
+                continue;
+            }
+
+            ranges.Add (new (urlStart, urlEnd - urlStart, candidate));
+            i = j;
+        }
+
+        return ranges;
     }
 
     private static int ParseEscapeSequence (string text, int start, int len)
--- a/Terminal.Gui/Drivers/Output/OutputBase.cs
+++ b/Terminal.Gui/Drivers/Output/OutputBase.cs
@@ -57,6 +57,7 @@ public abstract class OutputBase
 
     // Last URL used for tracking hyperlink state
     private string? _lastUrl = null;
+    private readonly HashSet<int> _rowsWithUrls = [];
 
     private readonly StringBuilder _lastOutputStringBuilder = new ();
     private bool _clearLastOutputPending;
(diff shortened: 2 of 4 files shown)


harder:
### Code review

Found 3 issues:

1. Auto-URL detection treats raw `char` offsets as column indices, which breaks for any multi-codepoint grapheme cluster (`.claude/rules/unicode-graphemes.md` says "Think in graphemes, not runes" and warns against treating char counts as terminal cells).

`SyncAutoUrlsForRowCore` builds `rowText` by appending each cell's `Grapheme` — that string can be 1 char (`"a"`), 2 chars (`"é"` = `'e'` + U+0301), or 8+ chars (ZWJ emoji like 👨‍👩‍👦). It then passes the concatenation to `FindUrls` and uses `range.Start` directly as a column index. For a row like `🚀 https://example.com`, the URL's char offset in `rowText` diverges from its column position, so `_autoUrlMap` tags the wrong cells and the link visibly drifts off the URL. Test coverage uses ASCII so this is invisible to the new tests.

https://github.com/gui-cs/Terminal.Gui/blob/ad1127de74a2d86f05b9d060d7f875f6a45aa4a7/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs#L478-L496

2. `_rowsWithUrls` (on `OutputBase`) is never cleared when the buffer is reset, so stale row indices survive across `ClearContents`/`SetSize`.

`OutputBufferImpl.ClearContentsCore` clears both URL maps, but `OutputBase._rowsWithUrls` lives in a different class and is not notified. After a resize or clear, the next render sees `rowHadUrlsPreviously = true` for stale rows; if the new content has no URLs there, a spurious `OSC_EndHyperlink()` is prepended at the start of those rows. Mostly cosmetic, but a real correctness drift between the two structures — consider exposing a `ClearUrlState()` on `OutputBase` that callers invoke from `ClearContents`/`SetSize`, or computing `rowHadUrlsPreviously` from the buffer's maps directly.

https://github.com/gui-cs/Terminal.Gui/blob/ad1127de74a2d86f05b9d060d7f875f6a45aa4a7/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs#L310-L315

3. The `GetCellUrl` fast-path is defeated for the lifetime of the buffer once any URL is ever set (a Copilot reviewer flagged the same anti-pattern in #5195).

`ClearContentsCore` calls `.Clear()` on both maps but does not null them, so after the first URL is ever assigned both `_explicitUrlMap` and `_autoUrlMap` stay non-null forever. The null fast-path in `GetCellUrl` then always misses, and the newly added `RowContainsUrls` adds another full-row scan per dirty row — roughly an extra 1920 lock acquisitions per frame on a 24×80 buffer with no URLs. Setting the maps to `null` in `ClearContentsCore` (and after `SyncAutoUrlsForRowCore` finds zero matches, where applicable) restores the optimization.

https://github.com/gui-cs/Terminal.Gui/blob/ad1127de74a2d86f05b9d060d7f875f6a45aa4a7/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs#L64-L82

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

harder:
I've extensively tested with multiple agents and models trying various angles to fix the two root issues observed in https://github.com/gui-cs/Terminal.Gui/issues/5275:
1. After hovering a link and scrolling with the mouse wheel, the underline can appear on the wrong text and move with the mouse cursor instead of staying on the actual link.
2. After selecting all text and deleting it, the deleted links can still appear hoverable at their former screen locations.

They did find and fix several small bugs related to OSC 8 handling that should make Terminal.Gui as compliant as possible.  But were not able to fix the root issue, and after testing with various terminals and AI agent analysis of the Windows.Terminal codebase, I believe the root cause is a Windows Terminal bug. 

Observations:
1. In Windows Terminal in Windows 11 on my Surface ARM-based laptop, I still see both issues above (1 and 2). 
2. In Warp terminal on MacOS, I see issue 1 above (have never seen issue 2). Claude found a known issue in Warp that explains issue 1, and Warp does it's own custom link regex parsing that prevents it from being affected by issue 2. 
3. In iTerm2 terminal on MacOS, I cannot reproduce either issue. iTerm2 is known for its reference-quality OSC8 support.

If anyone has other terminals they can test with, I'd appreciate it. 

I am going to file a bug report in the Microsoft Terminal project with repro details.  The other OSC8 related bug fixes in this PR can be reviewed and merged.

harder:
I filed a bug in the Windows Terminal project that is the root cause of this issue when running Terminal.Gui in Windows Terminal:

https://github.com/microsoft/terminal/issues/20219


This PR now contains 4 other related fixes that AI agents found while debugging and diagnosing that root cause, ,which I believe will improve our hyperlink handling and are worth reviewing and committing: 

1. Emit OSC 8 ; ; ST when a row stops carrying URLs (ad1127de7)
A protocol-correctness fix. Without it, a row that previously held a URL and was later overwritten with non-link text could leave cells whose _hyperlinkId was never cleared in the receiving terminal's buffer. The WT hover-cache bug makes this visible, but the underlying emission is wrong regardless — any spec-compliant terminal that inspects cell hyperlink metadata after the fact (logs, screen-readers, recorder tools, other emulators) is affected.

2. Grapheme-to-column mapping in SyncAutoUrlsForRowCore (f3dcdd94c)
Multi-codepoint graphemes (ZWJ emoji, base + combining marks) before a URL were shifting the auto-link interval onto the wrong cells.

3. Fast-path restoration in GetCellUrl / RowContainsUrls (f3dcdd94c)
_explicitUrlMap and _autoUrlMap were never nulled after ClearContentsCore or when they became empty, so the "no URLs anywhere" fast-path was dead. Every cell of a URL-free buffer was acquiring the contents lock. A perf fix.

4. _rowsWithUrls invalidation on buffer/dim/version change + Add/Remove placement (f3dcdd94c + 7d30ac718)
Two related bookkeeping bugs in OutputBase:
- After a resize/clear/buffer-swap, stale row indices in _rowsWithUrls triggered spurious OSC 8 ; ; ST emissions at the start of rows that no longer existed or no longer had URLs. Now tracked via a UrlStateVersion counter and buffer-identity check.
- The Add/Remove on _rowsWithUrls was placed after an early-exit that fired when the per-row builder had already been flushed mid-loop via WriteToConsole. Rows that took that path skipped the bookkeeping update, leaving stale entries that compounded over frames. Moved before the early-exit; regression test added.

tig (approved):
nice work.

PR #5277, opened 2026-05-08
Fixes #5144. Add bracketed-paste mode support

### LINKED ISSUE #5144: Bracketed Paste Mode support
If someone pastes between terminals or pastes a lot of text - it would be hand to know.

For performance and custom paste handling

For example with this it should be possible to copy paste TGD views

- without having to use non standard hotkey
- between instances in seperate terminal windows/tabs

Terminals support enabling this with a feature called Bracketed Paste Mode.

Enable it with esc sequence and pasted input will be bounded by special input codes.

```
Console.Write("\x1b[?2004h"); // enable

  // In your input read loop:
  // \x1b[200~ = paste start → buffer following chars
  // \x1b[201~ = paste end   → handle buffered string as paste

  Console.Write("\x1b[?2004l"); // disable on exit
```

This should be opt in and have nice public api like a OnPaste event.

The input processor would detect start and end of bracketted paste and send to event instead of input stream.

harder:
- Fixes #5144

## Summary

Adds support for the ANSI bracketed-paste mode (DECSET 2004) so apps can distinguish a clipboard paste from typed input. Pastes arrive as a single event instead of N synthetic keystrokes.

The driver enables `ESC[?2004h` at startup and disables it on shutdown / suspend / alt-screen exit. The existing `AnsiResponseParser` gains a new `InBracketedPaste` state that detects `ESC[200~`, accumulates raw bytes (including escapes and control chars) until `ESC[201~`, and raises a `Paste` event with the stripped payload. The event flows: parser → `InputProcessor` → `IDriver.Paste` → `IApplication.Paste` (via cancellable `PasteEventArgs`) → focused `View.NewPasteEvent` → `View.OnPasted` → `View.Pasted` event → bubble to `SuperView`.

`TextField.OnPasted` and `TextView.OnPasted` are wired with sensible defaults that match Windows Terminal's `FilterStringForPaste`: `TextField` takes the first line and strips C0/C1 control characters; `TextView` normalizes `\r` and `\r\n` to `\n` and strips control characters except tab and newline. Apps that want raw payloads can subscribe to `Application.Paste` directly — sanitization lives in the widget defaults, not the event.

A 5-second idle flush in `InputProcessorImpl.ProcessQueue` recovers from terminals that drop the end marker. A 1 MiB hard cap protects against runaway buffers.

On terminals that don't support 2004, the markers never arrive and pastes flow through the normal key pipeline as before — no probing, no fallback configuration needed.

## Design

The behavior was cross-checked against other libraries and terminals to get best practices, including Bubble Tea (`charmbracelet/x/input`), Windows Terminal (`Utils::FilterStringForPaste`), xterm.js, Alacritty, kitty, and the xterm spec. 

Findings:
- Application-side libraries trust the terminal to sanitize and deliver raw bytes — same model here at the `Application.Paste` layer.
- Terminals (WT, Alacritty, xterm.js) sanitize before bracketing — defense-in-depth at the widget layer mirrors that.
- Newline normalization is required because terminals send `\r` for paste line breaks (some send `\r\n`); xterm.js, WT, and `TextView` here all normalize. - `TextField`'s "first line only" matches the existing clipboard `Paste()` command on the same control.

See `docfx/docs/bracketed-paste.md` for the full user-facing model and security notes.

## What's in the PR

- Parser: `AnsiResponseParserBase` state machine + `Paste` event + `FlushStaleBracketedPaste`
- Driver: enable/disable wiring in `AnsiOutput`, `NetInput`, `NetOutput.Suspend`, `UnixTerminalHelper.Suspend`
- API: `IDriver.Paste`, `IInputProcessor.Paste`, `IApplication.Paste`, `PasteEventArgs`, `View.NewPasteEvent` / `OnPasted` / `Pasted`
- Defaults: `TextField.OnPasted`, `TextView.OnPasted`
- Tests: 26 new tests in `UnitTests.Parallelizable` covering parser state machine (start/end split at every byte boundary, back-to-back pastes, nested literal start markers, mid-paste reset, oversize truncation, stale flush) and view dispatch (CWP order, handled short-circuit, bubbling, disabled views, `TextField`/`TextView` sanitization)
- UICatalog: `BracketedPasteDemo` scenario in the Input category
- Docs: `docfx/docs/bracketed-paste.md` linked from the TOC

## Verification

- `dotnet build` clean — 0 warnings, 0 errors
- `dotnet test --project Tests/UnitTestsParallelizable` — 16,871 passed, 17 skipped, 0 failed

# To pull down this PR locally:
git fetch copilot feat/5144-bracketed-paste
git checkout copilot/feat/5144-bracketed-paste

### DIFF
--- a/Examples/UICatalog/Scenarios/BracketedPasteDemo.cs
+++ b/Examples/UICatalog/Scenarios/BracketedPasteDemo.cs
@@ -25,7 +25,7 @@ public sealed class BracketedPasteDemo : Scenario
             X = 0,
             Y = Pos.Bottom (hint) + 1,
             Width = Dim.Fill (),
-            Title = "Paste into the focused TextField — the default OnPasted inserts the text."
+            Title = "Paste into the focused TextField — the default Command.Paste handler inserts the text."
         };
 
         Label log = new ()
--- a/Examples/UICatalog/Scenarios/BracketedPasteDemo.cs
+++ b/Examples/UICatalog/Scenarios/BracketedPasteDemo.cs
@@ -1,0 +1,71 @@
+#nullable enable
+
+namespace UICatalog.Scenarios;
+
+[ScenarioMetadata ("Bracketed Paste", "Logs bracketed-paste payloads delivered by the terminal")]
+[ScenarioCategory ("Input")]
+public sealed class BracketedPasteDemo : Scenario
+{
+    public override void Main ()
+    {
+        ConfigurationManager.Enable (ConfigLocations.All);
+        using IApplication app = Application.Create ();
+        app.Init ();
+
+        using Window appWindow = new ()
+        {
+            Title = GetQuitKeyAndName ()
+        };
+        appWindow.AssignHotKeys = true;
+
+        Label hint = new ()
+        {
+            X = 0,
+            Y = 0,
+            Text = "Paste anything from your system clipboard. Pastes show up as a single delivery if "
+                   + "your terminal supports bracketed-paste mode (most modern terminals do)."
+        };
+
+        TextField field = new ()
+        {
+            X = 0,
+            Y = Pos.Bottom (hint) + 1,
+            Width = Dim.Fill (),
+            Title = "Paste into the focused TextField — the default OnPasted inserts the text."
+        };
+
+        Label log = new ()
+        {
+            X = 0,
+            Y = Pos.Bottom (field) + 1,
+            Width = Dim.Fill (),
+            Height = Dim.Fill (),
+            Text = "Application.Paste log:\n"
+        };
+
+        var counter = 0;
+
+        app.Paste += (_, args) =>
+                     {
+                         counter++;
+                         log.Text += $"[{counter}] {args.Text.Length} chars: {Truncate (args.Text)}\n";
+                     };
+
+        appWindow.Add (hint, field, log);
+
+        app.Run (appWindow);
+    }
+
+    private static string Truncate (string text)
+    {
+        // Replace control chars so the display stays one line per paste.
+        string flattened = text.Replace ("\r", "\\r").Replace ("\n", "\\n").Replace ("\t", "\\t");
+
+        if (flattened.Length <= 60)
+        {
+            return flattened;
+        }
+
+        return flattened [..60] + "…";
+    }
+}
--- a/Examples/UICatalog/Scenarios/BracketedPasteDemo.cs
+++ b/Examples/UICatalog/Scenarios/BracketedPasteDemo.cs
@@ -3,7 +3,7 @@
 namespace UICatalog.Scenarios;
 
 [ScenarioMetadata ("Bracketed Paste", "Logs bracketed-paste payloads delivered by the terminal")]
-[ScenarioCategory ("Input")]
+[ScenarioCategory ("Text and Formatting")]
 public sealed class BracketedPasteDemo : Scenario
 {
     public override void Main ()
@@ -18,13 +18,7 @@ public sealed class BracketedPasteDemo : Scenario
         };
         appWindow.AssignHotKeys = true;
 
-        Label hint = new ()
-        {
-            X = 0,
-            Y = 0,
-            Text = "Paste anything from your system clipboard. Pastes show up as a single delivery if "
-                   + "your terminal supports bracketed-paste mode (most modern terminals do)."
-        };
+        Label hint = CreateHintLabel ();
 
         TextField field = new ()
         {
@@ -40,20 +34,43 @@ public sealed class BracketedPasteDemo : Scenario
             Y = Pos.Bottom (field) + 1,
             Width = Dim.Fill (),
             Height = Dim.Fill (),
-            Text = "Application.Paste log:\n"
+            Text = "Application.Paste log (only bracketed paste events appear here):\n"
         };
 
-        var counter = 0;
+        int counter = 0;
 
         app.Paste += (_, args) =>
                      {
-                         counter++;
-                         log.Text += $"[{counter}] {args.Text.Length} chars: {Truncate (args.Text)}\n";
-                     };
+                          counter++;
+                          log.Text += $"{FormatPasteLogEntry (counter, args.Text)}\n";
+                      };
 
         appWindow.Add (hint, field, log);
 
         app.Run (appWindow);
+    }
+
+    public static Label CreateHintLabel ()
+    {
+        Label hint = new ()
+        {
+            X = 0,
+            Y = 0,
+            Width = Dim.Fill (),
+            Height = Dim.Auto (DimAutoStyle.Text),
+            Text = "Paste into the TextField below.\n"
+                   + "If Terminal.Gui detects bracketed paste, the paste is logged below as a single bracketed-paste event.\n"
+                   + "If text appears in the field but no new log entry is added, your terminal delivered it as normal input instead."
+        };
+
+        hint.TextFormatter.WordWrap = true;
+
+        return hint;
+    }
+
+    public static string FormatPasteLogEntry (int counter, string text)
+    {
+        return $"[{counter}] Bracketed paste event: {text.Length} chars: {Truncate (text)}";
     }
 
     private static string Truncate (string text)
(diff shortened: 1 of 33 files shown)


harder:
Keeping this in WIP/Draft mode for now so I can do more testing and code reviews, since it's a larger sized change.

harder:
I did a full code review with a different agent which found the issues below and addressed them in my most recent commit. Putting the description here for reference. I still need to do more manual testing and investigation.  (original commits done using Claude Code Opus 4.7,  code review and revised commits done with CoPilot CLI using Codex 5.4). 

-----

Confirmed issues
   1. Bug: stale bracketed-paste timeout is measured from paste start, not last byte received. A slow but valid paste can be flushed early after 5 seconds.
   2. Bug: oversized paste handling appears to flush at the 1 MiB cap and then return to normal parsing, which can let the remainder of the same paste leak into the normal input pipeline.
   3. Current failing CI: ViewBaseTests.ViewPasteTests.TextView_OnPasted_NormalizesCarriageReturnToLineFeed fails on Windows. Expected a\nb\nc\nd, actual a\r\nb\r\nc\r\nd.

Important follow-up concerns
   - Release () mid-paste silently discards buffered content, while stale flush emits a partial paste event. That inconsistency should be resolved or documented.
   - Raising Paste from inside parser state handling is a design risk; dispatching outside parser-internal state handling would be safer.
   - Literal \x1b[201~ inside pasted content is a protocol hazard and is not well covered by tests or docs.

What looks good
   - Enable/disable lifecycle for bracketed paste is consistent with common terminal/TUI practice.
   - Raw paste payload at the parser/application layer is the right model.
   - Widget-level sanitization in TextField and TextView is a sensible defense-in-depth approach.
   - No-probe fallback behavior is good: unsupported terminals naturally continue through the normal key pipeline.

Recommendation before merge
   1. Change stale flush to use last paste byte time, not StateChangedAt.
   2. On size-cap overflow, enter a discard-until-end-marker mode instead of returning directly to normal parsing.
   3. Resolve the Windows newline contract by fixing either the implementation, the test, or the documented behavior.
   4. Add tests for slow long-running paste, overflow with trailing bytes, and literal \x1b[201~ in payload.

 ---

This fixes the bracketed-paste issues from the review:
   - stale paste timeout now tracks the last paste byte, not paste start
   - oversized pastes now enter a discard remainder state so tail bytes do not leak into normal input
   - the end-marker split across the 1 MiB boundary is handled correctly
   - the Windows TextView newline test now matches the actual platform-facing Text contract
   - docs and regression coverage were updated, including idle-timeout, oversize, boundary-straddle, and discard-state cases

harder:
This is ready for review now. There's a new UICatalog scenario under Text and Formatting that demonstrates where you can test the bracketed paste. Let me know if you want that expanded/improved.

tig (commented):
I am curious about two things I see here:'

- no default paste sanitizer for ViewBase.
- no linkage to Command.Paste - speciiclaly, does Command.Paste have any need to bubble? Your calling into the SupreView in RaisePasteEvent implies yes, but you are not using the command bubbling mechansims.

I may be overthinking but would like to understand your thining on these lines. thanks.

harder:
> no default paste sanitizer for ViewBase

Good question. We could go either way. The approach I have here is for ViewBase to be agnostic as to the type of content pasted in. The payload is the exact terminal-delivered text after the bracket markers are stripped, and didn't want base View to automatically mutate that for every custom view. Global sanitization could be there, but then it would silently impose a text policy. Instead in this approach, `TextField.OnPasted` takes the first line and strips controls, and `TextView.OnPasted` normalizes CR/CRLF and strips controls except tab/newline. So it's "raw payload in the transport layer, sanitize at the text sink." 

On the flip side, doing sanitization in ViewBase could reduce duplicate code, and give a safer default if a custom control author forgets to think about paste safety. I'd guess most real paste content for a terminal library is going to be textual, so a base default isn't crazy. The only con would be if some custom view control wanted the exact terminal-delivered payload on paste. 

Thoughts?

harder:
Maybe a safer bet is to do a more basic/minimal sanitizer in ViewBase by default (not the full TextView semantics like normalizing line endings) and give an easy way to override with clear documentation. TextField and TextView could show that example by overriding with the implementation in this PR. 

If we wanted to go that route, we could either pass both raw and mutated payload in the PasteEventArgs, or have an explicit sanitizer hook (like `string SanizitizePastedText(string text)`) to override.

harder:
>no linkage to Command.Paste - speciiclaly, does Command.Paste have any need to bubble? Your calling into the SupreView in RaisePasteEvent implies yes, but you are not using the command bubbling mechansims.

Short answer is that yes, changing it to route through Command.Paste and everything use the bubbled contract instead of NewPasteEvent would make the architecture cleaner - if we redefined `Command.Paste` as the canonical paste abstraction and refactor things that touch it. That's definitely a bigger change.. but maybe the right one! 

This is all based on a core assumption I'm making that bracketed paste carries the exact payload the terminal already delivered, while`Command.Paste` means "go read from the clipboard and paste that." So unifying them requires redefining `Command.Paste` to be payload-aware instead of clipboard-only.

In the current implementation, bracketed paste is not routed through the command pipeline because `TextField.Paste` and `TextView.Paste` both read from App.Clipboard, they don't accept a payload argument. So with Bracketed paste, the terminal has already delivered the payload and wanted to preserve that exact string all the way through `Application.Paste` and `View.NewPasteEvent`. If we reuse Command.Paste as-is, the handlers would re-read the clipboard and lose the terminal-delivered payload. 

We could refactor so that Command.Paste would mean something more like "perform a paste operation, optionally using a provided payload" and modifying the layers / bubbling semantics to account for this (both a payload-bearing paste and the existing meaning to read from clipboard). Or I could also be missing something obvious and simpler, too. :)

BDisp:
In my opinion, I think there should be a virtual method that, by default, provides the actual pasting, treating the new line with the predefined new line value, as well as an event that is invoked with the same value. If a derived View wants to customize its pasting, it would have to override the virtual method and have the possibility to change the pasting value in the event before it is invoked. However, I agree with your inferences cited in the comments above.
By the way, the method has to return a Boolean to indicate whether the derived View consumed it or not.

harder:
I updated this PR to finish the bracketed-paste work by moving paste onto a single shared pipeline. It now has paste aligned with the rest of the T.Gui input model instead of adding a second event-routing system for bracketed paste.

Previously bracketed paste had its own parallel dispatch path:
- Application.RaisePasteEvent walked into the focused view through a dedicated paste event path
- View.NewPasteEvent called OnPasted, raised a cancellable Pasted event, and then manually recursed to SuperView
- TextField and TextView did their paste sanitization inside OnPasted
- Command.Paste / Ctrl+V was still a separate clipboard-only path

That meant terminal bracketed paste and keyboard/menu paste were similar in intent, but not actually using the same routing, sanitization, or event model.

Now, bracketed paste and Ctrl+V both go through the existing command pipeline:
- Application.RaisePasteEvent (text) raises Application.Paste with the raw payload
- if not handled there, the focused view receives Command.Paste
- a default Command.Paste handler on View now owns the common flow:
1. resolve the paste source
2. sanitize it
3. raise cancellable Pasting
4. call OnPaste
5. raise Pasted only if text was actually inserted

TextField and TextView now only override the two view-specific hooks they actually need:
- OnSanitizingPaste
- OnPaste

That removes the custom bracketed-paste event pipeline and makes terminal paste, Ctrl+V, and other paste entry points behave consistently.

Additional fixes included while finishing the refactor:
- introduced a typed PastePayload so bracketed-paste payloads cannot be confused with unrelated string-valued command context data
- moved parser-triggered paste callbacks so they fire outside the parser lock
- kept read-only text inputs consuming paste commands without incorrectly raising Pasted
- fixed TextView so payload-driven paste does not accidentally reuse clipboard-only “copy without selection” behavior
- fixed TextField so rewritten, filtered, empty-result, and identical-replacement pastes all leave cursor/selection state correct and 
report the right Pasted text
- aligned TextField sanitization with what TextField can actually insert, so Pasting, insertion, and Pasted all agree on the final text

tig (approved):

PR #5279, opened 2026-05-08
Fixes #5127. Anchor button delimiters to edges; fix highlight continuity

### LINKED ISSUE #5127: When a button has a fixed size then the delimiters should render at edges
## Describe the bug

When editting a Button and setting the Width to a fixed value e.g. 37

Then the `[` and `]` symbols should correctly be at the edges of the View

Instead the button just ignores the extra space and renders centred

Image

YourRobotOverlord:
<details><summary>Copilot Session</summary><p>a98f7d36-f362-457b-8f86-885ba3e011bc</p></details>

Fixes #5127.

## Problem

When a `Button` has a fixed `Width` wider than its natural size:
- The `[` and `]` delimiters were centered with the text instead of pinned to the view edges.
- When focused, the highlight attribute had gaps — cells before/after the centered text used the default attribute rather than `Focus`.

## Solution

Override `OnDrawingText` to take direct control of rendering:
- Draw `[` explicitly at `drawRect.X` (left edge) and `]` at `drawRect.X + Width - 1` (right edge).
- Call `Driver.FillRect(drawRect)` with the `normalAttr` (which resolves to `Focus` when focused) before drawing text, ensuring continuous highlight across the entire button area including all rows.
- Center only the interior text (between the brackets) using a dedicated `_interiorTextFormatter`.
- `UpdateTextFormatterText` still sets `TextFormatter.Text` to the full decorated string so `Dim.Auto` sizing continues to work correctly.

## Tests

Two new tests in `ButtonDrawingTests`:
- `FixedWidth_Anchors_Delimiters_To_Edges` — verifies `[` and `]` appear at columns 0 and 9 for a `Width=10` button, for both `IsDefault=false` and `IsDefault=true`.
- `Focused_FixedWidth_Button_Highlight_Is_Continuous` — verifies all 10 cells carry `Focus` or `HotFocus` attribute with no gaps.

### DIFF
--- a/Terminal.Gui/Views/Button.cs
+++ b/Terminal.Gui/Views/Button.cs
@@ -241,13 +241,62 @@ public class Button : View, IDesignable, IAcceptTarget
         Attribute hotAttr = HasFocus ? GetAttributeForRole (VisualRole.HotFocus) : GetAttributeForRole (VisualRole.HotNormal);
         string interiorText = GetInteriorText ();
 
-        _interiorTextFormatter.Text = interiorText;
-        _interiorTextFormatter.Alignment = TextAlignment;
-        _interiorTextFormatter.VerticalAlignment = VerticalTextAlignment;
-        _interiorTextFormatter.Direction = TextDirection;
-        _interiorTextFormatter.PreserveTrailingSpaces = PreserveTrailingSpaces;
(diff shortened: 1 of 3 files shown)


tig (approved):
lgtm

PR #5353, opened 2026-05-20
Fixes #5349. Ignore non-readable poll events during ANSI shutdown

### LINKED ISSUE #5349: ANSI app can hang in IApplication.Dispose() after RunAsync returns
This showed up while running an ANSI-driver integration smoke test in my SkillView tool while testing after upgrading to v2.2.1. However it looks like there is the same repro on 2.2.0-rc.3, hanging in the same place.

## Details
- Repo/version: gui-cs/Terminal.Gui v2.2.1
- Also reproduces on: v2.2.0-rc.3
- OS: macOS  26.4.1 arm64
- SDK/runtime: .NET SDK 10.0.201
- Driver: ANSI
- Expected: process exits after RunAsync and Dispose()
- Actual: RunAsync returns, window.Dispose() returns, app.Dispose() hangs

## Repro

```
 using Terminal.Gui.App;
 using Terminal.Gui.Views;
 
 internal static class Program
 {
     private static async Task<int> Main()
     {
         var app = Application.Create();
         app.Init("ansi");
         app.StopAfterFirstIteration = true;
 
         var window = new Window { Title = "TG repro" };
 
         Console.WriteLine("before RunAsync");
         await app.RunAsync(window, CancellationToken.None, null);
         Console.WriteLine("after RunAsync");
 
         Console.WriteLine("before window.Dispose");
         window.Dispose();
         Console.WriteLine("after window.Dispose");
 
         Console.WriteLine("before app.Dispose");
         app.Dispose();
         Console.WriteLine("after app.Dispose");
 
         return 0;
     }
 }
```

STEPS:
```
 dotnet new console -n TgDisposeRepro -f net10.0
 cd TgDisposeRepro
 dotnet add package Terminal.Gui --version 2.2.1
 # replace Program.cs with the repro above
 dotnet run
```

### Observed output

 before RunAsync
 after RunAsync
 before window.Dispose
 after window.Dispose
 before app.Dispose
_hangs here_

harder:
## Fixes

- Fixes #5349

## Proposed Changes
- Change `UnixIOHelper.IsInputAvailable ()` to return `true` only when `poll()` reports `POLLIN`
- Add `UnixIOHelperTests` to cover both a readable poll event and a non-readable poll event on Unix-like platforms
- Keep the fix at the shared helper so all ANSI input paths stop treating `POLLHUP`/`POLLERR`/`POLLNVAL`-only wakeups as readable input

## Root Cause

`AnsiInput.FlushInput ()` uses `UnixIOHelper.IsInputAvailable ()` to decide whether to keep draining input during shutdown.

On Unix/macOS, `poll()` can wake up with non-readable events such as `POLLHUP` or `POLLNVAL` without `POLLIN`. The old helper treated any `poll() > 0` result as readable input,   so shutdown could attempt another read even though no readable data was available. If that happened while `MainLoopCoordinator.Stop ()` was waiting for the input task to exit, `app.Dispose ()` could hang.

## OSS Comparison

This fix matches the defensive pattern used by other mature terminal codebases:
- `prompt_toolkit` fixed the same class of closed/unpollable-stdin bugs by treating EOF/unpollable stdin as terminal closure and unregistering the reader
- `kitty` currently has an active bug in the same family where non-readable poll events are effectively treated as readable
- `lnav`, `vim`, `zsh` ZLE, and `arcan` all gate reads on `POLLIN` and handle hangup/error events separately

That makes this change consistent with established terminal I/O practice: only read when `POLLIN` is set.

## Verification
- `dotnet build --no-restore`
- `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter-class "*UnixIOHelperTests"`
- `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter-class "*AllDriverTests"`
- `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter-class "*RunTests"`

No new warnings were introduced in the touched files. The build still reports an existing unrelated `IntegrationTests` analyzer warning (`AD0001`).

```bash
git fetch copilot fix-5349-ansi-dispose-hang
git checkout copilot/fix-5349-ansi-dispose-hang
```

### DIFF
--- a/Terminal.Gui/Drivers/UnixHelpers/UnixIOHelper.cs
+++ b/Terminal.Gui/Drivers/UnixHelpers/UnixIOHelper.cs
@@ -236,7 +236,20 @@ internal static class UnixIOHelper
         {
             int n = poll (pollMap, (uint)pollMap.Length, timeoutMs);
 
-            return n > 0;
+            if (n <= 0)
+            {
+                return false;
+            }
+
+            foreach (Pollfd pollfd in pollMap)
+            {
+                if ((pollfd.revents & (short)Condition.PollIn) != 0)
+                {
+                    return true;
+                }
+            }
+
+            return false;
         }
         catch
         {
--- a/Tests/UnitTestsParallelizable/Drivers/UnixIOHelperTests.cs
+++ b/Tests/UnitTestsParallelizable/Drivers/UnixIOHelperTests.cs
@@ -1,0 +1,85 @@
+using System.Runtime.InteropServices;
+using Terminal.Gui.Drivers;
+
+namespace DriverTests;
+
+/// <summary>
+///     Tests for <see cref="UnixIOHelper"/> polling semantics on Unix-like platforms.
+/// </summary>
+[Collection ("Driver Tests")]
+public class UnixIOHelperTests
+{
+    [DllImport ("libc", SetLastError = true)]
+    private static extern int pipe (int [] pipefd);
+
+    [DllImport ("libc", SetLastError = true)]
+    private static extern int close (int fd);
+
+    [Fact]
+    // Copilot
+    public void IsInputAvailable_ReturnsTrue_WhenPollReportsReadableData ()
+    {
+        if (!OperatingSystem.IsLinux () && !OperatingSystem.IsMacOS () && !OperatingSystem.IsFreeBSD ())
+        {
+            return;
+        }
+
+        int [] pipeFds = new int [2];
+        Assert.Equal (0, pipe (pipeFds));
+
+        try
+        {
+            byte [] payload = [0x41];
+            Assert.Equal (1, UnixIOHelper.write (pipeFds [1], payload, payload.Length));
+
+            UnixIOHelper.Pollfd [] pollMap =
+            [
+                new ()
+                {
+                    fd = pipeFds [0],
+                    events = (short)UnixIOHelper.Condition.PollIn
+                }
+            ];
+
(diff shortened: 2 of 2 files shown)


tig (approved):
The fix is clean and well-targeted. Here's my review:

## Summary

**Good PR.** The fix correctly addresses a hang during ANSI shutdown by checking `revents` for `POLLIN` instead of treating any poll wake-up (e.g., `POLLHUP`, `POLLERR`, `POLLNVAL`) as readable data. Previously, `poll()` returning > 0 was enough to signal "input available," which could cause a read attempt on a closed/errored fd — leading to a hang.

## Findings

The logic change is correct and the tests exercise both the happy path and the error-event path using real pipes + P/Invoke. Two minor observations:

1. **Test: double-close risk** — In `IsInputAvailable_ReturnsFalse_WhenPollReportsNonReadableEvent`, `pipeFds[0]` is closed inside the `try` block (the test's setup step), but the `finally` block doesn't guard against closing it again. Currently fine because `finally` only closes `pipeFds[1]`, but if someone refactors this, it's fragile. Minor — not blocking.

2. **Test: platform skip pattern** — The early-return skip pattern works but isn't self-documenting in test output (xUnit won't distinguish "skipped" from "passed"). A `Skip` attribute or `Assert.Skip()` would make CI results clearer. Minor — not blocking.

No bugs, no security issues, no logic errors. The change is minimal, correct, and well-tested.

Issue #4973, opened 2026-04-16
Draw cascade and layout propagation bugs cause O(N) full redraws on overlapping siblings


tig:
## Summary

During the draw pass, `ClearViewport()` calls `SetNeedsDraw()` which recursively propagates draw requests to **ALL overlapping subviews**. Combined with `DoDrawAdornments()` unconditionally escalating `NeedsDrawRect` to the full `Viewport`, this means any view with overlapping children (e.g., `Tabs` with 9 tab content views) forces **every** overlapping sibling to fully redraw on every frame — even when only one subview changed.

Additionally, `SetNeedsLayout()` cascades unnecessarily to all sibling subtrees, and `Layout()` unconditionally marks views for redraw even when nothing visual changed. These layout bugs compound the draw issue.

This was identified while investigating scroll performance in the `ConfigurationEditor` scenario vs the `Editor` scenario (PR #4723, related to #4522).

### The propagation chain

1. A `TextView` inside Tab 0 scrolls → `SetViewport()` → `SetNeedsDraw()` propagates up: `Tab0.SubViewNeedsDraw → Tabs.SubViewNeedsDraw → Window.SubViewNeedsDraw`
2. `Window.Draw()` enters the block (`NeedsDraw || SubViewNeedsDraw`)
3. **`DoDrawAdornments()`** (`View.Drawing.Adornments.cs:139-141`) escalates `NeedsDrawRect = Viewport` even though only `SubViewNeedsDraw` was true — this makes `NeedsDraw = true`
4. **`DoClearViewport()`** → **`ClearViewport()`** clears the entire viewport, then calls `SetNeedsDraw()`
5. `SetNeedsDraw(rect)` (`View.NeedsDraw.cs:123-133`) iterates ALL subviews that intersect the rect and recursively calls `SetNeedsDraw` on each
6. For a `Tabs` control with 9 overlapping tab views (all using `Dim.Fill`), ALL 9 tabs get `NeedsDraw = true`
7. ALL 9 tabs fully draw, even though only Tab 0 changed

### The key code paths

**`View.Drawing.cs` — `ClearViewport()` line ~290:**
```csharp
SetNeedsDraw ();  // Propagates to ALL overlapping subviews recursively
```

**`View.Drawing.Adornments.cs` — `DoDrawAdornments()` lines 139-141:**
```csharp
if (NeedsDrawRect == Rectangle.Empty)
{
[...]

PR #5364, opened 2026-05-21
Fixes #5356. Add tab fan-out layout/draw diagnostic tests

### LINKED ISSUE #5356: Add regression diagnostics for tabbed redraw/layout fan-out
## Summary

Add focused diagnostics/tests that prove the current tabbed redraw/layout fan-out problem and keep it from silently regressing again.

This issue is intentionally instrumentation-only. It should not change rendering or invalidation behavior.

Split from #4973.

## Problem

Issue #4973 and the follow-up discussion identified that scrolling in the active tab can still fan out into layout and draw work outside the active tab. Before making behavioral fixes, we need reliable diagnostics that capture the problem in a way that is safe for future refactoring.

The diagnostics must reflect the non-obvious rendering nuances called out in the thread:
- clipping, especially at the `IOutput` level
- `ViewportSettings.Transparent`
- margin shadows
- adornment subviews
- `Driver.Contents` is not equivalent to actual `IOutput` output

## Scope

- Add targeted tests and/or diagnostic helpers for tabbed redraw/layout fan-out.
- Measure active-tab vs inactive-tab layout/draw activity during scrolling.
- Add a reproducible comparison between a single-TextView scenario and a tabbed-TextView scenario.
- Keep this issue instrumentation-only. Do not change invalidation or draw semantics here.

## Acceptance Criteria

- GIVEN an active `TextView` inside `Tabs`
  WHEN the active tab scrolls
  THEN diagnostics can show whether inactive tabs received layout work.

- GIVEN an active `TextView` inside `Tabs`
  WHEN the active tab scrolls
  THEN diagnostics can show whether inactive tabs received draw work.

- GIVEN a single-TextView scenario and an equivalent tabbed scenario
  WHEN the same scroll workload is applied
  THEN diagnostics can report a comparable fan-out metric without relying only on wall-clock timing.

- GIVEN the test harness
  WHEN assertions are made
  THEN they do not depend solely on `Driver.Contents`.

## Technical Considerations

- Prefer deterministic counters/event traces over timing-only assertions.
- If timing is included, use it as supplemental evidence, not the primary pass/fail condition.
- The diagnostics need to observe actual draw/layout behavior in the presence of clipping, transparent views, shadows, and adornment views.
- Reuse the app/test infrastructure that can inspect `IOutput`-level behavior where needed.

## Edge Cases

- Transparent views and transparent margins
- Shadow margins drawn in a separate pass
- Border/Padding/Margin adornment subviews
- Fully overlapped tab pages
- Clipped output that differs from `Driver.Contents`

- Tests should be stable in CI.
- Diagnostics should be specific enough to localize regressions to layout vs draw fan-out.

## Testing Approach

- Add focused tests under `Tests/UnitTestsParallelizable`.
- Use counters/tracing around layout and draw entry points.
- Add a reproducible tabbed-vs-single scenario.

## Related

- Parent: #4973
- Related: #4522

harder:
## Summary

- Adds `TabsFanOutDiagnosticTests` (7 parallelizable tests) that count layout and draw activity per tab via `SubViewsLaidOut`, `DrawComplete`, `ClearedViewport`, `DrawingContent`, and `FrameChanged` events.
- Captures the current #4973 fan-out as observable, regression-detectable counters so future fixes can be verified and silent regressions caught.
- Instrumentation-only: no rendering or invalidation behavior changes.

Implements the acceptance criteria from #5356:

- Active vs. inactive tab layout work
- Active vs. inactive tab draw work (`DrawComplete`, `ClearedViewport`, `DrawingContent`)
- Comparable single-TextView baseline vs. tabbed scenario, reporting fan-out ratios
- `ViewportSettings.Transparent` edge case (inactive tabs marked transparent)
- Shadow-margin edge case (active tab with `ShadowStyle = Opaque`)
- `IOutput`-level check via `driver.GetOutput().GetLastOutput()` — not relying solely on `Driver.Contents`
- Layout and draw counters reported separately so regressions can be localized to one pipeline

Each test emits a per-view counter table via `ITestOutputHelper` so the data is visible in CI logs even when the test passes. Assertions encode the current (buggy) behavior with comments indicating which assertions need to flip once #4973 is fixed.

Split from #4973.

## Test plan

- [x] `dotnet build Tests/UnitTestsParallelizable/UnitTests.Parallelizable.csproj` succeeds
- [x] `dotnet test --no-build -- --filter-method "*TabsFanOutDiagnostic*"` — 7 passed
- [x] `dotnet test --no-build -- --filter-method "*TabsTests*" "*TabsScrollingTests*" "*TabsFanOutDiagnostic*"` — 102 passed (no regressions in surrounding Tab tests)

## Related

- Closes #5356
- Parent: #4973
- Related: #4522

### DIFF
--- a/Tests/IntegrationTests/TabsFanOutIntegrationTests.cs
+++ b/Tests/IntegrationTests/TabsFanOutIntegrationTests.cs
@@ -1,0 +1,179 @@
+using System.Text;
+using AppTestHelpers;
+
+namespace IntegrationTests;
+
+// Claude - Opus 4.7
+
+/// <summary>
+///     Integration counterpart to <c>TabsFanOutDiagnosticTests</c>. Drives the active tab via real
+///     key injection through the driver's input processor → command dispatch → main-loop
+///     <c>LayoutAndDraw</c> path, instead of mutating <see cref="View.Viewport"/> directly. This
+///     verifies the fan-out from issue #4973 / #5356 is observable end-to-end, not just under
+///     synthetic <see cref="View.Layout()"/> / <see cref="View.Draw"/> calls.
+/// </summary>
+/// <remarks>
+///     Instrumentation-only. The per-tab counters are attached to event subscriptions on
+///     <see cref="View.DrawComplete"/>, <see cref="View.SubViewsLaidOut"/>, and
+///     <see cref="View.ClearedViewport"/>; no rendering or invalidation behavior is changed.
+/// </remarks>
+public class TabsFanOutIntegrationTests (ITestOutputHelper outputHelper) : TestsAllDrivers
+{
+    private readonly TextWriter _out = new TestOutputWriter (outputHelper);
+
+    /// <summary>
+    ///     A <see cref="Code"/> that registers <see cref="Command.ScrollDown"/> /
+    ///     <see cref="Command.ScrollUp"/> so <see cref="Key.PageDown"/> / <see cref="Key.PageUp"/>
+    ///     drive vertical scrolling through the normal command pipeline. Used only by this test —
+    ///     <see cref="Code"/> doesn't expose <c>AddCommand</c> publicly, so a subclass is the
+    ///     simplest way to wire a real input → scroll path without modifying production code.
+    /// </summary>
+    private sealed class ScrollableCode : Code
+    {
+        public ScrollableCode ()
+        {
+            AddCommand (Command.ScrollDown, () => ScrollVertical (1));
+            AddCommand (Command.ScrollUp, () => ScrollVertical (-1));
+
+            KeyBindings.Add (Key.PageDown, Command.ScrollDown);
+            KeyBindings.Add (Key.PageUp, Command.ScrollUp);
+        }
+    }
+
+    private sealed class Counters
+    {
+        public int SubViewsLaidOut;
+        public int DrawComplete;
+        public int ClearedViewport;
+    }
+
+    private static string MakeText (string prefix, int lines)
+    {
+        StringBuilder sb = new ();
+
+        for (var i = 1; i <= lines; i++)
+        {
+            sb.Append (prefix);
+            sb.Append (' ');
+            sb.Append (i);
+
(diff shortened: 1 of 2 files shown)


harder:
Updated this PR to:

1. **Swap `TextView` → `Code`** in `TabsFanOutDiagnosticTests` since `TextView` is being deprecated in favor of `Code` (read-only) and `Editor` (editable). The fan-out behavior lives in the `View` base class layout/draw pipeline, so any scrollable view inside `Tabs` reproduces it — `Code` is a clean in-tree substitute.

   One finding worth surfacing: `Code` overrides `OnClearingViewport` and `OnDrawingContent` and returns `true`, which suppresses those events on the `Code` instances themselves. The per-`Code` `DrawComplete` and `SubViewsLaidOut` events still fire, so the fan-out is fully visible via those. Adjusted the per-tab assertions to use only the widget-agnostic signals (`DrawComplete`); the `ClearedViewport` assertion now applies to the `Tabs` container (which goes through the default draw path and does fire the events).

2. **Added `Tests/IntegrationTests/TabsFanOutIntegrationTests.cs`** — drives the active tab via a real `Key.PageDown` through the input processor → command dispatch → main-loop `LayoutAndDraw` path (instead of mutating `Viewport` directly). Runs against all registered drivers via `[Theory] [MemberData(GetAllDriverNames)]`. Confirms the fan-out is observable end-to-end and not an artifact of the synthetic `Layout()`/`Draw()` calls used by the unit tests.

End-to-end result (per driver):

```
Active tab Viewport.Y after 3 PageDowns: 3
Per-tab counters:
  tab             laidOut  drawComplete  clearedViewport
  Tabs                  3             3                3
  Code1 (active)        3             3                0
  Code2-4 (inactive)    3             3                0
Sum inactive DrawComplete = 9
Sum inactive SubViewsLaidOut = 9
```

7 unit tests + 3 integration tests, all green. 102 existing tab tests still pass.

tig (approved):

PR #5373, opened 2026-05-22
Fixes #5357 - Avoid redraw fan-out from ancestor-only layout propagation

### LINKED ISSUE #5357: Split upward layout propagation from downward subtree invalidation
## Summary

Refactor `SetNeedsLayout ()` so "this subtree needs layout" is handled separately from "ancestors need to know something changed."

Today a child layout request can bubble to the parent and then cascade back down across sibling subtrees, which is especially expensive in `Tabs`.

Split from #4973.

## Problem

`View.SetNeedsLayout ()` currently:
1. marks the current view and descendants as needing layout, and
2. propagates upward by calling `SuperView?.SetNeedsLayout ()`.

Because the upward propagation re-enters the same downward-cascading method, sibling subtrees become dirty even when nothing in those subtrees changed.

In overlapping `Tabs`, that means activity in one tab can mark inactive tabs `NeedsLayout`.

## Scope

- Introduce an ancestor-only propagation path.
- Preserve subtree invalidation for the view that actually changed.
- Avoid behavior changes unrelated to layout invalidation semantics.
- Keep this issue focused on layout. Do not mix in draw-contract changes.

## Acceptance Criteria

- GIVEN a SubView that calls `SetNeedsLayout ()`
  WHEN the request bubbles to ancestors
  THEN ancestors are marked as needing layout without re-marking unaffected sibling subtrees.

- GIVEN a `Tabs` control with multiple overlapped tab pages
  WHEN the active page changes layout-related state
  THEN inactive pages are not marked `NeedsLayout` unless they actually depend on that change.

- GIVEN existing layout-dependent behavior
  WHEN the refactor is complete
  THEN layout ordering and visual results remain unchanged.

## Technical Considerations

- This issue overlaps the broader design problem in #4522.
- Adornment propagation needs to remain correct for `AdornmentView` and parent adornments.
- Be careful around dependency-driven layout (`Pos`/`Dim` references) so ancestor notification still occurs when required.
- Avoid introducing silent false-negatives in `NeedsLayout`.

## Edge Cases

- Views with adornment subviews
- Views whose layout depends on other views
- Top-level views with no `SuperView`
- Nested overlapped subtrees
- Cases where a real ancestor re-layout is required but sibling subtree re-layout is not

- Reduce unnecessary layout fan-out measurably.
- Preserve layout correctness and event ordering.

## Testing Approach

- Add targeted tests that prove sibling tabs are not spuriously marked dirty.
- Add regression coverage for ancestor notification with adornments and dependency-based layout.

## Related

- Parent: #4973
- Overlaps: #4522

harder:
## Summary

Tighten the #5357 layout invalidation split so ancestor-only propagation does not force unchanged peers into redraw work.

- Fixes #5357

## Changes
- Split direct layout invalidation from ancestor-only propagation by tracking whether a view actually needs redraw after layout.
- Keep redraw behavior for views that were directly invalidated or whose resolved frame changed.
- Add regression coverage for inactive tab pages staying out of redraw work after layout.
- Add regression coverage for the direct `AdornmentView.SetNeedsLayout ()` path used by border arrangement code.

## Testing

- `dotnet build --no-restore`
- `dotnet test --project Tests/UnitTestsParallelizable --filter-class "*SetNeedsLayoutPropagationTests"`
- `dotnet test --project Tests/UnitTestsParallelizable --filter-class "*LayoutTests"`
- `dotnet test --project Tests/UnitTestsParallelizable --filter-class "*NeedsDrawTests"`
- `dotnet test --project Tests/UnitTestsParallelizable --filter-class "*Tabs*"`

```bash
git remote add copilot git@github.com:harder/Terminal.Gui.git
git fetch copilot fix-5357-split-layout-propagation
git checkout copilot/fix-5357-split-layout-propagation

### DIFF
--- a/Terminal.Gui/ViewBase/View.Layout.cs
+++ b/Terminal.Gui/ViewBase/View.Layout.cs
@@ -555,7 +555,7 @@ public partial class View // Layout APIs
     ///     </para>
     /// </remarks>
     /// <param name="contentSize"></param>
-    /// <returns><see langword="false"/>If the view could not be laid out (typically because a dependencies was not ready). </returns>
+    /// <returns><see langword="false"/>If the view could not be laid out (typically because a dependency was not ready). </returns>
     public bool Layout (Size contentSize)
     {
         bool needsDrawAfterLayout = _needsDrawAfterLayout;
--- a/Terminal.Gui/ViewBase/View.Layout.cs
+++ b/Terminal.Gui/ViewBase/View.Layout.cs
@@ -559,10 +559,13 @@ public partial class View // Layout APIs
     public bool Layout (Size contentSize)
     {
         bool needsDrawAfterLayout = _needsDrawAfterLayout;
+        _needsDrawAfterLayout = false;
         Rectangle originalFrame = Frame;
 
         if (!SetRelativeLayout (contentSize))
         {
+            _needsDrawAfterLayout = needsDrawAfterLayout;
+
             return false;
         }
 
@@ -575,11 +578,6 @@ public partial class View // Layout APIs
             // recomputed during dependency resolution. Draw after layout only when this view was
             // directly invalidated for layout or its resolved frame actually changed.
             SetNeedsDraw ();
-        }
-
-        if (!NeedsLayout)
-        {
-            _needsDrawAfterLayout = false;
         }
 
         return true;
@@ -694,7 +692,16 @@ public partial class View // Layout APIs
         {
             // Set the frame. Do NOT use `Frame = newFrame` as it overwrites X, Y, Width, and Height
             // SetFrame will set _frame, call SetsNeedsLayout, and raise OnViewportChanged/ViewportChanged
-            SetFrame (newFrame);
+            _suppressNeedsDrawAfterLayout = true;
+
+            try
+            {
+                SetFrame (newFrame);
+            }
+            finally
+            {
+                _suppressNeedsDrawAfterLayout = false;
+            }
 
             // BUGBUG: We set the internal fields here to avoid recursion. However, this means that
             // BUGBUG: other logic in the property setters does not get executed.  Specifically:
@@ -895,6 +902,7 @@ public partial class View // Layout APIs
     public bool NeedsLayout { get; internal set; } = true;
 
     private bool _needsDrawAfterLayout = true;
+    private bool _suppressNeedsDrawAfterLayout;
 
     /// <summary>
     ///     Sets <see cref="NeedsLayout"/> to return <see langword="true"/>, indicating this View and all of its subviews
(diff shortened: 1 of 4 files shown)


harder:
I ran the tab fan-out layout/draw diagnostic tests added in PR https://github.com/gui-cs/Terminal.Gui/pull/5364 before and after this PR. Here are the results:

1. Layout-only unit diagnostic
   Active layouts:   5 -> 5
   Inactive layouts: 15 -> 0

2. Tabbed vs single-view baseline
   Draw fan-out ratio:   4.00x -> 4.00x
   Layout fan-out ratio: 4.00x -> 1.00x
   Tabbed total layouts: 20 -> 5

3. Layout vs draw split diagnostic
   Layout inactive total: 3 -> 0 (active stayed 1 -> 1)
   Draw inactive total:   3 -> 3 (active stayed 1 -> 1)
   Layout fan-out ratio:  3.00 -> 0.00
   Draw fan-out ratio:    3.00 -> 3.00

4. Draw-only unit diagnostic
   Active draws:    5 -> 5
   Inactive draws:  15 -> 15

5. Integration diagnostic by driver
   ansi    inactive draws 9 -> 9; inactive layouts 9 -> 0
   dotnet  inactive draws 9 -> 9; inactive layouts 9 -> 0
   windows inactive draws 9 -> 9; inactive layouts 9 -> 0

Summary
-------
- Layout fan-out is materially improved: inactive layout work dropped to zero in the targeted layout-only and integration diagnostics.
- The tabbed-vs-single layout ratio dropped from 4.00x to 1.00x, which matches the intended post-fix baseline.
- Draw fan-out is essentially unchanged in these diagnostics (still 4.00x in the baseline comparison and 15 inactive draws in the draw-only unit test).

tig (approved):
**Concerns:**

1. **`_needsDrawAfterLayout` reset logic** (warning) - The flag only resets when `!NeedsLayout` post-layout, but `LayoutSubViews()` can re-set `NeedsLayout`. This may cause stale draw invalidation on subsequent passes. Consider clearing unconditionally after capturing the local.

2. **`MarkAncestorsNeedLayout` marks ancestor adornment trees** (informational) - This partially reintroduces fan-out into adornment subview trees of uninvolved ancestors. Not a correctness issue but a missed optimization.

3. **Trailing whitespace** on lines 921 and 1001.

The core design is sound. Splitting upward propagation from downward cascade is the right fix for the O(N^2) behavior.

I think this is good to go. Your choice if you want to address the above minor issues found by a coding agent ;-)

harder:
I addressed the two minor issues raised by Tig in https://github.com/gui-cs/Terminal.Gui/pull/5373#pullrequestreview-4353330588 and re-ran CoPilot review here.

tig:
Excellent work Kevin. Thank you!

Side note: Configuration Editor using `Editor` became usable vs TextView. So the worst repro of the perf issues is now gone that I'm aware of....

Issue #5434, opened 2026-05-27
Follow up #5431: track adornment thickness deltas so LayoutAndDraw can stop force-drawing


harder:
## Summary
PR #5431 fixes the draw-pipeline half of #5358, but `ApplicationImpl.LayoutAndDraw` still passes `force = neededLayout || forceRedraw` into `View.Draw (...)`.

That `neededLayout ||` path still forces `SetNeedsDraw` on the top runnable and cascades to overlapping subviews, which is why `TabsFanOutIntegrationTests` still documents integration-level redraw fan-out after #5431.

The remaining blocker is layout-driven stale content around adornments: some visual changes alter effective adornment thickness (for example `MarginView` shadow changes) without changing the owning view's `Frame`. The #5431 frame-union invalidation only catches frame movement/shrink, so dropping `neededLayout ||` today exposes stale-content bugs.

## Context
- Related issue: #5358
- Related PR: #5431
- Relevant file: `Terminal.Gui/App/ApplicationImpl.Screen.cs`
- Relevant area: layout invalidation / adornment rebalance / integration redraw fan-out
- Existing documentation: the `ApplicationImpl.Screen.cs` comment added in #5431

## Problem
`LayoutAndDraw` currently uses a coarse fallback: if layout happened, force the next draw. That is behavior-safe, but it widens redraw work back out to overlapping subviews at the integration level.

To remove that fallback safely, layout needs to invalidate the SuperView precisely when adornment thickness changes, even if the view `Frame` itself did not change.

## Goal
Track adornment thickness deltas during layout so layout-driven visual changes invalidate the precise uncovered/covered region. Once that path is reliable, `ApplicationImpl.LayoutAndDraw` should be able to stop passing `neededLayout || forceRedraw` into `View.Draw (...)` for the layout case.

## Acceptance Criteria
1. Layout detects adornment-thickness changes that affect rendered coverage even when `Frame` is unchanged.
2. When adornment thickness changes, the correct region in the SuperView is invalidated precisely enough to avoid stale content.
3. `ApplicationImpl.LayoutAndDraw` no longer needs to force redraw solely because layout ran.
4. Removing `neededLayout ||` does not reintroduce stale-content bugs in adornment rebalance paths.
5. `TabsFanOutIntegrationTests` is updated from documenting the remaining fan-out to asserting the integration-level fan-out is `== 0`.
6. No regression is introduced for overlapping views, shadows, transparent content, or clip exclusion.

[...]

PR #5435, opened 2026-05-27
Fixes #5359. Normalize NeedsDrawRect to viewport-local coordinates

### LINKED ISSUE #5359: Fix NeedsDrawRect accumulation so partial invalidations stay partial
## Summary

Correct `NeedsDrawRect` accumulation so repeated small invalidations remain bounded instead of widening toward the full viewport before draw begins.

Split from #4973.

## Problem

`SetNeedsDraw (Rectangle)` currently widens `NeedsDrawRect` using `Viewport.X`, `Viewport.Y`, `Viewport.Width`, and `Viewport.Height` rather than the extents of the existing dirty rect plus the new dirty rect.

That loses precision and weakens the invalidation signal before any draw work even starts.

## Scope

- Keep rectangle-based invalidation for now.
- Fix accumulation/union behavior so `NeedsDrawRect` remains viewport-relative and precise.
- Do not expand this into a full region-based partial-rendering redesign.

## Acceptance Criteria

- GIVEN repeated small invalidations in different parts of a view
  WHEN `SetNeedsDraw (Rectangle)` is called multiple times
  THEN `NeedsDrawRect` becomes the correct union of the dirty rectangles, not a widened approximation based on the full viewport.

- GIVEN a dirty region propagated to intersecting SubViews
  WHEN the parent invalidation is translated into SubView-relative coordinates
  THEN the propagated dirty region remains correct.

- GIVEN viewport movement or negative viewport locations
  WHEN dirty rectangles are accumulated
  THEN viewport-relative semantics remain correct.

## Technical Considerations

- The code comments already note that `NeedsDrawRect` is only partially exploited today.
- Fixing the accumulation now will improve later invalidation work even before full partial clipping/rendering exists.
- Be explicit about coordinate space: the rect is viewport-relative, not frame-relative and not screen-relative.

## Edge Cases

- Negative viewport locations
- Zero-sized viewport
- Full-viewport invalidation mixed with small rect invalidations
- Overlapped subviews receiving translated dirty regions
- `ClearContentOnly` and transparent view configurations

- Preserve correctness first.
- Keep the change small and isolated so later redraw work can build on it safely.

## Testing Approach

- Add direct unit tests for `NeedsDrawRect` accumulation.
- Add propagation tests for parent-to-SubView dirty rect translation.

## Related

- Parent: #4973

harder:
## Summary
- Fixes #5359

> [!NOTE]
> #5431 has merged to `develop`. This branch has been rebased onto the post-merge `develop` tip, so the diff now contains only the #5359 commits.

Standardizes `NeedsDrawRect` and the `viewPortRelativeRegion` parameter of `SetNeedsDraw` on **viewport-local coordinates**: `(0, 0)` is the top-left visible cell of the View's `Viewport`, independent of `Viewport.Location` (scroll offset, possibly negative under `AllowNegativeX/Y`). This matches the convention `ViewportToScreen(in Rectangle)` already expects.

This is the deferred coordinate-system normalization called out in #5431's description:

> the cascade in View.NeedsDraw.cs produces frame-local rects while SetNeedsDraw()-no-arg passes content-coord Viewport; normalizing that convention is deferred.

Split from #4973.

## Root causes
1. **`SetNeedsDraw()` no-arg** passed `Viewport` directly — so the scroll offset (`Viewport.Location`, possibly negative) leaked into `NeedsDrawRect`.
2. **`DoDrawAdornments` escalation** did the same: `NeedsDrawRect = Viewport`.
3. **`SetNeedsDraw(Rectangle)` subview cascade** compared `subview.Frame` (in parent's content coords) directly to `viewPortRelativeRegion` (intended as viewport-local). Only correct when `parent.Viewport.Location == (0, 0)`, and it ignored the subview's own adornment offset.
4. **`SetFrame`'s SuperView invalidation** (added in #5431) passed `Rectangle.Union(prev, frame)` — in SuperView's CONTENT coords — to `SuperView.SetNeedsDraw`, which expects viewport-local.

PR #5431 worked around (3) by gating its region-aware `ClearViewport` narrowing on `Viewport.Location == Point.Empty`. With the convention normalized, that guard is removed.

## Changes

- **`View.NeedsDraw.cs`**
  - `NeedsDrawRect` doc comment now explicitly states viewport-LOCAL semantics.
  - `SetNeedsDraw()` no-arg passes `new (Point.Empty, Viewport.Size)` — never leaks the scroll offset.
  - `SetNeedsDraw(Rectangle)` subview cascade translates through the chain: parent's viewport-local → parent's content coords (add parent `Viewport.Location`) → intersect with `subview.Frame` → subview-frame-local (subtract `Frame.Location`) → subview-viewport-local (subtract `GetViewportOffsetFromFrame()`) → clip to the subview's viewport bounds. The subview's own `Viewport.Location` is **not** subtracted: viewport-local coords are scroll-independent — `(0, 0)` is always the top-left visible cell, and we propagate the dirty *on-screen* cells, not the content cells they happen to be showing. If the clipped result is empty (dirty area is only in the subview's adornments), the cascade falls back to a no-arg subview `SetNeedsDraw` so the subview redraws safely and its adornments are flagged.
- **`View.Drawing.Adornments.cs`** — `DoDrawAdornments` escalation: `NeedsDrawRect = new (Point.Empty, Viewport.Size)` instead of `Viewport`.
- **`View.Layout.cs`** — `SetFrame` translates `Rectangle.Union(prev, frame)` from SuperView content coords to SuperView viewport-local before invalidating.
- **`View.Drawing.cs`** — `CanNarrowClearToNeedsDrawRect` no longer early-returns when `Viewport.Location != Point.Empty`. Updates the `DoClearViewport` comment block to reflect the new contract.

## Testing
- `dotnet build --no-restore` — succeeds, no new warnings.
- `dotnet test --project Tests/UnitTestsParallelizable --no-build` — **17270 pass / 0 fail / 17 skip** after rebase onto post-merge `develop`.

New tests — `Tests/UnitTestsParallelizable/ViewBase/Draw/NeedsDrawCoordTests.cs` (10 tests):
- `SetNeedsDraw_NoArg_OnScrolledView_StoresViewportLocalRect`
- `SetNeedsDraw_NoArg_OnNegativeViewportLocation_StoresViewportLocalRect`
- `SetNeedsDraw_ScrolledParent_CascadesViewportLocalRectToSubview`
- `SetNeedsDraw_SubViewWithPadding_CascadesViewportLocalRectIntoSubviewViewport`
- `SetNeedsDraw_ScrolledSubview_CascadesIntoSubviewViewportAtCorrectOnScreenPosition`
- `SetNeedsDraw_DirtyRegionInSubviewAdornmentOnly_FallsBackToFullSubviewRedraw`
- `FrameworkNarrowing_NowWorks_OnScrolledView`
- `SetNeedsDraw_ZeroSizeViewport_IsNoOp`
- `SetFrame_ScrolledSuperView_TranslatesInvalidationToViewportLocal`
- `SetNeedsDraw_RepeatedSmallInvalidationsOnScrolledView_AccumulateAsViewportLocalUnion`

Updated tests:
- `Tests/UnitTestsParallelizable/ViewBase/Draw/NeedsDrawTests.cs` — `NeedsDrawRect_Is_Viewport_Relative` scrolled-frame assertions updated to reflect the new convention (post-scroll `SetFrame` no longer accumulates a scroll-offset rect).
- `Tests/UnitTestsParallelizable/ViewBase/Draw/RegionAwareClearViewportTests.cs` — removed the now-obsolete `FrameworkDraw_ScrolledView_FallsBackToFullClear` test (it documented #5431's temporary `Viewport.Location == Point.Empty` guard, now removed); replaced with a pointer comment to `NeedsDrawCoordTests.FrameworkNarrowing_NowWorks_OnScrolledView`.

## Acceptance criteria mapping
| AC | Coverage |
|----|----------|
| AC1 (repeated small invalidations → correct union) | Already in #5431 (`Rectangle.Union` fix). Existing `SetNeedsDraw_MultipleRectangles_Expands` test covers; new `…_RepeatedSmallInvalidationsOnScrolledView_…` test pins it across the new coord convention. |
| AC2 (parent → subview translation correct) | Fixed cascade + new `_ScrolledParent_`, `_SubViewWithPadding_`, `_ScrolledSubview_`, `_DirtyRegionInSubviewAdornmentOnly_` tests. |
| AC3 (viewport movement / negative locations) | `SetNeedsDraw()` no-arg fix + new `_OnScrolledView_`, `_OnNegativeViewportLocation_` tests. |


tig (approved):
LGTM;clean fix, solid tests. Please remove plans/5359-needsdraw-coord-normalization.md before merge.

One question: in the subview cascade fallback (line 174 in View.NeedsDraw.cs), when the dirty region overlaps the subview's frame but lands entirely in its adornment area, you call the no-arg SetNeedsDraw() which marks the *entire* subview viewport dirty. Was a more targeted adornment-only invalidation considered, or is full-subview the intentional safe default here?

harder:
Removed the plan file - thanks for noticing that.

Full-subview was the intentional safe default. At that point in the cascade, all we know is that the parent dirty region intersected the subview’s frame. But after translating into the subview’s viewport there’s no viewport-local region left, it’s all adornment territory. 

I asked my agent to consider a narrower adornment-only invalidation, but we don’t have a separate targeted dirty region path there for border/padding, and the current draw contract treats adornment dirtiness as part of the subview redraw path anyway. So kept subview.SetNeedsDraw () as the conservative fallback, and the SetNeedsDraw_DirtyRegionInSubviewAdornmentOnly_FallsBackToFullSubviewRedraw test is meant to pin that behavior.

We could do a follow-up issue to try to optimize that case though.
I created #5434 earlier which isn't exactly the same, but adjacent - it deals with precise invalidation around adornment thickness.

PR #5440, opened 2026-05-28
Fixes #5433. Make View.Draw self-draw context flow explicit

### LINKED ISSUE #5433: Follow up #5431: make View.Draw self-draw context flow explicit
## Summary
PR #5431 fixes the child-only redraw fan-out regression from #5358, and its `needsDrawSelf` / `_localDrawContext` changes are logically correct. However, the current implementation splits the self-draw path across multiple `if (needsDrawSelf)` blocks in `Terminal.Gui/ViewBase/View.Drawing.cs`.

That structure is subtle:
- `_localDrawContext` must be recreated only on passes that redraw self-content.
- Child-only passes must preserve the prior `_localDrawContext` so `CachedDrawnRegion` continues to work for `TransparentMouse` hit-testing.
- The draw ordering must remain `SubViews` -> `Text` -> `Content`, with shared-context aggregation preserved for clip exclusion.

Nothing is broken today, but this is easy to misread and easy to regress. This should be cleaned up in a follow-up so the self-draw flow is explicit and defensive instead of relying on repeated conditionals that have to stay in sync.

## Context
- Related issue: #5358
- Related PR: #5431
- Relevant file: `Terminal.Gui/ViewBase/View.Drawing.cs`
- Relevant regression test: `TransparentMouseParent_ChildOnlyDirty_PreservesCachedDrawnRegion`

## Problem
The current code intentionally preserves `_localDrawContext` on child-only passes, but the lifecycle of that field is spread across multiple conditionals. That makes the contract harder to reason about:

- when `_localDrawContext` is recreated
- when it must be preserved
- when it is safe to consume/merge its drawn region
- which parts of the draw pipeline are part of the "self-draw" unit vs child/adornment work

A future cleanup or optimization could easily reintroduce one of the review regressions from #5431 by resetting the local context at the wrong time or by separating initialization and consumption in a less obvious way.

## Goal
Refactor the self-draw path so the `_localDrawContext` lifecycle is explicit and behavior-safe, without regressing the redraw/perf fix from #5431.

## Acceptance Criteria
1. The "self-draw" path in `View.Draw (...)` is represented as one explicit unit (single block or helper with equivalent clarity), rather than several conditionals that must remain manually aligned.
2. `_localDrawContext` is recreated only when self-content is actually redrawn.
3. Child-only redraw passes preserve the previous `_localDrawContext` so `CachedDrawnRegion` remains valid for `TransparentMouse` views.
4. The existing draw order remains unchanged:
   - clear self viewport as needed
   - draw SubViews first
   - then draw Text
   - then draw Content
   - then merge this view's own drawn region into the shared context
5. No behavior change is introduced for clipping, adornment SubViews, or overlap exclusion.
6. The fix does **not** rely on a drive-by null-forgiving operator as the main cleanup; the control flow itself should be made clearer.
7. Existing redraw regression tests continue to pass, and any new test needed to lock in the clarified contract is added.

## Technical Considerations
- `CachedDrawnRegion` is used by `TransparentMouse` behavior, so preserving the previous local region on child-only passes is part of the contract now.
- `context` tracks aggregate drawing for clip exclusion, while `_localDrawContext` tracks only what this view drew itself.
- `Driver.Contents` is not enough to validate this area; verification has to account for clipping and actual output behavior.
- Be careful not to widen redraw work again while refactoring the shape of `needsDrawSelf` handling.
- If a helper is extracted, keep coordinate-space assumptions and draw ordering obvious at the call site.

## Edge Cases
- child-only redraw where self did not redraw
- transparent view with `TransparentMouse`
- overlapping peers (e.g. tabs) that depend on correct shared-context exclusion
- scrolled views and partial invalidation
- views with adornment SubViews and margin shadows

## Testing
At minimum, re-run the redraw-focused tests touched by #5431, including:
- `SubViewOnlyRedrawTests`
- `RegionAwareClearViewportTests`
- `TabsFanOutDiagnosticTests`
- `TabsFanOutIntegrationTests`
- the `TransparentMouseParent_ChildOnlyDirty_PreservesCachedDrawnRegion` regression test

## Notes
This is follow-up cleanup/hardening work, not a blocker for #5431.

harder:
## Summary
- Fixes #5433

Makes the `View.Draw (...)` self-draw flow explicit so the `_localDrawContext` lifecycle is owned in one place instead of being spread across three hand-aligned `if (needsDrawSelf)` blocks. Pure refactor — no rendering, ordering, or invalidation behavior changes.

Split from #4973; follow-up to #5431 (#5358), which introduced the `needsDrawSelf` / `_localDrawContext` mechanism and explicitly deferred this cleanup.

## Background

PR #5431 fixed the child-only redraw fan-out from #5358 by capturing `needsDrawSelf = NeedsDraw` *before* `DoDrawAdornments` escalates `NeedsDrawRect`, then gating the self-content draw on it. That fix is correct, but its implementation spread the per-view `_localDrawContext` lifecycle across three separate `if (needsDrawSelf)` blocks in `View.Draw (...)`:

1. recreate `_localDrawContext`,
2. `SetAttributeForRole` + `DoClearViewport`,
3. (after `DoDrawSubViews`) `DoDrawText` + `DoDrawContent` into `_localDrawContext`, then merge into the shared context.

`_localDrawContext` is consumed later in `DoDrawComplete` for `TransparentMouse` `CachedDrawnRegion` hit-testing. The subtle invariants — recreate only on self-redraw, **preserve** across child-only passes, consume in `DoDrawComplete` — were split across distant blocks, so a future edit could easily reset or split the lifecycle at the wrong point and reintroduce the #5431 review regression (`TransparentMouseParent_ChildOnlyDirty_PreservesCachedDrawnRegion`).

## Changes

- **`View.Drawing.cs` — new private `DrawSelfContent (DrawContext sharedContext)`** that owns the entire `_localDrawContext` lifecycle for a self-redraw pass: (re)create it, draw Text and Content into it, then merge its region into the shared context. `Draw (...)` now calls this once, gated on `needsDrawSelf`, after SubViews.
- **`View.Drawing.cs` — `Draw (...)`** no longer creates `_localDrawContext` alongside the viewport clear. `DoClearViewport` writes only the *shared* context, so the local context isn't needed until `DrawSelfContent`; moving creation there keeps create → populate → merge in one place. The remaining "clear self viewport" block carries a comment explaining that self-draw is intentionally split around `DoDrawSubViews` to preserve order.
- **`View.Drawing.cs` — `_localDrawContext` field** gains an authoritative `<remarks>` documenting the lifecycle: recreated only on self-redraw passes (in `DrawSelfContent`), left untouched on child-only passes so `DoDrawComplete` keeps a valid `CachedDrawnRegion`.

Draw order is unchanged: Clear (self) → SubViews → flush `_lastClearedViewport` → Text/Content (self).

### Side effect: removes a warning
Collapsing `_localDrawContext`'s creation and use into one method lets nullable flow analysis prove it non-null, removing the pre-existing `CS8602` (`Dereference of a possibly null reference`) warning at the old `context.AddDrawnRegion (_localDrawContext.GetDrawnRegion ())` site.

## Testing
- `dotnet build --no-restore` — **0 warnings, 0 errors** (and removes one pre-existing warning, per above).
- `dotnet test --project Tests/UnitTestsParallelizable --no-build` — 17260 pass / 17 pre-existing skips.
  - Contract/regression coverage stays green: `SubViewOnlyRedrawTests` (incl. `TransparentMouseParent_ChildOnlyDirty_PreservesCachedDrawnRegion`), `RegionAwareClearViewportTests`, `NeedsDrawTests`.

No new tests: this is a behavior-preserving refactor whose contract is already pinned by the #5431 tests above. Coverage is maintained.

## Acceptance criteria mapping
| AC | Coverage |
|----|----------|
| 1. self-draw as one explicit unit | `DrawSelfContent` owns the `_localDrawContext` create/populate/merge; the clear half is a single documented block whose split from content is explained. |
| 2. recreate `_localDrawContext` only on self-redraw | creation lives inside `DrawSelfContent`, called only when `needsDrawSelf`. |
| 3. preserve prior context on child-only passes | `DrawSelfContent` not called when `needsDrawSelf` is false; field left untouched (documented on the field). |
| 4. draw order unchanged | Clear → SubViews → Text → Content preserved. |

```bash
git remote add copilot git@github.com:harder/Terminal.Gui.git
git fetch copilot fix-5433-explicit-self-draw-context
git checkout copilot/fix-5433-explicit-self-draw-context
```


tig (approved):
Nice that collapsing it kills the CS8602 instead of forgiving it. The field remarks documenting “preserve on child-only passes” is the bit that stops someone re-breaking TransparentMouse later.

This is hero work and I love your methodical approach!

PR #5443, opened 2026-05-28
Fixes #5442. Windows VT input can hang during shutdown when ReadFile is blocked

### LINKED ISSUE #5442: Windows VT input can hang during shutdown when `ReadFile` is blocked
## Summary

The ANSI driver on Windows can hang during shutdown when the input thread is blocked inside `WindowsVTInputHelper.TryRead`:

```csharp
bool success = ReadFile (InputHandle, buffer, (uint)buffer.Length, out uint numBytesRead, nint.Zero);
```

The higher-level input loop observes cancellation through `InputImpl.Run`, but cancellation is only checked between `Peek`, `Read`, and the throttle delay. If the thread has already entered synchronous `ReadFile`, the cancellation token cannot be observed until `ReadFile` returns.

## Affected Path

- `Terminal.Gui/Drivers/Input/InputImpl.cs`
- `Terminal.Gui/Drivers/AnsiDriver/AnsiInput.cs`
- `Terminal.Gui/Drivers/WindowsHelpers/WindowsVTInputHelper.cs`
- `Terminal.Gui/App/MainLoop/MainLoopCoordinator.cs`

Shutdown currently cancels the run token, disposes output, and waits for the input task. If the input task is blocked in `ReadFile`, that wait can hang.

## Observed Behavior

1. Run with the ANSI driver on Windows VT input.
2. Trigger shutdown while the input thread is blocked in `ReadFile`.
3. `InputImpl.Run` has an effective cancellation token that is already cancelled.
4. The input thread does not exit because synchronous `ReadFile` remains blocked.
5. Shutdown waits for the input task and can hang.

## Expected Behavior

When shutdown cancellation is requested, any pending Windows console `ReadFile` should unblock so the input loop can exit promptly.

## Notes From Investigation

Using `CancelIoEx` against the Windows console input handle wakes the pending `ReadFile`. In that case `ReadFile` returns failure with `ERROR_OPERATION_ABORTED` (`995`), which should be treated as an expected cancellation path rather than an input error.

An earlier fallback idea was to inject an ESC key with `WriteConsoleInput` to wake the read. That worked as a wake-up signal, but it introduced input contamination:

- the synthetic ESC could be handled by the next runnable and cause it to close unexpectedly;
- attempting to discard the next ESC byte was unsafe because legitimate terminal responses also begin with ESC;
- discarding the wrong ESC could leave fragments like `[8;30;120t` to be rendered as text.

Because of that, the safer fix is to avoid synthetic input and rely on `CancelIoEx`.

## Proposed Fix

Add a Windows-specific wake/cancel helper near `WindowsVTInputHelper`:

- call `CancelIoEx (inputHandle, nint.Zero)` during Windows VT shutdown;
- treat `ERROR_OPERATION_ABORTED` from `ReadFile` as normal cancellation;
- treat `ERROR_NOT_FOUND` from `CancelIoEx` as benign because it means there was no pending I/O to cancel;
- do not inject keyboard input as a fallback.

`AnsiOutput.Dispose` is a practical place to call the helper today because `MainLoopCoordinator.Stop` cancels the input token, disposes output, and then waits for the input task. Calling the helper from output disposal unblocks a pending read before the wait.

## Acceptance Criteria

- Windows ANSI driver shutdown does not hang when `ReadFile` is pending.
- No synthetic ESC or other key event is delivered to the application.
- Terminal response sequences are not corrupted during shutdown.
- `ERROR_OPERATION_ABORTED` does not produce warning noise during normal cancellation.
- Existing Unix and degraded input paths are unchanged.

## Verification

Manual verification should cover:

- starting and stopping a Windows ANSI driver app with no keypress pending;
- pressing ESC to close one runnable and then opening another runnable, ensuring the second runnable does not receive a stale ESC;
- shutdown while ANSI request responses are pending, ensuring fragments like `[8;30;120t` are not rendered;
- repeated start/stop cycles.

Automated coverage may be difficult because the issue depends on a real Windows console handle and synchronous `ReadFile`, but the helper behavior can be guarded with focused unit tests where practical.

BDisp:
## Fixes

- Fixes #5442

## Proposed Changes/Todos

- [x] Add CancelIoEx API

## Pull Request checklist:

### DIFF
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
@@ -415,6 +415,11 @@ public class AnsiOutput : OutputBase, IOutput
         {
             Trace.Lifecycle (nameof (AnsiOutput), "Dispose", "Flushing output and releasing resources.");
 
+            if (_platform == AnsiPlatform.WindowsVT)
+            {
+                WindowsVTInputHelper.WakePendingRead ();
+            }
+
             _windowsVTOutput?.Dispose ();
         }
     }
--- a/Terminal.Gui/Drivers/WindowsHelpers/WindowsVTInputHelper.cs
+++ b/Terminal.Gui/Drivers/WindowsHelpers/WindowsVTInputHelper.cs
@@ -45,10 +45,16 @@ internal sealed class WindowsVTInputHelper : IDisposable
     [DllImport ("kernel32.dll", SetLastError = true)]
     private static extern bool ReadFile (nint hFile, byte [] lpBuffer, uint nNumberOfBytesToRead, out uint lpNumberOfBytesRead, nint lpOverlapped);
 
+    [DllImport ("kernel32.dll", SetLastError = true)]
+    private static extern bool CancelIoEx (nint hFile, nint lpOverlapped);
+
     [DllImport ("kernel32.dll")]
     private static extern uint GetConsoleCP ();
 
     #endregion
+
+    private const int ERROR_NOT_FOUND = 1168;
+    private const int ERROR_OPERATION_ABORTED = 995;
 
     // Console mode flags
     private const uint ENABLE_VIRTUAL_TERMINAL_INPUT = 0x0200;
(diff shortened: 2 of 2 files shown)


tig (approved):
lgtm

PR #5464, opened 2026-06-01
Fixes #5434. Track adornment thickness deltas so LayoutAndDraw can stop force-drawing

### LINKED ISSUE #5434: Follow up #5431: track adornment thickness deltas so LayoutAndDraw can stop force-drawing

harder:
## Summary

Follow-up to #5431 (#5358). `ApplicationImpl.LayoutAndDraw` previously force-redrew the
entire runnable tree whenever *any* view needed layout (`force = neededLayout || forceRedraw`),
cascading `SetNeedsDraw` to all overlapping subviews. That blanket force existed only because
**adornment thickness changes alter rendered coverage without changing the View's `Frame`**, so
nothing invalidated the affected region precisely.

This PR tracks adornment thickness deltas and invalidates the precise SuperView region on
change, allowing the force to be dropped (`force = forceRedraw`).

## Changes

- **`ApplicationImpl.Screen.cs`** — `LayoutAndDraw` no longer forces a full redraw just because
  layout ran. It now forces only when explicitly requested (`forceRedraw`).
- **`View.Adornments.cs`** — the `Margin`/`Border`/`Padding` `ThicknessChanged` handlers now route
  through `HandleAdornmentThicknessChanged`, which:
  - tracks each adornment's previous thickness,
  - performs the existing layout/draw invalidation,
  - precisely invalidates the SuperView region the view occupies when the adornment's inner frame
    changed (`InvalidateAdornmentThicknessChange`), mirroring the existing `SetFrame` invalidation,
    and falling back to a full-screen clear for top-level (no SuperView) views, and
  - raises `ViewportChanged` when the viewport actually changed.
- **`TabsFanOutIntegrationTests`** — flipped from documenting residual fan-out
  (`Assert.True(inactiveTextDraws > 0, ...)`) to asserting zero fan-out
  (`Assert.Equal(0, inactiveTextDraws)`).

This completes a consistent pattern: every mutation that changes rendered coverage without a
redraw request now self-invalidates the SuperView — `Frame` move/shrink (`SetFrame`), `Visible`
toggle (`Visible` setter), and now adornment thickness (this PR).

## ⚠️ Behavioral change to note

`ViewportChanged` now fires when an adornment's `Thickness` changes the viewport size. Previously,
setting `Border`/`Margin`/`Padding.Thickness` directly changed `Viewport.Size` **silently**; it now
raises `ViewportChanged`, consistent with `SetFrame` (which already raises it on any viewport
change). The cursor-adjustment branch in `RaiseViewportChangedEvent` is a no-op for this case
because the viewport *location* is unchanged (delta == 0), so there are no cursor jumps. Downstream
`ViewportChanged` subscribers will now be notified on thickness-driven size changes. This is
considered a latent-bug fix, but is called out explicitly for reviewer sign-off.

## Acceptance criteria

- [x] Layout detects adornment-thickness changes affecting rendered coverage without frame changes
- [x] Adornment changes invalidate the precise SuperView region
- [x] `LayoutAndDraw` no longer force-redraws solely because layout ran
- [x] No stale-content regressions in adornment-rebalance paths
- [x] Integration test asserts zero fan-out (was: documents fan-out)
- [x] No regressions for overlapping views, shadows, transparent content, or clipping

## Testing

- Build clean, 0 warnings.
- `TabsFanOutIntegrationTests` (3) and `TabsFanOutDiagnosticTests` pass with zero-fan-out assertions.
- Regression-risk suites green: `ShadowTests` (35), `BorderViewTests` (78), `AdornmentTests` (116),
  `BorderTests` (9), `PaddingTests` (8), `MarginTests` (17), `ViewportTests` (143).
- Full `UnitTestsParallelizable` (17,300) and `UnitTests.NonParallelizable` (72, +2 skipped) pass.

### DIFF
--- a/Terminal.Gui/App/ApplicationImpl.Screen.cs
+++ b/Terminal.Gui/App/ApplicationImpl.Screen.cs
@@ -278,18 +278,16 @@ internal partial class ApplicationImpl
             Rectangle clipRect = Screen with { X = 0, Y = 0 };
             Driver.Clip = new Region (clipRect);
 
-            // Only force a complete redraw if needed (needsLayout or forceRedraw).
-            // Otherwise, just redraw views that need it.
+            // Only force a complete redraw when explicitly requested.
+            // Otherwise, redraw only the views that need it.
             //
             // NOTE (#5358): passing force=true here calls SetNeedsDraw on the top runnable, which
             // cascades to all overlapping subviews via the existing SetNeedsDraw recursion. This
             // is the remaining draw-fan-out source documented by TabsFanOutIntegrationTests. The
             // proper fix requires tracking adornment thickness changes (in addition to Frame
             // changes) so the SuperView can be invalidated precisely instead of force-redrawing
-            // the whole tree. Dropping force-on-neededLayout without that tracking exposes
-            // stale-content bugs in the shrink/move and adornment-rebalance paths (covered by
-            // ShadowTests / BorderViewTests).
-            View.Draw (views.ToArray ().Cast<View> (), neededLayout || forceRedraw);
+            // the whole tree.
+            View.Draw (views.ToArray ().Cast<View> (), forceRedraw);
 
             Driver.Clip = new Region (clipRect);
 
--- a/Terminal.Gui/ViewBase/View.Adornments.cs
+++ b/Terminal.Gui/ViewBase/View.Adornments.cs
@@ -23,7 +23,7 @@ public partial class View // Adornments
         // When any adornment's thickness changes, recompute frames and request layout + redraw.
         Margin.ThicknessChanged += (_, _) =>
                                    {
-                                      previousMarginThickness = HandleAdornmentThicknessChanged (Margin, previousMarginThickness);
+                                       previousMarginThickness = HandleAdornmentThicknessChanged (Margin, previousMarginThickness);
                                    };
 
         Border.ThicknessChanged += (_, _) =>
@@ -370,17 +370,18 @@ public partial class View // Adornments
     {
         Thickness thickness = GetAdornmentsThicknessForAdornment (changedAdornment, changedThickness);
 
-        return new Rectangle (_viewportLocation,
-                              new Size (Math.Max (0, Frame.Width - thickness.Horizontal), Math.Max (0, Frame.Height - thickness.Vertical)));
+        return new Rectangle (_viewportLocation, GetViewportSizeForThickness (thickness));
     }
 
     private Rectangle GetViewportFrameForAdornmentThickness (IAdornment changedAdornment, Thickness changedThickness)
     {
         Thickness thickness = GetAdornmentsThicknessForAdornment (changedAdornment, changedThickness);
 
-        return new Rectangle (new Point (thickness.Left, thickness.Top),
-                              new Size (Math.Max (0, Frame.Width - thickness.Horizontal), Math.Max (0, Frame.Height - thickness.Vertical)));
+        return new Rectangle (new Point (thickness.Left, thickness.Top), GetViewportSizeForThickness (thickness));
     }
+
+    private Size GetViewportSizeForThickness (Thickness thickness) =>
+        new (Math.Max (0, Frame.Width - thickness.Horizontal), Math.Max (0, Frame.Height - thickness.Vertical));
 
     private Thickness GetAdornmentsThicknessForAdornment (IAdornment changedAdornment, Thickness changedThickness)
     {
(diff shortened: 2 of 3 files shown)


tig (approved):
Clean and well-scoped. The approach is correct: track previous adornment thickness in closures, compare old vs new viewport frame on change, and invalidate precisely at the SuperView instead of force-drawing the entire tree. Removing `neededLayout` from the force condition in `LayoutAndDraw` is the payoff, and the integration test flipping from "documents remaining fan-out" to `Assert.Equal(0, inactiveTextDraws)` proves it works.

The closure pattern (local function returning the new thickness, reassigned to the captured variable) is a bit unusual but valid and keeps the event wiring concise. The null-SuperView fallback to `NeedsClearScreenNextIteration()` handles the top-level case correctly.

I did not have time to do a bunch of user testing. If you feel like you've played hard with these Scenarios I'm good: Adornments, Shadow Styles, Menus, Arrangements...

harder:
Thanks! Yeah, the closure pattern is a little unusual. It's written that way to keep the per-adornment previous-thickness tracking local to the wiring instead of adding three backing fields. But happy to refactor to fields if you'd rather have it more conventional.

I tested out Adornments, Shadow Styles, Menus, and Arrangements in the UICatalog and haven't noticed any bugs.

PR #5477, opened 2026-06-09
Fixes #5360. Cull fully occluded overlapped opaque subviews during draw

### LINKED ISSUE #5360: Cull fully occluded overlapped opaque subviews during draw
## Summary

In `DrawSubViews ()`, skip drawing SubViews that are fully obscured by higher-Z opaque peers.

This is an optimization issue and should come only after the invalidation semantics are stable.

Split from #4973.

## Problem

`DrawSubViews ()` currently iterates all visible overlapped SubViews and draws them all. In controls like `Tabs`, many pages occupy the same region, so fully hidden pages can still participate in draw work.

Even after invalidation fixes, this remains avoidable overhead.

## Scope

- Add safe occlusion culling for fully obscured overlapped opaque siblings.
- Limit the first pass to clearly safe cases.
- Do not mix this with unrelated invalidation refactors.

## Acceptance Criteria

- GIVEN overlapped opaque sibling views
  WHEN one sibling is fully covered by higher-Z opaque peers
  THEN the fully covered view is skipped during draw.

- GIVEN a view that is only partially covered
  WHEN draw runs
  THEN the view still draws correctly.

- GIVEN transparent views, shadows, or adornment-driven output
  WHEN draw runs
  THEN no incorrect culling occurs.

- GIVEN `Tabs`
  WHEN inactive tab pages are fully hidden by the active page
  THEN they can be skipped safely when the occlusion rules are satisfied.

## Technical Considerations

- This is the highest-risk optimization in the set.
- The thread explicitly warns about:
  - clipping at `IOutput`
  - transparent rendering
  - margin shadows
  - adornment subviews
- Line-canvas composition and reserved cells must remain correct.
- Define "opaque enough to cull behind" conservatively.

## Edge Cases

- Partially occluded views
- Transparent or semi-transparent rendering behavior
- Shadow margins drawn in separate passes
- Border/Padding/Margin adornment output outside content
- Views whose line canvas affects composed output

- No visual regressions.
- Optimization should be measurable in overlapped scenarios.

## Testing Approach

- Add focused overlap tests with safe fully-occluded cases.
- Add regression cases for partial coverage, transparent views, and shadow/adornment behavior.

## Related

- Parent: #4973

harder:
- Fixes #5360

## Summary

`DrawSubViews()` now skips drawing an Overlapped sibling when it is entirely covered by the higher-Z opaque peers already drawn. A fully-covered sibling produces **no visible output** - every cell it would draw (including its own `RenderLineCanvas`) is clipped away by the clip "holes" those peers punch in `DoDrawComplete` - so skipping its `Draw()` is output-neutral. This is the occlusion-culling step of the #4973 series, taken now that the invalidation semantics are stable (#5357, #5358, #5359, #5433, #5434).

The culled view's `NeedsDraw` is cleared (mirroring a drawn-but-fully-clipped pass) so it does not keep the SuperView perpetually dirty.

## Approach

- Occlusion is tracked with a `Region` of the screen area opaquely covered by higher-Z peers, accumulated as the existing reverse-Z (top-first) draw loop iterates.
- A new `IsOpaqueForOcclusion(out Rectangle)` reports whether a view fills a screen rectangle opaquely. The opaque region is the area **inside the (transparent-by-default) Margin** - `Border.FrameToScreen()` - matching exactly what `DoDrawComplete` excludes from the clip.
- `IsFullyCovered(rect, coverage)` uses `Region` difference (not single-rect `Contains`) so coverage spread across multiple peers is handled correctly.
- The whole check is gated on the presence of an `Overlapped` sibling, so the common (non-overlapping) draw path is **zero-overhead**.

## Why it's conservative (safety guards)

A sibling is culled only when **all** of these hold. Each guard maps to a hazard called out in the issue:

| Guard | Hazard it avoids |
|---|---|
| candidate is `Overlapped` | scope = overlapped siblings only |
| opaque (View/Border/Padding not `Transparent`) | `ViewportSettings.Transparent` - transparent views don't fully cover |
| `ShadowStyle is None` | margin shadows draw in a separate second pass, on top |
| **`!SuperViewRendersLineCanvas`** | line-canvas composition / reserved cells - **this is what keeps `Tabs` correct**: tab pages render their borders (and headers) through the SuperView's painters'-algorithm LineCanvas composition, so they must never be culled |
| full frame ⊆ accumulated opaque coverage | partial coverage still draws |

`Tabs` pages are `SuperViewRendersLineCanvas`, so they are deliberately **not** culled - preventing any header-line-art regression while still benefiting plain overlapped opaque siblings (overlapped Windows/panels, Arrangement-style stacks).

## Tests

- New `Tests/UnitTestsParallelizable/ViewBase/Draw/OcclusionCullingTests.cs` (11 tests, all green): fully-occluded cull, `NeedsDraw` clearing, partial coverage, transparent occluder, transparent candidate, shadowed candidate, `SuperViewRendersLineCanvas` candidate, non-Overlapped candidate, the no-Overlapped-siblings gate, output-neutrality, and a `Tabs` regression asserting inactive pages are never culled.
- Verification uses the headless in-process driver and `DrawComplete` event counting (which fires iff `Draw()` was invoked), not just `Driver.Contents`.
- Full suite green: 17,370 parallelizable + 74 non-parallelizable tests pass. No new build warnings.

## Visual & output-neutrality testing (tuirec + PTY)

Beyond the unit tests, I verified at the real-terminal layer (addressing the `Driver.Contents` ≠ IOutput caveat) using `Scripts/tuirec` recordings and `tmux capture-pane` frame diffs on the scenarios most likely affected:

- **Tabs Example** - overlapped `SuperViewRendersLineCanvas` pages (the guarded case)
- **Arrangement** - overlapped opaque views (where culling actually fires)
- **WideGlyphs** - overlapped views + clipping

**Method:** built the app **with** and **without** the change, drove each scenario identically under a PTY, and diffed the settled screen grids. A control (capturing the *same* build twice) isolates the scenarios' own non-determinism.

| Scenario | before (no cull) vs after (cull) | same build, run twice (control) | Result |
|---|---|---|---|
| Tabs | **0 differing lines** | 0 | byte-for-byte identical |
| Arrangement | 2 differing lines | 2 | diff == its animated progress bar only |
| WideGlyphs | 60 differing lines | 60 | diff == its random wide-glyph fill only |

Each before/after difference **exactly equals that scenario's own run-to-run variance** - culling adds nothing - and the regression-risk case (Tabs) is identical. Recordings:
tabs-after
arrangement-after
wideglyphs-after

```
git fetch copilot fix-5360-cull-occluded-subviews
git checkout copilot/fix-5360-cull-occluded-subviews
```

### DIFF
--- a/Terminal.Gui/ViewBase/View.Drawing.cs
+++ b/Terminal.Gui/ViewBase/View.Drawing.cs
@@ -660,9 +660,45 @@ public partial class View // Drawing APIs
         // SubViews earlier in the collection are drawn last (on top).
         // NOTE: Do not use SubViews or GetSubViews() here as GetSubViews can be overridden to return a different set of views or ordering.
         // NOTE: We need to draw exactly the views in InternalSubViews.
-        foreach (View view in InternalSubViews.Snapshot ().Where (v => v.Visible).Reverse ())
+        View [] drawOrder = InternalSubViews.Snapshot ().Where (v => v.Visible).Reverse ().ToArray ();
+
+        // Occlusion culling (issue #5360): when Overlapped opaque siblings stack (e.g. Tabs pages,
+        // overlapped Windows), a lower-Z sibling that is fully covered by the higher-Z opaque peers
+        // already drawn produces no visible output — every cell it would draw, including its own
+        // RenderLineCanvas, is clipped away by the clip "holes" those peers punched in DoDrawComplete.
+        // Skipping such a sibling's Draw is therefore output-neutral. opaqueCoverage accumulates the
+        // screen region opaquely covered by higher-Z peers (iterated highest-Z first). The whole
+        // check is gated on at least one Overlapped sibling so the common (non-overlapping) path is
+        // zero-overhead.
+        bool considerOcclusion = Driver is { } && drawOrder.Length > 1 && drawOrder.Any (v => v.Arrangement.FastHasFlags (ViewArrangement.Overlapped));
+        Region? opaqueCoverage = null;
+
+        foreach (View view in drawOrder)
         {
+            Rectangle coveredScreenRect = Rectangle.Empty;
+            bool opaque = considerOcclusion && view.IsOpaqueForOcclusion (out coveredScreenRect);
+
+            if (opaque
+                && view.Arrangement.FastHasFlags (ViewArrangement.Overlapped)
+                && !view.SuperViewRendersLineCanvas
+                && view.ShadowStyle is null or ShadowStyles.None
+                && IsFullyCovered (view.FrameToScreen (), opaqueCoverage))
+            {
+                // Fully occluded by higher-Z opaque peers. Clear the draw flags to mirror a
+                // drawn-but-fully-clipped pass (Draw() would have called ClearNeedsDraw at its end);
+                // otherwise this view would stay dirty forever and keep the SuperView redrawing.
+                view.ClearNeedsDraw ();
+
+                continue;
+            }
+
             view.Draw (context);
+
+            if (opaque)
+            {
+                opaqueCoverage ??= new ();
+                opaqueCoverage.Union (coveredScreenRect);
+            }
 
             if (!view.SuperViewRendersLineCanvas)
             {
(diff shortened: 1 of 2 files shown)


tig:
@harder what'd you think of the tuirec experience?

tig (approved):
Lovely, as usual.

harder:
> @harder what'd you think of the tuirec experience?

@tig  I like how easy it is to automate making screenshot videos now! I need to play with it more. I remember reading about it, but hadn't spent time trying it out yet until I was asking my agent to propose additional ways I could test and verify that this change didn't break any UI. It suggested using tuirec and proposed those 3 areas, then wired it all up. Very cool.

tig:
> > @harder what'd you think of the tuirec experience?
> 
> @tig I like how easy it is to automate making screenshot videos now! I need to play with it more. I remember reading about it, but hadn't spent time trying it out yet until I was asking my agent to propose additional ways I could test and verify that this change didn't break any UI. It suggested using tuirec and proposed those 3 areas, then wired it all up. Very cool.

Also, see the new-ish test infra that lets you provide Goldens as .ans files and then IDriver.StringToAnsi.

PR #5471, opened 2026-06-05
Regenerate docfx Mandelbrot.gif with rendered sixel

tig:
Recorded the UICatalog Mandelbrot scenario with tuirec (DA1 + window/cell-size handshake) and rendered it with a sixel-aware agg, so the GIF now shows the sixel Mandelbrot sized to its view. Leading blank alternate-screen frames are trimmed so the first frame shows content.

## Fixes

- Fixes #_____

## Proposed Changes/Todos

- [ ] Todo 1

## Pull Request checklist:

- [ ] I've named my PR in the form of "Fixes #issue. Terse description."
- [ ] My code follows the [style guidelines of Terminal.Gui](https://github.com/gui-cs/Terminal.Gui/blob/develop/.editorconfig) - if you use Visual Studio, hit `CTRL-K-D` to automatically reformat your files before committing.
- [ ] My code follows the [Terminal.Gui library design guidelines](https://github.com/gui-cs/Terminal.Gui/blob/develop/CONTRIBUTING.md)
- [ ] I ran `dotnet test` before commit
- [ ] I have made corresponding changes to the API documentation (using `///` style comments)
- [ ] My changes generate no new warnings
- [ ] I have checked my code and corrected any poor grammar or misspellings
- [ ] I conducted basic QA to assure all features are working

PR #5482, opened 2026-06-10
Auto-starts DOOM fire sixel animation in UICatalog About Box

tig:
## Summary

The UICatalog About Box now auto-starts the DOOM fire sixel animation at the bottom of the dialog when the terminal supports sixel — reusing `DoomFire`/`ConstPalette` from the Images scenario with the same encoder setup (`AvoidBottomScroll`, palette clamped to `MaxPaletteColors`, 30ms timer, every-2nd-frame encode).

## Behavior

- Starts automatically when `App.Driver.SixelSupport.IsSupported` is true; silently skipped otherwise — the About Box is unchanged on non-sixel terminals.
- The fire is sized to the dialog frame at first layout and bottom-aligned via `FrameToScreen ()`, recomputed every frame so it follows the (movable) dialog.
- On close, the timer stops and the `SixelToRender` is removed from the driver queue (rebuild — `ConcurrentQueue` has no Remove) so the last frame does not linger.
- `UICATALOG_FORCE_ABOUT_FIRE=1` forces the fire where the PTY cannot advertise sixel — needed for recording/verification through Windows ConPTY, which strips sixel DCS.

## Verification (tuirec, per Scripts/tuirec/README.md)

Recorded UICatalog with tuirec v0.8.1 and read the `.cast` back:

1. **Non-sixel path** (ConPTY does not advertise sixel): About opens via Help→About, renders normally, closes cleanly, app quits — no fire, no crash. Driver log confirms `Sixel support: False` and the guard short-circuits.
2. **Fire path** (`--env UICATALOG_FORCE_ABOUT_FIRE=1`): cast event timeline shows a steady ~21 output events/sec for the full 5s the dialog is open (the animation encode/redraw loop), dropping to zero immediately after Esc — confirming the timer stops and cleanup runs. 0 exceptions.

The flame pixels themselves cannot be captured on Windows (ConPTY strips sixel DCS — see `tuirec agent-guide`); a hero GIF needs a Linux/macOS recording, same as the Mandelbrot GIF in #5471.


tig:
this was just a test. closing.

PR #5519, opened 2026-06-22
Fix #5518: Padding adornment fills with stale attribute when parent skips ClearViewport (#5358)

### LINKED ISSUE #5518: ImageView (Sixel): on redraw, the column at Frame.X-1 is overdrawn with the image background
## Describe the bug

An `ImageView` using raster graphics (**Sixel**) overdraws the column **immediately to its left** — `ImageView.Frame.X - 1` — with the image's background color, but **only on a subsequent draw**, not the first. The initial render is correct; after the `ImageView` is redrawn (e.g. it loses then regains focus, or a sibling/SuperView triggers a re-render), the image's gray background bleeds **one column left** into the adjacent sibling/SuperView content.

> I'm fairly sure I've seen/fixed this before — filing because it's back (possible regression).

## To Reproduce

Repro'd in the winprint TUI (`wp`), which hosts an `ImageView` (raster/Sixel) as the page preview, with a settings panel to its left:

1. `WP_FORCE_SIXEL=1 wp tui testfiles/Program.cs` (drives the Sixel encode path).
   - Layout: left settings panel at `X = 0`; preview `ImageView` at `X = Pos.Right(panel)` → its `Frame.X` lands at screen **column 43**. So `Frame.X - 1` = **column 42**, which belongs to the settings panel (dark).
2. **Initial render:** column 42 is the settings panel background (dark). ✅
3. Move focus to **any** control in the left settings panel, cause the preview to render again, then move focus **back** to the preview (`ImageView`) so it redraws.
4. **Expected:** column 42 stays the settings-panel background; the `ImageView`'s gray canvas begins at `Frame.X` (column 43).
5. **Actual:** column 42 (`Frame.X - 1`) is overdrawn with the `ImageView` background **gray**, bleeding one column into the panel.

### Pixel evidence (captured via [tuirec](https://github.com/gui-cs/tuirec) + agg/Sixel)

Sampling the same screen column at mid-height, before vs. after the re-render:

| screen column | role | initial | after re-render |
|---|---|---|---|
| 41 | settings panel | `(26,26,26)` dark | `(26,26,26)` dark ✅ |
| **42** | **`ImageView.Frame.X - 1`** | `(26,26,26)` dark | **`(224,224,224)` gray ✗** |
| 43 | `ImageView.Frame.X` (canvas) | `(224,224,224)` gray | `(224,224,224)` gray ✅ |

Only column 42 changes — from the panel's dark to the image's gray — confirming a **1-column left overdraw** that appears only on redraw. (Before/after screenshots available; I'll attach them.)

## Likely cause

Looks like the Sixel emit on **redraw** starts one column too far left — either the Sixel placement origin (cursor position before the DCS) is off by one, or a pre-clear/erase rectangle uses `Frame.X - 1` — so the first paint is correct but a later repaint clears/fills the extra left column. Worth checking the `ImageView` raster draw path and the Sixel cursor-positioning on re-emit specifically (the first emit doesn't hit it).

## Minimal repro shape (not yet reduced to a standalone sample)

```csharp
var left = new View { X = 0, Y = 0, Width = 10, Height = Dim.Fill() };           // dark sibling
var img  = new ImageView { X = Pos.Right(left), Y = 0, Width = Dim.Fill(), Height = Dim.Fill(), UseRasterGraphics = true };
img.Image = /* a bitmap with a light/gray background */;
top.Add(left, img);
// 1) Run -> first draw: column (img.Frame.X - 1) shows `left`. OK.
// 2) Force img to redraw (move focus left sibling -> back, or img.SetNeedsDraw()).
// 3) Column (img.Frame.X - 1) is now overdrawn with img's background.
```

## Environment

- **Terminal.Gui:** 2.4.7
- **Renderer:** Sixel (agg, via tuirec) — **confirmed on Sixel**; Kitty path not verified, may share the bug.
- **OS:** macOS (Darwin 25.5.0)
- **.NET:** 10

tig:
Fixes #5518.

## Symptom
A view with a right `Padding` next to a raster `ImageView` (winprint's settings panel ↔ page preview) shows the image pane's background bleeding into the padding "seam" column after a partial (focus) redraw — on **both Sixel and Kitty**, so not encoder-specific. In fact it isn't ImageView-specific at all: a plain `View` reproduces it.

## Root cause
`Thickness.Draw` fills an adornment via `driver.CurrentAttribute` and **never sets an attribute itself** — it relies on the caller having set one. The parent's `ClearViewport` → `SetAttributeForRole` used to set `CurrentAttribute` to the parent's background right before adornments drew.

Since **#5358 (`584c02526`)** the parent **skips `ClearViewport` on a child-only redraw** (the intended fan-out optimization). So on a partial redraw the `Padding` adornment inherits a **stale `CurrentAttribute`** — `Color.None` left behind by a sibling raster `ImageView`'s transparent-marking (or the sibling's background for a plain `View`) — and fills the padding column with it. Driver-level tracing pinned the writer exactly:

```
WRITE r2c9 attr=[None,None,None] via:
   OutputBufferImpl.FillRect
   Thickness.Draw          ← the neighbour's Padding adornment
   View.DrawAdornments
```

(Bisect: passes at `584c02526^`, fails from `584c02526` on.)

## Fix
Set the view's `Normal` attribute before the no-subview `Padding.Thickness.Draw`, so the fill is deterministic regardless of prior draw state. It does **not** touch the #5358 clear optimization, so the fan-out tests keep passing.

The same latent pattern exists for `Margin`/`Border` `Thickness.Draw` (they also fill with the ambient attribute); I scoped this PR to the demonstrated `Padding` case and left those, since their "correct" attribute differs (border scheme, transparent margin) and no repro exercises them.

## Validation
- New regression test `ImageViewNeighbourPaddingTests` (fails before, passes after).
- **UnitTestsParallelizable: 17449 / 0 failed.** UnitTests.NonParallelizable: 74 / 0 failed. The fan-out / clipping tests that the naive "always clear" approach broke (`SubViewOnlyRedrawTests`, `TabsFanOutDiagnosticTests`, `ViewDrawingClippingTests`, `BorderArrangementTests`) all stay green.

### DIFF
--- a/Terminal.Gui/ViewBase/View.Drawing.Adornments.cs
+++ b/Terminal.Gui/ViewBase/View.Drawing.Adornments.cs
@@ -218,6 +218,12 @@ public partial class View
             }
             else if (Padding.Thickness != Thickness.Empty)
             {
+                // Thickness.Draw fills via the driver's CurrentAttribute and does not set one itself, so
+                // set the view's Normal here. Otherwise the padding inherits whatever attribute was last
+                // used — which is stale (e.g. a sibling raster view's Color.None) when the parent skipped
+                // its ClearViewport on a child-only redraw, bleeding foreign content into the padding
+                // column. (gui-cs/Terminal.Gui#5518.)
+                SetAttributeForRole (VisualRole.Normal);
                 Padding.Thickness.Draw (Driver, Padding.FrameToScreen (), Padding.Diagnostics);
                 Padding.LastDrawnRegion = null;
             }
--- a/Tests/UnitTestsParallelizable/Views/ImageViewNeighbourPaddingTests.cs
+++ b/Tests/UnitTestsParallelizable/Views/ImageViewNeighbourPaddingTests.cs
@@ -1,0 +1,86 @@
+namespace ViewsTests;
+
+/// <summary>
+///     Regression for gui-cs/Terminal.Gui#5518.
+///     <para>
+///         A raster <see cref="ImageView"/> placed at <see cref="Pos.Right"/> of a view that has a right
+///         <see cref="Adornment"/> padding column overwrites that padding column — the cell one to the
+///         LEFT of the ImageView's frame — when it redraws. The full first draw repaints the neighbour so
+///         it looks correct; a <em>partial</em> redraw (a focus round-trip here) leaves the overwrite
+///         behind, bleeding the image pane's background into the neighbour.
+///     </para>
+///     <para>
+///         Surfaced in winprint (gui-cs/Terminal.Gui#5518) as the page-preview pane bleeding its canvas
+///         colour into the settings panel's seam, on both Sixel and Kitty. It is not actually
+///         raster-specific — a plain <see cref="View"/> at <see cref="Pos.Right"/> reproduces it too — so
+///         the fix most likely belongs in the View draw/clip path, not the raster code.
+///     </para>
+/// </summary>
+public class ImageViewNeighbourPaddingTests
+{
+    [Fact]
+    public void Draw_RightOfPaddedNeighbour_DoesNotOverwriteNeighboursPaddingColumn ()
+    {
+        using IApplication app = Application.Create ();
+        app.Init (DriverRegistry.Names.ANSI);
+        app.Driver!.SetScreenSize (40, 12);
+
+        Runnable runnable = new () { Width = 40, Height = 12 };
+        runnable.SetScheme (new Scheme { Normal = new Attribute (new Color (255, 255, 255), new Color (26, 26, 26)) });
+        app.Begin (runnable);
+
+        DriverImpl driver = (DriverImpl)app.Driver!;
+        driver.SetSixelSupport (new SixelSupportResult { IsSupported = true, Resolution = new Size (9, 20), MaxPaletteColors = 256 });
+
+        // Left pane with a one-column right Padding — that padding column is the "seam".
+        View left = new () { X = 0, Y = 0, Width = 10, Height = Dim.Fill (), CanFocus = true };
+        left.Padding!.Thickness = new Thickness (0, 0, 1, 0);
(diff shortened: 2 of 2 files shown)


tig:
## Root cause: regression bisected to #5358 (`584c02526`)

`git bisect` pins the failure to **`584c02526` "Gate self-content draws on needsDrawSelf to stop child-only redraw fan-out" (#5358)**. The test **passes at its parent** (`584c02526^`) and **fails from that commit onward**.

### Mechanism

The neighbour pane has a **transparent right `Padding`** adornment (the "seam" column). With instrumentation:

- **Before #5358:** on the focus-driven partial redraw, the parent (`Runnable`) re-cleared its whole viewport every frame, so the seam column was re-established as the parent's background. The transparent padding composited against that fresh background → correct.
- **After #5358:** the parent enters `Draw()` only via `SubViewNeedsDraw` (`needsDrawSelf == false`), so `DoClearViewport` is skipped (the intended optimization). The neighbour's **transparent padding column now composites against stale backing** — it shows the image pane's content (`Color.None` raster-transparent for a Sixel/Kitty `ImageView`, or the pane's solid background for a plain `View`) instead of the parent's background.

So the seam cell goes from the window background to the image pane's background on the partial redraw — the bug winprint reported.

### Fix is non-trivial (don't just revert the gate)

The naive fix — always `DoClearViewport` — makes this test pass but **breaks the #5358 contract** (`SubViewOnlyRedrawTests.ChildOnlyDirty_ParentDoesNotClearOrRedrawSelf` and `ParentWithBorder_ChildOnlyDirty_ParentDoesNotClear` both assert the parent must **not** clear at all on a child-only redraw; 3 draw tests fail). The correct fix has to preserve the optimization while ensuring a transparent adornment/`ViewportSettings.Transparent` region composites against the parent's *current* background after a partial redraw — i.e. the fix belongs in the transparency-compositing / clip-exclusion path, not in re-enabling the parent clear.

Repro test in this PR stays as the red guard.

BDisp (approved):

PR #5549, opened 2026-06-27
Fixes #5499. Skip SetNeedsLayout when Dim.Auto text recompute yields unchanged size

### LINKED ISSUE #5499: Skip SetNeedsLayout when a Dim.Auto recompute yields an unchanged size
Split off from #4522 (the `NeedsLayout` lifecycle cleanup). Tracked separately because it's an independent optimization with its own redraw-correctness risk profile, not part of the lifecycle refactor.

## Problem

Raised by @tznind in #4522: changing the text of a `Dim.Auto`-sized view (e.g. a frequently-updating `Label` or a clock) calls `SetNeedsLayout()` unconditionally — even when the recomputed `Dim.Auto` dimensions are identical to the current `Frame`. That marks the view (and its ancestors) as needing layout and triggers a redraw every update, regardless of whether anything visually moved or resized.

The #5357/#5358/#5359 work already removed the *sibling* fan-out and made redraw region-aware, so this is no longer catastrophic. But the per-update layout/redraw still happens for content changes that don't actually change size.

## Proposed optimization

When a `Dim.Auto` recompute (or, more generally, a content/text change feeding into layout) produces the **same** resulting size, skip `SetNeedsLayout()` (or short-circuit so it doesn't propagate to ancestors). Only mark layout-needed when the computed `Frame` would actually differ.

## Risk / care

- Must not skip a redraw when the content changed but the size didn't — text content change still needs `SetNeedsDraw()` even if no `SetNeedsLayout()`. Keep the two signals distinct.
- `Dim.Auto` width/height interact with wrapping and `TextFormatter` constraints; "same size" must be evaluated after the constraint recompute, not before.
- Needs a diagnostic/test (similar to the `TabsFanOutDiagnosticTests` event-count approach) proving that an identical-size text update produces zero layout passes but still redraws.

## Relationship to #4522

Independent of the deterministic-lifecycle refactor; can land before or after it. Cross-linked for context.

harder:
Fixes #5499.

## Problem

The `View.Text` setter unconditionally called `SetNeedsLayout()`. That marks the view and its whole subtree as needing layout, sets `_needsDrawAfterLayout`, and **propagates up the entire ancestor chain** on every text change, even when the recomputed frame is identical to the current one (e.g. a once-per-second clock `Label`).

The layout machinery already no-ops the *frame mutation* when nothing changed (`SetRelativeLayout` only calls `SetFrame` when `Frame != newFrame`), but the layout pass, formatter setup, redraw scheduling, and ancestor invalidation were still being scheduled every tick.

### `Terminal.Gui/ViewBase/View.Layout.cs`

Extracted `SetRelativeLayout`'s `Pos`/`Dim` -> absolute-`Frame` resolution into a private `TryComputeRelativeFrame (Size, out Rectangle)`. `SetRelativeLayout` now calls it, and so does the `Text` setter's speculative check. This keeps frame prediction and real layout on the same computation path.

Also extracted `SetRelativeLayout`'s post-frame `TextFormatter` constraint cleanup into `FinalizeTextFormatterConstraints()`. This matters because `UpdateTextFormatterText()` clears `ConstrainToWidth`/`ConstrainToHeight`; if the text setter skips layout, the fast path must still restore the same formatter constraints a full layout pass would produce so wrapped/clipped text formats correctly.

### `Terminal.Gui/ViewBase/View.Text.cs`

The setter now calls `TryRedrawWithoutLayout()` after updating `TextFormatter.Text`.

When the view has already been laid out and **both** `Width` and `Height` are eligible, `TryRedrawWithoutLayout()` resolves the would-be `Frame` via `TryComputeRelativeFrame`. If the predicted frame equals the current `Frame`, it skips `SetNeedsLayout()` and ancestor propagation, then:

- finalizes `TextFormatter` constraints with `FinalizeTextFormatterConstraints()`;
- sets `TextFormatter.NeedsFormat = true`;
- calls `SetNeedsDraw()`;
- lets the setter raise `TextChanged` normally.

The comparison is the **full `Rectangle`** (position and size), not just dimensions. A `Pos` can depend on `Text` (e.g. `Pos.Func` reading `view.Text`), so a same-size text change can still move the view and must run layout.

Eligibility is exact by design:

```csharp
static bool IsEagerlyResolvable (Dim dim) => dim is DimAbsolute or DimAuto { Style: DimAutoStyle.Text };
```

- `DimAbsolute` is constant; a Text-only `DimAuto` depends on this view's `TextFormatter`.
- `DimAutoStyle.Content` / `DimAutoStyle.Auto` can lay out subviews while calculating, so those fall back to `SetNeedsLayout()`.
- Composite/nested dims such as `Dim.Auto(Text) + 2` are not bare `DimAbsolute`/`DimAuto`, so they also fall back.

This covers the intended safe cases: both-axis Text-auto, fixed-size text redraws, and one-axis-auto (`Width = Auto(Text), Height = 1`). The optimization only avoids *setting* `NeedsLayout`; it never clears a pre-existing pending layout request.

## Tests

Added focused layout and render coverage:

- `Tests/UnitTestsParallelizable/ViewBase/Layout/Dim.AutoTests.TextOptimization.cs`
  - ancestor diagnostic: same-size text change leaves `superView.NeedsLayout == false` while the view still redraws;
  - full-frame guard: `Pos.Func` depending on `Text` forces layout when same-size text moves the view;
  - exact guard: composite dims bypass the fast path;
  - fixed-size and one-axis-auto fast paths;
  - formatter constraint survival for fixed-size and one-axis-auto word-wrap cases;
  - clock-style repeated same-width updates with no frame drift or ancestor propagation;
  - reentrant `TextChanged` handling;
  - min/max, border adornment, uninitialized, `DimAutoStyle.Auto`, different-size, and pending-layout cases.

- `Tests/UnitTestsParallelizable/ViewBase/Layout/TextFastPathRenderTests.cs`
  - compares the isolated redraw-only fast path against a forced full-layout control;
  - verifies rendered driver contents match for fixed-size wrap, fixed-size clip, single-row clipping, and max-constrained Text-auto wrapping.

Additional exploratory probes were run and removed after conversion of the high-value cases into permanent tests. They compared fast-path vs full-layout behavior across repeated text updates, fixed/min/max/Text-auto dimensions, word-wrap on/off, explicit content-size constraints, `Button`, `CheckBox`, and reentrant handlers.

## Verification

Current local verification on `ae2047abec598989a10bea28d858e01e6b6b2f5f`:

- Build `Tests/UnitTestsParallelizable`: passed
- `DimAutoTests` text-focused tests: 67 passed
- `TextFastPathRenderTests`: 5 passed
- `SetRelativeLayoutTests`: 10 passed
- Full `UnitTestsParallelizable`: 17,484 total / 0 failed / 17 skipped
- Full `UnitTests.NonParallelizable`: 74 total / 0 failed / 2 skipped

## Pull Request checklist

- [x] My code follows the style guidelines of Terminal.Gui
- [x] My code follows the Terminal.Gui library design guidelines

### DIFF
--- a/Terminal.Gui/ViewBase/View.Layout.cs
+++ b/Terminal.Gui/ViewBase/View.Layout.cs
@@ -665,53 +665,11 @@ public partial class View // Layout APIs
         // TODO: Should move to View.LayoutSubViews?
         SetTextFormatterSize ();
 
-        int newX, newW, newY, newH;
-
-        try
+        if (!TryComputeRelativeFrame (superviewContentSize, out Rectangle newFrame))
         {
-            // Calculate the new X, Y, Width, and Height
-            // If the Width or Height is Dim.Auto, calculate the Width or Height first. Otherwise, calculate the X or Y first.
-            if (_width.Has<DimAuto> (out _))
-            {
-                newW = _width.Calculate (0, superviewContentSize.Width, this, Dimension.Width);
-                newX = _x.Calculate (superviewContentSize.Width, newW, this, Dimension.Width);
-
-                if (newW != Frame.Width)
-                {
-                    // Pos.Calculate got us a new position. We need to redo dimension
-                    newW = _width.Calculate (newX, superviewContentSize.Width, this, Dimension.Width);
-                }
-            }
-            else
-            {
-                newX = _x.Calculate (superviewContentSize.Width, _width, this, Dimension.Width);
-                newW = _width.Calculate (newX, superviewContentSize.Width, this, Dimension.Width);
-            }
-
-            if (_height.Has<DimAuto> (out _))
-            {
-                newH = _height.Calculate (0, superviewContentSize.Height, this, Dimension.Height);
-                newY = _y.Calculate (superviewContentSize.Height, newH, this, Dimension.Height);
-
-                if (newH != Frame.Height)
-                {
-                    // Pos.Calculate got us a new position. We need to redo dimension
-                    newH = _height.Calculate (newY, superviewContentSize.Height, this, Dimension.Height);
-                }
-            }
-            else
-            {
-                newY = _y.Calculate (superviewContentSize.Height, _height, this, Dimension.Height);
-                newH = _height.Calculate (newY, superviewContentSize.Height, this, Dimension.Height);
-            }
-        }
-        catch (LayoutException)
-        {
-            //Debug.WriteLine ($"A Dim/PosFunc function threw (typically this is because a dependent View was not laid out)\n{le}.");
+            //Debug.WriteLine ($"A Dim/PosFunc function threw (typically this is because a dependent View was not laid out)");
             return false;
         }
-
-        Rectangle newFrame = new (newX, newY, newW, newH);
 
         if (Frame != newFrame)
         {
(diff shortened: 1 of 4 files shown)


harder:
## P1 (formatter constraints lost on the widened fast path) — resolved in ae2047a

**Confirmed the bug** with a probe: a `View { Width = 5, Height = 2, WordWrap = true }` had constraints `5×2` after layout, but after a fast-path `Text` change `ConstrainToWidth`/`ConstrainToHeight` were both `null` while `NeedsLayout == false`. Since `GetLines()` treats a null width as `int.MaxValue`, wrapped/clipped fixed-size text could render incorrectly until some later layout restored the constraints.

**Root cause:** `UpdateTextFormatterText()` clears the constraints, and the widened fast path then skipped `SetRelativeLayout()` — including its post-resolve cleanup that restores them from `GetContentWidth()`/`GetContentHeight()`.

**Fix:**
- Extracted that cleanup into a shared `FinalizeTextFormatterConstraints()`, called by both `SetRelativeLayout()` and the fast path, so the fast path leaves `TextFormatter` in exactly the state a full layout pass produces.
- Renamed the helper to `TryRedrawWithoutLayout()` (it now owns the redraw-only handling: finalize constraints → mark `NeedsFormat` → `SetNeedsDraw`).

## Testing

Beyond the requested constraint-survival coverage (fixed-size and one-axis-auto, with `WordWrap`), I ran a deeper probe pass with temporary harnesses (since removed):

- **Render-equivalence sweep** comparing the *isolated* fast path (laid out only when the setter requests it — no masking `Layout()`) against a full-layout control on `Frame`, both `TextFormatter` constraints, **and rendered driver contents**, across width dims (Text-auto / min / max / fixed) × height dims (Text-auto / fixed) × `WordWrap` × a multi-step text sequence — all identical. Verified non-vacuous (removing the finalization makes the fixed-dim cases diverge).
- Also probed `CheckBox` (overrides `UpdateTextFormatterText`, now called eagerly — idempotent), `Pos.Align` sibling groups, fixed parents with subviews, reentrant `TextChanged`, and a 100-tick clock simulation (zero frame drift, no ancestor propagation). No unintended consequences.

The high-value probes were kept as permanent tests: `TextFastPathRenderTests` (render-equivalence for fixed-size wrap/clip and max-constrained auto wrapping), plus clock-drift, reentrancy, and constraint-survival cases.

**Results:** `UnitTestsParallelizable` 17467 passed / 0 failed; `UnitTests.NonParallelizable` 72 passed / 0 failed; no new warnings.

tig (approved):
Very nice.

PR #5556, opened 2026-06-28
Fixes #5555. Preserve explicit Scheme role reassignment

### LINKED ISSUE #5555: Copying a Scheme and reassigning an identical role removes its explicit value
## Describe the bug

When a `Scheme` is copied and an explicitly set role is reassigned the same `Attribute`, the copied scheme loses that explicit value. The role is then derived from `Normal`, producing an unexpected color.

This affects roles using `SetAttributeForRoleProperty`, including `Editable`.

## To Reproduce

```csharp
using Terminal.Gui.Drawing;

var editable = new Attribute(
    new Color("#00FF6A"),
    new Color("#292E42"));

var baseScheme = new Scheme
{
    Normal = new Attribute(
        new Color("#C0CAF5"),
        new Color("#1A1B26")),
    Editable = editable
};

var dialogScheme = new Scheme(baseScheme)
{
    Normal = new Attribute(
        new Color("#C0CAF5"),
        new Color("#16161E")),
    Editable = editable
};

Console.WriteLine(baseScheme.Editable.Background);
Console.WriteLine(dialogScheme.Editable.Background);
```

Expected output:

```text
#292E42
#292E42
```

Actual output with Terminal.Gui 2.4.12:

```text
#292E42
#19319F
```

## Cause

The copy constructor preserves `_editable`. During the object initializer, `SetAttributeForRoleProperty` compares the assigned value with the currently resolved value:

```csharp
if (GetAttributeForRoleCore(role, [], null) == value)
{
    return null;
}
```

Because the assigned value equals the copied explicit value, the initializer replaces `_editable` with `null`. `Editable` is subsequently derived from the new `Normal` attribute.

Reassigning an identical value should be idempotent and should not erase an explicit value inherited through the copy constructor.

## Workaround

Do not redundantly reassign the role:

```csharp
var dialogScheme = new Scheme(baseScheme)
{
    Normal = dialogNormal
};
```

## Environment

- Terminal.Gui: 2.4.12
- .NET SDK: 10.0.204
- PowerShell: 7.6.3
- OS: Windows

The behavior is in `Scheme`'s value-resolution logic and appears platform-independent.

YourRobotOverlord:
- Fixes #5555

## Description

Preserves an explicitly assigned Scheme role when a copied scheme
reassigns that role to the same Attribute.

Previously, SetAttributeForRoleProperty compared the assignment with the
resolved role value and converted a matching value to null. For copied
schemes, this erased the copied explicit value and caused the role to be
derived from Normal.

## Tests

Added a regression test covering an identical Editable reassignment after
copying and changing Normal.

- Full parallelizable suite: 17,453 passed, 17 skipped
- Scheme tests: 46 passed
- git diff --check: passed

```bash
git fetch copilot codex/fix-scheme-explicit-role-reassignment
git checkout copilot/codex/fix-scheme-explicit-role-reassignment
```

### DIFF
--- a/Terminal.Gui/Drawing/Scheme.cs
+++ b/Terminal.Gui/Drawing/Scheme.cs
@@ -545,6 +545,12 @@ public record Scheme : IEqualityOperators<Scheme, Scheme, bool>
     // Helper method for property _set implementation
     private Attribute? SetAttributeForRoleProperty (Attribute value, VisualRole role)
     {
+        // Reassigning an explicit value must not change the role to implicit.
+        if (TryGetExplicitlySetAttributeForRole (role, out Attribute? explicitValue) && explicitValue == value)
+        {
+            return value;
+        }
+
         // If value is the same as the algorithm value, use null
         if (GetAttributeForRoleCore (role, [], null) == value)
         {
--- a/Tests/UnitTestsParallelizable/Drawing/SchemeTests.cs
+++ b/Tests/UnitTestsParallelizable/Drawing/SchemeTests.cs
@@ -232,6 +232,28 @@ public class SchemeTests
     [Fact]
     public void CopyConstructor_Null_Throws () => Assert.Throws<ArgumentNullException> (() => new Scheme (null));
 
+    // Codex - GPT-5
+    [Fact]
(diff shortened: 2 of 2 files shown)


tig (approved):
LGTM. Good catch.

PR #5577, opened 2026-07-04
Kh/5498 width height event contract

### LINKED ISSUE #5498: Frame/layout-driven size changes bypass Width/Height change events and TextFormatter reset
Split off from #4522 (the `NeedsLayout` lifecycle cleanup). This is tracked separately because it's an event-semantics / breaking-change-policy decision in a different subsystem than the layout flag, and shouldn't block the lifecycle refactor.

## Problem

When a view's size changes via the `Frame` setter or via a layout pass, the `Width`/`Height` `Changing`/`Changed` events do **not** fire, and `TextFormatter.ConstrainToWidth`/`ConstrainToHeight` are not reset.

Both paths set the backing fields directly to avoid recursion, bypassing the `Width`/`Height` property setters (and thus the CWP event plumbing):

- `View.Layout.cs` — `Frame` setter:
  ```csharp
  // BUGBUG: ... other logic in the property setters does not get executed. Specifically:
  // BUGBUG: - Reset TextFormatter
  // BUGBUG: - If we add property change events for X/Y/Width/Height they will not be invoked
  _x = _frame\!.Value.X;
  _y = _frame\!.Value.Y;
  _width = _frame\!.Value.Width;
  _height = _frame\!.Value.Height;
  ```
- `View.Layout.cs` — `SetRelativeLayout` has the same direct-field assignment with the same BUGBUG block.

When #4522 was filed this was hypothetical ("if we add property change events…"). It is now **real**: `WidthChanging`/`WidthChanged` and `HeightChanging`/`HeightChanged` (plus `OnWidthChanging`/`OnWidthChanged`/`OnHeightChanging`/`OnHeightChanged`) exist today via `CWPPropertyHelper.ChangeProperty`. So a layout-driven size change silently skips them.

## Consequences

- `WidthChanging`/`HeightChanging` cannot intercept/cancel a layout-driven size change.
- `WidthChanged`/`HeightChanged` don't fire, so handlers / data-binding don't observe size changes that came from layout.
- Subclass `OnWidth*`/`OnHeight*` overrides are bypassed.
- `TextFormatter.ConstrainTo*` may retain a stale value.

## Decision needed

1. **Raise the events** (and reset `TextFormatter`) from the Frame/layout paths — manually invoke `OnWidthChanged`/`WidthChanged` etc. with the old/new `Dim` values when the absolute size changes during layout; **or**
2. **Document the limitation** — XML-doc that `Width/Height` change events do not fire for `Frame`/layout-driven changes and that consumers should subscribe to `FrameChanged` instead.

Option 1 is more correct but is a behavior change (handlers that never fired before will start firing during layout) and needs care around recursion. Option 2 is safe but leaves the API surprising.

## Scope

- Resolve the two BUGBUG blocks in `View.Layout.cs` (`Frame` setter + `SetRelativeLayout`).
- Either raise `Width/Height Changing/Changed` + reset `TextFormatter`, or document the gap.
- Tests covering whether the events fire (or are documented not to) when `Frame` is set directly and when `SetRelativeLayout` changes an absolute size.

harder:
### Background

#5498 flagged that frame/layout-driven size changes bypass the `Width`/`Height` `Changing`/`Changed` events and might leave `TextFormatter.ConstrainTo*` stale.

The issue asked for a decision:
1. Raise `Width`/`Height` events from the `Frame`/layout paths.
2. Document that those events only apply to declarative `Dim` assignment, and use `FrameChanged` for resolved-size changes.

This PR takes option 2: document the contract and lock the current behavior with tests.

### Findings
The `TextFormatter` concern is no longer a live bug. `SetRelativeLayout` now recomputes formatter constraints through `SetTextFormatterSize ()` and `FinalizeTextFormatterConstraints ()`. Direct `Frame` assignment also flows through that path via `Layout ()`, so the old `// BUGBUG: - Reset TextFormatter` comments were stale.

The event behavior is intentional once the model is separated into:
- Declarative layout: `Width`/`Height` as `Dim`
- Resolved layout: `Frame` as an absolute `Rectangle`

A layout pass can change `Frame.Width` while `Width` remains `Dim.Fill ()`. Raising `WidthChanged` in that case would imply the declarative `Dim` changed, which is not true. Direct `Frame` assignment does overwrite the internal `Dim` fields with absolute values, but that is bookkeeping after the `Frame` has already changed, not a cancellable declarative `Width`/`Height` assignment.

Contract:
`WidthChanging`/`WidthChanged` and `HeightChanging`/`HeightChanged` observe declarative `Dim` assignment only. `FrameChanged` is the event for resolved position/size changes from any cause.

### Changes

- Replaced the stale `BUGBUG` blocks in `View.Layout.cs` with design notes explaining the direct field writes.
- Added XML docs to `Width`, `Height`, and `FrameChanged` describing the event contract.
- Added contract tests covering:
  - `WidthChanged`/`HeightChanged` do not fire from direct `Frame` assignment.
  - `WidthChanging`/`HeightChanging` do not fire from direct `Frame` assignment.
  - `OnWidthChanging`/`OnWidthChanged` and `OnHeightChanging`/`OnHeightChanged` are not invoked from direct `Frame` assignment.
  - The same events/hooks do not fire from layout-driven resize.
  - `FrameChanged` does fire from direct `Frame` assignment, layout-driven resize, and declarative assignment after layout.

### Notes

`FrameChanged` fires after `Frame` has been updated. It should be used to observe the resolved `Frame`, not as a signal that every secondary layout/formatter side effect in the caller path has completed.

### Testing

- `dotnet test --project Tests/UnitTestsParallelizable --filter-class "*ViewLayoutEventTests"`  
  - Passed: 21/21
- `dotnet build Terminal.Gui/Terminal.Gui.csproj --no-restore -t:Rebuild -v:minimal`  
  - Passed with one existing CS0419 XML-doc warning in `View.Drawing.cs`
- `git diff --check upstream/develop...HEAD`  
  - Passed

Actual changes are comments/XML docs only; behavior is unchanged.

### DIFF
--- a/Terminal.Gui/ViewBase/View.Layout.cs
+++ b/Terminal.Gui/ViewBase/View.Layout.cs
@@ -67,11 +67,14 @@ public partial class View // Layout APIs
                 return;
             }
 
-            // BUGBUG: We set the internal fields here to avoid recursion. However, this means that
-            // BUGBUG: other logic in the property setters does not get executed.  Specifically:
-            // BUGBUG: - Reset TextFormatter
-            // BUGBUG: - SetLayoutNeeded (not an issue as we explicitly call Layout below)
-            // BUGBUG: - If we add property change events for X/Y/Width/Height they will not be invoked
+            // We set the internal fields directly (not via the X/Y/Width/Height setters) to avoid recursion.
+            // This is intentional and the two side effects of the setters are accounted for elsewhere (#5498):
+            // - TextFormatter constraints: recomputed by the Layout () -> SetRelativeLayout call below
+            //   (SetTextFormatterSize + FinalizeTextFormatterConstraints), so no reset is needed here.
+            // - Width/Height (and X/Y) change events: deliberately NOT raised here. Those events observe
+            //   declarative Dim/Pos assignment; this path overwrites them with Absolute values as bookkeeping
+            //   to keep state consistent with an imperatively-set Frame. Use FrameChanged (raised by SetFrame
+            //   above) to observe resolved-size changes from any cause.
             // If Frame gets set, set all Pos/Dim to Absolute values.
             _x = _frame!.Value.X;
             _y = _frame!.Value.Y;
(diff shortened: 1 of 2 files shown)


tig (approved):
Thank you. This is long overdue and the right fix.

PR #5602, opened 2026-07-25
Fixes #5601. Synchronize dirty-row flags after output

### LINKED ISSUE #5601: Optimize output flushing by synchronizing dirty-row flags
## Summary

The output buffer tracks pending rendering work at two levels:

- `Cell.IsDirty` for individual cells
- `DirtyLines` for rows that may contain pending dirty cells

Rows remain flagged in `DirtyLines` after their dirty cells have been written and cleared. Subsequent output passes then continue to scan those rows despite there being no cell-level work left.

## Proposed optimization

Maintain the row-level dirty-state lifecycle alongside the existing cell-level lifecycle:

1. Clear `DirtyLines[row]` after all pending cells in that row have been processed.
2. Mark `DirtyLines[row]` whenever a drawing operation dirties a cell in that row.

This allows the output pass to skip rows with no pending changes while preserving correct rendering for text writes, fills, and other buffer updates.

## Expected impact

A prototype benchmark using a warmed 340x65 buffer and a single-cell update measured:

| Metric | Current behavior | Optimized behavior |
|---|---:|---:|
| Mean output-buffer refresh time | 166.0 us | 3.315 us |
| Allocation per refresh | 270.35 KB | 5.85 KB |
| Dirty-row processing | Previously rendered rows remain eligible | Only rows with pending work are processed |

The benchmark uses a no-I/O output backend, so it isolates framework overhead.

## Acceptance criteria

- A row is skipped when it contains no pending dirty cells.
- A row becomes eligible for output whenever a drawing operation dirties one of its cells.
- Existing rendering behavior, including fills and wide-glyph handling, remains correct.
- Regression tests cover clearing a row after output and re-marking it after a subsequent update.

## Alternatives considered

Recomputing row state from every cell during each output pass would avoid stale flags, but would retain the full-screen scan that the row-level optimization is intended to avoid.

YourRobotOverlord:
- Fixes #5601

## Summary

- Clear a row's `DirtyLines` flag once its pending cells have been emitted.
- Keep cell and row dirty state synchronized through a single helper.
- Document dirty-row ownership for output-buffer implementations and direct `Contents` mutations.
- Add regression tests for idle output, dirty-row lifecycle, and inline-mode `FillRect` rendering.

## Root cause

`DirtyLines` remained set after a row's `Cell.IsDirty` flags were consumed. Output passes therefore continued scanning previously rendered rows even when no cells in them required output.

Conversely, `FillRect` marked cells dirty without marking their rows. Initially clean inline-mode buffers could therefore skip rows whose only content came from `FillRect`.

## Impact

The output layer now skips rows without pending work while preserving the invariant that every dirty cell belongs to a dirty row. This also ensures inline-mode fill, clear, border, margin, and padding drawing is eligible for emission.

A prototype 340x65 single-cell refresh benchmark measured:

| Metric | Before | After |
|---|---:|---:|
| Mean framework refresh time | 166.0 us | 3.315 us |
| Allocation per refresh | 270.35 KB | 5.85 KB |

The benchmark uses a no-I/O output backend and isolates framework overhead. It was rerun after the review changes and showed no regression.

## Validation

- `dotnet build --configuration Debug --no-restore`
- `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter-class "*OutputBaseTests" --verbosity normal` (71 passed)
- `dotnet test --project Tests/UnitTestsParallelizable --no-build --filter-class "*OutputBuffer*" --verbosity normal` (16 passed)
- `dotnet test --project Tests/UnitTests.NonParallelizable --no-build --verbosity normal` (72 passed, 2 skipped)
- Full parallel, non-parallel, and integration suites passed before the benchmark work.

# To pull down this PR locally:

```bash
git fetch copilot agent/retire-rendered-dirty-row-flags
git checkout copilot/agent/retire-rendered-dirty-row-flags
```

### DIFF
--- a/Terminal.Gui/Drivers/Output/IOutputBuffer.cs
+++ b/Terminal.Gui/Drivers/Output/IOutputBuffer.cs
@@ -44,6 +44,11 @@ public interface IOutputBuffer
     ///     Gets a value for each row indicating whether that row contains any dirty cells and should be rendered.
     ///     Used by <see cref="IOutput"/> to skip rows with no pending changes, avoiding unnecessary cursor moves.
     /// </summary>
+    /// <remarks>
+    ///     Output implementations clear each row's flag after emitting its dirty cells. Buffer implementations must set
+    ///     the corresponding row flag whenever a cell is marked dirty. Code that modifies <see cref="Contents"/> directly
+    ///     is responsible for updating the corresponding row flag.
+    /// </remarks>
     bool [] DirtyLines { get; }
 
     /// <summary>
--- a/Terminal.Gui/Drivers/Output/OutputBase.cs
+++ b/Terminal.Gui/Drivers/Output/OutputBase.cs
@@ -271,6 +271,10 @@ public abstract class OutputBase
                 }
             }
 
+            // Every dirty cell in this row has now been consumed. Drawing operations will
+            // set this flag again when they modify the row in a later frame.
+            buffer.DirtyLines [row] = false;
+
             // Flush buffered output for row. Even when nothing remains buffered, an OSC 8 hyperlink
             // may still be open in the terminal because it was started in a prior batch flushed by
             // WriteToConsole and the row ended (or only clean cells followed) before any cell with
--- a/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs
+++ b/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs
@@ -458,8 +458,14 @@ public class OutputBufferImpl : IOutputBuffer
             {
                 for (int col = visible.X; col < visible.Right; col++)
                 {
-                    Contents [row, col].IsDirty = isDirty;
-                    DirtyLines [row] |= isDirty;
+                    if (isDirty)
+                    {
+                        MarkDirty (col, row);
+
+                        continue;
+                    }
+
+                    Contents [row, col].IsDirty = false;
                 }
             }
         }
@@ -551,8 +557,6 @@ public class OutputBufferImpl : IOutputBuffer
                 string printableGrapheme = grapheme.MakePrintable ();
                 printableGraphemeWidth = printableGrapheme.GetColumns ();
                 WriteGraphemeByWidth (Col, Row, printableGrapheme, printableGraphemeWidth, clipRect);
-
-                DirtyLines [Row] = true;
             }
 
             // Always advance cursor (even if location was invalid)
@@ -575,7 +579,7 @@ public class OutputBufferImpl : IOutputBuffer
                 if (Contents [Row, Col].Attribute != CurrentAttribute)
                 {
                     Contents [Row, Col].Attribute = CurrentAttribute;
-                    Contents [Row, Col].IsDirty = true;
+                    MarkDirty (Col, Row);
                 }
             }
 
@@ -597,7 +601,18 @@ public class OutputBufferImpl : IOutputBuffer
             return;
         }
         Contents [row, col - 1].Grapheme = _column1ReplacementChar.ToString ();
-        Contents [row, col - 1].IsDirty = true;
+        MarkDirty (col - 1, row);
+    }
+
+    /// <summary>
+    ///     INTERNAL: Marks a cell and its row as dirty.
+    /// </summary>
+    /// <param name="col">The column.</param>
+    /// <param name="row">The row.</param>
+    private void MarkDirty (int col, int row)
+    {
+        Contents! [row, col].IsDirty = true;
+        DirtyLines [row] = true;
     }
 
     /// <summary>
@@ -608,8 +623,7 @@ public class OutputBufferImpl : IOutputBuffer
     private void SetAttributeAndDirty (int col, int row)
     {
         Contents! [row, col].Attribute = CurrentAttribute;
-        Contents [row, col].IsDirty = true;
-        DirtyLines [row] = true;
+        MarkDirty (col, row);
 
         // Update the URL map: store CurrentUrl, or clear any stale entry so cells
         // overdrawn by non-link content are not wrapped in OSC 8 sequences.
@@ -738,7 +752,7 @@ public class OutputBufferImpl : IOutputBuffer
         // Mark the next cell as dirty to ensure proper rendering of adjacent content
         if (col < clipRect.Right - 1 && col + 1 < Cols)
         {
-            Contents [row, col + 1].IsDirty = true;
+            MarkDirty (col + 1, row);
         }
     }
 
--- a/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs
+++ b/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs
@@ -609,6 +609,7 @@ public class OutputBufferImpl : IOutputBuffer
     {
         Contents! [row, col].Attribute = CurrentAttribute;
         Contents [row, col].IsDirty = true;
+        DirtyLines [row] = true;
 
         // Update the URL map: store CurrentUrl, or clear any stale entry so cells
         // overdrawn by non-link content are not wrapped in OSC 8 sequences.
--- a/Tests/UnitTestsParallelizable/Drivers/Output/OutputBaseTests.cs
+++ b/Tests/UnitTestsParallelizable/Drivers/Output/OutputBaseTests.cs
@@ -104,6 +104,38 @@ public class OutputBaseTests
         Assert.False (buffer.Contents! [0, 1].IsDirty);
     }
 
+    // CoPilot - GPT-5
+    [Fact]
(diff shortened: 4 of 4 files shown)


YourRobotOverlord:
Benchmark source used for the 340x65 single-cell refresh measurement (temporary file; not included in this PR):

```csharp
using BenchmarkDotNet.Attributes;
using System.Text;
using Terminal.Gui.Drivers;

namespace Terminal.Gui.Benchmarks.ConsoleDrivers.OutputBuffer;

[MemoryDiagnoser]
[BenchmarkCategory ("Output", "Latency")]
public class OutputRefreshLatencyBenchmark
{
    private IOutputBuffer _buffer = null!;
    private CountingOutput _output = null!;

    [GlobalSetup]
    public void Setup ()
    {
        _buffer = new OutputBufferImpl ();
        _buffer.SetSize (340, 65);
        _output = new CountingOutput ();
        _output.Write (_buffer);
        _output.ResetCounters ();
    }

    [Benchmark]
    public void SingleCellRefresh ()
    {
        _buffer.Move (0, 0);
        _buffer.AddStr ("X");
        _output.Write (_buffer);
    }

    private sealed class CountingOutput : OutputBase
    {
        public long CursorMoves { get; private set; }

        public long Writes { get; private set; }

        public void ResetCounters ()
        {
            CursorMoves = 0;
            Writes = 0;
        }

        protected override bool SetCursorPositionImpl (int screenPositionX, int screenPositionY)
        {
            CursorMoves++;

            return true;
        }

        protected override void Write (StringBuilder output) => Writes++;
    }
}
```

It was run in Release mode with:

```powershell
dotnet run --project Tests/Benchmarks --configuration Release --no-restore -- --filter "*OutputRefreshLatencyBenchmark*" --job short
```

The benchmark was run once against the baseline implementation and once with this PR's changes. `CountingOutput` intentionally avoids terminal I/O so the measurements isolate output-buffer framework work.

harder:
Thank you, this looks great! I did a code review with Claude Opus and it picked up a few nits (full automated code review below) that are optional if you feel like any of them are worthwhile.

---

### Code review

**Verdict: approve.** Correct, minimal, and the perf claim holds up under independent measurement. Five comments, none blocking.

#### Verified

- Full suites green: **Parallelizable** 17,540 passed / 0 failed (17 pre-existing skips), **NonParallelizable** 72/0, **Integration** 437/0. One flake on the first parallelizable run that did not reproduce across three subsequent runs.
- **Perf claim reproduced.** 340x65 buffer, steady state, 50 idle refreshes (Debug, arm64):

  | | Baseline | With PR |
  |---|---:|---:|
  | Alloc/frame | 268.84 KB | 0.20 KB |
  | Time/frame | 2021.7 us | 0.46 us |

  The mechanism is clear — pre-PR every *clean* cell in every row still ran `SetCursorPositionImpl`, one `CSI_SetCursorPosition` string per cell, 22,100 cells/frame:

  https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/Terminal.Gui/Drivers/Output/OutputBase.cs#L201-L213

- **Invariant audit.** Enumerated every site that sets `Cell.IsDirty = true` and confirmed each is covered by a `DirtyLines[row] = true` on the same row. Then ran probe tests over narrow/wide/combining glyphs, wide-glyph overlap invalidation, wide glyph at last column, clipping, `FillRect`, explicit + autolink URLs, and raster add/retire, asserting both "no dirty cell in a clean row" and "no cell left dirty after `Write`". All passed — no lost updates.
- Both `Write(IOutputBuffer)` overrides (`AnsiOutput`, `WindowsOutput`) delegate to `base.Write`, so the clear applies on every driver.
- Historical note: this restores semantics that existed until ab49fb8b4 (WindowsOutput rewrite) dropped `DirtyLines` from `IOutputBuffer`; baa12f62a (inline mode) reintroduced the read side only.

#### Findings

**1. This also fixes a real inline-mode rendering bug — undisclosed and untested.**

Before this change `FillRect` never touched `DirtyLines` at all, because it goes through `SetAttributeAndDirty`, which did not set it. In inline mode `ClearContents(initiallyDirty: false)` starts every row clean, so a row whose only content came from `FillRect` was **never emitted**. `FillRect` backs `View.ClearViewport` and `Thickness.Draw` (borders/margin/padding), so inline apps could silently drop filled or blank rows.

Confirmed: a probe asserting `DirtyLines[1]` after an inline-mode `FillRect` fails on `develop` and passes with this patch. Worth stating in the description, and worth an inline-mode regression test — `FillRect_MarksDirtyLine_AfterPreviousFlush` covers the flag but not the rendering consequence.

https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs#L610-L613

https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/Terminal.Gui/ViewBase/View.Drawing.cs#L397-L400

**2. `DirtyLines` is now load-bearing, with no guard.**

It used to be monotonic — set, never cleared — so a missed set was harmless. Now a missed set drops a cell permanently. Three sites set `Cell.IsDirty = true` without setting `DirtyLines`, and are safe *only* because their callers already dirtied the same row:

https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs#L577-L579

https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs#L599-L601

https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/Terminal.Gui/Drivers/Output/OutputBufferImpl.cs#L739-L742

Consider a `MarkDirty (col, row)` helper that does both, or a `Debug.Assert (DirtyLines [row])` at each site. Note `IOutputBuffer.Contents` is public and settable, so external code setting `IsDirty` directly now loses updates too.

**3. Document flag ownership.**

The existing doc already describes the *new* semantics, so this PR is really bringing code into line with the spec — a good framing for the description. But it does not say who clears the flags. Since `IOutputBuffer` is public and others may implement it, add a `<remarks>` noting that `IOutput.Write` clears each row's flag after emitting it, and implementations must set it whenever any cell in that row is dirtied.

https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/Terminal.Gui/Drivers/Output/IOutputBuffer.cs#L43-L48

**4. Nothing asserts the actual goal.** Both new tests check flag state, not the observable outcome — that an idle `Write` emits nothing. That cannot be asserted via `GetLastOutput()` today, because `SetCursorPositionImpl` writes through `Write(ReadOnlySpan<char>)` and bypasses `_lastOutputStringBuilder` (pre-existing, not this PR). A counting `IOutput` subclass would work. Optional.

**5. Nit: `Write_ClearsDirtyLines_AfterFlushing` is weak.** It never draws anything, so it only proves the initial all-dirty state gets cleared. Drawing first, as the `FillRect` test does, would make it meaningful.

#### Style

Compliant with [CLAUDE.md](https://github.com/tui-cs/Terminal.Gui/blob/4e555cc2cd0979b2aa531c4d2bb6e24cba2747cd/CLAUDE.md) — spacing before parens, target-typed `new ()`, AI-generated markers present, tests in `UnitTestsParallelizable`, no `Application.Init`.

#### Footnote

`OutputBufferImpl` guards `Contents` and `DirtyLines` with `_contentsLock`, but `OutputBase.Write` mutates both without it. Not a live race today — drawing and output both run on the main loop thread, only input is on a `Task.Run` — but the new write is what would make that window consequential if the threading model ever changes.

Generated with [Claude Code](https://claude.com/claude-code)

harder (approved):

YourRobotOverlord:
I will address those and update the PR, I think they're all worthwhile.

tig:
> I will address those and update the PR, I think they're all worthwhile.

converting to draft. ok?

YourRobotOverlord:
Addressed all five points:
- Added an inline-mode FillRect rendering regression test and updated the PR description.
- Added MarkDirty and routed all internal cell-dirtying paths through it.
- Documented row-flag ownership and direct Contents mutation requirements.
- Added a counting-output test proving idle writes emit no cursor moves or output.
- Strengthened the flush test to draw and verify dirty state before flushing.

The benchmark was rerun after these changes: 3.220 μs and 5.85 KB per refresh, with no regression.

tig (approved):

Issue #5604, opened 2026-07-28
Fix incorrect FreeBSD ARM64 ioctl workaround and modernize platform detection


kasperk81:
## Describe the bug

FreeBSD running on ARM64 hardware incorrectly triggers a specialized `ioctl_arm64` signature workaround path inside `UnixIOHelper.TryGetTerminalSize`. This workaround addresses a 64-bit varargs calling convention anomaly unique to macOS on Apple Silicon (detailed in dotnet/runtime #48796). On FreeBSD ARM64, this workaround is not required and instead causes terminal sizing initialization to behave incorrectly.

## To Reproduce

1. Run any Terminal.Gui application (like UICatalog) on a FreeBSD ARM64 machine.
2. Expected behavior: The terminal initializes size boundaries correctly using standard `ioctl`.
3. Actual behavior: The driver attempts an invalid `ioctl_arm64` call path, resulting in broken terminal layout detection or initial sizing failures.

### OS Information

**macOS/Linux:**
```bash
echo "OS: $(uname -s) $(uname -r)"
```

**Output:**
```
OS: FreeBSD 14.x-RELEASE (or similar ARM64 FreeBSD environment)
```

### Terminal Information

**Other terminals:**
```bash
echo $TERM
```

**Output:**
```
xterm-256color
```

### .NET Information

```bash
dotnet --version
dotnet --info
```

**Output:**
```
.NET 10.0 (ARM64)
```

### Terminal.Gui Version

**Version:**
```
v2 / Main branch (Pre-PR #5600)
```

## Screenshots, GIFs, or Terminal Output

The issue stems from the logic inside `UnixIOHelper.cs`:
```csharp
if (RuntimeInformation.OSArchitecture == Architecture.Arm64
    && (RuntimeInformation.IsOSPlatform (OSPlatform.OSX) || RuntimeInformation.IsOSPlatform (OSPlatform.FreeBSD)))
```
FreeBSD ARM64 uses regular `ioctl` and should not be bundled with macOS here.

## Additional context

This happens consistently on FreeBSD ARM64 architectures. 

Fixing this also opens up an opportunity to modernize the framework's overall platform management. Replacing our custom `PlatformDetection` wrappers with native `.NET` `OperatingSystem.IsLinux/Windows/MacOS()` APIs allows for JIT-time dead code elimination and naturally improves compatibility fallbacks for other non-Windows Unix OS environments (like NetBSD, OpenBSD, and Solaris).

PR #5606, opened 2026-07-30
Fix platform table and ARM64 ioctl guard defects following #5600

harder:
Follow-up to #5600. That PR fixed a FreeBSD/ARM64 bug (#5604) and the move to `OperatingSystem.Is*` is right; this corrects defects in the platform tables and predicates it touched. None are reachable from CI, which runs Windows/Linux/macOS only.

## Defects

| Defect | Was | Correct | Regression from #5600 |
|---|---|---|---|
| Haiku suspend signal | `21` (`SIGKILLTHR`) | `13` (`SIGTSTP`) | **Yes** — worse than the pre-#5600 no-op |
| Android `TIOCGWINSZ` | `0x40087468` | `0x5413` | **Yes** |
| Android suspend signal | `-1` | `20` | **Yes** |
| `PlatformDetection` public members | removed | restored | **Yes** — binary break |
| ARM64 ioctl guard architecture axis | `OSArchitecture` | `ProcessArchitecture` | No |
| ARM64 ioctl guard platform axis | `IsMacOS ()` | all Apple platforms | No |
| Haiku `TIOCGWINSZ` | `0x40087468` | `0x800C` | No |
| Solaris/illumos `TIOCGWINSZ` | `0x40087468` | `0x5468` | No |
| `AnsiOutput` descriptor leak | leaked per instance | closed | No |
| Linux `TIOCGWINSZ` on ppc64le | `0x5413` | `0x40087468` | No |
| `Suspend ()` return value | always `true` | `killpg` result | No |

Sources: [Haiku `signal.h`](https://github.com/haiku/haiku/blob/master/headers/posix/signal.h) (`SIGTSTP` 13, `SIGKILLTHR` 21) · [Haiku `termios.h`](https://github.com/haiku/haiku/blob/master/headers/posix/termios.h) (`TCGETA` `0x8000`, `TIOCGWINSZ` `TCGETA + 12`) · [`OperatingSystem.cs`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/OperatingSystem.cs) (`IsLinux()` is `#if TARGET_LINUX && !TARGET_ANDROID`) · [`pal_runtimeinformation.c`](https://github.com/dotnet/runtime/blob/main/src/native/libs/System.Native/pal_runtimeinformation.c) (`OSArchitecture` returns `ARCH_ARM64` when `sysctl.proc_translated`) · [dotnet/runtime#48796](https://github.com/dotnet/runtime/issues/48796#issuecomment-3695794860) (Apple ARM64 variadic ABI)

### Notes on the two non-obvious ones

**Android.** `OperatingSystem.IsLinux ()` returns `false` on Android. #5600 inverted the `TIOCGWINSZ` ternary to key off it, which changed the *default* for every platform name that is not `LINUX`/`OSX`/`FREEBSD`. Android uses asm-generic ioctl numbering, so it needs `0x5413`. Affects `net*-android` targets; not Termux, which runs a stock `linux-arm64` runtime.

**Rosetta.** .NET reports `OSArchitecture == Arm64` for an x64 process translated by Rosetta, but that process executes x64 code and passes variadic arguments in registers. The old guard therefore selected `ioctl_arm64`'s stack-passing signature for a process needing the register one: `winsize*` arrived as `NULL` in `rdx`, `ioctl` returned `EFAULT`, sizing fell back to 80×25 on any x64 process on Apple silicon. `ProcessArchitecture` is the correct axis and is a compile-time constant rather than a cached `uname` call. Separately, the stack-passing convention is an Apple ABI deviation, not a macOS one, so Mac Catalyst/iOS/tvOS need it too. watchOS is deliberately absent — .NET has no `WATCHOS` platform name.

## Restored public API

`IsLinux`, `IsMac`, `IsUnixLike`, `IsWindows` and `GetCurrentPlatform` shipped `public static` in v2.4.17 and were deleted by #5600, so consumers compiled against 2.4.17 hit `MissingMethodException` on upgrade. Restored as `[Obsolete]` shims with **v2.4.17 behaviour preserved exactly, not improved** — `IsUnixLike ()` still covers Linux/macOS/FreeBSD only, with the reason not to use it documented on the member rather than silently changing what it returns. Verified signature-identical to the tag. Also un-orphans `TuiPlatform`, which had no producer after #5600.

Two maintainer calls, both one-line edits: whether the shims should warn now or defer the warning to the next minor, and whether `TuiPlatform`/`GetCurrentPlatform` live on or are scheduled for v3.

## Design

Each lookup is now a pure function over the .NET OS platform name, so the tables are directly testable. All are once-per-process cached (`_suspendSignal`; `TIOCGWINSZ` and the guard are `static readonly`/const-folded), so this costs nothing measurable — the ~25 call sites where #5600's const-folding matters are untouched.

Every value carries its derivation in a comment, and encoding tests re-derive the BSD (`_IOR('t',104,winsize)`), Solaris (`TIOC|104`) and Haiku (`TCGETA+12`) constants from first principles rather than restating a magic number.

## Verification

- Full build clean, no new warnings; `CS0618` suppressed only inside the shim tests
- `UnitTestsParallelizable` 17,638 passed / 0 failed / 17 skipped (was 17,559)
- `UnitTests.NonParallelizable` 72 passed; Legacy, IntegrationTests, UICatalog build clean
- Mutation-checked: reintroducing the Haiku 21 and Android gaps fails 5 tests; removing the Haiku ioctl entry fails 2; narrowing the Apple set to macOS fails 3

Two honest limits: the ARM64 call-site swap to `ProcessArchitecture` is verified by inspection, not test, since the two values coincide on all non-Rosetta hardware. And every fix here is on a platform or configuration neither CI nor I can execute — they are verified against vendor headers and the .NET runtime source, not by running.

Unrelated finding while running the suite: `CollectionNavigatorTests.Delay` is timing-dependent (`Thread.Sleep (TypingDelay + 10)`) and fails roughly 1 run in 4 under parallel load. Untouched by this PR; worth its own issue.

## Deliberately excluded

- **PublicAPI baseline** (`Microsoft.CodeAnalysis.PublicApiAnalyzers`) — the systemic fix for how #5600's removal passed CI, but it generates a baseline for the entire public surface and gates every future PR, so it needs its own PR and your buy-in. A targeted reflection test pins `PlatformDetection`'s surface here in the meantime.
- **Linux `TIOCGWINSZ` on MIPS/SPARC/Alpha** — these use `0x40087468` like ppc64le, but .NET has no targets for them, so they stay documented in a code comment rather than handled.
- **Unused `using System.Runtime.InteropServices;`** left by #5600 in ~9 files — pure churn that would bury the substantive diff.

### DIFF
--- a/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
+++ b/Terminal.Gui/Drivers/AnsiDriver/AnsiOutput.cs
@@ -135,6 +135,10 @@ public class AnsiOutput : OutputBase, IOutput
                     return;
                 }
 
+                // The duplicate only proves the fd is usable; writes go through UnixIOHelper.TryWriteStdout, which
+                // resolves TerminalDevice.OutputFd itself. Close it so each AnsiOutput does not leak a descriptor.
+                UnixIOHelper.close (fdCopy);
+
                 _platform = AnsiPlatform.UnixRaw;
             }
 
--- a/Terminal.Gui/Drivers/DotNetDriver/NetOutput.cs
+++ b/Terminal.Gui/Drivers/DotNetDriver/NetOutput.cs
@@ -212,7 +212,10 @@ public class NetOutput : OutputBase, IOutput
 
             if (!SuspendHelper.Suspend ())
             {
-                return;
+                // Do NOT return here. The alternate buffer has already been left and the cursor shown, so returning
+                // would leave the terminal that way while the app keeps drawing into the scrollback buffer.
+                Logging.Warning ("NetOutput.Suspend: SuspendHelper.Suspend () returned false; the process was never "
+                                 + "stopped. Re-entering the alternate buffer anyway.");
             }
 
             //Enable alternative screen buffer.
--- a/Terminal.Gui/Drivers/PlatformDetection.cs
+++ b/Terminal.Gui/Drivers/PlatformDetection.cs
@@ -54,6 +54,66 @@ public static class PlatformDetection
     }
 
     /// <summary>
+    ///     Determines whether the current operating system is Linux.
+    /// </summary>
+    /// <remarks>
+    ///     This method returns <see langword="true"/> only when running on a Linux distribution. Other Unix-like
+    ///     platforms such as macOS and FreeBSD return <see langword="false"/>, as does Android.
+    /// </remarks>
+    /// <returns><see langword="true"/> if the operating system is Linux; otherwise, <see langword="false"/>.</returns>
+    [Obsolete ("Use OperatingSystem.IsLinux () instead. This shim exists for binary compatibility with v2.4.17 and earlier.")]
+    public static bool IsLinux () => OperatingSystem.IsLinux ();
+
+    /// <summary>
+    ///     Determines whether the current operating system is macOS.
+    /// </summary>
+    /// <returns><see langword="true"/> if the current operating system is macOS; otherwise, <see langword="false"/>.</returns>
+    [Obsolete ("Use OperatingSystem.IsMacOS () instead. This shim exists for binary compatibility with v2.4.17 and earlier.")]
+    public static bool IsMac () => OperatingSystem.IsMacOS ();
+
+    /// <summary>
+    ///     Determines if the current platform is Windows.
+    /// </summary>
+    /// <returns><see langword="true"/> if the operating system is Windows; otherwise, <see langword="false"/>.</returns>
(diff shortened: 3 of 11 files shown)


kasperk81:
> `MissingMethodException`

it is hard to beleive anyone is using tui-cs platform detection in their code to detect current platform? this looks like AI thinking too much into it. i suggest keep it simple. you can read all about it in their API proposal in dotnet/runtime in front of the entire .net community why they chose this api shape over "give me the name of OS" which is a faulty design.

PR #5610, opened 2026-08-04
Fix TIOCGWINSZ on other platforms

kasperk81:
Two values were incorrect in https://github.com/tui-cs/Terminal.Gui/pull/5600. Let's fix them instead of bringing back unnecessary ceremony in https://github.com/tui-cs/Terminal.Gui/pull/5606. All this manual signal mapping will go away once the Terminal.Gui framework moves past .NET 11 (`(int)PosixSignal.SIGTSTP`), so it's temporary.


kasperk81:
@tig, this one is ready.

PR #5618, opened 2026-08-13
Fixes #5616. Preserve repeated legacy input after kitty detection

### LINKED ISSUE #5616: Kitty-enabled ANSI input drops every second repeated legacy key
## Describe the bug

When kitty keyboard support has been detected, repeated fallback input with the same printable text
loses every second key. This occurs when a terminal reports kitty support but sends a key as legacy
input. Six paced `j` presses, for example, produce only three `KeyDown` events.

This is the kitty-enabled follow-up to #4976. The retained failure mode was called out as a watch
point in #4977.

## To Reproduce

Add this regression test to
`Tests/UnitTestsParallelizable/Drivers/AnsiHandling/KittyKeyboardPipelineTests.cs`:

```csharp
[Fact]
public void Pipeline_RepeatedLegacyPrintableInput_WithKittyEnabled_DoesNotDropCharacters ()
{
    (List<Key> down, List<Key> up) = InjectRawSequenceWithKittyEnabled ("jjjj");

    Assert.Equal ("jjjj", string.Concat (down.Select (key => key.GetPrintableText ())));
    Assert.Empty (up);
}
```

Run:

```bash
dotnet test --project Tests/UnitTestsParallelizable --filter-class "*KittyKeyboardPipelineTests"
```

**Expected:** four `KeyDown` events.

**Actual:** two `KeyDown` events.

The same behavior was reproduced through the public application pipeline with
`Terminal.Gui 2.4.18-develop.27`: six `j` presses spaced 500 ms apart inside an unconfigured Neovim
`:terminal` produced three `KeyDown` events. Running the same app directly produced six.

## Diagnosis

[`AnsiInputProcessor.ShouldSuppressFallbackKeyDown`](https://github.com/tui-cs/Terminal.Gui/blob/2e47b11478db083499917f2ad27c34d30efb0df1/Terminal.Gui/Drivers/AnsiDriver/AnsiInputProcessor.cs#L83-L105)
arms `_pendingPrintableSuppression` from the fallback key itself whenever
`_isKittyKeyboardEnabled` is true. The next identical fallback key is therefore treated as a
duplicate, even if it is a later, legitimate press:

1. first `j`: delivered and arms `"j"`;
2. second `j`: suppressed and clears the pending value;
3. subsequent keys repeat the cycle.

The pending value has no expiry, which is why the failure also occurs with seconds between presses.

## Recommended fix

Make suppression provenance-aware: a fallback key should only be suppressed when a matching parsed
kitty event armed the pending value. The existing `OnKeyboardEventParsed` path already provides
that evidence; the fallback path should not leave a suppression value pending indefinitely.

If fallback-only de-duplication is still required for #4918, scope it to evidence that both inputs
belong to one physical event - for example, the same input batch or a short deadline - and clear it at
that boundary. Add regression coverage for both cases:

- parsed kitty event plus its matching legacy fallback produces one `KeyDown`;
- kitty enabled plus separate repeated legacy keys produces every `KeyDown`.

## Environment

- Terminal.Gui: `2.4.18-develop.27`, commit `fffd19e08319998ca52bcb67ec4ae6ad35fc2903`
- Current `develop` checked: `2e47b11478db083499917f2ad27c34d30efb0df1` (the relevant code is unchanged)
- .NET SDK: `10.0.302`
- OS: NixOS 26.11, Linux 6.18.39 x86_64
- Reproducing terminal transport: Neovim v0.12.4 `:terminal`, `-u NONE`
- Direct-terminal control: no key loss

alsi-lawr:
Closes #5616

### DIFF
--- a/Terminal.Gui/App/MainLoop/MainLoopCoordinator.cs
+++ b/Terminal.Gui/App/MainLoop/MainLoopCoordinator.cs
@@ -222,11 +222,6 @@ internal class MainLoopCoordinator<TInputRecord> : IMainLoopCoordinator where TI
                                           {
                                               if (!result.IsSupported)
                                               {
-                                                  if (_inputProcessor is AnsiInputProcessor unsupportedAnsiInputProcessor)
-                                                  {
-                                                      unsupportedAnsiInputProcessor.SetKittyKeyboardEnabled (false);
-                                                  }
-
                                                   Trace.Lifecycle (app?.MainThreadId?.ToString (), "KittyKeyboard", "Kitty keyboard mode not enabled");
 
                                                   return;
@@ -234,12 +229,6 @@ internal class MainLoopCoordinator<TInputRecord> : IMainLoopCoordinator where TI
 
                                               // Kitty is supported. Store the capabilities and set the flags we care about.
                                               _driver?.SetKittyKeyboardCapabilities (result);
-
-                                              if (_inputProcessor is AnsiInputProcessor supportedAnsiInputProcessor)
-                                              {
-                                                  supportedAnsiInputProcessor.SetKittyKeyboardEnabled (true);
-                                              }
-
                                               kittyKeyboardDetector.Enable (EscSeqUtils.KittyKeyboardRequestedFlags);
 
                                               Trace.Lifecycle (app?.MainThreadId?.ToString (),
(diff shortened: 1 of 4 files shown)


tig (approved):
Thanks. Lgtm.

PR #5633, opened 2026-08-22
Addresses #5627. Skip clean runs during output flush

### LINKED ISSUE #5627: OutputBase.Write issues one cursor-move console write per clean cell, making sparse frames far more expensive than full ones
### Summary

`OutputBase.Write` emits one `SetCursorPositionImpl` call **per clean cell** in every row flagged dirty. Consecutive dirty cells are batched into a single write, but clean cells are not batched at all — each one steps the cursor forward a single column with its own console write.

The practical effect is that a frame is charged for its *gaps* rather than its content, so a sparse update costs far more than a full-screen one. On a large terminal this makes an application feel broken: in my case a single keystroke took ~240 ms while a full-screen scroll took ~70 ms.

### Where

[`Terminal.Gui/Drivers/Output/OutputBase.cs#L190-L213`](https://github.com/tui-cs/Terminal.Gui/blob/v2.4.17/Terminal.Gui/Drivers/Output/OutputBase.cs#L190-L213) (v2.4.17):

```csharp
for (; col < cols; col++)
{
    bool skipRasterBlank = IsRasterCoveredBlankCell (buffer, row, col, rasterCellRectangles);

    if (!buffer.Contents! [row, col].IsDirty || skipRasterBlank)
    {
        ...
        if (lastCol + 1 < cols)
        {
            lastCol++;
        }

        SetCursorPositionImpl (lastCol, row);   // once per clean cell

        continue;
    }
    ...
}
```

And [`WindowsOutput.SetCursorPositionImpl`](https://github.com/tui-cs/Terminal.Gui/blob/v2.4.17/Terminal.Gui/Drivers/WindowsDriver/WindowsOutput.cs#L217), which is what each of those calls costs:

```csharp
var sb = new StringBuilder ();
EscSeqUtils.CSI_AppendCursorPosition (sb, screenPositionY + 1, screenPositionX + 1);
Write (sb.ToString ());
```

A fresh `StringBuilder` and a separate write to the console, per clean cell.

### Measurements

Windows 11, Windows Terminal (ConPTY), .NET 10, Terminal.Gui 2.4.17, Release build, 72 rows x 270 columns. Timing is from the point every view has finished drawing to the point the frame is complete, so it excludes all application drawing (which measured 48 us).

| Frame | Cells painted | Clean cells stepped over | Cost |
|---|---|---|---|
| One keystroke | ~1,000, spread over all 72 rows | ~18,000 | **~240 ms** |
| Wheel scroll | ~15,000, contiguous | ~0 | **~70 ms** |

Fifteen times the content for a third of the time. The cost tracks the number of clean cells, at roughly 13 us each, which is about what a small write to ConPTY costs.

Because the cost scales with `rows x columns`, it gets worse as the terminal gets larger — the same keystroke cost ~45 ms in a 30-row window and ~240 ms full-screen.

This is independent of driver (`windows`, `ansi` and `dotnet` all measured the same, since they share `OutputBase`), and independent of colour depth (`Force16Colors` shortens every escape sequence roughly fourfold and changed the timing by nothing measurable, which is a further sign the cost is the number of writes rather than their size).

### Reproducing

Any frame that paints a small number of cells spread across many rows. A convenient shape is a full-screen `Window` with a border: the border adornment repaints whenever the window does, touching the first and last column of every row, which flags every row dirty. A single-character edit in a `TextField` then leaves each of those rows with a handful of dirty cells and several hundred clean ones, and every one of those clean cells becomes its own console write.

### Suggested fix

Don't walk clean cells one at a time. Within a dirty row, find the next dirty cell and issue a single cursor-move to it, rather than stepping column by column. That keeps the existing batching of dirty runs and removes the per-clean-cell cost entirely.

As a workaround from the application side, marking every cell of an already-dirty row as dirty before the flush is enough to avoid it: it converts several hundred cursor-move writes per row into one batched run. Doing that took the frame time above from ~240 ms to ~6 ms, a ~40x improvement, with no visible change.

YourRobotOverlord:
## Summary

Optimize `OutputBase.Write` so sparse dirty rows skip clean gaps instead of emitting one cursor-position operation per clean cell.

- Closes #5627
- Related work: #5601 and #5602 skip wholly clean rows; this PR handles clean gaps within dirty rows

## Changes

- Accumulate and flush consecutive dirty cells as output runs
- Position the cursor once at the beginning of each dirty run
- Preserve hyperlink, raster, wide-grapheme, legacy-console, and dirty-row behavior
- Add 270x72 full-frame and sparse-border BenchmarkDotNet coverage
- Add a Windows ConPTY benchmark command and regression tests

## Benchmarks

| Case | Before | After | Result |
|---|---:|---:|---:|
| Sparse BenchmarkDotNet mean | 2.741 ms | 0.704 ms | 3.89x faster |
| Sparse allocation | 3.83 MiB | 300.53 KiB | 92.3% lower |
| Sparse cursor calls | 19,440 | 144 | 135x fewer |
| ConPTY median | 347.196 ms | 5.713 ms | 60.8x faster |
| ConPTY p95 | 353.920 ms | 6.070 ms | 58.3x faster |

Full-frame BenchmarkDotNet performance remained effectively flat. Content writes remain unchanged at 144 for the sparse frame.

## Testing

- Debug build: 0 warnings, 0 errors
- `OutputBaseTests`: 73 passed
- Parallelizable unit tests: 17,575 passed, 17 skipped
- Non-parallelizable unit tests: 72 passed, 2 skipped
- Integration tests: 437 passed
- ConPTY captures used tuirec v0.9.1 at 270x72

```bash
git remote add yor https://github.com/YourRobotOverlord/Terminal.Gui.git
git fetch yor fix/skip-clean-runs
git checkout --track yor/fix/skip-clean-runs
```

### DIFF
--- a/Terminal.Gui/Drivers/Output/OutputBase.cs
+++ b/Terminal.Gui/Drivers/Output/OutputBase.cs
@@ -173,92 +173,81 @@ public abstract class OutputBase
                 outputStringBuilder.Append (EscSeqUtils.OSC_EndHyperlink ());
             }
 
+            lastCol = 0;
+            var outputWidth = 0;
+
             // Process columns in row
             for (int col = left; col < cols; col++)
             {
-                lastCol = -1;
-                var outputWidth = 0;
+                // Skip clean cells, plus blank cells owned by raster images. Raster-owned
+                // blanks must not be emitted because normal text output would erase Sixel
+                // images and is unnecessary over Kitty's transparent-cleared placement.
+                bool skipRasterBlank = IsRasterCoveredBlankCell (buffer, row, col, rasterCellRectangles);
 
-                // Batch consecutive dirty cells
-                for (; col < cols; col++)
+                if (!buffer.Contents! [row, col].IsDirty || skipRasterBlank)
                 {
-                    // Skip clean cells, plus blank cells owned by raster images. Raster-owned
-                    // blanks must not be emitted because normal text output would erase Sixel
-                    // images and is unnecessary over Kitty's transparent-cleared placement.
-                    bool skipRasterBlank = IsRasterCoveredBlankCell (buffer, row, col, rasterCellRectangles);
-
-                    if (!buffer.Contents! [row, col].IsDirty || skipRasterBlank)
+                    if (skipRasterBlank)
                     {
-                        if (skipRasterBlank)
-                        {
-                            buffer.Contents [row, col].IsDirty = false;
-                        }
-
-                        if (outputStringBuilder.Length > 0)
-                        {
-                            // This clears outputStringBuilder
-                            WriteToConsole (outputStringBuilder, ref lastCol, ref outputWidth);
-                        }
-                        else if (lastCol == -1)
-                        {
-                            lastCol = col;
-                        }
-
-                        if (lastCol + 1 < cols)
-                        {
-                            lastCol++;
-                        }
-
-                        SetCursorPositionImpl (lastCol, row);
-
-                        continue;
-                    }
-
-                    if (lastCol == -1)
-                    {
-                        lastCol = col;
-                    }
-
-                    // Handle URL hyperlink state changes
-                    if (!IsLegacyConsole)
-                    {
-                        string? cellUrl = buffer.GetCellUrl (col, row);
-
-                        if (cellUrl != _lastUrl)
-                        {
-                            // If we were in a hyperlink, end it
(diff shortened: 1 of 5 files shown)


YourRobotOverlord:
**Before:**
before
**After:**
after

tig (approved):
Well done. Thank you!

PR #5648, opened 2026-08-26
perf: opt-in reflection scan skip + config property cache

liuqihonggit:
## Summary

- Add ConfigProperty.RegisterExternalConfigPropertyHostTypes: consumers pre-register external host types to skip the runtime reflection scan (AOT/startup-friendly). Default behavior unchanged.
- Cache GetAllConfigProperties + CreateCopy to halve reflection in ConfigurationManager init.

## Files Changed
- Terminal.Gui/Configuration/ConfigProperty.cs (+47/-2)
- Terminal.Gui/Configuration/ConfigurationManager.cs (+13/-3)

## Motivation
Eliminates ~12% startup cost from runtime assembly scan when no external hosts exist. Consumers can pre-register their known host set for AOT compatibility.

### DIFF
--- a/Terminal.Gui/Configuration/ConfigProperty.cs
+++ b/Terminal.Gui/Configuration/ConfigProperty.cs
@@ -485,6 +485,34 @@ public class ConfigProperty
     private static ImmutableSortedDictionary<string, Type>? _classesWithConfigProps;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


liuqihonggit:
关闭,稍后重新提交

tig:
See 

- #5416 

With it, CM, and the issue you have goes away. 

It is close to ready to being merged...

PR #5684, opened 2026-09-17
Fixes #5683. Synchronize ListView keystroke navigation

### LINKED ISSUE #5683: ListView keyboard navigation retains stale ToList snapshot after CollectionChanged
## Describe the bug

`ListView` keeps a stale keyboard-navigation collection when a custom `IListDataSource.ToList()` returns a snapshot or projection. `CollectionChanged` updates the displayed list and content size, but does not refresh `KeystrokeNavigator.Collection`.

After items are removed, typing a letter in the list can return an index from the old snapshot and throw `ArgumentException` from `SelectedItem`. Reordering the items without changing their count can silently select the wrong row instead.

## To Reproduce

Create a .NET 10 console project referencing `Terminal.Gui` **2.5.0-beta.1**, enable implicit usings and nullable reference types, and use this `Program.cs`. No application loop, terminal driver, or external data is required.

```csharp
using System.Collections;
using System.Collections.ObjectModel;
using Terminal.Gui.Input;
using Terminal.Gui.Views;

ObservableCollection<string> items = new(["Alpha", "Bravo", "Zulu"]);
using var list = new ListView
{
    Source = new SnapshotSource(items),
    SelectedItem = 0,
};

// Both removals raise CollectionChanged. Only "Zulu" remains at index 0.
items.RemoveAt(0);
items.RemoveAt(0);
list.SelectedItem = 0;

if (args.Contains("--workaround"))
{
    list.KeystrokeNavigator!.Collection = list.Source!.ToList();
}

Console.WriteLine($"Source count: {list.Source!.Count}; navigator count: {list.KeystrokeNavigator!.Collection.Count}");
list.NewKeyDownEvent(Key.Z);
Console.WriteLine($"Selected item: {list.SelectedItem}");

// A custom data source may return a projection/snapshot for keyboard navigation.
sealed class SnapshotSource(ObservableCollection<string> items)
    : ListWrapper<string>(items), IListDataSource
{
    IList IListDataSource.ToList() => items.ToList();
}
```

Run `dotnet run -c Release`.

### Expected behavior

Keyboard navigation uses the current source contents. Pressing Z selects the remaining `Zulu` item at index 0 without throwing.

### Actual behavior

```text
Source count: 1; navigator count: 3
Unhandled exception. System.ArgumentException: SelectedItem must be greater than 0 or less than the number of items.
   at Terminal.Gui.Views.ListView.set_SelectedItem(Nullable`1 value)
   at Terminal.Gui.Views.ListView.OnKeyDown(Key key)
   at Terminal.Gui.ViewBase.View.<NewKeyDownEvent>g__RaiseKeyDown|517_0(Key k)
   at Terminal.Gui.ViewBase.View.NewKeyDownEvent(Key key)
```

Run `dotnet run -c Release -- --workaround` to refresh the navigator explicitly before the key event. This succeeds:

```text
Source count: 1; navigator count: 1
Selected item: 0
```

Both runs were verified against the NuGet package.

## Environment

- OS: Microsoft Windows 10.0.26200, x64
- .NET SDK: 10.0.204; target framework: net10.0
- PowerShell: 7.6.6
- Terminal.Gui NuGet package: 2.5.0-beta.1
- Package repository commit: `60ce99db27827da35b9661bb765f823304fd5982`
- Terminal: not required for this reproduction; it calls `NewKeyDownEvent` directly.

## Additional context

In the [package's ListView source](https://github.com/tui-cs/Terminal.Gui/blob/60ce99db27827da35b9661bb765f823304fd5982/Terminal.Gui/Views/ListView/ListView.cs), assigning `Source` initializes `KeystrokeNavigator.Collection` from `Source.ToList()`. `SourceOnCollectionChanged` updates layout/selection but does not replace that snapshot. `OnKeyDown` then assigns the returned index to `SelectedItem` without checking it against the current source count.

The same missing synchronization is present in the current `develop` source inspected when filing this report; the executable reproduction was tested against 2.5.0-beta.1 only.

This surfaced in a filtered list backed by a custom data source returning navigation strings. That source correctly raises a single `CollectionChanged(Reset)` after a bulk refresh. The standalone example above demonstrates that ordinary removal notifications also reproduce it, so batching is not required.

An application-side workaround is to assign `list.KeystrokeNavigator.Collection = list.Source.ToList()` after each refresh. The corresponding library fix would synchronize keyboard navigation when source contents change, while respecting a null/disabled navigator. An index bounds check alone would prevent the exception but would not fix incorrect navigation after a reorder.

Our application regression cases cover shrinking to one item, filtering to zero items, changing ranked order with the same item count, and restoring the original list. All pass with explicit navigator synchronization.

YourRobotOverlord:
## Summary

Refresh the ListView keystroke navigator whenever its data source raises CollectionChanged so snapshot and projection sources cannot become stale.

- Fixes #5683

## Changes

- Synchronize KeystrokeNavigator.Collection with Source.ToList() during collection changes.
- Add regression coverage for shrinking a snapshot source.
- Add regression coverage for same-count item reordering.

## Testing

- dotnet build --configuration Debug --no-restore: passed with 0 errors; existing dependency and XML-documentation warnings remain.
- Focused issue regression tests: 2 passed.
- ListViewTests: 110 passed.
- UnitTests.NonParallelizable: 32 passed.
- Full UnitTestsParallelizable run: 17,693 passed, 17 skipped, and 4 baseline ANSI-output failures. The same four failures reproduce on unmodified upstream/develop in this environment.

    git fetch copilot fix/5683-listview-keystroke-navigation
    git checkout copilot/fix/5683-listview-keystroke-navigation


BDisp:
## Deep review: concurrency and performance

**Conclusion: no new merge-blocking concern.** The PR fixes the correctness problem cleanly. Its main trade-off is intentional: for custom sources whose `ToList()` creates a projection or snapshot, every `CollectionChanged` notification now rebuilds that navigation collection. That can be costly for high-frequency, large-list updates, but it is required to guarantee correct type-to-search behavior under the existing `IListDataSource` contract.

### Concurrency

`CollectionNavigator` protects replacement and access to its `Collection` reference with `_collectionLock`:

```csharp name=CollectionNavigator.cs url=https://github.com/tui-cs/Terminal.Gui/blob/e8f636fc68cd69055f44cb95e6d9f21d730932a4/Terminal.Gui/Views/CollectionNavigation/CollectionNavigator.cs#L20-L55
public IList Collection
{
    get
    {
        lock (_collectionLock)
        {
            return _collection;
        }
    }
    set
    {
        lock (_collectionLock)
        {
            _collection = value;
        }
    }
}

protected override object ElementAt (int idx)
{
    lock (_collectionLock)
    {
        return Collection [idx];
    }
}

protected override int GetCollectionLength ()
{
    lock (_collectionLock)
    {
        return Collection.Count;
    }
}
```

That means assigning the fresh list is atomic at the reference level. However, it does **not** make a full navigation scan atomic: `CollectionNavigatorBase` obtains the count and reads individual items under separate lock acquisitions. If a background thread replaces the navigator collection between those operations, a scan could theoretically use an old count and index into a newer, shorter collection.

```csharp name=CollectionNavigatorBase.cs url=https://github.com/tui-cs/Terminal.Gui/blob/e8f636fc68cd69055f44cb95e6d9f21d730932a4/Terminal.Gui/Views/CollectionNavigation/CollectionNavigatorBase.cs#L164-L204
int collectionLength = GetCollectionLength ();

if (currentIndex.HasValue && currentIndex < collectionLength && Matcher.IsMatch (search, ElementAt (currentIndex.Value)))
{
    // ...
}

for (var i = 0; i < collectionLength; i++)
{
    if (Matcher.IsMatch (search, ElementAt (i)))
    {
        return i;
    }
}
```

This is pre-existing behavior, not introduced by this PR. It matters only if a data source raises `CollectionChanged` concurrently with keyboard processing—generally outside the expected single UI-thread model. `IListDataSource` does not promise thread safety, and `ObservableCollection<T>` should not be mutated concurrently with UI reads. I would **not** request a synchronization redesign in this small bug-fix PR.

### Performance

| Existing behavior | PR behavior |
|---|---|
| `CollectionChanged` recalculates size and selection state. | It additionally calls `Source.ToList()` and assigns the result to the navigator. |
| Default `ListWrapper<T>.ToList()` returns the backing collection without copying. | For the default source, the added cost is effectively a method call plus navigator assignment. |
| Custom sources may return a new snapshot/projection from `ToList()`. | Each notification may allocate and populate a new navigation list. |

```csharp name=ListView.cs url=https://github.com/tui-cs/Terminal.Gui/blob/e8f636fc68cd69055f44cb95e6d9f21d730932a4/Terminal.Gui/Views/ListView/ListView.cs#L214-L225
private void SourceOnCollectionChanged (object? sender, NotifyCollectionChangedEventArgs e)
{
    SetContentSize (new Size (EffectiveMaxItemLength, Source?.Count ?? Viewport.Height));
    KeystrokeNavigator?.Collection = Source?.ToList ();

    if (Source is { Count: > 0 } && SelectedItem.HasValue && SelectedItem > Source.Count - 1)
    {
        SelectedItem = Source.Count - 1;
    }

    SetNeedsDraw ();
    OnCollectionChanged (e);
}
```

For `ListWrapper<T>`, `ToList()` returns `_source` directly, so standard `ListView` usage avoids an O(*n*) copy:

```csharp name=ListWrapper.cs url=https://github.com/tui-cs/Terminal.Gui/blob/e8f636fc68cd69055f44cb95e6d9f21d730932a4/Terminal.Gui/Views/ListView/ListWrapper.cs#L101-L103
public IList ToList () => _source ?? [];
```

For a custom source like the added test fixture, however, it explicitly materializes a new list:

```csharp name=ListViewTests.cs url=https://github.com/tui-cs/Terminal.Gui/blob/e8f636fc68cd69055f44cb95e6d9f21d730932a4/Tests/UnitTestsParallelizable/Views/ListViewTests.cs#L935-L945
private sealed class SnapshotListDataSource : ListWrapper<string>, IListDataSource
{
    private readonly ObservableCollection<string> _source;

    public SnapshotListDataSource (ObservableCollection<string> source) : base (source)
    {
        _source = source;
    }

    IList IListDataSource.ToList () => _source.ToList ();
}
```

So, for a large filtered/projection-backed source that raises one event per item during bulk mutation, refresh work can become quadratic across the batch. That is not a regression in semantic intent—the PR must refresh the projection to fix the bug—but it makes batching important for custom high-volume sources. Such sources should suppress intermediate change events and emit a single final `Reset`/refresh notification when the projection is stable.

### Recommendation

**Approve as-is.** I would consider a non-blocking follow-up for documentation rather than changing the implementation:



- **`Terminal.Gui/Views/ListView/ListView.cs:217`** — Consider documenting that `IListDataSource.ToList()` is invoked after every delivered `CollectionChanged` event to refresh keystroke navigation. Custom sources that materialize snapshots or projections can then coalesce bulk updates into a final reset notification rather than paying for a full projection rebuild per item.

This is a performance-contract clarification, not a correctness defect. The PR’s actual implementation is the right fix for the stale navigation snapshot.

*Note: code-search results are capped and may not cover every navigator/data-source use site. You can inspect additional matches in the GitHub code search UI for `KeystrokeNavigator.Collection` within `tui-cs/Terminal.Gui`.*

`Tests/UnitTestsParallelizable/Views/ListViewTests.cs`  
Lines: 643-655

```text
Could we also cover the empty-source boundary here? After removing all items from the snapshot source, press a compatible key and assert that `NewKeyDownEvent` returns `false` without throwing and that the navigator collection is empty. This verifies the synchronization behavior at the lower bound and protects against stale entries when the source becomes empty.
```

### Why this matters

The current regression test verifies shrinking from three items to one, but an empty collection is a distinct boundary case. It confirms that the refreshed navigator does not retain stale items and that keyboard navigation safely handles a source with no searchable entries.

tig (approved):

